PR #11, opened 2016-01-28
EQ-26 - Implementing JWT

warrenbailey:
**What**
This pull request implements the following:
- JWT
- Signed JWT
- Encrypted JWT

I have created 3 different urls, one for each of the above. All urls expect a token parameter and parse this JWT token appropriately. If the Token is valid then the user is redirected to the main survey runner page. If the Token is missing an 401 Unauthorized challenge response is returned to the user. If the Token is invalid for any reason a 403 Forbidden response is returned to the user.

More details about the encryption process can be found here: https://digitaleq.atlassian.net/wiki/pages/viewpage.action?pageId=10780687

**How to test**
1. There are 4 Unit tests classes in  tests/app/authentication. These unit test perform the decryption process, one of them uses the example from RFC7516. All these tests should pass.
2. There is a test python script in the jwt-tests-keys directory. This script perform both the encryption and decryption process using a set of example public and private keys.
3. Finally use a browser or a REST client to send requests to the following url (note you need the survey runner application running locally):
- http://0.0.0.0:5000/jwt?token=XXXX
- http://0.0.0.0:5000/jwt_signed?token=XXXX
- http//0.0.0.0:5000/jwt_encrypted?token=XXXX

Where token=XXXX needs to be replaced with a real JWT token. To create a standard JWT  you can use http://jwt.io/. To create a Signed or Encrypted JWT use the test scripts provided in step 2.

**Who can review**


dhilton:
Can we prefix the environment variables with EQ_ ? Makes it easier to namespace the variables in the ENV as it can get a bit messy.

warrenbailey:
Done

dhilton:
Having tested this and reviewed the code more than happy to merge. Cracking stuff. :+1:

PR #20, opened 2016-02-09
EQ 69 Harden Portal Integration

warrenbailey:
**What**
Updates the JWT decoder and hardens the token handling. The token now must obtain all the rules in: https://digitaleq.atlassian.net/browse/EQ-69. The biggest changes are the inclusion of iat and exp in the payload and the inclusion of kid in the header.

There is now a production mode environment variable (EQ_PRODUCTION), when set to true all requests to the survey runner must set the url parameter 'token'. This should be an JWE encrypted signed JWT token.

The number of unit tests related to this functionality has been increase to cover all test cases in the story. Additional Flask unit tests have also been added.

**How to test**
1. Check out this branch
2. Run the unit tests by using the ./scripts/run_tests.sh script. All unit tests should pass.
3. Enable production mode by setting the EQ_PRODUCTION environment variable to true. Also set the public and private key variables:
- `export EQ_RRM_PUBLIC_KEY=cat ./jwt-test-keys/rrm-public.pem`
- `export EQ_SR_PRIVATE_KEY=cat ./jwt-test-keys/sr-private.pem`
1. Check that all survey runner urls are rejected with a 401 response.
2. Add a valid JWE token to the url and check the URLs now respond with a 200
3. Disable production mode by setting the EQ_PRODUCTION environment variable to false.
4. Check that all survey runner urls return a 200 response without a token
5. Check that all survey runner urls also accept a token parameter and that the token is used. (A token can be JWE encrypted signed JWT, an unencrypted signed JWT or an unencrypted unsigned JWT)

**Who can review**


dhilton:
Lgtm I'm :+1: on this but it'd be good to get others in the team to review.

.travis.yml
```diff
@@ -1,10 +1,17 @@
...
   - "3.5"
+before_install:
+  - sudo apt-get -qq update
+  - sudo apt-get install build-essential libssl-dev libffi-dev python-dev
+
 # command to install dependencies
 install:
   - pip install -r requirements.txt
   - pip install -r requirements_for_test.txt
+  - pip install cryptography==1.1.2 --upgrade --force
```
collisdigital: Is this still needed? We've determined that it wasn't the version that was causing the problem and requirements.txt should contain the correct version now (1.1.2), confirm that and remove this?

.travis.yml
```diff
@@ -1,10 +1,17 @@
 language: python
+sudo: required
+dist: trusty
 python:
   - "3.5"
+before_install:
+  - sudo apt-get -qq update
+  - sudo apt-get install build-essential libssl-dev libffi-dev python-dev
```
collisdigital: Was all this needed in the end, wasn't sure if you were trying different approaches to fix the initial cert load problem that is now resolved?

app/authentication/jwt_decoder.py
```diff
@@ -1,31 +1,63 @@
...
+        if not iat:
+            raise InvalidTokenException("Missing iat claim")
+        exp = token.get('exp')
+        if not exp:
+            raise InvalidTokenException("Missing exp claim")
+
+    def _check_payload(self, signed_token):
+        token = self.__to_str(signed_token)
+        try:
+            payload_data = token.split('.', 2)[1]
```
collisdigital: Consider making this more readable:

``` python
header_data, payload_data, sig_data = token.split('.', maxsplit=2)
```

app/authentication/jwt_decoder.py
```diff
@@ -37,22 +69,81 @@ def decode_jwt_token(self, token):
     def decode_signed_jwt_token(self, signed_token):
         try:
             if signed_token:
+                logging.debug("Decoding signed JWT " + self.__to_str(signed_token))
+                self._check_headers(signed_token)
+                self._check_payload(signed_token)
                 token = jwt.decode(signed_token, self.rrm_public_key, algorithms=['RS256'])
+                self._check_token(token)
+                if not token:
```
collisdigital, line 55: Looks like this check should be prior to calling _check_token()

app/authentication/jwt_decoder.py
```diff
@@ -37,22 +69,81 @@ def decode_jwt_token(self, token):
...
+        except (jwt.DecodeError,
+                jwt.exceptions.InvalidAlgorithmError,
+                jwt.exceptions.ExpiredSignatureError,
+                jwt.exceptions.InvalidIssuedAtError) as e:
             raise InvalidTokenException(repr(e))
-        except jwt.exceptions.InvalidAlgorithmError as e:
+
+    def _check_headers(self, signed_token):
+        token = self.__to_str(signed_token)
+        header_data = token.split('.', 1)[0]
```
collisdigital: Consider improving readability as per previous comment

app/authentication/jwt_decoder.py
```diff
@@ -37,22 +69,81 @@ def decode_jwt_token(self, token):
...
+        if "JWT" != header.get('typ').upper():
+            raise InvalidTokenException("Invalid Type")
+        if "RS256" != header.get('alg').upper():
+            raise InvalidTokenException("Invalid Algorithm")
+        if "EDCRRM" != header.get('kid').upper():
+            raise InvalidTokenException("Invalid Key Identifier")
+
+    def __check_for_duplicates(self, headers):
+        h = self.__to_str(headers)
+        if h.count("typ") > 1:
```
collisdigital: There is a risk with this approach that a header containing this value not as a 'key' will fail such as {"key1":"type1"} would hit this trigger because the value contains 'typ'. Something to consider.

app/authentication/jwt_decoder.py
```diff
@@ -67,16 +162,33 @@ def decrypt_jwt_token(self, token):
                 return self.decode_signed_jwt_token(signed_token)
             else:
                 raise NoTokenException("JWT Missing")
-        except jwt.DecodeError as e:
+        except (jwt.DecodeError, InvalidTag, InternalError, ValueError) as e:
             raise InvalidTokenException(repr(e))
 
+    def __check_jwe_protected_header(self, header):
+        header = self._base64_decode(header).decode()
+        if header.count("alg") == 0:
```
collisdigital: as per previous comment, someone putting 'alg' in value element would 'pass' this check but wouldn't be correct (i.e. could still be missing an 'alg' key in the header)

app/authentication/jwt_decoder.py
```diff
@@ -67,16 +162,33 @@ def decrypt_jwt_token(self, token):
...
+            raise InvalidTokenException("Missing Algorithm")
+        if header.count("RSA-OAEP") == 0:
+            raise InvalidTokenException("Invalid Algorithm")
+        if header.count("enc") == 0:
+            raise InvalidTokenException("Missing Encoding")
+        if header.count("A256GCM") == 0:
+            raise InvalidTokenException("Invalid Encoding")
+
+    def _check_iv_length(self, iv):
+        return len(iv) == 12
```
collisdigital: Really these magic numbers should be named as constants, probably at the top of the class definition e.g. IV_EXPECTED_LENGTH = 12

app/main/views/root.py
```diff
@@ -6,19 +6,34 @@
...
-with open(settings.EQ_SR_PRIVATE_KEY, "rb") as private_key_file:
-    sr_private_key = private_key_file.read()
-
-decoder = Decoder(rmm_public_key, sr_private_key, "digitaleq")
+@main.before_request
+def jwt():
+    if settings.EQ_PRODUCTION:
+        return jwt_decrypt()
+    else:
+        # DEVELOPER mode
```
collisdigital: Keep it clean, factor this else body into a separate function and call it here?

.travis.yml
warrenbailey: Just testing it without this now

.travis.yml
warrenbailey: As above

app/main/views/root.py
```diff
@@ -6,19 +6,34 @@
...
+        logging.warning("Developer mode")
+        if request.args.get('token'):
```
collisdigital: 'token' is used in quite a few places in this file, consider factoring it out into a constant e.g. EQ_URL_QUERY_STRING_JWT_FIELD_NAME

app/settings.py
```diff
@@ -1,7 +1,9 @@
 import os
 
-EQ_RRM_PUBLIC_KEY = os.getenv('EQ_RRM_PUBLIC_KEY', './jwt-test-keys/rrm-public.pem')
-EQ_SR_PRIVATE_KEY = os.getenv('EQ_SR_PRIVATE_KEY', './jwt-test-keys/sr-private.pem')
+
 EQ_RABBITMQ_URL = os.getenv('EQ_RABBITMQ_URL', 'amqp://admin:admin@localhost:5672/%2F')
 EQ_RABBITMQ_QUEUE_NAME = os.getenv('EQ_RABBITMQ_QUEUE_NAME', 'eq-submissions')
 EQ_RABBITMQ_TEST_QUEUE_NAME = os.getenv('EQ_RABBITMQ_TEST_QUEUE_NAME', 'eq-test')
+EQ_PRODUCTION = os.getenv("EQ_PRODUCTION", False)
```
collisdigital: Maybe this should default to True so we have to explicitly set to False to avoid accidentally pushing a debug build to Prod?

application.py
```diff
@@ -2,8 +2,15 @@
 
 import os
 from app import create_app
+from app import settings
+
 from flask.ext.script import Manager, Server
 
+if not settings.EQ_PRODUCTION:
+    with open(os.getcwd() + '/jwt-test-keys/rrm-public.pem', "r") as public_key_file:
```
collisdigital: Remove and keep only one method of loading keys (i.e. set the env value) ~ do it outside prior to running locally

tests/app/authentication/__init__.py
```diff
@@ -0,0 +1,82 @@
+SR_PUBLIC_PEM = """-----BEGIN PUBLIC KEY-----
```
collisdigital: Consider putting TEST_DO_NOT_USE or similar in the name so it's clear these mustn't be used in production.

tests/app/authentication/encoder.py
```diff
@@ -0,0 +1,78 @@
...
+from cryptography.hazmat.primitives.ciphers import Cipher, algorithms, modes
+from cryptography.hazmat.primitives import hashes
+from cryptography.hazmat.primitives.asymmetric import padding
+from tests.app.authentication import SR_PUBLIC_PEM, RRM_PRIVATE_PEM
+
+import jwt
+import os
+import base64
+
+backend = default_backend()
```
collisdigital: Switch to OpenSSL backend?

tests/app/authentication/encoder.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+class Encoder (object):
+    def __init__(self):
+        private_key = self.__to_bytes(RRM_PRIVATE_PEM)
+        public_key = self.__to_bytes(SR_PUBLIC_PEM)
+        self.rrm_privatekey = serialization.load_pem_private_key(private_key, password=b'digitaleq', backend=backend)
+        self.sr_publickey = serialization.load_pem_public_key(public_key, backend=default_backend())
```
collisdigital: Remove duplicate call to default_backend() from above, and switch to OpenSSL at same time?

.travis.yml
warrenbailey: They aren't needed I've removed from the travis file.

tests/app/authentication/test_jwe.py
```diff
@@ -0,0 +1,172 @@
...
+            "vPAhHsZEiGLQiKsy8WSoi8NC-CNs6FTqYrqcB4V77umi5XRU_JcdE_mqxcMkIHYUsUBDCqSMSA8xQgD56OOo41K4-vFMh99fU1u5zSbdq" \
+            "hqLtNeYTepJkOnCFi5aSjFPrtpfbvk9YPDJERJzEXAJkYbkUplmNudtObZY2IDmZglC5waGsTOq62tpudGPNb5B_FEq0s8ezAtMtf-C3h" \
+            "BslOcCLtAlGdRsptS6-L4YHxaGEAvwXD1heQADoAabjGCA-BZGPdc6h8A6y1QR7m2P33KEz83h141X4Ftyd9kTGxgjQ8-UrIGvu9b1wbm" \
+            "cjTWK3hiiyC3o0guBbzBq0SflX-RG9_zyK4sAIksiXj9mNRah1IUjUT5zr9cs1c5i4WwJdpJ28IzKvczgQ63B_t1xEpGsb6A35EhxzRkG" \
+            "7BQ.I9WExvUvZVXdwjxz8EPtCg"
+
+
+class JWETest(unittest.TestCase):
+    def test_valid_jwe(self):
+        decoder = Decoder(RRM_PUBLIC_PEM, SR_PRIVATE_PEM, "digitaleq")
```
collisdigital: Consider moving the password into the test.app.authentication.**init**.py as well?

app/authentication/jwt_decoder.py
warrenbailey: Done

app/authentication/jwt_decoder.py
warrenbailey, line 55: I've removed this check and made it part of _check_payload()

tests/app/authentication/test_jwt.py
```diff
@@ -0,0 +1,285 @@
+import unittest
+from tests.app.authentication import SR_PRIVATE_PEM
+from app.authentication.jwt_decoder import Decoder
+from app.authentication.invalid_token_exception import InvalidTokenException
+import base64
```
collisdigital, line 2: Generally speaking Python library imports normally come first before app specific ones (thought this was a PEP8/Lint thing?) I could be wrong...

tests/app/authentication/test_jwt.py
```diff
@@ -0,0 +1,285 @@
...
+    def test_jwt_does_not_contain_header(self):
+        token_without_header = "e30." + jwtio_payload + "." + jwtio_signature
+
+        decoder = Decoder(PUBLIC_KEY, SR_PRIVATE_PEM, "digitaleq")
+        with self.assertRaises(InvalidTokenException) as ite:
+            decoder.decode_signed_jwt_token(token_without_header)
+        self.assertIn("Missing Headers", ite.exception.value)
+
+    def test_jwt_does_not_contain_payload(self):
+        token_without_payload = jwtio_header + ".e30." + jwtio_signature
```
collisdigital, line 59: Is there a test that has an empty payload e.g. jwtio_header + ".." + jwtio_signature and also for header and sig?

tests/app/authentication/test_jwt.py
```diff
@@ -0,0 +1,285 @@
...
+        header = base64.urlsafe_b64encode(b'{"kid":"EDCRRM","typ":"JWT"}')
+        jwt = header.decode() + "." + jwtio_payload + "." + jwtio_signature
+
+        decoder = Decoder(PUBLIC_KEY, SR_PRIVATE_PEM, "digitaleq")
+        with self.assertRaises(InvalidTokenException) as ite:
+            decoder.decode_signed_jwt_token(jwt)
+        self.assertIn("Missing Algorithm", ite.exception.value)
+
+    def test_jose_header_invalid_alg(self):
+        header = base64.urlsafe_b64encode(b'{"alg":"invalid","kid":"EDCRRM","typ":"JWT"}')
```
collisdigital, line 117: do we have a test for 'alg':'None', just because this was a known issue with JWT in the past

tests/app/authentication/test_jwt_decoder.py
```diff
@@ -46,29 +46,26 @@
...
-                "Z0hI3.ar6axH5GB6L1K8UPNqQoJiLhwG0dZGL0Mgp_1fadQxIJXtBTmR2ocitgnzH5JxNJ" \
-                "1Z7vIxWFLU5IsEIjViUH97lXx_PjYGzW6w4ER4Tw8YXus-2wuTJUvaKnKnrZz_1TT9kn_s" \
-                "HgGmQdBfOktA-sGh_Kys9edAq7YtrnRCz8EanxWcEHGCqYeq-qT49ejpWNBJuVfaL7gH7d" \
-                "-l7XDlK-4L_RRYuf1_VhoQSCxhJyoZcTeFfrukDlQSWMmJlAEtrpBwKJe1JYSTEQVGm6Hg" \
-                "_x9nd9N_lBOmLHk8KUFCN02_Qg4PPnlTImDtbI-lmBoJroyIBC7wnQyDtJlx6rwdjKWGG6" \
-                "OCqx_1IuC1KAoSGip4SDOfnVvncennqcF6rprBWnTcOmFAl1RPZ8VoA81fZXREBCEDp91t" \
-                "ySEfPRkl1nJsWW2Mi6YYLTfsqn-lUU-W9smCRtKYTzVxwGXF2jIU-KK1c0FDJIrbzCuX-N" \
-                "2cTG8jEw-kTLf_DA11sL32YF_S_0Lpw.AN8XkD0BIi3anityXzbmhw"
+
+jwt = "eyJhbGciOiJub25lIn0.eyJleHAiOiIyMDc1Mjk3MTQ4IiwiaWF0IjoiMTQ1NDkzNzAyMyIsInVzZXIiOiJqaW1teSJ9."
```
collisdigital: At the top of this file the keys are in b'str\n' format, have we settled on triple quote with a byte conversion?

tests/app/authentication/test_jwt_decoder_using_rfc_example.py
```diff
@@ -87,7 +87,7 @@ def test_decrypt(self):
         cipher_text = decoder._base64_decode(tokens[3])
         tag = decoder._base64_decode(tokens[4])
 
-        decrypted_token = decoder._decrypt_cipher_text(cipher_text, iv, cek, tag, jwe_protected_header)
+        decrypted_token = decoder._decrypt_cipher_text(cipher_text, bytes(iv), bytes(cek), bytes(tag), jwe_protected_header)
```
collisdigital: Could be clearer to put the bytes() wrapping around the cek/iv when they are declared above rather than here. Also the tag doesn't need this conversion, but belt and braces you could wrap the decoder.__base64_decode call rather than here as well?

tests/app/views/test_jwt_authentication.py
```diff
@@ -0,0 +1,78 @@
...
+from app import create_app
+from tests.app.authentication.encoder import Encoder
+
+
+class FlaskClientAuthenticationTestCase(unittest.TestCase):
+
+    def setUp(self):
+        logging.basicConfig(level=logging.DEBUG)
+
+        with open(os.getcwd() + '/jwt-test-keys/rrm-public.pem', "r") as public_key_file:
```
collisdigital: Are these used/needed (see previous comments), move them to **init**.py alongside the others?

tests/conftest.py
```diff
@@ -1,11 +1,18 @@
...
 import mock
+import os
 from app import create_app
 from config import configs
+from app import settings
 
 
 @pytest.fixture(scope='session')
 def survey_runner(request):
+    with open(os.getcwd() + '/jwt-test-keys/rrm-public.pem', "rb") as public_key_file:
```
collisdigital: See previous on this

tests/conftest.py
```diff
@@ -1,11 +1,18 @@
 import pytest
 import mock
+import os
 from app import create_app
 from config import configs
+from app import settings
 
 
 @pytest.fixture(scope='session')
 def survey_runner(request):
```
collisdigital, line 11: Not sure about this file, is it needed etc?

app/authentication/jwt_decoder.py
warrenbailey: A very good point and that could lead to interesting bugs. I had to do this check as the library uses json loads underneath and that doesn't allow duplicate keys, only the second value is kept.

I'll have a rethink on this then.

app/authentication/jwt_decoder.py
warrenbailey: done

app/main/views/root.py
warrenbailey: Done

app/main/views/root.py
warrenbailey: Done

app/settings.py
warrenbailey: Done

application.py
warrenbailey: DONE

tests/app/authentication/__init__.py
warrenbailey: Done

tests/app/authentication/encoder.py
warrenbailey: done

tests/app/authentication/test_jwt.py
warrenbailey, line 2: Done

tests/app/authentication/test_jwe.py
warrenbailey: Done

tests/app/authentication/test_jwt.py
warrenbailey, line 59: Done

tests/app/authentication/test_jwt.py
warrenbailey, line 117: Done

tests/app/authentication/test_jwt_decoder.py
warrenbailey: Done

tests/app/authentication/test_jwt_decoder_using_rfc_example.py
warrenbailey: Done

tests/app/views/test_jwt_authentication.py
warrenbailey: Done

tests/conftest.py
warrenbailey: Done

tests/conftest.py
warrenbailey, line 11: Question for @dhilton

warrenbailey:
I've got a few more comments to do, don't merge yet :-1:

warrenbailey:
Ready for review :+1:

warrenbailey:
Callback is only called once :+1:

dhilton:
LGTM :+1:

PR #26, opened 2016-02-15
Fix git revision elasticbeanstalk

dhilton:
**What**

The change merged in 94315c9e5404798fb6d445a3ee21c934a41fa0d7 to add a git ref to the healthcheck, fails when deployed to elasticbeanstalk if the file doesn't exist. This change moves the file based approach to an environment variable - populating the elasticbeanstalk environment with the correct git ref for HEAD revision

This PR also fixes an issue when running the survey runner in development mode, whereby the check for  KEYMAT was always being run even if the developer explicitly switched off JWT support via the EQ_PRODUCTION feature flag. This now seems to behave.

**How to test**
1. Checkout this branch
2. With an existing elasticbeanstalk environment run: `eb init` 
3. Select `eu-west` region and your correct application 
4. Run `eb setenv EQ_PRODUCTION=false`
5. Run `eb setenv EQ_GIT_REF=1234567`
6. Run `eb deploy`

Wait till the application is deployed and then check the resulting healthcheck to ensure that the correct values are given for git ref and that you can access the pattern library without having to have a JWT in the url.

**Who can test**

Anyone but @dhilton

### DIFF
--- a/scripts/run_app.sh
+++ b/scripts/run_app.sh
@@ -26,7 +26,7 @@ fi
 echo "Environment variables in use:"
 env | grep EQ_
 
-#npm install
-#npm run compile
+npm install
+npm run compile
 
 python application.py runserver
--- a/scripts/run_app.sh
+++ b/scripts/run_app.sh
@@ -17,14 +17,16 @@ if [ -z "$EQ_SR_PRIVATE_KEY" ]; then
 fi
 
 # Output the current git revision
-git rev-parse HEAD > revision.log
+if [ -z "$EQ_GIT_REF" ]; then
+  export EQ_GIT_REF=`git rev-parse HEAD`
+fi
 
 # Use default environment vars for localhost if not already set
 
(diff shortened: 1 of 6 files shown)


scripts/run_app.sh
```diff
@@ -17,14 +17,16 @@ if [ -z "$EQ_SR_PRIVATE_KEY" ]; then
...
 
 # Use default environment vars for localhost if not already set
 
 echo "Environment variables in use:"
 env | grep EQ_
 
-npm install
-npm run compile
+#npm install
+#npm run compile
```
warrenbailey: I don't think this should be commented out.

warrenbailey:
Tested and this worked for me. In hindsight environment variables make much more sense. I'm just going to uncomment those npm commands in the run_app.sh and then I'll merge.

warrenbailey:
Looks good to me :+1:

PR #29, opened 2016-02-19
EQ 13 New Relic integration

warrenbailey:
**What**
This change install the New Relic agent programmatically. We need to do it this way as we're deploying into Elastic Beanstalk.

**How to test**
1. Create a new relic account
2. Follow the instructions on how to generate a newrelic.ini file - [New Relic](https://docs.newrelic.com/docs/agents/python-agent/getting-started/python-agent-quick-start)
3. Move the file to the top level directory of this repository (or point the EQ_NEW_RELIC_CONFIG_FILE to the location of that file).
4. Run the server
5. Browse to the [health check](http://localhost:5000/healthcheck)
6. Check new relic is receiving information.

**Who can review**
Anyone apart from @warrenmethods


dhilton:
LGTM - Do you think we'll ever add the newrelic.ini file to source control? If so is it worth adding it to gitignore, but including it in ebignore - so that a local ini file is deployed but never added to git?

warrenbailey:
The newrelic.ini file contains the license key, so it shouldn't be in GIT. I can add it to the .gitignore. I think we'll have to manually copy the file to the Jenkins box for deployment.

dhilton:
That works for me - just need to make sure we add it to .ebignore with a ! so that it is included when you run `eb deploy`

collisdigital:
@warrenbailey @dhilton could we put this key/value into VAULT later?

dhilton:
There is more than just a K/V in the settings file - we possibly could extract the license ID / key and just retain the settings - would need to look if the newrelic python lib supports environment variables.

PR #35, opened 2016-02-29
Update static url path

warrenbailey:
**What**
Fix for static files missing on Elastic beanstalk.

**How to test**
1. Deploy via terraform (changing the branch in deploy_survey_runner.sh to match this branch).
2. Check the static files are present

**Who can review**
Anyone apart from @weapdiv-david and @warrenbailey


hamishtaplin:
This will need a change in the CSS as all the paths are hardcoded. Will do it now.

PR #42, opened 2016-03-03
JWT flask session integration

warrenbailey:
**What**
Store the JWT token in the flask session, once stored the user will not need to provide the token again until the session timeouts. There are two new environment variables, one for setting the Flask secret key (used to sign the cookies) and one for the session timeout (defaulted to 30 minutes).

**How to test**
1. Checkout this branch
2. Set the following env variable `export EQ_PRODUCTION=True`
2. Browse to http://localhost:5000/session and check that you are asked to login
3. Navigate to http://localhost:5000/questionnaire/1 and check you are not allow to see the page
4. Browse to localhost:5000/session?token=<valid-jwt> and check that the home page is displayed
5. Navigate to the http://localhost:5000/questionnaire/1 and check it is displayed correctly
6. Wait until the session times out (you may wish to reduce the session timeout value)
7. Check that you can no longer navigate to #http://localhost:5000/questionnaire/1

**Who can review**


warrenbailey:
:-1:  Needs to change to middleware now that the main branch has changed to use multiple flask applications

warrenbailey:
Middleware doesn't do what we need it to do. I've decided to use the Flask Login extension instead, it makes the most sense and does what we need it to do. I'll update the description was I finish the final touches.

warrenbailey:
Ready for review

dhilton:
This LGTM - run all the tests described and checked the access control.

PR #43, opened 2016-03-04
Interfaces and Questionnaire Route

warrenbailey:
**What**
Adding the new interfaces (excluding the schema parse) and setting up the main questionnaire route

**How to test**
1. Checkout this branch
2. Run the server
3. Browse to http://localhost:5000/questionnaire/1 and check the base layout is rendered correctly.

**Who can review**

### DIFF
--- a/app/main/views/root.py
+++ b/app/main/views/root.py
@@ -84,6 +84,12 @@ def root():
     return render_template('index.html')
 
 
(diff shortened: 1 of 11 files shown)


LJBabbage:
If we are using these as interfaces, lets have a convention, "I" infront of each class seems logical, so someone does accidentally add functionality

weapdiv-david:
LGTM :+1:

PR #44, opened 2016-03-04
eq-parser initial MCI schema

LJBabbage:
**What**
First version of the json needed for MCI

**How to test**
Please check the format

__Who_
Anyone but @LJBabbage

### DIFF
--- a/json/mci.json
+++ b/json/mci.json
@@ -27,7 +27,7 @@
                                       "id": "29586b4c-fb0c-4755-b67d-b3cd398cb30a",
                                       "label": "Male employees working more than 30 hours per week?",
                                       "guidance": "How many men work for your company?",
-                                      "type": "Numeric"
+                                      "type": "Integer"
                                     }
                                 ]
                               }
--- a/json/mci.json
+++ b/json/mci.json
@@ -1,0 +1,41 @@
+{
+    "schema_version": "0.0.1",
+    "questionnaire_id": "22",
+    "survey_id": "23",
+    "title": "Monthly Business Survey - Retail Sales Index",
+    "description": "MCI description",
+    "groups": [
+        {
+            "id": "14ba4707-321d-441d-8d21-b8367366e766",
+            "title": "",
+            "blocks": [
+                {
+                    "id": "cd3b74d1-b687-4051-9634-a8f9ce10a27d",
+                    "title": "Monthly Business Survey",
+                    "sections": [
+                        {
+                          "id": "2cd99c83-186d-493a-a16d-17cb3c8bd302",
+                          "title":"",
+                          "questions": [
+                              {
+                                "id": "4ba2ec8a-582f-4985-b4ed-20355deba55a",
+                                "q_code": "110",
+                                "title": "On 12 January 2016 what was the number of employees for the business named above?",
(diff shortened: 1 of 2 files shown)


hamishtaplin:
For example: https://github.com/ONSdigital/eq-survey-runner/blob/master/app/main/data.json#L98

warrenbailey:
I though that the response in this schema is equivalent to your field?

hamishtaplin:
You're right, I confused myself there.

LGTM :+1:

json/mci.json
```diff
@@ -0,0 +1,41 @@
...
+                    "id": "cd3b74d1-b687-4051-9634-a8f9ce10a27d",
+                    "title": "Monthly Business Survey",
+                    "sections": [
+                        {
+                          "id": "2cd99c83-186d-493a-a16d-17cb3c8bd302",
+                          "title":"",
+                          "questions": [
+                              {
+                                "id": "4ba2ec8a-582f-4985-b4ed-20355deba55a",
+                                "q_code": "110",
```
dhilton, line 22: What's q_code and where has it come from?

json/mci.json
```diff
@@ -0,0 +1,41 @@
...
+                                "title": "On 12 January 2016 what was the number of employees for the business named above?",
+                                "description": "An employee is anyone aged 16 years or over that your organisation directly pays from its payroll(s), in return for carrying out a full-time or part-time job or being on a training scheme.",
+                                "responses": [
+                                    {
+                                      "id": "29586b4c-fb0c-4755-b67d-b3cd398cb30a",
+                                      "label": "Male employees working more than 30 hours per week?",
+                                      "guidance": "How many men work for your company?",
+                                      "type": "Numeric"
```
dhilton: Is numeric an integer, a float, what does this define as a numeric?

hamishtaplin:
Actually no, I do think fields are still relevant as they do have attributes that are possibly author-able such as validation messages or stuff that restricts user input, etc

json/mci.json
weapdiv-david, line 22: q_code came from the schema we discussed the other day in slack.  It's essentially the stat var identifier

json/mci.json
weapdiv-david: agree this needs to be clearer.  Lewis is changing it to an 'integer'

weapdiv-david:
@hamishtaplin there is no longer a one-to-one mapping between questionnaire responses and the html fields used to collect them.  Where a particular html widget needs additional authorable elements supplied to the html widget, it makes sense if we design and name a particular response type which can then contain the neccessary fields.

For simple response types (such as integer) there are attributes of the response that are implied by the type (e.g. non zero, whole number, numeric characters only allowed) that won't need additional labels displayed, but will effect the input control created in the front end.

For things like the date range "type", there will be additional fields, and perhaps the notion of conditional display that we discussed the other day, but we can cross that bridge when we get to it and will obviously need to have a discussion with you.

dhilton:
Can we change the encompassing folder to be called surveys - json doesn't mean anything apart from the format.

dhilton:
@hamishtaplin @weapdiv-david - I think you're talking at cross purposes, the field is still relevant at a schema model level, it's just not implemented here as it's a presentational decision about how a date field for example is presented. The form question then reassembles those field values into a single answer that is mapped to a statistical variable for downstream processing.

PR #49, opened 2016-03-08
Eq 1 add parser documentation

weapdiv-david:
**_What**_

Adds docstrings to parser code

**_How to test**_

1) Checkout branch
2) View parser code

**_Who can test**_

Anybody except @weapdiv-david

### DIFF
--- a/app/parser/parser_utils.py
+++ b/app/parser/parser_utils.py
@@ -1,42 +1,134 @@
+# -*- coding: utf-8 -*-
+'''Parser Utils Module
+
+This module defines a ParserUtils class with several static methods to be used
+by parser implementations.
+
+'''
+
 from .schema_parser_exception import SchemaParserException
 
 
 class ParserUtils(object):
-    # gets the value associated with key in the dict, or throws an exception if it's not present
+    '''ParserUtils
+
+    The ParserUtils class exposes a number of static utility methods that are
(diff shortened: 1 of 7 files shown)


warrenbailey:
I believe the standard python documentation is to document arguments and return values using :param: and :return:. That what my IDE auto fills anyway. I think we need to agree a standard and then either change this PR or what I've already committed to the code base.

e.g. 

"""Replaces template placeholder with values.

:param timestamp: formatted date to display
:param priority: priority number
:param priority_name: priority name
:param message: message to display
:returns: formatted string
"""

http://stackoverflow.com/questions/5334531/python-documentation-standard-for-docstring

app/parser/parser_utils.py
```diff
@@ -1,42 +1,134 @@
...
+    '''
+
     @staticmethod
     def get_required(obj, key):
+        '''Get a required value from the dict.
+
+        Gets the value associated with key in the dict, or raises a
+        SchemaParserException if it's not present.
+
+        Args:
```
warrenbailey: should be 
:param obj: A dict on json object
:param key: The name of the property to retrieve
:returns: the value if found

warrenbailey:
Lets have a discussion when you get in, needs rebasing too :-1: for now

warrenbailey:
LGTM :+1:

PR #51, opened 2016-03-08
Response Guidance

hamishtaplin:
This PR adds basic guidance to responses.
### Changes
- added guidance data to schema for various questions
- added section styling
- added guidance styling
- added guidance expand interaction
- re-worked typed inputs (eg. currency, percentage)
- moved rendering if `input`, `label` and `textarea` into macros
### Testing
- `npm run compile`
- Visit `/questionnaire/mci/`
- Check guidance works by clicking "show guidance for this question"


warrenbailey:
I think we need to remove the /questionnaire/mci route and just use /questionnaire/<questionnaire_id> going forward, as we want to integrate early. I've check them both and they are working so happy to merge this. I'll remove the mci route in another change set.

warrenbailey:
LGTM :+1:

PR #52, opened 2016-03-09
Eq model validation

LJBabbage:
**What**
Initial validation set up. n.b exceptions and full documentation to follow

**How to test**
1. Check the code and design patterns
2. Run the test_integer.py

**Who can test**
Anyone but @LJBabbage


app/model/block.py
```diff
@@ -3,7 +3,20 @@ def __init__(self):
...
         self.title = None
         self.sections = []
+        self.container = None
 
     def add_section(self, section):
         if section not in self.sections:
             self.sections.append(section)
+            section.container = self
+
+    def get_item_by_id(self, id):
```
collisdigital, line 15: Since this function is identical in all instances below except for the parameter being iterated it is a prime candidate for refactoring into a utility function and only being implemented/tested once.

collisdigital, line 15: If all 'elements' here have a 'get_item_by_id' that could possibly be an interface (or just use Python as-is to assume it exists in said utility function...)

collisdigital, line 15: Consider renaming this to 'find' rather than 'get' to emphasise the fact that this is how it is operating.

app/model/block.py
```diff
@@ -3,7 +3,20 @@ def __init__(self):
...
+        if id == self.id:
+            return self
+        else:
+            item = None
+            for section in self.sections:
+                candidate = section.get_item_by_id(id)
+                if candidate is not None:
+                    item = candidate
```
collisdigital: Return or break loop here? No need to keep iterating...

app/model/block.py
weapdiv-david, line 15: Agreed on all of that.  It makes sense to move this functionality into a single location that can be shared between all the model instances, and I thought as much when I wrote it.  Not sure yet whether to move it into a BaseModel class, have it as a Mixin, or move it out into another class yet.  There are more of these types of methods that will be added to model classes, so we're currently mulling over the best approach.

app/model/block.py
weapdiv-david: Agreed, will add the break

tests/a11y/index.html
```diff
@@ -0,0 +1,409 @@
+<!DOCTYPE html>
```
warrenbailey: Did you mean to check this file in?

tests/app/validation/test_integer.py
```diff
@@ -0,0 +1,169 @@
...
+        self.assertEquals(len(result.errors), 0)
+
+        # <space>
+        result = integer.validate(' ')
+        self.assertFalse(result.is_valid)
+        self.assertEquals(len(result.errors), 1)
+        self.assertEquals('\' \' is not an integer', result.errors[0])
+
+    # def test_integer_required_fail(self):
+    #
```
warrenbailey: Can we remove this code?

app/validation/integer.py
```diff
@@ -0,0 +1,22 @@
+from app.validation.i_validator import IValidator
+from app.validation.validation_result import ValidationResult
+
+
+class Integer(IValidator):
```
warrenbailey: I really think this should be called IntegerValidator or IntegerValidationRule or something, as Integer clashes with the Type Integer.

app/validation/required.py
```diff
@@ -0,0 +1,22 @@
+from app.validation.i_validator import IValidator
+from app.validation.validation_result import ValidationResult
+
+
+class Required(IValidator):
+
```
warrenbailey: For consistency, this probably should be named the same as whatever you change the Integer validator to.

app/model/questionnaire.py
```diff
@@ -5,8 +9,24 @@ def __init__(self):
...
+        self.validation = None
 
     def add_group(self, group):
         if group not in self.groups:
             self.groups.append(group)
-            group.questionnaire = self
+            group.container = self
+            self.register_item(group.id, group)
+
+    def register_item(self, item_id, item):
```
warrenbailey: is this method private? should it be _register_item?

app/model/questionnaire.py
```diff
@@ -5,8 +9,24 @@ def __init__(self):
...
+
+        if item_id in self.item_register.keys():
+            raise QuestionnaireException('Duplicate id found - ' + item_id)
+
+        self.item_register[item_id] = item
+        item.questionnaire = self
```
warrenbailey: I actually wonder if this method should be on a different class, maybe an 'ItemRegister' class

app/validation/validation_store.py
```diff
@@ -24,10 +26,27 @@ def get_result(self, key):
...
-        result.from_dict(session[key])
-        return result
+        if key in session[RESULTS].keys():
+            result = ValidationResult()
+            result.from_dict(session[RESULTS][key])
+            return result
+        return None
+
+
+class MockValidationStore(IValidationStore):
```
warrenbailey: I don't think this class should be in this file. It should be in the test file, then it won't get upload to elastic beanstalk

app/validation/validator.py
```diff
@@ -1,6 +1,98 @@
...
+        self._type_validator_factory_class = TypeValidatorFactory
+
+    def validate(self, user_data):
+        for item_id in user_data.keys():
+            logger.debug('Attempting to validate {}'.format(item_id))
+            item = self._schema.get_item_by_id(item_id)
+
+            # Currently the datefields trip this up
+            if not item:
+                logger.info('Ignoring {}'.format(item_id))
```
warrenbailey, line 30: Should this be at debug level?

tests/app/validation/test_integer.py
warrenbailey: Comment still stands

app/validation/required.py
warrenbailey: Comment still stands

app/validation/integer.py
warrenbailey: Comment still stands

app/validation/i_validator.py
```diff
@@ -0,0 +1,8 @@
+from abc import ABCMeta, abstractmethod
+
+
```
warrenbailey: I don't like the file name for this class, none of the other "interface" classes have i_ in the file name.

app/validation/validator.py
```diff
@@ -1,6 +1,98 @@
...
+import logging
+
+
+logger = logging.getLogger(__name__)
+
+
+class ValidationException(Exception):
+    pass
+
+
```
warrenbailey, line 13: I wonder if this "component" should be called ValidationEngine rather than Validator, then the file renamed validation.py and the i_validator.py renamed to validator.py. Ties in with the RoutingEngine naming convention then. What do you think?

warrenbailey:
A few minor comments, mainly around naming conventions. It's worth us all agreeing on it before merge. Overall happy with this now, so once those are address lets get it merged.

app/validation/validator.py
weapdiv-david, line 30: I've left it at info for now as it highlights the issue we have around date handling.  Ultimately, we shouldn't be ignoring anything without good reason

app/validation/validator.py
weapdiv-david, line 13: Haven't renamed it, but I have no strong convictions on this either way.  When this component was originally discussed we used the terms Validator and ValidationEngine interchangeably.

warrenbailey:
LGTM now :+1:

PR #54, opened 2016-03-09
EQ 107 JWT Generator

warrenbailey:
**What**
Removing the EQ_PRODUCTION flag as it's causing us issues. The application will now always be in production mode and required an encrypted JWT to access it. This pull request also provides a convenient way of generating that token.

**How to test**
1. Check out this branch
2. Run the run_app.sh script
3. It should output a URL with a valid token
4. Browse to that URL
5. You should be logged in and redirected to /questionnaire/1

**Who can review**

### DIFF
--- a/.ebignore
+++ b/.ebignore
@@ -25,4 +25,4 @@ tests/coverage
 !newrelic.ini
 
 # Token generation file
-token.py
+token_generator.py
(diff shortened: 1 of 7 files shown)


weapdiv-david:
LGTM :+1:  Have rebased and pushed, and will merge once tests pass

PR #58, opened 2016-03-10
Revert to read key material from file

warrenbailey:
**What**
Elastic Beanstalk does not like keys in environment variables. Temporarily revert back to loading key material from file with the environment variable point to the location of file on disk.

**How to test**
1. Check out this branch
2. start the server
3. make sure you can hit the session url

**Who can review**

### DIFF
--- a/app/settings.py
+++ b/app/settings.py
@@ -1,4 +1,7 @@
 import os
+import logging
+
+logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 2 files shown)


weapdiv-david:
LGTM :+1:  

We should revisit this after the key management stuff is resolved, but there's no point doing another temporary fix in the meantime as there are other areas we need to focus on.

PR #59, opened 2016-03-10
Eq 1 questionnaire manager

weapdiv-david:
**_What**_

Implements the QuestionnaireManager class and the interaction between main components needed to render a questionnaire and receive the users submission back.  Uses a stub Validator object but stores responses and validation results in the session.  Provides base empty implementations for all other components not currently needed for eq-1.

**_How to test**_

1) Check out the branch
2) Run the application
3) Login and view the questionnaire at `http://localhost:5000/questionnaire/1`
4) Enter some data and click one of the submit buttons.
5) Verify the post redirects you to the same page as a get request

**_Who can test**_

Anybody except @warrenbailey or @weapdiv-david, although probably @LJBabbage should look at this

**_Future Work**_

Python documentation and additional unit tests to follow in separate PR.  Integration with @LJBabbage work on validation


app/templates/questionnaire.html
```diff
@@ -28,8 +28,8 @@
 
     {% endfor %}
 
-  <button class="btn btn--neutral" type="button" name="button">Save &amp; Exit</button>
-  <button class="btn" type="button" name="button">Save &amp; Continue</button>
+  <button class="btn btn--neutral" type="submit" name="action[save_exit]">Save &amp; Exit</button>
```
dhilton, line 31: What does the `action[]` notation do? Is it boilerplate can it be removed?

app/main/views/root.py
```diff
@@ -17,3 +26,57 @@ def mci_survey():
...
+    schema = parser.parse()
+
+    # load the response store
+    response_store = ResponseStoreFactory.create_response_store()
+
+    # load the validation store
+    validation_store = ValidationStoreFactory.create_validation_store()
+
+    # Create the validator
+    class Validator(object):
```
dhilton, line 46: Does this class need to inside the function? Can we move it outside or separate validator module?

app/validation/validation_result.py
```diff
@@ -1,9 +1,26 @@
...
+    def __init__(self, is_valid=False):
+        self.is_valid = is_valid
+        self.errors = {}
+        self.warnings = {}
+
     def is_valid(self):
-        raise NotImplementedError()
+        return self.is_valid
 
     def get_errors(self):
```
dhilton, line 10: Why is this getter required when self.errors is not private? if later on we need to change the return we can turn it into a property via a decorator.

dhilton:
This doesn't have any tests, can we at least add one to check the render scenario as described in the how to test scenario above?

app/templates/questionnaire.html
weapdiv-david, line 31: it doesn't do anything at the moment, it was just added there so that we could differentiate between buttons clicked.

app/main/views/root.py
weapdiv-david, line 46: Lewis has been working on a separate validator class with the same interface.  This one is simply a mock that allows us to progress with the rest of the system

app/validation/validation_result.py
weapdiv-david, line 10: The validator classes will be refactored once lewis's changes have been merged in

weapdiv-david:
The rendering scenario at the end is still very much up for grabs.  Hamish is working on some bits around this area at the moment, based on what data we will be exposing i.e. schema, responses, and validation results.

weapdiv-david:
Architecturally, this is very much a first stab at it based on our current learnings from alpha and the outcomes of research conducted so far.  There is still research to be done around validation and a schema designed for this.  The main components and their interfaces will likely remain the same, but the implementations will be refactored once our understanding improves

dhilton:
OK - on the assurance that these will be addressed later, happy to merge this then @LJBabbage validation stuff.

PR #62, opened 2016-03-11
Pattern library restructure

hamishtaplin:
### Changes
- moves css for pattern library out of the main css file for the runner app
###  Test
- `npm run compile`
- check pattern library isn't upside down


warrenbailey:
LGTM :+1:

PR #64, opened 2016-03-15
EQ 1 JSON format

warrenbailey:
**What**
Implementing the integration with RRM, so extracting the required information from the JWT. Also converting the users responses to the required JSON format for the data integration team.

**How to test**
1. Run the application
2. Complete a survey
3. In the logs locate the JSON response being sent downstream and check the format

**Who can review**


weapdiv-david:
LGTM :+1:  Will merge when you're happy with the deployment @warrenbailey

app/submitter/converter.py
```diff
@@ -0,0 +1,98 @@
...
+                      PERIOD_KEY: user.get_period_id()}
+
+        paradata = {}
+
+        response = {TYPE_KEY: TYPE, VERSION_KEY: VERSION, ORIGIN_KEY: ORIGIN, SURVEY_ID_KEY: survey_id,
+                    SUBMITTED_AT_KEY: datetime.datetime.now().strftime(settings.DATETIME_FORMAT),
+                    COLLECTION_KEY: collection, METADATA_KEY: metadata,
+                    PARADATA_KEY: paradata, DATA_KEY: data}
+
+        logger.error(json.dumps(response))
```
dhilton: Need to be careful here as we're dumping actual data into a log file unencrypted. Anyway we can just hide this behind a feature flag so we explicitly need to turn it on?

warrenbailey: I can just remove it, it's only there for debugging purposes. It's at the wrong level anyway.

PR #65, opened 2016-03-15
Generic factory

warrenbailey:
**What**
Implementing a generic factory class and refactoring code to use.

**How to test**
1. Check out this branch
2. Run the application and make sure nothing is broken

**Who can review**


LJBabbage:
Looks good, but you need to include the current factories maybe

warrenbailey:
@LJBabbage could we do that as a separate PR? It'll be good to get this in as it sets a pattern for any new factories?

LJBabbage:
@warrenbailey not a problem, it is a nice pattern that we can just look at

PR #68, opened 2016-03-17
eq-8 optional response. Update schema/model/parser to allow optional …

LJBabbage:
**What**
Add optional questions, and internationalise error messages

**How to test**
1. Open up the survey and test the form, only "What was the value of the business’s total retail turnover?" should be required.
2.  run pybabel extract -F babel.cfg -o app/translations/messages.pot .  and make sure the pot is updated with results from integer_type_check.py and required_check.py

n.b date from and to will be required but at present they are being worked on in another story

**Who can test**
Anyone but @LJBabbage


dhilton:
LGTM :+1:

PR #70, opened 2016-03-17
EQ 192 Dynamic Survey Loading

warrenbailey:
**What**
This change introduces two concepts.
1. The survey schema is now loaded from the file system based on the eq_id and the form_type. These values must be present in as JWT claims.
2. A new development mode which adds an additional page on the url '/dev'. This form allows you to enter values for the JWT, it will then create a token for you and redirect you to the login page. This page is only enabled if the environment variable EQ_DEV_MODE is set to true.

**How to test**
1. Check out this branch
2. Execute the run_app.sh script and check you can login and complete the survey. This script generates a token with the 0205 form type, therefore check the fuel question is present.
3. Browse to /dev and check a 404 not found is return.
4. Stop the server
5. Set the environment variable EQ_DEV_MODE=True in your shell
6. Start the application
7. Browse to http://localhost:5000/dev
8. You should now see a lovely styled form which you can enter values for the JWT claims
9. Enter some values and submit
10. Check you are logged into the survey

**Who can review**


LJBabbage:
:-1: Old schema

LJBabbage:
We need to make sure we don't have 2 MCI jsons at this stage. Otherwise we will be updating both all the time during development.

PR #71, opened 2016-03-18
remove leading zeroes from q_code (don't align with downstream formats)

dgunthor:
we haven't changed our end yet (Andy Howells doing this), so maybe hold off merging for now


warrenbailey:
LGTM :+1:

PR #75, opened 2016-03-18
eq-1 date range entry

hamishtaplin:
## Changes
- refactored `<select>` elements in date entry template into a macro
- wired back-end validation in to date entries
- upon validation, user-entered data (ie. the date) is retained in date fields
## Test
- `npm run compile`
- submit page without entering a date
  -- confirm "this field is required" error message
- enter an invalid date (eg. 31 February 2016)
  -- confirm "this is not a valid date" error message

### DIFF
--- a/app/data/1_0205.json
+++ b/app/data/1_0205.json
@@ -32,7 +32,7 @@
                                   "mandatory":true
                                 },
                                 {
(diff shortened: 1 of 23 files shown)


warrenbailey:
LGTM :+1:

PR #76, opened 2016-03-21
EQ3 - Adding a text area validator to fix the comments submission issue

warrenbailey:
**What**
Fixes the comment entry box by adding the missing validator class. Also cleaning any user input using bleach, this should stop HTML and Javascript injection. I've also added a display properties section to the schema to allow us to specify max length for a text area.

**How to test**
1. Checkout this branch
2. Make sure you can complete a survey with comments
3. Check you cannot enter html or javascript code into the comments box

**Who can review**


weapdiv-david:
LGTM :+1:

PR #78, opened 2016-03-22
Fixing dev page and adding unit test

warrenbailey:
**What**
Fix for the dev page. Adding a check to the survey template for the data object, if it's not present don't show the extra info. Also adding a unit test to check the dev page is returning a 200.

**How to test**
1. Run the application and check you can hit http://localhost:5000/dev

**Who can review**


weapdiv-david:
LGTM :+1:

PR #79, opened 2016-03-22
Questionnaires without intro jump straight to the questionnaire

weapdiv-david:
**_What**_

Cover page is optional.  This change modifies the navigator behaviour to only send the user to the intro page if the schema contains an intro.

**_How to test**_

1) Check out the branch
2) Enable dev mode and visit the dev console
3) Check each questionnaire
4) mci_mini should direct straight to the question, whereas the others will display an intro page

**_Who can test**_

Anybody except @weapdiv-david


warrenbailey:
LGTM :+1:

PR #81, opened 2016-03-22
eq-9 review answers before submission

LJBabbage:
**What**
The ability to review your answers before submitting the questionnaire

**How to test**
1. Start the questionnaire and fill in as normal
2. Submit the questionnaire to find a review page is now presented (temporary till reskin)
3. Check the information is filled in right
4. Submit the page and check it goes to the thank you page

**Who can test**
Anyone but @LJBabbage

### DIFF
--- a/app/main/views/questionnaire.py
+++ b/app/main/views/questionnaire.py
@@ -18,7 +18,7 @@ def questionnaire():
 
     if request.method == 'POST':
         questionnaire_manager.process_incoming_responses(request.form)
-        current_location = questionnaire_manager.navigator.get_current_location()
+        current_location = questionnaire_manager.get_current_location()
         logger.debug("POST request question - current location %s", current_location)
 
         return _redirect_to_location(current_location)
@@ -26,7 +26,7 @@ def questionnaire():
     render_data = questionnaire_manager.get_rendering_context()
 
     return render_template('questionnaire.html', questionnaire=render_data, data={
-        "survey_code": questionnaire_manager.schema.survey_id,
+        "survey_code": questionnaire_manager.get_schema().survey_id,
         "period": current_user.get_period_str(),
         "respondent_id": current_user.get_ru_ref(),
     })
--- a/app/main/views/questionnaire.py
+++ b/app/main/views/questionnaire.py
@@ -1,16 +1,9 @@
+import logging
 from flask import render_template, request
 from flask_login import login_required, current_user
-from .. import main_blueprint
-from app.validation.validator import Validator
-from app.routing.routing_engine import RoutingEngine
-from app.navigation.navigator import Navigator
-from app.questionnaire.questionnaire_manager import QuestionnaireManager
-from app.main import errors
-from app.utilities.factory import factory
-import logging
-from app.main.views.root import _load_and_parse_schema
 from app.main.views.root import _redirect_to_location
-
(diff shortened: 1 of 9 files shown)


app/main/views/root.py
```diff
@@ -100,10 +96,11 @@ def _load_and_parse_schema(eq_id, form_type):
...
-    json_schema = load_schema(eq_id, form_type)
+
+    json_schema = load_schema(current_user.get_eq_id(), current_user.get_form_type())
     if json_schema:
         parser = SchemaParserFactory.create_parser(json_schema)
         schema = parser.parse()
         return schema
     else:
-        return None
+        return errors.page_not_found()
```
warrenbailey: I don't like this, it should either return a constructed schema model or none. It shouldn't return a rendered page.

app/questionnaire/create_questionnaire_manager.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+def create_questionnaire_manager():
+
+    eq_id = current_user.get_eq_id()
+    form_type = current_user.get_form_type()
+    logger.debug("Requested questionnaire %s for form type %s", eq_id, form_type)
+
+    schema = _load_and_parse_schema()
+    if not schema:
```
warrenbailey, line 21: This won't work with the new method as schema can never be none. Now you've refactored, this doesn't make sense. I think it should throw an exception if schema is none.

app/main/views/questionnaire.py
```diff
@@ -21,46 +14,11 @@ def questionnaire():
...
-                                                 validation_store,
-                                                 routing_engine,
-                                                 navigator,
-                                                 navigation_history)
+    questionnaire_manager = create_questionnaire_manager()
 
     if request.method == 'POST':
         questionnaire_manager.process_incoming_responses(request.form)
-        current_location = navigator.get_current_location()
+        current_location = questionnaire_manager.navigator.get_current_location()
```
warrenbailey: I wonder if it would be better to hide the navigator behind the questionnaire manager. So rather than do this, create a method on questionnaire manager called get_current_location() and then delegate the call through to the navigator component. This hides its internals away then.

app/questionnaire/questionnaire_manager.py
```diff
@@ -3,12 +3,12 @@
...
 class QuestionnaireManager(object):
     def __init__(self, schema, response_store, validator, validation_store, routing_engine, navigator, navigation_history):
-        self._schema = schema
+        self.schema = schema
         self._response_store = response_store
         self._validator = validator
         self._validation_store = validation_store
         self._routing_engine = routing_engine
-        self._navigator = navigator
+        self.navigator = navigator
```
warrenbailey: Can we put this back to 'private' and create a get_current_location() method?

app/main/views/root.py
```diff
@@ -53,7 +46,7 @@ def login():
             abort(404)
 
         # load the schema
-        schema = _load_and_parse_schema(eq_id, form_type)
+        schema = _load_and_parse_schema()
```
warrenbailey, line 49: This if statement is now broken

warrenbailey:
A couple of minor things mate, but overall looks good :-1:

warrenbailey:
Looks good to me :+1:

PR #82, opened 2016-03-23
eq-5 error messages template

hamishtaplin:
screen shot 2016-03-23 at 07 50 43
### Changes
- Error messages displayed at top of page inside `.alert` with links to their respective response.
### Test
- `npm run compile`
- Submit form with errors
- Check that output is as per screenshot


tests/app/renderer/test_renderer.py
```diff
@@ -7,8 +6,7 @@
 class RendererTests(unittest.TestCase):
 
     def test_render(self):
-        renderer = Renderer()
-        self.assertIsInstance(renderer.render(model=None), RenderingContext)
+        pass
```
warrenbailey: You've just removed the unit test, we really should replace this with a valid test.

warrenbailey:
Missing unit tests for the renderer :-1:

warrenbailey:
Needs rebasing too

tests/app/renderer/test_renderer.py
hamishtaplin: I blame @weapdiv-david :-1:

warrenbailey:
:+1:

PR #90, opened 2016-03-24
[hotfix] comment entry

hamishtaplin:
## defect
- upon page load, the comment entry character limit display wasn't taking into account characters already in the textarea
## test
- enter characters in the comment box
- don't enter anything else so the entry is invalid
- confirm "characters remaining" is correct (ie. not 2000)


warrenbailey:
LGTM :+1:

PR #91, opened 2016-03-24
Back button bug fix

LJBabbage:
**What**
If a user completes a questionnaire correctly and goes to the summary page but then goes back and changes a response which would then fail validation the page is still returning to the summary page. This is due to the navigator not updating its state

**How to test**
1. Correctly fill in a questionnaire and click through to the summary page
2. Use the back button to go back to the questionnaire page.
3. Change one of your answers to fail validation
4. Make sure it stays on the questionnaire page and highlights the error

**Who can test**
Anyone but @LJBabbage


warrenbailey:
LGTM :+1:

PR #92, opened 2016-03-24
Attempting to fix travis node issue

warrenbailey:
**What**
Attempting to fix the travis build issue by doing a node clean first. Also changing the order of the build, so that node is done first at least that means we will fail faster.

**Who can review**
Anyone but @hamishtaplin is probably best


dhilton:
LGTM - Travis seems happier.

PR #98, opened 2016-03-31
Serverside template testing

hamishtaplin:
## Changes
- Exposes a route `/dev/render-template` where an AJAX call can retrieve a given template rendered against data in the POST.
- Provides a helper function for interfacing with this in a f/e test
## Note
- ~~This branch is pretty untidy as there are commits from other things in here.~~

### DIFF
--- a/app/dev_mode/views.py
+++ b/app/dev_mode/views.py
@@ -2,12 +2,20 @@ from flask import render_template, redirect, request, abort
 from app.authentication.encoder import Encoder
 from app.authentication.user import UserConstants
 from app.schema_loader.schema_loader import available_schemas
+from flask.ext.cors import cross_origin
 from . import dev_mode_blueprint
 import os
 import time
 import logging
 
 logger = logging.getLogger(__name__)
+
+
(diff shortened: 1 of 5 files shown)


app/dev_mode/views.py
```diff
@@ -9,6 +10,15 @@
 
 logger = logging.getLogger(__name__)
 
+logging.getLogger('flask_cors').level = logging.DEBUG
+
+
+@dev_mode_blueprint.route('/dev/render-template', methods=['POST'])
+@cross_origin()
```
hamishtaplin: This could do with looking at by somebody who knows what they are doing (ie. not me), I just hacked this in to get it to work.

hamishtaplin:
Would be good if somebody could look at this to see if it's feasible. It all works fine locally but depends on the app being running to be able to hit the `/dev/render-template` route.

Also, I added some CORS stuff just to get it to work but have no idea what the security implications are.

weapdiv-david:
Can you rebase, and I'll take a look?

Issue #99, opened 2016-03-31
S-2 1st Page Date format display


stevevic:
The period date is to be displayed as sent through from RM (use field periodstr which should be a string e.g. April 2016)

weapdiv-david:
The period string is being displayed exactly as sent from the Repsondent Manager...

https://github.com/ONSdigital/eq-survey-runner/blob/master/app/main/views/cover_page.py#L28

Is this something we need to take up with them?

stevevic:
Based on code presented - Ok to close.  Thank you

Issue #104, opened 2016-03-31
S-4 - 1st Page Display Trading as name


stevevic:
Display trading as name where provided (as well as text to show it is the trading as name).  Please see story eQ-43 for more details 
https://trello.com/c/zbWNOwY4/43-eq-43-basic-business-landing-page-3-pts

PR #105, opened 2016-03-31
Fixes trading as, return by date formatting, details link, and #99

weapdiv-david:
**_What**_
- Remove the edit address details link (issue #103)
- Displays the Trading As name where supplied (issue #104)
- Formats the Return By date correctly
- Confirms that the period_str is being used by the template as referenced in #99 

**_How to test**_

1) Check out the branch and run the app
2) Confirm that all dates are formatted correctly
3) Confirm that the trading name appears
4) Confirm that the edit address has gone
5) Using dev mode, create a new token and confirm that the value you entered for period_str is displayed at the top right of the page.

**_Who can test**_

Anyone except @weapdiv-david

### DIFF
--- a/app/main/views/cover_page.py
+++ b/app/main/views/cover_page.py
@@ -23,13 +23,15 @@ def cover_page():
         "legal": questionnaire.introduction.legal,
         "description": questionnaire.introduction.description,
         "address": {
-            "name": current_user.get_ru_name()
+            "name": current_user.get_ru_name(),
+            "trading_as": current_user.get_trad_as()
         },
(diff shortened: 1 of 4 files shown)


weapdiv-david:
If Node.JS would stop breaking, this change would fix issues #99, #103, and #104

hamishtaplin:
This is a problem with phantomjs specifically, we need to find a way around it. Some people just ship a binary with the app, it's not ideal but maybe it would work.

hamishtaplin:
Hmm maybe this is it: https://github.com/Medium/phantomjs/issues/517

PR #118, opened 2016-04-04
eq-231, stop re-submission of questionnaire

LJBabbage:
**What**
Stop re-submission of questionnaire (i.e. it can only be submitted once in that session)

**How to test**
1. Make sure the questionnaire works and submits as expected
2. Now try to get back to the submission page anyway you can, back, url entry, etc

**Who can test**
Anyone but @LJBabbage

### DIFF
--- a/app/main/views/questionnaire.py
+++ b/app/main/views/questionnaire.py
@@ -1,8 +1,9 @@
 import logging
-from flask import render_template, request
+from flask import render_template, request, session
 from flask_login import login_required, current_user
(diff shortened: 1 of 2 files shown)


dhilton:
Could we add a test regression for this so that we can add it to the unit tests?

warrenbailey:
I think we should add this check and redirect to the /questionnaire url as well.

warrenbailey:
Perfect :+1:

PR #119, opened 2016-04-04
EQ-246: Session Timeout Extension

LJBabbage:
**What**
Extend session timeout to 8 hours (28800 seconds)

**How to test**
1. Start filling in the questionnaire, ....**STOP**....
2. Wait 7 hours and see if the session data is still there
3. Repeat but now wait over 8 hours and make sure the session data isn't there

**Who can test**
Anyone but @LJBabbage


warrenbailey:
Not sure how realistic it is to do that test locally. I've tested the timeout before and it works, so I suggest we just try this on pre-prod to confirm it works.

warrenbailey:
LGTM :+1:

PR #126, opened 2016-04-06
Eq 22 encrypted submissions

warrenbailey:
**What**
Encrypt submissions before they are placed on the rabbit mq.

**How to test**
1. Complete a survey
2. Check in the logs for the encrypted submission to rabbit mq

**Who can review**


app/settings.py
```diff
@@ -37,6 +34,16 @@ def get_key(key_name):
 EQ_SESSION_TIMEOUT = int(os.getenv('EQ_SESSION_TIMEOUT', '28800'))
 EQ_SECRET_KEY = os.getenv('EQ_SECRET_KEY', os.urandom(24))
 
+# keys for the RRM token exchange
+EQ_RRM_PUBLIC_KEY = get_key(os.getenv('EQ_RRM_PUBLIC_KEY', "./jwt-test-keys/rrm-public.pem"))
+EQ_SR_PRIVATE_KEY = get_key(os.getenv('EQ_SR_PRIVATE_KEY', "./jwt-test-keys/sr-private.pem"))
```
collisdigital: Might be worth naming these now to reflect their usage e.g. This is the user authentication encryption private key, the one above is the user authentication signing public key and the one below is the submission signing private key etc?

app/submitter/encrypter.py
```diff
@@ -0,0 +1,74 @@
...
+    def _encode_iv(self, iv):
+        return self._base_64_encode(iv)
+
+    def _base_64_encode(self, text):
+        # strip the trailing = as they are padding to make the result a multiple of 4
+        # the RFC does the same, as do other base64 libraries so this is a safe operation
+        return base64.urlsafe_b64encode(text).decode().strip("=").encode()
+
+    def _encode_and_signed(self, payload):
+        return jwt.encode(payload, self.private_key, algorithm="RS256", headers={'kid': 'EDCRRM', 'typ': 'jwt'})
```
collisdigital: Need to agree a different KID for this, not EDCRRM, and make it a const for the class? Perhaps EDCSR?

app/settings.py
warrenbailey: I was a bit reluctant to change environment variable names but I do agree with your comment. I'll change it now.

app/submitter/encrypter.py
warrenbailey: Agreed, missed that in my refactoring.

warrenbailey:
Hold off merging to post SDX demo

weapdiv-david:
LGTM :+1:

PR #128, opened 2016-04-06
eq-224 Contact Header

hamishtaplin:
## Screenshot

screenshot 2016-04-07 08 11 28
## Changes
- Updated contact header to latest design
## Test
- `npm run compile`
- Check header matches screenshot


app/templates/layouts/_survey.html
```diff
@@ -23,13 +23,22 @@ <h1 class="bar__title">Monthly Business Survey - Retail Sales Index</h1>
       </div>
 
     {% if data %}
+      {% set info = {
+        'Code': data.survey_code,
+        'Period': data.period_str,
```
hamishtaplin: For some reason `data.period_str` only pulls through on the landing page.

weapdiv-david: fixed in PR #132

hamishtaplin:
This is ready for merge now.

app/templates/layouts/_survey.html
```diff
@@ -24,12 +24,18 @@ <h1 class="bar__title">Monthly Business Survey - Retail Sales Index</h1>
...
-          <li class="info__item info__item--last">Reference: {{ data.respondent_id or '' }}</li>
-        </ul>
+        <h2 class="info__title">Survey Help</h2>
+        <div class="info__tel">
+          0300 1234 931
+        </div>
+        <dl class="info__list">
+          <dt class="info__dt">ID:</dt>
+          <dd class="info__dd">{{ data.survey_code}}</dd>
+          <dt class="info__dt">Period:</dt>
```
warrenbailey, line 34: You've removed the checks on these data values? This means it'll error if that piece of data is missing. Is that what you wanted to do?

hamishtaplin, line 34: We can't have empty values, under what circumstances would they fail?

warrenbailey, line 34: They come directly from RRM, we don't check them, so technically they could be missing, depends on how defensive we want to be.

warrenbailey, line 34: Actually checking the code, those values are required so the RRM has to provide them or the token gets rejected.

warrenbailey:
LGTM :+1:

PR #131, opened 2016-04-06
Update root view

weapdiv-david:
**_What**_

Updates the view returned by the root url, to remove the Hello World message, and replace it with a link to the ONS home page.

**_How to test**_

1) Checkout the branch
2) Run the app
3) Visit [home page](http://localhost:5000/) and verify the "Hello World" message has been replaced.

**_Who can test**_

Anybody except @weapdiv-david

### DIFF
--- a/app/main/errors.py
+++ b/app/main/errors.py
@@ -7,11 +7,18 @@ import logging
 logger = logging.getLogger(__name__)
 
 
+@main_blueprint.app_errorhandler(200)
+def index(error=None):
+    if error:
+        logger.error(error)
+    return _render_error_page(200)
+
+
 @main_blueprint.app_errorhandler(400)
 def bad_request(error=None):
     if error:
         logger.error(error)
-    return _render_error_page(500)
+    return _render_error_page(400)
 
 
 @main_blueprint.app_errorhandler(401)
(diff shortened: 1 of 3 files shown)


collisdigital:
I think this page should be our generic error page. No one should access eQ via the root page, that is an invalid action. We have error page on our backlog for this Sprint....

collisdigital:
This can be researched as part of the error page stuff in future. This was the story in this Sprint: https://trello.com/c/iLfxqu3u

weapdiv-david:
This has arisen out of a conversation with Steve, but I agree it needs to be researched further.  With regard to using an error page though, the elasticbeanstalk deployment seems to check the status of the root url before signalling whether a deployment was successful or not.  Treating the access of the root url as an erro may cause more problems than it solves.

collisdigital:
Yeah maybe return 200 code but same page?

warrenbailey:
So to clarify, we're going to display the error page but return a 200?

collisdigital:
That's what I would suggest; if we agree this is an error condition (hitting eQ directly); my expectation it's the same as going to any URL without a valid token

warrenbailey:
Error pages are now in place, so we should be able to progress this now.

warrenbailey:
Displaying an error page with a 200 code (needed for Elastic beanstalk)

warrenbailey:
Ready for review

PR #135, opened 2016-04-06
Only put submitted responses in the JSON payload to SDX

warrenbailey:
**What**
Only put submitted responses in the JSON payload to SDX. Currently unanswered questions are submitted with empty response, this change means they no longer get submitted. 

**How to test**
Complete a survey with only the mandatory questions completed and check the submitted json does not have the unanswered question.

**Who can review**
@LJBabbage ;-)

### DIFF
--- a/app/submitter/converter.py
+++ b/app/submitter/converter.py
@@ -82,7 +82,8 @@ class Converter(object):
             item = questionnaire.get_item_by_id(key)
             if item is not None:
                 value = responses[key]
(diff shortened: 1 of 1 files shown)


weapdiv-david:
LGTM :+1:

Will need to be rebased to include the encryption which would have made this much harder to test

warrenbailey:
When I rebase i'll add additional logging to log the submitted response

PR #136, opened 2016-04-06
Replaces the numeric error prefix with -

weapdiv-david:
**_What**_

Replaces the numeric error prefixes with a - as the numeric repfixes were too confusing for users who expected them to match to a question number

**_How to test**_

1) Check out the branch
2) Submit the survey with errors
3) Verify that the errors at the top of the page are prefixed with '-'

**_Who can test**_

Anybody except @weapdiv-david

### DIFF
--- a/app/templates/partials/alert.html
+++ b/app/templates/partials/alert.html
@@ -8,7 +8,7 @@
       {% for item_id, errors in all_errors %}
         {% set loop_index = loop.index %}
         {% for error in errors %}
-        <li>{{loop_index}}) {{ error }}. <a href="#response-{{item_id}}">{{_('Go to this error')}}</a></li>
+        <li>- {{ error }}. <a href="#response-{{item_id}}">{{_('Go to this error')}}</a></li>
         {% endfor %}
       {% endfor %}
     </ul>


collisdigital:
Can we please link these kinds of changes to a Trello card or GitHub issue, it's not clear where this has originated from or the research that has driven it, which we do need.

weapdiv-david:
Currently just anecdotal evidence from open door sessions as people expected the number for the error message to relate to the question number.  We are not sure where the number came from originally, so this pull request serves to replace it in the absence of good reason for leaving it alone.

Would be much happier if we could have a discussion around research done on this and why the numbers where there in the first place.

collisdigital:
There was a discussion around this in #design and jilly proposed to discuss, don't know if it happened though. It was tied up with question numbering in general. I would suggest that evidence from ONS staff is probably not a good basis for changes due to our pre-conceptions etc. without testing with respondents. Please check with Ben and Alex before making this change.

bcubbon:
Can we not push this through please. 
I was part of the conversation with both Lewis and David about this, open door people did show that the numbers confused people because they thought the numbers related to question numbers (which we don't have), we did discuss bullet points, but this was not a proposal that I wanted to go through with. It's a question that I want to test with Alex
The numbers have not come through as a requirement from DCM, which suggests that they part of Alex's design, so need to understand first of all where they have come from. Importantly we have not been able to test these error messages at all yet. 
Open door showed us a interesting view but importantly they aren't our users, so we have a hypothesis that we'll take out for research when we next test the patterns

warrenbailey:
We might as well close this PR and reopen it after @bcubbon discusses with @alexmorris :-1:

weapdiv-david:
you beat me to it @warrenbailey

PR #138, opened 2016-04-07
EQ-226 Use SHA256 instead of SHA1 for cookie signing

warrenbailey:
**What**
Force Flask cookie sessions to use SHA256 instead of SHA1

**How to test**
1. First run the application on master, check that the initial cookie on the cover page is about 321kb
2. Checkout this branch, run the application and check the cookie size again.
3. It should now be about 335kb, the additional bytes are due to the SHA256 signing algorithm

**Who can review**


dhilton:
LGTM - observed expected behaviour.

PR #146, opened 2016-04-08
Issue 133 Consistent Error messages

### LINKED ISSUE #133: S-10 Questionnaire - Error Message for numbers with decimal points
There is a little variation between numeric fields for the error message shown when entering a number with decimal point (or commas to suggest £thousands etc)

Message 1:  'This is not a whole number'
- displayed on all the Commodity fields - Food etc.

Message 2: 'Please only enter numbers into the field'
-  displayed on Total Retail Turnover, Internet Sales, Automotive Fuel

Please could we have a single error message for all fields covering this validation, wording as follows:.
'Please only enter whole numbers into the field'
- note - addition of word 'whole' means error wording cannot be generic for all instances where numbers are allowed - so there will be different scenarios where decimals are allowed for entry.

For future reference - the prototype displayed numerics with comma's - if this functionality is provided for in the future, then the presence of a comma to denote thousands, etc should not fail the number format validation.  It is noted that the summary does include commas, and as such allows respondents to check their answers and correct as appropriate.

LJBabbage:
### Changes

To fix #133 
### How to test

See  #133 
### Who can test

Anyone but @LJBabbage

### DIFF
--- a/app/data/1_0205.json
+++ b/app/data/1_0205.json
@@ -77,7 +77,7 @@
                                 "mandatory": false,
                                 "validation": {
                                   "messages": {
-                                    "NOT_INTEGER": "Please only enter numbers into the field.",
+                                    "NOT_INTEGER": "Please only enter whole numbers into the field.",
                                     "NEGATIVE_INTEGER": "The value cannot be negative. Please correct your answer."
                                   }
                                 }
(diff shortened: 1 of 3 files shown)


warrenbailey:
looks good :+1:

PR #148, opened 2016-04-08
Put messages on the queue as strings not bytes

warrenbailey:
### Changes

The message put on the queue is a string not a bytes object
### How to test

Test it's not broken
### Who can test

Anyone but @warrenbailey

### DIFF
--- a/app/submitter/encrypter.py
+++ b/app/submitter/encrypter.py
@@ -35,6 +35,13 @@ class Encrypter (object):
             value = bytes_or_str
         return value
 
+    def _to_str(self, bytes_or_str):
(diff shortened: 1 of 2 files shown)


app/submitter/encrypter.py
```diff
@@ -35,6 +35,13 @@ def _to_bytes(self, bytes_or_str):
             value = bytes_or_str
         return value
 
+    def _to_str(self, bytes_or_str):
```
LJBabbage, line 38: Hate the variable name bytes_or_str

LJBabbage:
:-1: unit test failure

PR #154, opened 2016-04-11
Update INTEGER_TOO_LARGE error message

weapdiv-david:
**_What**_

Updates the error message displayed when an number is entered that is too big for the system to handle

**_How to test**_

1) Check out the branch
2) Enter a large number into any of the numeric fields (greater than 10 digits)
3) Submit the form
4) Verify that the following message is displayed

`The maximum value allowed is 9999999999. Please correct your answer.`

**_Who can test**_

Anybody except @weapdiv-david

### DIFF
--- a/app/data/1_0205.json
+++ b/app/data/1_0205.json
@@ -78,7 +78,8 @@
                                 "validation": {
                                   "messages": {
                                     "NOT_INTEGER": "Please only enter whole numbers into the field.",
-                                    "NEGATIVE_INTEGER": "The value cannot be negative. Please correct your answer."
+                                    "NEGATIVE_INTEGER": "The value cannot be negative. Please correct your answer.",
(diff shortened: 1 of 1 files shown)


warrenbailey:
Any reason why we can't just update the default error message rather than the customizable ones? It would just be a single change then, rather than the same thing ten times?

weapdiv-david:
I've done it this way as the default error messages are deliberately terse in order to encourage authors to over-ride them.  I do hwever think that it might be better to override error messages once within the schema instead of on each response.  This is a larger question then, around schema design and system functionality which I think should be discussed post MCI.

warrenbailey:
OK lets merge this then, I've already added a tech debt backlog item for the error messages so we'll discuss then 👍

warrenbailey:
Rebased

PR #164, opened 2016-04-12
MCI 0203 updated to really version

LJBabbage:
### Changes

Update MCI 0203 form type Json
### How to test

Use /dev mode to get to the survey. Check that the survey 0203 is correct and works as expected 
### Who can test

Anyone but @LJBabbage


warrenbailey:
Looks good 👍

PR #165, opened 2016-04-12
Splunk

warrenbailey:
### Changes

Enables splunk logging
### How to test

Unfortunately needs to be deployed to pre-prod to test
### Who can test

@dhilton


LJBabbage:
Problem with hashes
In --require-hashes mode, all requirements must have their versions pinned with ==. These do not:

PR #166, opened 2016-04-12
eq 251, improve tests and add form type 0203

LJBabbage:
### Changes

Better Tests, including form type 0203
### How to test

Run the tests and make sure everything works
### Who can test

Anyone but @LJBabbage

### DIFF
--- a/tests/integration/create_token.py
+++ b/tests/integration/create_token.py
@@ -1,0 +1,36 @@
+from app.dev_mode.views import generate_token, create_payload
+
+# Set up some constants
+USER = "mci-integration-test"
+EQ_ID = "1"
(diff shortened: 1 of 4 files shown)


tests/integration/create_token.py
```diff
@@ -0,0 +1,36 @@
+from app.dev_mode.views import generate_token, create_payload
+
+# Set up some constants
+USER = "mci-integration-test"
+EQ_ID = "1"
+FORM_TYPE = "0205"
```
warrenbailey: as discussed, this is no longer used.

warrenbailey:
LGTM 👍 💯

PR #174, opened 2016-04-14
Pip hashes

warrenbailey:
**_Changes**_
Pins all pypi dependencies to SHA256 hashes to stop downstream tampering. Also upgrade the version of pip on elastic beanstalk

**_How to test**_
Run a clean install using pip install -U -I -r requirements.txt
Edit one of the hashes in requirements.txt
Re-run the command pip install -U -I -r requirements.txt
It should fail, warning of a tampering with one of the packages.
Deploy to elastic beanstalk and check the application still works.

**_Who can test**_

Anyone but @dhilton and @warrenbailey


weapdiv-david:
I get permission denied on some of the site-packages when trying to do the initial clean install?  I assume this is a problem with my virtualenv... Any ideas?

warrenbailey:
new virtual environment

PR #204, opened 2016-04-20
Renaming jenkins script for production and creating one for pre-prod

warrenbailey:
### Changes

Creating seperate jenkins scripts for prod and pre-prod. Also making the password of the DEV MODE rrm private configurable
### How to test
1. Deploy prod and pre-prod environments
### Who can test

@dhilton

### DIFF
--- a/jenkins-preprod.sh
+++ b/jenkins-preprod.sh
@@ -1,0 +1,59 @@
+#!/bin/bash -e
+
+echo  $Password | gpg -o secure-data.tar --passphrase-fd 0 --batch /opt/keys/secure-data.tar.gpg
+tar xvf secure-data.tar
+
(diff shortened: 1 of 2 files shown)


dhilton:
Can the bash scripts have explicit exit codes? That way we'll be able to know that a build set has explicitly failed or suceeded?

warrenbailey:
I added !/bin/bash -e to the top of the scripts. This means the script will exit if one of the command fails and the jenkins job will then subsequently fail. Is that enough? or do you want me to catch the exit code from the commands and then terminate the script with a specific error code?

dhilton:
Ah I didn't realise that worked. That's cool with me.

dhilton:
LGTM 👍

PR #205, opened 2016-04-20
Token Generation Updates

collisdigital:
### Changes
- Some changes made whilst working with pen testers (to give them some JWT tokens for testing)
- PIP complained about a duplicate package in requirements.txt, so removed it.
- token_generator exploded without EQ_DEV_MODE enabled so added a more useful error message when this scenario occurs
- Remove hard-coded key password for EQ_USER_AUTHENTICATION_RRM_PRIVATE_KEY used by token generator
- Factored out duplicate to_str and to_bytes functions into utility class with tests (needed to convert EQ_USER_AUTHENTICATION_RRM_PRIVATE_KEY_PASSWORD).
- Updated token generator expiration to 12 hours and made units clear (seconds)
### How to test
- Run unit tests, all should pass
- Run  token_generator.py without EQ_DEV_MODE set, observe useful exception message
- Set EQ_DEV_MODE and re-run token_generator.py, check token works.
- Run Flask app locally with EQ_DEV_MODE enabled and test via /dev, should work as normal.
- Try setting the keys/passwords settings as environment vars and see they are picked up correctly (e.g. set to invalid path/password and see failure)
### Who can test
- Anyone but @collisdigital

### DIFF
--- a/app/authentication/jwt_decoder.py
+++ b/app/authentication/jwt_decoder.py
@@ -8,6 +8,8 @@ from cryptography.exceptions import InvalidTag, InternalError
 from app.authentication.invalid_token_exception import InvalidTokenException
 from app.authentication.no_token_exception import NoTokenException
 from app import settings
+from app.utilities import strings
+
 import base64
 import jwt
 import logging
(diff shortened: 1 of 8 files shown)


app/authentication/jwt_decoder.py
```diff
@@ -8,6 +8,8 @@
 from app.authentication.invalid_token_exception import InvalidTokenException
 from app.authentication.no_token_exception import NoTokenException
 from app import settings
+from app.utilities import strings
```
warrenbailey, line 11: Nice, should of done that originally really 👍

app/settings.py
```diff
@@ -47,22 +47,34 @@ def get_key(key_name):
...
+    'EQ_SUBMISSION_SR_PRIVATE_SIGNING_KEY_PASSWORD':    "digitaleq",
+    'EQ_USER_AUTHENTICATION_RRM_PRIVATE_KEY_PASSWORD':  "digitaleq"
+}
+
+# Load keys and passwords, but only allow developer mode defaults if EQ_DEV_MODE
+# has been enabled explicitly...
+for key_name, dev_location in _KEYS.items():
+    path = os.getenv(key_name, dev_location if EQ_DEV_MODE else None)
+    vars()[key_name] = get_key(path)  # assigns attribute to this module
+
```
warrenbailey, line 74: DEV mode isn't set by terraform when it creates any environment (preprod, prod or dev). So this change means the terraform deploy won't work anymore, as the environment wouldn't have any keys  (or the unsecured preprod jenkins job for that matter). 

I like this change and it makes sense, so we either change terraform and add an environment variable for DEV_MODE set to true for dev and preprod or we remove the deploy step from terraform completely?

warrenbailey:
Tested, all works and looks good to me 👍 

We just need to decide what we're doing with Terraform.

PR #217, opened 2016-05-06
Basic navigation

weapdiv-david:
### Changes

Adds support for basic navigation across multi-page surveys

NOTE: The Url scheme used by the application _HAS CHANGED_

Tests have been refactored, but analytics will need to be updated post-deployment
### How to test

1) Check out the branch
2) Run the unit tests
3) Enable dev mode and choose the 999 survey
4) Navigate through the survey, checking validation and submission.  Check the functionality of the back button
### Who can test

Anyone except @weapdiv-david


app/main/views/questionnaire.py
```diff
@@ -1,38 +1,46 @@
...
+                               navigation=context['navigation']
+                               )
 
-    schema = questionnaire_manager.get_schema()
-    return render_template('questionnaire.html', questionnaire=render_data, data={
-        "survey_code": schema.survey_id,
-        "period_str": current_user.get_period_str(),
-        "respondent_id": current_user.get_ru_ref()
-    }, bar_title=schema.title)
+    except NavigationException:
```
LJBabbage: We need to catch these page not found exception with our 404 page, at the moment I get the raw frontend exception

app/main/views/root.py
```diff
@@ -77,15 +77,13 @@ def redirect_to_location(current_location):
     """
```
LJBabbage: Not sure what the purpose for this if block is, aren't all pages effectively going to /questionnaire/' + current_location?

app/navigation/navigator.py
```diff
@@ -20,8 +26,23 @@ def _valid_destination(self, destination):
...
 
         """
-        # TODO Addressable elements for navigation within a questionnaire need to be decided
-        return True
+        if destination == 'introduction':
+            if self._schema.introduction is not None:
+                return True
+            else:
+                return False
+
```
LJBabbage, line 35: This block is a bit odd, could do with a refactor, at least using less if statements

app/questionnaire/create_questionnaire_manager.py
```diff
@@ -39,8 +39,8 @@ def create_questionnaire_manager():
     # create the navigator
     navigator = Navigator(schema, navigation_history)
```
LJBabbage, line 43: Can we remove this if it is now redundant?

app/questionnaire/questionnaire_manager.py
```diff
@@ -11,11 +16,24 @@ def __init__(self, schema, response_store, validator, validation_store, routing_
...
         self._navigation_history = navigation_history
+        self.renderer = Renderer(self._schema, self._response_store, self._validation_store, self._navigator)
 
     def process_incoming_responses(self, post_data):
         # process incoming post data
-        user_responses = self._process_incoming_post_data(post_data)
-
+        user_action, user_responses = self._process_incoming_post_data(post_data)
+
+        # Are we submitting downstram?
```
LJBabbage, line 26: submit_answer should be constant. I would like to see the submission refactored into its own method

app/routing/routing_engine.py
```diff
@@ -1,9 +1,44 @@
...
 class RoutingEngine(object):
     def __init__(self, schema_model, response_store):
         self._schema = schema_model
         self._response_store = response_store
 
         # building routing rules
 
-    def get_next(self, current_location):
-        return "completed"
+    def get_next(self, current_location, user_action):
```
LJBabbage, line 11: Constant file please

app/main/views/root.py
```diff
@@ -77,15 +77,13 @@ def redirect_to_location(current_location):
...
-    elif current_location == 'questionnaire':
-        return redirect(url_for("main.questionnaire"))
+        return redirect('/questionnaire/introduction')
+    elif current_location == "summary":
+        return redirect('/questionnaire/summary')
+    elif current_location == "thank-you":
+        return redirect('/questionnaire/thank-you')
     else:
-        return errors.page_not_found()
+        return redirect('/questionnaire/' + current_location)
```
warrenbailey: Yeah agreed this block looks a little strange. 

However I think we can remove the entire method now, it's only referenced by the login method. That can just do a redirect redirect('/questionnaire/' + current_location)

app/navigation/navigator.py
```diff
@@ -32,12 +53,13 @@ def get_current_location(self):
...
 
-        # We need to have a conversation about what constitutes an addressable element within a questionnaire
+        # Blocks are the navigable elements within a questionnaire
         if current_location is None:
+            # If there is no introduction, go to the first block
             if self._schema.introduction is not None:
                 return 'introduction'
             else:
-                return 'questionnaire'
+                return self._schema.groups[0].blocks[0].id
```
warrenbailey, line 63: What happens if there are no groups or blocks?

app/navigation/navigator.py
```diff
@@ -20,8 +26,23 @@ def _valid_destination(self, destination):
         :returns: True if valid, False otherwise
 
         """
-        # TODO Addressable elements for navigation within a questionnaire need to be decided
-        return True
+        if destination == 'introduction':
```
warrenbailey, line 29: It wouldn't hurt to add some extra logging in here too

app/navigation/navigator.py
weapdiv-david, line 35: I'm not sure how else to do this... 'introduction' is only a valid destination if the schema defines one.  If there's no introduction defined, then it's not a valid destination within the schema

app/main/views/questionnaire.py
weapdiv-david: Thought the exception would be caught and the error page returned... Will update to return the error page

app/navigation/navigator.py
```diff
@@ -20,8 +26,23 @@ def _valid_destination(self, destination):
         :returns: True if valid, False otherwise
```
dhilton, line 26: This function is a membership check; is the string representing a step in our flow a member of our known steps. We can thus rewrite it as an element in a set check:

Something like this (zip just creates a new list from two lists, perhaps should append the custom_steps list). 

``` python
def _valid_destination(self, destination):
    RESERVED_STEPS = ['introduction', 'summary', 'thank-you']
    custom_steps = self._schema.get_steps() # Assumed this function exists somehow.
    step_group = zip(RESERVED_STEPS, custom_steps)
    return destination in step_group # This is a membership check that returns True or False dependent on result.
```

app/templates/layouts/_survey.html
```diff
@@ -7,7 +7,7 @@
   </div>
   <div class="bar__inner container">
     <div class="badge badge--amber u-mr-xs">BETA</div>
-    <div class="bar__title">{{ bar_title }}</div>
+    <div class="bar__title">{% if meta %}{{ meta.survey.title }}{% endif %}</div>
```
dhilton, line 10: Question for @hamishtaplin does this div always need to be present? It is empty if the survey has no title, can we just remove it from the DOM if no title is present?

warrenbailey:
Is this ready now @weapdiv-david ?

warrenbailey:
Looks good to me 👍

app/navigation/navigator.py
warrenbailey, line 26: Sorry Dan, missed your comment when I merged. I'll make the missed changes in a new PR

PR #221, opened 2016-05-10
Fix #189 whitespace removal

### LINKED ISSUE #189: Comments error
### Expected Behaviour

If respondent enters no data but one or more carriage returns into the comments field, the entry should be treated as a null value and no error raised.

Alternative consideration - Error is identified that no data entered (except one or more carriage returns) and the user is able to go to the error field.  The error message could say something along the lines of please provide comments? (optional)
### Actual Behaviour

If I enter one or more carriage returns to the Comments field the character count is decreased by 1 for each carriage return.  On submit (or return if outside the comments field) I get an error message which states this page has 1 error (see pics) but no go to error link and no guidance on why the error has been generated.

Enter one or more carriage returns to comments field and submit

if this is the only error on the form, and only one carriage return was entered, the returned screen has the erroneous carriage returns removed and character count shows 2000 (deafult) which means the respondent can resubmit without making changes.

if there are other errors on the form, this error is not specifically reported at the top of the screen (but is included in the error count), but the comments box is highlighted in red.  Only 1 carriage return is removed for each submission with existing errors on the page e.g. if I have 3 carriage returns (1997 chars left) and 1 date error, on submit I get 1 error and have 1998 chars left in comment box.  If I clear the date error, then submit I get the page error reported and character count decreases by 1.  Each subsequent commit clears another character from the comments box, but errors persist until character count reaches 20000 
#### Browser
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] MS Edge
- [ ] Internet Explorer 11
- [ ] Internet Explorer 10
- [ ] Internet Explorer 9
- [ ] Internet Explorer 8
- [ ] Internet Explorer 7
- [ ] Internet Explorer 6
- [ ] Other (please state below)
#### Operating System
- [x] Windows
- [ ] OS X
- [ ] iOS
- [ ] Android
### Screenshot

as only error
image
image

with other errors
image

warrenbailey:
### Changes

Fixes #189 : Remove leading and trailing whitespace in the comment box so stop bleach failing it when it's empty with carriage returns
### How to test
1. Complete a survey with valid data
2. Enter multiple carriage returns in the comments box
3. Check you can submit (on master it will error) 
### Who can test

### DIFF
--- a/app/questionnaire/questionnaire_manager.py
+++ b/app/questionnaire/questionnaire_manager.py
@@ -97,7 +97,11 @@ class QuestionnaireManager(object):
         return user_action, user_responses
 
     def _clean_input(self, value):
-        return bleach.clean(value)
+        if value:
+            whitespace_removed = value.strip()
+            return bleach.clean(whitespace_removed)
+        else:
+            return value
 
     def get_current_location(self):
         return self._navigator.get_current_location()
--- a/tests/integration/mci/test_empty_comments.py
+++ b/tests/integration/mci/test_empty_comments.py
@@ -1,0 +1,76 @@
+import unittest
+from app import create_app
+from tests.integration.create_token import create_token
+
+
+class TestEmptyComments(unittest.TestCase):
+
+    def setUp(self):
+        self.application = create_app('development')
+        self.client = self.application.test_client()
+
+    def test_empty_submission(self):
+        # Get a token
+        token = create_token('0205')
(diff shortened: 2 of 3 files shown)


LJBabbage:
23 test failures, bar that looks good

warrenbailey:
One failed due to rebase fun, so I've fixed that one. No idea why Travis failed the others

Issue #222, opened 2016-05-11
Change your answers link doesn't work


warrenbailey:
### Expected behaviour
1. Complete a survey 
2. On the review page, click 'change your answers'
3. You should be redirected back to the survey to change your answers
### Actual behaviour
1. Complete a survey 
2. On the review page, click 'change your answers'
3. An error 404 page is displayed
#### Browser
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] MS Edge
- [ ] Internet Explorer 11
- [ ] Internet Explorer 10
- [ ] Internet Explorer 9
- [ ] Internet Explorer 8
- [ ] Internet Explorer 7
- [ ] Internet Explorer 6
- [ ] Other (please state below)
#### Operating System
- [x] Windows
- [ ] OS X
- [ ] iOS
- [ ] Android

dhilton:
This was broken, but is now fixed. Closin'

PR #223, opened 2016-05-11
Fixes link on submission page

weapdiv-david:
### Changes

Fixes the change answers link on the summary page by introducing a new url... (Issue #222)

`/questionnaire/first` will always redirect to the first block in the questionnaire.
### How to test

1) Check out the branch
2) Run the tests
3) Run the app and fill in the questionniare and pause at the summary page.
4) Click the link to change your answers
5) Verify that you are back at the first page of the questionnaire
### Who can test

Anyone except @weapdiv-david

### DIFF
--- a/app/main/views/questionnaire.py
+++ b/app/main/views/questionnaire.py
@@ -8,6 +8,7 @@ from .. import main_blueprint
 from app.model.questionnaire import QuestionnaireException
 from app.main.errors import page_not_found
 
+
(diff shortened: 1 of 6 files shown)


warrenbailey:
Fix looks good, but I think we should have a integration test for this so it doesn't happen again. Can you add one?

warrenbailey:
LGTM 👍

PR #225, opened 2016-05-13
Eq 218 validation date

LJBabbage:
### Changes

Date type validation according to the story, which requires json changes to new structure (question types)
### How to test
1. Make sure every first works and looks right as mass design changes
2. Put in dates the same in the range and make sure an error message is generated
3. Put the from date later then the to date and make sure an error message is generated
### Who can test

Anyone but @LJBabbage


app/validation/date_range_check.py
```diff
@@ -0,0 +1,48 @@
+from app.validation.abstract_validator import AbstractValidator
+from app.validation.validation_result import ValidationResult
+from datetime import datetime
+
+DATE_FROM = "From"
```
warrenbailey: This will mean that the response labels have to be to and from for a data range question. Is that correct?

What happens when we do a welsh survey?

app/validation/validator.py
```diff
@@ -29,7 +30,9 @@ def __init__(self, schema, validation_store, response_store):
             AbstractValidator.MANDATORY: _("This field is mandatory."),
             AbstractValidator.INVALID_DATE: _("This is not a valid date."),
             AbstractValidator.NEGATIVE_INTEGER: _("Negative values are not allowed."),
-            AbstractValidator.INTEGER_TOO_LARGE: _('This number is too big.')
+            AbstractValidator.INTEGER_TOO_LARGE: _('This number is too big.'),
+            AbstractValidator.INVALID_DATE_RANGE_DIFF: _("The 'to' date cannot be before the 'from' date."),
```
warrenbailey: Again should the words 'to' and 'from' be in these defaults?

Also should these error messages be present in the schema?

app/validation/date_range_check.py
```diff
@@ -0,0 +1,48 @@
...
+DATE_TO = "To"
+
+
```
warrenbailey: I think we need some logging in this class

app/validation/validator.py
```diff
@@ -47,13 +50,17 @@ def validate(self, user_data):
...
 
         # Return true/False for the set of values
         for item_id in user_data.keys():
             result = self._validation_store.get_result(item_id)
+            item = self._schema.get_item_by_id(item_id)
+            container_result = self._validation_store.get_result(item.container.id)
+
             if result:
-                if not result.is_valid:
+                if not result.is_valid or not container_result.is_valid:
```
warrenbailey: At this point it's possible for container_result to be None and then the page explodes like this:

screen shot 2016-05-13 at 12 14 18 3

warrenbailey:
If I enter a to date thats after the from date then the pages errors correctly. However if I then change one of the dates to be invalid and resubmit I get a double error message (see screenshots). Interestingly depending on which date I make invalid I get different error messages. 

I check this with Jill and @collisdigital and we all agree it needs to be fixed and not so the double error message

screen shot 2016-05-13 at 12 17 54 3
screen shot 2016-05-13 at 12 17 13 3

app/validation/date_range_check.py
LJBabbage: True, but assuming the 1st response is always 'from' and the 2nd 'to' is a more dangerous position, I have to have some kind of handle and at present we don't have one bar the label

app/validation/validator.py
LJBabbage: We don't have the concept of Question error messages at the moment so they can't go in the schema.

weapdiv-david:
Happy with the code as is, but it does need to be refactored later to be implemented as validation rather then type-checking.  That will get rid of the double error message too.

warrenbailey:
LGTM, though I think we need to restructure some of the validation stuff asap 👍

warrenbailey:
Rebased

PR #226, opened 2016-05-13
Eq 332: Save and Resume - Unique per Collection

warrenbailey:
### Changes

The user id for save and resume is a hash value consisting of the ru_ref, collection exercise sid and eq_id.
### How to test
1. Enable save and resume; 
   `export EQ_SERVER_SIDE_STORAGE=True`
   `export EQ_SERVER_SIDE_STORAGE_TYPE=Database`
2. Check you can partially complete a survey and then return to it later
3. Check the file /tmp/questionnaire.db and ensure the user id is a hash value.
### Who can test

Anyone apart from @warren-bailey


app/authentication/user_id_generator.py
```diff
@@ -0,0 +1,30 @@
...
+class UserIDGenerator(object):
+
+    @staticmethod
+    def generate_id(token):
+        logger.debug("About to generate a user id")
+        ru_ref = token.get(MetaDataConstants.RU_REF)
+        collection_exercise_sid = token.get(MetaDataConstants.COLLECTION_EXERCISE_SID)
+        eq_id = token.get(MetaDataConstants.EQ_ID)
+
+        logger.debug("Using values %s, %s and %s with a salt", ru_ref, collection_exercise_sid, eq_id)
```
collisdigital: Worth adding an assert (maybe even if+exception?) here to ensure these values exist/aren't empty etc?

app/settings.py
```diff
@@ -40,11 +40,12 @@ def get_key(key_name):
...
 
-EQ_SERVER_SIDE_STORAGE = parse_mode(os.getenv('EQ_SERVER_SIDE_STORAGE', 'True'))
-EQ_SERVER_SIDE_STORAGE_TYPE = os.getenv('EQ_SERVER_SIDE_STORAGE_TYPE', 'memory')
-EQ_SERVER_SIDE_STORAGE_ENCRYPTION = parse_mode(os.getenv('EQ_SERVER_SIDE_STORAGE_ENCRYPTION', 'True'))
+EQ_SERVER_SIDE_STORAGE = parse_mode(os.getenv('EQ_SERVER_SIDE_STORAGE', 'False'))
+EQ_SERVER_SIDE_STORAGE_TYPE = os.getenv('EQ_SERVER_SIDE_STORAGE_TYPE', 'DATABASE')
+EQ_SERVER_SIDE_STORAGE_ENCRYPTION = parse_mode(os.getenv('EQ_SERVER_SIDE_STORAGE_ENCRYPTION', 'False'))
 EQ_SERVER_SIDE_STORAGE_ENCRYPTION_KEY = os.getenv('EQ_SERVER_SIDE_STORAGE_ENCRYPTION_KEY', "lord.darth.vader").encode()
 EQ_SERVER_SIDE_STORAGE_DATABASE_URL = os.getenv('EQ_SERVER_SIDE_STORAGE_DATABASE_URL', 'sqlite:////tmp/questionnaire.db')
+EQ_SERVER_SIDE_STORAGE_SALT = os.getenv('EQ_SERVER_SIDE_STORAGE_SALT', 'luke.skywalker.r2d2.c3p0')
```
collisdigital: Consider clarifying the name further e.g. USER_ID_SALT

tests/app/authentication/test_user_id_generator.py
```diff
@@ -0,0 +1,24 @@
+from app.authentication.user_id_generator import UserIDGenerator
+from app.metadata.metadata_store import MetaDataConstants
+from app import settings
+import unittest
+
+
+class TestUserIDGenerator(unittest.TestCase):
```
collisdigital, line 8: Consider adding a test that verifies that changing the SALT changes the hash for the same values and for different combinations of each element rather than just this one check (i.e. that each individual element contributes to the uniqueness)

app/authentication/user_id_generator.py
```diff
@@ -0,0 +1,30 @@
...
+import hashlib
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+class UserIDGenerator(object):
+
+    @staticmethod
+    def generate_id(token):
```
collisdigital, line 14: This could just be a method inside this file (i.e. outside of a class object), I suggest this would be a coding standard thing as to whether you always put things inside classes or not, as long as its consistent you can do either approach.

warrenbailey, line 14: I did consider that, I think its the Java in me that wants to put everything into classes.

tests/app/authentication/test_user_id_generator.py
warrenbailey, line 8: Good shout, will add one now

app/authentication/user_id_generator.py
warrenbailey: I'll add one, but the token has already been checked by this point, so would have been rejected if those values are missing.

app/settings.py
warrenbailey: Will do

warrenbailey:
@collisdigital: comments addressed.

All ready for re-review 👍

tests/app/authentication/test_user_id_generator.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def setUp(self):
+        settings.EQ_SERVER_SIDE_STORAGE = True
+
+    def test_generate_id(self):
+        user_id_1 = UserIDGenerator.generate_id(self.create_token('1', '2', '3'))
+        user_id_2 = UserIDGenerator.generate_id(self.create_token('1', '2', '3'))
+        user_id_3 = UserIDGenerator.generate_id(self.create_token('1', '2', '4'))
```
collisdigital, line 16: Still needs (2, 2, 4) != user_id_3 and
and (1,3,3) != user_id_2 to cover each element changing on its own....

collisdigital:
One last comment on ensuring hashing is working correctly across all 3 values in token + salt, otherwise LGTM

collisdigital:
:+1:

PR #228, opened 2016-05-16
Revert issue with change your answers link to drop you at the begining of the questionnaire

dhilton:
### Changes

Revert the change I introduced by merging the selenium tests, causing the fix to the `change your answers` link to fail.
### How to test
1. Check out this branch
2. Complete a survey and then click the change your answers link
3. You should be dropped back at the begining of the questionnaire process.
### Who can test

Anyone but @dhilton


warrenbailey:
LGTM 👍  💯

PR #231, opened 2016-05-17
Eq 322 harden security

warrenbailey:
### Changes

Hardening the security around the save and resume functionality, specifically:
1. Use a random IV
2. Generating a user id using a key derivation function
3. Generate a CEK passed on a salt and the generated user id
4. Enabling Flask database sessions to store the user id (the cookie will now only include the session id)
### How to test
1. Enable server side storage using the following env variables
   `export EQ_SERVER_SIDE_STORAGE=True`
   `export EQ_SERVER_SIDE_STORAGE_ENCRYPTION=True`
2. Clear all eq related cookies
3. Run the server and login
4. Check that in the eq cookie you can only see a session id
5. Check that all functionality of a survey still works
6. Check that you can partially complete a survey, get a new JWT and retrieve your old data.
### Who can test


app/authentication/user_id_generator.py
```diff
@@ -19,17 +22,15 @@ def generate_id(token):
...
-            sha256 = hashlib.sha256()
-            sha256.update(to_str(ru_ref).encode('utf-8'))
-            sha256.update(to_str(collection_exercise_sid).encode('utf-8'))
-            sha256.update(to_str(eq_id).encode('utf-8'))
-            sha256.update(to_str(salt).encode('utf-8'))
-
-            user_id = sha256.hexdigest()
-            logger.debug("User ID is %s", user_id)
-            return user_id
+            kdf = PBKDF2HMAC(algorithm=hashes.SHA256(), length=32, salt=salt, iterations=100000, backend=backend)
```
collisdigital: Should this be defaulting to 100,000 iterations? Also consider making it a setting and impose a minimum limit.

app/storage/encrypted_storage.py
```diff
@@ -1,34 +1,56 @@
...
 
     def __init__(self, server_storage):
-        self.encryption = JWEDirEncrypter(settings.EQ_SERVER_SIDE_STORAGE_ENCRYPTION_KEY)
-        self.decryption = JWEDirDecrypter(settings.EQ_SERVER_SIDE_STORAGE_ENCRYPTION_KEY)
+        self.encryption = JWEDirEncrypter()
+        self.decryption = JWEDirDecrypter()
         self.server_storage = server_storage
 
     def store(self, user_id, data):
         logger.debug("About to encrypt data %s", data)
```
collisdigital: Avoid logging this at all...

app/storage/encrypted_storage.py
```diff
@@ -1,34 +1,56 @@
...
+import json
+import hashlib
+import logging
+
+logger = logging.getLogger(__name__)
 
 
 class EncryptedServerStorageDecorator(AbstractServerStorage):
 
     ENCRYPTED_KEY = 'encrypted'
```
collisdigital: Really need some comments around this value, what is it for, on face value looks like a security problem but that's probably because it isn't clear wha it's for.

collisdigital: I think I see but perhaps a poor choice of naming here, maybe DATA_IS_ENCRYPTED_FLAG or something would be better (I.e this isn't actually an encryption key)

collisdigital: No I'm confused, probably best to discuss.

tests/app/storage/test_encrypted_storage.py
```diff
@@ -0,0 +1,33 @@
+from app.storage.encrypted_storage import EncryptedServerStorageDecorator
+from app.storage.memory_storage import InMemoryStorage
+
+
+import unittest
+
+
+class TestEncryptedServerStorageDecorator(unittest.TestCase):
+
+    def test_generate_cek(self):
```
collisdigital, line 11: Test changing the Salt changes the hash

app/authentication/user_id_generator.py
warrenbailey: No typo there, i'll change it to be a setting with a minimum

app/storage/encrypted_storage.py
warrenbailey: I can take out the logging, but it would be useful to be able to log this info when in dev mode with debug logging on as it makes development easier.

app/storage/encrypted_storage.py
warrenbailey: It's literally just a JSON wrapper round the encrypted data. It just allows me to use the the same classes to store the data whether it's encrypted or not. It was also a bit forward planning, I thought it would be easier to use the Postgres JSON column and/or switch to a NoSQL datastore in the future if the data we're saving is JSON.

It's obviously badly named so I'll change that.

tests/app/storage/test_encrypted_storage.py
warrenbailey, line 11: Will do

warrenbailey:
Ready for re-review

warrenbailey:
@collisdigital is happy with latest changes. Ready for review/merge

collisdigital:
LGTM 👍

PR #232, opened 2016-05-17
Eq 320 piping

weapdiv-david:
### Changes

Pipes exercise start and end dates into question text
### How to test

1) Check out the branch
2) Run the unit tests
3) Run the app and check the date period question
4) Create a new token using dev mode with new dates and check that the same questionnaire shows the dates in your token
### Who can test

Anybody except @weapdiv-david

### DIFF
--- a/app/data/1_0203.json
+++ b/app/data/1_0203.json
@@ -25,7 +25,7 @@
                             {
                               "id": "6cc86b54-330c-4465-99b2-34cc7073dc2c",
                               "title": "What are the dates of the sales period you are reporting for?",
-                              "description": "If possible, this should be for the period 3rd April 2016 to 30th April 2016.",
+                              "description": "If possible, this should be for the period {exercise.start_date:%-d %B %Y} to {exercise.end_date:%-d %B %Y}.",
                               "type" : "DateRange",
                               "responses": [
                                 {
--- a/app/metadata/metadata_store.py
+++ b/app/metadata/metadata_store.py
@@ -41,18 +41,14 @@ class MetaDataStore(object):
             self.user = current_user
 
     def store(self, key, value):
-        if self.user is not None:
-            data = self.user.get_questionnaire_data()
-            data[key] = value
+        data = self.user.get_questionnaire_data()
+        data[key] = value
 
     def store_all(self, token):
         for key in token.keys():
             self.store(key, token[key])
 
     def get(self, key):
-        if self.user is None:
-            return None
-
         data = self.user.get_questionnaire_data()
         if key in data:
             return data[key]
@@ -81,13 +77,9 @@ class MetaDataStore(object):
         return self.get(MetaDataConstants.FORM_TYPE)
 
     def get_ref_p_start_date(self):
-        if self.user is None:
-            return None
         return datetime.strptime(self.get(MetaDataConstants.REF_P_START_DATE), "%Y-%m-%d")
 
     def get_ref_p_end_date(self):
-        if self.user is None:
-            return None
         return datetime.strptime(self.get(MetaDataConstants.REF_P_END_DATE), "%Y-%m-%d")
 
     def get_trad_as(self):
--- a/app/metadata/metadata_store.py
+++ b/app/metadata/metadata_store.py
@@ -1,4 +1,5 @@
 from flask_login import current_user
+from datetime import datetime
 
 
 class MetaDataConstants(object):
@@ -76,16 +77,16 @@ class MetaDataStore(object):
         return self.get(MetaDataConstants.FORM_TYPE)
 
     def get_ref_p_start_date(self):
-        return self.get(MetaDataConstants.REF_P_START_DATE)
+        return datetime.strptime(self.get(MetaDataConstants.REF_P_START_DATE), "%Y-%m-%d")
 
     def get_ref_p_end_date(self):
-        return self.get(MetaDataConstants.REF_P_END_DATE)
+        return datetime.strptime(self.get(MetaDataConstants.REF_P_END_DATE), "%Y-%m-%d")
 
     def get_trad_as(self):
         return self.get(MetaDataConstants.TRAD_AS)
 
     def get_return_by(self):
-        return self.get(MetaDataConstants.RETURN_BY)
+        return datetime.strptime(self.get(MetaDataConstants.RETURN_BY), "%Y-%m-%d")
 
     @staticmethod
     def is_valid(token):
--- a/app/metadata/metadata_store.py
+++ b/app/metadata/metadata_store.py
@@ -41,14 +41,18 @@ class MetaDataStore(object):
             self.user = current_user
 
     def store(self, key, value):
-        data = self.user.get_questionnaire_data()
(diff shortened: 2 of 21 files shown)


app/data/1_0203.json
```diff
@@ -25,7 +25,7 @@
                             {
                               "id": "6cc86b54-330c-4465-99b2-34cc7073dc2c",
                               "title": "What are the dates of the sales period you are reporting for?",
-                              "description": "If possible, this should be for the period 3rd April 2016 to 30th April 2016.",
+                              "description": "If possible, this should be for the period {exercise.start_date:%-d %B %Y} to {exercise.end_date:%-d %B %Y}.",
```
LJBabbage, line 28: We need to escape { if anyone tries to use it in the text it blows up

app/piping/plumber.py
```diff
@@ -0,0 +1,35 @@
+import re
+
+
+class Plumber(object):
+    def __init__(self, context):
+        self._context = context
+
+    def plumb_item(self, item):
+        interesting_properties = self.get_properties_of_interest(item)
```
LJBabbage: Not sure what interesting_properties means, can you change the name to something more descriptive

app/data/1_0203.json
collisdigital, line 28: Consider adding a test that checks this situation.

app/renderer/renderer.py
```diff
@@ -112,47 +136,53 @@ def _render_navigation_meta(self):
...
+                    response.warnings = None
 
     def _augment_questionnaire(self):
         errors = OrderedDict()
         warnings = OrderedDict()
 
         # loops through the Schema and get errors and warnings in order
+        # augments each item in the schema as required
         for group in self._schema.groups:
+            self._plumber.plumb_item(group)
```
collisdigital: As a minimum these steps (lines 161-166) can be factored out onto a function and called for each of the repeats in the nested loops below. Further refactoring may be possible beyond this to clean this up.

app/metadata/metadata_store.py
```diff
@@ -76,16 +81,20 @@ def get_form_type(self):
         return self.get(MetaDataConstants.FORM_TYPE)
 
     def get_ref_p_start_date(self):
-        return self.get(MetaDataConstants.REF_P_START_DATE)
+        if self.user is None:
+            return None
+        return datetime.strptime(self.get(MetaDataConstants.REF_P_START_DATE), "%Y-%m-%d")
```
warrenbailey: I'd like to understand in what situations self.user is None as it shouldn't be.

Also you're already doing a self.user check in  self.get. I wonder if this would be better written:

```
def get_ref_p_start_date(self):
        start_date = self.get(MetaDataConstants.REF_P_START_DATE)       
        if start_date:
            return datetime.strptime(start_date, "%Y-%m-%d")
```

app/metadata/metadata_store.py
```diff
@@ -40,14 +41,18 @@ def __init__(self, user=None):
             self.user = current_user
 
     def store(self, key, value):
-        data = self.user.get_questionnaire_data()
-        data[key] = value
+        if self.user is not None:
```
warrenbailey: I wonder if we re-write this class and initialize all the variables on construction. That way it could be immutable and we wouldn't need these checks.

app/renderer/renderer.py
```diff
@@ -13,6 +13,31 @@ def __init__(self, schema, response_store, validation_store, navigator, metadata
...
+
+        try:
+            start_date = self._metadata.get_ref_p_start_date()
+            end_date = self._metadata.get_ref_p_end_date()
+        except:
+            pass
+
+        context = {
+            "exercise": ContextEntry({
+                # "start_date": '{dt.day} {dt:%B} {dt.year}'.format(dt=datetime.strptime(self._metadata.get_ref_p_start_date(), "%Y-%m-%d")),
```
warrenbailey: Lets just remove the commented out code if its not needed.

app/renderer/renderer.py
```diff
@@ -13,6 +13,31 @@ def __init__(self, schema, response_store, validation_store, navigator, metadata
         self._navigator = navigator
         self._metadata = metadata
 
+        # Create an object from a dict
+        class ContextEntry:
+            def __init__(self, entries):
+                self.__dict__ = entries
```
warrenbailey: You have exactly the same class called ObjectFromDict in test_plumber.py. Maybe refactor both into the app.utilities package.

app/renderer/renderer.py
```diff
@@ -13,6 +13,31 @@ def __init__(self, schema, response_store, validation_store, navigator, metadata
...
+
+        start_date = None
+        end_date = None
+
+        try:
+            start_date = self._metadata.get_ref_p_start_date()
+            end_date = self._metadata.get_ref_p_end_date()
+        except:
+            pass
```
warrenbailey, line 24: I don't really understand this try/except block. Either catch the exception and log it or at the very least comment whats going on as to me this looks like it will just quietly swallow all errors.

app/piping/plumber.py
```diff
@@ -0,0 +1,35 @@
...
+
+        for interesting_property in interesting_properties:
+            if self._needs_plumbing(item, interesting_property):
+                self._plumb(item, interesting_property)
+
+    def get_properties_of_interest(self, item):
+        if hasattr(item, 'templatable_properties'):
```
warrenbailey: Would it be possible to have templatable_propertes value on all aspects of the schema? We could set it to [] to start with?

app/piping/plumber.py
```diff
@@ -0,0 +1,35 @@
...
+            return item.templatable_properties
+        else:
+            return []
+
+    def _needs_plumbing(self, item, interesting_property):
+        value = getattr(item, interesting_property)
```
warrenbailey: maybe call this variable 'template' to match the one below

app/piping/plumber.py
```diff
@@ -0,0 +1,35 @@
+import re
+
+
+class Plumber(object):
+    def __init__(self, context):
```
warrenbailey, line 5: I'm finding this class a little hard to follow, maybe add some comments to explain whats going on.

app/piping/plumber.py
```diff
@@ -0,0 +1,35 @@
+import re
+
+
```
warrenbailey, line 3: I think this class needs debug logging.

app/renderer/renderer.py
warrenbailey, line 24: Could still do with explaining this.

app/piping/plumber.py
warrenbailey, line 5: With the variable name change it's definitely better, but a bit of documentation wouldn't hurt

warrenbailey:
Looks good now, a couple of minor things but my only comment now is it has no logging. I really think some debug logging in here, especially the plumber class, would be really useful. Even if it just tells us what properties are being replace with what values.

warrenbailey:
LGTM 👍

PR #234, opened 2016-05-18
Adding form type to the generated user id

warrenbailey:
### Changes

Adding form type to the list of values used to generate the user id. This is needed so that the user can complete multiple surveys with the same id but different form types.
### How to test
1. Check out this branch
2. Turn on server side storage
   `
   export EQ_SERVER_SIDE_STORAGE=True
   export EQ_SERVER_SIDE_STORAGE_ENCRYPTION=True
   `
3. Select the MCI survey with form type 0205
4. Partially complete the survey, save it and then check you can return to it later
5. With partially completed data select the MCI survey 0203 and check that you are taken to the landing page.
### Who can test

Anyone apart form @warrenbailey


weapdiv-david:
LGTM :+1:

PR #238, opened 2016-05-23
Add new safe health check that doesn't disclose any system info

dhilton:
### Changes

To enable us to turn healthchecks on in production without disclosing any sensitive information.We have added a new health check url that hides any detailed issues and
merely returns OK or DOWN for a healthcheck request, with
associated http codes of 200 and 500 respectively.
### How to test
1. Run the app, check `/health` you should see a simple output of `OK`
2. Change the DEV_MODE environment variable
3. ReRun the app and check that you can't see `/health` and instead can see the current full healthcheck at `/healthcheck`
### Who can test

Anyone but @dhilton

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -26,7 +26,7 @@ from logging.handlers import RotatingFileHandler
 from flask_analytics import Analytics
 from splunk_handler import SplunkHandler
 from app.analytics.custom_google_analytics import CustomGoogleAnalytics
-
+import json
 
 SECURE_HEADERS = {
     'Cache-Control': 'no-cache, no-store, must-revalidate',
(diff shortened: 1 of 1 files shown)


weapdiv-david:
I can't run this outside of dev mode as I don't have keys

dhilton:
Can you modify your local version to trigger that scenario without having to generate a new set of encryption credentials? I realise this isn't a true test but given the change that will give a true reflection of the functionality added.

weapdiv-david:
Right, got it!  The safe health check url is `/status` not `/health`, but I found it, I've seen it working, and I'm happy with it :)

weapdiv-david:
LGTM :+1: Mergin'!

PR #239, opened 2016-05-24
Database Session management and RDS Integration

warrenbailey:
### Changes

Various changes to support RDS integration:
1. Only store a randomly generated session id in the cookie. This eq-session-id is generated using UUID4
2. Create a mapping between this session id and the actual user id and store this in a database table. Then when the request arrives, extract the session id, load the user id from the table and use that to load the questionnaire data. 
3. Add the various postgres dependencies
4. Remove the Flask server side sessions code and dependency (as it doesn't work with postgres)
### How to test
1. Enable server side storage

```
export EQ_SERVER_SIDE_STORAGE=True
```
1. Partially complete a survey, then check you can return to it later. (You can use the dev page to do this)
2. Check you can partially complete different surveys (again use the dev page) and that the data does not get re-used between surveys.
### Who can review


app/authentication/session_management.py
```diff
@@ -33,30 +38,91 @@ def get_user_id(self):
...
+        logger.debug("DatabaseSessionManager get_user_id() - session %s", session)
+        if EQ_SESSION_ID in session:
+            eq_session_id = session[EQ_SESSION_ID]
+            eq_session = self._get_object(eq_session_id)
+            return eq_session.user_id
+        else:
+            return None
+
+    def create_session_id(self):
+        return str(uuid4())
```
collisdigital: Is there a check somewhere that a new id being created doesn't already exist in the db (could silently cause strange behaviour).

warrenbailey: No i'll add a check.

weapdiv-david:
LGTM :+1:

app/authentication/session_management.py
```diff
@@ -33,30 +38,103 @@ def get_user_id(self):
...
+        logger.debug("DatabaseSessionManager get_user_id() - session %s", session)
+        if EQ_SESSION_ID in session:
+            eq_session_id = session[EQ_SESSION_ID]
+            eq_session = self._get_object(eq_session_id)
+            return eq_session.user_id
+        else:
+            return None
+
+    def create_session_id(self):
+        while True:
```
warrenbailey, line 92: Change this to be max retry count, as this risks locking up the whole instance

collisdigital, line 92: Since this should probably rarely occur, if at all; if it does happen after multiple retires it would suggest something more serious is wrong (log error/throw exception?)

PR #244, opened 2016-06-08
Eq 374 serialiser

weapdiv-david:
### Changes

Adds the MCI Schema as python code, along with to_json() methods to generate json from model.
Also overrides **eq** and **ne** methods to check for equivalence rather than strict equality.
### How to test

1) Check out the branch
2) Run the unit tests
3) Verify they all pass
### Who can test

Anyone except @weapdiv-david

### DIFF
--- a/app/model/block.py
+++ b/app/model/block.py
@@ -65,7 +65,7 @@ class Block(object):
                     if validation != other.validation[index]:
                         return False
 
-            templatable_properties_match = False
+            templatable_properties_match = True
             if len(self.templatable_properties) != len(other.templatable_properties):
                 return False
 
--- a/app/model/block.py
+++ b/app/model/block.py
@@ -76,7 +76,10 @@ class Block(object):
             return properties_match and sections_match and validations_match and templatable_properties_match
 
         else:
-            return False
+            return NotImplemented
 
     def __ne__(self, other):
-        return not self.__eq__(other)
+        return not self == other
+
+    def __hash__(self):
+        return hash((self.id, self.title))
--- a/app/model/block.py
+++ b/app/model/block.py
@@ -10,7 +10,7 @@ class Block(object):
         self.container = None
         self.questionnaire = None
         self.validation = None
-        self.questionniare = None
+        self.questionnaire = None
         self.templatable_properties = []
         self.display = Display()
 
@@ -18,3 +18,61 @@ class Block(object):
         if section not in self.sections:
             self.sections.append(section)
             section.container = self
+
+    def to_json(self):
+        json_dict = {
+            "id": self.id,
+            "title": self.title,
+            "sections": [],
+            "validation": [],
+            "display": {}
+        }
+
+        for section in self.sections:
+            json_dict['sections'].append(section.to_json())
+
+        if self.validation is not None:
+            for validation in self.validation:
+                json_dict['validation'].append(validation.to_json())
+
+        if self.display is not None:
+            json_dict['display'] = self.display.to_json()
+
+        return json_dict
+
+    def __eq__(self, other):
+        if isinstance(other, Block):
+            properties_match = self.id == other.id and \
+                               self.title == other.title
+
+            sectionns_match = True
+            if len(self.sections) != len(other.sections):
+                return False
+
+            for index, section in enumerate(self.sections):
+                if section != other.sections[index]:
+                    return False
+
+            validations_match = True
+            if self.validation is not None and other.validation is not None:
+                if len(self.validation) != len(other.validation):
+                    return False
+
+                for index, validation in enumerate(self.validation):
+                    if validation != other.validation[index]:
+                        return False
+            else:
+                return False
+
+            templatable_properties_match = False
+            if len(self.templatable_properties) != len(other.templatable_properties):
+                return False
+
+            for index, templatable_property in enumerate(self.templatable_properties):
+                if templatable_property not in other.templatable_properties:
+                    return False
+
+            return properties_match and sections_match and validations_match and templatable_properties_match
+
+        else:
+            return False
--- a/app/model/block.py
+++ b/app/model/block.py
@@ -41,11 +41,14 @@ class Block(object):
         return json_dict
 
     def __eq__(self, other):
+        if id(self) == id(other):
+            return True
+
         if isinstance(other, Block):
             properties_match = self.id == other.id and \
                                self.title == other.title
 
-            sectionns_match = True
+            sections_match = True
             if len(self.sections) != len(other.sections):
                 return False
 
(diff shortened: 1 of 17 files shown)


app/model/block.py
```diff
@@ -10,11 +10,73 @@ def __init__(self):
...
+        if self.validation is not None:
+            for validation in self.validation:
+                json_dict['validation'].append(validation.to_json())
+
+        if self.display is not None:
+            json_dict['display'] = self.display.to_json()
+
+        return json_dict
+
+    def __eq__(self, other):
```
collisdigital, line 43: You've got to be careful with this, anywhere you'd normally use a == (or code that performs the same comparison) is going to call this code and not compare the object pointers; which may not be obvious to the caller and give strange behaviour if not expected, if you do stick with it then you probably need pretty good unit tests around the operator to ensure it is working correctly (i.e. verifies each change in property results in a not equal outcome etc.). This seems like a lot of code to create and maintain just to test the json stuff; do you really want to do/keep these methods, another approach would be to run this through the to_json method, dump the resulting JSON text and paste that text into the unit test to verify that the code generates the expected json text or manually in the unit test check the properties of the objects etc. Worth discussing with @dhilton @LJBabbage ?

app/surveys/mci/mci_0205.py
```diff
@@ -0,0 +1,353 @@
...
+s2.description = "<p>- You should enter figures for the reporting period stated above<br>- You should round your figures to the nearest (£) pound</p><h4>Include</h4><ul><li>VAT</li><li>Internet Sales</li></ul>"  # NOQA
+
+block.add_section(s2)
+questionnaire.register(s2)
+
+q2 = Question()
+q2.id = "eedb9f9f-2f2e-41f4-9186-ba7110d9e6a4"
+q2.title = ""
+q2.description = ""
+q2.type = "Currency"
```
collisdigital, line 112: In the latest agreed model all Questions would be generic/general and only specialised question types like DateRange would have a different type. Maybe not to do now but at some point (sooner rather than later) we should make the switch to this approach.

app/surveys/mci/mci_0205.py
```diff
@@ -0,0 +1,353 @@
...
+
+s2.add_question(q2)
+questionnaire.register(q2)
+
+q2r = Response()
```
collisdigital, line 117: We discussed the possibility of the different Response types having their own classes (e.g. CurrencyResponse) etc. Another one of those when would be the right time to make that jump?

app/surveys/mci/mci_0205.py
```diff
@@ -0,0 +1,353 @@
...
+q3r = Response()
+q3r.id = "fee0b9fe-4c3a-4c14-9611-4fa9e2e9578a"
+q3r.code = "23"
+q3r.label = "What was the value of the business's total sales of alcohol, confectionery and tobacco?"
+q3r.guidance = "<div> <h4>Include</h4> <ul> <li>alcoholic drink</li> <li>chocolate and sugar confectionery</li> <li>tobacco and smokers’ requisites</li></ul> </div>"  # NOQA
+q3r.type = "Currency"
+q3r.options = []
+q3r.mandatory = False
+q3r.messages = {
+    "NOT_INTEGER": "Please only enter whole numbers into the field.",
```
collisdigital: Are these keys stored anywhere, I mean "NOT_INTEGER" that could be used instead (enum?) instead of pasting in the strings directly?

app/model/block.py
```diff
@@ -10,11 +10,73 @@ def __init__(self):
...
+        self.questionnaire = None
         self.templatable_properties = []
         self.display = Display()
 
     def add_section(self, section):
         if section not in self.sections:
             self.sections.append(section)
             section.container = self
+
+    def to_json(self):
```
collisdigital, line 22: When we did the design review with Warren it was agreed to also implement a from_json method that did the reverse operation so you could do a round trip to/from JSON, is this still in scope?

app/model/block.py
weapdiv-david, line 43: Definitely worth discussing with the larger team for the reasons you mentioned.  I had to change a number of the unittests to ensure that some objects being created were no longer equivalent, where the tests had simply required that they were not the same object.

I'm open to alternative approaches

app/surveys/mci/mci_0205.py
weapdiv-david, line 112: Agreed this needs to happen soon, but was left out of this in order to limit the scope.  Otherwise, the parser needs to change, etc...

app/surveys/mci/mci_0205.py
weapdiv-david, line 117: Similar to above, this was left out _for now_ in order to limit the scope and minimise the impact of other parts of the system.  It should probably be done sooner rather than later though, but perhaps a different PR

app/surveys/mci/mci_0205.py
weapdiv-david: My Bad!  They are stored as constants... I'll refactor them now

app/model/block.py
weapdiv-david, line 22: Could still be in scope... I'm not sure.  Currently the to_json() method works to output json in the format expected by the parser, so the round trip can be done through that as seen in the unit-tests.

It's not quite as clean as having a from_json() method, but while so much of this is up for grabs, I'm not sure how much value it adds at this stage to the implement that method.

Any thoughts from @dhilton, @LJBabbage ?

app/model/block.py
dhilton, line 43: Reading around the subject, I still think this is the right approach for this instance. We also need to add `__ne__` for completeness, but given we're writing a serialisation / deserialisation from JSON we need to be able to compare the objects generated correctly. 

I'm trying to think of an example where we'd compare two objects that would require us to check the internal python id of that object (which is how python does this check by default when comparing objects).

weapdiv-david, line 43: `__ne__` has been implemented for each of the model objects.

weapdiv-david, line 43: With regard to doing a comparision at json string level, that approach didn't work as the order in which the properties appeared in the json seemed to vary, making the string comparison impossible without resorting to lots and lots of fragile regexps.

Parsing the generated schema and doing an object model comparison seemed like the better way forward but comes with the caveats discussed above around equivalence vs equality

weapdiv-david, line 43: As for checking the internal id, the overridden `__eq__` method check that check first as a shortcut before comparing the equivalence of properties (which are also compared for equivalence).

The main issue to be aware of is with collections.  Two equivalent but separate objects may not be stored in a set, for example.  Also, if an object is already in a list, checking that a different but equivalent object is in the list will return true, even through they are different objects

collisdigital, line 43: If you stick with the **eq** approach also check **hash** e.g. https://docs.python.org/3/reference/datamodel.html#object.__hash__ and I believe (but could be wrong that **ne** by default calls **eq** and inverts the result (i.e. no need to explicitly add it). Also consider better unit tests to cover the operators.

app/model/block.py
dhilton, line 22: I think right now unless you've this locally, let's cut that from this story and migrate to a new one.

dhilton:
LGTM - 👍

PR #245, opened 2016-06-09
EQ-382: Dynamic Employment Date

LJBabbage:
### Changes

Add employment_date into jwt token decode (default if not present) and pipe result into 1_0215.json. Update dev page to include option to change employment_date
### How to test

Use dev to select 1_0215 and change the employment_date at the bottom, make sure the survey works and that the date selected is piped into the employment section
### Who can test

Anyone but @LJBabbage


dhilton:
LGTM - can we squash some of the commits @LJBabbage ? just makes the commit history cleaner and clearer to future devs.

PR #252, opened 2016-06-15
eq 303, create mci-0215, write build time json, update python model, …

LJBabbage:
### Changes

Created 0215 and 0213 (update 0203, 0205 and star wars) to python model. Introduce build time json creator. Fix 
### How to test
1. Run tests
2. Open surveys 0215, 0213, 0203, 0205 and Star Wars in app  making sure the models are being generated and the front end works as expected
### Who can test

Anyone but @LJBabbage


generate_json.py
```diff
@@ -0,0 +1,48 @@
...
+
+def write_json_to_file(json_name, json_survey):
+    try:
+        path = os.path.join(JSON_PATH, json_name + ".json")
+        f = open(path, "w")
+        f.write(json_survey)
+        f.close()
+
+    except IOError:
+        raise IOError("Can't write file to disk")
```
warrenbailey: I'm not sure there's any point catching and re-throwing the same exception?

app/surveys/mci_0213.py
```diff
@@ -0,0 +1,20 @@
+import copy
+
+from app.surveys.mci_0215 import mci_0215
+
+eq_id = "1"
+form_type = "0213"
+
```
warrenbailey, line 7: Would this be easier the other way round? I.e. add the additional question on 0215?

generate_json.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+if __name__ == '__main__':
+    # Loop through all the files in the survey folder, generating json for each, before writing to disk
+    for filename in os.listdir(PYTHON_MODEL_PATH):
```
warrenbailey, line 47: Might be nice to refactor this out to a method called generate_all_json()

warrenbailey:
Apart from a few very minor things I'm happy with this.

LGTM 👍

app/surveys/mci_0213.py
LJBabbage, line 7: Was done due to question numbering and position, if we created 0213 then added a question to get to 0215, getting the number and position in the json is far more complicated

warrenbailey:
Thanks for doing the review comments 👍

Issue #256, opened 2016-06-16
Summary does not include Comments


stevevic:
### Expected behaviour

The Summary screen displays any comments made by respondents (or N/A if no comments provided)
### Actual behaviour

Row for comments and any comments are not displayed on Summary screen
### Steps to reproduce the behaviour

Complete survey and submit through to summary page
#### Browser
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] MS Edge
- [ ] Internet Explorer 11
- [ ] Internet Explorer 10
- [ ] Internet Explorer 9
- [ ] Internet Explorer 8
- [ ] Internet Explorer 7
- [ ] Internet Explorer 6
- [ ] Other (please state below)
#### Operating System
- [x] Windows
- [ ] OS X
- [ ] iOS
- [ ] Android
### Screenshots

Comments being added:
![image](https://cloud.githubusercontent.com/assets/17979130/16119478/9536e6ae-33d3-11e6-93da-633893f68794.png)

Summary screen without comments:
![image](https://cloud.githubusercontent.com/assets/17979130/16119504/b5d01b06-33d3-11e6-84cd-883f08d22acd.png)

PR #261, opened 2016-06-17
Eq 303 text changes visuals and issue 256

LJBabbage:
### Changes

to fix issue  #256 
### How to test

Make sure the wording is correct for 1_0215 and that if a comment is made in the comment box it appears on the summary page
### Who can test

Anyone but @LJBabbage


warrenbailey:
LGTM 👍

PR #272, opened 2016-06-23
Support schema files names with no form type

warrenbailey:
### Changes

Updating the dev page so we can support schema files which don't have the form_type in their name. Also changing the default value for EQ_SCHEMA_BUCKET so that it matches the author application.
### How to test
1. Check out this branch
2. Load a schema file from s3 which is of the form (1.json) - you can use the author application to do this
3. Check you can complete the survey
### Who can test

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,8 @@ sudo: required
 dist: trusty
 python:
 - '3.5'
+env:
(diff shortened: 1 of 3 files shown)


weapdiv-david:
LGTM :+1:

PR #271, opened 2016-06-23
Improvements to Star wars tests

LJBabbage:
### Changes

Improvements to Stars Integration tests
### How to test

Run script/run_tests.sh and make sure they work
### Who can test

Anyone but @LJBabbage

PR #274, opened 2016-06-24
Remove un-used json files and absorb multi page test into Star Wars

LJBabbage:
### Changes

Remove the non python model generated json from app/data and absorb the missing backwards test of multi pages. (N.B the others which are being deleted  are now tested in #271 as we are not going to maintain 999.json)
### How to test

Make sure the tests run successfully
### Who can test

Anyone but @LJBabbage


dhilton:
I've rebased this - looks good to me once CI passes 👍

PR #275, opened 2016-06-24
Only show error messages related to page

### LINKED ISSUE #254: Multi page error messages
### Expected behaviour

Only the errors for a page should be shown
### Actual behaviour

If you generate an error on the second page then go back to the first page and cause an error on that page both will be displayed on the first page

See above
#### Browser

Chrome
#### Operating System

Windows

weapdiv-david:
### Changes

Fixes #254 Multi page error messages
### How to test

1) Checkout the branch
2) Run the app and choose a multi-page validated survey from the dev console (999 is a good one)
3) Complete the first page of the survey.
4) On the second page of the survey trigger a validation error
5) Once the validation error is triggered, go back to the first page
6) Trigger a validation error on the first page also
7) Confirm that you only see the errors applicable to the current page
### Who can test

Anyone except @weapdiv-david


tests/integration/multipage/test_errors.py
```diff
@@ -0,0 +1,145 @@
+import unittest
+from app import create_app
+from tests.integration.create_token import create_token
+
+
```
warrenbailey: All integration tests now need to extend IntegrationTestCase as that setups up an in-memory sql database. otherwise the tests will just fail

app/renderer/renderer.py
```diff
@@ -24,9 +26,17 @@ def __init__(self, schema, response_store, validation_store, navigator, metadata
             end_date = self._metadata.get_ref_p_end_date()
             employment_date = self._metadata.get_employment_date()
             return_by = self._metadata.get_return_by()
+
         except:
             pass
 
+        try:
+            self._current_block = self._schema.get_item_by_id(self._navigator.get_current_location())
+        except:
```
warrenbailey: This should catch QuestionnaireException rather than catch all.

app/renderer/renderer.py
```diff
@@ -150,16 +158,22 @@ def _augment_questionnaire(self):
...
+            if self._current_group.id == group.id:
+                self._augment_item(group, errors, warnings)
+                for block in group.blocks:
+                    if self._current_block.id == block.id:
+                        self._augment_item(block, errors, warnings)
+                        for section in block.sections:
+                            self._augment_item(section, errors, warnings)
+                            for question in section.questions:
+                                self._augment_item(question, errors, warnings)
+                                for response in question.responses:
```
warrenbailey, line 170: Do you actually need this last for loop if you're augmenting all the response in the next step anyway. Seems like duplication?

app/renderer/renderer.py
```diff
@@ -129,10 +139,8 @@ def _render_navigation_meta(self):
...
 
         try:
-            current_block = self._schema.get_item_by_id(self._navigator.get_current_location())
-            current_group = current_block.container
-            navigation_meta["current_block_id"] = self._navigator.get_current_location()
-            navigation_meta["current_group_id"] = current_group.id
+            navigation_meta["current_block_id"] = self._current_block.id
+            navigation_meta["current_group_id"] = self._current_group.id
         except:
             pass
```
warrenbailey: Confused as to why you need this except all block?

warrenbailey:
This class could really benefit from some documentation and logging.

app/renderer/renderer.py
weapdiv-david, line 170: Unfortunately yes.  We only need to build the error messages for the current group/block, but the system crashes unless _all_ responses are augmented due to the summary page needing them.

This is one of the main reasons why I wanted a design review

app/renderer/renderer.py
weapdiv-david: Will remove.  Was originally needed when calling get_item_by_id(...)

LJBabbage:
These tests need to be moved into Star wars as agreed by the team. Survey 999 no longer exists

app/renderer/renderer.py
warrenbailey: Yeah but aren't you doing it twice for the responses on that block?

warrenbailey, line 170: Yeah but aren't you doing it twice for the responses on that block?

warrenbailey, line 170: Ah nevermind. I see the difference now, its the responses from the schema and the response store responses. They are different things (this is why hated that naming convention). Ok this makes sense now.

warrenbailey:
I think these git commits could be squashed, especially the 'refactor test' ones.

warrenbailey:
I still think this class needs better documentation and logging, but we'll address this as part of the Software Design story. Therefore I'm going to merge this. 👍

PR #279, opened 2016-06-27
Rabbit MQ - Persist messages between restarts

warrenbailey:

Changing the submitter so that it sends persistent messages to rabbit mq and sets the queue to be durable. This means messages are not lost on restart of the Rabbit MQ application as they are now written to disk.
### How to review
1. Check out this branch
2. Run a local copy of rabbit MQ
3. Set the environment variable `export EQ_RABBITMQ_ENABLED=True`
4. Complete a survey
5. Check a message exists on the Rabbit MQ server (`/rabbitmqctl list_queues`)
6. Restart the Rabbit MQ server and check the message still exists
### Who worked on the PR

@warrenbailey

### DIFF
--- a/app/submitter/submitter.py
+++ b/app/submitter/submitter.py
@@ -94,8 +94,14 @@ class RabbitMQSubmitter(Submitter):
         try:
             self._connect()
             channel = self.connection.channel()
-            channel.queue_declare(queue=queue)
-            published = channel.basic_publish(exchange='', routing_key=queue, body=message_as_string, mandatory=True)
+            channel.queue_declare(queue=queue, durable=True)
+            published = channel.basic_publish(exchange='',
+                                              routing_key=queue,
+                                              body=message_as_string,
+                                              mandatory=True,
+                                              properties=pika.BasicProperties(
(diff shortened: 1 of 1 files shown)


app/submitter/submitter.py
```diff
@@ -94,8 +94,14 @@ def send_message(self, message, queue):
...
-            channel.queue_declare(queue=queue)
-            published = channel.basic_publish(exchange='', routing_key=queue, body=message_as_string, mandatory=True)
+            channel.queue_declare(queue=queue, durable=True)
+            published = channel.basic_publish(exchange='',
+                                              routing_key=queue,
+                                              body=message_as_string,
+                                              mandatory=True,
+                                              properties=pika.BasicProperties(
+                                                  delivery_mode=2
+                                              ))
```
dhilton, line 104: Does this also need to make use of publisher confirms? https://www.rabbitmq.com/confirms.html That will (according to  documentation) block until the message is persisted to disk. That would at least give us an option of redisplaying to the user to try submission again (not that we currently support this).

Otherwise LGTM 👍

Issue #280, opened 2016-06-27
When generating json schemas from code, data directory needs to be present or dev page fails.


dhilton:
### Expected behaviour

When you hit `/dev` you're presented with a development page in development environments and preprod.
### Actual behaviour

That page error's 500.
### Steps to reproduce the behaviour

Visit preprod url `/dev`
### Technical information

Logs report an error due to lack of `data` directory.
#### Browser

ALL
#### Operating System

ALL

warrenbailey:
It's broken again, elastic beanstalk ignores the .gitignore file

dhilton:
OK - revert the .gitignore change to the README change but keep the filename filter aspect. Bother.

dhilton:
This has been fixed. Closin'

PR #282, opened 2016-06-28
Replacing README with .gitignore

warrenbailey:

Fixs #280 - Replacing README and removing hidden files from the dev page
### How to review

1. Check out this branch and check the app/data folder exists
2. Run the application and check that .gitignore is not present in the dev page
### Who worked on the PR

@warrenbailey

### DIFF
--- a/app/schema_loader/schema_loader.py
+++ b/app/schema_loader/schema_loader.py
@@ -52,7 +52,7 @@ def available_local_schemas(files):
     for file in os.listdir(settings.EQ_SCHEMA_DIRECTORY):
         if os.path.isfile(os.path.join(settings.EQ_SCHEMA_DIRECTORY, file)):
             # ignore hidden file
-            if not file.startswith("."):
+            if file.endswith(".json"):
                 files.append(file)
 
 
--- a/app/schema_loader/schema_loader.py
+++ b/app/schema_loader/schema_loader.py
@@ -42,15 +42,25 @@ def load_schema(eq_id, form_type):
 
 def available_schemas():
     files = []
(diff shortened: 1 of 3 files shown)


app/schema_loader/schema_loader.py
```diff
@@ -42,15 +42,25 @@ def load_schema(eq_id, form_type):
...
-        except botocore.exceptions.ClientError as e:
-            logging.error("S3 error: %s", e.response['Error']['Code'])
+        available_s3_schemas(files)
     return files
+
+
+def available_local_schemas(files):
+    for file in os.listdir(settings.EQ_SCHEMA_DIRECTORY):
+        if os.path.isfile(os.path.join(settings.EQ_SCHEMA_DIRECTORY, file)):
+            # ignore hidden file
```
dhilton, line 54: I think a "Deny all, accept only json" approach should be used here as that is a recommended secure pattern. Allowing only json files is good, but we only do filename checks, not actual filetype checks but this will be addressed in the coming survey schema validation stories.

app/schema_loader/schema_loader.py
```diff
@@ -42,15 +42,25 @@ def load_schema(eq_id, form_type):
...
+            if not file.startswith("."):
```
dhilton: If we swap this for :

```
    if file.endswith(".json"):

```

That is a whitelist approach which should then clear out all other possible files (not that any files apart from json files should be in that directory / bucket).

warrenbailey:
Sounds good i'll make that change

dhilton:
LGTM 👍

PR #284, opened 2016-06-30
Renames "Response" to "Answer" as agreed by team

weapdiv-david:

Following our software design review the team felt there were a number of changes that needed to be made to the existing codebase to make it clearer to understand, and better support the features that we have planned.  This pull request is the first of a series of pull requests which address the changes we discussed and agreed.

Specifically, this PR changes the name of the "Response" object used in our data model to "Answer" in the hope of clearing up any confusion created when discussing responses.
### How to review

1) Check out this branch
2) Run the tests
3) Run the app and verify that it continues to function as before
### Who worked on the PR

@weapdiv-david worked on this PR, so it should be reviewed by all other members of the team.


hamishtaplin:
This should probably also change the class names used in the CSS as they correlate to the use of "response" as name.

weapdiv-david:
I haven't change the css names for two reasons...

1) There are further refactors to come around the Fields/Answers/Question classes and the relationships between them which will prompt an overhaul to the current template layout and structure
2) That @hamishtaplin and I will need to collaborate on those changes and evaluate whether any larger changes to markup and styling are required once those changes are done.

app/model/question.py
```diff
@@ -6,8 +6,8 @@ def __init__(self):
         self.id = None
         self.title = None
         self.description = ""
-        self.responses = []
-        self.children = self.responses
+        self.answers = []
```
warrenbailey, line 9: Is this correct? The schema isn't changing?

warrenbailey:
Maybe I misunderstood this, I thought we were renaming the response store to answer store but keeping the name response in the schema. Therefore keeping a clean separation between a response represented in the schema and the actual answer provided by the user?

At the minute it's difficult to distinguish between the two in the codebase

hamishtaplin:
@weapdiv-david Ok fair enough, happy to defer re-naming for now.

weapdiv-david:
@hamishtaplin cool, there are loads more changes to come following the design review.  I look forward to your input.

@warrenbailey Thanks for posting that up, I think it's important that we try and have a record of the discussion and the decisions made by the team around this area as the name "response" as been a huge source of confusion and it looks like it's going to continue for a while yet.

My view of a response is that it should be a singular piece of information provided by the user in response to a question.  Those responses need to be stored somewhere, and so we store them in a response_store.  With this PR, substitute the word "response" with "answer" and that model still holds true.

My understanding behind the need for this refactor was based on the confusion between a response (answer provided by a user in response to a question prompt) and the HttpResponse object that provides the low-level transport between client and server.

My view is that the response/answer defined in the schema should only define the data type of the information the user is to provide. e.g. Currency response, Integer response, Email response.  That response/answer can then be type-checked and/or stored for submission downstream or redisplay to the user.  I think that the classes involved in this should all use the same root naming to minimise further confusion.

When dealing with HttpResponse objects and the key-value pairs, there is already confusion around relating key-value pairs to response/answer objects and this is likely to get worse as we explore different relationships between questions/responses and the fields that illicit those responses from the user.

I await your response (lol) and look forward to contributions from the wider team

warrenbailey:
I agree the term response has too much mean in a web application and is easily confused with a HTTP response.

However I think we also need a clean separation between the response 'type' defined in the schema (e.g. IntegerResponse) and the actual response 'value'  (e.g. 1) provided by the user. It's all too easy to confuse them in the current code base. 

What we should call them is up for debate, but I think answer definitely should be one of them.

app/submitter/converter.py
```diff
@@ -95,15 +95,15 @@ def prepare_responses(user, metadata_store, questionnaire, responses):
...
+                   SubmitterConstants.ORIGIN_KEY: SubmitterConstants.ORIGIN,
+                   SubmitterConstants.SURVEY_ID_KEY: survey_id,
+                   SubmitterConstants.SUBMITTED_AT_KEY: submitted_at.strftime(settings.DATETIME_FORMAT),
+                   SubmitterConstants.COLLECTION_KEY: collection,
+                   SubmitterConstants.METADATA_KEY: metadata,
+                   SubmitterConstants.PARADATA_KEY: paradata,
+                   SubmitterConstants.DATA_KEY: data}
+
+        logging.debug("Converted answer ready for submission %s", json.dumps(payload))
+        return answer, submitted_at
```
warrenbailey: This should be payload not answer. I can't see how this even works?

warrenbailey:
The converter change doesn't work, see my comment

warrenbailey:
Thanks for fixing the converter. As agreed with the team this morning we'll merge this now and then next sprint spend time looking at Question and Answer Types in the next sprint. 👍

PR #291, opened 2016-07-04
Eq 399 software design add mccabe

dhilton:

To help us write and refactor our codebase to reduce complexity, we're implementing a CI test for the McCabe index (https://en.wikipedia.org/wiki/Cyclomatic_complexity). The aim of this check is to minimise the complexity of any individual function, which has been shown to be correlated to numbers of bugs in a codebase.

We're initially taking the baseline of 10 as our limit.
### How to review
1. Checkout this branch
2. Run the test scripts `./scripts/run_test.sh`
3. Check the output from Flake8 passes and that the McCabe index is set to 10.
### Who worked on the PR

@dhilton but some of the other function refactors originally found were done under different PR's.

### DIFF
--- a/app/authentication/jwt_decoder.py
+++ b/app/authentication/jwt_decoder.py
@@ -74,19 +74,7 @@ class JWTDecryptor(JWERSAOAEPDecryptor):
         self._check_payload(token_as_str)
         return
 
-    def _check_headers(self, token):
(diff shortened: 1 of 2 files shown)


weapdiv-david:
LGTM :+1:

Issue #288, opened 2016-07-04
Pipe an answer bug


LJBabbage:
### Expected behaviour

Summary page should show the piped answer in the question text
### Actual behaviour

It show the script, {answers.chewies_age}
### Steps to reproduce the behaviour

Complete Star wars survey

PR #294, opened 2016-07-04
Fixes issue #288 - Pipe an answer on summary page

weapdiv-david:

This change splits out the multiple roles of the TemplatePreProcessor into discrete functions and ensures those functions are called at the correct points in the users journey through a questionnaire.  This PR solves the issue #228 which showed that piped values were not being passed into questions when being displayed on the summary page.
### How to review

1) Checkout the branch and run the unit tests
2) Run the app and use the dev console to select the star wars survey
3) Fill in the survey and verify that Chewbaccas age is correctly displayed in the question text
4) Go to the summary page, and confirm that Chewbaccas age is _also_ displayed on the summary page
- [x] Is all new code unit tested?
- [x] Are there integration tests if needed?
### Who worked on the PR

@weapdiv-david worked on this PR

### DIFF
--- a/app/templating/template_pre_processor.py
+++ b/app/templating/template_pre_processor.py
@@ -46,7 +46,17 @@ class TemplatePreProcessor(object):
             return 'questionnaire.html'
 
     def build_view_data(self):
-        self._augment_questionnaire()
+        # Always plumb the questionnaire
+        self._plumb_questionnaire()
+
+        # Collect the answers for all pages except the intro and thank you pages
+        current_location = self._navigator.get_current_location()
+        if current_location != 'introduction' and current_location != 'thank-you':
+            self._collect_answers()
+
+            # We only need to augment the questionnaire if we are still inside it
+            if current_location != 'summary':
+                self._augment_questionnaire()
 
         render_data = {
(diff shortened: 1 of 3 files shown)


app/templating/template_pre_processor.py
```diff
@@ -46,7 +46,17 @@ def get_template_name(self):
             return 'questionnaire.html'
 
     def build_view_data(self):
-        self._augment_questionnaire()
+        # Always plumb the questionnaire
+        self._plumb_questionnaire()
+
```
LJBabbage, line 51: These 2 if statements are related, both requiring the user to be in the questionnaire, the only difference is the 2nd one just excludes the summary page. So I would consider

```
   # We only need the answers if we are in the questionnaire
    if current_location != 'introduction' and current_location != 'thank-you':
        self._collect_answers()

        # We don't need to augment the summary page
        if current_location != 'summary':
            self._augment_questionnaire()
```

LJBabbage:
A test in Star wars on the summary page for the value would be good

LJBabbage:
Looks good otherwise :thumbsup:

PR #304, opened 2016-07-07
Eq 399 Refactoring Navigator component

warrenbailey:

Refactoring the navigation and routing components and consolidating the user action functionality into a single place.

The navigation state object is now an actual state pattern, which maintains the current user position in the questionnaire.

The routing component was actually dealing with user actions which isn't its purpose. That should be handled elsewhere and it should only do routing (which we're yet to implement).

The questionnaire manager was also dealing with user actions making it difficult to isolate how they are processed.

The user actions are now handled by the navigator using a chain of responsibility pattern. This means adding a new user action should simply be a case of creating a new UserAction class and placing it in the correct location in the chain.

I appreciate this is a big refactor but I believe it leads us to a more elegant solution going forward. I'm open to comments as there's definitely still room for improvements.
### How to review
1. Check out this branch
2. Check all the tests still work
3. Complete a survey and check nothing is broken
### Who worked on the PR

@warrenbailey

### DIFF
--- a/app/navigation/navigator.py
+++ b/app/navigation/navigator.py
@@ -24,6 +24,8 @@ class UserAction(object):
         self._schema = schema
         self._metadata = metadata
         self.next_action = None
+        # subclasses should provide the name of the action it can support
+        self.action = None
 
     def set_next_action(self, action):
         self.next_action = action
@@ -31,7 +33,7 @@ class UserAction(object):
 
     def process_action(self, action, current_location):
         # work through the user actions to determine which one matches the request action
-        if action == self.get_action():
+        if action == self.action:
             # perform any functionality necessary to complete that action
             self.perform_action()
             # then provide the next location
@@ -39,10 +41,6 @@ class UserAction(object):
         else:
             # request the next location
             return self.next_action.process_action(action, current_location)
-
-    def get_action(self):
-        # subclasses should provide the name of the action it can support
-        pass
 
     def perform_action(self):
         # subclasses should perform any functionality necessary to execute the user action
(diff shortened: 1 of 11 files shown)


warrenbailey:
Adding a WIP label but i'd still value some feedback on it, once I'm happy with it I'll squash the commits ready for merge.

weapdiv-david:
Looks good so far, although I don't fully understand it all yet.  It makes perfect sense to split the user action processing out from the data, and I'm currently looking at better ways to process the form data in view of the changes we want to make around fields/responses/questions.

Keep going!

warrenbailey:
Thanks for the feedback @weapdiv-david, if anyone else has time to review I'd appreciate it. Otherwise I'll fix the minor quantified code issues and squash the commits ready for merging.

warrenbailey:
Rebased and squashed the commits ready for merge. I liked to take this further but wonder if it would be better to merge this now and then make another incremental refactor, otherwise this risks getting out of date with master very quickly.

weapdiv-david:
Happy to merge this... Any other team members wanting to review it? (clue: you should)

app/navigation/navigator.py
```diff
@@ -1,82 +1,156 @@
...
+            return self.provide_next_location(current_location)
+        else:
+            # request the next location
+            return self.next_action.process_action(action, current_location)
 
-    # destination  "group:block:section:question:<repetition>"
-    def _valid_destination(self, destination):
-        """
-        Returns a boolean indicating whether a proposed destination is a valid one
+    def get_action(self):
```
dhilton: Does this getter always return a string? If so, is it not a property that is set rather than a method?

dhilton:
Apart from my comment above 👍 from me.

app/navigation/navigator.py
warrenbailey: Yep always returns a string. I'll change it to be a property.

warrenbailey:
Fixing code review comments and rebasing

dhilton:
Happy to merge this LGTM 👍

PR #306, opened 2016-07-08
Eq 399 improve test coverage

weapdiv-david:

Minor improvements to test coverage, fixes a bug that was uncovered during this.
### How to review

1) Checkout branch
2) Run unit tests
### Who worked on the PR

@weapdiv-david worked on this


app/validation/mandatory_check.py
```diff
@@ -16,16 +16,21 @@ def validate(self, user_answer):
...
+
+            # No valid values in list
+            validation_result.errors.append(AbstractValidator.MANDATORY)
         else:
             if user_answer and not str(user_answer).isspace():
                 validation_result.is_valid = True
             else:
                 validation_result.is_valid = False
                 validation_result.errors.append(AbstractValidator.MANDATORY)
+
```
warrenbailey, line 25: There is duplicated functionality in this method, I think it would be better to extract a method and have the loop repeatedly call it. For example:

```
    def validate(self, user_answer):
        if isinstance(user_answer, list):
            for answer in user_answer:
                validation_result = self._validate(answer)
                if validation_result.is_valid:
                    break
        else:
            validation_result = self._validate(user_answer)

        return validation_result

    def _validate(self, user_answer):
        validation_result = ValidationResult()

        if user_answer and not str(user_answer).isspace():
            validation_result.is_valid = True
        else:
            validation_result.is_valid = False
            validation_result.errors.append(AbstractValidator.MANDATORY)

        return validation_result
```

warrenbailey:
I think the code in the Mandatory check could be simplified and refactored out into an additional method. Happy with the rest of it though.

warrenbailey:
Excellent, I think that new version is better 👍

PR #313, opened 2016-07-15
EQ 410 User Journey Manager

warrenbailey:

The introduction of the User Journey Manager. This is essentially a data structure for keeping track of a users interaction with a survey. It uses a doubly linked list to maintain the position in the survey and state objects to store information about each page, such as the responses the user has answered to the question, validation state and the typed value of each response.

This change has also refactored the current answer store to use the new data structure. This is a temporary solution until a wider refactor happens making the answer store obsolete.

The next step is to refactor the navigation and validation components to use this list. However in order to make this pull request manageable it is desirable to review and merge the refactoring at this point. As such there are various TODO scattered around the code base.
### How to review
1. Confirm all the tests and integration tests pass
2. Check you can still complete a survey successfully.
### Who worked on the PR

@warrenbailey and @weapdiv-david both worked on this PR. So this change needs to be reviewed by @dhilton and @LJBabbage


app/answers/answer_store.py
```diff
@@ -25,29 +30,36 @@ def clear_answers(self):
...
-        if ANSWERS not in data.keys():
-            data[ANSWERS] = {}
-            return None
-        if key not in data[ANSWERS].keys():
+        # ujm = UserJourneyManager.get_instance()
+        # current_state = ujm.get_current_state()
+        # answer = current_state.get_answer(key)
+        # return answer.input
+        #
+        answers = self.get_answers()
```
dhilton, line 41: Why not have answers as a property of an AnswerStore? That way this code code be refined (even removed as you could then call `answerStoreInstance.answers[key]` which assumes that .answers has a dict like interface but would simplify this code and other instances of it.

app/questionnaire_state/item.py
```diff
@@ -0,0 +1,28 @@
...
+        self.children = None
+        self.is_valid = None
+        self.errors = None
+        self.warnings = None
+
+    def update_state(self, user_input):
+        for child in self.children:
+            child.update_state(user_input)
+
+    def get_answer(self, id):
```
dhilton, line 16: This will always return the first found answer - is there any likelyhood that this _won't_ be the answer? Should it check through all possibles and then return? Raise an exception? Return both values?

app/questionnaire_state/question.py
```diff
@@ -1,6 +1,23 @@
...
     def __init__(self, id):
-        self.id = id
+        super().__init__(id=id)
         self.answers = []
+        self.children = self.answers
+
+    def get_answers(self):
+        answers = []
+        for child in self.children:
+            answers.append(child.get_answer())
```
dhilton: Do you mean append or do you mean to use extend? get_answers seems to return a list even if a single value is found, so I think you end up with a list of lists rather than just a single list of answers. Also I can't see the difference between this method and the parent method on the Item class shown in https://github.com/ONSdigital/eq-survey-runner/pull/313/files#diff-5f2cdd4ed301f540243d3474c5f36f01R24

app/answers/answer_store.py
warrenbailey, line 41: That makes a lot sense and I would be open to that change. However AnswerStore is one of the interfaces we're planning to remove shortly. Therefore it's probably not worth refactoring the code to do that at the minute.

app/questionnaire_state/item.py
warrenbailey, line 16: I wasn't fussed on this method either. I will only return an answer and it returns the first one it encounters. However I take your point about checking that is definitely an answer. I'll have a rethink on this one.

app/questionnaire_state/question.py
warrenbailey: At the question level it is .append rather than .extend as it's calling child.get_answer() rather than child.getanswers().

On reflection this isn't clear and maybe returning a list containing a single element would be clearer, keeping the interface clean and allowing this class to use the parent method.

dhilton:
I can't seem to find a test for the json pickle of an object. My worry is we change the object and that jsonpickle can't handle the serialisation and we get an issue. If we have a test for that serialisation of a survey state, as we update that object we should catch if that object can't be jsonpickle serialisation issue.

warrenbailey:
Agree thats a good idea, i'll add one.

warrenbailey:
I've address your review comments, apart from the first one. There's still work to do but I believe this is a good step in the right direction.

About to rebase

dhilton:
LGTM 👍  Mergin'

PR #318, opened 2016-07-19
Eq 410 navigator

warrenbailey:

Removing the navigator and replacing it with the User Journey Manager. Also cleaning up the User action processor and moving the routing components into the routing engine. 
### How to review
1. Check out this branch
2. Check the application still works
3. Ensure all tests pass
### Who worked on the PR

@warrenbailey


app/questionnaire_state/user_journey_manager.py
```diff
@@ -63,22 +77,33 @@ def go_to_state(self, item_id):
             self._create_new_state(item_id)
 
     def _create_new_state(self, item_id):
-        item = self._schema.get_item_by_id(item_id)
+
         logger.debug("Creating new state for %s", item_id)
-        if isinstance(item, SchemaBlock):
-            state = StateBlock.construct_state(item)
+        if item_id in self._valid_locations:
```
dhilton, line 80: I wonder if there is any way we can remove this case style statement and use polymorphism? 
http://refactoring.com/catalog/replaceConditionalWithPolymorphism.html

app/routing/routing_engine.py
```diff
@@ -1,10 +1,40 @@
...
     ''' The routing engine will apply any routing rules dependant on where the user is in the schema
     and what answers they have provided.
     '''
     def __init__(self, schema_model):
         self._schema = schema_model
 
-    def get_next(self, current_location):
-        # however we're yet to implement routing so always return the current location
-        return current_location
+    def get_next_location(self, current_location):
```
dhilton, line 12: When we talk about the next location, we really mean the next block to display? Would it ever be anything else?

warrenbailey, line 12: Technically no, but the system is flexible enough to cope with us changing it to question or section for example. That's why we haven't called it get_next_block

app/questionnaire_state/user_journey_manager.py
warrenbailey, line 80: That doesn't really work in this case as item isn't a StateBlock its SchemaBlock. We could modify the schema model objects to have a construct state method, which could work.

dhilton:
Happy to merge this now?

warrenbailey:
Yep am now

PR #319, opened 2016-07-19
Eq 410 composite question types

weapdiv-david:

This PR builds upon PR #318 

This PR replaces all questions in all schemas with either a "General" or "DateRange" question type.  Two question subclasses are created to represent them.  The parser is also updated to throw an exception when encountering a question type that is NOT one of these types. It also subclasses the answer types currently supported.  Existing templates are not not affected.

This PR lays one of the foundation stones for the larger display-widget refactoring that's currently taking place.
### How to review

1) Check out the branch
2) Run the unit tests
3) Run the app and ensure the questionnaires are rendered as before
### Who worked on the PR

### DIFF
--- a/app/schema/answer.py
+++ b/app/schema/answer.py
@@ -25,44 +25,5 @@ class Answer(Item):
     def get_state_class(self):
         return State
 
-    @staticmethod
-    def get_instance(answer_type):
-        # Do the imports here to avoid circular dependencies
-        # TODO: Refactor out into a factory of some sort
-        from app.schema.questionnaire import QuestionnaireException
-
-        from app.schema.answers.checkbox_answer import CheckboxAnswer
-        from app.schema.answers.currency_answer import CurrencyAnswer
-        from app.schema.answers.date_answer import DateAnswer
-        from app.schema.answers.integer_answer import IntegerAnswer
-        from app.schema.answers.percentage_answer import PercentageAnswer
-        from app.schema.answers.positiveinteger_answer import PositiveIntegerAnswer
-        from app.schema.answers.radio_answer import RadioAnswer
-        from app.schema.answers.textarea_answer import TextareaAnswer
-        from app.schema.answers.textfield_answer import TextfieldAnswer
-
-        answer_type = answer_type.upper()
-
-        if answer_type == 'CHECKBOX':
-            return CheckboxAnswer()
-        elif answer_type == 'CURRENCY':
-            return CurrencyAnswer()
-        elif answer_type == 'DATE':
-            return DateAnswer()
-        elif answer_type == 'INTEGER':
-            return IntegerAnswer()
-        elif answer_type == 'PERCENTAGE':
-            return PercentageAnswer()
-        elif answer_type == 'POSITIVEINTEGER':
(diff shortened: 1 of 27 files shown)


app/schema/answer.py
```diff
@@ -15,3 +14,45 @@ def __init__(self):
         self.templatable_properties = []
         self.options = []
         self.alias = None
+
+    @staticmethod
+    def get_instance(answer_type):
+        # Do the imports here to avoid circular dependencies
+        # TODO: Refactor out into a factory of some sort
```
warrenbailey: Maybe add the factory now, this looks pretty messy.

app/schema/answer.py
```diff
@@ -15,3 +14,45 @@ def __init__(self):
...
+        from app.schema.answers.integer_answer import IntegerAnswer
+        from app.schema.answers.percentage_answer import PercentageAnswer
+        from app.schema.answers.positiveinteger_answer import PositiveIntegerAnswer
+        from app.schema.answers.radio_answer import RadioAnswer
+        from app.schema.answers.textarea_answer import TextareaAnswer
+        from app.schema.answers.textfield_answer import TextfieldAnswer
+
+        answer_type = answer_type.upper()
+
+        if answer_type == 'CHECKBOX':
```
warrenbailey: If we implemented the factory we could avoid all these elif block by implementing a register/create pattern

For example:

```
class AnswerFactory(object):
    def __init(self):
         self.classes = {}
    def register(self, name, class):
         self.classes[name] = class

    def create(name):
        return self.classes[name]()
```

Then register all the types like this:

```
answer_factory.register("CURRENCY", CurrencyAnswer)
```

app/schema/answer_factory.py
```diff
@@ -0,0 +1,34 @@
...
+from app.schema.answers.integer_answer import IntegerAnswer
+from app.schema.answers.percentage_answer import PercentageAnswer
+from app.schema.answers.positiveinteger_answer import PositiveIntegerAnswer
+from app.schema.answers.radio_answer import RadioAnswer
+from app.schema.answers.textarea_answer import TextareaAnswer
+from app.schema.answers.textfield_answer import TextfieldAnswer
+
+
+class AnswerFactory(object):
+    @staticmethod
```
collisdigital: There's a generic factory in (https://github.com/ONSdigital/eq-survey-runner/blob/master/app/utilities/factory.py) that does this, you can just create one of them here and register the types e.g

AnswerFactory = factory.Factory()
AnswerFactory.register_all(...)

app/schema/question_factory.py
```diff
@@ -0,0 +1,19 @@
+from app.schema.questionnaire import QuestionnaireException
+from app.schema.questions.general_question import GeneralQuestion
+from app.schema.questions.date_range_question import DateRangeQuestion
+
+
+class QuestionFactory(object):
```
collisdigital: See other comment about https://github.com/ONSdigital/eq-survey-runner/blob/master/app/utilities/factory.py

warrenbailey:
LGTM 👍

PR #326, opened 2016-07-21
captures issue with None employment date and correctly logs exception

weapdiv-david:

During the course of testing we discovered that if RRM did not send the employment date and the schema required it, then the runner would crash.

This PR increased the logging around the piping component and it nows correctly throws an exception under these circumstances.
### How to review

1) Checkout the branch
2) Run the tests and ensure they all pass
### Who worked on the PR

@weapdiv-david worked on this

### DIFF
--- a/app/piping/plumber.py
+++ b/app/piping/plumber.py
@@ -1,4 +1,8 @@
 import re
+import logging
+
+
+logger = logging.getLogger(__name__)
 
 
 class Plumber(object):
@@ -36,10 +40,16 @@ class Plumber(object):
         Perform the string replacement directly on the object
         '''
         if hasattr(item, templatable_property):
+            logger.debug('Piping property "{}" of item "{}"'.format(templatable_property, item.id))
             try:
                 template = getattr(item, templatable_property)
                 formatting_data = self._context
                 plumbed_value = template.format(**formatting_data)
                 setattr(item, templatable_property, plumbed_value)
             except KeyError:
+                logger.warn('Property "{}" cannot be piped'.format(templatable_property))
                 pass    # Do nothing, leave the propery as is
+            except Exception as e:
+                logger.error("An exception has been thrown")
+                logger.exception(e)
+                raise e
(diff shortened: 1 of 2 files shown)


app/piping/plumber.py
```diff
@@ -36,10 +40,16 @@ def _plumb(self, item, templatable_property):
...
             try:
                 template = getattr(item, templatable_property)
                 formatting_data = self._context
                 plumbed_value = template.format(**formatting_data)
                 setattr(item, templatable_property, plumbed_value)
             except KeyError:
+                logger.warn('Property "{}" cannot be piped'.format(templatable_property))
                 pass    # Do nothing, leave the propery as is
+            except Exception as e:
+                logger.error("An exception has been thrown")
```
warrenbailey: Not sure how useful this error message is?

Maybe have something similar to the one in KeyError:

 logger.error('Property "{}" cannot be piped'.format(templatable_property) as it is missing)

weapdiv-david:
I wasn't sure what this message should be.  It's an exceptional circumstance, but the property simply not being there is handled in the above except block.  The only current circumstances which trigger this are trying to pipe a date object that should be there, but isn't a date. That to me seems too specific to be handled here, but is implemented in this way as the rest of the system can't currently guarantee the dates availability.

Longer term, I would rather remove this exception handling altogether as I don't think it really belongs here (missing values should simply be ignored).  While I agree the message is not helpful, the corresponding stack trace is intended to provide more information.  

A slightly more verbose but accurate error message might be...

`Data required to pipe property "{}" of item "{}" is invalid.  The template is "{}"`

Which would identify the item, the property, and the property value but still doesn't give access to the type of data that's causing the problem.  It would, however, point us in the right direction.

Thoughts??

warrenbailey:
At least with that more verbose error message we can immediately identify the properties, values and template that has caused the error which will be displayed in the stack trace.

weapdiv-david:
Agreed, although I think this issue particularly needs to be addressed at service level.  I'll make the changes above and push it up.

warrenbailey:
What we really want is someway of telling RRM what the mandatory values are. Something like a Survey Registry might come in useful then ;-)

weapdiv-david:
A survey registry?  What a wonderful idea :)

dhilton:
@weapdiv-david @warrenbailey you two and your logic 😆

warrenbailey:
LGTM 👍

PR #338, opened 2016-07-27
Eq 410 validation refactor

weapdiv-david:

This PR is a work-in-progress.  It is being created to allow other members of the team to review the large scale changes that are taking place within the appliation and provide a means for team members to comment and arrange their thoughts.

So what's changed? The short answer is "pretty much everything", but the longer answer is as follows:
- Transitioning the application from "Thin Model" to "Thick Model"
- Removal of the central `validator` component
- Specific subclasses of `Question`, `Answer`, and `Widget` have been introduced
- Type-checking of incoming user data is now performed by those subclasses
- User state is managed by the `Answer` subclasses
- Error messages are handled by the `Questionnaire` schema component allowing error messages to be default, provided by the questionnaire, or at the component level
- Refactoring of the `TemplatePreProcessor` to combine `State` objects with `Schema` objects prior to rendering, in order to mininize the amount of rework to the template layer
- Collecting error messages across components has been delegated to the "Thick Model" rather than taking place within the `TemplatePreProcessor'
### Still to do
- [x] Tidy up the API around error message propagation and storage
- [x] Handle `DateAnswer`s correctly
- [x] Implement `DateRangeAnswer` special case validation
- [x] Implement the Summary Page
- [x] Identify and remove unneeded unit tests
- [ ] Identify and implement tests in need of extension

At that point, the functionality offered should be comparable with the code base previously, but with the groundwork laid to support additional widgets, and question types and vastly simplify navigation, routing and repetition.
### How to review

1) Check out the branch
2) Run the failing unit tests
3) Run the app
4) Post comments against the PR

### Who worked on the PR


app/questionnaire_state/answer.py
```diff
@@ -9,9 +9,33 @@ def __init__(self, id):
...
-        if self.id in user_input.keys():
-            self.input = user_input[self.id]
+    def update_state(self, user_input, schema_item):
+        # Do we have the item or it's containing block?
+        if schema_item.id != self.id:
+            schema_item = schema_item.questionnaire.get_item_by_id(self.id)
+
+        # Get the user input
+        self.input = schema_item.get_user_input(user_input)
+
```
warrenbailey, line 20: I think the bit below should be refactored out into a valid_input method

app/questionnaire_state/answer.py
```diff
@@ -9,9 +9,33 @@ def __init__(self, id):
...
+            self.is_valid = False
+
+        # Try and get the typed value
+        if self.input:
+            try:
+                self.value = schema_item.get_typed_value(user_input)
+                self.is_valid = True
+                self.errors = None
+                self.warnings = None
+            except Exception as e:
```
warrenbailey: I'm not keen on the use of Exception here, we probably want to define our own exceptions for validation failures

app/questionnaire_state/user_journey_manager.py
```diff
@@ -181,3 +190,19 @@ def get_answer(self, id):
...
             return None
+
+    def validate(self):
+        # get the current location in the questionnaire
+        current_location = self.get_current_location()
+        if self.is_valid_location(current_location):
+            try:
+                current_state = self.get_state(current_location)
+                schema_item = self._schema.get_item_by_id(current_state.item_id)
+
```
warrenbailey: I wonder if we need an is_valid_item method on the schema now, then we can avoid all these try / except pass true statements

app/schema/item.py
```diff
@@ -17,3 +15,48 @@ def construct_state(self):
...
+                child_schema = self.questionnaire.get_item_by_id(child_state.id)
+                child_valid = child_schema.validate(child_state)
+                if not child_valid:
+                    is_valid = False
+
+            return is_valid
+        else:
+            import pdb
+            pdb.set_trace()
+            raise Exception('Cannot validate - incorrect state class')
```
warrenbailey: As above, should raise a more specific exception

app/schema/item.py
```diff
@@ -17,3 +15,48 @@ def construct_state(self):
 
     def get_state_class(self):
         pass
+
+    def validate(self, state):
+        if isinstance(state, self.get_state_class()):
+            is_valid = True
+            for child_state in state.children:
+                child_schema = self.questionnaire.get_item_by_id(child_state.id)
+                child_valid = child_schema.validate(child_state)
```
warrenbailey, line 28: can this just be simplified to is_valid = child.schema.validate(child_state) ?

app/schema/item.py
```diff
@@ -17,3 +15,48 @@ def construct_state(self):
...
+    def augment_with_state(self, state):
+        if state.id == self.id:
+            self.is_valid = state.is_valid
+            self.errors = state.errors
+            self.warnings = state.warnings
+            for child_state in state.children:
+                child_schema = self.questionnaire.get_item_by_id(child_state.id)
+                if child_schema:
+                    child_schema.augment_with_state(child_state)
+                else:
```
warrenbailey: I think the state object should contain everything that is needed to render the page, and it would be better to expand that rather than the schema object.

Though after our discussion, that might be best left to the display widget branch. So maybe add a TODO here to remove this.

app/templating/template_pre_processor.py
```diff
@@ -145,28 +142,15 @@ def _augment_answer(self, answer):
...
-                            self._collect_errors(section, errors, warnings)
-                            for question in section.questions:
-                                self._collect_errors(question, errors, warnings)
-                                for answer in question.answers:
-                                    self._collect_errors(answer, errors, warnings)
-
-        self._schema.errors = errors
-        self._schema.warnings = warnings
+        # This now feels wrong, but I can't think of a better way of doing this without ripping apart the whole rendering pipeline
+        schema_item = self._schema.get_item_by_id(current_state.item_id)
```
warrenbailey: Maybe add a TODO here, as we're probably going to have to do this work next

app/templates/partials/answers/date.html
```diff
@@ -12,7 +12,7 @@
     {% with name="day" %}
     <div class="field field--{{name}}">
       <label class="label label--small" for="{{answer.id}}-{{name}}">{{_('Day')}}</label>
-      {{form.select(name, (range(1, 32)), answer_value[0]|int, answer)}}
+      {{form.select(name, (range(1, 32)), answer_value[2]|int, answer)}}
```
warrenbailey: Why has this has changed from answer_value[0] to answer_value[2]?

app/questionnaire_state/user_journey_manager.py
warrenbailey: Actually we can use the item_exists method that's already there

app/schema/answers/date_answer.py
```diff
@@ -1,6 +1,30 @@
...
+        month = post_data.get(self.id + '-month', '')
+        year = post_data.get(self.id + '-year', '')
+
+        user_input = year + '/' + month + '/' + day
+
+        for checker in self.type_checkers:
+            result = checker.validate(user_input)
+            if not result.is_valid:
+                raise Exception(result.errors[0])
+                return None
```
warrenbailey: This return statement isn't needed

app/questionnaire_state/answer.py
weapdiv-david, line 20: What would the valid_input do?  if the response is an empty string or whatnot, then it already gets set to none.  if the input is none, there's no point doing the type-checking, and the mandatory check is already handled elsewhere.

I'm not sure how to break this down further, although there maybe opportunities to do so once this is handled by the display widgets

app/schema/item.py
weapdiv-david, line 28: possibly, but at present the state object doesn't hold a reference to it's associated schema counterpart.  This may change when we look at the augmentation strategy for rendering

warrenbailey:
I noticed you've changed the schema to bring the validation messages to the top. Does that mean its no longer backwards capability? 

If so, schemas created by the author application won't work with this implementation and we'd also need to update the JSON schema in the author application to match. I'm just wondering if its worth leaving as is for now?

app/questionnaire_state/answer.py
warrenbailey, line 20: I was just suggesting extracting out the code below into a sub-method as it is specifically related to type checking, something like _validate_input() or _extract_type() would be viable method names. It just keeps the size of the method down, limits complexity and makes the code more readable.

app/schema/item.py
warrenbailey, line 28: I'm not sure I understand, as far as I can work out child_valid will always be true or false, so this if statement seems unnecessary?

warrenbailey:
Couple of minor comments, but looking good. 

It might be worth squashing some commits though, it's quite a big change now?

app/validation/messages.py
```diff
@@ -0,0 +1,13 @@
+from flask.ext.babel import gettext as _
+
+# Set up default error and warning messages
```
dhilton: Why messages not errors? These are all error states plus soft checks are still errors (of sorts). When I saw this I instantly thought it was to do with translation.

tests/app/data_manually_created/0_star_wars.json
```diff
@@ -1,424 +1,514 @@
...
+                                                    "value": "Cow"
+                                                },
+                                                {
+                                                    "label": "Hippo",
+                                                    "value": "Hippo"
+                                                }
+                                            ],
+                                            "q_code": "4",
+                                            "type": "Radio",
+                                            "validation": {
```
dhilton: Is validation a required value here? I.e. even if there are no validation messages is this still required?

tests/app/templating/test_template_pre_processor.py
```diff
@@ -82,66 +73,66 @@ def test_augment_answer(self):
...
-        pre_proc._augment_questionnaire()
-
-        # check the questionnaire has been augmented correctly
-        self.assertIsInstance(self.schema.errors, OrderedDict, 'Errors should be an OrderedDict')
-        self.assertIsInstance(self.schema.warnings, OrderedDict, 'Warnings should be an OrderedDict')
-
-        self.assertListEqual(self.schema.errors['answer-2'], ['There is an error'])
-        self.assertListEqual(self.schema.errors['answer-3'], ['This is a required field'])
-        self.assertListEqual(self.schema.warnings['answer-2'], ['There is a warning'])
+    # def test_collect_errors(self):
```
dhilton: Why is this commented out? Can we remove it completely?

weapdiv-david: Test was commented out as the method it tested no longer exists, however the functionality under test is still required but is performed elsewhere.  I commented out the test and plan to refactor it to test the functionality required

app/schema/item.py
weapdiv-david, line 28: Wasn't sure I understood where you were coming from... The reason I've done it this way is to allow `is_valid` to represent the state of the current item, while the method itself interrogates all children.  If any of the children are in valid, then the current_item is invalid, but we presume it to be valid until a child fails.

We are not setting the is_valid attribute on the children, only summising whether the tree is valid.  I suppose a better method name might have been is_valid(), but that's already the name of the property.

weapdiv-david:
With regard to the validation messages at the top, this is a backwards compatible change.  The validation attribute at answer level remains optional and the new messages attribute at questionnaire level is also optional.

There will be a small change required to the author schema validator to allow these attributes to be added or missed as neccessary, but none of the existing officially supported schemas should be effected.

tests/app/templating/test_template_pre_processor.py
weapdiv-david: on reflection, I've removed one of these tests as this wasn't the correct place to do it.  The other test has been simplified and reinstated, but it now only checks that the global errors object is an OrderedDict

tests/app/data_manually_created/0_star_wars.json
weapdiv-david: Removed in the latest commit

app/validation/messages.py
weapdiv-david: I've renamed this module to error_messages.  It provides a central place for error messages that can easily be localised

warrenbailey:
We've discussed the outstanding issues, happy to merge as is.

warrenbailey:
LGTM 👍

PR #341, opened 2016-07-28
Display widgets (Based on work in 410)

weapdiv-david:

This PR adds display widgets on top of the previous refactoring.  It also fixes the handling of list type values (checkboxes).
### Widgets Currently Supported

[x] CurrencyWidget
[x] TextWidget, used by Text, Integer and PositiveInteger answers
[x] TextareaWidget
[x] DateWidget
### How to review

1) Check out the branch
2) Run the unit tests
3) Run the app

### Who worked on the PR

@weapdiv-david worked on this

### DIFF
--- a/app/schema/answers/checkbox_answer.py
+++ b/app/schema/answers/checkbox_answer.py
@@ -1,6 +1,31 @@
 from app.schema.answer import Answer
+from app.schema.widgets.checkbox_group_widget import CheckboxGroupWidget
+from app.questionnaire_state.exceptions import StateException
 
 
 class CheckboxAnswer(Answer):
-    def __init__(self):
-        super().__init__()
+    def __init__(self, answer_id=None):
+        super().__init__(answer_id)
+        self.widget = CheckboxGroupWidget(self.id)
+
+    def get_typed_value(self, post_vars):
+        # We cannot tyoe cast a list of values, so just return the user_input
+        return self.get_user_input(post_vars)
+
+    def get_user_input(self, post_vars):
+        return self.widget.get_user_input(post_vars)
+
+    def validate(self, state):
+        if isinstance(state, self.get_state_class()):
+
+            # Mandatory check
+            if self.mandatory and len(state.input) == 0:
+                state.errors = []
+                state.errors.append(self.questionnaire.get_error_message('MANDATORY', self.id))
+                state.is_valid = False
+                return False
+
+            # Here we just report on whether the answer has passed type checking
+            return state.is_valid
+        else:
+            raise StateException('Cannot validate - incorrect state class')
--- a/app/schema/answers/checkbox_answer.py
+++ b/app/schema/answers/checkbox_answer.py
@@ -23,7 +23,6 @@ class CheckboxAnswer(Answer):
                 state.errors = []
                 state.errors.append(self.questionnaire.get_error_message('MANDATORY', self.id))
                 state.is_valid = False
-                return False
 
             # Here we just report on whether the answer has passed type checking
             return state.is_valid
--- a/app/schema/answers/integer_answer.py
+++ b/app/schema/answers/integer_answer.py
@@ -1,5 +1,4 @@
 from app.schema.answer import Answer
-from app.validation.positive_integer_type_check import PositiveIntegerTypeCheck
 from app.validation.integer_type_check import IntegerTypeCheck
 from app.schema.widgets.text_widget import TextWidget
 
@@ -8,16 +7,6 @@ class IntegerAnswer(Answer):
     def __init__(self, answer_id=None):
         super().__init__(answer_id)
         self.type_checkers.append(IntegerTypeCheck())
-        self.widget = TextWidget(self.id)
-
-    def _cast_user_input(self, user_input):
-        return int(user_input)
-
-
-class PositiveIntegerAnswer(Answer):
-    def __init__(self, answer_id=None):
-        super().__init__(answer_id)
-        self.type_checkers.append(PositiveIntegerTypeCheck())
         self.widget = TextWidget(self.id)
 
     def _cast_user_input(self, user_input):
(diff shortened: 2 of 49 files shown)


app/schema/widgets/currency_widget.py
```diff
@@ -0,0 +1,30 @@
...
+logger = logging.getLogger(__name__)
+
+
+class CurrencyWidget(Widget):
+    def __init__(self, name):
+        super().__init__(name)
+
+    def render(self, schema, state):
+        widget_params = {
+            'answer': {
```
warrenbailey: This answer object seems to repeat a lot. Is it possible to move it to the super class?

weapdiv-david: Am waiting for Hamish to commit his changes to the DateWidget, and for me to finish the checkbox and radio group widgets.  Once those are in it would make sense to try refactor the interface between python and html.

weapdiv-david: Following on from this, the internal structures of the widget is subject to being refactored somewhat once we have some additional widget types to play with.  The next step would be to add an "Other" option to the radio group and checkbox group widgets allowing user-specified options.

Hamish is planning to continue tidying up the template layer which will also drive refactoring opportunities within widgets.

warrenbailey:
Happy with this now

weapdiv-david:
One comment... This does change the output format used for the Census 2016 slightly...

```
...
"data": {
        "device": ["1", "2", "3", "5", ""],
        "help2": ["1", "3", ""],
        "whypaper": ["2", "5", ""]
       ...
}
...
```

becomes

```
...
"data": {
        "device": ["1", "2", "3", "5"],
        "help2": ["1", "3"],
        "whypaper": ["2", "5"]
       ...
}
...
```

The blank option "" is no longer needed and so is excluded from the list that is sent downstream.

This is also true of any other survey that uses checkboxes and radios, although there are currently none in production apart from the census test.

warrenbailey:
I think the blank option was causing SDX a problem anyway, so I would imagine they'll be happy with that. Have a chat with @ajmaddaford

weapdiv-david:
Just had a chat with him.  He said they'd noticed it, but hadn't really looked into it properly yet.  He was happy that it's no longer going to be there though

app/schema/answers/checkbox_answer.py
```diff
@@ -1,6 +1,31 @@
...
+        return self.widget.get_user_input(post_vars)
+
+    def validate(self, state):
+        if isinstance(state, self.get_state_class()):
+
+            # Mandatory check
+            if self.mandatory and len(state.input) == 0:
+                state.errors = []
+                state.errors.append(self.questionnaire.get_error_message('MANDATORY', self.id))
+                state.is_valid = False
```
dhilton, line 25: Why not return state.is_valid here rather than False? Seems a small duplication we can get rid of.

app/schema/answers/integer_answer.py
```diff
@@ -1,21 +1,24 @@
...
+    def __init__(self, answer_id=None):
+        super().__init__(answer_id)
         self.type_checkers.append(IntegerTypeCheck())
+        self.widget = TextWidget(self.id)
 
     def _cast_user_input(self, user_input):
         return int(user_input)
 
 
 class PositiveIntegerAnswer(Answer):
```
dhilton: Why does this not subclass from IntegerAnswer? The TypeCheck could then make use of mixins to check for positive number and IntegerChecks. Not that big a deal here but as we get to more complex types would reduce boilerplate.

app/schema/widget.py
```diff
@@ -0,0 +1,9 @@
+class Widget(object):
+    def __init__(self, name):
+        self.name = name
+
+    def render(self, schema, state):
+        pass
```
dhilton: Can we tighten up this interface contract so that we return an NotImplemented exception if (and only if) we expect all widgets to implement the `render` and `get_user_input` methods? For reference on the exception: https://dwieeb.com/2015/06/02/usages-of-notimplemented-and-notimplementederror-in-python/

app/schema/widgets/checkbox_group_widget.py
```diff
@@ -0,0 +1,48 @@
+from app.schema.widget import Widget
+from flask import render_template
+from app.libs.utils import ObjectFromDict
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+class CheckboxGroupWidget(Widget):
+    def __init__(self, name):
```
dhilton: This doesn't seem to differ from the base class method? Why is it restated here?

app/schema/widgets/currency_widget.py
```diff
@@ -0,0 +1,31 @@
+from app.schema.widget import Widget
+from flask import render_template
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+class CurrencyWidget(Widget):
+    def __init__(self, name):
```
dhilton: Again not sure why the **init** method is called again here?

app/schema/widgets/radio_group_widget.py
```diff
@@ -0,0 +1,47 @@
...
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+class RadioGroupWidget(Widget):
+    def __init__(self, name):
+        super().__init__(name)
+
+    def get_user_input(self, post_vars):
```
dhilton: Across a couple of the widgets this seems boilerplate - can we move it to the base class to make this DRY'er?

app/schema/answers/checkbox_answer.py
weapdiv-david, line 25: agreed and removed

dhilton:
The Choice Widgets seem to share a bit of boilerplate (RadioChoices and CheckboxChoices) around how they generate options - I think we need to be careful how we ensure that the widgets are reusable and follow a sensible object oriented pattern.

weapdiv-david:
Agreed.  Both of these are candidates for refactoring when we get stories that describe how they should work.  Currently, they've just been kind of hacked in to support census and are not used by any of our supported surveys.

dhilton:
LGTM 👍 I'll add more stories for the composite questionTypes and associated widgets.

PR #342, opened 2016-07-29
EQ 427, basic routing and integration tests

LJBabbage:

Basic routing for survey runner
(N.B The summary page also has a minor visual issue, produces single lines for missing blocks, this won't affect any other survey bar star wars)
### How to review

Use dev to open star wars, play around with the routing on the first few question, make sure it all works as expected
### Who worked on the PR

@LJBabbage

### DIFF
--- a/app/data/0_star_wars.json
+++ b/app/data/0_star_wars.json
@@ -25,6 +25,530 @@
                     "display": {
                         "properties": {}
                     },
+                    "id": "f22b1ba4-d15f-48b8-a1f3-db62b6f34cc0",
+                    "sections": [
+                        {
+                            "description": "",
+                            "display": {
+                                "properties": {}
+                            },
+                            "id": "ed3e200a-0735-4e8d-9eea-627c1d908697",
+                            "questions": [
+                                {
+                                    "answers": [
+                                        {
+                                            "display": {
+                                                "properties": {
+                                                    "columns": true
+                                                }
+                                            },
+                                            "guidance": "",
+                                            "id": "ca3ce3a3-ae44-4e30-8f85-5b6a7a2fb23c",
+                                            "label": "Choose a side",
+                                            "mandatory": true,
+                                            "options": [
+                                                {
+                                                    "label": "Light Side",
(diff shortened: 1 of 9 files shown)


app/data/0_star_wars.json
```diff
@@ -20,6 +20,530 @@
...
+                            ],
+                            "title": "Choose your side",
+                            "validation": []
+                        }
+                    ],
+                    "routing_rules":[
+                        {
+                            "goto": {
+                                "id" : "96682325-47ab-41e4-a56e-8315a19ffe2a",
+                                "when": {
```
collisdigital, line 88: Just a thought, would this be better of being called "if" rather than "when"?

app/routing/routing_engine.py
```diff
@@ -19,22 +26,62 @@ def get_next_location(self, current_location):
...
+                    if index + 1 < len(self._schema.groups):
+                        # return the first block in the next group
+                        return self._schema.groups[index + 1].blocks[0].id
+        # There are no more blocks or groups, go to summary
+        return 'summary'
+
+    @staticmethod
+    def process_routing_rules(routing_rules):
+
+        logger.debug("Processing routing rules", routing_rules)
```
collisdigital: Parameter isn't logged (doesn't Flake8 identify this?)

dhilton: No - Although perhaps we should look again at our Flake8 usage, this is a good example of how to configure it: https://bugzilla.mozilla.org/show_bug.cgi?id=1237977

app/data/0_star_wars.json
LJBabbage, line 88: Can do, I used 'when' because it made the code look a bit nicer
fugly alert:
 if 'if' not in rule['goto'].keys():

app/routing/routing_engine.py
collisdigital: Looks like a winner; add a new card?

dhilton:
Have tested - great work 👍

PR #348, opened 2016-08-03
Replacing User Journey Manager with Questionnaire Manager

warrenbailey:

In our progression towards a simpler code base we're refactoring the user journey manager and combining it with the questionnaire manager. 

I've also introduced an optimisation that means the schema is only parsed once per session and then pickled to the database alongside the user's state.
### How to review
1. Check out this branch
2. Confirm the test pass
3. Check the star wars and MCI surveys and make sure nothing it broken
### Who worked on the PR

@warrenbailey


app/main/views/root.py
```diff
@@ -48,14 +47,7 @@ def login():
...
 
-        # load the schema
-        schema = load_and_parse_schema(eq_id, form_type)
-        if not schema:
-            return errors.page_not_found()
-
-        user_journey_manager = UserJourneyManager.get_instance()
-        if not user_journey_manager:
-            user_journey_manager = UserJourneyManager.new_instance(schema)
+        user_journey_manager = create_questionnaire_manager()
```
LJBabbage: Thought we were trying to remove the term user_joruney_manager

warrenbailey: will fix

app/questionnaire/questionnaire_manager.py
```diff
@@ -1,38 +1,242 @@
...
+        self.submitted_at = None
         self._schema = schema
-        self._user_journey_manager = user_journey_manager
-        self._metadata = metadata
-        self._routing_engine = routing_engine
-        self._user_action_processor = UserActionProcessor(self._schema, self._metadata, self._user_journey_manager)
-        self._pre_processor = TemplatePreProcessor(self._schema, self._user_journey_manager, self._metadata)
+        self._current = None  # the latest page
+        self._first = None  # the first page in the doubly linked list
+        self._archive = {}  # a dict of completed or discarded pages
```
LJBabbage: Completed or discarded pages? not sure the comment makes sense, is this right?

app/questionnaire/questionnaire_manager.py
```diff
@@ -1,38 +1,242 @@
...
+        self._valid_locations = self._build_valid_locations()
+        self._pre_processor = TemplatePreProcessor(self._schema, self)
+
+    @staticmethod
+    def new_instance(schema):
```
LJBabbage, line 40: Don't like the name here as its also saving the state to database, save_new_questionnaire_manager, sounds a bit better. Also another appearance from user_journey_manager

app/questionnaire/questionnaire_manager.py
```diff
@@ -1,38 +1,242 @@
...
+        user_journey_manager = QuestionnaireManager(schema)
+        # immediately save it to the database
+        StateManager.save_state(user_journey_manager)
+        logger.debug("Constructing new state")
+        return user_journey_manager
+
+    @staticmethod
+    def get_instance():
```
LJBabbage: Don't like the get_instance name, too generic

LJBabbage:
Good work, all works as expected. A few minor adjustments

app/questionnaire/questionnaire_manager.py
warrenbailey: will change

app/questionnaire/questionnaire_manager.py
warrenbailey, line 40: new_instance and get_instance is a standard singleton pattern though. It's just using the DB to maintain the instance rather than a variable. I'm reluctant to change this one to be honest.

I'll fix the variable name though

PR #354, opened 2016-08-04
Improve debugging for routing rules

### LINKED ISSUE #353: Routing, improve debugging
### Expected behaviour

The exception thrown on a missing route could be better
### Actual behaviour

Everything works fine, the exception thrown just isn't quite right

Create a schema with a missing routing rule

LJBabbage:

fixes #353 
### How to review
1. Make sure the star wars routing survey works
2. Add a new option in the first question without a routing rules and make sure the error is valid
### Who worked on the PR

@LJBabbage


warrenbailey:
LGTM 👍

PR #356, opened 2016-08-04
Add test for types sent downstream

weapdiv-david:

This PR adds a test of the data types being sent downstream to SDX.  While these have not changed, this PR adds a test that will serve as a warning flag if they do.
### How to review

1) Checkout the branch
2) Run the tests
3) Verify that they all pass
### Who worked on the PR

@weapdiv-david worked on this

### DIFF
--- a/tests/integration/downstream/downstream_test_case.py
+++ b/tests/integration/downstream/downstream_test_case.py
@@ -1,0 +1,40 @@
+from tests.integration.integration_test_case import IntegrationTestCase
+
+from app.submitter.submitter import SubmitterFactory, Submitter
+
+
+class MyMockSubmitter(Submitter):
+    def __init__(self):
+        self._message = None
+        self._submitted_at = None
+
+    def send(self, message):
+        self._message = message
+        return True
+
+    def encrypt_message(self, message):
+        return message
+
+    def get_message(self):
+        return self._message
+
+
+class DownstreamTestCase(IntegrationTestCase):
+    '''
+    Overrides the application SubmitterFactory to provide our own which
+    allows the message to be captured and interrogated.
+    '''
+    _submitter = MyMockSubmitter()
+
+    @staticmethod
+    def get_submitter():
+        return DownstreamTestCase._submitter
+
+    def setUp(self):
+        super().setUp()
+        self._old_method = SubmitterFactory.get_submitter
+        SubmitterFactory.get_submitter = DownstreamTestCase.get_submitter
+
+    def tearDown(self):
+        DownstreamTestCase._submitter._message = None
+        SubmitterFactory.get_submitter = self._old_method
--- a/app/submitter/submitter.py
+++ b/app/submitter/submitter.py
@@ -35,8 +35,7 @@ class Submitter(object):
         '''
         message, submitted_at = Converter.prepare_answers(user, metadata, schema, answers)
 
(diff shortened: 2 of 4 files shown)


tests/integration/downstream/downstream_test_case.py
```diff
@@ -0,0 +1,38 @@
+from tests.integration.integration_test_case import IntegrationTestCase
+
+from app.submitter.submitter import SubmitterFactory, Submitter
+from app.submitter.converter import Converter
+
+
+class MyMockSubmitter(Submitter):
+    def __init__(self):
+        self._message = None
+        self._submitted_at = None
```
warrenbailey, line 9: Is there a reuse for self._submitted_at? It doesn't appear to be used anywhere, apart from in the teardown method where it's just set back to None?

tests/integration/downstream/downstream_test_case.py
```diff
@@ -0,0 +1,38 @@
...
+
+    def send_answers(self, user, metadata, schema, answers):
+        message, submitted_at = Converter.prepare_answers(user, metadata, schema, answers)
```
warrenbailey: I still think it would of been better to refactor the submitter, extract the encrypting functionality into a separate method and then override that and the send message method instead. This just means that if the send_answers method or the Converter.prepare_answer method signature changes we've got to remember to change it here too.

tests/integration/downstream/downstream_test_case.py
```diff
@@ -0,0 +1,38 @@
...
+
+class DownstreamTestCase(IntegrationTestCase):
+    '''
+    Overrides the application SubmitterFactory to provide our own which
+    allows the message to be captured and interrogated.
+    '''
+    _submitter = MyMockSubmitter()
+
+    @staticmethod
+    def create_submitter():
```
warrenbailey: This method implies it creates a new submitter when actually it just returns the single instance, this would be better name get_submitter()?

tests/integration/downstream/downstream_test_case.py
```diff
@@ -0,0 +1,38 @@
...
+        return DownstreamTestCase._submitter
+
+    def setUp(self):
+        super().setUp()
+        self._old_method = SubmitterFactory.get_submitter
+        SubmitterFactory.get_submitter = DownstreamTestCase.create_submitter
```
warrenbailey: Which would actually make the above line read better to.

warrenbailey:
Good work, we needed this as it was very much lacking. A few minor comments related to code style but then good to go.

tests/integration/downstream/downstream_test_case.py
weapdiv-david, line 9: I plan to add tests for the submitted date as it's included in the message

tests/integration/downstream/downstream_test_case.py
weapdiv-david: Agreed, I'll make that change now

tests/integration/downstream/downstream_test_case.py
weapdiv-david, line 9: Actually, I've removed it as part of the rebase to the encrypt message...  The hooks no longer allow us to get the submission date

warrenbailey:
LGTM 👍

Issue #263, opened 2016-06-20
Preceeding zeros are not handled properly


shilpabasu:
### Expected behaviour

When a user enters preceding zeros, they should be removed
### Actual behaviour

Preceding zeros are not removed and passed on to the downstream system
### Steps to reproduce the behaviour -
1. Open an eQ survey
2. Enter preeceding zeros (0000000000000000000000001) as test data in various fields.
3. Submit Survey
4. Verify the image file and other files generated for down stream system.
#### Browser

IE11
#### Operating System

Windows 7
### Screenshot

![form_page1](https://cloud.githubusercontent.com/assets/17923832/16190505/7c063d68-36d7-11e6-80d3-6ab5ff26d810.PNG)
![form_page2](https://cloud.githubusercontent.com/assets/17923832/16190689/73be0f40-36d8-11e6-9228-5698511b4526.PNG)
[Output_DataFile.txt](https://github.com/ONSdigital/eq-survey-runner/files/323163/Output_DataFile.txt)
![output_imagefile](https://cloud.githubusercontent.com/assets/17923832/16190690/73c2a2ee-36d8-11e6-87c5-8ef55a37fdc5.jpg)
![summary1](https://cloud.githubusercontent.com/assets/17923832/16190688/73bbecd8-36d8-11e6-90b3-cfd08cbfdcc8.PNG)
![summary2](https://cloud.githubusercontent.com/assets/17923832/16190691/73c70e42-36d8-11e6-97e4-1ddb8daf8ad5.PNG)

dhilton:
Has this been verified to be fixed by #341 ? @weapdiv-david

weapdiv-david:
No, we're still sending strings downstream at present.  Will need to verify with the SDX team that we can change the format to send integers, then update our downstream tests accordingly

collisdigital:
Confirmed this is fixed by #368 on pre-prod. Closing.

PR #368, opened 2016-08-11
Fix issue with proceeding zeros

weapdiv-david:

Fixes issue #263.  Updates the converter, questionnaire manager and widgets to use typed values rather than input values.  Modifies the downstream data test to check the proceeding zeros case.
### How to review

1) Checkout the branch
2) Run the tests
3) Go to the dev console and select a survey that allows you to enter integer/currency values
4) Fill an input box with a load of proceeeding zeros and a valid integer
5) Cause a validation error elsewhere on the page and submit, causing the page to be redisplayed
6) Verify that the proceeding zeros for your valid integer answer have been removed
7) Complete the survey and verify the proceeding zeros are not present
### Who worked on the PR

@weapdiv-david worked on this

### DIFF
--- a/app/submitter/converter.py
+++ b/app/submitter/converter.py
@@ -47,11 +47,6 @@ class SubmitterConstants(object):
 
 class Converter(object):
 
-    """
-    Date Format expected by SDX
-    """
-    SDX_DATE_FORMAT = "%d/%m/%Y"
-
     @staticmethod
     def prepare_answers(user, metadata_store, questionnaire, answers):
         """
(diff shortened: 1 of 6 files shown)


app/submitter/converter.py
```diff
@@ -47,6 +47,11 @@ class SubmitterConstants(object):
 
 class Converter(object):
 
+    """
+    Date Format expected by SDX
+    """
+    SDX_DATE_FORMAT = "%d/%m/%Y"
```
warrenbailey: I think this should be configurable, can we move it to settings.py and expose as an environment variable?

warrenbailey:
Looks good, lets just make the format configurable then it's good to go

PR #371, opened 2016-08-12
eq 471 repeating elements

LJBabbage:

Repeating blocks, registry for rules and updated tests
### How to review
1. Open Star wars survey
2. Navigate to the dark side, there are 2 repeating examples, one behind the no ship question (crawler example) and one behind yes ships. One is an example of routing after the repeating question the other is no routing after.
3. Check both work
4. Check you can go back and forward as expected and change your answers, including the amount of repeating.
5. Navigate to the submission page and make sure everything looks right.
### Who worked on the PR

@LJBabbage

### DIFF
--- a/app/questionnaire/questionnaire_manager.py
+++ b/app/questionnaire/questionnaire_manager.py
@@ -23,10 +23,8 @@ class QuestionnaireManager(object):
     This class represents a user journey through a survey. It models the request/response process of the web application
     using a doubly linked list. Each node in the list is essentially a page displayed to the user. A new node is created
     by a GET request and subsequently updated via a POST request.
-
     The doubly linked list approach allows us to maintain the path the user has taken through the question. If that path
     changes we archive off the nodes in case the user revisits that path.
-
     '''
     def __init__(self, schema):
         self.submitted_at = None
@@ -358,3 +356,17 @@ class QuestionnaireManager(object):
         current_user.delete_questionnaire_data()
         # and clear out the session state
         session_manager.clear()
+
+    def register_element_in_schema(self, item):
+        self._schema.register(item)
+
+    def check_item_exists_in_schema(self, item_id):
+        return self._schema.item_exists(item_id)
+
+    def get_schema_item_by_id(self, item_id):
+        return self._schema.get_item_by_id(item_id)
+
+    def add_repeating_element(self, item):
+        self._schema.register(item)
+        self._valid_locations.append(item.id)
+        self._create_new_state(item.id)
--- a/app/repeating_element/repeating_element.py
+++ b/app/repeating_element/repeating_element.py
@@ -1,0 +1,69 @@
+import logging
+
+from copy import deepcopy
+
+from app.schema.answer import Answer
+
+
+logger = logging.getLogger(__name__)
+
+
+class RepeatingElementException(Exception):
+    pass
+
+
+class RepeatingElement(object):
+    ''' The repeating element manager will create a new element with new ids and q-codes using the
+        repetition number as an identifier (e.g. item.id_repetition number)
+    '''
+
+    def __init__(self, questionnaire_manager):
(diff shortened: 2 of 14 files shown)


app/questionnaire/questionnaire_manager.py
```diff
@@ -126,7 +125,7 @@ def update_state(self, item_id, user_input):
         else:
             raise TypeError("Can only handle blocks")
 
-    def _append(self, node):
+    def append(self, node):
```
warrenbailey: bit confused to why you've renamed this method?

app/questionnaire/questionnaire_manager.py
```diff
@@ -22,18 +22,16 @@ class QuestionnaireManager(object):
...
         self.submitted_at = None
-        self._schema = schema
-        self._current = None  # the latest node
-        self._first = None  # the first node in the doubly linked list
-        self._archive = {}  # a dict of discarded nodes for later use (if needed)
-        self._valid_locations = self._build_valid_locations()
+        self.schema = schema
+        self._current = None  # the latest page
+        self._first = None  # the first page in the doubly linked list
+        self.archive = {}  # a dict of discarded pages for later use (if needed)
```
warrenbailey: Not sure we should be renaming these variables, as this is essentially exposing the internals of the class

app/repeating_element/repeating_element.py
```diff
@@ -0,0 +1,72 @@
...
+        next_repetition = current_element.repetition + 1
+        new_element_id = self._rename_id(current_element.id, next_repetition)
+
+        # Check to see if we can create the element
+        if not self._questionnaire_manager.schema.item_exists(new_element_id):
+            new_element = self._clone_element(current_element, new_element_id, next_repetition)
+            self._update_state(new_element)
+        elif new_element_id in self._questionnaire_manager.archive:
+            # re-attach it, if it is in the archive
+            self._questionnaire_manager.append(self._questionnaire_manager.archive[new_element_id])
```
warrenbailey: We shouldn't be modifying the questionnaire manager internals outside the class. This would be better moved into the questionnaire manager itself and then exposed as a function.

app/repeating_element/repeating_element.py
```diff
@@ -0,0 +1,72 @@
...
+    @staticmethod
+    def _rename_id(current_id, repetition):
+        logging.debug("Updating id for %s, repetition %s", current_id, repetition)
+        # Take of any previous repetition numbers
+        return current_id.rsplit('_', 1)[0] + "_" + str(repetition)
+
+    def _update_state(self, new_element):
+        logging.debug("Registering new element %s in schema and updating state", new_element.id)
+        self._questionnaire_manager.schema.register(new_element)
+        self._questionnaire_manager.valid_locations.append(new_element.id)
```
warrenbailey: Again, I think this needs an add_valid_location method on the questionnaire manager

app/repeating_element/repeating_element.py
```diff
@@ -0,0 +1,72 @@
...
+class RepeatingElementException(Exception):
+    pass
+
+
+class RepeatingElement(object):
+    ''' The repeating element manager will create a new element with new ids and q-codes using the
+        repetition number as an identifier (e.g. item.id_repetition number)
+    '''
+
+    def __init__(self, questionnaire_manager):
```
warrenbailey, line 20: I think this class should be constructed with the schema, rather than using the one inside questionnaire manager

warrenbailey:
LGTM 👍

app/repeating_element/repeating_element.py
warrenbailey, line 20: Not needed anymore

warrenbailey:
Sorry Lewis, it's now failing the new linting rules 👎

warrenbailey:
Looks good 👍

PR #376, opened 2016-08-15
Issue 366 post data length

jjeane:

Post data payload limited to 64k
### How to review

Run eq-survey-runner
Check you can't submit more than 64k
### Who worked on the PR

@jjeane & @warrenbailey

### DIFF
--- a/app/main/views/questionnaire.py
+++ b/app/main/views/questionnaire.py
@@ -77,6 +77,9 @@ def do_get(questionnaire_manager, location):
 
 def do_post(collection_id, eq_id, location, questionnaire_manager):
     logger.debug("POST request question - current location %s", location)
+
+    logger.debug("POST request length %s", request.content_length)
+
     questionnaire_manager.process_incoming_answers(location, request.form)
     next_location = questionnaire_manager.get_current_location()
     metadata = MetaDataStore.get_instance(current_user)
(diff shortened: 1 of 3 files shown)


app/schema/widgets/textarea_widget.py
```diff
@@ -16,7 +16,7 @@ def render(self, state):
                 'label': state.schema_item.label,
                 'value': state.input or '',
                 'placeholder': '',
-                'maxChars': 2000,
+                'maxChars': state.schema_item.max_chars or 2000,
```
LJBabbage: We seem to still be using magic numbers, we need to get away from the 2000 and stick it as a default higher up

warrenbailey:
There is already a section in the schema called max_length we can use. However for some reason it's not parser by  the schema parser. We should just correct schema parser and use that value instead:

Have a look at 01_0203.json:

```
"display": {
  "properties": {
    "max_length": "2000"
   }
},
```

app/main/views/questionnaire.py
```diff
@@ -77,6 +77,9 @@ def do_get(questionnaire_manager, location):
 
 def do_post(collection_id, eq_id, location, questionnaire_manager):
     logger.debug("POST request question - current location %s", location)
+
+    logger.error("POST request length %s", request.content_length)
```
warrenbailey: maybe change to debug oops

LJBabbage:
Think we need to redirected correctly on error.

PR #381, opened 2016-08-16
Fix for Database blips

### LINKED ISSUE #337: Exception when Database blips
### Expected behaviour

When the database disconnects/reconnects operations resume normally
### Actual behaviour

There is an exception seen

This occurred when changing the instance size of the RDS database; may occurs in other scenarios too, such as switching from master to slave, needs checking.
### Technical information

See attached stack trace:
[database_exception.txt](https://github.com/ONSdigital/eq-survey-runner/files/385769/database_exception.txt)

warrenbailey:

Fixes #337 

Test database connections when they are requested from the pool, if they are invalid try them again as this should reinitialise the connection. If that still doesn't work then forcibly close them.
### How to review
1. Deploy this branch to elastic beanstalk
2. Restart the RDS instance
3. Check you can still complete a survey
### Who worked on the PR

@warrenbailey

…

### DIFF
--- a/app/data_model/database.py
+++ b/app/data_model/database.py
@@ -8,6 +8,9 @@ from sqlalchemy import Column
 from sqlalchemy import DateTime
 from sqlalchemy import String
 from sqlalchemy import create_engine
+from sqlalchemy import exc
+from sqlalchemy import event
+from sqlalchemy import select
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.orm import scoped_session
 from sqlalchemy.orm import sessionmaker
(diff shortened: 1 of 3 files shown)


app/data_model/database.py
```diff
@@ -68,8 +71,32 @@ def create_session_and_engine():
...
 
 base.metadata.create_all(engine)
+
+
+@event.listens_for(engine, "engine_connect")
+def test_connection(connection, branch):
+    if branch:
+        # "branch" refers to a sub-connection of a connection,
+        # we don't want to bother testing on these.
+        return
```
dhilton, line 85: What does this return? If the function has finished perhaps return None or True? Further on the code returns errors, so be good to have a return value under any scenario.

warrenbailey, line 85: It's a listener function, it doesn't return anything. The idea being it allows you to execute some functionality when a connection is checked out of the pool. In this case we test the connection, if it fails tries to reconnect and if the reconnect fails it then forcibly closes the connection.

It's a modified version of:
http://docs.sqlalchemy.org/en/latest/core/pooling.html#disconnect-handling-pessimistic

app/main/views/root.py
```diff
@@ -68,5 +68,5 @@ def login():
     except InvalidTokenException as e:
         logger.warning("Invalid Token provided")
         return errors.forbidden(e)
-    except RuntimeError as e:
+    except (Exception, RuntimeError) as e:
```
warrenbailey, line 71: Different file ;-)

app/data_model/database.py
dhilton: Ah ok cool - 👍 from me then.

warrenbailey:
Rebasing

iwootten:
So, having spent the best part to the afternoon getting this up, this appears to fail for me when I restart RDS. 

I can get go through the motions of a survey initially, and successfully restart RDS. When I attempt another questionnaire, can get to the introduction, but on proceeding, get an internal server error.

warrenbailey:
@iwootten I've tested this again this morning and it works for me. I do get one internal server error while the database is restarting but once it's up its all back to normal.

iwootten:
Yep works for me today too. Guess I was too eager after restarting the db yesterday.

Issue #383, opened 2016-08-16
Values causing validation errors aren't stored/displayed to user


collisdigital:
### Expected behaviour
- Changing a value from valid to something that causes a validation is stored and re-presented to the user on page refresh.
### Actual behaviour
- Changing a value from valid to something that cause a validation error is not stored and is not re-presented to the user on page refresh, instead they are presented the previous valid value entered prior to the validation error triggering.
### Steps to reproduce the behaviour

These steps involve two fields but the issue occurs with just one field but using two allows the problem to be shown without having to navigate around the questionnaire.
1. Launch a survey with currency fields
2. Enter a valid answer (whole number e.g. 1000) for one question on the page
3. Enter an invalid answer (e.g. 1000.00) for a different question on the page
4. Submit the page
5. Observe the single error report for the invalid answer
6. Swap the values in the two fields
7. Submit the page
8. Observe that both fields now have "1000" as the value but the field that "1000.00" was last entered into shows an error but the data is correct i.e. the "1000.00" has been lost
9. Change the field with the error to any other value e.g. "Jar Jar"
10. Submit the form
11. Observe that the value entered is lost i.e. "Jar Jar", but the error is displayed and the last valid value is shown i.e. "1000"
#### Browser

All
#### Operating System

All
### Screenshot

<img width="749" alt="screen shot 2016-08-16 at 20 46 46" src="https://cloud.githubusercontent.com/assets/14041600/17713596/a4783500-63f3-11e6-893c-25ac4ad22b14.png">

<img width="738" alt="screen shot 2016-08-16 at 20 48 19" src="https://cloud.githubusercontent.com/assets/14041600/17713603/a9d335c2-63f3-11e6-9d83-c119fe7ed158.png">

<img width="744" alt="screen shot 2016-08-16 at 20 48 31" src="https://cloud.githubusercontent.com/assets/14041600/17713605/aef9d6fa-63f3-11e6-89ec-1265b618cec5.png">

<img width="758" alt="screen shot 2016-08-16 at 20 49 08" src="https://cloud.githubusercontent.com/assets/14041600/17713613/b460f18c-63f3-11e6-8028-685da25a736e.png">

collisdigital:
Having spent some more time coming up with the steps to reproduce I think this is a duplicate of #374 which #377 was merged to fix, but the problem still occurs even with this fix.

collisdigital:
Confirmed this is fixed by #385 on pre-prod. Closing.

PR #385, opened 2016-08-17
Fix clearing values on update state

weapdiv-david:

Fixes issue #383, whereby if a valid value that had previously passed validation was replaced by an invalid one, the valid value was what was redisplayed to the end-user.
### How to review

1) Checkout the branch
2) Run the tests
3) Start the app and start a survey
4) Fill a valid value into a required field, but allow validation to fail across the page as a whole.
5) Replace the valid value with an invalid one and resubmit
6) Verify that your invalid value is redisplayed.
### Who worked on the PR

@weapdiv-david

### DIFF
--- a/app/questionnaire_state/answer.py
+++ b/app/questionnaire_state/answer.py
@@ -12,6 +12,7 @@ class Answer(Item):
 
     def update_state(self, user_input):
         # Get the user input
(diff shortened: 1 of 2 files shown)


warrenbailey:
LGTM 👍

PR #386, opened 2016-08-17
Use ISO 8601 for submitted_at

iwootten:

Dropped the use of DATETIME_FORMAT setting. This was only used for submitted_at within survey-runner and incorrectly specified UTC. Instead use datetimes isoformat method to set a ISO 8601 formatted version of the the submission time, as specified in the ons sdx-eq schema docs.
### How to review

Given this time is based on the time the test is run, unsure how we'd modify current integration tests to check this.
### Who worked on the PR

@iwootten

### DIFF
--- a/app/questionnaire/user_action_processor.py
+++ b/app/questionnaire/user_action_processor.py
@@ -97,7 +97,8 @@ class SubmitAnswers(UserAction):
         if is_valid:
             submitter = SubmitterFactory.get_submitter()
             submitted_at = submitter.send_answers(current_user, self._metadata, self._schema, answers)
-            logger.debug("setting submitted at %s", submitted_at)
-            self._questionnaire_manager.submitted_at = submitted_at.strftime(settings.DISPLAY_DATETIME_FORMAT)
+            local_submitted_at = submitted_at.astimezone(settings.EUROPE_LONDON)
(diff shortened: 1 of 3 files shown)


warrenbailey:
We discussed using utc_now to fix this defect:
'''
submitted_at = datetime.utcnow()
'''

This change updates the format of the date rather than the actual date, so it's now: 'submitted_at': '2016-08-18T08:30:09.164001+01:00. Is that correct?

iwootten:
Yes, there's no need to enforce the use of UTC for sdx. So long as it's correctly formatted ISO 8601 date, sdx doesn't care. The problem is at the moment that eq is formatting a BST date as UTC, because UTC is hardcoded in the format string. See https://github.com/ONSdigital/eq-survey-runner/blob/master/app/settings.py#L108. 

Given the problem was caused by an incorrect formatting string, why use it at all if there's a method for it already?

We can use utcnow() as well as use isoformat() to make it both UTC and formatted correctly, if eq needs that.

iwootten:
Discussed with @ajmaddaford and agreed to use utcnow() also.

warrenbailey:
Rebasing

warrenbailey:
LGTM 👍

PR #395, opened 2016-08-24
Code cov

jjeane:

Added codecov to survey runner
Travis can be set up to fail under certain code coverage %
### How to review

Ensure codecov is running on this branch
In travis.yml uncomment out

`- coverage report --fail-under=80` 

and change value to ensure that build fails/passes as per the criteria
### Who worked on the PR

@jjeane

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,6 @@ python:
 - '3.5'
 env:
 - EQ_SCHEMA_BUCKET=""
-before_install:
-  - pip install codecov
 install:
 - ". $HOME/.nvm/nvm.sh"
 - rm -rf node_modules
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -5,6 +5,8 @@ python:
 - '3.5'
 env:
 - EQ_SCHEMA_BUCKET=""
+before_install:
+  - pip install codecov
```
warrenbailey: This dependency could be added to the requirements.txt instead

.travis.yml
```diff
@@ -19,6 +21,8 @@ before_script:
 - "./scripts/run_app.sh"
 script:
 - "./scripts/run_tests.sh"
+# fail travis if under X% test coverage
+# - coverage report --fail-under=80
```
warrenbailey: Might as well enable this, but set it to the around current code coverage level, so 77%?

At least that makes sure people don't make it worse, we can always up it to 80% when we get there.

warrenbailey:
A few minor changes and then this is good to go. Also needs rebasing to master.

I like this code coverage tool :)

dhilton:
@warrenbailey Are you ok to make those changes you've mentioned? They all look sensible amendments (IMHO). So looks good to merge 👍

warrenbailey:
👍

PR #399, opened 2016-08-26
Remove folder created for the Python model, but not removed when it was

LJBabbage:

We used the manually created JSON folder as a comparison for tests when we converted from a Python model to JSON. We no longer do this, so this folder is redundant and holds out of date JSON, it should be removed to avoid confusion
### How to review

Make sure the tests work
### Who worked on the PR

@LJBabbage


dhilton:
LGTM 👍

PR #401, opened 2016-08-26
Recover State by replaying post data

### LINKED ISSUE #364: Saved data not retrieved after new Deployment
### Expected behaviour

EQ window should launch with the saved information where we closed the EQ window in point 9
### Actual behaviour
1. EQ launched from Starting without the saved information
2. Filled all the data and clicked on Save & Continue button
3. Error 500 displayed
1. Register a user for MCI survey with instrument file( MCI)
2. Launch and login to the external
3. Verify the status as 'NOT STARTED'
4. Click on 'START' button
5. EQ window is launched
6. Start the survey and complete some fields with data
7. Click on 'Save and Continue' button
8. Close the window
9. Logout from the application
10. Mean while EQ deployment is done for the new release
11. Login into External
    13 Click on 'START' button on portfolio page
#### Browser

IE 11,  Pre-prod
#### Operating System

Windows 7
### Screenshot

c--users-raom-downloads-error500

warrenbailey:

Fixes #364 

Questionnaire state is serialization via jsonpickle to the database. In most case that is used to restore questionnaire state. However if a deployment happens and the GIT revision changes then we replay post data, which we are now also saving, to recover the state. This solves the issue when the python classes can get out of sync between deployments.
### How to review
1. Start a survey
2. Submit at least a single page
3. Manually set the EQ_GIT_REVISION to a value other than the git head revision
4. Restart and the survey and check that the state is recovered.

Part 2 - Safe prod/pre-prod migration
1. Using master and not this branch, start a questionnaire, completing a couple of pages
2. Checkout this branch and try to restore the same session with the same questionnaire
3. It should not be able to restore the state but take you to the first page
4. A record should be included in the logs
### Who worked on the PR

@warrenbailey and @LJBabbage

### DIFF
--- a/app/questionnaire/state_manager.py
+++ b/app/questionnaire/state_manager.py
@@ -14,9 +14,9 @@ STATE = "state"
 class StateManager(object):
     '''
     This class is responsible for saving the state of the User Journey Manager into the database.
-    It does this by pickling the Python object graph into JSON and storing that. Thi means that code
+    It does this by pickling the Python object graph into JSON and storing that. This means that code
     changes between deployments can cause odd behaviour if the database isn't wipe, as when it deserializes
-    state from old python code to new python code attributes can be missing.
+    state from old python code to new python code attributes can be missing. Hence why we have the StateRecovery class
     '''
     @staticmethod
     def has_state():
--- a/app/questionnaire/state_manager.py
+++ b/app/questionnaire/state_manager.py
@@ -129,8 +129,8 @@ class DatabaseStateManager(StateManager):
             logger.debug("Replaying post data %s", post_data)
             location = post_data['location']
             data_to_replay = DatabaseStateManager._convert_to_multi_dict(post_data['post_data'])
-            questionnaire_manager.go_to(location)
             location = questionnaire_manager.process_incoming_answers(location, data_to_replay, replay=True)
+            questionnaire_manager.go_to(location)
             logger.debug("Location %s", location)
         logger.debug("Post data replayed")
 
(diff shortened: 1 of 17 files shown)


app/questionnaire/state_manager.py
```diff
@@ -14,7 +14,7 @@
 class StateManager(object):
     '''
     This class is responsible for saving the state of the User Journey Manager into the database.
-    It does this by pickling the Python object graph into JSON and storing that. This means that code
+    It does this by pickling the Python object graph into JSON and storing that. Thi means that code
```
dhilton: Small thing: missing 's'

app/questionnaire/state_recovery.py
```diff
@@ -0,0 +1,70 @@
...
+
+from flask_login import current_user
+
+
+from werkzeug.datastructures import ImmutableMultiDict, MultiDict
+
+logger = logging.getLogger(__name__)
+
+POST_DATA = "post_data"
+MAX_NO_OF_REPLAYS = 50
```
warrenbailey: Thinking this should be configurable i'll change

collisdigital:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;CAQgrouT&#x2F;503-eq-survey-runner-364-saved-data-not-retrieved-after-new-deployment-issue"> eq-survey-runner: #364 - Saved data not retrieved after new Deployment (Issue)</a>

LJBabbage:
Works ok, but we need to find a good time to check it in, as it will break anything saved or anyone in flight

dhilton:
👍  for rebasing commits </unwatches>

ajmaddaford:
LGTM 👍

PR #389, opened 2016-08-23
eq-525 summary page improvements

weapdiv-david:
### What is the context of this PR?
1. Fixes a bug whereby if a user POSTed to the summary page, then they would automatically be taken to the thank-you page.  
2. Restructures the summary page rendering for easier customisation, simpler templates, and better performance with screen readers.
3. Adds new layout to summary page including sectioning and an 'edit' link for each answer.
4. Improvements to accessibility of summary.
### How to review
1. Checkout the branch
2. Run the tests
3. Complete several surveys and check answer formatting is as expected.
4. Test 'Edit' links
- [ ] Do all commits have sensible commit messages?
- [ ] Is all new code unit tested?
- [ ] Are there integration tests if needed?
- [ ] Is there sufficient logging?
- [ ] Is there documentation or is the code self-describing?
### Who worked on the PR

@weapdiv-david
@hamishtaplin

PR #404, opened 2016-08-31
Routing tests

weapdiv-david:

This PR provides a base test class which could be used to simplify the writing of further integration tests.  It also provides a test which automatically explores all the routes within a questionnaire and compares the summary page against what was submitted.

This is a work in progress and is not yet finished.
### Still to do

1) Fix an issue with checkbox/radio items where the label/value differ. The label is displayed on the summary page, while the value is what gets submitted
2) Fix an issue with date range where TWO answers are contained within a single link, causing the second answer to not appear in the list of answers scraped from the summary page.
3) Integrate the test with the parser, so that a parsed version of the schema is available to the test.
4) The rules for routing are hard coded, but should be easy to collect from the schema through the parser
5) The inputs to the application need to be loaded from somewhere, currently they are hard coded.
### How to review

1) Checkout the branch
2) Review the test_routing file carefully
3) Fix the above issues and re-test
### Who worked on the PR

@weapdiv-david worked on this, but ran out of time

### DIFF
--- a/app/main/views/questionnaire.py
+++ b/app/main/views/questionnaire.py
@@ -35,8 +35,9 @@ def survey(eq_id, collection_id, location):
 
         # redirect to the first block if the change your answers link is clicked
         if location == 'first' or location == 'previous':
+            location = questionnaire_manager.resolve_location(location)
             questionnaire_manager.go_to(location)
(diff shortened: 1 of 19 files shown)


iwootten:
Closing due to integration with https://github.com/ONSdigital/eq-survey-runner/pull/389

PR #407, opened 2016-09-02
Move all surveys to new base theme

collisdigital:
Blocked by #408

Moved all surveys (MCI/RSI etc.) over to the new base theme that is being rolled out.
### How to review

Launch each survey via /dev and check that the new basetheme is being used. Star Wars surveys should still use the starwars theme. See Screenshot.

screen shot 2016-09-02 at 12 49 43
### Who worked on the PR

@collisdigital


LJBabbage:
The Survey title only appears once in the new theme compared to 2 in the old theme, the tests are checking for 2 appearances, so they are all failing

hamishtaplin:
I'm not sure this is the best approach. The plan was to move the code from this theme to the main css/template code once the theme was ready, not to keep it around as a theme.

collisdigital:
Yes agreed; but until that happens it would be good to get all surveys shifted over now so we can start to identify any problems and gain wider visibility of the updates. We can always convert these back to "default" later.

hamishtaplin:
Righto. This definitely needs extensive cross-browser testing before merging then.

collisdigital:
Benn testing it now

bennpearson:
@yoganandkunche tested this and created an issue here: https://github.com/ONSdigital/eq-survey-runner/issues/412

Merging this now as it is a minor issue related to lte IE9 and Galaxy Tab

PR #418, opened 2016-09-08
Fix for issue 417 non-mandatory error causing error after being emptied

### LINKED ISSUE #417: Non mandatory fields fail validation if a previous value is enter then removed
### Expected behaviour

If a non mandatory field is populated, which fails validation but then the value is removed, validation should pass
### Actual behaviour

The first validation failure is presented to the user and you can only preceed by refilling the field

Open 1_205.json, fill in the field 'What was the value of the business's total sales of clothing and footwear?' with text. See the error message appear, now empty the question and submit, see the error message still exists even though it is non-mandatory

LJBabbage:

Fixes #417 
### How to review

see #417 
### Who worked on the PR

@LJBabbage

### DIFF
--- a/app/questionnaire_state/answer.py
+++ b/app/questionnaire_state/answer.py
@@ -11,16 +11,18 @@ class Answer(Item):
         self.input = None
 
     def update_state(self, user_input):
+
+        # Clear any previous value and validation results
+        self.value = None
+        self.is_valid = True
+        self.errors = []
         # Get the user input
-        self.value = None  # Clear any previous value
         self.input = self.schema_item.get_user_input(user_input)
 
         # Try and get the typed value
         if self.input:
             try:
                 self.value = self.schema_item.get_typed_value(user_input)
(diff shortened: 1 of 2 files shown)


app/questionnaire_state/answer.py
```diff
@@ -11,16 +11,18 @@ def __init__(self, id, schema_item):
...
 
         # Try and get the typed value
         if self.input:
             try:
                 self.value = self.schema_item.get_typed_value(user_input)
-                self.is_valid = True
-                self.errors = []
             except TypeCheckingException as e:
                 self.is_valid = False
                 self.errors = []
```
iwootten, line 28: Is this necessary? Doesn't errors default to empty?

LJBabbage:
Thats where the problem lies, old errors are being recorded in state and then aren't being cleared.

iwootten:
Looks good to me!

PR #424, opened 2016-09-12
Issue 422 numeric zero sdx

### LINKED ISSUE #422: RSI - '0' values entered for input are not sent to sdx
### Expected behaviour

All zero '0' values entered by the the user are passed as part of the payload to sdx.
### Actual behaviour

Zero value fields are not sent in payload.

Launch an RSI 0112 form type enter '0' for Total Retail sales and submit the survey.
### Screenshot

https://cloud.githubusercontent.com/assets/21335236/18432642/00ff3854-78db-11e6-8edb-3d02033cec18.png

On further investigation, it would seem all values entered as '0' are not passed to sdx. The screenshot below show the data sent when entering 0 values for all numeric fields on rsi.

screen shot 2016-09-12 at 12 44 15

iwootten:

Allow non-null zero values to be encoded in the converter. Fixes #422. 
### How to review
- Add a log line to the submitter to app/submitter/submitter.py to print the unencrypted JSON payload before submission
- Complete a survey with zeros in relevant inputs and confirm the generated JSON contains the zeros
- [x] Is all new code unit tested?
- [x] Are there integration tests if needed?
- [x] Is there sufficient logging?
### Who worked on the PR

@iwootten

### DIFF
--- a/tests/app/submitter/test_converter.py
+++ b/tests/app/submitter/test_converter.py
@@ -97,7 +97,7 @@ class TestConverter(SurveyRunnerTestCase):
             questionniare.register(answer_1)
             questionniare.register(answer_2)
 
-            answer_object, submitted_at = Converter.prepare_answers(user, metadata, questionniare, user_answer)
+            answer_object, submitted_at = Converter.prepare_answers(metadata, questionniare, user_answer)
 
             self.assertEquals(EXPECTED_RESPONSE["type"], answer_object["type"])
             self.assertEquals(EXPECTED_RESPONSE["version"], answer_object["version"])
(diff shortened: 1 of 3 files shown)


tests/app/submitter/test_converter.py
```diff
@@ -108,5 +106,44 @@ def test_prepare_answers(self):
             self.assertEquals(EXPECTED_RESPONSE["paradata"], answer_object["paradata"])
             self.assertEquals(EXPECTED_RESPONSE["data"], answer_object["data"])
 
+    def test_zero_answers(self):
```
ajmaddaford: I think this method name implies we're testing a survey with no answers, rather than an answer that is zero. Maybe something like `test_answer_with_zero`

ajmaddaford (approved):
LGTM 👍

PR #426, opened 2016-09-13
EQ-525 Summary page improvements

LJBabbage:

Include links on the summary page and introduce a registry for formatting rules
### How to review

Using star wars, complete the questionnaire and make sure everything works and the summary page displays all the questions asked and answer, also that the links work. 

(n.b. It is known that you must complete the survey again if using a link, e.g. you can't return to the summary page directly, this will be picked up in another story)
### Who worked on the PR

@LJBabbage


app/templating/summary/formatters/abstract_formater.py
```diff
@@ -0,0 +1,9 @@
+from abc import ABCMeta, abstractmethod
+
+
+class AbstractFormater(metaclass=ABCMeta):
```
iwootten: Typo here.

LJBabbage:
Is the rest ok then?

iwootten:
@LJBabbage think there may be a couple of styling issues here that haven't yet made it across from the other branch. Here's what I get on the summary page at the mo after a compile of the frontend. 

Been through with @hamishtaplin and he's happy to jump in if necessary.

screen shot 2016-09-13 at 16 23 02

app/templating/summary/summary_item.py
```diff
@@ -0,0 +1,32 @@
...
+
+        if user_answer:
+            answer_type = self.schema.answers[0].type.upper()
+
+            # find out if the question type and answer has a formatter
+            formatter_class = FormatterRegistry.get_formatter(answer_type, self.question_type)
+            if formatter_class:
+                formatter_rule = formatter_class()
+                return formatter_rule.format(self.schema, self.state, user_answer)
+            else:
```
iwootten: Could lose the else here and just fallback to return user_answer

LJBabbage:
Star wars theme is good, but the default one is not

app/templating/summary/formatters/check_box.py
```diff
@@ -0,0 +1,13 @@
+from app.templating.summary.formatters.abstract_formater import AbstractFormater
+
+
+class CheckBoxFormatter(AbstractFormater):
+
+    @staticmethod
+    def format(schema, state, user_answer):
+        answers = []
+        for option in schema.answers[0].options:
```
iwootten: We only seem to be using answers here, not the schema

app/templating/summary/formatters/date_range.py
```diff
@@ -0,0 +1,15 @@
+from app.templating.summary.formatters.abstract_formater import AbstractFormater
+
+
+class DateRangeFormatter(AbstractFormater):
+
+    @staticmethod
+    def format(schema, state, user_answer):
```
iwootten: We only seem to be using the answers from both schema, state - maybe we should only pass them? Each of the format methods on these class children appear to have different requirements for parameters, which means we'll need to mock them all out as part of unit tests, even though some params won't be used.

app/templating/summary/formatters/radio_button.py
```diff
@@ -0,0 +1,10 @@
+from app.templating.summary.formatters.abstract_formater import AbstractFormater
+
+
+class RadioButtonFormatter(AbstractFormater):
+
+    @staticmethod
+    def format(schema, state, user_answer):
```
iwootten: Again, we only seem to be using an answer from schema and user_answer

app/templating/summary/formatters/check_box.py
LJBabbage: We use schema in schema.answers[0].options, but we can tighten up as suggested on the other comments to just pass schema.answers.

iwootten: Ah, yes my mistake.

hamishtaplin:
@LJBabbage this should be ok now

app/templates/summary.html
```diff
@@ -0,0 +1,82 @@
...
+
+              {% else %}
+
+                {{not_answered}}
+
+              {% endif %}
+
+            {% elif item.type == 'textarea' %}
+
+              {{item.answer|nl2br or not_answered}}
```
hamishtaplin: I'm getting a string "None" here instead of a falsy value when the comments aren't entered.

hamishtaplin: Turns out this was caused by the filter, have fixed.

app/templates/confirmation.html
```diff
@@ -3,10 +3,11 @@
 {% block page_title -%}Summary{% endblock -%}
 
 {% block survey -%}
-  <div class="header">
-    <h1 class="header__title h1">Confirmation</h1>
-    <h2 class="header__subtitle">Thank you for your answers, do you wish to submit</h2>
-  </div>
+
+<div class="header">
+  <h1 class="header__title h1">Confirmation</h1>
```
bennpearson: Should be planet name

app/templates/confirmation.html
```diff
@@ -3,10 +3,11 @@
...
+  <h2 class="header__subtitle">Thank you for your answers, do you wish to submit</h2>
```
bennpearson, line 9: Should contain planet class

app/assets/styles/partials/components/_summary.scss
```diff
@@ -0,0 +1,121 @@
...
+    bottom: 0;
+  }
+}
+
+.summary__edit-link {
+  padding: 0.5rem 2rem;
+  display: flex;
+  align-items: center;
+  &:hover {
+    background-color: #efefef;
```
bennpearson: Should be a var?

app/templates/submission.html
```diff
@@ -1,4 +1,4 @@
-{% extends theme('layouts/_survey.html') -%}
+{% extends 'layouts/_survey.html' -%}
```
bennpearson: Note: Should this be theme()

yoganandkunche:
@hamishtaplin @LJBabbage  as discussed can you please change the 'Not answered' to 'No answer provided' to match requirements in acceptance criteria.

app/templating/summary/formatter_registry.py
```diff
@@ -0,0 +1,28 @@
+from app.templating.summary.formatters.check_box import CheckBoxFormatter
+from app.templating.summary.formatters.currency import CurrencyFormatter
+from app.templating.summary.formatters.date_range import DateRangeFormatter
+from app.templating.summary.formatters.radio_button import RadioButtonFormatter
+
+
+KNOWN_FORMATTERS = {
+    "GENERAL": {
```
iwootten, line 8: These formatters all appear to apply to single answer fields, whilst daterange is comprised of multiple. Would it be better to name these keys 'SINGLE"/"MULTIPLE" or similar and change the subkey "DATE" to "DATERANGE"? 'GENERAL' could cover everything.

app/templating/summary/formatters/abstract_formater.py
```diff
@@ -0,0 +1,9 @@
+from abc import ABCMeta, abstractmethod
+
+
+class AbstractFormatter(metaclass=ABCMeta):
+
+    @staticmethod
+    @abstractmethod
+    def format(schema_answers, state_answers, user_answer):
```
iwootten, line 8: Are we able to make use of keyword arguments here in order to allow child classes to omit variables that aren't used in formatting?

tests/app/templating/test_question_formatters.py
```diff
@@ -0,0 +1,114 @@
...
+
+    def test_general_text_area(self):
+
+        original_answer = 'text area test'
+        expected_format = 'text area test'
+        Question_type = 'GENERAL'
+        answer_type = 'TEXTAREA'
+        self.check_formatter(answer_type, Question_type, expected_format, original_answer)
+
+    def test_general_Radio(self):
```
iwootten: Seems to be some weird camelcasing of method names throughout this file.

tests/integration/routing/test_routing.py
```diff
@@ -0,0 +1,75 @@
...
+from tests.integration.create_token import create_token
+from tests.integration.integration_test_case import IntegrationTestCase
+
+
+class TestRouting(IntegrationTestCase):
+
+    def setUp(self):
+        super().setUp()
+
+    def test_routing_paths(self):
```
iwootten, line 11: This is a boatload easier to understand than what we had previously!

iwootten (changes requested):
In addition to my comments, have yogi's concerns been addressed - do we need to change the content of no_answers?

app/templating/summary/formatter_registry.py
LJBabbage: It matches the JSON format, so possibly something for the future

hamishtaplin:
@iwootten Wording was changed here, as per Yogi's comment: https://github.com/ONSdigital/eq-survey-runner/pull/426/files#diff-9877ee549e610942cbd6fa832b7411b7R5

iwootten:
@hamishtaplin cheers!

app/templating/summary/formatters/abstract_formater.py
LJBabbage: As discussed, this doesn't really lend itself at the moment, with the updated questionnaire state model this should improve

app/templating/summary/summary_item.py
```diff
@@ -0,0 +1,37 @@
...
+        self.link = schema.container.container.id + '#' + schema.answers[0].id
+        self.type = self.schema.answers[0].type.lower()
+
+    # Formatting the answer for the frontend e.g. currency needs £
+    def format_answer(self):
+
+        user_answer = self.state_answers[0].value
+
+        if user_answer:
+            answer_type = self.schema.answers[0].type.upper()
```
ajmaddaford: I think this should be uppercased in the `get_formatter` method as it's only being uppercased to do the lookup against the dictionary in the `FormatterRegistry`.

app/templating/summary/summary_item.py
```diff
@@ -0,0 +1,37 @@
...
+
+            # find out if the question type and answer has a formatter
+            formatter_class = FormatterRegistry.get_formatter(answer_type, self.question_type)
+            if formatter_class:
+                logger.debug("formatting answer %s", user_answer)
+                formatter_rule = formatter_class()
```
ajmaddaford: Not sure where rule comes from, would be better to go with `formatter = formatter_class()`

app/templating/summary/section_summary.py
```diff
@@ -0,0 +1,30 @@
+import logging
+
+from app.templating.summary.summary_item import SummaryItem
+
+logger = logging.getLogger(__name__)
+
+
+class SectionSummary(object):
```
ajmaddaford: This should be called `SummarySection` to be consistent with `SummaryItem`

app/templating/summary_template_preprocessor.py
```diff
@@ -14,19 +14,24 @@ def build_view_data(self, node, schema):
...
-            node = node.next
+        while node:
+            # Not all nodes (pages) will have sections e.g. Introduction
             if node.state.display_on_summary:
-                states.append(node.state)
-        return states
+                for section in node.state.sections:
+                    section_schema = schema.get_item_by_id(section.id)
+                    section_summary = SectionSummary(section_schema, section)
+                    summary_section.append(section_summary)
```
ajmaddaford: If `SectionSummary` is renamed `SummarySection`, `summary_section` should be renamed `summary_sections`

app/templating/summary/formatter_registry.py
```diff
@@ -0,0 +1,28 @@
...
+    "DATERANGE": {
+                'DATE': DateRangeFormatter,
+    },
+}
+
+
+class FormatterRegistry(object):
+
+    @staticmethod
+    def get_formatter(answer_type, question_type):
```
ajmaddaford: Minor, `question_type` should be the first argument as it's the more significant parameter (the logic of the method uses it first and doesn't carry on if it doesn't match)

ajmaddaford (approved):

iwootten (approved):

iwootten:
[Update to Python 3.8](https://trello.com/c/5pBfWFfU/3849-update-to-python-38)

PR #427, opened 2016-09-13
Add classes to allow selenium to drive page

### LINKED ISSUE #423: Some fields on http://localhost:5000/dev launch page does not have 'class' or 'ID'
### Expected behaviour

All fielde must have a class/ID so that selenium can drive the tests
### Actual behaviour

Some fields (example: Period ID and Period String) does not have class/ID making it impossible for selenium to drive

Launch http://localhost:5000/dev and inspect on the fields

**Business Requirement**
Selenium cannot drive without ID/Class

iwootten:

Add classes to fields without them on /dev. Fixes #423 
### How to review

Check that classes exist for each input
### Who worked on the PR

@iwootten

### DIFF
--- a/app/dev_mode/templates/dev-page.html
+++ b/app/dev_mode/templates/dev-page.html
@@ -19,15 +19,15 @@
         </p>
         <p>
           Token Expiry (seconds):
-          <input name="exp" type="text" value="1800">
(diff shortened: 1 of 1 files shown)


yoganandkunche:
This defect passed re-test. Happy to QA passed.

PR #435, opened 2016-09-16
Issue 379 Multiple surveys being used

### LINKED ISSUE #379: Undefined Behaviour with Simultaneous Sessions
### Expected behaviour

An error is displayed when attempting to alternate between two simultaneous sessions.
### Actual behaviour

Data is submitted to the wrong questionnaire and silently works/ignores incorrect POST fields and maps over any whose ids that match
- Launch two different questionnaires in the same browser in different tabs
- The last one launched will be the 'active' one
- Try POSTing data to the first one launched and see that your are directed to the second one
- If you find two questionnaires with the same id for fields (such as Data Range) you can do the above steps and see the data is submitted to the 'other' questionnaire.
### Technical information
- Due to the session info stored in the cookie a user cannot launch two questionnaires at the same time in the same browser (they could do it with an incognito/private browser tab or separate browsers). 
- The session id in the cookie will be whichever was last set (i.e. the last authenticated session); they will have essentially been logged out of the first session when launching the second but the browser will still be displaying the first questionnaire content on-screen and allow the user to perform a POST.
- The proposal here isn't to support simultaneous questionnaires but instead to present an error when submitting the 'wrong' questionnaire (i.e. when the session has been replaced by another one).
#### Browser

All
#### Operating System

All
### Screenshot

N/A

hamishtaplin:

Fixes #379.

screenshot 2016-09-19 09 27 24
### How to review

As per issue reproduction steps.
### Who worked on the PR


hamishtaplin:
@LJBabbage tests are failing but it doesn't seem to be related to what I've done?

alexmorris:
info

@hamishtaplin this will work for now

app/frontend_messages.py
```diff
@@ -0,0 +1,4 @@
+messages = {
+  'MULTIPLE_SURVEYS': "Unfortunately you can only complete one survey at a time.",
```
collisdigital: Don't forget to use gettext here so strings can be translated (prefix them with _ etc).

collisdigital: You may want to consider more descriptive/long constant names if this dictionary is going to grow

app/main/system_message.py
```diff
@@ -0,0 +1,25 @@
...
+from flask_login import current_user
+
+from ua_parser import user_agent_parser
+
+logger = logging.getLogger(__name__)
+
+
+def system_message(message):
+    logger.debug(message)
+    tx_id = None
```
collisdigital: If tx is isn't being shown anymore is this needed?

app/main/system_message.py
```diff
@@ -0,0 +1,25 @@
...
+    metadata = MetaDataStore.get_instance(current_user)
+    if metadata:
+        tx_id = convert_tx_id(metadata.tx_id)
+    user_agent = user_agent_parser.Parse(request.headers.get('User-Agent', ''))
```
collisdigital: If user agent isn't being shown is this needed?

app/templates/layouts/_error.html
```diff
@@ -2,29 +2,39 @@
...
-      <p>Please quote this reference for any support queries.</p>
-      <ul class="list list--boxes">
-        {%- for id_part in tx_id -%}
-          <li class="list__item">{{id_part}}</li>
-        {%- endfor -%}
-      </ul>
-    {%- endif -%}
+
+    {% block error_message %}
+      <strong>Sorry, something has gone wrong.</strong>
```
collisdigital, line 15: If your moving the message into a dict and perhaps worth doing these too now?

app/templates/system_message.html
```diff
@@ -0,0 +1,17 @@
+{% extends 'layouts/_survey.html' %}
+
+{% block page_title %}Information{% endblock %}
```
collisdigital: Should these strings also move to the dict?

app/templates/system_message.html
```diff
@@ -0,0 +1,17 @@
...
+
+{% block main %}
+  <div class="header">
+    <h1 class="header__title">
+      {% block error_title %}Information{% endblock %}
```
collisdigital: And here

app/main/system_message.py
```diff
@@ -0,0 +1,25 @@
...
+from flask.ext.themes2 import render_theme_template
+
+from flask_login import current_user
+
+from ua_parser import user_agent_parser
+
+logger = logging.getLogger(__name__)
+
+
+def system_message(message):
```
collisdigital: Codecov is reporting no test coverage of this function, this implies we have no component test to verify this is working, see https://codecov.io/gh/ONSdigital/eq-survey-runner/src/0539aa4bce732111df8de49221be6a741b33c909/app/main/system_message.py

PR #442, opened 2016-09-19
Move info_to_provide copy beneath title

### LINKED ISSUE #441: Star Wars Landing page has some repetition
### Expected behaviour

Configured information_to_provide list appears in same position as upon other surveys.
### Actual behaviour

Appears above title, with repetition of "If actual figures.." copy.

Open the star wars survey.
### Screenshot

screen shot 2016-09-19 at 15 54 51

iwootten:

Moved copy for information_to_provide beneath the landing page title, inline with rest of copy. Reuse existing header html. Fixes #441.
### How to review

Open the Star Wars survey, see if there is any duplication. See if it renders as per screenshot
screen shot 2016-09-19 at 16 03 07
### Who worked on the PR

@iwootten

### DIFF
--- a/app/templates/landing-page.html
+++ b/app/templates/landing-page.html
@@ -45,21 +45,6 @@
     <div>
 
       <div class="header">
-        {% if meta.survey.information_to_provide %}
-        <h3>You will be asked to provide information for the business, including:</h3>
-        <ul>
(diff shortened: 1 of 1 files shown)


hamishtaplin:
@iwootten Unrelated but the font isn't loading for some reason. Would be good to find out why in case it's a problem that happens in prod.

PR #454, opened 2016-09-22
Fix for issue-379 Multiple surveys open at one time

### LINKED ISSUE #379: Undefined Behaviour with Simultaneous Sessions

LJBabbage:

fixes #379. Changes the URL on survey runner to include the form_type and the period string. This is then used to compare against the metadata store to see if they are different
### How to review

see #379. Open up one survey and start work on it. Open a separate window with the same session information and start filling it in. Revert back to the first one and make sure it sends you to the information page
### Who worked on the PR

@LJBabbage and @hamishtaplin


app/frontend_messages.py
```diff
@@ -0,0 +1,15 @@
...
+MESSAGES = {
+            'multiple-surveys': {
+                'TITLE': _("Information"),
+                'MSG': _("Unfortunately you can only complete one survey at a time."),
+                'INSTRUCTIONS': _("Close this window to continue with your current survey."),
+                },
+            }
+
+
+def get_message(message):
```
collisdigital: Zen: Explicit is better than implicit: What does this return if message isn't found

collisdigital: Zen: Readability counts: is this getting a single message or messages, what is it looking up? i.e. is the function name and parameter clear as to what it is doing and required?

app/main/views/root.py
```diff
@@ -25,6 +28,16 @@ def root():
     return errors.index()
 
 
+@main_blueprint.route('/information/<error>', methods=['GET'])
```
collisdigital: Note: these aren't 'errors' they are system message.

app/templates/information.html
```diff
@@ -0,0 +1,17 @@
+{% extends 'layouts/_survey.html' %}
+
+{% block page_title %}{{messages.TITLE}}{% endblock %}
+
+{% block main %}
+  <div class="header">
+    <h1 class="header__title">
+      {% block error_title %}{{messages.TITLE}}{% endblock %}
```
collisdigital: again, not an error here.

collisdigital (commented):
Zen!

app/templates/information.html
```diff
@@ -0,0 +1,17 @@
...
+    </h1>
+  </div>
+
+  <div class="alert alert--simple alert--error alert--spacious">
```
iwootten, line 12: Following on from @collisdigital - if this isn't an error - are these css classes therefore correct?

tests/integration/mci/test_empty_submission.py
```diff
@@ -23,11 +23,11 @@ def empty_submission(self, form_type_id, eq_id):
...
             'action[start_questionnaire]': 'Start Questionnaire'
         }
-        resp = self.client.post('/questionnaire/' + eq_id + '/789/introduction', data=post_data, follow_redirects=False)
+        resp = self.client.post('/questionnaire/' + eq_id + '/0205/201604/789/introduction', data=post_data, follow_redirects=False)
         self.assertEquals(resp.status_code, 302)
 
         # We proceed to the questionnaire
-        resp = self.client.get('/questionnaire/' + eq_id + '/789/cd3b74d1-b687-4051-9634-a8f9ce10a27d', follow_redirects=True)
+        resp = self.client.get('/questionnaire/' + eq_id + '/0205/201604/789/cd3b74d1-b687-4051-9634-a8f9ce10a27d', follow_redirects=True)
         self.assertEquals(resp.status_code, 200)
```
iwootten, line 32: There's a whole bunch of assertions going on in this single method, seems like it was as this originally. It would be better to break down into stages. Additionally the empty_submission method is only ever called once, so it's unnecessary for it to be broken out into another method.

tests/integration/mci/test_information_page.py
```diff
@@ -0,0 +1,52 @@
...
+        self.assertRegexpMatches(content, 'Information')
+        self.assertRegexpMatches(content, 'Unfortunately you can only complete one survey at a time.')
+        self.assertRegexpMatches(content, 'Close this window to continue with your current survey.')
+
+    def test_information_page_missing_message(self):
+
+        resp = self.client.get('/information/test', follow_redirects=False)
+        self.assertEquals(resp.status_code, 404)
+
+    def test_different_metadat_store_to_url(self):
```
iwootten: Dropped an 'a' on this method name

tests/integration/mci/test_empty_submission.py
```diff
@@ -23,11 +23,11 @@ def empty_submission(self, form_type_id, eq_id):
         post_data = {
             'action[start_questionnaire]': 'Start Questionnaire'
         }
-        resp = self.client.post('/questionnaire/' + eq_id + '/789/introduction', data=post_data, follow_redirects=False)
+        resp = self.client.post('/questionnaire/' + eq_id + '/0205/201604/789/introduction', data=post_data, follow_redirects=False)
```
iwootten: The base of the urls here are the same throughout this file, we could wrap it up as a var to save on line length :) 

This goes for the other test files changed too.

app/frontend_messages.py
```diff
@@ -0,0 +1,15 @@
+from flask.ext.babel import gettext as _
+
+
+MESSAGES = {
+            'multiple-surveys': {
+                'TITLE': _("Information"),
+                'MSG': _("Unfortunately you can only complete one survey at a time."),
```
iwootten: I was under the impression the copy for this message had been changed? Why's it unfortunate?

iwootten: Additionally, shouldn't we be adding all existing messages to this structure rather than creating something new?

tests/test_front_end_messages.py
```diff
@@ -0,0 +1,14 @@
...
+from flask.ext.babel import gettext as _
+
+EXPECTED_MESSAGE = {
+                  'TITLE': _("Information"),
+                  'MSG': _("Unfortunately you can only complete one survey at a time."),
+                  'INSTRUCTIONS': _("Close this window to continue with your current survey."),
+                  }
+
+class TestFrontEndMessage(unittest.TestCase):
+    def test_front_end_messages(self):
```
iwootten: This test seems a bit redundant. i.e. Does the get_message lookup the only message we have coded it to return?

iwootten (changes requested):

tests/integration/mci/test_empty_submission.py
LJBabbage: Done, but I can't get them all, as some are dynamic

tests/integration/mci/test_empty_submission.py
LJBabbage: One for a different story as its not connect to the work in this one

app/frontend_messages.py
LJBabbage: We need another story for all the other messages

iwootten (approved):

app/frontend_messages.py
```diff
@@ -0,0 +1,20 @@
...
+from flask.ext.babel import gettext as _
+
+logger = logging.getLogger(__name__)
+
+MESSAGES = {
+            'multiple-surveys': {
+                'TITLE': _("Information"),
+                'MSG': _("Unfortunately you can only complete one survey at a time."),
+                'INSTRUCTIONS': _("Close this window to continue with your current survey."),
+                },
```
ajmaddaford, line 12: I think that the keys should be lower cased and `MSG` should be changed to `message`

app/frontend_messages.py
```diff
@@ -0,0 +1,20 @@
...
+            }
+
+
+def get_messages(system_message):
```
ajmaddaford: `system_message` is only valid for our current use case, as this has been developed as a generic message loader, this would be more appropriately named something like `message_identifier`

app/frontend_messages.py
```diff
@@ -0,0 +1,20 @@
...
+    if system_message in MESSAGES:
+        return MESSAGES['multiple-surveys']
```
ajmaddaford: This will always return the same messages regardless of what was passed in!

ajmaddaford (changes requested):

app/templates/information.html
hamishtaplin: It may not be an error but it looks like one (it's red) so the class is valid.

ajmaddaford (approved):

PR #464, opened 2016-09-29
Remove redundant code

iwootten:

Remove code that is no longer used. Specifically the /app/patternlib and /app/surveys folders.
### How to review

Check whether jenkins tests pass. Manually run any survey and ensure works as previous.


benjefferies:
Looks good to me, also pushes up our code coverage from 78 to 89

PR #466, opened 2016-09-29
Remove saving of Questionnaire data from auth class

iwootten:

Break the saving of questionnaire data out of our extension to the flask auth class and put it into it's own dedicated class, QuestionnaireStore. Create and store this object against the flask global, 'g' to be used on subsequent questionnaire data queries within the context of a request. More info here: http://flask.pocoo.org/docs/0.11/appcontext/

[Edit] Additionally have removed the save/get/encode/decode responsibility of the metadata store class, making it part of the QuestionnaireStore behaviour. This means the MetadataStore can have a single responsibility of parsing the metadata, rather than doing this as part of save_instance() and thus is renamed MetadataParser.
### How to review

No behaviour change has been introduced, so check unit tests continue to pass.


app/authentication/authenticator.py
```diff
@@ -55,18 +55,16 @@ def jwt_login(self, request):
...
+        user = User(user_id, user_ik)
+
         # store the meta data
         metadata = MetaDataStore.save_instance(user, token)
 
-        user.save()
-
         logger.info("User authenticated with tx_id=%s", metadata.tx_id)
 
         return user
```
benjefferies: Do we need to create and return the user object as it's not used?

app/data_model/questionnaire_store.py
```diff
@@ -0,0 +1,49 @@
...
+        else:
+            logger.debug("User %s does not have previous data creating", user_id)
+            self.save()
+
+    def delete(self):
+        logger.debug("Deleting questionnaire data for %s", self.user_id)
+        self.data = {}
+        self.storage.delete(self.user_id)
+
+    def save(self):
```
benjefferies, line 34: Would it make sense to take the data as a parameter here instead of mutating the field and then calling this?

app/data_model/questionnaire_store.py
```diff
@@ -0,0 +1,49 @@
...
+            self.user_ik = user_ik
+        else:
+            raise ValueError("No user_id or user_ik found in session")
+
+        self.storage = StorageFactory.get_storage_mechanism()
+
+        if self.storage.has_data(self.user_id):
+            logger.debug("User %s has previous data loading", user_id)
+            self.data = self.storage.get(self.user_id, self.user_ik)
+        else:
```
benjefferies: There is going to be no data at this point, do we need to called save?

app/questionnaire/user_action_processor.py
```diff
@@ -79,7 +79,7 @@ def __init__(self, schema, metadata):
         self.action = 'save_continue'
 
     def perform_action(self):
-        current_user.save()
+        pass
```
benjefferies, line 82: Observation: Using this pattern for UserAction seems somewhat like overkill now there is only one concrete implementation for perform_action. Would a rest api for submit not suffice here?

app/data_model/questionnaire_store.py
iwootten: Don't think so, - the delete method removes the user questionnaire.

app/authentication/authenticator.py
iwootten: Nope, I'll remove this - will require refactoring the metadatastore.save_instance() too.

app/authentication/authenticator.py
```diff
@@ -55,22 +55,16 @@ def jwt_login(self, request):
...
 
         # store the meta data
-        metadata = MetaDataStore.save_instance(user, token)
-
-        user.save()
+        metadata = MetaDataStore.save_instance(user_id, user_ik, token)
 
         logger.info("User authenticated with tx_id=%s", metadata.tx_id)
 
-        return user
```
benjefferies, line 72: Worth updating the return doc on line 39

app/data_model/questionnaire_store.py
iwootten: I think I prefer this approach, given the data in the class will be used globally and it doesn't require returning and modifying attributes and then passing then back. e.g. If say you wanted to just update metadata.

app/questionnaire/user_action_processor.py
iwootten: Possibly - maybe should be tackled as another piece of work?

app/authentication/authenticator.py
iwootten: Done

app/data_model/questionnaire_store.py
benjefferies: So in theory we don't need to call save then?

app/data_model/questionnaire_store.py
benjefferies: Fair enough, maybe it just feels weird for me coming from the java world

app/questionnaire/user_action_processor.py
benjefferies: Definitely

app/data_model/questionnaire_store.py
```diff
@@ -0,0 +1,70 @@
...
+from app.storage.storage_factory import StorageFactory
+
+from flask import g
+
+import jsonpickle
+
+logger = logging.getLogger(__name__)
+
+
+class QuestionnaireStore:
```
benjefferies, line 12: Should we have a test for this class?

tests/app/templating/test_summary_template_preprocessor.py
```diff
@@ -1,38 +1,36 @@
...
+            MetadataConstants.TRAD_AS.claim_id: "Apple",
+            MetadataConstants.RETURN_BY.claim_id: "2016-07-07",
+            MetadataConstants.TRANSACTION_ID.claim_id: "4ec3aa9e-e8ac-4c8d-9793-6ed88b957c2f"
         }
         with self.application.test_request_context():
-            user = User("1", "2")
-            self.metadata_store = MetaDataStore.save_instance(user, self.jwt)
+            self.metadata = MetadataParser.parse_token(self.jwt)
 
     def get_metadata_store(self):
```
benjefferies: Is it with renaming to get_metadata?

tests/app/templating/test_questionnaire_template_preprocessor.py
```diff
@@ -1,37 +1,35 @@
...
+            MetadataConstants.TRANSACTION_ID.claim_id: "4ec3aa9e-e8ac-4c8d-9793-6ed88b957c2f"
         }
         with self.application.test_request_context():
-            user = User("1", "2")
-            self.metadata_store = MetaDataStore.save_instance(user, self.jwt)
+            self.metadata = MetadataParser.parse_token(self.jwt)
 
     def get_metadata_store(self):
-        return self.metadata_store
+        return self.metadata
```
benjefferies, line 32: Is it with renaming to get_metadata?

tests/app/templating/test_metadata_template_preprocessor.py
```diff
@@ -1,34 +1,31 @@
...
+            MetadataConstants.TRAD_AS.claim_id: "Apple",
+            MetadataConstants.RETURN_BY.claim_id: "2016-07-07",
+            MetadataConstants.TRANSACTION_ID.claim_id: "4ec3aa9e-e8ac-4c8d-9793-6ed88b957c2f"
         }
         with self.application.test_request_context():
-            user = User("1", "2")
-            self.metadata_store = MetaDataStore.save_instance(user, self.jwt)
+            self.metadata = MetadataParser.parse_token(self.jwt)
 
     def get_metadata_store(self):
```
benjefferies: Is it with renaming to get_metadata?

ajmaddaford (approved):
This LGTM, much clearer 👍

PR #470, opened 2016-09-30
issue-469 Fix "ga is not defined" error

### LINKED ISSUE #469: Javascript fails when Google Analytics is not loaded
### Expected behaviour

Javascript execution should not be affected by the absence of the global Google Analytic `ga` object. In short: failure for the GA code to load should not affect the rest of the app.
### Actual behaviour

Javascript execution halts with error `Uncaught ReferenceError: ga is not defined`, causing subsequent JS to not execute.
1. Make sure you **do not** have the `EQ_UA_ID` environment variable set as this will cause the app to load Google Analytics.
2. Visit any survey.
3. Invoke a validation error, which will cause the app to attempt to send an event to GA.
4. Confirm JHS error in console
### Screenshot

screenshot 2016-09-30 15 11 13

hamishtaplin:

Fixes #469. 

Analytics event tracking now checks if the global `ga` object is defined before attempting to execute code. This fixes an issue whereby an environment that does not have Google Analytics JS loaded (eg. Sauce Labs) will cause all further JS execution to halt.
### How to review

As per reproduction steps in the bug.

Also, we need to check that the event tracking _does_ still work as expected when the environment variable is set.

### DIFF
--- a/app/assets/js/app/modules/analytics.js
+++ b/app/assets/js/app/modules/analytics.js
@@ -2,6 +2,10 @@ import forEach from 'lodash/forEach'
 import domready from './domready'
 
 export default function initAnalytics() {
+  if (typeof ga === 'undefined') {
+    return false
+  }
+
   const errors = document.querySelectorAll('[data-error=true]')
   const guidances = document.querySelectorAll('[data-guidance]')
 


benjefferies:
Looks good and tested by cherry picked change to webdriver-page-model branch and it fixes the issue I was seeing with ga

bennpearson (approved):

PR #474, opened 2016-10-04
Docker

iwootten:

Added a docker-compose.yml to represent postgres, runner and rabbit in single configuration. I'm aware we could cleanup build dependencies after the build, but for some reason cryptography requires cache to build. Given the compilation time of each layer, it makes sense to make it all available as the overhead is a few hundred megabytes. Also had to update to 0.3 of flake8-mock due to failed compilation.
### How to review

Suggest any optimisations that stand out.


docker-compose.yml
```diff
@@ -0,0 +1,43 @@
+version: '2'
+services:
+  rabbit:
+    image: rabbitmq:3-management
+    ports:
+      - "5672:5672"
```
benjefferies, line 6: Do we need to externally expose these ports. I think survey runner should be able to access the container as they should be linked

docker-compose.yml
```diff
@@ -0,0 +1,43 @@
...
+      RABBITMQ_DEFAULT_USER: digitaleq
+      RABBITMQ_DEFAULT_PASS: digitaleq
+  db: 
+    image: postgres
+    environment:
+      POSTGRES_PASSWORD: digitaleq
+      POSTGRES_USER: digitaleq
+      POSTGRES_DB: runner
+    ports:
+      - "5433:5432"
```
benjefferies, line 20: Why 5433 to 5432?

docker-compose.yml
```diff
@@ -0,0 +1,43 @@
...
+    volumes:
+      - ./:/code
+    depends_on:
+      - db
+      - rabbit
+    networks:
+      - eq-env
+    ports:
+      - "5000:5000"
+    command: ./scripts/run_app.sh
```
benjefferies: Should this be /code/run_app.sh?

docker-compose.yml
jonnyshaw89: if we go down that route we then also won't need the management version

docker-compose.yml
iwootten: So a docker doesn't attempt to use the same port as a locally installed postgres.

docker-compose.yml
iwootten: Yes, so we can inspect the queue from our local machines.

docker-compose.yml
iwootten: No :) It executes from within the work dir so /code/scripts/run_app.sh

benjefferies: On second thoughts is it necessary because it's an entrypoint for the container anyway?

docker-compose.yml
benjefferies: Ah, makes sense. I guess this is if we want to connect to it externally to inspect the db

docker-compose.yml
iwootten: Ah, yeah no, it's not necessary - can be removed.

jonnyshaw89 (approved):
This all looks good to me.

Dockerfile
```diff
@@ -0,0 +1,25 @@
+FROM ubuntu:15.04
+
+ENV RUNTIME_PACKAGES="python3 postgresql python-psycopg2 git"
+ENV BUILD_PACKAGES="curl build-essential python3-dev libpq-dev ca-certificates libffi-dev libssl-dev"
+
+WORKDIR /code
+
+EXPOSE 5000
```
jonnyshaw89, line 8: I added expose here

iwootten: Cool, guess that's for if you decide to run standalone from compose?

jonnyshaw89: Yeh, it makes it clear what ports the container can expose even if your not exposing them

scripts/run_app.sh
```diff
@@ -44,11 +51,11 @@ if [ ! -s "app/static" ]; then
   npm run compile
 fi
 
-url="`python token_generator.py`"
+url="`run_python token_generator.py`"
 
-if [ -z "${TRAVIS}" ]; then
+if [ -z "${TRAVIS}"] || [-z "${DOCKER}"]; then
   open_url $url &
```
jonnyshaw89, line 49: You need to move this command into the docker conditional. 
As it tries to launch chrome in the container.

iwootten: Can't we just make the "||" here a "&&"?

iwootten: Additionally, I guess the "DOCKER" env var should be declared in the dockerfile and not the compose one.

jonnyshaw89: yeh or that

benjefferies: Or we could abandon the use of the run_app.sh script when using docker and change the entrypoint to python3 application.py runserver?

jonnyshaw89:
@benjefferies You happy to merge this?

benjefferies (approved):

PR #475, opened 2016-10-05
eq-364 Other option for radio/checkboxes

hamishtaplin:

[Trello card](https://trello.com/c/hNeyHf3v/642-5-respondent-radio-other-option-generates-a-free-text)

Adds an "other" option to radio buttons/checkboxes. This is defined in the schema as:

``` json
"options": [
    {
        "label": "Deep pan",
        "value": "deep pan"
    },
    {
        "label": "Thin",
        "value": "thin"
    },
    {
        "label": "Other",
        "value": "other",
        "other": {
          "label": "Please specify other"
        }
    }
],
```

This will generate the following pattern:

screenshot 2016-10-11 08 38 18
### Additional changes
- Implemented new styling for radios/checkboxes with a minimum-width and a border around the selectable area.
- Re-factored `checkbox_group_widget.py` and `radio_group_widget.py` into a single `checkbox_radio_group_widget.py` as they were functionally the same.
### How to review
1. Start '0_checkbox_other' survey.
2. Confirm extra option works on both questions/pages.
3. Enter a value in "other".
4. Invoke a validation error on that page and check that the "other" value is retained.
5. Complete survey and check that "other" value is handled by the summary page.


bennpearson (approved):

app/schema/answer.py
```diff
@@ -38,6 +38,15 @@ def get_state_class(self):
 
     def get_user_input(self, post_vars):
         user_input = self.widget.get_user_input(post_vars)
+        return self._input_or_none(user_input)
+
+    def get_other_value(self, post_vars):
```
LJBabbage, line 43: Would prefer a better name here or some documentation, get_other_value doesn't really mean anything if I didn't know the story

app/schema/answer.py
```diff
@@ -38,6 +38,15 @@ def get_state_class(self):
...
+        if self.type == 'Radio' or self.type == 'Checkbox':
+            user_input = self.widget.get_other_input(post_vars)
+            return self._input_or_none(user_input)
+
+    @staticmethod
+    def _input_or_none(user_input):
```
LJBabbage: Same as above

LJBabbage:
We perhaps need to use more relevant survey examples

app/schema/answer.py
```diff
@@ -70,6 +79,11 @@ def validate(self, state):
                 state.errors = []
                 state.errors.append(self.questionnaire.get_error_message('MANDATORY', self.id))
                 state.is_valid = False
+            elif self.mandatory and self.type == 'Radio' and state.value == 'other' and state.input == 'other' and state.other is None:
```
LJBabbage: Would prefer something like, easier to read and not repetitive

elif self.mandatory and state.input is None:
                self.mandatory_error(state)
            elif self.mandatory and self.type == 'Radio' and state.input == 'other' and state.other is None :
                self.mandatory_error(state)

 def mandatory_error(self, state):
        state.errors = []
        state.errors.append(self.questionnaire.get_error_message('MANDATORY', self.id))
        state.is_valid = False

app/schema/answers/checkbox_answer.py
```diff
@@ -31,6 +31,10 @@ def validate(self, state):
                 state.errors = []
                 state.errors.append(self.questionnaire.get_error_message('MANDATORY', self.id))
                 state.is_valid = False
+            elif self.mandatory and 'other' in state.input and not state.other:
```
LJBabbage, line 32: Same as above

app/schema/widget.py
```diff
@@ -15,3 +15,12 @@ def get_user_input(self, post_vars):
         user_input = post_vars.get(self.name, None)
         logger.debug('Getting user input for "{}", value is "{}"'.format(self.name, user_input))
         return user_input
+
+    def get_other_input(self, post_vars):
+        if not hasattr(post_vars, 'getlist'):
+            return None
+        list_of_values = post_vars.getlist(self.name)
+        logger.debug('There are {} entries in the list: {}'.format(len(list_of_values), list_of_values))
```
LJBabbage: Not sure knowing how many entries in a list would add much when debugging

app/questionnaire/questionnaire_manager.py
```diff
@@ -206,7 +206,16 @@ def get_answers(self):
             node = node.next
 
         for answer in answers:
-            answers_dict[answer.id] = answer.value
+            if not answer.other:
+                answers_dict[answer.id] = answer.value
+            elif answer.other and isinstance(answer.value, list):
+                list_without_other = answer.value[:]
+                if 'other' in list_without_other:
```
LJBabbage: Looks odd, searching list_without_other the word 'other', feels like you would never find it, although I know you can

app/questionnaire_state/answer.py
```diff
@@ -23,6 +24,7 @@ def update_state(self, user_input):
         if self.input:
             try:
                 self.value = self.schema_item.get_typed_value(user_input)
+                self.other = self.schema_item.get_other_value(user_input)
```
LJBabbage, line 27: Can this fall over?, if not it doesn't need to be in the try.

app/schema/answer.py
samiwel: Added some documentation around the get_other_value function.

app/schema/answer.py
LJBabbage: And only for Radio?

app/schema/widgets/checkbox_group_widget.py
```diff
@@ -43,6 +26,8 @@ def _build_options(self, answer_schema, answer_state):
                     'value': option['value'],
                     'label': option['label'],
                     'selected': option_selected,
+                    'other': option['other'] if 'other' in option else None,
+                    'othervalue': answer_state.other if hasattr(answer_state, 'other') else None,
```
LJBabbage: other_value instead of othervalue?

app/schema/widgets/radio_group_widget.py
```diff
@@ -36,6 +16,8 @@ def _build_options(self, answer_schema, answer_state):
                     'value': option['value'],
                     'label': option['label'],
                     'selected': option['value'] == answer_state.input,
+                    'other': option['other'] if 'other' in option else None,
+                    'othervalue': answer_state.other if hasattr(answer_state, 'other') else None,
```
LJBabbage: As above

app/settings.py
```diff
@@ -50,7 +50,7 @@ def get_key(key_name):
 EQ_CLOUDWATCH_LOGGING = parse_mode(os.getenv("EQ_CLOUDWATCH_LOGGING", 'True'))
 EQ_SCHEMA_DIRECTORY = os.getenv('EQ_SCHEMA_DIRECTORY', 'app/data')
 EQ_SESSION_TIMEOUT = int(os.getenv('EQ_SESSION_TIMEOUT', '28800'))
-EQ_SECRET_KEY = os.getenv('EQ_SECRET_KEY', os.urandom(24))
+EQ_SECRET_KEY = os.getenv('EQ_SECRET_KEY', 'test.test.test.test')
```
LJBabbage: Contentious

tests/integration/star_wars/star_wars_tests.py
```diff
@@ -154,7 +165,7 @@ def check_repeating_answer(self, repeating_answer_page, question_id):
         self.assertRegexpMatches(content, question_id )
 
 
-    def check_choose_your_side(self, start_page):
+    def mem(self, start_page):
```
LJBabbage: Not sure what is going on here

app/schema/answer.py
samiwel: This validation is only applicable to Radio at this point since there is a separate validation routine for a checkbox in the checkbox sub-class.

app/questionnaire/questionnaire_manager.py
samiwel: Renamed list to answer_values for clarity.

tests/integration/star_wars/star_wars_tests.py
samiwel: Think I must have unintentionally renamed this method, reverted the change.

PR #484, opened 2016-10-10
Metadata refactor

iwootten:

Work necessary in order to support saving metadata as a dict, rather than pickling objects.

Return a dict of keys from MetadataParser instead of an object with attributes. Additionally, remove the use of MetadataConstants and replace its use throughout with variable names. This makes the majority of code dependent on the constants far easier to read. 

Additionally, move date conversion responsibility to preprocessors, allowing for simple serialisation instead of pickling using json encode/decode.

This work was dependent on my recent auth/questionnaire store work and would have further complicated that work, so I've opened up this PR for it to keep things separate.
### How to test

Run the python tests.


iwootten:
Hmm, this patch coverage from codecov is a little misleading - the coverage is the same as before (on the files changed), with 75.51% of files covered. We probably should make that be set to greater than or equal to the patch files previous coverage, rather than the same as the rest of the project.

app/parser/metadata_parser.py
```diff
@@ -23,67 +25,100 @@ def id_generator():
...
+            if field.mandatory and key not in token:
+                return False, key
+        return True, ""
+
+
+def serialise_metadata(metadata):
+    # Strip all 'none' values
+    metadata = {k: v for k, v in metadata.items() if v is not None}
+    serialised_metadata = metadata
+    date_fields = [(k, metadata[k]) for k, v in metadata_fields.items() if v.parser == iso_8601_data_parser and k in metadata]
```
collisdigital: You could add a serialise and deserialise method to the MetadatField class to keep the logic for each field self-contained (with a default that just returns the value unchanged); means new fields/types can be added without changing this code/logic or associates tests and removes field specific logic from here leaving this as just a simple for-loop calling serialise on each field and the opposite for deserialise below (possibly boils down to be a single line list comprehension then)

iwootten: Not with the current implementation - the metadata parse method returns a dict, thus what is finally passed to the final encode_metadata is a datetime object or string for each key, rather than a metadatafield which could be serialised in the way you describe. 

The flat dict is more desirable than the previous object and I think the metadatafield is really trying to do multiple jobs of validation/generation where we should probably be using a python module for the validation in this and other areas of the submitter. 

The commit above was really an interim step in making this class a collection of helper methods.

collisdigital: I meant something like this:

`[v.serialise(metadata[k]) for k, v in metadata_fields.items() if k in metadata]`

NB: totally made up code probably not valid.

But completely see what you're saying; if the responsibilities are going to be split further it doesn't make sense to do this.

app/authentication/authenticator.py
```diff
@@ -80,6 +80,6 @@ def _jwt_decrypt(self, request):
         return token
 
     def _check_user_data(self, token):
-        valid, reason = MetadataParser.is_valid(token)
+        valid, reason = is_valid_metadata(token)
```
ajmaddaford: The missing metadata field (if any) is returned, not a reason

app/parser/metadata_parser.py
```diff
@@ -23,77 +19,69 @@ def id_generator():
...
+            parsed[key] = attr_value
+    except (RuntimeError, ValueError, TypeError) as e:
+        logger.error("parse_metadata: Unable to parse")
+        logger.exception(e)
+        raise InvalidTokenException("Incorrect data in token")
+
+    return parsed
+
+
+def is_valid_metadata(metadata):
```
ajmaddaford, line 84: This method only checks mandatory fields, and a subsequent call to parse_metadata could fail if one of the validators fails. I can see this is how it worked before, so I'd be happy for it to be looked at outside of this pull request.

iwootten: Yeah, I wanted to rename this method (has_mandatory_metadata_fields?), but couldn't think of anything concise. It makes sense that all the validation is done in this method, rather than parse, but it'll touch lots of unit tests.

app/data_model/questionnaire_store.py
```diff
@@ -35,14 +33,14 @@ def save(self):
...
-    def encode_metadata(self, metadata):
-        self.data["METADATA"] = jsonpickle.encode(metadata)
-
-    def decode_metadata(self):
+    @property
+    def metadata(self):
         if "METADATA" in self.data:
-            return jsonpickle.decode(self.data["METADATA"])
-        else:
-            raise RuntimeError("No metadata for user %s", self.user_id)
```
collisdigital, line 45: If this is removed; is the same error/message/check handled somewhere else (e.g. do all callers check for None as a return value ~ they may not have before etc. ); I.e if this occurs will it be easily identified in the log/stack trace.

app/data_model/questionnaire_store.py
```diff
@@ -35,14 +33,14 @@ def save(self):
         logger.debug("Saving user data %s for user id %s", self.data, self.user_id)
         self.storage.store(data=self.data, user_id=self.user_id, user_ik=self.user_ik)
 
-    def encode_metadata(self, metadata):
-        self.data["METADATA"] = jsonpickle.encode(metadata)
-
-    def decode_metadata(self):
+    @property
+    def metadata(self):
         if "METADATA" in self.data:
```
collisdigital, line 38: PEP8 has explicit return values, even if None: "explicit return statement should be present at the end of the function (if reachable)."; up to you and team to agree if this is something you want to stick to or not.

app/parser/metadata_parser.py
```diff
@@ -23,77 +19,69 @@ def id_generator():
...
-                        logger.debug("No value provide for %s but this is not mandatory, setting to None", attr_name)
-                        attr_value = None
-                setattr(metadata, attr_name, attr_value)
-            return metadata
-        except (RuntimeError, ValueError, TypeError) as e:
-            logger.error("Unable to parse Metadata")
-            logger.exception(e)
-            raise InvalidTokenException("Incorrect data in token")
+
+    def validate(self, original_value):
```
collisdigital, line 31: Same comment as before about explicit return values here.

app/parser/metadata_parser.py
```diff
@@ -23,77 +19,69 @@ def id_generator():
...
+        if self.mandatory and original_value is None:
+            raise ValueError("Missing mandatory field value")
+        if self._validator:
+            return self._validator(original_value)
+
+    def generate(self):
```
collisdigital, line 38: Same comment as before about explicit return values here.

collisdigital (commented):
On explicit return values; if you decide to adhere to that I assume there may be a flake8 check that can be enabled to look for when it's missed.

app/data_model/questionnaire_store.py
iwootten: The set method is _only_ accessed once as part of the authenticator call in jwt_login. 

It is called once metadata has been validated through parse_metadata, so couldn't ever be anything than a valid metadata dict, meaning we can't save anything other than valid metadata.

The call to get_metadata is the main accessor of the metadata property on the questionnaire_store and it returns None when hitting an anonymous user. But given we've only ever saved valid metadata, it shouldn't be returning anything else.

app/parser/metadata_parser.py
iwootten: Validate shouldn't be returning anything and simply throwing exceptions when validation fails. I've corrected this.

app/templating/metadata_template_preprocessor.py
```diff
@@ -18,17 +19,19 @@ def build_metadata(self, schema):
...
+            name = metadata["ru_name"]
+            trading_as = metadata["trad_as"]
         else:
             respondent_id = None
             name = None
             trading_as = None
 
         respondent_meta = {
-            "tx_id": convert_tx_id(self._get_metadata().tx_id),
+            "tx_id": convert_tx_id(metadata["tx_id"]),
```
collisdigital, line 34: should this move to inside the 'if metadata' above like the other values (i.e. will probably fall over if metadata is None)?

iwootten: Possibly, though it raises further issues if we do so - should we generate a tx_id when metadata is none? Going to leave as is for now as it's way better than pulling the data from the store repeatedly.

collisdigital: Originally we said we should generate a tx_id if none is supplied, but this was mostly whilst the other system (RRM Classic) was in the process of adding it the JWT and in this case you'd still expect there to be valid metadata but with a missing tx_id (this is already handled somewhere else ~ generator in the MetadataField). The areas that tx_id are used handle it being None I believe and just don't include it in the page (e.g. Error pages)

ajmaddaford: This is no different to the way it worked previously - tx_id will always be set in the metadata. If metadata doesn't exist a RuntimeError is raised.

ajmaddaford (approved):
LGTM 👍

PR #485, opened 2016-10-11
Enable Selenium tests on Travis

hamishtaplin:

Enables selenium tests via webdriver to be run Sauce Labs on Travis. 
### How to review

Confirms Travis is running the tests and that they all pass


.travis.yml
```diff
@@ -16,21 +16,22 @@ install:
...
 before_script:
+- npm run compile
 - "./scripts/run_app.sh"
 script:
 - "./scripts/run_tests.sh"
 # fail travis if under X% test coverage
 - coverage report --fail-under=77
 notifications:
--  slack: digitaleq:LDxJXRGeVmg3PpdUhp90eY3Z
+- slack: digitaleq:LDxJXRGeVmg3PpdUhp90eY3Z
```
benjefferies, line 26: Are the notifications working? travis docs suggests that it should notify every time https://docs.travis-ci.com/user/notifications/ but I haven't got a notification or I am not subscribed to the correct channel?

gulp/tests.js
```diff
@@ -4,26 +4,10 @@ import path from 'path'
...
 import {paths} from './paths'
 import webdriver from 'gulp-webdriver'
-import selenium from 'selenium-standalone'
 import yargs from 'yargs'
 
 const KarmaServer = karma.Server
 
-export let seleniumServer
-
-export function startSeleniumServer(done) {
```
benjefferies, line 14: Is this going to cause issues running the functional tests locally? I am getting ERROR: Couldn't connect to selenium server when running the tests locally.

benjefferies: Ignore this as I have done a `npm install` and I no longer see this issue

iwootten:
Trying to run this locally all I get is "An element could not be located on the page using the given search parameters." after a npm install. Chrome fires up with "data;" in the address bar in each scenario.

ajmaddaford:
@iwootten you need to run npm install, then run the app (./scripts/run_app.sh) and then run the tests (npm run test_functional).

ajmaddaford:
Looking at the Travis output the selenium tests took 5.27 min, as we increase the number of tests I think this will add too much time to the build. It may be worth thinking about having a subset of selenium tests run on every build and have an overnight job to run through all tests.

.travis.yml
hamishtaplin: @benjefferies I think we disabled the notifications in Slack itself.

gulp/tests.js
hamishtaplin: @benjefferies Yeah this code was removed as newer versions of webdriver include the Selenium server so you don't have to spin one up manually anymore.

hamishtaplin:
@ajmaddaford What would the subset of tests be?

ajmaddaford:
@hamishtaplin until we get a more complete set of feature tests, possibly just mci and rsi-saverestore-0112?

hamishtaplin:
@ajmaddaford Done.

tests/functional/config.js
```diff
@@ -1,82 +1,57 @@
...
 }
 
 if (process.env.TRAVIS === 'true') {
   config = {
     ...config,
     ...sauceLabsConfig,
-    logLevel: 'debug'
+    logLevel: 'debug',
+    capabilities: [firefox],
+    specs: ['mci', 'rsi-saverestore-0112'].map(spec => `${paths.test.wdioSpec}/${spec}.js`)
```
iwootten: This fails for me:

pattern tests/functional/spec/mci.js did not match any file
pattern tests/functional/spec/rsi-saverestore-0112.js did not match any file

Do you need to specify the ".spec"?

hamishtaplin: @iwootten Oops. Done.

ajmaddaford (approved):
LGTM 👍

PR #489, opened 2016-10-13
Add git to Docker

iwootten:

Putting git back in dockerfile due to npm needing it during build (spotted by @samiwel) 
### How to review

Run 

   docker-compose build --no-cache 

To confirm it builds as expected


samiwel (approved):
Tested the change on the branch, git is installed and so the npm install step works as expected.

PR #490, opened 2016-10-13
Eq 651 remove db abstractions

iwootten:

Remove unnecessary storage classes and rename appropriately. Move to using DBStorage for tests previously extending InMemoryStorage. Additionally, make the StorageMedium use an abstract base class so classes extending it need to implement abstract methods.
### How to review

Confirm tests pass, check if you agree with the naming.


app/storage/encrypted_storage.py
```diff
@@ -30,7 +28,7 @@ def store(self, data, user_id, user_ik):
     def get(self, user_id, user_ik):
```
benjefferies, line 36: Pycharm hinted to a potential issue here, user_ik is define as a optional argument in storage medium but it's not here. That makes it firstly confusing and secondly it can potentially break if someone calls it assuming optional.

Given that it makes sense to make it mandatory given the only place it's used defines the argument

iwootten: Yes, seems daft, will update.

app/storage/storage_medium.py
```diff
@@ -0,0 +1,29 @@
+import logging
+from abc import ABCMeta, abstractmethod
+
+logger = logging.getLogger(__name__)
+
+
+class StorageMedium(metaclass=ABCMeta):
```
ajmaddaford: Why not AbstractStorage?

iwootten: Agree, but as per our offline conversation, think that this would be better off moving to removing this abstract model and making encrypted storage a child of database storage. Given these two classes don't actually share the same method specification.

app/storage/encrypted_storage.py
```diff
@@ -5,80 +5,58 @@
...
-from app.storage.abstract_server_storage import AbstractServerStorage
+from app.storage.database_storage import DatabaseStorage
 from app.utilities.strings import to_bytes
 from app.utilities.strings import to_str
 
 logger = logging.getLogger(__name__)
 
 
-class EncryptedServerStorageDecorator(AbstractServerStorage):
+def _safe_logging(msg, param):
```
benjefferies: Shall we get rid of the logging if it's not used in production. It only adds risk to logging if misconfigured. In theory we can run without encryption locally

iwootten: Agreed, we can always put it back if it's an issue.

app/storage/encrypted_storage.py
```diff
@@ -5,80 +5,48 @@
...
+    sha256.update(to_str(pepper).encode('utf-8'))
 
-    JSON_DATA_KEY = 'data'
+    # we only need the first 32 characters for the CEK
+    cek = sha256.hexdigest()[:32]
+    return to_bytes(cek)
 
-    def __init__(self, server_storage):
+
+class EncryptedStorage(DatabaseStorage):
```
benjefferies, line 26: By removing the store and get API contract I don't get why we have the get_storage factory method. If the method signatures differ then surely the encrypted storage toggle off function won't work? It was in this state before we started work on it so has it ever worked without encryption?

Going further than this if there isn't a IS-A relationship do we disadvantage ourselves by tightly coupling the two classes and composition would be favoured here?

benjefferies: I guess what I am saying is if I am right in thinking it doesn't work then why not delete the non encrypted data until there is a need for it

benjefferies: Stacktrace from trying to run with Q_SERVER_SIDE_STORAGE_ENCRYPTION=False

```
[2016-10-17 15:38:23,558] ERROR [app.main.errors.log_exception:61] store() got an unexpected keyword argument 'user_ik'
[2016-10-17 15:38:23,558] ERROR [app.main.errors.log_exception:62] store() got an unexpected keyword argument 'user_ik'
  File "/Users/benjefferies/source/github/eq-survey-runner/app/main/views/root.py", line 56, in login
    authenticator.jwt_login(request)
  File "/Users/benjefferies/source/github/eq-survey-runner/app/authentication/authenticator.py", line 72, in jwt_login
    questionnaire_store.save()
  File "/Users/benjefferies/source/github/eq-survey-runner/app/data_model/questionnaire_store.py", line 36, in save
    self.storage.store(data=self.data, user_id=self.user_id, user_ik=self.user_ik)
TypeError: store() got an unexpected keyword argument 'user_ik'
```

iwootten: Had a look at master and doesn't appear to trip up there, presumably because the abstract class wasn't using a proper base class and the signatures didn't appear as different. Will have to have a think on what the best approach here is.

iwootten: As discussed, think I'm going to go for overloading the db storage methods to allow being called with an user_ik. That way we fix the issue described and calling classes can either not care what the storage type is, or invoke the storage type with known params directly. Maybe at a later date we can look at moving the choice of storage medium into the models.

benjefferies:
Looks good to me

benjefferies (approved):

PR #491, opened 2016-10-13
added json schema validation as a test to validate all files in app/data

jonnyshaw89:

Have added a test that loads all of the json files in app/data and validates them against the schema
### How to review

Check that the test is validating the json files


app/schema_loader/schema_loader.py
```diff
@@ -72,3 +74,10 @@ def available_s3_schemas(files):
             files.append(key.key)
     except botocore.exceptions.ClientError as e:
         logging.error("S3 error: %s", e.response['Error']['Code'])
+
+
+def validate_json_against_schema(json_to_validate):
```
ajmaddaford: Is there a reason this method is here rather than in the test?

tests/app/data/test_schema_validation.py
```diff
@@ -0,0 +1,38 @@
...
+      errors = []
+
+      for file in schema_loader.available_schemas():
+
+        try:
+          filename = os.path.splitext(file)[0]
+
+          eq_id = filename.split("_")[0]
+
+          json_schema = schema_loader.load_schema(eq_id, filename[len(eq_id)+1:])
```
ajmaddaford: I think it would be clearer if we loaded the schema file directly here, rather than extract the eq_id and form_type and use the load_schema method.

ajmaddaford (changes requested):

app/schema_loader/schema_loader.py
jonnyshaw89: I put it in here as i thought it would be useful for when we validate the schemas on load. But i guess as its not being used now ill move it into the tests.

tests/app/data/test_schema_validation.py
```diff
@@ -0,0 +1,47 @@
...
+logger = logging.getLogger(__name__)
+
+
+class TestSchemaValidation(unittest.TestCase):
+
+    def test_schema(self):
+
+      errors = []
+
+      for file in schema_loader.available_schemas():
```
ajmaddaford: It might be better to just read the json files directly from the data directory here - I got an error trying to read an s3 schema from the data dir:
`FileNotFoundError: [Errno 2] No such file or directory: 'app/data/1.json'`

tests/app/data/test_schema_validation.py
```diff
@@ -0,0 +1,47 @@
...
+
+        try:
+
+          schema_file = open(os.path.join(settings.EQ_SCHEMA_DIRECTORY, file), encoding="utf8")
+
+          self.validate_json_against_schema(json.load(schema_file))
+
+        except jsonschema.exceptions.ValidationError as e:
```
ajmaddaford: The try and exception here should be around the `jsonschema.validate` method below.

tests/app/data/test_schema_validation.py
```diff
@@ -0,0 +1,47 @@
...
+
+      if errors:
+        for error in errors:
+          logger.error(error)
+
+        self.fail("{} Schema Validation Errors.".format(len(errors)))
+
+    def validate_json_against_schema(self, json_to_validate):
+        schema_file = open(os.path.join(settings.EQ_SCHEMA_DIRECTORY, "schema/schema-v1.json"), encoding="utf8")
+        schema = json.load(schema_file)
```
ajmaddaford: We should load the schema once, rather than for every survey JSON we validate.

tests/app/data/test_schema_validation.py
```diff
@@ -0,0 +1,47 @@
...
+
+        validate(json_to_validate, schema)
```
ajmaddaford: When the exception handling is moved, this method could return the error.

ajmaddaford (changes requested):

tests/app/data/test_schema_validation.py
benjefferies: Looks fixed to me

tests/app/data/test_schema_validation.py
benjefferies: Looks fixed to me

tests/app/data/test_schema_validation.py
benjefferies: Looks fixed to me

tests/app/data/test_schema_validation.py
benjefferies: Looks fixed to me

benjefferies (approved):

iwootten:
Do you think you could cherry pick your work and rebase this against master? Most of these commits just seem to be merges.

ajmaddaford (approved):

PR #497, opened 2016-10-17
Split out end points

benjefferies:

- Split questionnaire submit and get requests
- Pull out thank-you page
- Pull out previous
- Remove first end point
  - first end point only ever appeared to be used in integration test
- Delete unused methods
### How to review

- [x]  Run tests


app/main/views/questionnaire.py
```diff
@@ -22,79 +18,91 @@
...
-        metadata = get_metadata(current_user)
-        current_survey = eq_id + form_type + period_id + collection_id
-        metadata_survey = metadata.eq_id + metadata.form_type + metadata.period_id + metadata.collection_exercise_sid
 
-        if not current_survey == metadata_survey:
-            return redirect("/information/multiple-surveys")
+@main_blueprint.route('/questionnaire/<eq_id>/<form_type>/<period_id>/<collection_id>/<location>', methods=["POST"])
+@login_required
+def post_questionnaire(eq_id, form_type, period_id, collection_id, location):
+    logger.debug("Post location : /questionnaire/%s/%s/%s/%s/%s", eq_id, form_type, period_id, collection_id, location)
```
iwootten: Isn't flask (werkzeug) logging the url visited by default?

benjefferies: From the look of my logs yes. Will remove them.

app/main/views/questionnaire.py
```diff
@@ -22,79 +18,90 @@
...
-        return page_not_found(e)
-    except InvalidLocationException as e:
-        return do_redirect(eq_id, form_type, period_id, collection_id, questionnaire_manager.get_current_location())
-    except SubmissionFailedException as e:
-        # Rabbit MQ connection issue
-        return service_unavailable(e)
-    except Exception as e:
-        return internal_server_error(e)
+    page = render_page('thank-you', questionnaire_manager)
+    questionnaire_manager.delete_user_data()
```
iwootten: I feel that it may be good to move the questionnaire manager behaviour behind delete_user_data into this main route? It's only ever called here.

benjefferies: Certainly, didn't catch that it was only used here.

app/main/views/questionnaire.py
```diff
@@ -79,10 +80,15 @@ def get_thank_you(eq_id, form_type, period_id, collection_id):
...
 
     page = render_page('thank-you', questionnaire_manager)
-    questionnaire_manager.delete_user_data()
+    delete_user_data()
     return page
 
 
+def delete_user_data():
+    get_questionnaire_store(current_user.user_id, current_user.user_ik).delete()
+    session_manager.clear()
```
iwootten, line 96: Think you've missed a line here- need to call delete on the questionnaire store once it's retrieved.

iwootten: Oh I see!

iwootten:
I think this change highlights how much questionnaire_manager is trying to do - for me I feel that some of that possibly should be tackled as part of this work?

benjefferies:
I am happy with this to be merged given the merge conflicts are resolved

app/questionnaire/questionnaire_manager.py
```diff
@@ -345,29 +324,32 @@ def process_incoming_answers(self, location, post_data, replay=False):
             raise InvalidLocationException()
 
     def get_rendering_context(self, location):
```
iwootten: If we're going to be splitting these into two methods, it probably makes sense that if the location is unknown within a call to get_rendering_context it should be capable of handling a location which is 'summary'. Maybe it should therefore check on the first line and return get_summary_rendering_context() if so?

benjefferies: done

tests/integration/mci/test_change_answers.py
```diff
@@ -77,8 +77,3 @@ def happy_path(self, form_type_id, eq_id):
         self.assertRegexpMatches(content, '>Your responses<')
         self.assertRegexpMatches(content, 'Please check carefully before submission')
         self.assertRegexpMatches(content, '>Submit answers<')
-
```
benjefferies, line 80: The first end point is only ever used in integration tests so the end point and related tests have been removed

tests/app/views/test_health_check.py
```diff
@@ -2,9 +2,3 @@ def test_healthcheck(survey_runner):
     response = survey_runner.test_client().get('/healthcheck')
     assert response.status_code == 200
     assert "success" in response.get_data(as_text=True)
-
-
-def test_errors_404(survey_runner):
```
benjefferies, line 7: This exact test live in test_errors.py

app/main/views/questionnaire.py
```diff
@@ -1,103 +1,148 @@
...
+def go_to_page(collection_id, eq_id, form_type, period_id, location, is_valid=True):
+    questionnaire_manager = QuestionnaireManagerFactory.get_instance()
+    try:
+        questionnaire_manager.go_to(location)
+    except InvalidLocationException:
+        return do_redirect(eq_id, form_type, period_id, collection_id, questionnaire_manager.get_current_location())
+    return get_page(questionnaire_manager, location, is_valid)
+
+
+def get_page(questionnaire_manager, location, is_valid):
```
ajmaddaford: get_page is confusing, I think render_page would be clearer

app/main/views/questionnaire.py
```diff
@@ -1,103 +1,148 @@
...
-    next_location = questionnaire_manager.get_current_location()
-
-    if is_valid:
-        metadata = get_metadata(current_user)
-        logger.info("Redirecting user to next location %s with tx_id=%s", next_location, metadata["tx_id"])
-        return do_redirect(eq_id, form_type, period_id, collection_id, next_location)
-    else:
-        return do_get(questionnaire_manager, next_location, is_valid)
+def get_rendering_template(location):
+    return TemplateRegistry.get_template_name(location)
```
ajmaddaford: Any reason not to use this directly rather than via the `get_rendering_template` method?

app/main/views/questionnaire.py
```diff
@@ -1,103 +1,148 @@
...
+        elif not same_survey(kwargs['eq_id'], kwargs['form_type'], kwargs['period_id'], kwargs['collection_id']):
+            return redirect("/information/multiple-surveys")
+        else:
+            return func(*args, **kwargs)
 
-        # If the metadata doesn't match the current survey, multiple surveys must be open, which we don't support
-        metadata = get_metadata(current_user)
-        current_survey = eq_id + form_type + period_id + collection_id
-        metadata_survey = metadata["eq_id"] + metadata["form_type"] + metadata["period_id"] + metadata["collection_exercise_sid"]
+    return decorated_function
```
ajmaddaford: I think we should re-implement this using the `url_value_preprocessor` blueprint decorator function (see http://exploreflask.com/en/latest/blueprints.html#using-a-dynamic-url-prefix). We should also set the questionnaire_manager on the global application (request) context so that the routes that use this function don't need to re-instantiate the manager.

app/main/views/questionnaire.py
```diff
@@ -1,103 +1,148 @@
...
+    return go_to_page(collection_id, eq_id, form_type, period_id, 'summary')
+
+
+def delete_user_data():
+    get_questionnaire_store(current_user.user_id, current_user.user_ik).delete()
+    session_manager.clear()
 
 
 def do_redirect(eq_id, form_type, period_id, collection_id,  location):
     return redirect('/questionnaire/' + eq_id + '/' + form_type + '/' + period_id + '/' + collection_id + '/' + location)
```
ajmaddaford: We should be using the Flask url_for method for generating URLs for known routes. Also, is there a way to get the current route rather than keep having to pass all of the parameters to this method?

app/main/views/questionnaire.py
```diff
@@ -1,103 +1,148 @@
...
-def do_get(questionnaire_manager, location, is_valid=True):
-    questionnaire_manager.go_to(location)
+def same_survey(eq_id, form_type, period_id, collection_id):
+    metadata = get_metadata(current_user)
+    current_survey = eq_id + form_type + period_id + collection_id
+    metadata_survey = metadata["eq_id"] + metadata["form_type"] + metadata["period_id"] + metadata["collection_exercise_sid"]
+    return current_survey == metadata_survey
+
+
+def go_to_page(collection_id, eq_id, form_type, period_id, location, is_valid=True):
```
ajmaddaford: This method go's to a page **and** renders it. I think we should either rename the method or move the rendering out to be called independently.

app/main/views/errors.py
```diff
@@ -13,47 +17,39 @@
 
 logger = logging.getLogger(__name__)
 
+errors_blueprint = Blueprint('errors', __name__)
```
ajmaddaford, line 25: Should the errors blueprint be setting the cache headers like the root and questionnaire blueprints?

ajmaddaford (changes requested):

app/main/views/errors.py
benjefferies: What is the reason we are indicating to the browser not to cache? I can add it just to blanket cover this too anyway

app/main/views/questionnaire.py

app/main/views/questionnaire.py

app/main/views/questionnaire.py
benjefferies: Nope, copy and paste initially and then forgot to remove

app/main/views/questionnaire.py
benjefferies: Good call

app/main/views/questionnaire.py
benjefferies: Not a great fan of methods with `and` suggests it's doing too much which could be right in this case so I've pulled out

benjefferies:
Superseded by https://github.com/ONSdigital/eq-survey-runner/pull/512

PR #498, opened 2016-10-17
Remove Schema and State from linked list 

LJBabbage:

This PR removes the Schema and State from the linked list. Instead the node holds the id of the block and the answers associated with that block. When there is a need for the State (validation, or rendering) it finds it in the parsed version of the schema and generates the state on the fly, with any previous answers. Json pickle now only saves the linked list

Validation is now in one place, as it used to be under validation (Mandatory) and state update (Type Checking). We no longer redirect on failed validation, instead just recalling the renderer, meaning we don't have to reload the state again.

I have renamed a number of classes and functions as well to make the code more readable and easy to follow

**What has been removed**
All repeating elements and their tests
Other option from checkboxes and radio button, due to the use of state in linked list

**Behaviour change**
Error messages do not appear on a get request
### How to review

Full system testing


app/main/views/questionnaire.py
```diff
@@ -92,9 +92,12 @@ def do_post(collection_id, eq_id, form_type, period_id, location, questionnaire_
...
+    is_valid = questionnaire_manager.process_incoming_answers(location, request.form)
     next_location = questionnaire_manager.get_current_location()
-    metadata = get_metadata(current_user)
-    logger.info("Redirecting user to next location %s with tx_id=%s", next_location, metadata.tx_id)
-    return redirect('/questionnaire/' + eq_id + '/' + form_type + '/' + period_id + '/' + collection_id + '/' + next_location)
+
+    if is_valid:
+        metadata = get_metadata(current_user)
+        logger.info("Redirecting user to next location %s with tx_id=%s", next_location, metadata.tx_id)
+        return redirect('/questionnaire/' + eq_id + '/' + form_type + '/' + period_id + '/' + collection_id + '/' + next_location)
```
benjefferies: do_redirect could be made use of here

app/questionnaire/questionnaire_manager.py
```diff
@@ -184,78 +155,69 @@ def get_first_location(self):
...
-            answers.extend(node_answers)
+            answers_dict.update(node.answers)
             node = node.next
-
-        for answer in answers:
-            if not answer.other:
-                answers_dict[answer.id] = answer.value
-            elif answer.other and isinstance(answer.value, list):
-                answer_values = answer.value[:]
-                if 'other' in answer_values:
```
benjefferies, line 213: Are we losing the 'other' functionality @samiwel introduced here

app/questionnaire/questionnaire_manager.py
```diff
@@ -184,78 +155,69 @@ def get_first_location(self):
...
-                if not valid:
-                    current_location = node.item_id
-                    logger.debug("Failed validation with current location %s", current_location)
+            self.build_state(node, node.answers)
+            if self.schema_item:
+                self._conditional_display(self.state)
+                is_valid = self.schema_item.validate(self.state)
+                if not is_valid:
+                    location = node.item_id
+                    logger.error("Failed validation with current location %s", location)
```
benjefferies: What was the reason to up the level of logging for a validation error from debug to error? Doesn't seem necessary to me.

app/questionnaire/questionnaire_manager.py
```diff
@@ -328,62 +284,71 @@ def process_incoming_answers(self, location, post_data, replay=False):
...
-
-    def _plumbing_preprocessing(self, node):
-        '''
-        Run the current state through the plumbing preprocessor
-        :return:
-        '''
+        node = self._current
+        state_items = []
+        if is_valid:
+            if location == 'summary':
```
benjefferies, line 311: Would it be cleaner to break this out into get_summary_rendering_context and get_questionnaire_rendering_context then push the conditional logic further up?

tests/app/questionnaire/test_questionnaire_manager.py
```diff
@@ -29,110 +31,20 @@ def setUp(self):
     def test_get_state_is_none(self):
         questionnaire_manager = QuestionnaireManager(self.questionnaire)
         block1 = self.questionnaire.children[0].children[0]
-        self.assertIsNone(questionnaire_manager.get_state(block1.id))
-
-    def test_single_value_answer_with_other_returns_other(self):
```
benjefferies, line 34: Looks like this may have been merged out?

tests/app/questionnaire/test_questionnaire_manager_factory.py
```diff
@@ -14,45 +12,24 @@ class TestQuestionnaireManagerFactory(SurveyRunnerTestCase):
...
-            del state.revision
-            StateManager.get_state = MagicMock(return_value=state)
-            logger_name = 'app.questionnaire.questionnaire_manager_factory'
-            logging.getLogger(logger_name).error = MagicMock()
-            QuestionnaireManagerFactory.get_instance()
-            logging.getLogger(logger_name).error.assert_called_with('Data not recoverable, old version of state without revision attribute')
-
-    def mock_state(self):
-        StateManager.has_state = MagicMock(return_value="True")
+    def mock_user_journey(self):
```
benjefferies: This appears not to be used

tests/app/questionnaire/test_questionnaire_manager_factory.py
```diff
@@ -14,45 +12,24 @@ class TestQuestionnaireManagerFactory(SurveyRunnerTestCase):
...
     def tearDown(self):
         super().tearDown()
-        StateManager.has_state = self.has_state
-        StateManager.get_state = self.get_state
-        StateRecovery.recover_from_post_data = self.recover_from_post_data
+        UserJourneyManager.has_user_journey = self.has_user_journey
+        UserJourneyManager.get_user_journey = self.get_user_journey
         QuestionnaireManagerFactory._get_schema = self.get_schema
 
     def test_get_instance(self):
```
benjefferies: Not sure how useful this test's being now

tests/app/templating/test_summary_template_preprocessor.py
```diff
@@ -55,26 +58,20 @@ def test_build_view_data(self):
...
-        node1 = Node("1", block1)
-        node2 = Node("2", block2)
-
+        node1 = Node("1")
+        node2 = Node("2")
+        state = StateBlock("1", None)
         node1.next = node2
 
-        summary_template_preprocessor = SummaryTemplatePreprocessor()
+        questionnaire_template_preprocessor = SummaryTemplatePreprocessor()
```
benjefferies: Variable has been renamed incorrectly also test class name appears to be wrong?

tests/app/templating/test_summary_template_preprocessor.py
```diff
@@ -55,26 +58,20 @@ def test_build_view_data(self):
...
         original_get_metadata_method = MetaDataTemplatePreprocessor._get_metadata
-        MetaDataTemplatePreprocessor._get_metadata = self.get_metadata
 
-        render_data = summary_template_preprocessor.build_view_data(node1, self.questionnaire)
+        MetaDataTemplatePreprocessor._get_metadata = self.get_metadata
+        state_items = [state]
+        render_data = questionnaire_template_preprocessor.build_view_data(node1, self.questionnaire, state_items)
         self.assertIsNotNone(render_data)
-        blocks = render_data['content']
```
benjefferies, line 69: Should we still have these assertions?

tests/integration/star_wars/test_radio_boxes.py
```diff
@@ -85,34 +81,3 @@ def test_radio_boxes_mandatory_empty(self):
         # Check we are on the next page
         content = resp.get_data(True)
         self.assertRegexpMatches(content, 'Why doesn&#39;t Chewbacca receive a medal at the end of A New Hope?')
-
-    def test_radio_boxes_other_is_mandatory(self):
```
benjefferies, line 89: Looks like this has been merged out

benjefferies (changes requested):

app/questionnaire/questionnaire_manager.py
LJBabbage: I pulled it out as it was using state on the find answer, he is rewriting it now

app/questionnaire/questionnaire_manager.py
LJBabbage: True, left over from me fixing it

tests/app/questionnaire/test_questionnaire_manager.py
LJBabbage: Good spot

LJBabbage:
Definitely some loose ends that need tiding up, the focus wasn't on getting it perfect, just working for people to work of. So I guess a 2 pronged approach, what needs to be done now to get it in and what we have to do but later

app/questionnaire/questionnaire_manager.py
```diff
@@ -328,62 +284,71 @@ def process_incoming_answers(self, location, post_data, replay=False):
...
+        state_items.append(self.state)
+
+        return preprocessor.build_view_data(node, self._schema, state_items)
+
+    def build_state(self, node, answers):
+        # Build the state from the linked list and the answers
+        self.state = None
+        self.schema_item = None
+        if self._schema.item_exists(node.item_id):
+            self.schema_item = self._schema.get_item_by_id(node.item_id)
```
ajmaddaford: I don't think we need to set schema_item on self. Once we've created the state, isn't it accessible via `state.schema_item`?

app/questionnaire/questionnaire_manager.py
```diff
@@ -328,62 +284,71 @@ def process_incoming_answers(self, location, post_data, replay=False):
...
+                self.build_state(node, node.answers)
+
+        if self.schema_item:
+            self._plumbing_preprocessing(self.state)
+            self._conditional_display(self.state)
+        state_items.append(self.state)
+
+        return preprocessor.build_view_data(node, self._schema, state_items)
+
+    def build_state(self, node, answers):
```
ajmaddaford, line 336: We should minimise what we pass to methods. This method only uses the node to get the item_id (more specifically a block_id?), so it would be better to pass just that.

app/questionnaire/user_journey_manager.py
```diff
@@ -0,0 +1,85 @@
+import logging
```
ajmaddaford, line 1: This functionality should be implemented along the same lines as metadata in data_model/questionnaire_store.

app/schema/answers/date_answer.py
```diff
@@ -12,7 +12,8 @@ def __init__(self, answer_id=None):
         self.type_checkers.append(DateTypeCheck())
         self.widget = DateWidget(self.id)
 
-    def get_typed_value(self, post_data):
+    def type_validation(self, post_data):
```
ajmaddaford: I don't think this method should have been renamed - it returns a typed value.

ajmaddaford (changes requested):

app/questionnaire/questionnaire_manager.py
LJBabbage: Yep, I used to use node.answer in this method, I removed the need for that, so you are correct it only uses the id

app/schema/answers/date_answer.py
LJBabbage: True, but its also doing validation, this refers back to the note I made in the code, that validation and type value casting is being done in the same place which isn't right

app/questionnaire_state/state_answer.py
```diff
@@ -22,6 +22,18 @@ def update_state(self, user_input):
...
         self.value = self.schema_item.get_user_input(user_input)
         self.other = self.schema_item.get_other_value(user_input)
+        if self.schema_item.type == 'Radio' and self.input:
+            self._restore_other_value(user_input)
+
+    def _restore_other_value(self, user_input):
+        # Get radio options from the schema.
+        for option in self.schema_item.options:
+            # If the radio answer has an Other option...
+            if option['label'] == 'Other' and 'other' in option and self.input not in [option['value'] for option in self.schema_item.options]:
```
iwootten: This if statement is pretty difficult to read, due to it's length. Do you think you could break the conditions out onto multiple lines @samiwel?

app/schema/widgets/multiple_choice_widget.py
```diff
@@ -26,3 +26,12 @@ def render(self, answer_state):
         }
 
         return render_template('/partials/widgets/multiple_choice_widget.html', **widget_params)
+
+    def get_other_input(self, post_vars):
+        if len(post_vars) == 0:
+            return None
+
+        other_values = post_vars.getlist(self.name) if hasattr(post_vars, 'getlist') else post_vars.get(self.name)
```
iwootten: Similarly @samiwel, these two if conditions would be better if they were over multiple lines.

app/questionnaire_state/state_answer.py
samiwel: @iwootten I'll take a look, thanks.

app/schema/widgets/multiple_choice_widget.py
```diff
@@ -28,10 +28,23 @@ def render(self, answer_state):
...
+        is_immutable_multi_dict = hasattr(post_vars, 'getlist')
+        posted_data = post_vars.getlist(self.name) if is_immutable_multi_dict else post_vars.get(self.name)
+
+        has_multiple_values = isinstance(posted_data, list) and len(posted_data) > 0
+        if has_multiple_values:
+            last_element = posted_data[-1:][0]
+            other_value = last_element
+
         logger.debug('Getting user input for "{}", value is "{}"'.format(self.name, other_value))
         return other_value
```
iwootten: @samiwel You could reduce this down a bit to:

```
return posted_data[-1:][0] if has_multiple_values else None
```

and lose the other_value assignments.

app/questionnaire/questionnaire_manager.py
LJBabbage: Sam has now rewritten this

tests/app/questionnaire/test_questionnaire_manager.py
LJBabbage: Sam has added the extra tests as well

tests/integration/star_wars/test_radio_boxes.py
LJBabbage: As above

app/questionnaire/user_journey_manager.py
LJBabbage: Not tackled as this was only name changes to the class and will be refactored out later anyway, as dicussed

ajmaddaford (approved):

benjefferies (approved):

PR #500, opened 2016-10-18
Issue 499 Improve date parsing

### LINKED ISSUE #499: Plumbing Preprocessor falls over on missing Employment Date
### Expected behaviour

The plumbing preprocessor should fallback gracefully and use "None" for dates which are missing from metadata.
### Actual behaviour

The plumber preprocessor throws a error.

Invoke the preprocessor with metadata containing any missing date.

iwootten:

Improved the way dates are parsed within preprocessors to use shared util method. This fixes #499.
### How to review

I've added additional unit tests to cover the new behaviour so run them.


app/utilities/date_utils.py
```diff
@@ -0,0 +1,18 @@
...
+from datetime import datetime
+
+logger = logging.getLogger(__name__)
+
+
+def to_date(input_date_string, date_format="%Y-%m-%d"):
+    formatted_date = None
+
+    try:
+        # employment date is optional this is why string is checked
```
benjefferies: Comment no longer necessary

benjefferies: Is the try catch necessary now too?

benjefferies (changes requested):

benjefferies (approved):

PR #503, opened 2016-10-19
eq-520 Fieldset/legend

hamishtaplin:

[Trello card](https://trello.com/c/TTUzhhCq/520-2-improve-markup-for-fieldset-and-legend-etc)

This PR adds `fieldset` and `legend` markup to radio button/checkbox questions to improve document semantics and therefore accessibility of the questions/answers.

It also fixes some issues where the validity of HTML was affecting the fieldset\legend being actually parsed correctly by clients.
### How to review
1. Visit Star Wars survey
2. Navigate to first section "Choose a side"
3. Using Voiceover, confirm that the user is informed that they are inside a "group" (`fieldset`) when navigating to the radio button controls. 
4. Confirm that you are able to navigate outside the group (ctrl+option+shift+up) and that it reads the `legend` correctly ("Which side of the force are you on?")


hamishtaplin:
I've added the basic `fieldset` and `legend` to the markup but there's currently no way (that I can see) of wiring it up to the question title to act as the legend.

This also raises the issue of some questions possibly not having a title as seen in this question on Star Wars:

screenshot 2016-10-19 10 30 24

There's only a section title here. This raises the question, do we enforce question titles?

bennpearson (dismissed):

LJBabbage:
As discussed, implement the quick fix for now, either use answer.parent.title or assign the question title at a higher level and use it. We will tackle it properly when we get to the front end render story

bennpearson (approved):

collisdigital:
@yoganandkunche does this still "needs-eq", if not we can remove the label...

yoganandkunche:
@collisdigital As this is already merged I will be testing this in Pre-prod. I have removed the label

yoganandkunche:
Tested this defect in Pre-prod. Voice over for fieldset works as expected. However voice over for legend add extra text "Stop interacting Which side of the force are you on? " instead of "Which side of the force are you on?" discussed with Hamish and raised in scrum. As DAC will be doing testing today and tomorrow we will wait for their feedback.

PR #504, opened 2016-10-19
Docker documentation

jonnyshaw89:

Getting the docker setup in to a place where it can reliably be used for day-to-day development
### How to review

Can you easily get it running?
Can you change code without having to rebuild/re-deploy?

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,49 @@
-node_modules/*
 .git
+.idea
 Dockerfile
+docker-compose.yml
+
+# Installer logs
+pip-log.txt
+
+# Unit test / coverage reports
+.coverage
+.tox
+coverage/
+
+.cache
+
+*.log
+eq.log.*
+
+# Virtualenvs
+env
+env*
+venv
+
+# intellij
+.idea/
+*.ipr
+*.iml
+*.iws
+
+.DS_Store
+
+node_modules/
+app/static/
+tests/results
+tests/coverage
+tests/a11y
+
+app/styles/base/_sprite.scss
+
+## OS X
+._*
+.Spotlight-V100
+.Trashes
+
+# Elastic Beanstalk Files
+.elasticbeanstalk/*
+!.elasticbeanstalk/*.cfg.yml
+!.elasticbeanstalk/*.global.yml
+newrelic.ini
--- a/README.md
+++ b/README.md
@@ -121,7 +121,8 @@ You can then start the containers.
 docker-compose up -d
 ```
 
-When the containers are running you are able to access the application as normal, and code changes will be reflected in the running application
+When the containers are running you are able to access the application as normal, and code changes will be reflected in the running application.
+However, any new dependencies that are added would require a re-build.
 
 ## Internationalisation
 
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -110,6 +110,20 @@ This will create the environment and spin up the application . Once the applicat
 eb open
 ```
 
+## Run with Docker
```
iwootten: I think if we're looking to make this the default then it probably should be at the/toward the top of the readme?

jonnyshaw89: good point.

.dockerignore
```diff
@@ -1,3 +1,49 @@
...
+
+.DS_Store
+
+node_modules/
+app/static/
+tests/results
+tests/coverage
+tests/a11y
+
+app/styles/base/_sprite.scss
```
iwootten, line 38: How come we've specifically excluded this one partial?

jonnyshaw89: It was in the .gitignore so i assumed it was something that was generated by a script?

README.md
```diff
@@ -1,9 +1,22 @@
 # eQ Survey Runner
 [![Build Status](https://travis-ci.org/ONSdigital/eq-survey-runner.svg?branch=master)](https://travis-ci.org/ONSdigital/eq-survey-runner) [![Code Issues](https://www.quantifiedcode.com/api/v1/project/1709e9d582cc479a86568a043117d4d0/badge.svg)](https://www.quantifiedcode.com/app/project/1709e9d582cc479a86568a043117d4d0) [![codecov](https://codecov.io/gh/ONSdigital/eq-survey-runner/branch/master/graph/badge.svg)](https://codecov.io/gh/ONSdigital/eq-survey-runner) [![Dependency Status](https://gemnasium.com/badges/github.com/ONSdigital/eq-survey-runner.svg)](https://gemnasium.com/github.com/ONSdigital/eq-survey-runner)
 
-Based on python 3
+## Run with Docker
+
+To use docker to run eq-survey-runner you first need to build the container (this may take a few minutes on first run)
```
benjefferies: Don't think this step is required, the docs suggest it builds it if it's missing - https://docs.docker.com/compose/reference/up/. Appears to just work with `docker-compose up -d` when I tried it

benjefferies (changes requested):

README.md
iwootten: Yeah, "up" will build if it hasn't yet.

jonnyshaw89: but if you change the code and do up it won't re-build.

benjefferies: Is it worth rewriting as

To run docker...
To rebuild docker...

A one command up and running sounds quite appealing albeit the two commands aren't bad

iwootten: It's also worth mentioning 

```
docker-compose build --no-cache
```

if anyone needs to rebuild from scratch.

jonnyshaw89: you don't need to rebuild from scratch its just that `docker-compose up -d` doesn't check the layers for rebuild. 
If you do `docker-compose build` again then it will use the cache and just rebuild the changed layers.

iwootten: From experience, there may be upstream changes that don't appear as changed to docker-compose until you don't use the cache - i.e the postgres image we're using is latest and could be updated, and similarly some of the npm deps aren't locked on specific version numbers. In those cases using cached layers doesn't correctly represent a new build. For those cases, it's worth being verbose in the docs on usage.

jonnyshaw89: For the updated version of postgres we would need to do `docker-compose pull` to update that. Maybe we could lock into the version used in RDS?
Ill make a note of the other caveats in the readme though.

benjefferies (approved):

Issue #483, opened 2016-10-10
System Information page title missing


collisdigital:
### Expected behaviour
- https://preprod-surveys.eq.ons.digital/information/multiple-surveys has a sensible page title
### Actual behaviour
- https://preprod-surveys.eq.ons.digital/information/multiple-surveys has no page title set.
### Steps to reproduce the behaviour
- Go here: https://preprod-surveys.eq.ons.digital/information/multiple-surveys
- See page title is empty
### Technical information
- Issue appears to be CAPITAL used when it should be lowercase on {{messages.TITLE}} here:
- https://github.com/ONSdigital/eq-survey-runner/blob/master/app/templates/information.html#L3
#### Browser

All
#### Operating System

All
### Screenshot

<img width="175" alt="screen shot 2016-09-30 at 20 57 31" src="https://cloud.githubusercontent.com/assets/14041600/19238167/2402add0-8ef8-11e6-9606-88c094fe5ad8.png">

collisdigital:
Confirmed fixed in pre-prod.

PR #505, opened 2016-10-19
corrected incorrect case in information title

jonnyshaw89:

Fixing defect https://github.com/ONSdigital/eq-survey-runner/issues/483
### How to review

Check that expected behaviour of defect is met


iwootten (approved):
I can't imagine a better solution to the the problem.

PR #515, opened 2016-10-24
Allow zeros to be entered in a currency field

### LINKED ISSUE #511: Answer with value '0' not stored
### Expected behaviour
- Entering '0' for an answer is stored correctly and presented back to the user.
### Actual behaviour
- Entering '0' for an answer is not displayed on page refresh or on the summary screen.
- Enter 0 for any numeric answer
- Navigate away from the page (save and continue)
- Navigate to the summary screen and see "no answer provided"
- Return to the page with the question on and observe the answer is empty
#### Browser

Chrome
#### Operating System

macOS Sierra
### Screenshot

screen shot 2016-10-20 at 12 29 46
screen shot 2016-10-20 at 12 29 38

LJBabbage:

Fixes #511 . Python sees integer 0 as false therefore our current code is returning None or empty string for currency questions, this therefore means the frontend shows the data wrong, a None check is better in this situation. The state input and the state value don't need both checking, so I have removed one
### How to review

Open any survey and start entering the data making sure you enter '0' into a currency question, submit the survey and make sure '0' appears on the summary page and on the original page when you navigate back to it

### DIFF
--- a/tests/functional/spec/summary-screen.spec.js
+++ b/tests/functional/spec/summary-screen.spec.js
@@ -63,4 +63,29 @@ describe('RSI - summary screen edit test', function() {
     expect(rsiSummaryPage.getChangeInRetailTurnoverSummary()).to.contain('This is to test edit links on summary screen - edited')
 
   })
+  it('Given the RSI survery 0102 when a 0 is entered in a currency field then the summary screen should show £0 and the original page should show 0', function(done) {
+
+    //Given the RSI business survey 0102 is started
+    startQuestionnaire('1_0102.json')
+
+    // when data is entered for the survey
+    reportingPeriod.setFromReportingPeriodDay(2)
+      .setToReportingPeriodDay(2)
+      .setFromReportingPeriodYear(2016)
+      .setToReportingPeriodYear(2017)
+      .submit()
+    retailTurnoverPage.setRetailTurnover(1000)
+      .submit()
+    internetSalesPage.setInternetSales(0)
+      .submit()
+    changeInRetailTurnover.setChangesInRetailTurnover('')
+      .submit()
+
+    // Then summary screen and the original page shows the data entered
+    expect(rsiSummaryPage.getRetailTurnoverSummary()).to.contain('£1,000')
+    expect(rsiSummaryPage.getInternetSalesSummary()).to.contain('£' + 0)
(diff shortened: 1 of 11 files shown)


ajmaddaford:
If I enter zero for the total employees field on the 0112 survey, proceed to the summary, and then click the edit link to go back to the employees question, the total employees field is blank. The currency fields in this survey work correctly with zero values.

tests/functional/spec/summary-screen.spec.js
```diff
@@ -63,4 +63,28 @@ describe('RSI - summary screen edit test', function() {
     expect(rsiSummaryPage.getChangeInRetailTurnoverSummary()).to.contain('This is to test edit links on summary screen - edited')
 
   })
+  it('Given a 0 is entered in a currency field it should still be 0 on the summary page and the original page it was entered', function(done) {
```
benjefferies: For consistency it might be worth continuing with the suggested Given When Then structure examples are here http://martinfowler.com/bliki/GivenWhenThen.html

benjefferies (changes requested):

tests/functional/spec/summary-screen.spec.js
```diff
@@ -63,4 +63,29 @@ describe('RSI - summary screen edit test', function() {
     expect(rsiSummaryPage.getChangeInRetailTurnoverSummary()).to.contain('This is to test edit links on summary screen - edited')
 
   })
+  it('Given a 0 is entered in a currency field when on the summary page and the original page it still is shown as 0', function(done) {
```
ajmaddaford: This test name isn't consistent with the others, should be something more like:
"Given the RSI survey 0102 when a 0 is entered in a currency field then the summary screen should show £0 and the original page should show 0"

ajmaddaford (changes requested):

tests/functional/spec/summary-screen.spec.js
LJBabbage: Good point

yoganandkunche:
Passed QA

ajmaddaford (approved):

benjefferies (approved):

benjefferies:
Rebased master on to branch to update

PR #520, opened 2016-10-24
Moved static content up a directory

jonnyshaw89:

Currently the docker container builds the static assets but when the volume was mounted it lost the compiled assets.
This change moves those static assets so they they are no longer in the app directory
### How to review

When building the docker containers from a clean checkout, does your application have styles when running?


jonnyshaw89:
Confirmed working on ElasticBeanstalk

hamishtaplin (changes requested):
@jonnyshaw89 Line 77 of README needs updating

hamishtaplin:
@jonnyshaw89 Also, the location of the system-generated `_sprite.scss` is wrong throughout. Is it worth fixing that as part of this? If so, I can push those changes up.

benjefferies:
It's not directly related to this but it might be worth adding instructions how to remotely debug within the docker container if you found out in the end

hamishtaplin (approved):

PR #523, opened 2016-10-25
Eq 650 Implement Previous + Next from Schema

iwootten:

Moved to using the schema for navigation, rather than methods on the questionnaire manager. Added a new class Navigator, which deals with finding block and location paths through a survey from a schema. Refactored all the previous routing behaviour to use new navigator and removed redundant dependencies. 

It's been necessary to modify the way a schema is loaded, so it returns a tuple of the parsed schema and original json, so the navigator is able to take the json as an argument.
### How to review

Review the approach, confirm that tests pass and test coverage is appropriate.


app/main/errors.py
```diff
@@ -0,0 +1,73 @@
+import logging
```
benjefferies: Think this can be removed as it's superseded by app/main/view/errors.py

iwootten:
Codecov's patch is complaining as I've refactored a number of method names in preexisting files. I've managed to bump them quite a lot by mocking out boto, but not enough to get our current target of 90.3%. 

I think the patch blocking a PR isn't helpful, given it's difficult to know what the coverage of a patch will be before letting codecov run, so we're forever making updates to a PR to get it to increase.

collisdigital:
This is the coverage report: https://codecov.io/gh/ONSdigital/eq-survey-runner/compare/b9b82d14406a662b6c8bede15bc3ee72e18347c4...9f5190fde5b3680b688116dffcf12997fe2a8240

We can either disable blocking the PR or change the codecov setting (i.e. just fix it at 85% rather than always the same or higher than the parent branch...

app/questionnaire/navigator.py
```diff
@@ -0,0 +1,176 @@
...
+    return path
+
+
+class Navigator:
+
+    def __init__(self, survey_json):
+        self.survey_json = survey_json
+        self.blocks = self.get_blocks()
+        self.first_block_id = self.blocks[0]['id']
+        self.location_path = ['introduction']
```
benjefferies: Is this making the assumption there will always be a introduction, is that the case now?

app/questionnaire/navigator.py
```diff
@@ -0,0 +1,176 @@
...
+        """
+        Returns a list of url locations visited based on answers provided
+        :param answers:
+        :return:
+        """
+        answers = answers or {}
+        routing_path = visit_block(self.blocks, self.first_block_id, answers, [])
+        location_path = self.location_path
+        location_path.extend(routing_path)
+        location_path.extend(['summary', 'thank-you'])
```
benjefferies: Is this making the assumption there will always be a summary and thank you page, is that the case now?

app/questionnaire/questionnaire_manager.py
```diff
@@ -42,6 +42,8 @@ def __init__(self, schema, current=None, first=None, tail=None, archive=None, va
         else:
             self._valid_locations = self._build_valid_locations()
```
benjefferies, line 43: Does build_path() do something similar to this when given no answers? It would be worth using that I guess if it is the case

app/questionnaire/questionnaire_manager.py
```diff
@@ -341,7 +325,15 @@ def _conditional_display(self, item):
 
         if item.schema_item:
 
-            item.skipped = ConditionalDisplay.is_skipped(item.schema_item, self)
+            item.skipped = False
+
+            if hasattr(item.schema_item, 'skip_condition') and item.schema_item.skip_condition:
+                rule = item.schema_item.skip_condition.__dict__
+                rule['when'] = rule['when'].__dict__
```
jonnyshaw89, line 332: I get the following error from this line when completing the star wars survey
`File "/Users/jonshaw/code/eq-survey-runner/app/questionnaire/questionnaire_manager.py", line 332, in _conditional_display
    rule['when'] = rule['when'].__dict__
AttributeError: 'dict' object has no attribute '__dict__'`

LJBabbage: It works for me oddly, and I can see all whens in rules are dicts

jonnyshaw89: It also no longer breaks for me. Worth investigating? or ignoring and putting it down to switching between branches etc?

LJBabbage:
If I click on the first page I prefer Star Trek, it gives me a 500 error
block_id_index = next(index for (index, b) in enumerate(blocks) if b["id"] == block_id)
StopIteration

app/globals.py
```diff
@@ -0,0 +1,47 @@
...
+    if schemas is None:
+        try:
+            schemas = g._schemas = load_all_schemas()
+        except Exception as e:
+            logger.error("schemas failed to init", exception=repr(e))
+
+    return schemas
+
+
+def get_schema(eq_id, form_type):
```
LJBabbage: Is this function being called? questionnaire_factory still seems to be pointed at the old get_schema in schema.py

app/globals.py
```diff
@@ -0,0 +1,47 @@
...
+from app.schema_loader.schema_loader import load_all_schemas
+
+from flask import g
+
+logger = logging.getLogger(__name__)
+
+
+def get_schemas():
+    # Sets up a single schema accessor instance throughout app.
+    schemas = g.get('_schemas')
```
LJBabbage: This is for request scope, not sure why we would load every schema is we lose it after the request

LJBabbage: Won't this always be None as its requests scope, also why are we loading all schemas in request scope

app/globals.py
iwootten: Good spot - I was using flask globals but ended up ditching the approach given the questionnaire manager is tested outside of a flask request.

app/globals.py
iwootten: No, worked the same as the other get globals. Have removed this all for now, given it's not used. Finally codecov is happy now.

iwootten:
@LJBabbage - The recursive visit_block method was only working with block ids and had no idea interstitial pages could be jumped to. Have added a test to catch it and handled the case.

app/questionnaire/navigator.py
```diff
@@ -0,0 +1,182 @@
...
+    def get_first_block_id(self):
+        return self.survey_json['groups'][0]['blocks'][0]['id']
+
+    def get_blocks(self):
+        blocks = []
+        for group in self.survey_json['groups']:
+            blocks.extend([block for block in group['blocks']])
+        return blocks
+
+    def get_block_ids(self):
```
collisdigital: There's no test coverage on this function; which is suspicious, would have thought other tests would have hit it indirectly, is this function redundant ?

https://codecov.io/gh/ONSdigital/eq-survey-runner/compare/b9b82d14406a662b6c8bede15bc3ee72e18347c4...333b0e5b685e2fab03a8eb7ad9cdaeac72c01c1a#6170702F7175657374696F6E6E616972652F6E6176696761746F722E7079-108

app/questionnaire/navigator.py
```diff
@@ -0,0 +1,182 @@
...
+
+    # Evaluate the condition on the routing rule
+    if condition == 'equals' and match_value == answer:
+        return True
+    elif condition == 'not equals' and match_value != answer:
+        return True
+    return False
+
+
+def visit_block(blocks, block_id, answers, path):
```
ajmaddaford: I think this function name should more clearly reflect what is does, something like `build_path` (which it does by visiting blocks)

app/questionnaire/navigator.py
```diff
@@ -0,0 +1,182 @@
...
+                return visit_block(blocks, goto_rule['id'], answers, path)
+    else:
+        if block_id_index != len(blocks) - 1:
+            next_block_id = blocks[block_id_index + 1]['id']
+            return visit_block(blocks, next_block_id, answers, path)
+    return path
+
+
+class Navigator:
+    PROCEEDING_INTERSTITIAL_PATH = ['introduction']
```
ajmaddaford: I think this should be preceding rather than proceeding?

app/questionnaire/navigator.py
```diff
@@ -0,0 +1,182 @@
...
+        """
+        Returns a list of the block ids visited based on answers provided
+        :return: List of block ids
+        """
+        answers = answers or {}
+        routing_path = []
+
+        return visit_block(self.blocks, self.first_block_id, answers, routing_path)
+
+    def get_location_path(self, answers=None):
```
ajmaddaford, line 114: I think `get_locations_visited` would be a better name

app/questionnaire/navigator.py
```diff
@@ -0,0 +1,182 @@
...
+
+            if current_block_id in routing_path:
+                block_route_index = routing_path.index(current_block_id)
+
+                if block_route_index != 0:
+                    return routing_path[block_route_index - 1]
+
+        return current_block_id
+
+    def get_next_block_id(self, answers=None, current_block_id=None):
```
ajmaddaford: I can only see this method used in the tests - has this been superseded by `get_next_location`?

app/schema_loader/schema_loader.py
```diff
@@ -27,48 +27,64 @@ def load_schema(eq_id, form_type):
...
         s3 = boto3.resource('s3')
         schemas_bucket = s3.Bucket(settings.EQ_SCHEMA_BUCKET)
         for key in schemas_bucket.objects.all():
             files.append(key.key)
     except botocore.exceptions.ClientError as e:
         logging.error("S3 error: %s", e.response['Error']['Code'])
+    return files
+
+
+def load_all_schemas():
```
ajmaddaford: Unused method?

ajmaddaford (changes requested):

app/questionnaire/navigator.py
iwootten: I'm not sure, given the list of locations haven't yet been visited, at least from a respondents point of view.

app/questionnaire/navigator.py
iwootten: Sure thing.

app/schema_loader/schema_loader.py
iwootten: Yup

app/questionnaire/navigator.py
iwootten: Yes, you're right.

app/questionnaire/navigator.py
iwootten: Yes.

app/questionnaire/navigator.py
iwootten: Yes.

app/questionnaire/questionnaire_manager.py
iwootten: Unless this can be replicated (I've never hit this issue and don't on this branch), then I can't see how we would pursue it.

app/utilities/schema.py
```diff
@@ -16,11 +16,11 @@ def get_schema():
...
-    if not schema:
+    json_schema, schema = load_and_parse_schema(eq_id, form_type)
+    if not json_schema:
         raise ValueError("No schema available")
     logger.debug("Constructing brand new User Journey Manager")
-    return schema
+    return json_schema, schema
 
 
 def load_and_parse_schema(eq_id, form_type):
```
benjefferies, line 30: The docs are missing the other return object

benjefferies (changes requested):

app/questionnaire/questionnaire_manager.py
iwootten: Yes, leaving this for @jonnyshaw89's PR branching from this.

ajmaddaford (approved):

app/questionnaire/navigator.py
```diff
@@ -0,0 +1,174 @@
...
+        a set of answers
+        :param routing_path:
+        :param answers:
+        :return:
+        """
+        answers = answers or {}
+        routing_path = routing_path or build_path(self.blocks, self.first_block_id, answers, [])
+        last_routing_block_id = routing_path[len(routing_path) - 1]
+
+        if self.blocks[len(self.blocks) - 1]['id'] == last_routing_block_id:
```
benjefferies: self.blocks[-1]['id'] could work here I think?

app/questionnaire/navigator.py
```diff
@@ -0,0 +1,174 @@
...
+        """
+        Determines whether the end of a given routing path can be reached given
+        a set of answers
+        :param routing_path:
+        :param answers:
+        :return:
+        """
+        answers = answers or {}
+        routing_path = routing_path or build_path(self.blocks, self.first_block_id, answers, [])
+        last_routing_block_id = routing_path[len(routing_path) - 1]
```
benjefferies: routing_path[-1] could work here I think?

benjefferies (approved):

PR #529, opened 2016-10-27
Refactor summary to use answer dict and schema dict

benjefferies:

### How to review
* Refactoring summary model logic to use answer and schema dict which is
  essential a json representation of the answers and schema json.
* Refactoring plumber/template logic to use jinja instead of homegrown
 template implementation.
* Add new jinja filter for pretty dates.
* Summary page data-qa attribute hooks use element ids instead of an id
  based on the loop iteration
* Removing formatter registry for summary page.

All tests should still pass


app/templates/partials/summary/checkbox.html
```diff
@@ -1,5 +1,5 @@
 <ul class="u-m-no">
-  {% for answer in sub_item.answer %}
+  {% for answer in answer.answer %}
```
iwootten: This is pretty confusing..

app/templates/partials/summary/currency.html
```diff
@@ -0,0 +1 @@
+{{answer.answer|format_currency}}
```
iwootten: Similarly here.

app/templates/partials/summary/date.html
```diff
@@ -1,3 +1,3 @@
 {% set date_format = '%d %b %Y' %}
 
-{{sub_item.answer[0].value.strftime(date_format)}} to {{sub_item.answer[1].value.strftime(date_format)}}
+{{answer.answer['from'].strftime(date_format)}} to {{answer.answer['to'].strftime(date_format)}}
```
iwootten: and here...

app/templating/model_builder.py
```diff
@@ -0,0 +1,55 @@
...
+
+    return render_data
+
+
+def _build_summary(questionnaire_schema):
+    q_manager = g.questionnaire_manager
+    answers = q_manager.get_answers()
+    path = q_manager.navigator.get_routing_path(answers)
+    sections = []
+    for group in questionnaire_schema['groups']:
```
iwootten, line 51: I wonder if we should group all the carving up of the schema into a simple helper class that can then be reused? Currently a lot of similar stuff is done in Navigator too.

app/templating/summary/question.py
```diff
@@ -0,0 +1,76 @@
...
+        self.id = question_schema['id']
+        self.type = question_schema['type']
+        self.skip_condition = question_schema.get('skip_condition')
+        answer_schema = question_schema['answers']
+        self.title = question_schema['title'] or answer_schema[0]['label']
+        self.answers = Question._build_answers(block_id, question_schema, answer_schema, answers)
+
+    def is_skipped(self, answers):
+        if self.skip_condition is not None:
+            answer = answers[self.skip_condition['when']['id']]
```
iwootten: Will we always have an answer in answers that has a matching id?

app/templating/summary/question.py
```diff
@@ -0,0 +1,76 @@
...
+            return evaluate_rule(self.skip_condition, answer)
+
+        return False
+
+    @staticmethod
+    def _build_answers(block_id, question_schema, answer_schema, answers):
```
iwootten: Does it make sense that these methods are instance methods and the class holds the arguments when initialised, so we don't have to keep passing them in?

app/templating/summary/section.py
```diff
@@ -0,0 +1,18 @@
...
+
+class Section:
+
+    def __init__(self, block_id, section_schema, answers):
+        self.id = section_schema['id']
+        self.title = section_schema['title']
+        self.questions = Section._build_questions(block_id, section_schema, answers)
+
+    @staticmethod
+    def _build_questions(block_id, section_schema, answers):
```
iwootten, line 12: Same comment regarding instance vs static methods here too

tests/app/piping/test_plumber.py
```diff
@@ -23,74 +26,10 @@ def test_plumb_item(self):
...
-        self.plumber.plumb_item(item)
+        with patch('app.piping.plumber.render_template_string', return_value="Property One is value one, while Property Two is value two"):
+            self.plumber.plumb_item(item)
 
         self.assertEquals(item.description, "Property One is value one, while Property Two is value two")
-
-        item2 = ObjectFromDict({
-            "id": "item2",
-            "templatable_properties": ['property_one', 'property_two', 'property_three'],
-            "property_one": "Date is {dates.first_april_2016:%-d %B %Y}",
```
iwootten, line 38: We probably should keep some tests to demonstrate dates are correctly formatted.

iwootten (changes requested):

app/templating/summary/question.py
benjefferies: I think we should have an answer but I have used .get() incase

app/templating/summary/question.py
benjefferies: Debatable?

app/templating/summary/section.py
benjefferies: Debatable?

tests/app/piping/test_plumber.py
benjefferies: I have added a unit test for the function and we already have integration tests covering this

app/jinja_filters.py
```diff
@@ -4,78 +4,21 @@
...
+def format_currency(value):
     return "£{:,}".format(value)
 
-blueprint.add_app_template_filter(format_currency)
-
 
-@contextfilter
 @blueprint.app_template_filter()
-def prettify(context, code):
+def prettify(code):
```
ajmaddaford: This method is only ever used in `templates/macros/patterns.html`, which is no longer used since all the patterns stuff was removed in an earlier pull request. I think we should remove this method and the template as part of this pull request.

app/piping/plumber.py
```diff
@@ -44,7 +46,7 @@ def _plumb(self, item, templatable_property):
             template = getattr(item, templatable_property)
```
ajmaddaford: I think this variable would be clearer named `template_string` rather than `template`.

app/piping/plumber.py
```diff
@@ -2,6 +2,8 @@
 
 import re
 
+from flask import render_template_string
```
ajmaddaford, line 5: We should consider using jinja directly rather than via flask here, maybe as part of a future piece of work though.

app/piping/plumbing_preprocessor.py
```diff
@@ -12,13 +11,51 @@
...
+        value = questionnaire_manager.find_answer(item_id)
+        if value is None:
+            value = ""  # Empty string
+        values[alias] = value
+    return values
+
+
 class PlumbingPreprocessor(object):
 
     def plumb_current_state(self, questionnaire_manager, state, schema):
```
ajmaddaford, line 57: Can we remove the passing of `questionnaire_manager` and replace with the `answers`? Currently for every alias that exists the full linked list is iterated over to return a single answer.

app/templating/model_builder.py
benjefferies: After discussion @iwootten has started creating helper functions on his branch so will hold off

app/templates/partials/summary/date.html
```diff
@@ -1,3 +1,3 @@
 {% set date_format = '%d %b %Y' %}
 
-{{sub_item.answer[0].value.strftime(date_format)}} to {{sub_item.answer[1].value.strftime(date_format)}}
+{{answer.value['from'].strftime(date_format)}} to {{answer.value['to'].strftime(date_format)}}
```
ajmaddaford: Can this be piped to a jinja filter like the prettify_date filter?

app/templating/summary/question.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def __init__(self, block_id, question_schema, answers):
+        self.id = question_schema['id']
+        self.type = question_schema['type']
+        self.skip_condition = question_schema.get('skip_condition')
+        answer_schema = question_schema['answers']
+        self.title = question_schema['title'] or answer_schema[0]['label']
+        self.answers = Question._build_answers(block_id, question_schema, answer_schema, answers)
+
+    def is_skipped(self, answers):
```
ajmaddaford: Why do we pass answers to this method rather than use `self.answers`?

ajmaddaford (changes requested):

app/jinja_filters.py
benjefferies: Done

app/piping/plumber.py
benjefferies: Maybe we could do this as a part of removing plumbing stuff. Spoke to @iwootten about this and it was making testing quite difficult as a part of the testing.

app/piping/plumber.py
benjefferies: Done

app/piping/plumbing_preprocessor.py
benjefferies: I started doing this and reverted it because I thought the PR was big enough as it is. Maybe we can visit as a part of the removing of the plumbing stuff; it will probably naturally fall away?

app/templates/partials/summary/date.html
benjefferies: Good spot, done

app/templating/summary/question.py
benjefferies: The question only holds reference to its own answers, we could have reference to all answers but I think we'd be holding state for the sake of the function instead of passing it

app/piping/plumbing_preprocessor.py
ajmaddaford: Fair point - it has been removed in #534 😄

app/templating/summary/question.py
ajmaddaford: I hadn't realised that! Maybe renaming it something like `all_answers` would make this clearer

ajmaddaford (changes requested):

app/templating/summary/question.py
benjefferies: Makes perfect sense, will do

iwootten (approved):

ajmaddaford (approved):

PR #534, opened 2016-11-01
Remove linked list from questionnaire manager

jonnyshaw89:
This is the final change to remove the state storage and pickling from the system to enable more flexibility in software upgrades/changes and more robust data handling.

### How to review 
 - No reference to the linked list in the system anymore
 - JSON pickle is no longer a dependency
 - Answers are stored via QuestionnaireStore as simple JSON/text (not pickled).


app/main/views/questionnaire.py
```diff
@@ -47,41 +47,28 @@ def add_cache_control(response):
...
-    g.questionnaire_manager.go_to(location)
     return render_page(location, True)
 
 
 @questionnaire_blueprint.route('<location>', methods=["POST"])
 @login_required
 def post_questionnaire(eq_id, form_type, period_id, collection_id, location):
-    valid = g.questionnaire_manager.process_incoming_answers(location, request.form)
-    if not valid:
+    valid, next_location = g.questionnaire_manager.process_incoming_answers(location, request.form)
```
ajmaddaford: I think `process_incoming_answers` should just return whether the questionnaire answers are valid. We could combine the next location logic from this method with the code that follows.

tests/integration/star_wars/test_previous_link.py
```diff
@@ -1,89 +0,0 @@
-from tests.integration.star_wars.star_wars_tests import StarWarsTestCase
-from tests.integration.star_wars import star_wars_test_urls
-
-class TestPreviousLink(StarWarsTestCase):
-
-    def test_previous_link_path(self):
```
iwootten, line 6: Has this integration test been removed and the behaviour tested elsewhere?

jonnyshaw89: The previous page is now just a href so no value testing that with integration tests but its covered in selenium

app/questionnaire/questionnaire_manager.py
```diff
@@ -178,130 +50,79 @@ def validate(self, post_data):
...
-    def go_to(self, location):
-        if self._current:
-            logger.debug("Attempting to go to location %s with current location as %s", location, self._current.item_id)
-        if location == 'previous':
-            location = self.navigator.get_previous_location(self.get_answers(), current_location_id=location)
-        elif location == 'summary' and self._current.item_id != 'summary':
-            metadata = get_metadata(current_user)
-            if metadata:
-                logger.warning("User with tx_id %s tried to submit in an invalid state", metadata["tx_id"])
-            raise InvalidLocationException()
```
benjefferies, line 219: InvalidLocationException can be deleted and the exception handler in errors.py can be to

app/questionnaire/questionnaire_manager.py
```diff
@@ -178,130 +50,79 @@ def validate(self, post_data):
...
             self._plumbing_preprocessing(self.state)
             self._conditional_display(self.state)
 
-        if self.state not in state_items:
-            state_items.append(self.state)
+        previous_location = self.navigator.get_previous_location(get_answers(current_user), location)
 
         # look up the preprocessor and then build the view data
-        return build_questionnaire_model(self._json, state_items)
+        return build_questionnaire_model(self._json, [self.state], previous_location)
```
benjefferies: Is it worth changing so it's not an array and amending the downstream of that

app/data_model/questionnaire_store.py
```diff
@@ -41,3 +41,17 @@ def metadata(self):
...
+
+    @property
+    def answers(self):
+        if "ANSWERS" not in self.data:
+            self.data["ANSWERS"] = {}
+
+        return self.data["ANSWERS"]
+
+    @property
+    def visited_blocks(self):
```
ajmaddaford: I think `completed_blocks` be a better name, as a block could be viewed as visited in a GET request for a block.

app/main/views/questionnaire.py
```diff
@@ -91,8 +78,11 @@ def get_thank_you(eq_id, form_type, period_id, collection_id):
 @questionnaire_blueprint.route('summary', methods=["GET"])
 @login_required
 def get_summary(eq_id, form_type, period_id, collection_id):
-    g.questionnaire_manager.go_to('summary')
-    return render_page('summary', True)
+    current_location = g.questionnaire_manager.navigator.get_current_location(get_answers(current_user), get_visited_blocks(current_user))
```
ajmaddaford: I think `current_location` should be renamed something like `latest_location` (both the variable and method name), I think `current_location` could get confused with the currently requested location from the URL.

app/main/views/root.py
```diff
@@ -75,7 +75,14 @@ def login():
...
-    current_location = questionnaire_manager.get_current_location()
+    current_location = questionnaire_manager.navigator.get_first_location()
+
+    visited_blocks = get_visited_blocks(current_user)
+
+    if visited_blocks:
+        incomplete_blocks = [item for item in questionnaire_manager.navigator.get_location_path(get_answers(current_user)) if item not in visited_blocks]
+
+        if incomplete_blocks:
+            current_location = incomplete_blocks[0]
```
ajmaddaford: I think all of this should be done in `questionnaire.py`. Also, isn't all of this replaceable with a call to `get_current_location` in `navigator.py`?

app/questionnaire/navigator.py
```diff
@@ -172,3 +172,21 @@ def get_previous_location(self, answers=None, current_location_id=None):
 
             if current_location_index != 0:
                 return location_path[current_location_index - 1]
+
+    def get_current_location(self, answers=None, visited_blocks=None):
```
ajmaddaford: As previous comment, I think current should be renamed something like latest

app/questionnaire/questionnaire_manager.py
```diff
@@ -24,151 +21,26 @@ class InvalidLocationException(Exception):
 class QuestionnaireManager(object):
     '''
     This class represents a user journey through a survey. It models the request/response process of the web application
-    using a doubly linked list. Each node in the list is a reference to a schema item and the answer the user has entered.
-    A new node is created by a GET request and subsequently updated via a POST request.
-    The doubly linked list approach allows us to maintain the path the user has taken through the question. If that path
-    changes we archive off the nodes in case the user revisits that path.
     '''
-    def __init__(self, schema, current=None, first=None, tail=None, archive=None, valid_locations=None, submitted_at=None, json=None):
+    def __init__(self, schema, submitted_at=None, json=None):
```
ajmaddaford: I think `submitted_at` can go? `QuestionnaireManager` is only initialised in one place (`QuestionnaireManagerFactory`) and it doesn't pass `submitted_at`

app/questionnaire/questionnaire_manager.py
```diff
@@ -178,130 +50,79 @@ def validate(self, post_data):
...
+
+        is_valid = self.validate(location, post_data)
+        # run the validator to update the validation_store
+        if is_valid:
+
+            # Store answers in QuestionnaireStore
+            questionnaire_store = get_questionnaire_store(current_user.user_id, current_user.user_ik)
+
+            if location not in questionnaire_store.visited_blocks:
+                questionnaire_store.visited_blocks.append(location)
```
ajmaddaford: If we always append, does this mean we can have duplicates if someone goes back and re-submits a block?

app/questionnaire/questionnaire_manager.py
```diff
@@ -178,130 +50,79 @@ def validate(self, post_data):
...
 
     def get_summary_rendering_context(self):
         schema_template_context = get_schema_template_context(self, self._schema)
         rendered_questionnaire_schema_json = render_template_string(json.dumps(self._json), **schema_template_context)
+
+        previous_location = self.navigator.get_previous_location(get_answers(current_user), "summary")
+
         # look up the preprocessor and then build the view data
-        return build_summary_model(json.loads(rendered_questionnaire_schema_json))
+        return build_summary_model(json.loads(rendered_questionnaire_schema_json), previous_location)
```
ajmaddaford: I think we should add `previous_location` to the model returned by `build_summary_model` rather than pass it to the method to add it - it's not really anything to do with the summary model.

app/questionnaire/questionnaire_manager_factory.py
```diff
@@ -14,28 +13,7 @@ class QuestionnaireManagerFactory(object):
...
-        logger.debug("QuestionnaireManagerFactory saving state")
-        return questionnaire_manager
+        json, schema = get_schema()
 
-    @staticmethod
-    def _get_schema():
-        # exists to facilitate testing
-        return get_schema()
+        logger.debug("QuestionnaireManagerFactory - constructing new instance")
+        return QuestionnaireManager(schema, json=json)
```
ajmaddaford: Maybe it's time to remove the QuestionnaireManagerFactory?

app/questionnaire/questionnaire_manager.py
```diff
@@ -319,8 +140,9 @@ def _conditional_display(self, item):
 
             if hasattr(item.schema_item, 'skip_condition') and item.schema_item.skip_condition:
                 rule = item.schema_item.skip_condition.__dict__
-                rule['when'] = rule['when'].__dict__
-                answer = self.find_answer(rule['when']['id'])
+                # Sometimes its a dict and sometimes its an object?
+                rule['when'] = rule['when'].__dict__ if not isinstance(rule['when'], dict) else rule['when']
```
ajmaddaford: It looks like this has been fixed in https://github.com/ONSdigital/eq-survey-runner/issues/540

app/schema/answer.py
```diff
@@ -86,6 +89,8 @@ def validate(self, state):
             if state.input:
                 try:
                     state.value = self.get_typed_value(state.input)
+
+                    get_answers(current_user)[self.id] = state.other or state.value
```
ajmaddaford: Why is this needed here? We shouldn't be setting a value in the answer store on validate.

app/schema/answers/checkbox_answer.py
```diff
@@ -31,6 +34,7 @@ def validate(self, state):
                 super(CheckboxAnswer, self).mandatory_error(state)
 
             # Here we just report on whether the answer has passed type checking
+            get_answers(current_user)[self.id] = self.get_typed_value(state.input)
```
ajmaddaford: Same as above.

tests/app/questionnaire/test_questionnaire_manager.py
```diff
@@ -23,199 +23,3 @@ def setUp(self):
         parser = SchemaParserFactory.create_parser(self.json)
```
benjefferies: We should probably add some unit test coverage for questionnaire_manager without any the file is redundant

app/templating/model_builder.py
```diff
@@ -1,41 +1,48 @@
...
     :param questionnaire_schema: variable substituted questionnaire schema
     :param state_items:
+    :param previous_location:
     :return: context dict for questionnaire page variable substitution.
     '''
     metadata_template_preprocessor = MetaDataTemplatePreprocessor()
     render_data = {
       "meta": metadata_template_preprocessor.build_metadata(questionnaire_schema),
       "content": state_items[0],
+      "previous_location": previous_location,
```
ajmaddaford: Same as previous comment on this, would be better to do this outside of this method - previous location has nothing to do with building the questionnaire model. We should also look at removing the metadata stuff as well some time soon.

app/questionnaire/questionnaire_manager_factory.py
```diff
@@ -14,28 +13,7 @@ class QuestionnaireManagerFactory(object):
...
-                                                         json=json)
-        else:
-            logger.debug("QuestionnaireManagerFactory - constructing new instance")
-            questionnaire_manager = QuestionnaireManager(schema, json=json)
-            questionnaire_manager.go_to(questionnaire_manager.get_current_location())
-
-        # immediately save it to the database
-        logger.debug("QuestionnaireManagerFactory saving state")
-        return questionnaire_manager
+        json, schema = get_schema()
```
benjefferies: I think we can delete the factory class. It doesn't do anything special anymore

app/questionnaire/user_action_processor.py
```diff
@@ -1,7 +1,7 @@
 import logging
```
benjefferies: With little work we could make this class redundant but you could argue it's out of scope

tests/integration/mci/test_empty_questionnaire.py
```diff
@@ -20,12 +20,21 @@ def test_empty_questionnaire(self):
         content = resp.get_data(True)
         self.assertRegexpMatches(content, '>Start survey<')
 
+        post_data = {
```
benjefferies, line 23: I can't follow this. Why did we need to change the test?

tests/integration/mci/test_empty_questionnaire.py
```diff
@@ -20,12 +20,21 @@ def test_empty_questionnaire(self):
...
+          'action[start_questionnaire]': "Submit Answers"
+        }
+
+        resp = self.client.post(intro_page_url, data=post_data, follow_redirects=False)
+        self.assertEquals(resp.status_code, 302)
+
+        block_one_url = resp.headers['Location']
```
ajmaddaford: I think `block_one_url` is misleading, as won't this be the introduction page?

tests/integration/mci/test_submission_with_errors.py
```diff
@@ -96,7 +59,7 @@ def test_submission_with_errors(self):
         post_data = {
             "action[submit_answers]": "Submit answers"
         }
-        resp = self.client.post(summary_url, data=post_data, follow_redirects=False)
+        resp = self.client.post(mci_test_urls.MCI_0205_SUMMARY, data=post_data, follow_redirects=False)
         self.assertEquals(resp.status_code, 302)
```
benjefferies, line 63: Am I mad in thinking this test doesn't even test it has any errors?

tests/integration/mci/test_submission_with_errors.py
```diff
@@ -37,43 +37,6 @@ def test_submission_with_errors(self):
...
-        self.assertEquals(resp.status_code, 200)
-
-        # We are on the review answers page
-        content = resp.get_data(True)
-        self.assertRegexpMatches(content, '<title>Summary</title>')
-        self.assertRegexpMatches(content, '>Monthly Business Survey - Retail Sales Index</')
-        self.assertRegexpMatches(content, '>Your responses<')
-        self.assertRegexpMatches(content, 'Please check carefully before submission')
-        self.assertRegexpMatches(content, '>Submit answers<')
-
```
ajmaddaford, line 76: Are we removing all these tests because they are covered by the selenium tests?

benjefferies (changes requested):

ajmaddaford (changes requested):

app/questionnaire/questionnaire_manager.py
jonnyshaw89: Yes it can

app/questionnaire/questionnaire_manager.py
benjefferies: Looks like it checks if it's already in there

app/questionnaire/questionnaire_manager.py
jonnyshaw89: Yes i can now. i left it like that as it was being used for summary, but now isn't so ill update that.

app/questionnaire/user_action_processor.py
jonnyshaw89: Yeh i think that would make this PR too broad.

tests/integration/mci/test_empty_questionnaire.py
jonnyshaw89: To get the location that we should be redirected to.

LJBabbage:
Not sure if this story has caused this, but choosing 'other' on the first page throws a 500

LJBabbage:
Are we saying that once a block is valid you can never make it invalid again?

app/main/views/questionnaire.py
jonnyshaw89: Updated

app/data_model/questionnaire_store.py
jonnyshaw89: done

app/main/views/root.py
jonnyshaw89: It is replaceable, i must have missed this when i moved the logic into get_current_location

app/questionnaire/questionnaire_manager.py
jonnyshaw89: Yeh must have missed that one.

app/questionnaire/questionnaire_manager.py
jonnyshaw89: it checks existence first.

app/questionnaire/questionnaire_manager.py
jonnyshaw89: Yeh that makes sense

app/questionnaire/questionnaire_manager.py
jonnyshaw89: It has, i will remove my little hack.

tests/integration/mci/test_submission_with_errors.py
benjefferies: As discussed the 302 implicitly means it has an error, would it be more explicit to check for 'Page has x errors'

app/questionnaire/questionnaire_manager_factory.py
jonnyshaw89: Removed

tests/integration/mci/test_submission_with_errors.py
jonnyshaw89: It was submitting valid answers the submitting invalid answers, but as we no longer save invalid answers the validation on submission didn't fail as the answers were still valid.
so if we don't submit valid answers first then the submission has errors so redirects back to the invalid page.

tests/integration/mci/test_submission_with_errors.py
jonnyshaw89: It tests that it redirects to the invalid page.

benjefferies: but would a self.assertRegex(resp.get_data(True), 'Page has 1 error') on 104 be more explicit and make it more obvious

jonnyshaw89: because its a browser redirect it won't have the validation errors

app/main/views/questionnaire.py
benjefferies: Good point, more cohesion would be a good thing

app/main/views/questionnaire.py
```diff
@@ -58,21 +56,22 @@ def post_questionnaire(eq_id, form_type, period_id, collection_id, location):
...
-def go_to_previous_page(eq_id, form_type, period_id, collection_id):
-    q_manager = g.questionnaire_manager
-    current_location = q_manager.get_current_location()
-    answers = q_manager.get_answers()
-    previous_location = q_manager.navigator.get_previous_location(answers, current_location)
-    g.questionnaire_manager.go_to(previous_location)
-    return redirect_to_questionnaire_page(eq_id, form_type, period_id, collection_id, previous_location)
+def get_summary(eq_id, form_type, period_id, collection_id):
+    navigator = g.questionnaire_manager.navigator
+    latest_location = navigator.get_latest_location(get_answers(current_user), get_completed_blocks(current_user))
```
benjefferies, line 70: Do we have a test to check that you can't skip to the summary page if that's what this is supposed to be limiting?

app/main/views/questionnaire.py
```diff
@@ -81,18 +80,27 @@ def get_thank_you(eq_id, form_type, period_id, collection_id):
...
+    answers = get_answers(current_user)
+
+    # check that all the answers we have are valid before submitting the data
+    is_valid, invalid_location = g.questionnaire_manager.validate_all_answers()
+
+    if is_valid:
+        submitter = SubmitterFactory.get_submitter()
+        submitter.send_answers(current_user, get_metadata(current_user), g.questionnaire_manager._schema, answers)
+
+        return redirect_to_questionnaire_page(eq_id, form_type, period_id, collection_id, "thank-you")
```
benjefferies: We seem to be inconsistent in the quotes we are using, my consistency OCD is coming out

app/main/views/questionnaire.py
jonnyshaw89: We do. TestEmptyQuestionnaire

app/main/views/questionnaire.py
```diff
@@ -73,14 +86,21 @@ def get_thank_you(eq_id, form_type, period_id, collection_id):
...
 @login_required
-def get_summary(eq_id, form_type, period_id, collection_id):
-    latest_location = g.questionnaire_manager.navigator.get_latest_location(get_answers(current_user), get_completed_blocks(current_user))
-    if latest_location is 'summary':
-        return render_page('summary', True)
+def submit_answers(eq_id, form_type, period_id, collection_id):
+    answers = get_answers(current_user)
 
-    return redirect_to_questionnaire_page(eq_id, form_type, period_id, collection_id, latest_location)
+    # check that all the answers we have are valid before submitting the data
```
ajmaddaford, line 94: I don't think we need this comment 😛

app/main/views/questionnaire.py
```diff
@@ -73,14 +86,21 @@ def get_thank_you(eq_id, form_type, period_id, collection_id):
...
+    is_valid, invalid_location = g.questionnaire_manager.validate_all_answers()
+
+    if is_valid:
+        submitter = SubmitterFactory.get_submitter()
+        submitter.send_answers(current_user, get_metadata(current_user), g.questionnaire_manager._schema, answers)
+
```
ajmaddaford: Unnecessary new line (it looks odd because the statement that follows is closer to the else than the line it's related to)

app/main/views/questionnaire.py
```diff
@@ -81,18 +80,27 @@ def get_thank_you(eq_id, form_type, period_id, collection_id):
...
-    return render_page('summary', True)
+def submit_answers(eq_id, form_type, period_id, collection_id):
+    answers = get_answers(current_user)
+
+    # check that all the answers we have are valid before submitting the data
+    is_valid, invalid_location = g.questionnaire_manager.validate_all_answers()
+
+    if is_valid:
+        submitter = SubmitterFactory.get_submitter()
+        submitter.send_answers(current_user, get_metadata(current_user), g.questionnaire_manager._schema, answers)
```
ajmaddaford: `current_user` isn't used by the submitter. Also `_schema` is a 'private' property of questionnaire manager, we should either rename the property `schema` or create an accessor.

app/questionnaire/navigator.py
```diff
@@ -172,3 +172,21 @@ def get_previous_location(self, answers=None, current_location_id=None):
 
             if current_location_index != 0:
                 return location_path[current_location_index - 1]
+
+    def get_latest_location(self, answers=None, completed_blocks=None):
+        """
+        Returns the current 'location' based on the location path and previously visited blocks
```
ajmaddaford: current -> latest, visited -> completed

app/questionnaire/questionnaire_manager.py
```diff
@@ -178,134 +44,76 @@ def validate(self, post_data):
...
-        self.go_to_node(location)
-
-    def is_known_node(self, location):
-        node = self._tail
-        while node:
-            if node.item_id == location:
-                return True
-            node = node.previous
-        return False
+        for location in self.navigator.get_location_path(get_answers(current_user)):
```
ajmaddaford: `get_answers` is called potentially many times, it would be better to call it once and assign it to an `answers` variable. This would make it clear that we don't expect get_answers to return something different on each loop iteration.

app/questionnaire/questionnaire_manager.py
```diff
@@ -178,134 +44,76 @@ def validate(self, post_data):
...
         self.state = None
-        if self._schema.item_exists(node.item_id):
-            schema_item = self._schema.get_item_by_id(node.item_id)
+        if self._schema.item_exists(item_id):
+            schema_item = self._schema.get_item_by_id(item_id)
             self.state = schema_item.construct_state()
             self.state.update_state(answers)
 
-    def _plumbing_preprocessing(self, state):
+    def get_answers(self, item_id):
```
ajmaddaford: I think this method name could be confused with the global `get_answers`, `get_state_answers` or `get_answers_from_state` would be better.

ajmaddaford (changes requested):

ajmaddaford (approved):

benjefferies (approved):

PR #537, opened 2016-11-02
Spike household composition composite

samiwel:
Pull request to review Spike work for the household composition question against the Spike objectives.

### How to review 
Rather than reviewing/correcting individual code snippets, it would probably be more useful to review the overall approach and comment on high level changes that could be made?
Does the approach taken (CompositeAnswer and HouseholdQuestion) make sense? Is there a better way that has been overlooked?

### DIFF
--- a/app/data/census_household_composition.json
+++ b/app/data/census_household_composition.json
@@ -42,14 +42,14 @@
                                   "validation": [],
                                   "answers": [
                                     {
-                                        "id": "composite_answer",
+                                        "id": "person",
                                         "type": "Composite",
                                         "q_code": "9999",
                                         "mandatory": false,
                                         "answers": [
                                           {
                                             "display": {},
-                                            "id": "414699da-1667-44fd-8e98-7606966884db",
+                                            "id": "first",
                                             "label": "First name",
                                             "mandatory": true,
                                             "options": [],
@@ -59,7 +59,7 @@
                                         },
                                         {
                                             "display": {},
-                                            "id": "414699da-1667-44fd-8e98-7606966884d",
+                                            "id": "middle",
                                             "label": "Middle name/s",
                                             "mandatory": false,
                                             "options": [],
@@ -69,7 +69,7 @@
                                         },
                                         {
                                             "display": {},
-                                            "id": "414699da-1667-44fd-8e98-760696688",
+                                            "id": "last",
                                             "label": "Last name",
                                             "mandatory": true,
                                             "options": [],
@@ -77,8 +77,7 @@
                                             "type": "Textfield",
                                             "validation": {}
                                         }]
-                                    }
-                                  ]
+                                    }]
 
                                 }
                             ],
--- a/app/data/census_household_composition.json
+++ b/app/data/census_household_composition.json
@@ -1,0 +1,94 @@
+{
+    "mime_type": "application/json/ons/eq",
+    "questionnaire_id": "23",
+    "schema_version": "0.0.1",
+    "survey_id": "023",
+    "title": "Census Theme",
+    "description": "",
+    "theme": "census",
+    "introduction": {
+        "description": "",
+        "information_to_provide": []
+    },
+    "display": {
+        "properties": {}
+    },
+    "eq_id": "1",
+    "groups": [
+        {
+            "blocks": [
+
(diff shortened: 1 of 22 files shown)


app/main/views/questionnaire.py
```diff
@@ -58,6 +58,9 @@ def post_questionnaire(eq_id, form_type, period_id, collection_id, location):
         return render_page(location, False)
 
     next_location = g.questionnaire_manager.get_current_location()
+    if next_location == location:
```
benjefferies, line 61: If we created another end point for adding another composite field we wouldn't have to do this.

As discussed with @jonnyshaw89 it would also mean we could handle the logic explicitly there, therefore we wouldn't need to depend on the user action pattern which means we are in a better position for removing it

app/data/census_household_composition.json
```diff
@@ -0,0 +1,100 @@
...
+                                  "description": "<br> <div> <h3>Include:</h3> <ul> <li>Yourself, if this is your permanent or family home </li> <li>Family members including partners, children and babies born on or before 9 April 2017</li> <li>Students and/or school children who live away from home during term time</li> <li>Housemates, tenants or lodgers</li> </ul> </div>",
+                                  "type": "Household",
+                                  "validation": [],
+                                  "answers": [
+                                    {
+                                        "id": "person",
+                                        "type": "Composite",
+                                        "q_code": "9999",
+                                        "mandatory": false,
+                                        "answers": [
```
benjefferies, line 49: Maybe we should call these something else from answers, maybe something like parts or something?

app/questionnaire/user_action_processor.py
```diff
@@ -73,6 +75,28 @@ def perform_action(self):
         pass
 
 
+class AddAnotherPerson(UserAction):
```
benjefferies, line 78: As above, could we do this as endpoints instead of actions

samiwel:
I noticed whilst looking at implementing the validation on `CompositeAnswer` that the `state.input` value does not lend itself very well to be passed through to each of its child answers.

The way I originally envisioned implementing validation for `CompositeAnswer` was in terms of its children, e.g.

```
    def validate(self, state):
        is_valid = True
        for answer in self.answers:
            is_valid &= answer.validate(state)

        return is_valid
```

but at the point validation gets called, the state.input contains data of the wrong type to pass through to its children. i.e. `state.input` for `CompositeAnswer` is a `dict`, but each child answer expects a value. I could perform check in `Answer`, but perhaps a better solution would be to re-visit how `state.input` is built for `CompositeAnswer`. Anyone have any opinions/ideas about this?

app/main/views/questionnaire.py
samiwel: Thanks @benjefferies  and @jonnyshaw89. I guess this makes sense if we're planning on getting rid of the User Action pattern altogether. Some actions are never used, so perhaps dedicated end points for actions, possibly an actions Blueprint instead?

app/data/census_household_composition.json
LJBabbage: Agree, but I would prefer not to use a generic term like parts, maybe something like composite_answers

app/parser/v0_0_1/schema_parser.py
```diff
@@ -296,7 +309,7 @@ def _parse_answer(self, schema, questionnaire):
             answer_id = ParserUtils.get_required_string(schema, 'id')
             answer = self.answer_factory.create(answer_type.upper(), answer_id)
             answer.type = answer_type
-            answer.code = ParserUtils.get_required_string(schema, 'q_code')
+            answer.codecode = ParserUtils.get_required_string(schema, 'q_code')
```
LJBabbage: ?

app/data/census_household_composition.json
samiwel: Yeah, I agree. Answers within answers can become confusing. We just need to agree on nomenclature that makes sense. As long as we don't lose sight that what is actually inside `CompositeAnswer` are, in fact, `Answer`s.

ajmaddaford:
Outcomes of the spike have been documented in Confluence, @samiwel is starting work on a new  branch from master for this feature.

PR #539, opened 2016-11-03
Update JWT token to accept region_code

LJBabbage:
To include region_code in the JWT and store the result in metadata["region_code"]. It has been agreed that it will be optional to begin with and then mandatory at a later date. If it is not passed in the JWT it will be stored as None. Surveys which are currently inflight, which don't have region_codes have been protected against and will still work.  
This functionality will be used in the future so front end tests are not possible at this time.

### How to review 
We currently don't use this functionality so the best way is to add {{ meta.survey.region_code}} somewhere on the landing page(can be any page) and use /dev to make sure it changes when the dev page region code drop down changes.

### DIFF
--- a/app/templating/metadata_template_preprocessor.py
+++ b/app/templating/metadata_template_preprocessor.py
@@ -55,6 +55,7 @@ class MetaDataTemplatePreprocessor(object):
             "start_date":  self._format_date(metadata["ref_p_start_date"]),
             "end_date": self._format_date(metadata["ref_p_end_date"]),
             "employment_date": self._format_date(metadata["employment_date"]),
+            "region_code": metadata["region_code"] if 'region_code' in metadata else None,
             "period_str": metadata["period_str"],
         }
         return survey_meta
--- a/tests/integration/create_token.py
+++ b/tests/integration/create_token.py
@@ -13,7 +13,8 @@ RETURN_BY = "2016-05-06"
 TRAD_AS = "Integration Tests"
 EMPLOYMENT_P_DATE = "1983-06-02"
 
-def create_token(form_type_id, eq_id, start_date=None, end_date=None, employment_date=None):
+
+def create_token(form_type_id, eq_id, start_date=None, end_date=None, employment_date=None, region_code=None):
         user = USER
         exp_time = 3600                         # one hour from now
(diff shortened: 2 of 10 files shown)


LJBabbage:
In WIP till https://github.com/ONSdigital/ons-schema-definitions/pull/11 is merged

app/templating/metadata_template_preprocessor.py
```diff
@@ -55,6 +55,7 @@ def _build_survey_meta(self, questionnaire_schema):
             "start_date":  self._format_date(metadata["ref_p_start_date"]),
             "end_date": self._format_date(metadata["ref_p_end_date"]),
             "employment_date": self._format_date(metadata["employment_date"]),
+            "region_code": metadata["region_code"] if 'region_code' in metadata else None,
```
benjefferies, line 58: Would this not fail with KeyError if region_code wasn't in metadata or is that a misunderstanding?

tests/integration/create_token.py
```diff
@@ -37,13 +38,18 @@ def create_token(form_type_id, eq_id, start_date=None, end_date=None, employment
         else:
             ref_p_employment_date = employment_date
 
+        if region_code is None:
```
benjefferies: We could tidy the conditional statements up and put the defaults as the constants possibly

benjefferies (commented):
Can we not merge this in before https://github.com/ONSdigital/ons-schema-definitions/pull/11/files as it's only documentation updates?

tests/integration/create_token.py
LJBabbage: This is actually a mistake and should have been taken out, there should be no default. It was agreed it should be None if a value isn't provided

pricem14pc:
Ok from us Lewis. We will  be releasing our changes to introduce regions_code into our EQ JWT in release 5.5.3 currently planned for w/c 21 Nov

app/templating/metadata_template_preprocessor.py
LJBabbage: As discussed this is the standard format for an if else on one line in python

ajmaddaford (approved):

Issue #538, opened 2016-11-03
Validation error shows server error page when validation errors on page with skip condition question


benjefferies:
### Expected behaviour
Validation errors relating to questions should appear at top of page

### Actual behaviour
Server error page is displayed

### Steps to reproduce the behaviour

1. Start star wars survey
1. Choose Leyoda, lightside
1. Complete page 2
1. Complete page 3 selecting `Awesome, I love them all` to question `What do you think of the prequel series?`
1. Try to `Save and continue` without selecting an answer

### Technical information
Error occurs at https://github.com/ONSdigital/eq-survey-runner/blob/master/app/questionnaire/questionnaire_manager.py#L322. The reason is because we cannot use __dict__ on a dict. Initially rule['when'] is an object but we use __dict__ to get the dict of the attributes but we also reassign rule['when'] to a dict therefore when we try to convert this to a dict using __dict__ the second time it fails as it is already a dict.

### Screenshot
![image](https://cloud.githubusercontent.com/assets/15050019/19962809/4a48ef74-a1b1-11e6-816f-4d876ac711a4.png)

collisdigital:
Fixed by #540

yoganandkunche:
Defect passed retest in pre-prod

PR #540, opened 2016-11-03
Fixing dict conversion bug

benjefferies:

When posting invalid answers that contain questions with skip_conditions
state modification causes a error to be raised which redirects the user
to a server error page when they should see the errors.
We now create a deep copy of the object so when we change state it
doesn't affect the state item.

### How to review 
See steps to replicated in #538

### DIFF
--- a/app/schema/skip_condition.py
+++ b/app/schema/skip_condition.py
@@ -1,5 +1,11 @@
+import copy
 
 
 class SkipCondition(object):
     def __init__(self):
         self.when = None
+
+    def as_dict(self):
+        self_copy = copy.deepcopy(self.__dict__)
+        self_copy['when'] = self_copy['when'].__dict__ if self.when else None
+        return self_copy
--- a/app/data/test_skip_condition.json
+++ b/app/data/test_skip_condition.json
@@ -1,0 +1,184 @@
+{
+   "mime_type":"application/json/ons/eq",
+   "questionnaire_id":"0",
+   "schema_version":"0.0.1",
+   "survey_id":"0",
+   "title":"Test skip condition",
+   "theme":"default",
+   "description":"A questionnaire to test skip condition.",
+   "introduction":{
+      "description":"",
+      "information_to_provide":[
+         "Breakfast foods",
+         "Other info"
+      ]
+   },
+   "display":{
+      "properties":{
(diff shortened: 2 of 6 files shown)


app/parser/v0_0_1/schema_parser.py
```diff
@@ -251,7 +248,7 @@ def _parse_question(self, schema, questionnaire):
             question.id = ParserUtils.get_required_string(schema, "id")
             question.title = ParserUtils.get_required_string(schema, "title")
             question.description = ParserUtils.get_required_string(schema, "description")
-            question.skip_condition = self._parse_skip_condition(ParserUtils.get_optional(schema, "skip_condition"))
+            question.skip_condition = schema.get("skip_condition")
```
ajmaddaford: I don't think we should be changing how one property of the schema object works, this would make it inconsistent with every other property. Is it possible to revert this and add an as_dict method to `schema/skip_condition.py`?

ajmaddaford (changes requested):

app/parser/v0_0_1/schema_parser.py
benjefferies: I'd agree it's against the grain of how the schema is now but knowing that we are going to be removing the schema model imminently to use a dict anyway is it worth it also it's only being used in that one place?

app/schema/skip_condition.py
```diff
@@ -1,5 +1,11 @@
+import copy
 
 
 class SkipCondition(object):
     def __init__(self):
         self.when = None
+
+    def copy_as_dict(self):
```
ajmaddaford: Do we need to know it's a copy, or can we just call the method `as_dict`?

ajmaddaford (changes requested):

app/schema/skip_condition.py
benjefferies: I would question why they weren't using .**dict** if I saw it that's all

jonnyshaw89: Why does it need to be a copy? Are we modifying it? if so, should we be?

jonnyshaw89 (approved):

ajmaddaford (approved):

PR #542, opened 2016-11-04
Remove period id from URL and multiple survey checks

LJBabbage:
Removal of the period ID from the URL and an update to multiple surveys check

### How to review 
1. Open any survey from /dev and look at the URL, make sure the url no longer has the period id. Make sure all the tests work
2. Make sure the multiple survey check still works (see how to recreate https://github.com/ONSdigital/eq-survey-runner/issues/379)


samiwel (approved):
These all seem like sensible modifications to me.
`period_id` value doesn't appear to be used anywhere so makes sense to remove it.

ajmaddaford (approved):

Issue #524, opened 2016-10-26
Get metadata can throw a RuntimeException during error handling causing stacktrace to be displayed in browser


benjefferies:
### Expected behaviour

When an error occurs for a user with no metadata in the application we should see a error page with an error code
### Actual behaviour

When an error occurs for a user with no metadata in the application we should see a stacktrace
### Steps to reproduce the behaviour
- Start questionnaire
- Invalidate metadata (Unsure how to do this)
- Refresh page
### Technical information

We throw a runtime exception when there is no meta data here https://github.com/ONSdigital/eq-survey-runner/blob/master/app/data_model/questionnaire_store.py#L37
#### Browser

Chrome
53.0.2785.143
#### Operating System

OS X 10.11.6
### Screenshot

[RuntimeError_ ('No metadata for user %s', '5a4ede5dea984997fe13bf1138ebaff4addcf389e1d5d918bbec411b7db5a0aa') __ Werkzeug Debugger.pdf](https://github.com/ONSdigital/eq-survey-runner/files/553755/RuntimeError_.No.metadata.for.user.s.5a4ede5dea984997fe13bf1138ebaff4addcf389e1d5d918bbec411b7db5a0aa.__.Werkzeug.Debugger.pdf)

collisdigital:
We believe this to be caused by having two different browsers/users open against the same session (ru_ref); we plan to prevent this scenario soon; will review this defect when that happens.

jonnyshaw89:
Also we appear not to be catching exceptions properly and displaying 500 error page.

ajmaddaford:
If we find there is no metadata then we should redirect to /session-expired. Card added to Trello https://trello.com/c/aMZBGZy2/1411-log-user-out-when-no-metadata

PR #543, opened 2016-11-04
Removing plumber

benjefferies:

[trello](https://trello.com/c/au4LlbSV/706-5-replacing-plumber-with-jinja-templating)

There were two strategies for resolving variable substitution in the
application causing overheads in maintenance.

* Using python string substitution
* Using jinja templating engine

The only templating strategy implemented now is using jinja.

### How to review 
We shouldn't be hand rolling our own templating.
Jinja should be used for templating
Tests should exist for templating filters


app/questionnaire/questionnaire_manager.py
```diff
@@ -85,14 +85,15 @@ def get_rendering_context(self, location, is_valid=True):
...
         if self.state:
-            self._plumbing_preprocessing()
+            context = self._get_schema_template_context()
+            self._plumb(context, self.state)
             self._conditional_display(self.state)
 
         # look up the preprocessor and then build the view data
         return build_questionnaire_model(self._json, self.state)
 
     def get_summary_rendering_context(self):
```
benjefferies: Considering moving all of the rendering context thus plumbing related logic out into it's own file. Apart from that the `piping` has been stripped back to only what is required

app/main/views/questionnaire.py
```diff
@@ -115,26 +125,37 @@ def redirect_to_questionnaire_page(eq_id, form_type, collection_id, location):
...
+def _same_survey(eq_id, form_type, collection_id):
     metadata = get_metadata(current_user)
     current_survey = eq_id + form_type + collection_id
     metadata_survey = metadata["eq_id"] + metadata["form_type"] + metadata["collection_exercise_sid"]
     return current_survey == metadata_survey
 
 
-def render_page(location, is_valid):
-    context = g.questionnaire_manager.get_rendering_context(location, is_valid)
+def _render_page(location, context):
```
ajmaddaford: I think we should remove the TemplateRegistry. Some routes know what template they should use (summary and thank-you), confirmation isn't used, and introduction should probably be it's own route.

app/main/views/questionnaire.py
```diff
@@ -70,39 +78,41 @@ def get_summary(eq_id, form_type, collection_id):
     navigator = g.questionnaire_manager.navigator
     latest_location = navigator.get_latest_location(get_answers(current_user), get_completed_blocks(current_user))
     if latest_location is 'summary':
-        return render_page('summary', True)
+        metadata = get_metadata(current_user)
+        answers = get_answers(current_user)
+        schema_json = _render_schema(g.schema_json, answers, metadata)
```
jonnyshaw89, line 92: These lines are basically the same as in _questionnaire_rendering_context()
Could we combine these somehow?

app/templating/metadata_context.py
```diff
@@ -0,0 +1,76 @@
...
+        respondent_id = metadata["ru_ref"]
+        name = metadata["ru_name"]
+        trading_as = metadata["trad_as"]
+    else:
+        respondent_id = None
+        name = None
+        trading_as = None
+
+    respondent_meta = {
+        "tx_id": convert_tx_id(metadata["tx_id"]),
```
ajmaddaford, line 32: What if metadata is None?

app/main/views/questionnaire.py
```diff
@@ -30,11 +36,11 @@
 @questionnaire_blueprint.before_request
 @login_required
 def check_survey_state():
-    json, schema = get_schema()
-    g.questionnaire_manager = QuestionnaireManager(schema, json=json)
+    g.schema_json, g.schema = get_schema()
```
iwootten, line 40: Have we determined a way to bind at an application level, rather than per request in order to cache the schema?

app/main/views/questionnaire.py
```diff
@@ -47,15 +53,17 @@ def add_cache_control(response):
 @questionnaire_blueprint.route('<location>', methods=["GET"])
 @login_required
 def get_questionnaire(eq_id, form_type, collection_id, location):
-    return render_page(location, True)
+    g.questionnaire_manager.build_state(location, get_answers(current_user))
```
iwootten: I kind of feel we shouldn't be accessing flasks globals in this way. Flask has pretty clear guidance on using the app context http://flask.pocoo.org/docs/0.11/appcontext/. 

I'm aware this has crept in elsewhere, but I think we should be creating accessors as per the questionnaire store approach in globals.py.

app/main/views/questionnaire.py
collisdigital: For info, in the Alpha when using Cassandra the connection pool was setup once at startup; this was done outside of any Flask mechanism as it turned out the request and application contexts in Flask are all per-request so aren't cached between requests. There may be a more 'Flask' way of doing it however, the PR related to this is here: https://github.com/ONSdigital/alpha-eq-survey-runner/pull/54

ajmaddaford: We definitely need to do this, but not in this pull request

app/templating/metadata_context.py
```diff
@@ -0,0 +1,76 @@
...
+def _build_survey_meta(questionnaire_schema, metadata):
+    introduction = questionnaire_schema['introduction']
+
+    survey_meta = {
+        "title": questionnaire_schema['title'],
+        "survey_code": questionnaire_schema['survey_id'],
+        "description": _get_description(introduction),
+        "information_to_provide": _get_info_to_provide(introduction),
+        "theme": questionnaire_schema['theme'],
+        "return_by": _format_date(metadata["return_by"]),
```
ajmaddaford: We shouldn't be formatting dates here, it should be done in the template. This will be important when we need to output dates in Welsh.

app/templating/template_context.py
```diff
@@ -0,0 +1,85 @@
+import logging
```
ajmaddaford: template_context is very generic - would it be worth having `questionnaire_context` and `summary_context`?

app/templating/template_context.py
```diff
@@ -0,0 +1,85 @@
...
+    """
+    return {
+      "exercise": _build_exercise(metadata),
+      "answers": _map_alias_to_answers(aliases, answers),
+    }
+
+
+def _build_rendering_context(meta, content):
+    return {
+      "meta": meta,
```
ajmaddaford: I don't think we should be returning meta and content from this method, I think we should just return the content. It would make testing this class a lot simpler (no need to mock metadata) and the metadata is always treated separately to the content (we explicitly pull it out of this dict to pass it to the template).

app/main/views/questionnaire.py
benjefferies: From what I can see there is no 'flask' way of sharing state between requests. This would leave us with a similar option to where @collisdigital has pointed us where we manage our own global state. There are a couple of things we would have to consider here as the schema_json is modified. Mutability and multithreading isn't fun

ajmaddaford (changes requested):

app/main/views/questionnaire.py
benjefferies: Fair enough, boy scout rule has convinced me to pick this one up. I'll create an accessor for the questionnaire_manager

app/templating/metadata_context.py
benjefferies: At this moment in time can't be. It raises an error if it doesn't exist

app/templating/template_context.py
benjefferies: Good point will do this

app/templating/template_context.py
benjefferies: Fair point, should of clocked on but had tunnel vision trying to get it in!

app/templating/metadata_context.py
benjefferies: Shall we do this now or raise a ticket to action after?

app/main/views/questionnaire.py
```diff
@@ -115,26 +125,37 @@ def redirect_to_questionnaire_page(eq_id, form_type, collection_id, location):
...
 
 
-def render_page(location, is_valid):
-    context = g.questionnaire_manager.get_rendering_context(location, is_valid)
+def _render_page(location, context):
     template = TemplateRegistry.get_template_name(location)
     navigator = g.questionnaire_manager.navigator
     previous_location = navigator.get_previous_location(get_answers(current_user), location)
-    return render_template(template, context, previous_location)
+    return _render_template(template, context, previous_location)
```
ajmaddaford: I think we need to rationalise these methods `_render_page > render_template > render_theme_template`, each one is returning a call to another method. How about we refactor `_render_template` into a method to get the theme, and call `render_theme_template` from this method?

app/templating/metadata_context.py
```diff
@@ -0,0 +1,76 @@
+import logging
+
+from app.libs.utils import convert_tx_id
+from app.utilities.date_utils import to_date
+
+logger = logging.getLogger(__name__)
+
+
+def build_metadata(schema_json, metadata):
```
ajmaddaford: I don't think we should be passing schema_json to this method, it looks like it would be worth introducing an introduction context rather than overload the metadata context with it.

ajmaddaford (changes requested):

app/templating/metadata_context.py
benjefferies: Metadata can't be None at this point

app/templating/metadata_context.py
benjefferies: Done

app/templating/template_context.py
benjefferies: Done

app/templating/template_context.py
benjefferies: Done

app/main/views/questionnaire.py

app/templating/metadata_context.py
jonnyshaw89: Can it be?

collisdigital: Related? #524

app/main/views/questionnaire.py
benjefferies: Will look at the options, the trouble is the context is different

app/main/views/questionnaire.py

app/templating/metadata_context.py
benjefferies: Done

app/questionnaire/questionnaire_manager.py
```diff
@@ -1,19 +1,27 @@
...
 
-from flask import render_template_string
+from flask import g
 
 from flask_login import current_user
 
 logger = logging.getLogger(__name__)
 
 
+def get_questionnaire_manager(schema, schema_json):
```
benjefferies, line 15: @iwootten I would have put this in globals but it introduces a circular dependency. Also at the moment the globals contains questionnaire store related stuff which is nice without this polluting it however not sure where else to put this. The other option is it can live in its own file

app/main/views/questionnaire.py
```diff
@@ -115,26 +135,33 @@ def redirect_to_questionnaire_page(eq_id, form_type, collection_id, location):
...
-def render_page(location, is_valid):
-    context = g.questionnaire_manager.get_rendering_context(location, is_valid)
-    template = TemplateRegistry.get_template_name(location)
-    navigator = g.questionnaire_manager.navigator
-    previous_location = navigator.get_previous_location(get_answers(current_user), location)
-    return render_template(template, context, previous_location)
-
-
-def render_template(template, context, previous_location):
+def _render_template(location, context, rendered_schema_json=None, page=None):
```
ajmaddaford: - I think the `page` parameter should be renamed `template` and explicitly pass in the filename 'questionnaire.html'
- `rendered_schema_json` doesn't appear to be passed in by anything

app/questionnaire/questionnaire_manager.py
```diff
@@ -79,36 +84,17 @@ def process_incoming_answers(self, location, post_data):
...
         # Build the state from the answers
         self.state = None
         if self._schema.item_exists(item_id):
             schema_item = self._schema.get_item_by_id(item_id)
             self.state = schema_item.construct_state()
             self.state.update_state(answers)
+            self._conditional_display(self.state)
+        if self.state:
+            context = build_schema_context(get_metadata(current_user), self._schema.aliases, answers)
+            self.renderer.render_state_item(self.state, context)
```
ajmaddaford: I think we should rename `render_state_item` to `render_state`, we don't refer to state items anywhere else.

ajmaddaford (changes requested):

app/main/views/questionnaire.py

app/questionnaire/questionnaire_manager.py
benjefferies: Done

ajmaddaford (approved):

Issue #492, opened 2016-10-13
Two same surveys for different months when launched contains 'data' even when Get Started is clicked for the first time.


yoganandkunche:
### Expected behaviour

So data/errors exists for these surveys even though they are just started.

**_Note**_ Issue is intermittent.
### Actual behaviour

Both the surveys must not contain any previously saved data.
### Steps to reproduce the behaviour

Launch two surveys of the same type simultaneously (each for different period (month))
Click on [Get Started] for both the survey's

collisdigital:
Has anyone seen this recently? Suggest we close this if no reccurrance in the last month?

benjefferies:
I haven't noticed the test fail

jonnyshaw89:
I see no harm in closing and re-opening if it occurs again

collisdigital:
Closing as not seen recently.

PR #553, opened 2016-11-09
Add Selenium test to simulate two months being worked on at the same …

LJBabbage:
A test for #492 to see if we can recreate the problem after the refactoring of survey runner. Two different months being worked on at the same time using the same form type 

### How to review 
Make sure all the frontend tests work correctly "yarn test"

### DIFF
--- a/tests/functional/spec/same-survey-diff-month.spec.js
+++ b/tests/functional/spec/same-survey-diff-month.spec.js
@@ -1,0 +1,76 @@
+import assert from 'assert'
+import chai from 'chai'
+import {getRandomString, startQuestionnaire} from '../helpers'
+import reportingPeriod from '../pages/surveys/rsi/0102/reporting-period.page'
+import retailTurnoverPage from '../pages/surveys/rsi/0102/retail-turnover.page'
+import internetSalesPage from '../pages/surveys/rsi/0102/internet-sales.page'
+import changeInRetailTurnover from '../pages/surveys/rsi/0102/changes-in-retail-turnover.page'
+import rsiSummaryPage from '../pages/surveys/rsi/0102/rsi-summary.page'
+import devPage from '../pages/dev.page'
+import landingPage from '../pages/landing.page'
+import multipleSurveys from '../pages/multiple-survey.page'
+
+const expect = chai.expect
+
+describe('RSI - Multiple months of one form type being worked on at the same time ', function() {
+
+    it('Given the RSI survey 0102 when a second surveys is open for a different month then the first survey shows the multiple survey page and the 2nd survey continues to work and has the correct data', function() {
+
+        //Given that RSI business survey 0102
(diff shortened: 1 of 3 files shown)


tests/functional/spec/same-survey-diff-month.spec.js
```diff
@@ -0,0 +1,73 @@
...
+const expect = chai.expect
+
+describe('RSI - Multiple months of one form type being worked on at the same time ', function() {
+
+    it('Given the RSI survey 0102 when a second surveys is open for a different month then the first survey shows the multiple survey page and the 2nd survey continues to work and has the correct data', function() {
+
+        //Given that RSI business survey 0102
+        devPage.open()
+          .setUserId(getRandomString(10))
+          .setCollectionId(getRandomString(3))
```
collisdigital: Possibility that both launches will contain the same values because they are random; however unlikely it could cause a false positive. What about prefixing each with something unique e.g. "A" + getRandom...

tests/functional/spec/same-survey-diff-month.spec.js
```diff
@@ -0,0 +1,73 @@
...
+
+const expect = chai.expect
+
+describe('RSI - Multiple months of one form type being worked on at the same time ', function() {
+
+    it('Given the RSI survey 0102 when a second surveys is open for a different month then the first survey shows the multiple survey page and the 2nd survey continues to work and has the correct data', function() {
+
+        //Given that RSI business survey 0102
+        devPage.open()
+          .setUserId(getRandomString(10))
```
collisdigital: I thought the issue was the same user id doing two periods; not different ones? Not sure it matters though?

LJBabbage: Would make it a more realistic test, has been updated

tests/functional/spec/same-survey-diff-month.spec.js
LJBabbage: Also an improvement

tests/functional/spec/same-survey-diff-month.spec.js
```diff
@@ -0,0 +1,76 @@
...
+        //Then the 1st survey shows the multiple survey page and the 2nd survey continues to works and has the correct data
+        browser.switchTab()
+        reportingPeriod.submit()
+        expect(multipleSurveys.isOpen()).to.be.true
+
+        browser.switchTab('second_survey')
+
+        retailTurnoverPage.setRetailTurnover(12345)
+        retailTurnoverPage.submit()
+        internetSalesPage.setInternetSales(1234)
```
benjefferies: Do we need to complete the survey or just submit the page and assert that the internet sales page is open?

benjefferies (approved):
Is it worth squashing down commits and this needs a rebase against master

ajmaddaford (approved):

PR #556, opened 2016-11-10
Fixing SQLAlchemy transaction handling

benjefferies:

We saw a error in production which suggested a transaction had already
been closed `sqlalchemy.exc.InvalidRequestError: This Session's
transaction has been rolled back due to a previous exception during
flush`. On further investigation this was because of a duplicate
questionnaire PK when committing a transaction. Further down the line
the application tried to make use of the session which due to the error
meant we had no transaction to use as it was left in a dirty state
without being rolled back. This caused a stacktrace to be displayed on
simulation.

* Removing db_session when application dies and at the end of a request
  as per [flask documentation]
  (http://flask.pocoo.org/docs/0.11/patterns/sqlalchemy/#declarative).
* Catching errors on commit and rolling back then re-raising error.
* Removing unused interface
* Logging tx_id against errors

Further reading: On investigating different transaction handling
strategies one suggestion which seems quite applicable in our scenario
was to map the transaction to the http request which makes sense when
following the `unit of work` pattern. The benefits of this would be we
wouldn't have to manually handle our transactions for all of the
database code and we'd have one place to handle it thus one area to test
regarding transactions.

### How to review 

#### Simulating duplicate PK (developer)

1. Add threaded configuration as `Server` parameter at https://github.com/ONSdigital/eq-survey-runner/blob/master/application.py#L16. See https://arusahni.net/blog/2013/10/flask-multithreading.html
1. Run application in debug with a breakpoint at https://github.com/ONSdigital/eq-survey-runner/blob/eq-494-fix-database-transaction-handling/app/storage/database_storage.py#L24
1. Start a new survey as a new unique user
1. You should hit the break point (Do not resume it yet)
1. Open another tab and start another survey as the same user
1. Resume both of the threads
1. Observe 500 error instead of stacktrace

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -153,7 +153,7 @@ def create_app(config_name):
     Themes(application, app_identifier="surveyrunner")
 
     @application.teardown_appcontext
-    def shutdown_session(exception=None):
+    def shutdown_session(exception=None):    # pylint: disable=unused-argument
         db_session.remove()
 
     return application
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -153,7 +153,7 @@ def create_app(config_name):
     Themes(application, app_identifier="surveyrunner")
 
     @application.teardown_appcontext
-    def shutdown_session(exception=None):    # pylint: disable=unused-argument
+    def shutdown_session(exception=None):    # pylint: disable=unused-variable
         db_session.remove()
 
     return application
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -10,6 +10,7 @@ from app import settings
 from app.analytics.custom_google_analytics import CustomGoogleAnalytics
 from app.authentication.authenticator import Authenticator
 from app.authentication.cookie_session import SHA256SecureCookieSessionInterface
+from app.data_model.database import db_session
 from app.libs.utils import get_locale
 from app.submitter.submitter import SubmitterFactory
 
@@ -150,6 +151,10 @@ def create_app(config_name):
     # Add theme manager
     application.config['THEME_PATHS'] = os.path.dirname(os.path.abspath(__file__))
     Themes(application, app_identifier="surveyrunner")
+
+    @application.teardown_appcontext
+    def shutdown_session(exception=None):
+        db_session.remove()
 
     return application
 
--- a/app/authentication/session_management.py
+++ b/app/authentication/session_management.py
@@ -1,12 +1,10 @@
 import logging
 from uuid import uuid4
 
-from app.data_model.database import EQSession
+from app.data_model.database import EQSession, commit_or_rollback
 from app.data_model.database import db_session
 
 from flask import session
-
-from sqlalchemy.exc import SQLAlchemyError
 
 USER_ID = "user_id"
 USER_IK = "user_ik"
@@ -39,15 +37,10 @@ class DatabaseSessionManager:
             logger.debug("Found eq_session_id %s in session", eq_session_id)
             eq_session = self._get_object(eq_session_id)
             logger.debug("Loaded object eq session %s", eq_session)
-            logger.debug("About to commit to database")
 
-        try:
+        logger.debug("About to commit to database")
+        with commit_or_rollback(db_session):
             db_session.add(eq_session)
-            db_session.commit()
-            logger.debug("Committed")
-        except SQLAlchemyError:
-            db_session.rollback()
-            raise
 
     def has_user_id(self):
         """
@@ -72,12 +65,8 @@ class DatabaseSessionManager:
             eq_session = self._get_object(eq_session_id)
             logger.debug("About to delete entry from eq_session table %s", eq_session)
 
-            try:
+            with commit_or_rollback(db_session):
                 db_session.delete(eq_session)
-                db_session.commit()
-            except SQLAlchemyError:
-                db_session.rollback()
-                raise
         else:
             logger.warning("No eq session id exists")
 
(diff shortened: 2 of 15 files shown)


app/authentication/session_management.py
```diff
@@ -84,12 +39,21 @@ def store_user_id(self, user_id):
             logger.debug("Found eq_session_id %s in session", eq_session_id)
             eq_session = self._get_object(eq_session_id)
             logger.debug("Loaded object eq session %s", eq_session)
-        logger.debug("About to commit to database")
-        db_session.add(eq_session)
-        db_session.commit()
-        logger.debug("Committed")
+            logger.debug("About to commit to database")
+
+        try:
```
collisdigital: These repeated patterns below could benefit from a "with" context manager utility e.g.

``` python
@contextmanager
def db_committed_or_rolledback(db_session):
    try:
        yield db_session
        db_session.commit()
        logger.debug("Committed")
    except SQLAlchemyError:
        db_session.rollback()
        raise

...

with db_committed_or_rolledback(db_session):
    db_session.add(eq_session)
```

or similar...

collisdigital: See http://preshing.com/20110920/the-python-with-statement-by-example/ for examples

collisdigital:
Do your steps to reproduce suggest that in a scenario when a user launches the survey and for example hits refresh before the page loads (or opens two tabs etc.) we may get two identical requests and the second will fail with a 500 error because of a race condition of which request creates the session in the db first? If so it is worth discussing whether a 500 error is appropriate here or if a more graceful recovery (such as then looking up the duplicate PK and returning the just created session?)

app/authentication/session_management.py
benjefferies: Definitely a better way of doing it.

The other side of this is that we could start a transaction at the beginning of the flask request and commit/rollback at the end but I guess it depends what we class as a unit of work. At the moment there is only one place where we do more than one DML action on login.

benjefferies:
Yes that is the case at this moment in time (better than the previous which was a stacktrace) but we can talk about how we want to handle this scenario. Loading the current session would probably be sensible

jonnyshaw89:
We should probably question if we should ever throw a 500 error?
If we lost the database and the web servers started returning 500 errors, then they will get removed from the Load Balancer?
Maybe we need a static html error page to show in worst cases?

app/__init__.py
```diff
@@ -151,6 +152,10 @@ def override_url_for():
     application.config['THEME_PATHS'] = os.path.dirname(os.path.abspath(__file__))
     Themes(application, app_identifier="surveyrunner")
 
+    @application.teardown_appcontext
+    def shutdown_session(exception=None):
```
collisdigital: May need a pylint exception (https://pylint.readthedocs.io/en/latest/user_guide/message-control.html)

benjefferies: Thanks! I was just looking for this

tests/app/authentication/test_database_session_manager.py
```diff
@@ -0,0 +1,47 @@
...
+
+from mock import patch, Mock, MagicMock
+from sqlalchemy.exc import IntegrityError
+
+from app.authentication.session_management import DatabaseSessionManager, EQ_SESSION_ID
+
+
+class TestDatabaseSessionManager(unittest.TestCase):
+
+    def test_store_user_id_rollback(self):
```
benjefferies: Not sure what the preference is here for test methods? Do we make all test methods static or functions?

I guess grouping tests within a class is quite useful in some circumstances

app/authentication/session_management.py
```diff
@@ -18,60 +18,13 @@ class NoUniqueIDError(RuntimeError):
     pass
 
 
-class SessionManagement(object):
+class DatabaseSessionManager:
```
ajmaddaford: I think as there is only one type of session manager now, I think we should rename this class `SessionManager` (the file should also be renamed `session_manager.py`).

app/__init__.py
benjefferies: # pylint: disable=unused-argument didn't work on codacy

collisdigital: it's not an unused argument though, unused-variable ?

collisdigital: https://docs.pylint.org/en/1.6.0/features.html 

unused-argument (W0613):
    Unused argument %r Used when a function or method argument is not used.
unused-variable (W0612):
    Unused variable %r Used when a variable is defined but not used.

collisdigital: May also need to be on the end of the line?

collisdigital: Also I clicked "Ignore" in the GUI to see if it would show a snippet of how to do it; but it didn't (it just hid the issue) and now can't un-ignore it, sorry!

benjefferies: It should appear on next build again. I copied what was on pylint which you'd assume is correct?

tests/app/authentication/test_database_session_manager.py
```diff
@@ -0,0 +1,49 @@
...
+class TestDatabaseSessionManager(unittest.TestCase):
+
+    @staticmethod
+    def test_store_user_id_rollback():
+        # Given
+        session_manager = DatabaseSessionManager()
+        user_id = "1"
+
+        with patch('app.authentication.session_management.session'), \
+                patch('app.authentication.session_management.db_session') as db_session:
```
ajmaddaford: We should add `autospec=True` to the creation of the mock object here to ensure the assertions we make areactually testing what we are expeting them to (see https://engineeringblog.yelp.com/2015/02/assert_called_once-threat-or-menace.html)

ajmaddaford (changes requested):

ajmaddaford (approved):

PR #557, opened 2016-11-11
Added ability to add interstitial pages in to questionnaire

jonnyshaw89:
The PR adds the functionality to be able to add interstitial pages into the questionnaire

### How to review 
Launch test survey test_interstitial_page.json.
When completing the survey, an interstitial page should be shown between the questions.

### DIFF
--- a/app/data/test_interstitial_page.json
+++ b/app/data/test_interstitial_page.json
@@ -1,0 +1,157 @@
+{
+    "mime_type": "application/json/ons/eq",
+    "questionnaire_id": "0",
+    "schema_version": "0.0.1",
+    "survey_id": "0",
+    "title": "Interstitial Pages",
+    "theme": "default",
+    "description": "A questionnaire to demo interstitial pages.",
+    "introduction": {
+        "description": "",
+        "information_to_provide": []
+    },
+    "display": {
+        "properties": {}
+    },
+    "eq_id": "0",
+    "messages": {
+        "INTEGER_TOO_LARGE": "Number is too large",
+        "NEGATIVE_INTEGER": "Number cannot be less than zero",
+        "NOT_INTEGER": "Please enter an integer"
+    },
+    "groups": [
+        {
+            "blocks": [
+                {
+                    "type": "questionnaire",
+                    "display": {
+                        "properties": {}
+                    },
+                    "id": "breakfast-block",
+                    "sections": [
+                        {
+                            "description": "",
+                            "display": {
+                                "properties": {}
+                            },
+                            "id": "breakfast-section",
+                            "questions": [
+                                {
+                                    "answers": [
+                                        {
+                                            "display": {},
+                                            "guidance": "",
+                                            "id": "favourite-breakfast",
+                                            "label": "What is your favourite breakfast food",
+                                            "mandatory": false,
+                                            "options": [],
+                                            "q_code": "0",
+                                            "type": "TextField"
+                                        }
+                                    ],
+                                    "description": "",
+                                    "display": {
+                                        "properties": {}
+                                    },
+                                    "id": "favourite-breakfast-question",
+                                    "title": "",
+                                    "type": "General",
+                                    "validation": []
+                                }
+                            ],
+                            "title": "What is your favourite breakfast food",
+                            "validation": []
+                        }
+                    ],
+                    "routing_rules": [
+                    ],
+                    "title": "",
+                    "validation": []
+                },
+                {
+                    "type": "interstitial",
+                    "id": "breakfast-interstitial",
+                    "title": "Breakfast Interstitial Page",
+                    "display": {
+                        "properties": {}
+                    },
+                    "sections": [
+                        {
+                            "description": "",
+                            "display": {
+                                "properties": {}
+                            },
+                            "id": "interstitial-section",
+                            "questions": [
+                                {
+                                    "description": "You have successfully completed the breakfast section. Next we want to know about your lunch.",
+                                    "display": {
+                                        "properties": {}
+                                    },
+                                    "answers": [],
+                                    "id": "interstitial-question",
+                                    "title": "Breakfast Interstitial Page",
+                                    "type": "General",
+                                    "validation": []
+                                }
+                            ],
+                            "title": "Breakfast Interstitial Page",
+                            "validation": []
+                        }
+                    ]
+                },
+                {
+                    "type": "questionnaire",
+                    "display": {
+                        "properties": {}
+                    },
+                    "id": "lunch-block",
+                    "sections": [
+                        {
+                            "description": "",
+                            "display": {
+                                "properties": {}
+                            },
+                            "id": "lunch-section",
+                            "questions": [
+                                {
+                                    "answers": [
+                                        {
+                                            "display": {},
+                                            "guidance": "",
+                                            "id": "favourite-lunch",
+                                            "label": "What is your favourite lunchtime food",
+                                            "mandatory": false,
+                                            "options": [],
+                                            "q_code": "0",
+                                            "type": "TextField"
+                                        }
+                                    ],
+                                    "description": "",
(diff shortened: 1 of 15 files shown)


app/main/views/questionnaire.py
```diff
@@ -62,9 +62,16 @@ def get_introduction(eq_id, form_type, collection_id):
...
 @login_required
 def get_questionnaire(eq_id, form_type, collection_id, location):
-    get_questionnaire_manager(g.schema, g.schema_json).build_state(location, get_answers(current_user))
+    questionnaire_manager = get_questionnaire_manager(g.schema, g.schema_json)
 
-    return _render_template(location, get_questionnaire_manager(g.schema, g.schema_json).state, template='questionnaire')
+    questionnaire_manager.build_state(location, get_answers(current_user))
+
+    schema = questionnaire_manager.get_schema()
+    block = schema.get_item_by_id(location) if schema.item_exists(location) else None
```
benjefferies: schema.get_item_by_id(location) will raise a QuestionnaireException if the ID doesn't exists which will in turn show an error page. I think it will be fine just to get the block and then you can make the assumption it will not be none

app/parser/v0_0_1/schema_parser.py
```diff
@@ -183,6 +183,8 @@ def _parse_block(self, schema, questionnaire):
         try:
             block.id = ParserUtils.get_required_string(schema, "id")
             block.title = ParserUtils.get_optional_string(schema, "title")
+            block.description = ParserUtils.get_optional_string(schema, "description")
```
benjefferies: We should update the schema to match the new items. It's interesting the tests didn't fail?

app/templates/interstitial.html
```diff
@@ -0,0 +1,22 @@
...
+</div>
+
+<div>
+    {{content.schema_item.description}}
+</div>
+
+<div>
+    <form action="" method="POST" novalidate>
+        <button class="btn btn--secondary qa-btn-submit venus" type="submit" name="action[save_continue]">Continue</button>
+        <a class="u-dib" href="{{previous_location}}">Change your answers</a>
```
benjefferies: Link looks inconsistent with the other pages for previous location as it appears by the side

app/templates/interstitial.html
```diff
@@ -0,0 +1,22 @@
+{% extends 'layouts/_survey.html' -%}
+
+{% block page_title -%}Summary{% endblock -%}
```
benjefferies: Should the title be {{content.schema_item.title}}?

app/templates/interstitial.html
```diff
@@ -0,0 +1,22 @@
...
+  <h1 class="header__title saturn">{{content.schema_item.title}}</h1>
+</div>
+
+<div>
+    {{content.schema_item.description}}
+</div>
+
+<div>
+    <form action="" method="POST" novalidate>
+        <button class="btn btn--secondary qa-btn-submit venus" type="submit" name="action[save_continue]">Continue</button>
```
benjefferies: Do we need the action in the name?

tests/functional/spec/interstitial.spec.js
```diff
@@ -0,0 +1,29 @@
...
+    it('Given I am completing a survey with an interstitial page, when I arrive at the interstitial page, then i should be able to continue the survey.', function () {
+        //Given
+        startQuestionnaire(interstitial_schema)
+        InterstitialMultipleChoiceWithOtherPage.clickOther().setOtherInputField('Cereal').submit()
+
+        // When
+        expect(InterstitialPage.getTitle()).to.contain('Breakfast Interstitial Page')
+
+        //Then
+        InterstitialPage.submit()
```
benjefferies, line 22: Are we doing more than we need to in this test by going to the summary page and checking it? Are we covered by other tests on that? It means we can delete some code if not

benjefferies (changes requested):

app/main/views/questionnaire.py
jonnyshaw89: it won't call schema.get_item_by_id(location) if it doesn't exist as it checks if schema.item_exists(location)

app/parser/v0_0_1/schema_parser.py
jonnyshaw89: The tests don't fail as the schema defines the mandatory

app/templates/interstitial.html
jonnyshaw89: Yes

app/templates/interstitial.html
jonnyshaw89: Not anymore but it maintains consistency with the other for now.

tests/functional/spec/interstitial.spec.js
jonnyshaw89: None of the other tests test that button on the interstitial page.

app/parser/v0_0_1/schema_parser.py
benjefferies: We should also define the optional as it will potentially be a tool for building schemas

jonnyshaw89: Its now added

app/data/test_interstitial_page.json
```diff
@@ -0,0 +1,193 @@
+{
```
ajmaddaford, line 1: I think we should aim to create the simplest possible schemas to demonstrate features. Using a simple text input per page rather than a radio options would be better here.

app/main/views/questionnaire.py
```diff
@@ -62,9 +62,16 @@ def get_introduction(eq_id, form_type, collection_id):
 @questionnaire_blueprint.route('<location>', methods=["GET"])
 @login_required
 def get_questionnaire(eq_id, form_type, collection_id, location):
-    get_questionnaire_manager(g.schema, g.schema_json).build_state(location, get_answers(current_user))
+    questionnaire_manager = get_questionnaire_manager(g.schema, g.schema_json)
 
-    return _render_template(location, get_questionnaire_manager(g.schema, g.schema_json).state, template='questionnaire')
+    questionnaire_manager.build_state(location, get_answers(current_user))
```
ajmaddaford: Aren't we unnecessarily building state for the interstitial pages? The same question applies to the POST.

app/main/views/questionnaire.py
```diff
@@ -62,9 +62,16 @@ def get_introduction(eq_id, form_type, collection_id):
...
+
+    schema = questionnaire_manager.get_schema()
+    block = schema.get_item_by_id(location)
```
ajmaddaford: The schema is available via `g.schema` (rather than using the `get_schema` method)

tests/functional/pages/surveys/interstitial/interstitial.page.js
```diff
@@ -0,0 +1,12 @@
+import InterstitialPage from '../../interstitial.page'
```
ajmaddaford: The file should be named `breakfast-interstitial.page.js`

ajmaddaford (changes requested):

ajmaddaford (approved):

app/data/test_interstitial_page.json
```diff
@@ -0,0 +1,157 @@
...
+    },
+    "groups": [
+        {
+            "blocks": [
+                {
+                    "type": "questionnaire",
+                    "display": {
+                        "properties": {}
+                    },
+                    "id": "f22b1ba4-d15f-48b8-a1f3-db62b6f34cc0",
```
benjefferies: It would be nice if we had readable names in schemas going forward

jonnyshaw89: Agreed, do you want me to change it in this test one?

benjefferies: I think we should, I mentioned it on sams PR too

benjefferies (approved):

PR #558, opened 2016-11-11
Added ability to have a Final Confirmation to Submit at the end

jonnyshaw89:
Added ability to have a Final Confirmation to Submit at the end of the questionnaire

### How to review 
Launch test survey test_final_confirmation.json.
At the end of the survey you will be asked before the survey is submitted.

### DIFF
--- a/app/data/test_final_confirmation.json
+++ b/app/data/test_final_confirmation.json
@@ -1,0 +1,85 @@
+{
+    "mime_type": "application/json/ons/eq",
+    "questionnaire_id": "0",
+    "schema_version": "0.0.1",
+    "survey_id": "0",
+    "title": "Final confirmation to submit",
+    "theme": "default",
+    "description": "A questionnaire to demo final confirmation to submit.",
+    "introduction": {
+        "description": ""
+    },
+    "display": {
+        "properties": {}
+    },
+    "eq_id": "0",
+    "messages": {
+        "INTEGER_TOO_LARGE": "Number is too large",
+        "NEGATIVE_INTEGER": "Number cannot be less than zero",
+        "NOT_INTEGER": "Please enter an integer"
+    },
+    "groups": [
+        {
+            "blocks": [
+                {
+                    "display": {
+                        "properties": {}
+                    },
+                    "id": "f22b1ba4-d15f-48b8-a1f3-db62b6f34cc0",
+                    "sections": [
+                        {
+                            "description": "",
+                            "display": {
+                                "properties": {}
+                            },
+                            "id": "ed3e200a-0735-4e8d-9eea-627c1d908697",
+                            "questions": [
+                                {
+                                    "answers": [
+                                        {
+                                            "display": {},
+                                            "guidance": "",
+                                            "id": "ca3ce3a3-ae44-4e30-8f85-5b6a7a2fb23c",
+                                            "label": "What is your favourite breakfast food",
+                                            "mandatory": false,
+                                            "options": [],
+                                            "q_code": "0",
+                                            "type": "TextField"
+                                        }
+                                    ],
+                                    "description": "",
+                                    "display": {
+                                        "properties": {}
+                                    },
+                                    "id": "4bfa0229-0ce8-4190-817b-535f55ad2817",
+                                    "title": "",
+                                    "type": "General",
+                                    "validation": []
+                                }
+                            ],
+                            "title": "What is your favourite breakfast food",
+                            "validation": []
+                        }
+                    ],
+                    "routing_rules": [
+                    ],
+                    "title": "",
+                    "validation": []
+                },
+                {
+                    "display": {
+                        "properties": {}
+                    },
+                    "sections": [],
+                    "id": "confirmation",
+                    "title": ""
+                }
+            ],
+            "display": {
+                "properties": {}
+            },
+            "id": "14ba4707-321d-441d-8d21-b8367366e766",
+            "title": ""
+        }
+    ]
+}
--- a/app/main/views/questionnaire.py
+++ b/app/main/views/questionnaire.py
@@ -1,7 +1,7 @@
 import logging
 
 from app.authentication.session_manager import session_manager
-from app.globals import get_answers, get_completed_blocks, get_metadata, get_questionnaire_store
+from app.globals import get_answers, get_completed_blocks, get_metadata, get_navigator, get_questionnaire_store
 from app.questionnaire.questionnaire_manager import get_questionnaire_manager
 from app.submitter.submitter import SubmitterFactory
 from app.templating.introduction_context import get_introduction_context
@@ -59,7 +59,8 @@ def get_introduction(eq_id, form_type, collection_id):
 @questionnaire_blueprint.route('<location>', methods=["GET"])
 @login_required
 def get_questionnaire(eq_id, form_type, collection_id, location):
-    questionnaire_manager = get_questionnaire_manager(g.schema, g.schema_json)
+    navigator = get_navigator(g.schema_json)
+    questionnaire_manager = get_questionnaire_manager(navigator, g.schema, g.schema_json)
 
     questionnaire_manager.build_state(location, get_answers(current_user))
 
@@ -67,17 +68,18 @@ def get_questionnaire(eq_id, form_type, collection_id, location):
 
     template = block.type if block and block.type else 'questionnaire'
 
-    return _render_template(location, get_questionnaire_manager(g.schema, g.schema_json).state, template=template)
+    return _render_template(location, questionnaire_manager.state, template=template)
 
 
 @questionnaire_blueprint.route('<location>', methods=["POST"])
 @login_required
 def post_questionnaire(eq_id, form_type, collection_id, location):
-    valid = get_questionnaire_manager(g.schema, g.schema_json).process_incoming_answers(location, request.form)
+    navigator = get_navigator(g.schema_json)
+    questionnaire_manager = get_questionnaire_manager(navigator, g.schema, g.schema_json)
+    valid = questionnaire_manager.process_incoming_answers(location, request.form)
     if not valid:
-        return _render_template(location, get_questionnaire_manager(g.schema, g.schema_json).state, template='questionnaire')
+        return _render_template(location, questionnaire_manager.state, template='questionnaire')
 
-    navigator = get_questionnaire_manager(g.schema, g.schema_json).navigator
     next_location = navigator.get_next_location(get_answers(current_user), location)
     metadata = get_metadata(current_user)
     logger.info("Redirecting user to next location %s with tx_id=%s", next_location, metadata["tx_id"])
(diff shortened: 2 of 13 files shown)


app/data/test_final_confirmation.json
```diff
@@ -0,0 +1,183 @@
+{
```
benjefferies, line 1: Having a smaller schema for this test will emphasis the final confirmation configuration when someone comes to create it for real and it's also just going to be more questions than needed for testing

tests/integration/questionnaire/test_questionnaire_final_confirmation.py
```diff
@@ -0,0 +1,68 @@
...
+
+        # Given
+        token = create_token('final_confirmation', 'test')
+        self.client.get('/session?token=' + token.decode(), follow_redirects=False)
+
+        # When
+        # We proceed through the questionnaire
+        post_data = {
+            'action[start_questionnaire]': 'Start Questionnaire'
+        }
```
benjefferies, line 18: To back my point up above this test would be smaller and simpler if the schema only had one question as the questions are not the focus of the test

app/main/views/questionnaire.py
```diff
@@ -96,6 +96,19 @@ def get_summary(eq_id, form_type, collection_id):
     return redirect_to_questionnaire_page(eq_id, form_type, collection_id, latest_location)
 
 
+@questionnaire_blueprint.route('confirmation', methods=["GET"])
+@login_required
+def get_confirmation(eq_id, form_type, collection_id):
+    navigator = get_questionnaire_manager(g.schema, g.schema_json).navigator
```
benjefferies: We seem to use navigator a lot but we are getting it via questionnaire_manager - should we consider a global for navigator by now?

app/main/views/questionnaire.py
```diff
@@ -96,6 +96,19 @@ def get_summary(eq_id, form_type, collection_id):
...
+    latest_location = navigator.get_latest_location(get_answers(current_user), get_completed_blocks(current_user))
+    if latest_location == 'confirmation':
+        metadata = get_metadata(current_user)
+        answers = get_answers(current_user)
+        schema_json = _render_schema(g.schema_json, answers, metadata)
+        return _render_template('confirmation', get_introduction_context(schema_json), rendered_schema_json=schema_json)
+    return redirect_to_questionnaire_page(eq_id, form_type, collection_id, latest_location)
```
benjefferies, line 114: Should we refactor the 'check if the user is actually on summary/thank you/confirmation page or redirect' logic as we do it in a few places now

app/main/views/questionnaire.py
```diff
@@ -96,6 +96,19 @@ def get_summary(eq_id, form_type, collection_id):
...
+@questionnaire_blueprint.route('confirmation', methods=["GET"])
+@login_required
+def get_confirmation(eq_id, form_type, collection_id):
+    navigator = get_questionnaire_manager(g.schema, g.schema_json).navigator
+    latest_location = navigator.get_latest_location(get_answers(current_user), get_completed_blocks(current_user))
+    if latest_location == 'confirmation':
+        metadata = get_metadata(current_user)
+        answers = get_answers(current_user)
+        schema_json = _render_schema(g.schema_json, answers, metadata)
+        return _render_template('confirmation', get_introduction_context(schema_json), rendered_schema_json=schema_json)
```
benjefferies: I don't think we need the context `get_introduction_context(schema_json)` for the confirmation page

benjefferies (changes requested):

app/data/test_final_confirmation.json
jonnyshaw89: done

tests/integration/questionnaire/test_questionnaire_final_confirmation.py
jonnyshaw89: done

tests/functional/pages/confirmation.page.js
```diff
@@ -0,0 +1,22 @@
+class Confirmation {
+
+  isOpen() {
+    const url = browser.url().value
+    return url.indexOf('confirmation') > -1
+  }
+
+  getH2Text() {
```
benjefferies: Minor: getSubtitle might be a better method name

benjefferies (commented):
Happy to merge after squash of the commits

benjefferies (approved):

benjefferies:
@yoganandkunche can you approve and merge if you are happy with this from a QA perspective?

app/main/views/questionnaire.py
jonnyshaw89: I have moved this off of questionnaire manager now as we use it more outside of there than we did inside.

app/data/test_final_confirmation.json
ajmaddaford: Using a simple text input rather than radio options would reduce the size of the schema further.

app/questionnaire/navigator.py
```diff
@@ -59,6 +61,14 @@ def build_path(blocks, block_id, answers, path):
     return path
 
 
+def get_navigator(schema_json):
```
ajmaddaford: Can this method be moved to globals.py?

app/questionnaire/questionnaire_manager.py
```diff
@@ -30,13 +30,11 @@ def __init__(self, schema, json=None):
         self._schema = schema
         self.state = None
 
-        self.navigator = Navigator(self._json)
```
ajmaddaford, line 33: Can we still set a navigator property on questionnaire manager, but pass it in instead of initialising it on creation. This would make the navigator usage consistent with the global json and schema.

tests/functional/pages/thank-you.page.js
```diff
@@ -5,6 +5,10 @@ class ThankYou {
     return url.indexOf('thank-you') > -1
   }
 
+  getTitle() {
```
ajmaddaford: This may get confused with getting the contents of the HTML `<title>` element, maybe `getMainHeading` would be better?

tests/functional/pages/confirmation.page.js
```diff
@@ -0,0 +1,22 @@
+class Confirmation {
```
ajmaddaford: I think if we named this class `ConfirmationPage` it would make the places it's used in the tests more readable

tests/functional/spec/final-confirmation.spec.js
```diff
@@ -0,0 +1,46 @@
...
+
+        // When
+        MandatoryRadioPage.clickOther().setOtherInputField('Hello').submit()
+
+        //Then
+        expect(Confirmation.getSubtitle()).to.contain('Thank you for your answers, do you wish to submit')
+
+        Confirmation.submit()
+
+        expect(ThankYou.getTitle()).to.contain('Submission Successful')
```
ajmaddaford: I think this should be a separate test

tests/functional/spec/final-confirmation.spec.js
```diff
@@ -0,0 +1,46 @@
...
+
+    it('Given I successfully complete a questionnaire and an on the confirmation page, when I click change answers, then I should be returned to the questionnaire', function () {
+        //Given
+        startQuestionnaire(confirmation_schema)
+
+        // When
+        MandatoryRadioPage.clickOther().setOtherInputField('Hello').submit()
+
+        //Then
+        expect(Confirmation.getSubtitle()).to.contain('Thank you for your answers, do you wish to submit')
```
ajmaddaford: This is tested in the first test, the 'Then' should be the expectation that we're on the page for editing

ajmaddaford (changes requested):

app/questionnaire/navigator.py
benjefferies: We might not be able to. I put the get_questionnaire_manager in questionnaire_manager.py because of a circular import dependency i was facing

tests/functional/spec/final-confirmation.spec.js
```diff
@@ -0,0 +1,49 @@
...
+    it('Given I successfully complete a questionnaire and am on the confirmation page, when I click change answers, then I should be returned to the questionnaire', function () {
+        //Given
+        startQuestionnaire(confirmation_schema)
+        FinalConfirmationSurveyPage.setBreakfastFood('Bacon').submit()
+
+        // When
+        Confirmation.changeAnswers()
+
+        //Then
+        expect(FinalConfirmationSurveyPage)
```
collisdigital: Is this an actual check, don't know chai but all other expect calls have a chain on them?

benjefferies: Good spot, this doesn't look right

ajmaddaford (approved):

tests/functional/spec/final-confirmation.spec.js
collisdigital: Raise the question of how to prevent these kind of false-positive test assertions @ajmaddaford any thoughts?

tests/functional/spec/final-confirmation.spec.js
```diff
@@ -0,0 +1,49 @@
...
+    it('Given I successfully complete a questionnaire and am on the confirmation page, when I click change answers, then I should be returned to the questionnaire', function () {
+        //Given
+        startQuestionnaire(confirmation_schema)
+        FinalConfirmationSurveyPage.setBreakfastFood('Bacon').submit()
+
+        // When
+        Confirmation.changeAnswers()
+
+        //Then
+        expect(FinalConfirmationSurveyPage.isOpen())
```
benjefferies: Think this needs to.be.true on the end

benjefferies (changes requested):

tests/functional/spec/final-confirmation.spec.js
```diff
@@ -0,0 +1,49 @@
...
+    it('Given I successfully complete a questionnaire and am on the confirmation page, when I click change answers, then I should be returned to the questionnaire', function () {
+        //Given
+        startQuestionnaire(confirmation_schema)
+        FinalConfirmationSurveyPage.setBreakfastFood('Bacon').submit()
+
+        // When
+        Confirmation.changeAnswers()
+
+        //Then
+        expect(FinalConfirmationSurveyPage.isOpen().to.be.true)
```
benjefferies: should be `expect(FinalConfirmationSurveyPage.isOpen()).to.be.true`

benjefferies:
This branch needs rebasing against master

benjefferies (approved):

yoganandkunche:
A defect is raised '&' entered in Text is displayed as &amp;
Also need to discuss with @willrichards4 regarding the 'Change your answers' link
@jonathanshaw6 As discussed can you please add another question to the test schema so that I can test the navigation before signing off this piece of functionality.

yoganandkunche:
Following the discussion with @willrichards4 @jonathanshaw6 removed the 'Change your answers' link and added a 'Previous' link - Tested this functionality and it works as Expected.

A defect is raised '&' entered in Text is displayed as &amp; is in 'Open' State

I am removing the 'needs-qa' label now. It can be merged if the dev reviewing this pull request is happy.

PR #559, opened 2016-11-11
AnswerStore

iwootten:

Added an AnswerStore to support the various repeating question requirements for census Household composition questions. 

Migrated the existing calls to get_answers to use AnswerStores map() method in order to support original structure.

### How to review 

Check the approach, see if the tests cover behaviour.

### DIFF
--- a/app/data_model/answer_store.py
+++ b/app/data_model/answer_store.py
@@ -1,0 +1,202 @@
+import re
+
+from collections import OrderedDict
+
+
+class Answer(object):
+    def __init__(self, answer_id, block_id, value, answer_instance=0):
+        self.answer_id = answer_id
+        self.block_id = block_id
+        self.answer_instance = answer_instance
+        self.value = value
+
+    def matches(self, answer):
+        """
+        Check to see if two answers match.
+        Two answers are considered to match if they share the same block, answer and instance id.
+
+        :param answer: An answer to compare
+        :return: True if both answers match, otherwise False.
+        """
+        return self.block_id == answer.block_id and \
+            self.answer_id == answer.answer_id and \
+            self.answer_instance == answer.answer_instance
+
+    def matches_dict(self, answer_dict):
+        """
+        Check to see if a dict describes an answer the same as this object.
+
+        :param answer_dict: A dict providing
+        :return:
+        """
+
+        return self.matches(Answer(
+            answer_dict['answer_id'],
+            answer_dict['block_id'],
+            "",
+            answer_dict['answer_instance'],
+        ))
+
+
+class AnswerStore(object):
+    """
+    An object that stores and updates a collection of answers, ready for serialisation
+    via the Questionnaire Store.
+    """
+    def __init__(self, existing_answers=None):
+        self.answers = existing_answers or []
+
+    @staticmethod
+    def _validate(answer):
+        if not isinstance(answer, Answer):
+            raise TypeError("Method only supports Answer argument type")
+
+    def add(self, answer):
+        """
+        Add a new answer into the answer store.
+
+        :param answer: A dict of flattened answer details.
+        """
+        self._validate(answer)
+
+        answer_to_add = answer.__dict__.copy()
+
+        if self.exists(answer):
+            raise ValueError("Answer instance already exists in store")
+        else:
+            self.answers.append(answer_to_add)
+
+    def update(self, answer):
+        """
+        Update the value of an answer already in the answer store.
+
+        :param answer: A dict of flattened answer details.
+        """
+        position = self.find(answer)
+
+        if position is None:
+            raise ValueError("Answer instance does not exist in store")
+        else:
+            self.answers[position]['value'] = answer.value
+
+    def add_or_update(self, answer):
+        """
+        Add a new answer into the answer store, or update if it exists.
+
+        :param answer: An answer object.
+        """
+        if self.exists(answer):
+            self.update(answer)
+        else:
+            self.add(answer)
+
+    def get(self, answer):
+        """
+        Returns the value of an answer.
+
+        :param answer: The ids for the answer to return
+        :return: The value of the answer found
+        """
+        position = self.find(answer)
+
+        if position is None:
+            raise ValueError("Answer instance does not exist in store")
+        else:
+            return self.answers[position]['value']
+
+    def find(self, answer):
+        """
+        Returns the position of an answer if it exists
+
+        :param answer: The answer to search for
+        :return: The position the answer exists at, None if it doesn't exist
+        """
+        self._validate(answer)
+
+        for index, existing in enumerate(self.answers):
+            if answer.matches_dict(existing):
+                return index
+
+        return None
+
+    def exists(self, answer):
+        """
+        Checks to see if an answer exists in the answer store.
+
+        :param answer: A dict of flattened answer details.
+        :return: True if the answer is in the store, False if not.
+        """
+        return self.find(answer) is not None
+
+    def count(self, answer):
+        """
+        Count of the number of instances of an answer in the answer store.
+
+        :param answer: A dict of flattened answer details.
+        :return: 0 if the answer doesn't exist, otherwise the number of instances.
+        """
+        self._validate(answer)
+
+        return len(self.filter(answer.answer_id, answer.block_id, answer.answer_instance))
+
+    def filter(self, answer_id=None, block_id=None, answer_instance=None):
+        """
+        Find all answers in the answer store for a given set of filter parameter matches.
+
+        :param answer_id:
+        :param block_id:
+        :param answer_instance:
+        :return:
+        """
+        filtered = []
+        filter_vars = {
+            "answer_id": answer_id,
+            "block_id": block_id,
+            "answer_instance": answer_instance,
+        }
+        for answer in self.answers:
+            matches = True
+            for k, v in filter_vars.items():
+                if v is not None:
+                    matches = matches and answer[k] == v
+            if matches:
+                filtered.append(answer)
+        return filtered
+
+    def clear(self):
+        self.answers.clear()
+
+    def map(self, answer_id=None, block_id=None, answer_instance=None):
+        """
+        Maps the answers in this store to a dictionary of key, value answers. Keys include instance
+        id's when the instance id is non zero.
+
+        :param answer_id:
+        :param block_id:
+        :param answer_instance:
+        :return:
+        """
+        result = {}
+        use_filter = (answer_id or block_id or answer_instance) is not None
+        answers = self.filter(answer_id, block_id, answer_instance) if use_filter else self.answers
+
+        for answer in answers:
+            answer_id = answer['answer_id']
+            answer_id += "_" + str(answer['answer_instance']) if answer['answer_instance'] > 0 else ''
+
+            result[answer_id] = answer['value']
+
+        return OrderedDict(sorted(result.items(), key=lambda t: natural_order(t[0])))
+
+
+def number_else_string(text):
+    return int(text) if text.isdigit() else text
+
+
+def natural_order(key):
+    """
+    Orders a set of items according to
+    :param key:
+    :return:
+    """
+    return [number_else_string(c) for c in re.split('(\d+)', key)]
--- a/app/globals.py
+++ b/app/globals.py
@@ -28,9 +28,14 @@ def get_metadata(user):
     return questionnaire_store.metadata
 
 
-def get_answers(user):
+def get_answer_store(user):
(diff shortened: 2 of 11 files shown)


app/data_model/answer_store.py
```diff
@@ -0,0 +1,134 @@
...
+
+    def __init__(self, existing_answers=None):
+        self.answers = existing_answers or []
+
+    def add(self, answer):
+        """
+        Add a new answer into the answer store.
+
+        :param answer: A dict of flattened answer details.
+        :return: None.
```
benjefferies: Do we need to explicitly state the return here?

app/data_model/answer_store.py
```diff
@@ -0,0 +1,134 @@
...
+
+        :param answer: A dict of flattened answer details.
+        :return:
+        """
+        for index, existing in enumerate(self.answers):
+            if self.same(existing, answer):
+                self.answers[index]['value'] = answer['value']
+                break
+
+    def find(self, answer):
```
benjefferies, line 107: In what scenario do we have multiple instances of an answer?

app/data_model/answer_store.py
```diff
@@ -0,0 +1,134 @@
...
+        for answer in answers:
+            answer_id = answer['answer_id']
+            answer_id += "_" + str(answer['answer_instance']) if answer['answer_instance'] > 0 else ''
+
+            result[answer_id] = answer['value']
+
+        return OrderedDict(sorted(result.items(), key=lambda t: natural_order(t[0])))
+
+
+def atoi(text):
```
benjefferies: atoi isn't the friendliest name, maybe number_else_string might be a bit more descriptive?

app/data_model/answer_store.py
```diff
@@ -0,0 +1,135 @@
...
+        Returns the value of an answer.
+
+        :param answer:The alue of an answer
+        :return:
+        """
+        found = False
+
+        for existing in self.answers:
+            if self.same(answer, existing):
+                found = True
```
jonnyshaw89: Could you not call self.exists here?

app/data_model/answer_store.py
```diff
@@ -0,0 +1,135 @@
...
+
+        return existing['value'] if found else None
```
jonnyshaw89: where is existing set? as we are out of the for loop her?

app/data_model/answer_store.py
iwootten: self.exists also loops the entire set of answers, so self.same should be more performant.

app/data_model/answer_store.py
iwootten: Yes we are, I've updated this so that it returns immediately.

app/data_model/answer_store.py
```diff
@@ -0,0 +1,133 @@
...
+            self.update(answer)
+        else:
+            self.answers.append(answer_to_add)
+
+    def update(self, answer):
+        """
+        Update the value of an answer already in the answer store.
+
+        :param answer: A dict of flattened answer details.
+        :return:
```
benjefferies: Do we need to explicitly state the return here?

app/data_model/answer_store.py
```diff
@@ -0,0 +1,133 @@
...
+                if k in ['block_id', 'question_id', 'answer_id', 'answer_instance']:
+                    matches = matches and answer[k] == v
+            if matches:
+                filtered.append(answer)
+        return filtered
+
+    def clear(self):
+        self.answers.clear()
+
+    def map(self, filter_vars=None):
```
benjefferies: Not sure if the method would be easier to use if the parameters were explicit on the contrary if not we should define what the filter_vars can be. i.e. 'block_id', 'answer_id'

app/questionnaire_state/state_answer.py
```diff
@@ -42,3 +42,12 @@ def _restore_other_value(self, user_input):
...
     def get_answers(self):
         return [self]
+
+    def flatten(self):
+        return {
+            'block_id': self.parent.parent.parent.id,
+            'question_id': self.parent.id,
+            'answer_id': self.id,
+            'answer_instance': self.instance,
+            'value': self.value if self.value is not None else None,
```
benjefferies: `'value': self.value = self.value`?

tests/app/data_model/test_answer_store.py
```diff
@@ -0,0 +1,246 @@
...
+            'answer_instance': 1,
+            'value': [23, 45, 67],
+        }
+
+        self.store.add(answer_1)
+        value = self.store.get(answer_1)
+
+        self.assertEqual(value, [23, 45, 67])
+
+        updated_answer = {
```
benjefferies: Does the latter part of this test add much value, is it not testing the same thing in theory?

app/data_model/answer_store.py
```diff
@@ -0,0 +1,133 @@
...
+    def add(self, answer):
+        """
+        Add a new answer into the answer store.
+
+        :param answer: A dict of flattened answer details.
+        :return: None.
+        """
+        answer_to_add = answer.copy()
+
+        if self.exists(answer_to_add):
```
benjefferies: Does it make sense to merge the add/update functionality. It makes update kind of redundant.

tests/app/data_model/test_answer_store.py
```diff
@@ -0,0 +1,246 @@
...
+
+    def test_returns_ordered_map(self):
+        answer = {
+            'block_id': "1",
+            'answer_id': "2",
+            'question_id': "3",
+            'value': 25,
+        }
+
+        for i in range(0, 1000):
```
benjefferies: Is this overkill? Could you not add two answers one with an id of 10 then 1 and check the order? It would simplify the test

benjefferies (changes requested):

app/data_model/answer_store.py
iwootten: I think they should be kept distinct, if anything the add behaviour should be modified to only add if it doesn't yet exist..

app/data_model/answer_store.py
iwootten: Maybe filtered could raise/log a warning on filtered variables it doesn't recognise?

app/data_model/answer_store.py
benjefferies: Sorry that's what I meant. Should have been clearer

app/data_model/answer_store.py
iwootten: Sure, this method is ripped straight from stackoverflow.

app/data_model/answer_store.py
benjefferies: Shouldn't we raise an error at that point as it would be programmer error and shouldn't happen?

tests/app/data_model/test_answer_store.py
iwootten: Belt and braces - a dict can potentially return in any order and may return the items in the order they were inserted, this test covers every eventuality. Suppose it could be reduced to 100, because even that will cover it.

tests/app/data_model/test_answer_store.py
iwootten: Yeah, this was originally here during debugging - don't think it's necessary any longer.

app/data_model/answer_store.py
```diff
@@ -0,0 +1,192 @@
...
+class Answer(object):
+    def __init__(self, answer_id, block_id, value, answer_instance=None):
+        answer_instance = answer_instance or 0
+
+        self.answer_id = answer_id
+        self.block_id = block_id
+        self.answer_instance = answer_instance
+        self.value = value
+
+    def equals(self, answer):
```
benjefferies: We could implement **eq** and **ne** to provide equality with the == operator

collisdigital: If you do this, be aware of adding a `__hash__` function (https://docs.python.org/3.6/reference/datamodel.html#object.__hash__) and also that you don't need to define `__ne__`: "By default, **ne**() delegates to **eq**() and inverts the result unless it is NotImplemented."

You may also want some kind of test harness that checks Answer's `__dict__` property and tries changing each value in turn to ensure the operator works; past experience shows it can slip through the net to update the operator when adding new properties to a class which can lead to strange/irritate behaviour (although the same applies even with the current implementation)

benjefferies: Oh just noticed that it's not really testing equality more that it's the same answer identifier so might not make sense here. Maybe the old `same` name was more reflective of that?

app/data_model/answer_store.py
```diff
@@ -0,0 +1,192 @@
...
+        :return:
+        """
+        filtered = []
+        filter_vars = {
+            "answer_id": answer_id,
+            "block_id": block_id,
+            "answer_instance": answer_instance,
+        }
+        for answer in self.answers:
+            matches = True
```
jonnyshaw89, line 158: Do we need to define this bool? on the first match can't we just append the answer and continue?

benjefferies: I think you need it because it will likely need to match all the criteria to be added

jonnyshaw89: ah, yes.

app/data_model/answer_store.py
```diff
@@ -0,0 +1,192 @@
...
+                self.answers[index]['value'] = answer.value
+                break
+
+    def add_or_update(self, answer):
+        """
+        Add a new answer into the answer store, or update if it exists.
+
+        :param answer: An answer object.
+        """
+        self._validate(answer)
```
jonnyshaw89: Do we need to validate here? as we are also calling validate in add and update

app/data_model/answer_store.py
```diff
@@ -0,0 +1,192 @@
...
+        return None
+
+    def exists(self, answer):
+        """
+        Checks to see if an answer exists in the answer store.
+
+        :param answer: A dict of flattened answer details.
+        :return: True if the answer is in the store, False if not.
+        """
+        self._validate(answer)
```
jonnyshaw89: do we need to call validate here as its called in count()

app/globals.py
```diff
@@ -28,9 +28,16 @@ def get_metadata(user):
...
+    questionnaire_store = get_questionnaire_store(user.user_id, user.user_ik)
+
+    return questionnaire_store.answer_store
+
+
 def get_answers(user):
     questionnaire_store = get_questionnaire_store(user.user_id, user.user_ik)
-    return questionnaire_store.answers
+
+    return questionnaire_store.answer_store.map()
```
jonnyshaw89: can we call get_answer_store(user).map() to reduce duplication

app/questionnaire/questionnaire_manager.py
```diff
@@ -55,31 +55,33 @@ def validate_all_answers(self):
...
                 logger.debug("Failed validation with current location %s", location)
                 return False, location
 
         return True, None
 
+    def update_questionnaire_store(self, location):
+        # Store answers in QuestionnaireStore
+        questionnaire_store = get_questionnaire_store(current_user.user_id, current_user.user_ik)
+
+        for answer in self.get_state_answers(location):
```
jonnyshaw89, line 70: should we call self.get_state_answers(location) before the for and pass the value in so that the function isn't called on every iteration?

iwootten: The function is evaluated before it's iterated over in the loop.

app/data_model/answer_store.py
```diff
@@ -0,0 +1,204 @@
...
+        if self.exists(answer):
+            self.update(answer)
+        else:
+            self.add(answer)
+
+    def get(self, answer):
+        """
+        Returns the value of an answer.
+
+        :param answer: The ids for the answer ro return
```
benjefferies: Minor typo `ro` -> `to`

app/data_model/answer_store.py
```diff
@@ -0,0 +1,204 @@
+import re
+
+from collections import OrderedDict
+
+
+class Answer(object):
+    def __init__(self, answer_id, block_id, value, answer_instance=None):
```
benjefferies: answer_instance could just default to 0?

iwootten: Yep.

benjefferies (approved):
Looks good lets squash those commits and get it in unless anyone else finds anything else?

app/questionnaire/questionnaire_manager.py
jonnyshaw89: It is indeed, a previous comment by Andrew on one of my PRs confused me about this functionality.

ajmaddaford (approved):
LGTM 👍

PR #562, opened 2016-11-14
Add new date type(Month/Year) and update tests

LJBabbage:
To add in the month/year answer type "mm/yyyy". A new widget has been introduced to deal with the alternative format and the summary . A test schema has also been set up to handle all dates 

### How to review 
Using dev, open up the test_dates.json and make sure it shows both the date range and the new format above. Add  invalid data into each and make sure the right messages are shown. Correct the errors and make sure the summary page is correct.

(N.B there is a malfunctioning test_date.json file as well and an associated selenium test already in master. This needs removing/fixing, but this is outside the scope of this story and will be done in EQ-718 date of birth)

### DIFF
--- a/app/data/test_dates.json
+++ b/app/data/test_dates.json
@@ -1,0 +1,124 @@
+{
+    "mime_type": "application/json/ons/eq",
+    "questionnaire_id": "23",
+    "schema_version": "0.0.1",
+    "survey_id": "023",
+    "title": "Date formats",
+    "description": "A test schema for different date formats",
+    "theme": "default",
+    "introduction": {
+        "description": "",
+        "information_to_provide": [
+          "dates"
+        ]
+    },
+    "display": {
+        "properties": {}
+    },
+    "eq_id": "0",
+    "groups": [
+        {
+            "blocks": [
+                {
+                    "display": {
+                        "properties": {}
+                    },
+                    "id": "b10c8c7a-97fe-4bee-a4fb-e75f3a6f56c0",
+                    "sections": [
+                        {
+                            "description": "",
+                            "display": {
+                                "properties": {}
+                            },
+                            "id": "7352fd10-bb24-407a-9f13-23cc8b1cb51d",
+                            "questions": [
+                                {
+                                    "answers": [
+                                        {
+                                            "display": {},
+                                            "guidance": "",
+                                            "id": "5b12aea2-ae3b-467d-9291-4d803a444d25",
+                                            "label": "Period from",
+                                            "mandatory": true,
+                                            "options": [],
+                                            "q_code": "11",
+                                            "type": "Date",
+                                            "validation": {
+                                                "messages": {
+                                                    "INVALID_DATE": "The date entered is not valid.  Please correct your answer.",
+                                                    "MANDATORY": "Please answer before continuing."
+                                                }
+                                            }
+                                        },
+                                        {
+                                            "display": {},
+                                            "guidance": "",
+                                            "id": "ce42c1f7-a32a-46d5-8183-f54e38616617",
+                                            "label": "Period to",
+                                            "mandatory": true,
+                                            "options": [],
+                                            "q_code": "12",
+                                            "type": "Date",
+                                            "validation": {
+                                                "messages": {
+                                                    "INVALID_DATE": "The date entered is not valid.  Please correct your answer.",
+                                                    "MANDATORY": "Please answer before continuing."
+                                                }
+                                            }
+                                        }
+                                    ],
+                                    "description": "",
+                                    "display": {
+                                        "properties": {}
+                                    },
+                                    "id": "e9130586-f72a-4b68-8373-885ad51d6cd5",
+                                    "title": "Date range",
+                                    "type": "DateRange",
+                                    "validation": []
+                                },
+                              {
+                                    "answers": [
+                                        {
+                                            "display": {},
+                                            "guidance": "",
+                                            "id": "eade17ef-3ec6-46c2-afc7-dffeb07e2e5e",
+                                            "label": "Date",
+                                            "mandatory": true,
+                                            "options": [],
+                                            "q_code": "11",
+                                            "type": "MonthYearDate",
+                                            "validation": {
+                                                "messages": {
+                                                    "INVALID_DATE": "The date entered is not valid.  Please correct your answer.",
+                                                    "MANDATORY": "Please answer before continuing."
+                                                }
+                                            }
+                                        }
+                                    ],
+                                    "description": "",
+                                    "display": {
+                                        "properties": {}
+                                    },
+                                    "id": "9858220e-0514-4868-984f-8bb2481fde8f",
+                                    "title": "Date with month and year",
+                                    "type": "General",
+                                    "validation": []
+                                }
+                            ],
+                            "title": "Date Examples",
+                            "validation": []
+                        }
+                    ],
+                    "title": "Date Examples",
+                    "validation": []
+                }
+            ],
+            "display": {
+                "properties": {}
+            },
+            "id": "a23d36db-6b07-4ce0-94b2-a843369511e3",
+            "title": "",
+            "validation": []
+        }
+    ]
+}
--- a/app/schema/widgets/date_widget.py
+++ b/app/schema/widgets/date_widget.py
@@ -10,22 +10,77 @@ logger = logging.getLogger(__name__)
 
 
 class DateWidget(Widget):
-    def _get_days(self, selected_day):
-        if selected_day:
-            selected_day = int(selected_day)
 
-        days = []
-        for day in range(1, 32):
-            days.append(ObjectFromDict({
-                'type': 'text',
-                'value': day,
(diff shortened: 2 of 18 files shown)


tests/functional/pages/surveys/dates/dates-answers.page.js
```diff
@@ -0,0 +1,47 @@
...
+    browser.setValue('[name="ce42c1f7-a32a-46d5-8183-f54e38616617-month"]', month)
+    return this
+  }
+
+  setToReportingPeriodYear(year) {
+    browser.setValue('[name="ce42c1f7-a32a-46d5-8183-f54e38616617-year"]', year)
+    return this
+  }
+
+  setMonthYearDay(month) {
```
jonnyshaw89: Is this name correct?

app/schema/widgets/date_widget.py
```diff
@@ -10,85 +10,90 @@
...
-                'disabled': False,
-            }))
-        return months
 
     def render(self, answer_state):
+
         if answer_state.input:
             parts = answer_state.input.split('/')
         else:
             parts = [None, None, '']
```
jonnyshaw89: Is this correct given that the day is a text box?

app/schema/widgets/month_year_date_widget.py
```diff
@@ -0,0 +1,47 @@
...
+    def _get_date_fields(self, parts):
+
+        fields = {}
+        fields['month'] = self._get_month_field(parts[0])
+        fields['year'] = self._get_year_field(parts[1])
+        return fields
+
+    def get_user_input(self, post_vars):
+
+        if isinstance(post_vars, dict):
```
benjefferies, line 34: Does this logic have to be complex i.e. Do we need to check if it's a dict or not? Under what circumstance will it not be a dict or will it be the other case?

tests/functional/pages/surveys/dates/dates-answers.page.js
LJBabbage: Oh yeah, thanks

tests/app/validation/test_date_month_year.py
```diff
@@ -0,0 +1,54 @@
...
+from app.validation.abstract_validator import AbstractValidator
+
+
+class DateMonthYearTest(unittest.TestCase):
+
+    def testmonth_year_date_validator(self):
+        month_year_date_type = MonthYearDateTypeCheck()
+
+        # validate None
+        result = month_year_date_type.validate(None)
```
benjefferies, line 11: This feels like 7/8 separate tests to me as we are testing 7/8 different scenarios? The comments are not needed when they are separate tests and it's easier to track broken functionality because it will say test_empty_string will fail in travis?

tests/app/validation/test_date_month_year.py
```diff
@@ -0,0 +1,54 @@
...
+        self.assertEquals(len(result.errors), 1)
+        self.assertEquals(AbstractValidator.INVALID_DATE, result.errors[0])
+
+        # valid date
+        result = month_year_date_type.validate('01/2017')
+        self.assertTrue(result.is_valid)
+        self.assertEquals(len(result.errors), 0)
+
+
+if __name__ == '__main__':
```
benjefferies: I think 

```
if __name__ == '__main__':
    ...
```

is redundant

tests/functional/spec/dates.spec.js
```diff
@@ -0,0 +1,129 @@
+import chai from 'chai'
+import {startQuestionnaire} from '../helpers'
+
+import DatesPage from '../pages/surveys/dates/dates-answers.page'
+import SummaryPage from '../pages/surveys/dates/dates-summary.page'
+
+const expect = chai.expect
+
+describe('Checkbox with "other" option', function() {
```
benjefferies: Describe text is wrong

tests/functional/spec/dates.spec.js
```diff
@@ -0,0 +1,129 @@
+import chai from 'chai'
```
benjefferies, line 1: Do we need to exercise all of this logic as selenium tests, can the validation logic not be exercised as unit tests?

benjefferies (changes requested):

app/schema/widgets/month_year_date_widget.py
LJBabbage: Its the old/current pattern, but it does get them in the different types if memory serves correctly. As discussed, one for technical debt

tests/app/validation/test_date_month_year.py
LJBabbage: No problem will separate them

tests/functional/spec/dates.spec.js
LJBabbage: Did both

tests/app/validation/test_date_month_year.py
LJBabbage: Has been removed

tests/functional/spec/dates.spec.js
LJBabbage: Updated

app/schema/widgets/month_year_date_widget.py
benjefferies: Cool, I have checked and it's still used in both patterns. It's bit nasty and confusing really but not scope of this PR

app/validation/month_year_date_type_check.py
```diff
@@ -0,0 +1,24 @@
...
+    """
+    Validate that the users answer is a valid date
+    :param user_answer: The answer the user provided for the response
+    :return: ValidationResult(): An object containing the result of the validation
+    """
+
+    def validate(self, user_answer):
+        result = ValidationResult(False)
+        try:
+            time.strptime(user_answer, "%m/%Y")  # NOQA
```
jonnyshaw89: Is this just cheating? # NOQA
Maybe if we add this we should also say why?

LJBabbage: Its not needed, and has been removed

app/data/test_dates.json
```diff
@@ -0,0 +1,124 @@
+{
```
hamishtaplin, line 1: This adds a `test_dates` schema but we already have `test_date`—should these be combined?

hamishtaplin (dismissed):

app/schema/widgets/date_widget.py
```diff
@@ -10,85 +10,90 @@
...
+
+        fields = {}
+        fields['day'] = self._get_day_field(parts[0])
+        fields['month'] = self._get_month_field(parts[1])
+        fields['year'] = self._get_year_field(parts[2])
+
+        return fields
+
+    def _get_day_field(self, day):
+        day = {
```
benjefferies: inline return would be nicer than reassigning day to be something of a totally different type

app/schema/widgets/date_widget.py
```diff
@@ -10,85 +10,90 @@
...
                   },
-                },
-                'month': {
-                  'label': {
+                }
+        return day
+
+    def _get_month_field(self, month):
+
+        month = {
```
benjefferies: inline return would be nicer than reassigning month to be something of a totally different type

app/schema/widgets/date_widget.py
```diff
@@ -10,85 +10,90 @@
...
                     'id': self.name + '-month',
                   },
-                },
-                'year': {
-                  'label': {
+                }
+        return month
+
+    def _get_year_field(self, year):
+        year = {
```
benjefferies: inline return would be nicer than reassigning year to be something of a totally different type

app/schema/widgets/date_widget.py
```diff
@@ -10,85 +10,90 @@
...
-                },
-            },
-        }
+                }
+        return year
 
-        return render_template('partials/widgets/date_widget.html', **widget_params)
+    @staticmethod
+    def _get_months(selected_month):
+        if selected_month:
```
benjefferies, line 84: How will the widget handle it if the selected month isn't selected? Should there always be a selected month?

app/data/test_dates.json
LJBabbage: Yep, but I am doing it in the next story as the note above, just because I need to fix the summary screen for the single date before combining them

app/schema/widgets/date_widget.py
LJBabbage: Its the refilling of pre-existing data, if there is no month then it should not have a selected value and just behave as a normal drop down

benjefferies (approved):

app/schema/widgets/month_year_date_widget.py
samiwel: Just to supplement this discussion, the repeating single answer story and new answer store mechanism should hopefully ensure that only dict will be passed down to the widget going forward.

jonnyshaw89 (changes requested):
All good once the codacy issues are fixed

jonnyshaw89 (approved):

PR #563, opened 2016-11-14
Household composition repeat single

samiwel:
Added a new question type for repeating answers.
Created a new schema to demonstrate the household composition question.
Implement repeated single answers using the new Answer Store.

### How to review 
Should be able to add and remove people on the household question screen.
The answers should be persisted in the Answer Store.
N.B. This version does not make use of Javascript and only deals with single answers repeating. Multiple answers and introducing Javascript are covered by different stories.


jonnyshaw89:
Think you committed your code coverage report?

app/templates/partials/question.html
```diff
@@ -24,13 +24,58 @@ <h3 class="question__subtitle mars">{{question_subtitle}}</h3>
...
+            <span class="js-household-action">You</span>
+         {%- else -%}
+         <button class="btn btn--link"
+            formaction="{{ url_for('action.remove_answer',
+              eq_id=meta.survey.eq_id,
+              collection_id=meta.survey.collection_id,
+              form_type=meta.survey.form_type,
+              location=content.id,
+              question=question.schema_item.id) }}"
+            name="action[remove_answer]" value="{{ loop.index0 }}">Remove <span class="u-vh">Person {{loop.index0}}</span></button>
```
collisdigital: These kinds of strings "Remove" and "Person" need to translatable to Welsh.

app/templates/partials/question.html
```diff
@@ -24,13 +24,58 @@ <h3 class="question__subtitle mars">{{question_subtitle}}</h3>
...
+<div class="question__actions">
+  {%- if question.schema_item.type == 'RepeatingAnswer' %}
+  <button class="btn btn--border btn--secondary venus js-household-btn" type="submit"
+          formaction="{{ url_for('action.add_answer',
+            eq_id=meta.survey.eq_id,
+            collection_id=meta.survey.collection_id,
+            form_type=meta.survey.form_type,
+            location=content.id,
+            question=question.schema_item.id) }}"
+          name="action[add_answer]">Add another person</button>
```
collisdigital: Same for translating this string

app/data/test_household_question.json
```diff
@@ -0,0 +1,74 @@
...
+                                  },
+                                  "id": "question",
+                                  "title": "Who usually lives here?",
+                                  "description": "<br> <div> <h3>Include:</h3> <ul> <li>Yourself, if this is your permanent or family home </li> <li>Family members including partners, children and babies born on or before 9 April 2017</li> <li>Students and/or school children who live away from home during term time</li> <li>Housemates, tenants or lodgers</li> </ul> </div>",
+                                  "type": "RepeatingAnswer",
+                                  "validation": [],
+                                  "answers": [
+                                    {
+                                        "display": {},
+                                        "id": "414699da-1667-44fd-8e98-7606966884db",
```
benjefferies: Is it worth calling this household_full_name

app/templates/partials/question.html
```diff
@@ -24,13 +24,58 @@ <h3 class="question__subtitle mars">{{question_subtitle}}</h3>
...
+   {%- if question.schema_item.type == 'RepeatingAnswer' %}
+     {%- for answer in question.answers %}
+        <div class="question__answer js-household-person">
+       <h3 class="neptune">Person <span class="js-household-loopindex">{{ loop.index }}</span>
+         <small>(
+         {%- if loop.index == 1 -%}
+            <span class="js-household-action">{{ _('You') }}</span>
+         {%- else -%}
+         <button class="btn btn--link"
+            formaction="{{ url_for('action.remove_answer',
```
hamishtaplin: We can't rely on `formaction` as it's a HTML5 attribute and so won't work in older browsers.

app/assets/js/app/main.js
```diff
@@ -4,3 +4,4 @@ import './modules/accordion'
 import './modules/analytics'
 import './modules/inpagelink'
 import './modules/focus-styles'
+// import './modules/household'
```
jonnyshaw89, line 7: Remove this comment

app/templates/partials/answer.html
```diff
@@ -1,7 +1,8 @@
 {% import 'macros/helpers.html' as helpers %}
 
-<div class="answer answer--{{answer.schema_item.type|lower}}" id="{{answer.id}}">
-  {% set invalid = answer.is_valid == False %}
+{% set invalid = answer.is_valid == False %}
+
+<div class="answer answer--{{answer.schema_item.type|lower}} {{'js-has-errors' if invalid}}" id="{{answer.id}}">
```
hamishtaplin, line 5: The `answer.id` here is not unique when we add new answers. This needs to be fixed at a high level (not in the template) as that id is then used to generate id's further down the tree.

hamishtaplin (changes requested):
A couple of issues you may already know about:
1. When you add multiple people and then submit without entering a name, the error at the top of the page doesn't reflect this (only one error).
2. When you submit with errors and then add another person, the errors aren't maintained.

app/assets/js/app/main.js
benjefferies, line 7: Can this be deleted?

app/templates/partials/question.html
```diff
@@ -24,13 +24,55 @@ <h3 class="question__subtitle mars">{{question_subtitle}}</h3>
...
+   {%- if question.schema_item.type == 'RepeatingAnswer' %}
+     {%- for answer in question.answers %}
+       <div class="question__answer js-household-person">
+         <h3 class="neptune">Person <span class="js-household-loopindex">{{ loop.index }}</span>
+           <small>(
+           {%- if loop.index == 1 -%}
+              <span class="js-household-action">{{ _('You') }}</span>
+           {%- else -%}
+           <button class="btn btn--link"
+              formaction="{{ url_for('action.remove_answer',
```
hamishtaplin: `formaction` is a html5 attribute and can't be relied on.

hamishtaplin (changes requested):

app/main/views/questionnaire.py
```diff
@@ -119,13 +125,31 @@ def submit_answers(eq_id, form_type, collection_id):
         return redirect_to_questionnaire_page(eq_id, form_type, collection_id, invalid_location)
 
 
+@action_blueprint.route('<location>/<question>/add', methods=["POST"])
+@login_required
+def add_answer(eq_id, form_type, collection_id, location, question):
+    answer_store = get_answer_store(current_user)
```
jonnyshaw89, line 147: we are just loading answer store from global to pass it into questionnaire_manager. should we just load it in questionnaire_manager?

tests/app/data_model/test_answer_store.py
```diff
@@ -267,3 +267,93 @@ def test_returns_ordered_map(self):
...
+        )
+
+        self.store.add(answer_1)
+        self.store.add(answer_2)
+
+        expected_answers = {
+            "2_1": 25,
+            "5_1": 65,
+        }
+        self.assertEqual(self.store.map(), expected_answers)
```
benjefferies: Do we need this assertion, does that mean this test is testing the add and remove functionality?

app/main/views/questionnaire.py
```diff
@@ -30,8 +30,13 @@
                                     import_name=__name__,
                                     url_prefix='/questionnaire/<eq_id>/<form_type>/<collection_id>/')
 
+action_blueprint = Blueprint(name='action',
```
benjefferies: Would it be cleaner if we have the action_blueprint in another file?

app/questionnaire/questionnaire_manager.py
```diff
@@ -125,3 +126,44 @@ def get_schema_item_by_id(self, item_id):
...
+        new_answer.schema_item = new_answer_schema
+        new_answer.parent = question_state
+        new_answer.instance = next_instance_id
+        return new_answer
+
+    def remove_answer(self, block, post_data, answer_store):
+        if self.state is None:
+            self.process_incoming_answers(block, post_data)
+
+        index_to_remove = post_data.get('action[remove_answer]')
```
jonnyshaw89: maybe we can pass the index through to this function? IMO references to the name of a button should probably by up at the 1st level rather than being extracted lower down.

app/questionnaire_state/state_item.py
```diff
@@ -38,3 +38,10 @@ def get_errors(self):
 
         logger.debug("Item errors list is %s", errors)
         return errors
+
+    def find_state_item(self, schema_item):
+        for child in self.children:
+            if child.schema_item == schema_item:
+                return child
+            else:
+                return child.find_state_item(schema_item)
```
jonnyshaw89, line 48: what if none of the children contain the schema_item?

app/questionnaire/questionnaire_manager.py
```diff
@@ -125,3 +126,44 @@ def get_schema_item_by_id(self, item_id):
 
     def get_schema(self):
         return self._schema
+
+    def add_answer(self, block, question, post_data, answer_store):
+        if self.state is None:
```
benjefferies: Should we call process_incoming_answers in questionnaire.py as that would be consistent with what we are currently doing

app/questionnaire/questionnaire_manager.py
```diff
@@ -125,3 +126,44 @@ def get_schema_item_by_id(self, item_id):
...
+        next_instance_id = 0 if len(last_answer) == 0 else int(last_answer[0]['answer_instance']) + 1
+        new_answer_schema = copy.deepcopy(new_answer.schema_item)
+        new_answer_schema.widget.name += '_' + str(next_instance_id)
+        new_answer.schema_item = new_answer_schema
+        new_answer.parent = question_state
+        new_answer.instance = next_instance_id
+        return new_answer
+
+    def remove_answer(self, block, post_data, answer_store):
+        if self.state is None:
```
benjefferies: Should we call process_incoming_answers in questionnaire.py as that would be consistent with what we are currently doing

app/questionnaire_state/state_question.py
```diff
@@ -6,3 +8,54 @@ def __init__(self, id, schema_item):
...
+            self.answers.remove(answer)
+
+    def update_state(self, user_input):
+
+        if self.schema_item.type == 'RepeatingAnswer':
+            state_answers = self.children[:]
+
+            for child in state_answers:
+                answer_keys = [key if key.startswith(child.schema_item.id) else None for key in user_input]
+                answer_instances = sorted(list(filter(None, answer_keys)))
```
jonnyshaw89: What is this line doing? filtering by nothing?

app/templates/partials/answer.html
jonnyshaw89, line 5: Instead of `if invalid` can we not do something like `if not answer.is_valid`?
Im not sure if thats possible

app/questionnaire_state/state_question.py
```diff
@@ -6,3 +8,54 @@ def __init__(self, id, schema_item):
...
         self.children = self.answers
+
+    def remove_answer(self, answer):
+        if answer in self.answers:
+            self.answers.remove(answer)
+
+    def update_state(self, user_input):
+
+        if self.schema_item.type == 'RepeatingAnswer':
+            state_answers = self.children[:]
```
benjefferies: Maybe this refactored into a method with a name like build_repeating_answers or something would give that extra bit of context as there is quite a bit of logic here

jonnyshaw89 (changes requested):

app/questionnaire_state/state_question.py
```diff
@@ -6,3 +8,54 @@ def __init__(self, id, schema_item):
...
+
+            for child in state_answers:
+                answer_keys = [key if key.startswith(child.schema_item.id) else None for key in user_input]
+                answer_instances = sorted(list(filter(None, answer_keys)))
+                num_instances = len(answer_instances)
+
+                if num_instances > 1:
```
benjefferies: Do we need `if num_instances > 1:`? Wouldn't it just iterate once and call child.update_state

app/submitter/converter.py
```diff
@@ -88,7 +88,13 @@ def prepare_answers(metadata_store, questionnaire, answers):
             if item is not None:
                 value = answers[key]
                 if value is not None:
-                    data[item.code] = Converter._encode_value(value)
+                    if item.code not in data:
```
benjefferies, line 92: After a quick discussion with @ajmaddaford it might makes sense to not change anything here until we know the format we send this to the downstream system in

app/questionnaire/questionnaire_manager.py
```diff
@@ -125,3 +126,44 @@ def get_schema_item_by_id(self, item_id):
 
     def get_schema(self):
         return self._schema
+
+    def add_answer(self, block, question, post_data, answer_store):
```
ajmaddaford: It would be clearer to rename `block` to `block_id` and `question` to `question_id` to make it clear they are id's and not objects

app/questionnaire/questionnaire_manager.py
```diff
@@ -125,3 +126,44 @@ def get_schema_item_by_id(self, item_id):
...
+        if self.state is None:
+            self.process_incoming_answers(block, post_data)
+
+        question_schema = self._schema.get_item_by_id(question)
+        question_state = self.state.find_state_item(question_schema)
+
+        answer_schema = question_schema.answers[0]  # Single answer for now.
```
ajmaddaford: I think renaming `answer_schema` to `existing_answer_schema` would make things clearer

app/templates/partials/question.html
benjefferies: Not sure how I feel able this, maybe we should just hard code the url with the dynamic parts substituted or create a add/remove variable in the context

app/questionnaire/questionnaire_manager.py
```diff
@@ -125,3 +126,44 @@ def get_schema_item_by_id(self, item_id):
...
+        new_answer = self._create_new_answer(answer_schema, question_state, answer_store)
```
ajmaddaford: I think renaming `new_answer` to `new_answer_state` and `_ create_new_answer` to `_create_new_answer_state` would make it clear we are creating a new state item here.

app/templates/partials/question.html
```diff
@@ -24,13 +24,55 @@ <h3 class="question__subtitle mars">{{question_subtitle}}</h3>
...
+     {%- endfor -%}
+   {%- endif %}
+ </div>
+{% endset %}
+
+{% set question_actions %}
+  {%- if question.schema_item.type == 'RepeatingAnswer' %}
+  <div class="question__actions">
+    <button class="btn btn--border btn--secondary venus js-household-btn" type="submit"
+          formaction="{{ url_for('action.add_answer',
```
benjefferies: Not sure how I feel able this, maybe we should just hard code the url with the dynamic parts substituted or create a add/remove variable in the context

app/questionnaire/questionnaire_manager.py
```diff
@@ -125,3 +126,44 @@ def get_schema_item_by_id(self, item_id):
...
+
+        question_answers = question_state.children
+        question_answers.append(new_answer)
```
ajmaddaford: Can we just do `question_state.answers.append(new_answer_state)`?

tests/app/data_model/test_answer_store.py
```diff
@@ -267,3 +267,93 @@ def test_returns_ordered_map(self):
...
+
+        self.store.remove(answer_2)
+        expected_answers = {
+            "2_1": 25,
+        }
+        self.assertEqual(self.store.map(), expected_answers)
+
+    def test_remove_answer_(self):
```
benjefferies: This is the same as above?

app/templates/partials/answer.html
hamishtaplin: @jonnyshaw89 That `invalid` value is used in two places so this caches the result so the logic isn't repeated.

tests/app/questionnaire/test_questionnaire_manager.py
```diff
@@ -0,0 +1,145 @@
...
+
+        self.answer_store.filter = MagicMock(return_value=existing_answers)
+        # When
+        new_answer = self.questionnaire_manager._create_new_answer(answer_schema, question_state, self.answer_store)
+
+        # Then
+        self.assertEqual(new_answer.schema_item.widget.name, 'answer_2')
+        self.assertEqual(new_answer.instance, 2)
+
+    def setUp(self):
```
benjefferies, line 115: Minor point but it's a bit confusing reading the tests with this at the bottom as you have references to things you don't expect to exist but that might be a preference thing

tests/app/questionnaire_state/test_state_item.py
```diff
@@ -0,0 +1,44 @@
...
+        state_item = StateItem('item_id', schema_item)
+        state_item.children = [child]
+
+        # When
+        result = state_item.find_state_item(search_schema)
+
+        # Then
+        self.assertEqual(result, child)
+
+    def test_find_schema_item_recursive_call_if_not_found(self):
```
benjefferies: Test names might be nicer test_find_schema_item_in_question and test_find_schema_item_in_block as I think that's what it's testing right? Again minor (possibly preference)

tests/app/questionnaire_state/test_state_question.py
```diff
@@ -0,0 +1,152 @@
...
+
+class TestStateQuestion(unittest.TestCase):
+
+    def test_remove_answer_single_answer(self):
+        answer = MagicMock()
+        question = StateQuestion('question_id', MagicMock())
+
+        question.answers.append(answer)
+
+        self.assertTrue(answer in question.answers)
```
benjefferies, line 19: Should these assertions be another test?

tests/app/questionnaire_state/test_state_question.py
```diff
@@ -0,0 +1,152 @@
...
+
+    def test_remove_answer_multiple_answers(self):
+        answer1 = MagicMock()
+        answer2 = MagicMock()
+
+        question = StateQuestion('question_id', MagicMock())
+        question.answers.append(answer1)
+        question.answers.append(answer2)
+
+        self.assertEqual(len(question.answers), 2)
```
benjefferies: Again is this not testing the add functionality therefore be another test?

tests/app/questionnaire_state/test_state_question.py
```diff
@@ -0,0 +1,152 @@
...
+
+        question.update_state({
+            'answer_one': 'answer_one_value',
+            'answer_one_1': 'answer_one_value_1',
+            'answer_two': 'answer_two_value',
+        })
+
+        self.assertEqual(len(question.children), 3)
+
+        question.answers.clear()
```
benjefferies: Does the second part of this test below test anything different therefore needed?

tests/functional/pages/surveys/household_composition/summary.page.js
```diff
@@ -0,0 +1,36 @@
...
+  clickEdit() {
+    browser.click('[data-qa="414699da-1667-44fd-8e98-7606966884db-edit"]')
+    return this
+  }
+
+  isNameDisplayed(name) {
+    return browser.element('[data-qa="414699da-1667-44fd-8e98-7606966884db-answer"]').getText() === name
+  }
+
+  areNamesDisplayed(names) {
```
benjefferies: It might be nicer if this was getHouseholdNames with a contains assertion instead of us writing that logic ourself

tests/functional/spec/household-composition.spec.js
```diff
@@ -0,0 +1,114 @@
...
+    startQuestionnaire(household_composition_schema)
+
+    //When
+    HouseholdCompositionPage.setPersonName(0, 'Person One').submit()
+
+    // Then
+    expect(HouseholdCompositionSummary.isNameDisplayed('Person One')).to.be.true
+  })
+
+  it('Given no people added, when enter a name and add another person, then there should be two input fields displayed.', function() {
```
benjefferies: Given a name has been added, when I enter another name?

tests/functional/spec/household-composition.spec.js
```diff
@@ -0,0 +1,114 @@
...
+        .setPersonName(0, 'Person One')
+        .addPerson()
+        .setPersonName(1, 'Person Two')
+        .submit()
+
+    // Then
+    expect(HouseholdCompositionSummary.areNamesDisplayed(['Person One', 'Person Two'])).to.be.true
+  })
+
+  it('Given three people added, when submitted, all three names should appear on summary.', function() {
```
benjefferies, line 37: Maybe we don't need this test as it doesn't exercise anything more than `Given two people added, when submitted, both names should appear on summary.`

tests/functional/spec/household-composition.spec.js
```diff
@@ -0,0 +1,114 @@
...
+
+    // When
+    HouseholdCompositionSummary.clickEdit()
+    HouseholdCompositionPage.removePerson(1).submit()
+
+    // Then
+    expect(HouseholdCompositionSummary.areNamesDisplayed(['Person One'])).to.be.true
+  })
+
+   it('Given three people added, when I remove second person, first and third person should appear on summary.', function() {
```
benjefferies, line 82: Does this exercise anymore logic than `Given two people added, when I remove second person, only first person should appear on summary.` does?

benjefferies (changes requested):

app/questionnaire_state/state_item.py
benjefferies: Python has an implicit none return but should we be explicit there is a valid question

app/questionnaire/questionnaire_manager.py
```diff
@@ -125,3 +126,44 @@ def get_schema_item_by_id(self, item_id):
...
+
+        existing = answer_store.filter(answer_id=answer_schema.id)
+        last_answer = existing[-1:]
+        next_instance_id = 0 if len(last_answer) == 0 else int(last_answer[0]['answer_instance']) + 1
+        new_answer_schema = copy.deepcopy(new_answer.schema_item)
+        new_answer_schema.widget.name += '_' + str(next_instance_id)
+        new_answer.schema_item = new_answer_schema
+        new_answer.parent = question_state
+        new_answer.instance = next_instance_id
+        return new_answer
```
ajmaddaford: Can we adapt the `_add_new_answer` method in the `StateQuestion` class to create the new answer state item?

app/questionnaire_state/state_question.py
```diff
@@ -6,3 +8,54 @@ def __init__(self, id, schema_item):
...
+
+        instance_id = self._get_instance_id(instances, answer_schema.id, index)
+        new_answer.instance = instance_id
+
+        new_answer_schema = copy.deepcopy(answer_schema)
+        new_answer_schema.widget.name += '_' + str(instance_id) if instance_id > 0 else ''
+        new_answer.schema_item = new_answer_schema
+
+        self.children.append(new_answer)
+        return new_answer
```
ajmaddaford: If this method was modified to create a new answer state (rather than create and append) , could this method be moved to `StateAnswer` class?

ajmaddaford (changes requested):

app/assets/js/app/main.js
samiwel: @jonnyshaw89 This needs to remain commented out for now. It should be uncommented when we work on **Household Composition Entry: Realtime Changes** story.

app/main/views/questionnaire.py
samiwel: The reason for passing it in is that it is easier to use a mock answer store during unit tests. If you can show me how to mock the global call to get_answer_store within questionnaire_manager.add_answer, then I'm happy to move it.

app/questionnaire_state/state_question.py
samiwel: Filters out answers starting with the answer Id from the schema. The filter on None discards values in user_input that are not related to the current child.

app/data/test_household_question.json
samiwel: Yes, probably if we're deciding to no longer use GUIDs.

app/assets/js/app/main.js
samiwel: See https://github.com/ONSdigital/eq-survey-runner/pull/563#discussion_r88448316

tests/app/data_model/test_answer_store.py
samiwel: The first assert is probably superfluous as I think there is already a separate test for the add. All I'm doing here is asserting that we did have 2 answers and after the remove we only have one.

app/main/views/questionnaire.py
samiwel: Not sure if we need the blueprint at all. There's been some discussion about having a single endpoint to avoid using HTML5 formaction attribute in the template. If I re-work this so that it posts back to the same endpoint then we may just be able to use the questionnaire endpoint or some variation on it.

app/questionnaire_state/state_question.py
samiwel: Yeah, I think you're right. I'll remove it.

app/submitter/converter.py
samiwel: Had to do something or you get an error when you try to submit. Happy to have a discussion about what the downstream format should/needs to be.

tests/app/questionnaire_state/test_state_item.py
samiwel: Agreed, changed.

tests/functional/spec/household-composition.spec.js
samiwel: It just catches the case where you remove an item from the middle. There was a bug with this during the initial spike, so I just wanted to make sure it was covered.

app/questionnaire_state/state_question.py
samiwel: Actually, I still think we need this. Imagine what happens in the case of {} being passed in.

tests/functional/spec/household-composition.spec.js
benjefferies: That's fair enough. That sounds like it could be caught at unit test level. I am worried we might start using selenium to test logic which could be tested in a more isolated, faster way

app/questionnaire_state/state_question.py
benjefferies: Would num_instances not be 0 then and it wouldn't iterate anyway?

app/data/test_household_question.json
ajmaddaford: Yes, but use hyphens rather than underscores

app/questionnaire/questionnaire_manager.py
samiwel: Changed.

app/questionnaire/questionnaire_manager.py
samiwel: Changed

app/questionnaire/questionnaire_manager.py
samiwel: Changed

app/questionnaire/questionnaire_manager.py
samiwel: Changed

app/questionnaire_state/state_question.py
samiwel: I'll give it a test it might be okay to remove it after all. Confusing myself because it's changed to build the state on update_state now rather than construct state.

tests/functional/spec/household-composition.spec.js
samiwel: I'll try put together an unit test to cover this functionality.

benjefferies:
Not sure if it was intended but the focused element after adding a person is the remove person so if you tap enter to submit you remove a person

app/questionnaire/questionnaire_manager.py
samiwel: Moved this logic into questionnaire.py.

app/questionnaire_state/state_question.py
samiwel: Done.

tests/app/questionnaire_state/test_state_question.py
samiwel: I've split this out into two tests as they're two different scenarios. Second one will come into play when middle and last names are added in the next story, I was thinking ahead it seems!

collisdigital:
With respect to errors; check with Louise, but I believe there shouldn't be any; i.e. you can enter nothing and still proceed or click Add person and leave blank and that's fine. However, since this PR is huge perhaps don't change any behaviour but add a new card after discussing with Louise.

app/templating/summary/question.py
```diff
@@ -24,13 +24,20 @@ def _build_answers(cls, block_id, question_schema, answer_schema, answers):
         summary_answers = []
         answers_iterator = iter(answer_schema)
         for answer_schema in answers_iterator:
-            answer = cls._build_answer(question_schema, answer_schema, answers, answers_iterator)
-            summary_answers.append(Answer(block_id, answer_schema, answer))
+            if question_schema['type'] == 'RepeatingAnswer':
+                for k, v in answers.items():
```
benjefferies: If we don't use `v` I think the pythonic way of ignoring it is with an underscore `k, _`

app/questionnaire/questionnaire_manager.py
samiwel: Done.

app/questionnaire_state/state_item.py
samiwel: I've made this more explicit so that it returns None if schema item is not found.

tests/app/data_model/test_answer_store.py
samiwel: Removed duplicate test.

tests/app/questionnaire_state/test_state_question.py
samiwel: I've split these out into two separate tests.

tests/app/questionnaire_state/test_state_question.py
samiwel: I've split these out into two separate tests.

tests/functional/spec/household-composition.spec.js
samiwel: I've removed the two person one and stuck with the 3 people test.

tests/functional/spec/household-composition.spec.js
samiwel: I've re-worded this.

app/templates/partials/answer.html
samiwel: I've modified the backend code to return unique Id's for each instance of an answer when it renders the widget.

app/data/test_household_question.json
```diff
@@ -0,0 +1,74 @@
...
+                        }
+                    ],
+                    "title": "Household",
+                    "validation": []
+                }
+            ],
+            "display": {
+                "properties": {}
+            },
+            "id": "multiple_questions_group",
```
collisdigital: Should be multiple-questions-group with - not _

app/main/views/questionnaire.py
```diff
@@ -119,13 +115,44 @@ def submit_answers(eq_id, form_type, collection_id):
...
+        return redirect_to_questionnaire_page(eq_id, form_type, collection_id, 'household-composition')
+
+    if not valid:
+        return _render_template('household-composition', questionnaire_manager.state, template='questionnaire')
+
+    return _go_to_next_block(current_location='household-composition', eq_id=eq_id,
+                             form_type=form_type, collection_id=collection_id)
+
+
+def _go_to_next_block(current_location, eq_id, form_type, collection_id, ):
```
collisdigital: Trailing comma, in args list.

app/questionnaire/questionnaire_manager.py
```diff
@@ -125,3 +125,29 @@ def get_schema_item_by_id(self, item_id):
...
+
+        answer_schema = question_schema.answers[0]  # Single answer for now.
+
+        next_answer_instance_id = self._get_next_answer_instance(answer_store, answer_schema)
+        answer_schema.create_new_answer_state(answer_instance=next_answer_instance_id, parent=question_state)
+
+        self.update_questionnaire_store(block_id)
+
+    @staticmethod
+    def _get_next_answer_instance(answer_store, existing_answer_schema):
```
ajmaddaford: We shouldn't be passing an `existing_answer_schema` object when we only need the `answer_id`. Also, I think we should move this method to the AnswerStore.

app/schema/questionnaire.py
```diff
@@ -28,10 +28,13 @@ def add_group(self, group):
             group.container = self
 
     def get_item_by_id(self, item_id):
+        answer_without_instance_id = item_id.split('_')[0]
         if item_id == self.id:
             return self
         elif item_id in self.items_by_id.keys():
             return self.items_by_id[item_id]
+        elif answer_without_instance_id in self.items_by_id.keys():
+            return self.items_by_id[answer_without_instance_id]
```
ajmaddaford: Do we make use of this anywhere? The only place I can see we pass an answer with an instance id is in the tests. The concern for me here is no item id's exist in the schema with an instance id, if we need to find an item with an instance id we should remove it before calling this method.

app/schema/widget.py
```diff
@@ -18,3 +17,8 @@ def get_user_input(self, post_vars):
 
     def get_other_input(self, post_vars):
         pass
+
+    @staticmethod
+    def get_id(state):
+        schema_id_with_suffix = '_'.join([state.schema_item.id, str(state.answer_instance)])
+        return state.schema_item.id if state.answer_instance == 0 else schema_id_with_suffix
```
ajmaddaford: Can we set the widget id at the same time we create the widget name?

app/parser/v0_0_1/schema_parser.py
```diff
@@ -74,6 +74,7 @@ def __init__(self, schema):
         self.question_factory.register_all({
             'GENERAL': GeneralQuestion,
             'DATERANGE': DateRangeQuestion,
+            'REPEATINGANSWER': GeneralQuestion,
```
ajmaddaford: We should create a specific question type class for this

ajmaddaford (changes requested):

app/templates/partials/question.html
```diff
@@ -24,13 +24,45 @@ <h3 class="question__subtitle mars">{{question_subtitle}}</h3>
...
 {% set question_answers %}
-  <div class="question__answers">
-    {%- for answer in question.answers -%}
-      {% include theme('partials/answer.html') %}
-    {%- endfor -%}
+ <div class="question__answers">
+   {%- if question.schema_item.type == 'RepeatingAnswer' %}
+     {%- for answer in question.answers %}
+       <div class="question__answer js-household-person">
+         <h3 class="neptune">Person <span class="js-household-loopindex">{{ loop.index }}</span>
```
ajmaddaford: This should be `_('Person ')`

app/questionnaire_state/state_question.py
```diff
@@ -1,8 +1,54 @@
+import re
+
+from app.data_model.answer_store import natural_order
 from app.questionnaire_state.state_item import StateItem
 
 
 class StateQuestion(StateItem):
     def __init__(self, id, schema_item):
-        super().__init__(id=id, schema_item=schema_item)
+        super(StateQuestion, self).__init__(id=id, schema_item=schema_item)
```
collisdigital: can be `super().__init__(id=id, schema_item=schema_item)` in Python3

app/questionnaire_state/state_question.py
```diff
@@ -1,8 +1,54 @@
...
+        if answer in self.answers:
+            self.answers.remove(answer)
+
+    def update_state(self, user_input):
+        if self.schema_item.type == 'RepeatingAnswer':
+            self.build_repeating_state(user_input)
+            for state_answer in self.answers:
+                state_answer.update_state(user_input)
+        else:
+            super(StateQuestion, self).update_state(user_input)
```
collisdigital: Can be `super().update_state(user_input)` in Python3

app/questionnaire_state/state_question.py
```diff
@@ -1,8 +1,54 @@
...
+                    if new_state:
+                        new_state.parent = self
+                        self.answers.append(new_state)
+                    break
+
+    @classmethod
+    def _iterate_over_instance_ids(cls, user_input):
+
+        answer_instance_ids = [key for key in user_input]
+        answer_instance_ids = sorted(answer_instance_ids, key=natural_order)
```
collisdigital: These two lines can be replaced with:

``` python
        answer_instance_ids = sorted(user_input.keys(), key=natural_order)
```

app/schema/questionnaire.py
```diff
@@ -28,10 +28,13 @@ def add_group(self, group):
             group.container = self
 
     def get_item_by_id(self, item_id):
+        answer_without_instance_id = item_id.split('_')[0]
```
collisdigital: Does this need to use the `_extract_answer_instance_id(...)` method; this simple split is likely to give false positives etc.

app/templates/partials/question.html
collisdigital: Even better, any way this could come from the Schema (probably one for a later PR)?

app/templating/summary/question.py
```diff
@@ -24,13 +24,20 @@ def _build_answers(cls, block_id, question_schema, answer_schema, answers):
         summary_answers = []
         answers_iterator = iter(answer_schema)
         for answer_schema in answers_iterator:
-            answer = cls._build_answer(question_schema, answer_schema, answers, answers_iterator)
-            summary_answers.append(Answer(block_id, answer_schema, answer))
+            if question_schema['type'] == 'RepeatingAnswer':
+                for k, _ in answers.items():
+                    if k.startswith(answer_schema['id']):
```
collisdigital: this simple check will suffer the same problems as the original code before `_extract_answer_instance_id` was introduced; could re-use `_iterate_over_instance_ids`; if it is modified to take user_input.keys() instead of user_input then it could also take answers.keys() here as well to get the same effect?

collisdigital:
97 comments, that century is in sight 💯!

app/questionnaire_state/state_question.py
collisdigital: at which stage it might be better to just pass user_input.keys() into this function and add a pydoc saying it expects to receive a list of raw answer_instance_ids; this ties up with a comment later in this review about re-using this function elsewhere.

app/questionnaire/questionnaire_manager.py
samiwel: Changed this so that it passes the Id in directly.

app/schema/questionnaire.py
samiwel: Removed this and stripping off the instance Id from the answer Id before this method is called.

app/main/views/questionnaire.py
```diff
@@ -125,13 +121,44 @@ def submit_answers(eq_id, form_type, collection_id):
...
+@questionnaire_blueprint.route('household-composition', methods=["POST"])
+@login_required
+def post_household_composition(eq_id, form_type, collection_id):
+    questionnaire_manager = get_questionnaire_manager(g.schema, g.schema_json)
+    answer_store = get_answer_store(current_user)
+    valid = questionnaire_manager.process_incoming_answers('household-composition', request.form)
+
+    if 'action[add_answer]' in request.form:
+        questionnaire_manager.add_answer('household-composition', 'question', answer_store)
+        return redirect_to_questionnaire_page(eq_id, form_type, collection_id, 'household-composition')
```
benjefferies: I think we could just return get_questionnaire(...) here to avoid the redirects as they will stay on the same url

app/main/views/questionnaire.py
```diff
@@ -125,13 +121,44 @@ def submit_answers(eq_id, form_type, collection_id):
...
+
+    elif 'action[remove_answer]' in request.form:
+        index_to_remove = request.form.get('action[remove_answer]')
+        questionnaire_manager.remove_answer('household-composition', answer_store, index_to_remove)
+        return redirect_to_questionnaire_page(eq_id, form_type, collection_id, 'household-composition')
```
benjefferies: I think we could just return get_questionnaire(...) here to avoid the redirects as they will stay on the same url

app/templates/partials/question.html
samiwel: I have changed this to  `_('Person ')` for now, happy to revisit in future if necessary.

app/data/test_household_question.json
samiwel: Changed

app/main/views/questionnaire.py
samiwel: Removed

app/questionnaire_state/state_question.py
samiwel: Changed

app/questionnaire_state/state_question.py
samiwel: Changed

app/schema/widget.py
samiwel: Added Id attribute to widget. Updated each widget's render method to reference self.id instead of using the schema item id.

app/templating/summary/question.py
```diff
@@ -24,13 +24,20 @@ def _build_answers(cls, block_id, question_schema, answer_schema, answers):
         summary_answers = []
         answers_iterator = iter(answer_schema)
         for answer_schema in answers_iterator:
-            answer = cls._build_answer(question_schema, answer_schema, answers, answers_iterator)
-            summary_answers.append(Answer(block_id, answer_schema, answer))
+            if question_schema['type'] == 'RepeatingAnswer':
+                for k, _ in answers.items():
```
benjefferies: I guess this could be just answers.keys()

tests/app/data_model/test_answer_store.py
samiwel: I split these out into two tests.

app/parser/v0_0_1/schema_parser.py
samiwel: @ajmaddaford I started doing this, but it doesn't feel like there's much benefit to splitting it out into it's own class. There's very little being done by the Question schema item class. so having another class which also doesn't do very much doesn't make much sense. What would be more useful would be to have a specific StateQuestion type for the repeating answer question, something like `RepeatingAnswerStateQuestion`, this would need to be created during the `construct_state` phase.

app/submitter/converter.py
ajmaddaford: This is fine for now, we need to refactor the submitter to use the answer store, but that can be done in another pull request.

ajmaddaford (approved):

tests/functional/pages/surveys/household_composition/summary.page.js
```diff
@@ -0,0 +1,36 @@
...
+  clickEdit() {
+    browser.click('[data-qa="household-full-name-edit"]')
+    return this
+  }
+
+  isNameDisplayed(name) {
+    return browser.element('[data-qa="household-full-name-answer"]').getText() === name
+  }
+
+  areNamesDisplayed(names) {
```
benjefferies: I think we are implementing complex logic here which is unnecessarily. We should use the chai testing framework to assert all values exist? i.e. return an array of names and assert those

tests/integration/household_composition/test_household_question.py
```diff
@@ -0,0 +1,152 @@
...
+
+        resp = self.navigate_to_page(first_page)
+        content = resp.get_data(True)
+        self.assertRegex(content, 'household-full-name_1')
+        self.assertRegex(content, 'household-full-name_2')
+        self.assertRegex(content, 'household-full-name_3')
+
+        return first_page
+
+    def test_regular_submit(self):
```
benjefferies: Can be deleted

samiwel: Do you mean the `test_regular_submit` test? or the `return` statement?
The test was added to cover each path through the `post_household_composition` endpoint.
Without the test the coverage drops below the acceptable level set by codecov.

samiwel: Removed.

jonnyshaw89 (approved):

benjefferies (approved):

hamishtaplin (approved):

samiwel:
Closing this PR, opening new PR #579 for a cleaner merge into master.

PR #567, opened 2016-11-16
added group_id and group_instance to answer store

jonnyshaw89:
We need to store answers with the group_id and group_instance so we can develop repeating groups functionality (used by relationship questions and household member personal details)

### How to review 
Check that AnswerStore stores group_id and group_instance for all answers

### DIFF
--- a/app/data_model/answer_store.py
+++ b/app/data_model/answer_store.py
@@ -4,22 +4,26 @@ from collections import OrderedDict
 
 
 class Answer(object):
-    def __init__(self, answer_id, block_id, value, answer_instance=0):
-        self.answer_id = answer_id
+    def __init__(self, group_id, block_id, answer_id, value, group_instance=0, answer_instance=0):
+        self.group_id = group_id
         self.block_id = block_id
+        self.answer_id = answer_id
+        self.group_instance = group_instance
         self.answer_instance = answer_instance
         self.value = value
 
     def matches(self, answer):
         """
         Check to see if two answers match.
-        Two answers are considered to match if they share the same block, answer and instance id.
+        Two answers are considered to match if they share the same block_id, answer, answer_instance, group_id and group_instance.
 
         :param answer: An answer to compare
         :return: True if both answers match, otherwise False.
         """
-        return self.block_id == answer.block_id and \
+        return self.group_id == answer.group_id and \
+            self.block_id == answer.block_id and \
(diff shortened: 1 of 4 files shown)


app/data_model/answer_store.py
```diff
@@ -20,7 +22,9 @@ def matches(self, answer):
         """
         return self.block_id == answer.block_id and \
             self.answer_id == answer.answer_id and \
-            self.answer_instance == answer.answer_instance
+            self.answer_instance == answer.answer_instance and \
+            self.group_id == answer.group_id and \
```
benjefferies: Docs need updating

benjefferies (changes requested):

benjefferies (approved):

app/data_model/answer_store.py
```diff
@@ -4,23 +4,27 @@
 
 
 class Answer(object):
-    def __init__(self, answer_id, block_id, value, answer_instance=0):
+    def __init__(self, answer_id, block_id, group_id, value, answer_instance=0, group_instance=0):
```
samiwel: This is only a small point, but semantically, does it make sense to order these in the same order that we'd expect them to appear in the schema e.g. group, block, answer?

iwootten: I'd say yes - in my branch I've updated the ones there to be kwargs, because entering a bunch of uuids into an answer makes things really difficult to read where answers are created.

jonnyshaw89: Done

samiwel (approved):

app/questionnaire_state/state_item.py
```diff
@@ -14,6 +14,7 @@ def __init__(self, id, schema_item):
         self.skipped = False
         self.display_on_summary = True
         self.instance = 0
```
ajmaddaford: We should rename this `answer_instance`, now we have a `group_instance`

app/questionnaire_state/state_answer.py
```diff
@@ -48,5 +48,7 @@ def flatten(self):
             block_id=self.parent.parent.parent.id,
             answer_id=self.id,
             answer_instance=self.instance,
```
ajmaddaford: `instance` -> `answer_instance`

ajmaddaford (changes requested):

ajmaddaford (approved):

PR #570, opened 2016-11-17
Add single date type question

LJBabbage:
Adding a single date type question; combining all the date tests into one Selenium test and renaming the Jinja templates when formatting date to be more explicit.

### How to review 
Using Dev, use test_dates.json and make sure the survey generates an error message if the date of birth question is either left empty of is an invalid date. Correct the error and then make sure the summary page works and presents the date correct and then submits.


tests/app/test_jinja_filters.py
```diff
@@ -1,26 +1,46 @@
...
+    def test_format_date(self):
         # Given
         date = datetime.strptime('01/01/17', '%d/%m/%y')
 
         # When
-        pretty_value = pretty_date(date)
+        format_value = format_date(date)
 
-        self.assertEquals(pretty_value, '1 January 2017')
+        self.assertEquals(format_value, '1 January 2017')
```
jonnyshaw89, line 19: Why does this give full month name and the others are short month names. maybe the filter names should suggest what you will get?

jonnyshaw89 (changes requested):

tests/app/test_jinja_filters.py
collisdigital: Please check with UX on this; I would assume full month names would be preferable to abbreviations (everywhere)?

LJBabbage: All changed to long months as agreed

app/jinja_filters.py
```diff
@@ -26,21 +26,26 @@ def nl2br(context, value):
...
 
 @blueprint.app_template_filter()
-def pretty_date(value):
+def format_date(value):
     return value.strftime('%-d %B %Y')
 
 
-def pretty_short_date(value):
-    return datetime.strptime(value, "%d/%m/%Y").strftime('%d %b %Y')
+def format_str_as_short_date(value):
```
benjefferies, line 33: Do we need the str_as part?

jonnyshaw89 (approved):

app/templates/partials/summary/date.html
```diff
@@ -1 +1,5 @@
-{{answer.value|pretty_date_range}}
+{% if question.type == 'DateRange' %}
```
benjefferies, line 1: It would be nice if we could pull the formatting out into different templates so we didn't have this logic in the template

benjefferies (changes requested):

app/jinja_filters.py
LJBabbage: Making it explicit due to the fact we have other Jinja filters formating dates that aren't strings

app/templates/partials/summary/date.html
LJBabbage: Due to a refactor of summary page after Census test, this was the agreed approach for now

app/data/test_dates.json
```diff
@@ -103,6 +103,34 @@
                                     "title": "Date with month and year",
                                     "type": "General",
                                     "validation": []
+                                    },
+                                    {
+                                    "answers": [
+                                        {
+                                            "display": {},
+                                            "guidance": "",
+                                            "id": "a51b1c7b-3f25-49dd-9275-2ea742eee510",
```
benjefferies, line 112: Shall we give the schema ids more descriptive names?

LJBabbage: Agree, but lets agree a format, time and a place to do it and raise a story to cover all the id's just not the odd one in a middle of a story

benjefferies: Boy scout rule?

LJBabbage: Happy to have a meeting with the team and discuss id structure for answers, blocks, question etc and the process we migrate them from what we have to that new structure and raise a card, but doing it for one answer in one of 20 schemas which all have GUID's seems a little out of scope of this story

benjefferies (approved):
Can you squash your commits down?

It might be worth setting up a meeting to address how we are going to name things in test schemas. I have just been going by common sense in the past as they are just test schemas and nothing public facing.

PR #571, opened 2016-11-17
eq-722 Non-mandatory dropdowns

hamishtaplin:

[Trello card](https://trello.com/c/As8A9W3K/722-5-dropdowns-non-mandatory-answer)

This PR enables users to progress beyond a non-mandatory question with a dropdown answer (`select`) inside by adding a basic "placeholder" functionality via a 'please select' type option.

If a question is mandatory and a user selects an option and then goes back or invokes an error, they are not allowed to the reselect the placeholder (it isn't displayed)/

If a question is non-mandatory, the user should be able to reselect the placeholder—effectively removing their answer.

### How to review 

TBA


app/schema/widgets/date_widget.py
```diff
@@ -57,6 +57,9 @@ def _get_month_field(self, month):
                     'text': 'Month',
                 },
                 'select': {
+                    'mandatory': False,   # This needs wiring up
```
hamishtaplin: This needs wiring up.

PR #572, opened 2016-11-17
Enabled config of werkzeug log level and reduce TRAVIS CI logging

collisdigital:
Travis builds logs are full of INFO logging from the app and werkzeug making it hard to see errors and test output. This change enables config of the "werkzeug" logging level (which was logging every GET) and sets both it and the normal app logging to WARNING for TRAVIS builds as well as reducing the selenium log output slightly.

I also moved login init outside of logging setup function as it looks to have been put there by mistake (logging/login??)

### How to review 
- Check Travis build log for this PR and confirm only WARNING/ERROR messages are displayed.
- Confirm if we're happy to not log INFO on TRAVIS before merging.

### DIFF
--- a/app/settings.py
+++ b/app/settings.py
@@ -48,6 +48,7 @@ EQ_NEW_RELIC_CONFIG_FILE = os.getenv('EQ_NEW_RELIC_CONFIG_FILE', './newrelic.ini
 EQ_SR_LOG_GROUP = os.getenv('EQ_SR_LOG_GROUP', os.getenv('USER', 'UNKNOWN') + '-local')
 EQ_LOG_LEVEL = os.getenv('EQ_LOG_LEVEL', 'INFO')
 EQ_CLOUDWATCH_LOGGING = parse_mode(os.getenv("EQ_CLOUDWATCH_LOGGING", 'True'))
+EQ_WERKZEUG_LOG_LEVEL = os.getenv('EQ_WERKZEUG_LOG_LEVEL', 'INFO')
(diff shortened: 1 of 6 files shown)


tests/functional/config.js
```diff
@@ -42,7 +42,7 @@ const sauceLabsConfig = {
 if (process.env.TRAVIS === 'true') {
   config = {
     ...config,
-    logLevel: 'debug',
+    logLevel: 'silent',
```
collisdigital, line 45: Note that `silent` is the default logLevel for WebDriver.IO http://webdriver.io/guide/getstarted/configuration.html

app/settings.py
```diff
@@ -48,6 +48,7 @@ def get_key(key_name):
 EQ_SR_LOG_GROUP = os.getenv('EQ_SR_LOG_GROUP', os.getenv('USER', 'UNKNOWN') + '-local')
 EQ_LOG_LEVEL = os.getenv('EQ_LOG_LEVEL', 'INFO')
 EQ_CLOUDWATCH_LOGGING = parse_mode(os.getenv("EQ_CLOUDWATCH_LOGGING", 'True'))
+EQ_WERKZEUG_LOG_LEVEL = os.getenv('EQ_WERKZEUG_LOG_LEVEL', 'INFO')
```
jonnyshaw89, line 51: This new environment variable should be added to the Readme

jonnyshaw89 (changes requested):

app/settings.py
collisdigital: Done

jonnyshaw89 (approved):

PR #575, opened 2016-11-18
Add non-mandatory date type

LJBabbage:

[Trello card](https://trello.com/c/As8A9W3K/722-5-dropdowns-non-mandatory-answer)

This PR enables users to progress beyond a non-mandatory question with a dropdown answer (`select`) inside by adding a basic "placeholder" functionality via a 'please select' type option.

If a question is mandatory and a user selects an option and then goes back or invokes an error, they are not allowed to the reselect the placeholder (it isn't displayed)/

If a question is non-mandatory, the user should be able to reselect the placeholder—effectively removing their answer.

### How to review
Using dev, open up test_dates. Using the bottom date (Non Mandatory), make sure it submits without being filled in, but generates date errors if partly filled in


LJBabbage:
Selenium tests will be fixed on Monday

yoganandkunche:
Passed QA on Branch. 
Waiting for confirmation of the error message from @jennyprice2 and @louisemorris5

app/data/test_dates.json
```diff
@@ -127,10 +127,38 @@
                                     "display": {
                                         "properties": {}
                                     },
-                                    "id": "question",
+                                    "id": "3b7d43d1-8e0e-499d-928e-7cd78be8525e",
```
jonnyshaw89, line 130: Why did we remove a text based id and replace it with a GUID given that our approach going forward is to use descriptive ids?

jonnyshaw89 (changes requested):

app/templates/partials/forms/select.html
```diff
@@ -1,12 +1,17 @@
 <select class='input input--select' id='input-{{select.name}}' name='{{select.name}}'>
+
+  {% if not (select.selected and select.mandatory) %}
+      <option value="">{{select.placeholder}}</option>
+  {% endif %}
+
```
ajmaddaford: I think we should be adding the placeholder text to the select.options in `date_widget.py`. If we need to do any logic about not displaying the placeholder text that can be done in the widget.

tests/functional/spec/dates.spec.js
```diff
@@ -15,18 +15,23 @@ describe('Date checks', function() {
```
ajmaddaford, line 15: I may have missed it, but I can't see a test for when the 'Please select' placeholder isn't in the month dropdown for mandatory dates once a month is selected.

ajmaddaford (changes requested):

app/templates/partials/forms/select.html
collisdigital: Can you check with a designer and Hamish about this behaviour; I thought the placeholder would always be in the list but just not selectable (disabled) for mandatory questions; i.e. should you actually be removing the placeholder?

tests/functional/spec/dates.spec.js
collisdigital: See previous comment

app/schema/widgets/date_widget.py
```diff
@@ -49,14 +48,17 @@ def _get_day_field(self, day):
...
 
         return {
                 'label': {
                     'for': self.name + '-month',
                     'text': 'Month',
                 },
                 'select': {
+                    'mandatory': is_mandatory,
+                    'placeholder': 'Select month',
+                    'selected': bool(month),
```
samiwel: This probably needs to be made translatable?

app/templates/partials/forms/select.html
LJBabbage: Hamish is happy. Non-mandatory always has a placeholder regardless if content is present or not. Mandatory has a place holder up until something has been entered then it gets removed

app/schema/widgets/date_widget.py
LJBabbage: Think we need to do this across the board, as all dates, labels etc are all non translated at the minute, I reckon a new story

app/data/test_dates.json
LJBabbage: Question was never a valid ID. With the potential for a question library in the future, we do need to name everything in a correct manor otherwise we will just cause more headaches down the line, at least GUID's are virtually unique. Lets agree a format, and then start doing it.

tests/functional/spec/dates.spec.js
LJBabbage: I have now put some unit tests in

yoganandkunche:
Following the error message update card Passed QA on Branch.

app/schema/widgets/date_widget.py
```diff
@@ -80,11 +80,18 @@ def _get_year_field(self, year):
...
             selected_month = int(selected_month)
 
         months = []
+
+        if not (selected_month and is_mandatory):
+            months.append(ObjectFromDict({
+                'value': '',
+                'text': 'Select month',
+            }))
+
```
ajmaddaford, line 92: I did some research for the accessibility of the placeholder text for the drop down and found http://ux.stackexchange.com/questions/90642/accessibility-of-disabled-placeholders-in-dropdown. Following that advice and changing the `if not` block to the code below improves the usability of this for screen readers. The placeholder text is disabled for mandatory dates, but the option is selected until a month is chosen. 
```
        months.append(ObjectFromDict({
            'value': '',
            'text': 'Select month',
            'disabled': True if is_mandatory else False,
            'selected': False if selected_month else True
        }))
```

ajmaddaford (changes requested):

ajmaddaford (approved):

jonnyshaw89 (approved):

jonnyshaw89:
@LJBabbage can you rebase and squash?

PR #576, opened 2016-11-20
Fixed most pylint issues

collisdigital:
Worked through most of the issues reported by Pylint in #574 and fixed them. Not all issues are resolved yet.

### How to review 
Review changes, decide if it is worth merging them in now given it will likely cause merge problems for everyone for a period due to number of files touched.

### DIFF
--- a/app/authentication/jwt_decoder.py
+++ b/app/authentication/jwt_decoder.py
@@ -19,15 +19,15 @@ CEK_EXPECT_LENGTH = 32
 
 
 class JWTDecryptor(JWERSAOAEPDecryptor):
-    '''
+    """
     JWT signed with JWS RS256 And encrypted with JWE RSA-OAEP
-    '''
+    """
     def __init__(self):
         self.logger = logging.getLogger(__name__)
         if settings.EQ_USER_AUTHENTICATION_RRM_PUBLIC_KEY is None or settings.EQ_USER_AUTHENTICATION_SR_PRIVATE_KEY is None \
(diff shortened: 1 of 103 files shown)


collisdigital:
Was planning to resolve the `Redefining built-in 'id'` in a later PR

app/authentication/jwt_decoder.py
```diff
@@ -72,15 +73,15 @@ def _check_header_values(self, header):
...
-        if "RS256" != header.get('alg').upper():
+        if header.get('alg').upper() != 'RS256':
             raise InvalidTokenException("Invalid Algorithm")
-        if "EDCRRM" != header.get('kid').upper():
+        if header.get('kid').upper() != 'EDCRRM':
             raise InvalidTokenException("Invalid Key Identifier")
 
     def _check_headers(self, token):
-        header_data, payload_data, signature_data = token.split('.', maxsplit=2)
+        header_data, _, _ = token.split('.', maxsplit=2)  # header_data, payload_data, signature_data
```
ajmaddaford: For this method and the `_check_payload` method it would be better to split the token before the method calls and only pass `header_data` or `payload_data` to the methods

ajmaddaford (changes requested):

app/authentication/jwt_decoder.py
collisdigital: Although this function also uses the token at the end as well; hmmm...

collisdigital:
PR Horribly Broken, sorry!

PR #578, opened 2016-11-22
Eq 725 Household Repeating Groups

iwootten:

Adds repeating group behaviour to both flask routes and the navigator. It's been necessary to update the navigator to return dict items so routes are able to be constructed appropriately. Navigator also now takes an answer store on initialisation, rather than answers within each method call. I've also removed the navigator from the questionnaire_manager, as it's possible to move some of the previous path queries to the flask route level. 

Several new flask endpoints are necessary in order route correctly, including a post to interstitial pages - we could get rid of this if interstitial pages simply linked to the next location, rather than submitted empty post data.

Another side effect is that links to previous on all pages and edit on the summary page are generated in the backend, due to the fact group context is required in order to support them. Previously a block id sufficed and simply the final segment of a url was needed.

### How to review 
Unit tests have been updated to use new structure, and new unit tests have been added within navigator to cover repeated groups. A new test_repeating_household.json schema has been added.


app/data_model/answer_store.py
```diff
@@ -4,7 +4,12 @@
 
 
 class Answer(object):
-    def __init__(self, group_id, block_id, answer_id, value, group_instance=0, answer_instance=0):
+    def __init__(self, group_id=None, block_id=None, answer_id=None, value=None, group_instance=0, answer_instance=0):
+        valid = (group_id or answer_id or block_id or value) is not None
```
collisdigital, line 8: should this be `and` otherwise if any of the values are not None valid will be True (is there a unit test to verify this?)

collisdigital: alternative something like ```if None in (group_id, answer_id, ...):```

app/questionnaire/navigator.py
```diff
@@ -1,192 +1,298 @@
...
+                "group_id": group['id'],
+                "group_instance": 0,
+                "block": block,
+            } for block in group['blocks']])
+
+            if 'routing_rules' in group:
+                for rule in group['routing_rules']:
+                    if 'repeat' in rule.keys():
+                        no_of_times = evaluate_repeat(rule['repeat'], self.answer_store)
+                        for i in range(no_of_times - 1):
```
collisdigital: since your doing `i + 1` below; perhaps `range(1, no_of_times)` would be better (and remove -1 and +1)?

app/data_model/answer_store.py
iwootten: Well, from the error message raised, this was the intended behaviour - but really it only makes sense that group_id and block_id are optional. I think this may have originally been to support filtering on the answer store.

collisdigital: Ah! I got it backwards, I thought "Missing one of" meant any one of them was missing causing it to be invalid, perhaps it could say "At least one of ... must be set"

app/data/test_repeating_household.json
```diff
@@ -0,0 +1,136 @@
...
+    "title": "Test Repeating",
+    "description": "",
+    "theme": "default",
+    "eq_id": "1",
+    "introduction": {
+        "description": ""
+    },
+    "groups": [{
+        "blocks": [{
+            "id": "980b148e-0856-4e50-9afe-67a4fa6ae13b",
```
benjefferies: UUID police!

app/main/views/questionnaire.py
```diff
@@ -47,6 +48,55 @@ def add_cache_control(response):
...
+
+    valid_location = this_block in navigator.get_location_path()
+    valid_data = q_manager.validate(block_id, request.form)
+
+    if not valid_location or not valid_data:
+        return _render_template(block_id, q_manager.state, template='questionnaire')
+    else:
+        q_manager.update_questionnaire_store(this_block)
+
+    next_location = navigator.get_next_location(current_group_id=group_id, current_iteration=group_instance, current_block_id=block_id)
```
benjefferies, line 89: Shall we rename to get_next_block?

app/main/views/questionnaire.py
```diff
@@ -47,6 +48,55 @@ def add_cache_control(response):
...
+
+    metadata = get_metadata(current_user)
+
+    next_block_id = None if next_location is None else next_location['block_id']
```
benjefferies, line 93: We should probably raise NotFound if next_location is not found?

app/main/views/questionnaire.py
```diff
@@ -47,6 +48,55 @@ def add_cache_control(response):
     return response
 
 
+@questionnaire_blueprint.route('<group_id>/<int:group_instance>/<block_id>', methods=["GET"])
```
benjefferies, line 51: I wonder if the url structure would be nicer with

```
@questionnaire_blueprint.route('<group_id>/<block_id>', methods=["GET"], defaults={'group_instance': 0})
@questionnaire_blueprint.route('<group_id>/<int:group_instance>/<block_id>', methods=["GET"])
def get_block(eq_id, form_type, collection_id, group_id, group_instance, block_id):
```

app/main/views/questionnaire.py
```diff
@@ -56,59 +106,71 @@ def get_introduction(eq_id, form_type, collection_id):
...
-    navigator = get_navigator(g.schema_json)
-    questionnaire_manager = get_questionnaire_manager(navigator, g.schema, g.schema_json)
+def post_interstitial(eq_id, form_type, collection_id, block_id):
+    navigator = Navigator(g.schema_json, get_answer_store(current_user))
+    q_manager = get_questionnaire_manager(g.schema, g.schema_json)
 
-    questionnaire_manager.build_state(location, get_answers(current_user))
+    this_block = {
+        'block_id': block_id,
+        'group_id': navigator.get_first_group_id(),
```
benjefferies, line 123: Will this always be the first group id?

app/main/views/questionnaire.py
```diff
@@ -127,61 +188,120 @@ def get_thank_you(eq_id, form_type, collection_id):
...
 
 
 def _delete_user_data():
     get_questionnaire_store(current_user.user_id, current_user.user_ik).delete()
     session_manager.clear()
 
 
-def redirect_to_questionnaire_page(eq_id, form_type, collection_id, location):
-    return redirect(url_for('questionnaire.get_questionnaire',
+def redirect_to_introduction(eq_id, form_type, collection_id):
```
benjefferies: This looks to be unused

app/main/views/questionnaire.py
```diff
@@ -127,61 +188,120 @@ def get_thank_you(eq_id, form_type, collection_id):
...
+
+def redirect_to_summary(eq_id, form_type, collection_id):
+    return redirect(interstitial_url(eq_id, form_type, collection_id, 'summary'))
+
+
+def redirect_to_thank_you(eq_id, form_type, collection_id):
+    return redirect(interstitial_url(eq_id, form_type, collection_id, 'thank-you'))
+
+
+def redirect_to_confirmation(eq_id, form_type, collection_id):
```
benjefferies, line 287: This looks to be unused

app/main/views/questionnaire.py
```diff
@@ -127,61 +188,120 @@ def get_thank_you(eq_id, form_type, collection_id):
...
                             form_type=form_type,
                             collection_id=collection_id,
-                            location=location))
+                            group_id=group_id,
+                            group_instance=group_instance,
+                            block_id=block_id))
+
+
+def interstitial_url(eq_id, form_type, collection_id, block_id):
+    if block_id == 'summary':
```
benjefferies, line 292: Could we just redirect like this?

```
return url_for('.get_' + block_id,
                       eq_id=eq_id,
                       form_type=form_type,
                       collection_id=collection_id,
                       )
```

app/main/views/questionnaire.py
```diff
@@ -195,8 +315,26 @@ def _render_template(location, context, rendered_schema_json=None, template=None
     metadata = get_metadata(current_user)
     answers = get_answers(current_user)
     metadata_context = build_metadata_context(metadata)
-    previous_location = get_navigator(g.schema_json).get_previous_location(answers, location)
+
+    navigator = Navigator(g.schema_json, get_answer_store(current_user))
+    previous_location = navigator.get_previous_location(current_block_id=location)
```
benjefferies: It feels a bit nasty that this returns None when the previous location is the introduction. Couldn't it return the introduction in that case?

app/main/views/root.py
```diff
@@ -68,13 +71,41 @@ def login():
...
-    navigator = get_navigator(json)
-    current_location = navigator.get_latest_location(get_answers(current_user), get_completed_blocks(current_user))
-
-    return redirect('/questionnaire/' + eq_id + '/' + form_type + '/' + collection_id + '/' + current_location)
+    json, _ = get_schema()
+
+    navigator = Navigator(json, get_answer_store(current_user))
+    current_location = navigator.get_latest_location(get_completed_blocks(current_user))
+
+    if current_location['block_id'] == 'introduction':
```
benjefferies, line 84: It might be tidy to condense to the below?

```
block = current_location['block_id']
if block in ['introduction'...]:
    return redirect(url_for('questionnaire.get_' + block,
                                 eq_id=eq_id,
                                 form_type=form_type,
                                 collection_id=collection_id))
```

app/questionnaire/navigator.py
```diff
@@ -1,193 +1,293 @@
...
+        filtered = answers.filter(answer_id=answer_index, group_instance=group_instance)
+        if len(filtered) == 1:
+            answer = filtered[0]
+            if evaluate_rule(goto_rule, answer['value']):
+                return True
+        else:
+            return False
+    elif 'id' in goto_rule.keys():
+        return True
+    return None
```
benjefferies: It would be better if we only had true false return values for a boolean evaluation

app/questionnaire/navigator.py
```diff
@@ -1,193 +1,293 @@
...
         """
-        answers = answers or {}
-        routing_path = routing_path or build_path(self.blocks, self.first_block_id, answers, [])
-        last_routing_block_id = routing_path[-1]
+        blocks = self.get_blocks()
+        routing_path = routing_path or self.build_path(blocks, self.first_group_id, 0, self.first_block_id, [])
+        last_routing_block_id = routing_path[-1]['block_id']
 
-        if self.blocks[-1]['id'] == last_routing_block_id:
+        if blocks[-1]['block']['id'] == last_routing_block_id:
```
benjefferies: blocks[-1]['block']['id'] might be clearer as summary_block_id = blocks[-1]['block']['id']

app/questionnaire/navigator.py
```diff
@@ -1,193 +1,293 @@
...
+            block_ids = [index for (index, path_item) in enumerate(location_path) if path_item['block_id'] == current_block_id]
+
+            # Return the index of the block id to be visited
+            current_location_index = block_ids[current_iteration]
 
             if current_location_index < len(location_path) - 1:
                 return location_path[current_location_index + 1]
 
-    def get_previous_location(self, answers=None, current_location_id=None):
+    def get_previous_location(self, current_group_id=None, current_block_id=None, current_iteration=0):
```
benjefferies, line 251: Lots of logic between get_previous_location and get_next_location is shared. We should refactor those bits

benjefferies (changes requested):

app/main/views/questionnaire.py
iwootten: It would require wrangling the block_id, given thank-you url is get_thank_you. I think this is clearer..

app/main/views/root.py
iwootten: As above...

app/main/views/questionnaire.py
iwootten: Yes :) The interstitial pages always are on the first group.

app/main/views/questionnaire.py
iwootten: It could, and I think previously would have without the group iteration behaviour before. Not sure why that method doesn't any longer. Will investigate.

app/main/views/questionnaire.py
benjefferies: Ah yes, didn't spot the thank-you difference

app/questionnaire/questionnaire_manager.py
```diff
@@ -25,40 +26,31 @@ class QuestionnaireManager(object):
...
-            if self.state:
-                # Todo, this doesn't feel right, validation is casting the user values to their type.
-                return self.state.schema_item.validate(self.state)
-            else:
-                # Item has node, but is not in schema: must be introduction, thank you or summary
-                return True
+        if self.state:
+            # Todo, this doesn't feel right, validation is casting the user values to their type.
+            return self.state.schema_item.validate(self.state)
         else:
```
benjefferies, line 41: I wonder if we should put a condition to exlude interstituatal pages from validation  instead of deriving it from the state. I think this would be more explicit. Something like
```
if location not in interstitual_pages
```

app/questionnaire/questionnaire_manager.py
```diff
@@ -67,7 +59,9 @@ def update_questionnaire_store(self, location):
         # Store answers in QuestionnaireStore
         questionnaire_store = get_questionnaire_store(current_user.user_id, current_user.user_ik)
 
-        for answer in self.get_state_answers(location):
+        for answer in self.get_state_answers(location['block_id']):
+            answer.group_id = location['group_id']
```
benjefferies, line 63: Should we be setting the group/group_instance state as a part of the construct_state logic instead of after the state has been created?

app/questionnaire/questionnaire_manager.py
```diff
@@ -125,15 +123,15 @@ def get_schema_item_by_id(self, item_id):
     def get_schema(self):
         return self._schema
 
-    def add_answer(self, block_id, question_id, answer_store):
+    def add_answer(self, location, question_id, answer_store):
```
benjefferies, line 126: I'm getting a bit confused between what we are calling block_id and location as they are one and the same from my understanding

tests/app/confirmation_page/test_confirmation_page.py
```diff
@@ -1,19 +1,21 @@
...
 from app.schema_loader.schema_loader import load_schema_file
 import unittest
 
 
 class TestConfirmationPage(unittest.TestCase):
 
-    def setUp(self):
-        survey = load_schema_file("0_rogue_one.json")
+    def test_get_next_location_confirmation(self):
+        current_block_id = "5e633f04-073a-44c0-a9da-a7cc2116b937"
```
benjefferies: This was confusing me until I realised it was the last block. It might be more readable as last_block_id

tests/app/data_model/test_answer_store.py
```diff
@@ -2,6 +2,55 @@
 from app.data_model.answer_store import Answer, AnswerStore
 
 
+class TestAnswer(unittest.TestCase):
+    def test_raises_error_on_invalid(self):
+
+        with self.assertRaises(ValueError) as ite:
+            answer = Answer()
```
benjefferies:
```
with self.assertRaises(ValueError) as ite:
    Answer()
```

Shouldn't create a unused variable and still work

tests/app/questionnaire/test_navigator.py
```diff
@@ -2,109 +2,151 @@
 
 from app.questionnaire.navigator import Navigator
 from app.schema_loader.schema_loader import load_schema_file
+from app.data_model.answer_store import Answer, AnswerStore
 
 
 class TestNavigator(unittest.TestCase):
 
     def test_first_block(self):
```
benjefferies, line 10: Maybe split out into test_first_block and test_first_group?

tests/app/questionnaire/test_navigator.py
```diff
@@ -114,188 +156,631 @@ def test_routing_conditional_path(self):
...
+            {
+                "block_id": "a7dcbb30-1187-4276-a49c-9284730ba4ed",
+                "group_id": "f22b1ba4-d15f-48b8-a1f3-db62b6f34cc0",
+                'group_instance': 2
+            }
+        ]
+
+        answer = Answer(
+            group_id="multiple-questions-group",
+            group_instance=0,
```
benjefferies, line 523: It doesn't affect the test but for accuracy I think this should be answer_instance

tests/app/questionnaire/test_navigator.py
```diff
@@ -114,188 +156,631 @@ def test_routing_conditional_path(self):
...
+        answers.add(answer)
+        answers.add(answer_2)
+
+        navigator = Navigator(survey, answers)
+
+        self.assertEqual(expected_previous_location, navigator.get_previous_location(current_group_id=current_group_id,
+                                                                                     current_block_id=current_block_id,
+                                                                                     current_iteration=current_iteration))
+
+    def test_repeating_groups_next_location(self):
```
benjefferies, line 642: Is this to test repeating a group only once so not repeating? Not sure how we could clarify that in the test name but might be good to as it might be a bit misleading

tests/app/templating/summary/test_section.py
```diff
@@ -18,6 +18,12 @@ def build_question_schema(id, answer_schemas):
 
 class TestSection(TestCase):
 
+    metadata = {
```
benjefferies: Looks unused?

tests/app/templating/summary/test_section.py
```diff
@@ -26,7 +32,7 @@ def test_create_section(self):
         section_schema = {'id': 'section_id', 'title': 'A section title', 'questions': [question_schema]}
 
         # When
-        section = Section('1', section_schema, answers)
+        section = Section(section_schema, answers, "some_link")
 
         # Then
         self.assertEqual(section.id, 'section_id')
```
benjefferies, line 32: We could add an assertion for the link self.assertEqual(section.link, "some_link")

app/questionnaire/navigator.py
iwootten: True, except it isn't the summary_block_id, simply the last block_id - updated to that effect.

app/questionnaire/questionnaire_manager.py
iwootten: Not any more :) I've been using location as the dictionary describing a group_id/group_instance/block_id and block_id as the, er, block_id.

app/main/views/questionnaire.py
benjefferies: To clarify if we weren't doing repeating we could have a URL like household/composition instead of household/0/composition

app/questionnaire/questionnaire_manager.py
benjefferies: Cool good to clarify!

app/main/views/questionnaire.py
iwootten: I think having two ways of getting to the same block is possibly asking for trouble...which url will we get when we call get_block in url_for?

app/questionnaire/navigator.py
```diff
@@ -1,193 +1,293 @@
...
+
+        if 'routing_rules' in block and len(block['routing_rules']) > 0:
+            for rule in block['routing_rules']:
+                if 'goto' in rule:
+                    should_go = evaluate_goto(rule['goto'], self.answer_store, group_instance)
+                    if should_go is True:
+                        return self.build_path(blocks, group_id, group_instance, rule['goto']['id'], path)
+                    elif should_go is False:
+                        return path
+        elif block_id_index != len(blocks) - 1:
```
ajmaddaford, line 113: It might be beneficial to add a comment to highlight this is 'if not last block'

app/questionnaire/navigator.py
```diff
@@ -1,193 +1,293 @@
...
         """
         Returns a list of the block ids visited based on answers provided
         :return: List of block ids
         """
-        answers = answers or {}
         routing_path = []
+        blocks = self.get_blocks()
 
-        return build_path(self.blocks, self.first_block_id, answers, routing_path)
+        return self.build_path(blocks, self.first_group_id, 0, self.first_block_id, routing_path)
```
ajmaddaford: Any reason for passing `routing_path` rather than `[]` (as done in the methods below)?

tests/app/questionnaire/test_navigator.py
iwootten: Really? Why? Note: The routing rule is updated as part of the test to be based on the value of the answer..

app/questionnaire/navigator.py
```diff
@@ -1,193 +1,288 @@
...
-                        return True
+                    return evaluate_goto(goto_rule, self.answer_store, 0)
         return False
 
-    def get_location_path(self, answers=None):
+    def get_location_path(self):
         """
         Returns a list of url locations visited based on answers provided
-        :param answers:
         :return:
```
ajmaddaford: I think it would be worthwhile adding to the docstring that this returns the routing_path with the preceding and closing interstitial pages added.

app/questionnaire/navigator.py
```diff
@@ -1,193 +1,288 @@
...
+
+        this_block = {
+            "block_id": current_block_id,
+            "group_id": current_group_id,
+            "group_instance": current_iteration,
+        }
+
+        if this_block in self.location_path:
+            return self.location_path.index(this_block)
+        return -1
```
ajmaddaford: Any reason for returning `None` rather than `-1`?

app/questionnaire/navigator.py
```diff
@@ -1,193 +1,288 @@
...
-        answers = answers or {}
-        location_path = self.get_location_path(answers)
-        if current_location_id in location_path:
-            current_location_index = location_path.index(current_location_id)
+        current_location_index = self._get_current_location_index(current_group_id, current_block_id, current_iteration)
 
-            if current_location_index < len(location_path) - 1:
-                return location_path[current_location_index + 1]
+        if current_location_index != -1 and current_location_index < len(self.location_path) - 1:
+            return self.location_path[current_location_index + 1]
```
ajmaddaford, line 248: There should be an explicit return here when the if condition isn't satisfied

app/questionnaire/navigator.py
```diff
@@ -1,193 +1,288 @@
...
-        answers = answers or {}
-        location_path = self.get_location_path(answers)
-        if current_location_id in location_path:
-            current_location_index = location_path.index(current_location_id)
+        current_location_index = self._get_current_location_index(current_group_id, current_block_id, current_iteration)
 
-            if current_location_index != 0:
-                return location_path[current_location_index - 1]
+        if current_location_index != -1 and current_location_index != 0:
+            return self.location_path[current_location_index - 1]
```
ajmaddaford, line 262: There should be an explicit return here if the if condition isn't satisfied

app/questionnaire/questionnaire_manager.py
ajmaddaford: We should, but I'd suggest not in this pull request. This behaviour matches what it does currently.

app/questionnaire/navigator.py
```diff
@@ -1,193 +1,288 @@
...
-        answers = answers or {}
-        location_path = self.get_location_path(answers)
-        if current_location_id in location_path:
-            current_location_index = location_path.index(current_location_id)
+        current_location_index = self._get_current_location_index(current_group_id, current_block_id, current_iteration)
 
-            if current_location_index != 0:
-                return location_path[current_location_index - 1]
+        if current_location_index > 0:
+            return self.location_path[current_location_index - 1]
```
ajmaddaford, line 262: There should be an explicit return here when the if condition isn't satisfied

ajmaddaford (changes requested):

app/questionnaire/navigator.py
iwootten: Huh? You mean -1 vs None? Simply that then we're only dealing with integer types returning from the method.

tests/app/confirmation_page/test_confirmation_page.py
iwootten: Sure, I can do that. This isn't a new test though - simply an updated one to use new navigator init approach.

app/questionnaire/navigator.py
```diff
@@ -1,193 +1,286 @@
...
+
+        if 'routing_rules' in block and len(block['routing_rules']) > 0:
+            for rule in block['routing_rules']:
+                if 'goto' in rule:
+                    should_go = evaluate_goto(rule['goto'], self.answer_store, group_instance)
+                    if should_go is True:
+                        return self.build_path(blocks, group_id, group_instance, rule['goto']['id'], path)
+                    elif should_go is False:
+                        return path
+        # If this isn't the last block in the set evaluated
```
benjefferies, line 112: Alternatively to adding a comment we could have refactored to a variable that won't go stale!

```
last_block = block_id_index != len(blocks) - 1
elif not last_block:
```

app/questionnaire/navigator.py
```diff
@@ -248,25 +244,27 @@ def get_next_location(self, current_group_id=None, current_block_id=None, curren
...
+        return None
 
     def get_previous_location(self, current_group_id=None, current_block_id=None, current_iteration=0):
         """
         Returns the next 'location' to visit given a set of user answers
         :param current_group_id:
         :param current_block_id:
-        :param current_iteration:
+        :return: The previous location as a dict
         :return:
```
ajmaddaford, line 257: The`current_iteration` param has been overwritten

app/questionnaire/navigator.py
iwootten: Comments only go stale if you ignore them :)

app/questionnaire/navigator.py
ajmaddaford: Hah, yeah, that's what I meant 😛  I think returning `None` says 'there is no current location index' more clearly than `-1`

app/main/views/questionnaire.py
iwootten: I think None is actually a valid response for previous/next links where we're looking at introduction or a summary page say. The template isn't using them, but the render method requires that they're specified.

ajmaddaford (approved):

jonnyshaw89 (approved):

benjefferies (approved):

PR #582, opened 2016-11-23
Fixes #581 Underscores in answer Id submission failure.

### LINKED ISSUE #581: Underscores in answer_ids cause submitter to fail
### Expected behaviour

A schema that uses underscores within answer_ids in the survey json is able to have them submitted as per standard behaviour or a schema that uses underscores fails gracefully (or is unable to be selected?) upon validation.

### Actual behaviour

Submission fails with 'app.schema.exceptions.QuestionnaireException: Unknown id ' message.

Create a survey using an answer_id with an underscore. Attempt to follow through the survey and submit.

### Technical information

This is failing because we use underscores to delimit between an answer_id and answer_instance when calling .map(). We should perhaps update the converter to use the answer_store, rather than the flattened list of answers. https://github.com/ONSdigital/eq-survey-runner/blob/master/app/submitter/converter.py#L87

samiwel:
Fixes #581 .

Submitter was doing too simplistic a match to determine the answer instance id.

### How to review 
Submission of answers containing _ should succeed.

### DIFF
--- a/app/submitter/converter.py
+++ b/app/submitter/converter.py
@@ -4,7 +4,7 @@ import logging
 from datetime import datetime, timezone
 
 from app import settings
-from app.questionnaire_state.state_repeating_answer_question import _extract_answer_instance_id
+from app.questionnaire_state.state_repeating_answer_question import extract_answer_instance_id
 
 
 logger = logging.getLogger(__name__)
@@ -86,7 +86,7 @@ class Converter(object):
         data = {}
 
         for key in answers.keys():
-            answer_id, _ = _extract_answer_instance_id(key)
+            answer_id, _ = extract_answer_instance_id(key)
             item = questionnaire.get_item_by_id(answer_id)
             if item is not None:
                 value = answers[key]
--- a/app/submitter/converter.py
+++ b/app/submitter/converter.py
@@ -4,6 +4,8 @@ import logging
 from datetime import datetime, timezone
 
 from app import settings
+from app.questionnaire_state.state_repeating_answer_question import _extract_answer_instance_id
+
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


app/submitter/converter.py
```diff
@@ -4,6 +4,8 @@
 from datetime import datetime, timezone
 
 from app import settings
+from app.questionnaire_state.state_repeating_answer_question import _extract_answer_instance_id
```
jonnyshaw89: the underscore at the start of _extract_answer_instance_id implies that this is a "private" function?

app/submitter/converter.py
```diff
@@ -84,8 +86,8 @@ def prepare_answers(metadata_store, questionnaire, answers):
         data = {}
 
         for key in answers.keys():
-            key_without_index = key.rsplit('_', 1)[0]
-            item = questionnaire.get_item_by_id(key_without_index)
+            answer_id, answer_index = _extract_answer_instance_id(key)
```
jonnyshaw89: answer_index is unused, Codacy will complain?

jonnyshaw89 (changes requested):

app/submitter/converter.py
samiwel: It did, resolved.

app/submitter/converter.py
iwootten: I like how the review of the thing to select the right underscore is to remove the underscore from the method name :)

samiwel: Renamed 👍

jonnyshaw89 (approved):

iwootten (commented):
LGTM 👍

iwootten (approved):

Issue #566, opened 2016-11-16
Changing radio option from Other fails


collisdigital:
### Expected behaviour
- After selecting Other and entering a value, changing to a different radio option updates my select.

### Actual behaviour
- After selecting Other and entering a value, any changes of radio selection are not stored unless the Other value is cleared manually.

### Steps to reproduce the behaviour
- Launch test_radio.json
- On "What is your favourite breakfast food" select "Other" and enter a value
- Select the "Eggs" radio option
- Save and continue
- See that the next question incorrectly shows the text entered into "Other" in the "Do you really want ____ for breakfast?"
- Click previous and see the wrong option selected

#### Browser
- Chrome

#### Operating System
- macOS Sierra

### Screenshot
- None

yoganandkunche:
Defect passed re-test in pre-prod

PR #583, opened 2016-11-24
Eq 731 other checkbox defects

### LINKED ISSUE #565: Unticking mandatory checkbox Other doesn't trigger an error
### Expected behaviour
- When only the "Other" checkbox is ticked in a checkbox field with a value entered; unticking the "other" option but not deleting the text causes a validation failure (question mandatory)

### Actual behaviour
- Unticking the "Other" checkbox but leaving the text in the value does not cause a validation failure; returning to the question and attempting to submit again (without making any changes) does cause the failure to occur.

- Launch test_checkbox.json
- Select "Other" checkbox on the "Mandatory other option" screen
- Enter some text into the "Please specify other" field
- Untick the "Other" checkbox but don't clear the "Please specify other" field
- Save and continue
- Observe no errors

**Issue 2:**
Select Other and enter some data in the text field
select two other checkboxes for the same question and
Click save and continue

**Expected:** Must navigate to the next Question

**Actual:**
Wipe the text entered in other field and throws an error ' This field is mandatory.'

#### Browser
- Chrome

#### Operating System
- macOS Sierra

### Screenshot
- None

LJBabbage:
Fixes #565. The get other was assuming that 'other' was in position 1 and the 'other value' in last position, which meant checkboxes would never work correctly if more then one was select. This PR changes that to search the list to find the value which isn't in the option list, which must be the other value. It also updates the checkbox validation to check for valid options, and valid 'other' values if required. The test suite for checkboxes has also been updated and the some of the Star wars test have been changed to align with the structure of all the rest.

### How to review 
Using test_checkbox.json
other_field: The text field that appears
other : the 'other' checkbox
**Mandatory question**
1. Try to submit with no values - (validation error)
2. Submit with just a pre-defined option (pass)
3. Tick other but leave other_field blank (validation error)
4. Add just other_field, now un-tick other (validation error)
5. Select pre-defined option, select other but leave other_field blank (validation error)
6. Select pre-defined option, put value in other_field and un-tick other. Navigate to end and make sure just the pre-defined option value is shown (pass)
7. Go back and now put value in pre-defined option and tick other and put value in other_field, navigate to the end and make sure both values appear. (pass)
8. Try to break it
**Non-Mandatory question**
1. Enter value in other_field, but untick other, the summary should show 'No answer provided'
2. Repeat as above but also select a pre-defined option, the summary page should just show the pre-defined option
3. Select pre-defined option, but also select other, put no text in other_field. Summary should show the pre-defined option and the words 'Other'
4. Select pre-defined option, select other and put value in other_field. Summary should show the pre-defined option and the value entered in the other_field.

re-test https://github.com/ONSdigital/eq-survey-runner/issues/566 and https://github.com/ONSdigital/eq-survey-runner/issues/565

### DIFF
--- a/app/schema/answer.py
+++ b/app/schema/answer.py
@@ -46,7 +46,7 @@ class Answer(Item):
 
         :returns a str for the Other value or None"""
         if self.type == 'Radio' or self.type == 'Checkbox':
-            user_input = self.widget.get_other_input(post_vars)
+            user_input = self.widget.get_other_input(post_vars, self.options)
             return self.check_user_input(user_input)
 
     @staticmethod
--- a/app/schema/answers/checkbox_answer.py
+++ b/app/schema/answers/checkbox_answer.py
@@ -19,19 +19,23 @@ class CheckboxAnswer(Answer):
     def validate(self, state):
         if isinstance(state, self.get_state_class()):
             question = state.parent
+            options = state.schema_item.options
             logger.debug("Checkbox Question is skipped %s", question.skipped)
             # Mandatory check
+
             if question.skipped:
                 state.is_valid = True
-            elif self.mandatory and state.input is None:
-                super(CheckboxAnswer, self).mandatory_error(state)
-            elif self.mandatory and state.input is not None and len(state.input) == 0:
-                super(CheckboxAnswer, self).mandatory_error(state)
-            elif self.mandatory and state.input is not None and 'other' in state.input and not state.other:
-                super(CheckboxAnswer, self).mandatory_error(state)
+            elif self.mandatory:
+                if not state.input:
+                    super(CheckboxAnswer, self).mandatory_error(state)
+                elif 'other' not in state.input and not self._valid_option_selected(state.input, options):
+                    super(CheckboxAnswer, self).mandatory_error(state)
+                elif 'other' in state.input and not self.widget.find_other_value(state.input, options):
+                    super(CheckboxAnswer, self).mandatory_error(state)
 
             # Here we just report on whether the answer has passed type checking
             state.value = self.get_typed_value(state.input)
+
             return state.is_valid
         else:
             raise StateException('Cannot validate - incorrect state class')
@@ -41,3 +45,14 @@ class CheckboxAnswer(Answer):
         return user_input if (Answer.check_user_input(user_input) and
                               user_input != [''] and
                               user_input != [None]) else None
+
+    @staticmethod
+    def _valid_option_selected(state_input, options):
+        # Check to see if the user input is one of the options in the schema
+        valid_option_selected = False
+        if state_input:
+            for answer in state_input:
+                if answer and any(option['value'] == answer for option in options) and answer != 'other':
+                    valid_option_selected = True
+                    break
+        return valid_option_selected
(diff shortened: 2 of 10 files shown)


iwootten:
All the scenarios listed above and #566, #565 seem to pass. Two observations though:

If you just enter just whitespace in either 'other' question type, it leads to empty entries in the summary screen.

Entering a value in 'other' for the mandatory field, then deselecting all options leads to the other field losing the content entered when the error is returned. Is that expected?

LJBabbage:
Thanks Ian, the first one needs fixing in this PR, I will try to fix the other one as well as I believe we should try to keep the data, but it seems to have always behaved like that, and it might need another issue raised as its more to do with how state is getting updated.

LJBabbage:
Agreed with Louise and Chris that "Entering a value in 'other' for the mandatory field, then deselecting all options leads to the other field losing the content entered when the error is returned." That this behaviour is fine for this story

app/schema/answer.py
```diff
@@ -100,3 +100,14 @@ def mandatory_error(self, state):
         state.errors = []
         state.errors.append(self.questionnaire.get_error_message('MANDATORY', self.id))
         state.is_valid = False
+
+    @staticmethod
+    def _valid_option_selected(state_input, options):
```
benjefferies: This can be localised to `app/schema/answers/checkbox_answer.py` as it's not used anywhere else

tests/app/templating/summary/test_question.py
```diff
@@ -136,6 +136,8 @@ def test_checkbox_button_other_option_empty(self):
         self.assertEqual(len(question.answers[0].value), 1)
         self.assertEqual(question.answers[0].value[0], 'Other option label')
 
+
```
benjefferies: Is it worth removing these two lines otherwise there will be an unnecessary touch on the file in the git history?

benjefferies (changes requested):

app/schema/answer.py
LJBabbage: Put it here after sitting in a meeting that said other option value will always be visible, this brings radio boxes into play and they will also need this

tests/app/templating/summary/test_question.py
LJBabbage: Sure can do

app/schema/answer.py
benjefferies: Even in that case I don't believe it's logic which belongs to an answer.py. This function would only apply to radio and checkbox answers. The functionality is not dependent on state and could go in a separate file at that point otherwise putting the logic in answer.py suggests this is functionality applicable to all answers?

LJBabbage: No worries, will move it

benjefferies: Cool, other than that it all looks good and you've made a good effort in backfilling the checkbox tests!

LJBabbage: Decided not to try to predict the future and put it in the only place it is currently used checkbox_answer.py

benjefferies (approved):
Great stuff lets squash and I'll press merge

iwootten:
Hold fire ben #596 relates to my original comment and has yet to be resolved.

benjefferies:
Ok @iwootten holding fire

iwootten (approved):
Cool this fixes #596 too also. Can we get the commits squashed into one for a merge?

PR #580, opened 2016-11-22
Added basic census schemas.

jonnyshaw89:
### What is the context of this PR?
Added basic census schemas.
Mainly Household survey but includes skeleton schemas for individual and communal establishment.

PR #586, opened 2016-11-25
Eq 737 census sexual id question

jonnyshaw89:
Added ability to evaluate metadata in routing rules and skip conditions to allow for sexual identity question to be conditionally asked and for welsh questions to be shown in place of alternative.

The branch is blocked until https://github.com/ONSdigital/eq-survey-runner/pull/580 is merged, But creating PR for visibility.

### How to review 
Launch the census household survey with and without the sexual identity flag and observe the sexual identity question being conditionally asked.
Launch the census household survey with welsh region code and see that welsh questions are displayed in place of alternate questions.


app/questionnaire/navigator.py
```diff
@@ -24,7 +24,7 @@ def evaluate_rule(rule, answer):
     return False
 
 
-def build_path(blocks, block_id, answers, path):
+def build_path(blocks, block_id, metadata, answers, path):
```
iwootten: The navigator has changed quite a bit for the repeating work - it no longer takes answers as method arguments and uses them when initialised. If we're having to evaluate metadata too, it probably makes sense for that to be an init arg too.

app/questionnaire/navigator.py
```diff
@@ -46,21 +46,55 @@ def build_path(blocks, block_id, answers, path):
...
-                return build_path(blocks, goto_rule['id'], answers, path)
+                return build_path(blocks, goto_rule['id'], metadata, answers, path)
     elif block_id_index != len(blocks) - 1:
         next_block_id = blocks[block_id_index + 1]['id']
-        return build_path(blocks, next_block_id, answers, path)
+        return build_path(blocks, next_block_id, metadata, answers, path)
     return path
 
 
+def evaluate_answer_rule(goto_rule, answers):
```
iwootten: Seems like we're getting to the point now that there are so many rules they should live in another module.

iwootten (changes requested):

PR #589, opened 2016-11-27
Add support for piping based on current group_instance Id.

samiwel:
Added support for piping based on group instance Id.

### How to review 
Start the household repeating questionnaire.
Enter some names for people in the household.
Observe that the names are displayed in the titles of the subsequent repeated groups.

### DIFF
--- a/app/main/views/questionnaire.py
+++ b/app/main/views/questionnaire.py
@@ -57,7 +57,7 @@ def get_block(eq_id, form_type, collection_id, group_id, group_instance, block_i
     answers = answer_store.map(group_id=group_id, group_instance=group_instance, block_id=block_id)
 
     q_manager = get_questionnaire_manager(g.schema, g.schema_json)
-    q_manager.build_state(block_id, answers)
+    q_manager.build_state(block_id, answers, group_instance)
 
     block = g.schema.get_item_by_id(block_id)
     template = block.type if block and block.type else 'questionnaire'
@@ -152,7 +152,7 @@ def get_summary(eq_id, form_type, collection_id):
 
     if latest_location['block_id'] is 'summary':
         metadata = get_metadata(current_user)
-        answers = get_answers(current_user)
+        answers = get_answer_store(current_user)
         schema_context = build_schema_context(metadata, g.schema.aliases, answers)
         rendered_schema_json = renderer.render(g.schema_json, **schema_context)
         summary_context = build_summary_rendering_context(rendered_schema_json, answer_store, metadata)
--- a/app/main/views/questionnaire.py
+++ b/app/main/views/questionnaire.py
@@ -218,15 +218,15 @@ def submit_answers(eq_id, form_type, collection_id):
                                   block_id=invalid_location['block_id']))
 
 
-@questionnaire_blueprint.route('multiple-questions-group/0/household-composition', methods=["POST"])
+@questionnaire_blueprint.route('<group_id>/0/household-composition', methods=["POST"])
 @login_required
-def post_household_composition(eq_id, form_type, collection_id):
+def post_household_composition(eq_id, form_type, collection_id, group_id):
     questionnaire_manager = get_questionnaire_manager(g.schema, g.schema_json)
     answer_store = get_answer_store(current_user)
 
     this_block = {
         'block_id': 'household-composition',
-        'group_id': 'multiple-questions-group',
+        'group_id': group_id,
         'group_instance': 0,
     }
 
@@ -234,15 +234,15 @@ def post_household_composition(eq_id, form_type, collection_id):
 
     if 'action[add_answer]' in request.form:
         questionnaire_manager.add_answer(this_block, 'household-composition-question', answer_store)
-        return get_block(eq_id, form_type, collection_id, 'multiple-questions-group', 0, 'household-composition')
+        return get_block(eq_id, form_type, collection_id, group_id, 0, 'household-composition')
 
     elif 'action[remove_answer]' in request.form:
         index_to_remove = request.form.get('action[remove_answer]')
         questionnaire_manager.remove_answer(this_block, answer_store, index_to_remove)
-        return get_block(eq_id, form_type, collection_id, 'multiple-questions-group', 0, 'household-composition')
+        return get_block(eq_id, form_type, collection_id, group_id, 0, 'household-composition')
 
     if not valid:
-        return _render_template(questionnaire_manager.state, 'multiple-questions-group', 0, 'household-composition', template='questionnaire')
+        return _render_template(questionnaire_manager.state, group_id, 0, 'household-composition', template='questionnaire')
 
     return _go_to_next_block(current_block_id=this_block, eq_id=eq_id,
                              form_type=form_type, collection_id=collection_id)
--- a/app/parser/v0_0_1/schema_parser.py
+++ b/app/parser/v0_0_1/schema_parser.py
@@ -124,6 +124,8 @@ class SchemaParser(AbstractSchemaParser):
             # re-use the parse validation method
             self._parse_validation(questionnaire, self._schema)
 
+        questionnaire.register_aliases()
+
         return questionnaire
 
     @staticmethod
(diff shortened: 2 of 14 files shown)


app/main/views/questionnaire.py
```diff
@@ -369,5 +369,5 @@ def _render_template(context, group_id=None, group_instance=0, block_id=None, re
 
 
 def _render_schema(schema_json, answers, metadata):
-    schema_context = build_schema_context(metadata, g.schema.aliases, answers)
+    schema_context = build_schema_context(metadata, g.schema.aliases, answers, request.view_args)
```
benjefferies: Shall we limit this to just the group_instance while we are only using that?

app/questionnaire/questionnaire_manager.py
```diff
@@ -91,7 +91,7 @@ def build_state(self, item_id, answers):
             self.state.update_state(answers)
             self._conditional_display(self.state)
 
-            context = build_schema_context(metadata, self._schema.aliases, all_answers)
+            context = build_schema_context(metadata, self._schema.aliases, all_answers, request.view_args)
```
jonnyshaw89: Should we be passing these from the view as opposed to grabbing them from the request lower down?

jonnyshaw89 (changes requested):

app/templating/schema_context.py
```diff
@@ -1,25 +1,43 @@
+from app.questionnaire_state.state_repeating_answer_question import iterate_over_instance_ids
 from app.utilities.date_utils import to_date
 
 
-def build_schema_context(metadata, aliases, answers):
+def build_schema_context(metadata, aliases, answers, group_instance=None):
```
benjefferies: Should the default group_instance be 0?

app/questionnaire/questionnaire_manager.py
```diff
@@ -79,7 +79,7 @@ def process_incoming_answers(self, location, post_data):
 
         return is_valid
 
-    def build_state(self, item_id, answers):
+    def build_state(self, item_id, answers, group_instance=None):
```
benjefferies: Should the default group_instance be 0?

app/main/views/questionnaire.py
```diff
@@ -356,3 +356,8 @@ def _render_template(context, group_id=None, group_instance=0, block_id=None, te
                                  content=context,
                                  previous_location=previous_url,
                                  schema=g.schema_json)
+
+
+def _render_schema(schema_json, answers, metadata):
```
benjefferies: This isn't used anywhere

app/templating/schema_context.py
```diff
@@ -1,25 +1,43 @@
...
 
 
 def _map_alias_to_answers(aliases, answers):
     values = {}
     for alias, item_id in aliases.items():
-        value = answers.get(item_id)
-        values[alias] = value if value else ""
+        matching_answers = []
+        for answer_id, answer_index in iterate_over_instance_ids(answers):
+            if answer_id == item_id:
```
benjefferies: Maybe we should pass the answer_store instead of the flattened answers. Then we could just filter for answers with the same item_id otherwise we are implementing filtering logic on top of the flattened answers

app/templating/schema_context.py
```diff
@@ -1,25 +1,43 @@
...
+                key = '_'.join([answer_id, str(answer_index)]) if answer_index > 0 else answer_id
+                matching_answers.append(answers.get(key))
+
+        number_of_matches = len(matching_answers)
+
+        repeating_answer_alias = alias.startswith('_')
```
benjefferies: I don't think there is any need to a add a magic underscore to identify repeating answers is there?

tests/functional/pages/surveys/repeating_household/repeating-household.page.js
```diff
@@ -0,0 +1,12 @@
+import QuestionPage from '../question.page'
+
+class RepeatingHouseholdPage extends QuestionPage {
+
+  isNameDisplayed(name) {
+    var title = browser.getText('h1[class="section__title saturn"]')
```
benjefferies: Might be worth adding an identifiable name or id for titles which takes the id of the schema item

tests/functional/spec/repeating-household.spec.js
```diff
@@ -0,0 +1,64 @@
...
+
+  it('Given I enter one name, when I navigate through the subsequent group, I should see the name on each block.', function() {
+    // Given
+    startQuestionnaire(repeating_household_schema)
+    var name = 'Person One'
+
+    // When
+    HouseholdCompositionPage.setPersonName(0, name).submit()
+
+    expect(RepeatingHouseholdPage.isNameDisplayed(name)).to.be.true
```
benjefferies: Will chai it is better do use to.equal(name) because if/when it fails you get the comparison in the logs as debug otherwise you will just get expected true was false

benjefferies (changes requested):
Good effort on the tests they're really readable 👍

app/main/views/questionnaire.py
samiwel: I think this must have been left over from a recent merge, I've removed it.

app/questionnaire/questionnaire_manager.py
samiwel: Changed.

app/templating/schema_context.py
samiwel: Changed.

tests/functional/spec/repeating-household.spec.js
samiwel: Good point. I've changed this now so that it uses chai's `equal` instead.

app/schema/questionnaire.py
```diff
@@ -43,16 +43,22 @@ def register(self, item):
             raise QuestionnaireException('{} is a duplicate id'.format(item.id))
 
         self.items_by_id[item.id] = item
-        self._register_alias(item)
 
         # Build the two-way relationship
         item.questionnaire = self
 
+    def register_aliases(self):
+        for _, item in self.items_by_id.items():
```
iwootten: Minor thing, but if the key is unused, then we don't need to call .items() to return key, value pairs.

app/schema/questionnaire.py
```diff
@@ -43,16 +43,22 @@ def register(self, item):
...
+            self._register_alias(item)
+
     def _register_alias(self, item):
         # Register the alias if the item has one
         if isinstance(item, Answer) and item.alias is not None:
             if item.alias not in self.aliases.keys():
-                self.aliases[item.alias] = item.id
+                self.aliases[item.alias] = {
+                    "item_id": item.id,
```
iwootten: Is this actually an answer_id? - it's later used to filter answer store against answer_ids. Are there any cases where item_id points to something other than an answer?

iwootten (changes requested):

app/schema/questionnaire.py
samiwel: Yeah it's an answer_id. As far as I know, we only use aliases on Answers right now, so I'll rename this to answer_id for clarity.

iwootten (approved):

app/parser/v0_0_1/schema_parser.py
```diff
@@ -306,6 +308,7 @@ def _parse_answer(self, schema, questionnaire):
             answer.mandatory = ParserUtils.get_required_boolean(schema, 'mandatory')
             answer.options = ParserUtils.get_optional_array(schema, 'options')
             answer.alias = ParserUtils.get_optional_string(schema, 'alias')
+            answer.repeats = ParserUtils.get_optional(schema, "repeats")
```
benjefferies: I guess this is not needed now

benjefferies (approved):
Just @jonnyshaw89 left on this one now

jonnyshaw89:
Were we going to update the census schema in this PR?

benjefferies:
Good point we should do that @samiwel

collisdigital:
Can we also extend Census schema selenium tests as we go?

jonnyshaw89 (approved):

jonnyshaw89:
Do you want to squash these commits?

PR #590, opened 2016-11-27
Adds support for question guidance

collisdigital:
Trello card: https://trello.com/c/Gaqrltzr

Adding support for question guidance within questions as per the latest tested design prototypes
- See http://onsdigital.github.io/eq-prototypes/ukis/v3/section-11/ for example
- Or http://onsdigital.github.io/eq-prototypes/census-lite/v7/section-4/

screen shot 2016-11-27 at 22 47 14

screen shot 2016-11-27 at 22 46 58

Guidance is attached to questions and can contain optional titles, descriptions and bullet lists of items to show. Multiple guidance elements are allowed per question (they stack below each other).

Here's an example of the JSON to add gudiance to a question object:

```Javascript
"guidance": [
    {
        "title": "Include",
        "description": "<p>Guidance <b>include</b> description text</p>",
        "list": [
            "Item Include 1",
            "Item Include 2",
            "Item Include 3",
            "Item Include 4"
        ]
    },
    {
        "title": "Exclude",
        "description": "<p>Guidance <b>exclude</b> description text</p>",
        "list": [
            "Item Exclude 1",
            "Item Exclude 2",
            "Item Exclude 3",
            "Item Exclude 4"
        ]
    }
```

The general approach (in the schema) was agreed with @ajmaddaford and @hamishtaplin.

### Help needed
This PR needs help on the selenium test, see the strawman approach, is this the right way to go? Should we be checking raw HTML in the test like this?

The frontend changes and SCSS need a FEer to review; noting that the existing alert style used a 90% hint to handle the background, which didn't match the actual approach to the design, so I've added a default to use 90% and the ability to override the alert background if needed (and used that here).

The Census theme hasn't been updated to include the guidance style; it's not clear how best to do this looking at the existing theme (maybe wait for Census Header, Footer & Landing Page Design (8) card?), discuss with @hamishtaplin.

Finally the schemas in #580 will need updating (after that merges) to break out the include/exclude lists etc.

### How to review 
- Run the `test_question_guidance.json` schema to see various examples of question guidance.
- Run each of the existing business surveys schemas (0102, 0112, 0215 etc.) and verify the guidance have been applied correctly.
- Review the schema approach taken and names of elements used, are they acceptable?
- Review the frontend changes to the question.html
- Review the SCSS changes

### DIFF
--- a/app/data/1_0102.json
+++ b/app/data/1_0102.json
@@ -120,7 +120,26 @@
                                             }
                                         }
                                     ],
-                                    "description": "Round your figures to the nearest whole pound (£). Even if your figures are zero, please still complete.<div><br/> <h4>Include</h4> <ul> <li>VAT</li> <li>internet sales</li></ul> </div> <div> <h4>Exclude</h4> <ul> <li>revenue from mobile phone network commission and top up </li> <li>sales from catering facilities used by customers</li>  <li>lottery sales and commission from lottery sales</li> <li>sales of car accessories and motor vehicles</li> <li>NHS receipts</li> </ul> </div>",
+                                    "description": "<p>Round your figures to the nearest whole pound (£). Even if your figures are zero, please still complete.</p>",
+                                    "guidance": [
+                                        {
+                                            "title": "Include",
+                                            "list": [
+                                                "VAT",
+                                                "internet sales"
+                                            ]
+                                        },
+                                        {
+                                            "title": "Exclude",
+                                            "list": [
+                                                "revenue from mobile phone network commission and top up ",
+                                                "sales from catering facilities used by customers",
+                                                "lottery sales and commission from lottery sales",
+                                                "sales of car accessories and motor vehicles",
+                                                "NHS receipts"
+                                            ]
+                                        }
+                                    ],
                                     "display": {
                                         "properties": {}
                                     },
@@ -171,7 +190,13 @@
                                             }
                                         }
                                     ],
-                                    "description": "Round your figures to the nearest whole pound (£). Even if your figures are zero, please still complete.<div><br/> <h4>Include</h4><ul> <li>VAT</li></ul></div>",
+                                    "description": "<p>Round your figures to the nearest whole pound (£). Even if your figures are zero, please still complete.</p>",
+                                    "guidance": [
+                                        {
+                                            "title": "Include",
+                                            "list": [ "VAT" ]
+                                        }
+                                    ],
                                     "display": {
                                         "properties": {}
                                     },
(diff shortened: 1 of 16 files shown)


app/templates/partials/question.html
```diff
@@ -12,17 +13,41 @@ <h2 class="question__title neptune">{{question_title}}</h2>
...
+          {%- if highlight.title -%}
+            <h3 class="venus">{{highlight.title}}</h3>
+          {% endif %}
+          {%- if highlight.description -%}
+            <div class="mars">{{highlight.description|safe}}</div>
+          {% endif %}
+          {%- if highlight.list -%}
+            <ul>
+              {%- for item in highlight.list -%}
+                <li class="mars">{{item|safe}}</li>
```
hamishtaplin: We don't need to explicitly use the `mars` class as it's the default body text style (ignore how the prototypes approach this)

app/assets/styles/partials/vars/_colors.scss
```diff
@@ -18,6 +18,8 @@ $color-red: #c50b0b;
 $color-yellow: yellow;
 $color-amber: #fe781f;
 $color-blue: #4263c2;
+$color-highlight: #003d59;
```
hamishtaplin: I'm not sure we need to store this colour as a variable as it won't be used again elsewhere. Need to check why it's not just the brand shade of blue as it's very close. 

In fact, there's already a style for an alert using the brand colour of the theme here: https://onsdigital.github.io/eq-survey-runner-patterns/components/detail/alert--default.html

app/data/1_0102.json
```diff
@@ -171,7 +190,13 @@
                                             }
                                         }
                                     ],
-                                    "description": "Round your figures to the nearest whole pound (£). Even if your figures are zero, please still complete.<div><br/> <h4>Include</h4><ul> <li>VAT</li></ul></div>",
+                                    "description": "Round your figures to the nearest whole pound (£). Even if your figures are zero, please still complete.",
```
hamishtaplin: I realise this isn't a change we introduced here but we should be consistent with how we apply the paragraphs in the description as this doesn't have a wrapping `<p>` tag but others do.

hamishtaplin (changes requested):

hamishtaplin (approved):

jonnyshaw89 (changes requested):
Should be be updating the current Census schemas with these changes?

collisdigital:
Will make the changes to the household schema as part of this PR

jonnyshaw89 (approved):

jonnyshaw89:
@collisdigital Are the labels on this PR still relevant?

collisdigital:
@jonnyshaw89 Yes, whilst @hamishtaplin  and @bennpearson  review the spacing and alignment of the boxes

yoganandkunche:
QA for this PR is done on the branch.
Cross browser testing is done: Window (IE,Chrome,Firefox) Mac (Safari, Firefox, Chrome) Android (Samsung S7) iOS (Iphone 7s) Ipad.
One minor issue identified for Survey 0102 & 0112. Please find attached screenshot.
screen shot 2016-11-29 at 11 46 06

collisdigital:
@yoganandkunche fixed 0112 and 0102 schemas.

Agreed with @hamishtaplin and @bennpearson that margins and whitespace issues will be resolved as part of https://trello.com/c/PBOnBqqL (vertical rhythm).

PR #591, opened 2016-11-28
Eq 737 census sexual id question

jonnyshaw89:
Added ability to evaluate metadata in routing rules and skip conditions to allow for sexual identity question to be conditionally asked and for welsh questions to be shown in place of alternative.

### How to review 
Launch the census household survey with and without the sexual identity flag and observe the sexual identity question being conditionally asked.
Launch the census household survey with welsh region code and see that welsh questions are displayed in place of alternate questions.


app/data/census_household.json
```diff
@@ -78,7 +78,7 @@
                     "title": "Who lives here?",
                     "sections": [
                         {
-                            "id": "else-permanent-or-family-home-section",
+                            "id": "else-permanent-or-family-home_section",
```
benjefferies: Looks like you've merged the _ back in

app/data/census_household.json
```diff
@@ -192,24 +192,24 @@
...
                         },
                         {
                             "goto": {
-                                "id": "who-lives-at-this-address",
+                                "id": "household-composition",
                                 "when": {
-                                    "id" : "everyone-at-address-confirmation-answer",
+                                    "id": "everyone-at-address-confirmation-answer",
                                     "condition": "equals",
-                                    "value":"No - I need to add another person"
```
benjefferies, line 256: Looks like `No - I need to add another person` has been merged out

app/data/census_household.json
```diff
@@ -192,24 +192,24 @@
...
+                                    "value": "Not"
```
ajmaddaford: Should this be 'No'?

app/data/census_individual.json
```diff
@@ -863,60 +647,127 @@
...
+                                "id": "carer",
+                                "when": {
+                                    "id": "country-of-birth-england-answer",
+                                    "condition": "equals",
+                                    "value": "Northern Ireland"
+                                }
+                            }
+                        },
+                        {
+                            "goto": {
```
benjefferies, line 701: The gotos are duplicated?

app/questionnaire/navigator.py
```diff
@@ -170,10 +200,10 @@ def get_location_path(self):
...
         # Make sure we don't update original
         location_path = [{
-            "block_id": block_id,
-            "group_id": self.first_group_id,
-            "group_instance": 0,
-        } for block_id in Navigator.PRECEEDING_INTERSTITIAL_PATH]
+                             "block_id": block_id,
+                             "group_id": self.first_group_id,
+                             "group_instance": 0,
+                         } for block_id in Navigator.PRECEEDING_INTERSTITIAL_PATH]
```
ajmaddaford: Not sure what's happened to the spacing here?

app/questionnaire/navigator.py
```diff
@@ -33,14 +33,41 @@ def evaluate_goto(goto_rule, answers, group_instance):
...
+            if _contains_in_dict(metadata, meta_index):
+                metadata_value = _get_from_dict(metadata, meta_index)
+                return evaluate_rule(goto_rule, metadata_value)
+
         return False
     return True
 
 
+def _get_from_dict(d, keys):
+    if "." in keys:
```
benjefferies, line 59: Why have we got the dot separated variables i.e. variant.sexual_identity? Can it just be sexual_identity or has this already been agreed with upstream systems?

tests/app/questionnaire/test_questionnaire_manager.py
```diff
@@ -145,3 +149,47 @@ def setUp(self):
     def tearDown(self):
         # Reset some behaviours.
         QuestionnaireManager.update_questionnaire_store = self.original_update_questionnaire_store
+
+    def test_conditional_display_skips_when_equals(self):
```
benjefferies, line 153: Good effort for backfilling the tests!

tests/functional/pages/surveys/census/household/household-composition.page.js
```diff
@@ -0,0 +1,38 @@
+import QuestionPage from '../../question.page'
+
+class HouseholdCompositionPage extends QuestionPage {
```
benjefferies, line 3: There should already be a page model for this

tests/functional/spec/census-household.spec.js
```diff
@@ -139,5 +216,67 @@ describe('Example Test', function () {
...
+        Over16.clickOver16AnswerYes().submit()
+        MaritalStatus.clickMaritalStatusAnswerMarried().submit()
+        AnotherAddress.clickAnotherAddressAnswerNo().submit()
+        InEducation.clickInEducationAnswerNo().submit()
+        CountryOfBirth.clickCountryOfBirthWalesAnswerWales().submit()
+        Carer.clickCarerAnswerNo().submit()
+        NationalIdentity.clickNationalIdentityAnswerBritish().submit()
+        EthnicGroup.clickEthnicGroupAnswerWhite().submit()
+        WhiteEthnicGroup.clickWhiteEthnicGroupAnswerEnglishWelshScottishNorthernIrishBritish().submit()
+        // SexualIdentity.clickSexualIdentityAnswerHeterosexualOrStraight().submit()
```
benjefferies: Can be deleted?

tests/functional/spec/census-household.spec.js
```diff
@@ -76,14 +78,89 @@ import ThankYou from '../pages/thank-you.page'
 
 const expect = chai.expect
 
-describe('Example Test', function () {
+describe('Census Household', function () {
 
-    it('Given a census household survey, When i enter valid data, Then the survey should submit successfully', function () {
-        startQuestionnaire('census_household.json')
+    const openAndStartQuestionnaire = (schema, sexual_identity = false, region = 'GB-ENG') => {
```
benjefferies: We should pull this out into the helper class and parameterise that

tests/functional/spec/census-individual.spec.js
```diff
@@ -0,0 +1,149 @@
...
+        Over16.clickOver16AnswerYes().submit()
+        MaritalStatus.clickMaritalStatusAnswerMarried().submit()
+        AnotherAddress.clickAnotherAddressAnswerNo().submit()
+        InEducation.clickInEducationAnswerNo().submit()
+        CountryOfBirth.clickCountryOfBirthEnglandAnswerEngland().submit()
+        Carer.clickCarerAnswerNo().submit()
+        NationalIdentity.clickNationalIdentityAnswerBritish().submit()
+        EthnicGroup.clickEthnicGroupAnswerWhite().submit()
+        WhiteEthnicGroup.clickWhiteEthnicGroupAnswerEnglishWelshScottishNorthernIrishBritish().submit()
+        // SexualIdentity.clickSexualIdentityAnswerHeterosexualOrStraight().submit()
```
benjefferies: Can be deleted?

tests/functional/spec/census-individual.spec.js
```diff
@@ -0,0 +1,149 @@
...
+        Religion.clickReligionAnswerNoReligion().submit()
+        PastUsualAddress.clickPastUsualAddressAnswerThisAddress().submit()
+        Passports.clickPassportsAnswerUnitedKingdom().submit()
+        Disability.clickDisabilityAnswerNo().submit()
+        Qualifications.clickQualificationsAnswerUndergraduateDegree().submit()
+        Volunteering.clickVolunteeringAnswerNo().submit()
+        EmploymentType.clickEmploymentTypeAnswerWorkingAsAnEmployee().submit()
+        HouseholdMemberCompleted.submit()
+
+        browser.debug
```
benjefferies: Debug not needed

benjefferies (changes requested):
Good effort on backfilling tests and putting well named ids into the schemas

app/data/census_individual.json
jonnyshaw89: no, the ids are different, some for england and some for wales

app/questionnaire/navigator.py
jonnyshaw89: already defined in here https://github.com/ONSdigital/ons-schema-definitions

app/data/census_individual.json
benjefferies: Think this was on a different line before. It was on one about england wales scotland or northern ireland but it had them twice

app/questionnaire/navigator.py
```diff
@@ -33,14 +33,45 @@ def evaluate_goto(goto_rule, answers, group_instance):
...
 
+def evaluate_metadata_rule(goto_rule, metadata):
+    meta_index = goto_rule['when']['meta']
+    if _contains_in_dict(metadata, meta_index):
+        metadata_value = _get_from_dict(metadata, meta_index)
+        return evaluate_rule(goto_rule, metadata_value)
+
+
+def _get_from_dict(d, keys):
+    if "." in keys:
```
benjefferies, line 59: Do we need the recursion then could we not do keys.split(".")[-1] to get the last part of the key?

benjefferies: It just feels like we are trying to support cases on the metadata key which we don't know if they exist with this recursive call

jonnyshaw89: for the region it is defined at the root of the json but the variant_flags are nested so we need both scenarios

app/data/census_individual.json
jonnyshaw89: yeh they need to be defined for each question

app/questionnaire/navigator.py
benjefferies: Fair enough point retracted

collisdigital:
tests/functional/pages/surveys/census/invividual/arrive-in-uk.page.js

**invividual** << ^^

Got a lisp?

app/dev_mode/templates/dev-page.html
```diff
@@ -70,6 +70,11 @@
             <option name="GB-WLS" value="GB-WLS">GB-WLS</option>
           </select>
         </p>
+        <p>
+            Sexual Identity Flag:
+            <input type="checkbox" name="sexual_identity" class="qa-sexual-identity" value="true">
+        </p>
+
```
ajmaddaford, line 77: This is probably good enough for now to get this in, but we should create a card to re-visit this to amend the dev page to allow any custom variant flags

tests/app/confirmation_page/test_confirmation_page.py
```diff
@@ -15,7 +15,7 @@ def test_get_next_location_confirmation(self):
         answer_store = AnswerStore()
         answer_store.add(answer)
 
-        navigator = Navigator(load_schema_file("0_rogue_one.json"), answer_store)
+        navigator = Navigator(load_schema_file("0_rogue_one.json"), {}, answer_store)
```
iwootten, line 18: Would it make sense to defaulting metadata to an empty dict in Navigator and pass answer_store in as a kwarg to prevent passing nothing?

tests/functional/helpers.js
```diff
@@ -6,6 +6,20 @@ export const getUri = uri => browser.options.baseUrl + uri
 
 export const getRandomString = length => sampleSize('ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789', length).join('')
 
+export const openAndStartCensusQuestionnaire = (schema, sexual_identity = false, region = 'GB-ENG') => {
```
ajmaddaford: I don't think this should be in the helpers used by all functional tests

tests/app/confirmation_page/test_confirmation_page.py
jonnyshaw89: I did think about defaulting the metadata, but would there ever be a case when we need to use the default other than in unit testing? as a user couldn't get into the system without metadata?

benjefferies:
Thanks @jonnyshaw89, now that's sorted I have clojure...

app/questionnaire/navigator.py
```diff
@@ -33,14 +33,45 @@ def evaluate_goto(goto_rule, answers, group_instance):
...
-        if len(filtered) == 1:
-            return evaluate_rule(goto_rule, filtered[0]['value'])
+
+        if 'id' in goto_rule['when']:
+            answer_index = goto_rule['when']['id']
+            filtered = answers.filter(answer_id=answer_index, group_instance=group_instance)
+            if len(filtered) == 1:
+                return evaluate_rule(goto_rule, filtered[0]['value'])
+
+        if 'meta' in goto_rule['when']:
```
iwootten: Shouldn't this be an elif? Is there ever a case that both would be specified and we'd want to fallback to an if with meta?

tests/functional/pages/surveys/census/household/household-composition.page.js
jonnyshaw89: There is, it is slightly different. I thought we were going down the lines of a page model per page in survey.

benjefferies: Not sure what has been agreed. I have reused them where possible. I guess generally the benefit of the page models are that they are reusable bits of logic

tests/app/confirmation_page/test_confirmation_page.py
benjefferies: I'd agree with @jonnyshaw89 here. Looks like metadata is a fundamental component in the navigator in which it wouldn't function properly without so it should be required?

app/questionnaire/navigator.py
```diff
@@ -63,8 +94,9 @@ class Navigator:
     PRECEEDING_INTERSTITIAL_PATH = ['introduction']
     CLOSING_INTERSTITIAL_PATH = ['summary', 'thank-you']
 
-    def __init__(self, survey_json, answer_store=None):
+    def __init__(self, survey_json, metadata, answer_store=None):
```
iwootten: Given answer_store defaults to empty, I think metadata should do the same

iwootten (changes requested):

benjefferies (approved):

tests/app/confirmation_page/test_confirmation_page.py
iwootten: Nope, I disagree :) Navigator should work with both an empty AnswerStore and metadata. It's valid for me to request a path without metadata, which would be informed only by what's in the json. Similarly, it's valid for me to request a path based on answer store and the survey only...

benjefferies: but nowhere do we use the navigator without an AnswerStore so why support it? It just means more logic to support/test

app/questionnaire/navigator.py
```diff
@@ -33,14 +33,45 @@ def evaluate_goto(goto_rule, answers, group_instance):
...
-            return evaluate_rule(goto_rule, filtered[0]['value'])
+
+        if 'id' in goto_rule['when']:
+            answer_index = goto_rule['when']['id']
+            filtered = answers.filter(answer_id=answer_index, group_instance=group_instance)
+            if len(filtered) == 1:
+                return evaluate_rule(goto_rule, filtered[0]['value'])
+
+        if 'meta' in goto_rule['when']:
+            return evaluate_metadata_rule(goto_rule, metadata)
```
ajmaddaford: I think it would be more easily understood if this matched what we do for the id rule above - have a method to get the value from the metadata, and then use the `evaluate_rule`, something like:
```
key = goto_rule['when']['meta']
return evaluate_rule(goto_rule, value)
```

app/questionnaire/navigator.py
```diff
@@ -24,23 +24,55 @@ def evaluate_rule(rule, answer_value):
...
-            return evaluate_rule(goto_rule, filtered[0]['value'])
+
+        if 'id' in goto_rule['when']:
+            answer_index = goto_rule['when']['id']
+            filtered = answers.filter(answer_id=answer_index, group_instance=group_instance)
+            if len(filtered) == 1:
+                return evaluate_rule(goto_rule, filtered[0]['value'])
+
+        elif 'meta' in goto_rule['when']:
+            return evaluate_metadata_rule(goto_rule, metadata)
```
ajmaddaford: I think it would be more easily understood if this matched what we do for the id rule above - a method to get the value from the metadata, and then use the `evaluate_rule`, something like:
```
key = goto_rule['when']['meta']
return evaluate_rule(goto_rule, value)
```

app/questionnaire/questionnaire_manager.py
```diff
@@ -110,9 +110,13 @@ def _conditional_display(self, item):
...
                 rule = item.schema_item.skip_condition.as_dict()
-                answer = get_answers(current_user).get(rule['when']['id'])
 
-                item.skipped = evaluate_rule(rule, answer)
+                if 'id' in rule['when'] and rule['when']['id']:
+                    answer = get_answers(current_user).get(rule['when']['id'])
+                    item.skipped = evaluate_rule(rule, answer)
+
+                if 'meta' in rule['when'] and rule['when']['meta']:
+                    item.skipped = evaluate_metadata_rule(rule, get_metadata(current_user))
```
ajmaddaford: This logic looks the same as used in the `evaluate_goto` method in Navigator, any way we could rationalise this?

ajmaddaford (changes requested):

app/dev_mode/templates/dev-page.html
jonnyshaw89: Do we have confirmation on the name of the flag that will be passed to us?

tests/app/questionnaire/test_navigator.py
```diff
@@ -822,3 +822,36 @@ def test_next_with_conditional_path_based_on_metadata(self):
                                                                     current_block_id=current_block_id,
                                                                     current_iteration=current_iteration))
 
+    def test_next_with_conditional_path_based_on_metadata_when_value_not_in_metadata(self):
```
iwootten: Hmm, I think the title of this test is a little wordy no?

app/dev_mode/templates/dev-page.html
ajmaddaford: For this yes, but the variant flags (if used) will probably be different for every survey.

tests/app/confirmation_page/test_confirmation_page.py
ajmaddaford: I think that while there is an element of functionality that exists in Navigator that works without both metadata and and AnswerStore we should support them being optional. I'm not convinced there's more logic to test as when optional they default to `{}` and an `AnswerStore` instance, if they were mandatory should still test with `{}`and an empty `AnswerStore`  instance.

ajmaddaford (approved):

iwootten (approved):

PR #593, opened 2016-11-28
Schema Helper

iwootten:
Broke the schema interaction logic out from the navigator to assist in other classes interacting with schemas.

This was originally part of my spike for repeating, so I've brought it back in as I felt it would be useful to put common logic here rather than adding it all to the navigator.

### How to review 
Unit tests have been added, check whether they're appropriate.

### DIFF
--- a/app/helpers/schema_helper.py
+++ b/app/helpers/schema_helper.py
@@ -1,0 +1,39 @@
+class SchemaHelper(object):
+
+    @staticmethod
+    def get_first_group_id(survey_json):
+        return survey_json['groups'][0]['id']
+
+    @staticmethod
+    def get_last_group_id(survey_json):
+        return survey_json['groups'][-1]['id']
+
+    @staticmethod
+    def get_first_block_id(survey_json):
+        return survey_json['groups'][0]['blocks'][0]['id']
+
+    @staticmethod
+    def get_blocks(survey_json):
+        for group in survey_json['groups']:
+            for block in group['blocks']:
+                yield block
+
+    @staticmethod
+    def get_groups(survey_json):
+        for group in survey_json['groups']:
+            yield group
+
+    @staticmethod
+    def get_repeat_rules(group):
+        if 'routing_rules' in group:
+            for rule in group['routing_rules']:
+                if 'repeat' in rule.keys():
+                    yield rule
+
+    @classmethod
+    def get_group(cls, survey_json, group_id):
+        return next(g for g in cls.get_groups(survey_json) if g["id"] == group_id)
+
+    @classmethod
+    def get_block(cls, survey_json, block_id):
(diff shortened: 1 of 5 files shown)


app/helpers/schema_helper.py
```diff
@@ -0,0 +1,46 @@
...
+    @staticmethod
+    def get_blocks(survey_json):
+        blocks = []
+        for group in survey_json['groups']:
+            for block in group['blocks']:
+                blocks.append(block)
+        return blocks
+
+    @staticmethod
+    def get_groups(survey_json):
```
collisdigital, line 22: Worth considering if a generator would be better for these kinds of iterator/return a list methods, to avoid building up the whole list up-front and then iterating over it vs. iterating all elements one at a time when needed, which also allows you to quit the iterator before compiling the whole list if you find what you need earlier e.g.

```python
def get_groups(survey_json):
    for group in survey_json['groups']:
        yield group
```
Usage
---
```python
for group in SchemaHelper.get_groups(g.schema_json):
    ...

for index, group in enumerate(SchemaHelper.get_groups(g.schema_json)):
    ...
```
Ref
---
- http://anandology.com/python-practice-book/iterators.html

app/helpers/schema_helper.py
```diff
@@ -0,0 +1,46 @@
...
+        groups = []
+        for group in survey_json['groups']:
+            groups.append(group)
+        return groups
+
+    @staticmethod
+    def get_repeat_rules(survey_json, group_index):
```
benjefferies: Would it make sense to pass in the group here?

tests/app/questionnaire/test_navigator.py
```diff
@@ -400,7 +383,6 @@ def test_next_location_empty_routing_rules(self):
         navigator = Navigator(survey, answers)
 
         # Force some empty routing rules
```
benjefferies: Comments only go stale if you ignore them 👎

benjefferies (changes requested):

tests/app/questionnaire/test_navigator.py
iwootten: Ha!

iwootten: This was a test.

app/helpers/schema_helper.py
iwootten: Annoyingly, I originally implemented it this way and backtracked as I was having trouble getting it working with Navigator. Great minds eh? I've managed to get it working now though.

jonnyshaw89 (approved):

benjefferies (approved):

benjefferies:
@iwootten can you squash please

PR #597, opened 2016-11-30
Use unique template for interstitial pages

### LINKED ISSUE #587: On Interstitial Page the button must read as Continue rather than Save and Continue
### Expected behaviour
Interstitial Page must display the button as [Continue] according to the prototypes
### Actual behaviour
The button on the Interstitial Page reads as [Save and Continue] when we are actually not saving anything on that page
use test_Interstitial Page.json
complete the survey and get to Interstitial Page
#### Browser
All Browsers
#### Operating System
Mac
### Screenshot
screen shot 2016-11-25 at 15 12 07

iwootten:

Replace the interstitial template content with its own content, unique to the interstitial page. Fixes #587.

### How to review 

Use test_interstitial_page.json. Complete the survey and get to Interstitial page - observe the button text is different to that where content is entered.


ajmaddaford (approved):

tests/functional/spec/interstitial.spec.js
```diff
@@ -17,6 +17,7 @@ describe('Interstitial Pages', function () {
 
         // When
         expect(InterstitialPage.getMainHeading()).to.contain('Breakfast Interstitial Page')
```
benjefferies: These two assertions look like they should be in the then section?

benjefferies (changes requested):
@hamishtaplin should also run his eyes over this

tests/functional/spec/interstitial.spec.js
jonnyshaw89: i disagree given the test it that you can "continue" from that page.

benjefferies: From http://martinfowler.com/bliki/GivenWhenThen.html

- The given part describes the state of the world before you begin the behavior you're specifying in this scenario. You can think of it as the pre-conditions to the test.
- The when section is that behavior that you're specifying.
- Finally the then section describes the changes you expect due to the specified behavior.

Maybe the BDD scenario is what is making this confusing, maybe it should be...

Given I am on the interstitial page
When I continue
Then I should go to the next question

Then it would be split out like...

```
//Given
startQuestionnaire(interstitial_schema)
InterstitialSurveyPage.setBreakfastFood('Cereal').submit()

// When
InterstitialPage.submit()

// Then
expect(NextPage.isOpen()).to.be.true
```

jonnyshaw89: Yeh i guess if you change the statement then that works.

app/templates/interstitial.html
```diff
@@ -1 +1,45 @@
-{% include 'questionnaire.html' %}
+{% extends theme('layouts/_survey.html') %}
```
hamishtaplin: I think this would be better solved by using template inheritance and setting a variable for the content of the button text. This way we can avoid duplicating elements of the template across files.

tests/functional/spec/interstitial.spec.js
iwootten: Modifying it in that way removes any expectation on button content and relevance to this PR. You want it reverted?

benjefferies: We could have another one, given I am answering a questionnaire with an interstitial page, when I am on the interstitial page, then the button should say continue

hamishtaplin (changes requested):
I think this would be better solved by using template inheritance and setting a variable for the content of the button text. This way we can avoid duplicating elements of the template across files.

benjefferies (approved):

hamishtaplin (approved):

PR #598, opened 2016-11-30
Making census individual match defined content

jonnyshaw89:
Implementing content to match desired content in slides


app/data/census_individual.json
```diff
@@ -1519,6 +1446,11 @@
                                     "value": "English or Welsh"
                                 }
                             }
+                        },
+                        {
```
iwootten, line 1531: Given this routing rule just goes to the next block with no conditions, is there any reason why it should be specified?

jonnyshaw89: without specifying this it errors

iwootten: Ok, it shouldn't really and I think specifying it is confusing. Doesn't this mean it's possible that two routing rules are able to be satisfied?

jonnyshaw89: In theory yes, but they get evaluated in order so the 1st would take precedence?

app/questionnaire/navigator.py
```diff
@@ -133,8 +133,11 @@ def build_path(self, blocks, group_id, group_instance, block_id, path):
         path.append(this_block)
 
         # Return the index of the block id to be visited
-        block_id_index = next(index for (index, b) in enumerate(blocks) if b["block"]["id"] == block_id and
-                              b["group_id"] == group_id and b['group_instance'] == group_instance)
+        block_id_index = next((index for (index, b) in enumerate(blocks) if b["block"]["id"] == block_id and
+                              b["group_id"] == group_id and b['group_instance'] == group_instance), None)
+
+        if block_id_index is None:
```
iwootten: Wouldn't it be better to throw an error here and catch/log at the point the path was built? Otherwise the recursive call will still carry on infinitely, just logging.

jonnyshaw89: Is there any point catching it? i know thats not usually very good, but what are we gonna do to recover in this event?

iwootten: I don't know what we'd do to recover, but it would save a horrible recursive call as soon as we know when we can't build a path any further.

app/data/census_individual.json
```diff
@@ -125,11 +44,11 @@
                 },
                 {
                     "id": "sex",
-                    "title": "John Doe",
+                    "title": "{{ answers.person_full_name }}",
```
benjefferies: Is this correct shouldn't it be something like {{answers.person_full_name[current.group_instance]}}?

ajmaddaford:
There's no routing on the over-16 question, so you get asked the marital-status question regardless of answer

ajmaddaford:
The confirmation screen isn't styled like the rest of the survey

app/data/census_individual.json
jonnyshaw89: There is only 1 person, and no repeating group.

benjefferies: That would make sense

jonnyshaw89:
@ajmaddaford the styling appears to be raised as a separate issue https://github.com/ONSdigital/eq-survey-runner/issues/605

app/data/census_individual.json
```diff
@@ -160,11 +79,11 @@
                 },
                 {
                     "id": "date-of-birth",
-                    "title": "John Doe",
+                    "title": "{{ answers.person_full_name }}",
```
ajmaddaford: I don't think we should be setting the block title to the person's name. I don't think we use the block title anywhere, but if we did it would probably be in something like the HTML title element, and we don't want people's names in there.

ajmaddaford (changes requested):

app/data/census_individual.json
jonnyshaw89: we don't appear to use it in the html title. however it is mandatory in the schema. should this be changed?

ajmaddaford: Sorry, I meant if we used it for something like the HTML title in a future version. I think the title should be the title for the page, so for this block it would be 'Date of birth'.

ajmaddaford (approved):

iwootten (approved):

PR #608, opened 2016-12-01
Extend schema validation to validate routing rules are valid

jonnyshaw89:
Extend the schema validation test to ensure that the routing rules are semantically correct and cover all scenarios. 

Fixes 601 and Fixes 602

### How to review 
Re-test issues.


app/data/0_star_wars.json
```diff
@@ -127,6 +127,11 @@
                                 }
                             }
 
+                        },
+                        {
+                            "goto": {
+                                "id": "96682325-47ab-41e4-a56e-8315a19ffe2a"
+                            }
```
samiwel, line 134: Do we need to specify a `goto` route if it's just going to the next block? Assuming the previous routing rules don't fire, doesn't it just route to the next block anyway?

tests/app/data/test_schema_validation.py
```diff
@@ -41,15 +47,67 @@ def test_schema(self):
...
         except JSONDecodeError as e:
             return "JSON Parse Error! Could not parse [{}]. Error [{}]".format(file, e)
 
 
+    @staticmethod
+    def validate_schema_contains_valid_routing_rules(file, json_to_validate):
+        for block in SchemaHelper.get_blocks(json_to_validate):
+            if 'routing_rules' in block and len(block['routing_rules']) > 0:
+                for rule in block['routing_rules']:
+                    is_goto_rule = 'goto' in rule and 'when' in rule['goto'].keys() or 'id' in rule['goto'].keys()
```
samiwel: Do we need this check? `'when' in rule['goto'].keys() or ` Isn't it sufficient just to check for existence of `'id'`?

jonnyshaw89: agree, ill change this

app/data/0_star_wars.json
jonnyshaw89: we need to specify, the way the navigator works if routing_rules are defined then 1 of them needs to match or default

samiwel: Okay, fair enough.

tests/app/data/test_schema_validation.py
```diff
@@ -28,11 +29,16 @@ def test_schema(self):
         for file in files:
 
             json_file = open(os.path.join(settings.EQ_SCHEMA_DIRECTORY, file), encoding="utf8")
+            json_to_validate = json.load(json_file)
+
+            error = self.validate_json_against_schema(file, json_to_validate, schema)
+
+            error = self.validate_schema_contains_valid_routing_rules(file, json_to_validate)
```
collisdigital: Each time error is set is overriding the last before the list is extended finally below (will only lost last error found if there were multiple); maybe extend errors immediately? Or alternatively only call later checks of previous check had no errors?

collisdigital: Only **list** ...

jonnyshaw89: A keen eye you have.

tests/app/data/test_schema_validation.py
```diff
@@ -41,15 +44,67 @@ def test_schema(self):
...
+                            if rule['goto']['id'] == 'summary':
+                                continue
+
+                            SchemaHelper.get_block(json_to_validate, rule['goto']['id'])
+                        except StopIteration:
+                            return "Schema Error. File[{}] Routing rule routes to invalid block [{}]".format(file, rule['goto']['id'])
+
+
+    @staticmethod
+    def validate_schema_containing_routing_rules_has_default_if_nots_all_answers_routed(file, json_to_validate):
```
benjefferies: validate_schema_containing_routing_rules_has_default_if_nots_all_answers_routed -> validate_schema_containing_routing_rules_has_default_if_not_all_answers_routed

tests/app/data/test_schema_validation.py
```diff
@@ -41,15 +44,72 @@ def test_schema(self):
...
+    def validate_schema_contains_valid_routing_rules(file, json_to_validate):
+
+        errors = []
+
+        for block in SchemaHelper.get_blocks(json_to_validate):
+            if 'routing_rules' in block and len(block['routing_rules']) > 0:
+                for rule in block['routing_rules']:
+                    is_goto_rule = 'goto' in rule and 'id' in rule['goto'].keys()
+                    if is_goto_rule:
+                        try:
```
benjefferies: This try except block seems a bit nasty and hard to understand. I think we should create a has_block method in the test file instead

tests/app/data/test_schema_validation.py
```diff
@@ -41,15 +44,72 @@ def test_schema(self):
...
+
+                            SchemaHelper.get_block(json_to_validate, rule['goto']['id'])
+                        except StopIteration:
+                            errors.append("Schema Error. File[{}] Routing rule routes to invalid block [{}]".format(file, rule['goto']['id']))
+
+        return errors
+
+
+    @staticmethod
+    def validate_schema_containing_routing_rules_has_default_if_nots_all_answers_routed(file, json_to_validate):
```
benjefferies: validate_schema_containing_routing_rules_has_default_if_nots_all_answers_routed -> validate_schema_containing_routing_rules_has_default_if_not_all_answers_routed

tests/app/data/test_schema_validation.py
```diff
@@ -41,15 +44,72 @@ def test_schema(self):
...
+
+        for block in SchemaHelper.get_blocks(json_to_validate):
+
+            if 'sections' in block:
+                for section in block['sections']:
+                    if 'questions' in section:
+                        for question in section['questions']:
+                            if 'answers' in question:
+                                for answer in question['answers']:
+                                    if answer['type'] == 'Radio' or answer['type'] == 'Checkbox':
```
benjefferies: I think we should refactor this to a method called get_radio_and_checkbox_answers so it is more readable in what it's doing

tests/app/data/test_schema_validation.py
```diff
@@ -41,15 +44,73 @@ def test_schema(self):
...
+
+                                        if 'routing_rules' in block and len(block['routing_rules']) > 0:
+
+                                            options = []
```
benjefferies: Could do some nice bit of python magic here

```
options = [option['value'] for option in answer['options']]
```

tests/app/data/test_schema_validation.py
```diff
@@ -41,15 +44,73 @@ def test_schema(self):
...
+                                                if 'goto' in rule:
+
+                                                    if 'when' in rule['goto'].keys():
+                                                        when = rule['goto']['when']
+                                                        if 'id' in when and when['id'] == answer['id'] and when['value'] in options:
+                                                            options.remove(when['value'])
+                                                    else:
+                                                        options = []
+
+                                            if len(options) > 0 and len(options) != len(answer['options']):
```
benjefferies: Might be nice to refactor to boolean variable no_routing_or_all_routes for clarity

benjefferies (changes requested):

benjefferies (approved):

tests/app/data/test_schema_validation.py
```diff
@@ -28,27 +29,89 @@ def test_schema(self):
...
+
+    @staticmethod
+    def validate_schema_containing_routing_rules_has_default_if_not_all_answers_routed(file, json_to_validate):
+
+        errors = []
+
+        for block in SchemaHelper.get_blocks(json_to_validate):
+
+            if 'sections' in block:
+                for section in block['sections']:
```
ajmaddaford: Can we remove some of the nesting by doing something like this for each loop:
```
for section in (s for s in block['sections'] if 'sections' in block)
```

tests/app/data/test_schema_validation.py
```diff
@@ -28,27 +29,89 @@ def test_schema(self):
...
+                        if rule['goto']['id'] == 'summary':
+                            continue
+
+                        if not cls.has_block(json_to_validate, rule['goto']['id']):
+                            errors.append("Schema Error. File[{}] Routing rule routes to invalid block [{}]".format(file, rule['goto']['id']))
+
+        return errors
+
+    @staticmethod
+    def validate_schema_containing_routing_rules_has_default_if_not_all_answers_routed(file, json_to_validate):
```
ajmaddaford: I think `validate_routing_rules_has_default_if_not_all_answers_routed` says the same in a few less words

tests/app/data/test_schema_validation.py
```diff
@@ -28,27 +29,89 @@ def test_schema(self):
...
+    @classmethod
+    def validate_schema_contains_valid_routing_rules(cls, file, json_to_validate):
+
+        errors = []
+
+        for block in SchemaHelper.get_blocks(json_to_validate):
+            if 'routing_rules' in block and len(block['routing_rules']) > 0:
+                for rule in block['routing_rules']:
+                    is_goto_rule = 'goto' in rule and 'id' in rule['goto'].keys()
+                    if is_goto_rule:
```
ajmaddaford: Can this not just be `if 'goto' in rule and 'id' in rule['goto'].keys():`?

tests/app/data/test_schema_validation.py
```diff
@@ -28,27 +29,89 @@ def test_schema(self):
...
+                        if rule['goto']['id'] == 'summary':
+                            continue
+
+                        if not cls.has_block(json_to_validate, rule['goto']['id']):
```
ajmaddaford: I think this method would be clearer as `schema_has_block`

tests/app/data/test_schema_validation.py
```diff
@@ -28,27 +29,89 @@ def test_schema(self):
...
+    def validate_schema_contains_valid_routing_rules(cls, file, json_to_validate):
+
+        errors = []
+
+        for block in SchemaHelper.get_blocks(json_to_validate):
+            if 'routing_rules' in block and len(block['routing_rules']) > 0:
+                for rule in block['routing_rules']:
+                    is_goto_rule = 'goto' in rule and 'id' in rule['goto'].keys()
+                    if is_goto_rule:
+                        if rule['goto']['id'] == 'summary':
```
ajmaddaford: `rule['goto']['id']` is used three times, might be neater using a variable e.g. `block_id`

ajmaddaford (changes requested):

ajmaddaford (approved):

Issue #614, opened 2016-12-02
Census: Error message on Number of Visitors screen not displayed correct


yoganandkunche:
### Expected behaviour
When no answer is provided an alert "Please enter a value even if that value is 0' must be  displayed (according to the spec)
### Actual behaviour
Current Alert/error message: This field is mandatory.

### Steps to reproduce the behaviour
Launch Census survey
Navigate to Visitors screen
Click [Save and Continue] with entering a value

*** Note: There is a branch created where you can find a selenium test to recreate the issue ***

https://github.com/ONSdigital/eq-survey-runner/tree/census-num-of-visitor-alert-msg

#### Browser
All Browsers

yoganandkunche:
Re-test of this Defect Passed QA.

PR #615, opened 2016-12-02
Census num of visitor alert msg

jonnyshaw89:
Fixes https://github.com/ONSdigital/eq-survey-runner/issues/614


ajmaddaford (approved):

benjefferies (approved):

tests/functional/pages/surveys/census/household/overnight-visitors.page.js
```diff
@@ -11,6 +11,10 @@ class OvernightVisitorsPage extends QuestionPage {
     return browser.element('[name="overnight-visitors-answer"]').getValue()
   }
 
+  getErrorMsg() {
```
collisdigital: If we re-generate the Selenium pages then this will probably get lost, is it something that is generic to all pages (could be in the base page model / added to a new base page to derive all these from?)

jonnyshaw89: Agreed, Done

PR #618, opened 2016-12-02
Add question numbers to schema json

jonnyshaw89:
Have added question numbers to the json schema

### How to review 
Question numbers should not be in the question title bit should be loaded through the number value in the json schema


app/templates/partials/question.html
```diff
@@ -2,13 +2,14 @@
...
 
+{% set question_number = question.schema_item.number %}
 {% set question_title = question.schema_item.title %}
 {% set question_subtitle = question.schema_item.subtitle %}
 {% set question_guidance = "" %}
 
 {% set question_titles %}
   {%- if question_title -%}
-    <h2 class="question__title neptune">{{question_title}}</h2>
+<h2 class="question__title neptune"><span>{{question_number}}. </span>{{question_title}}</h2>
```
ajmaddaford: Question number is optional, so we shouldn't output the span if it's not in the schema.

app/templates/partials/section.html
```diff
@@ -1,6 +1,6 @@
 <div class="section" id="{{section.id}}">
   {% if section.schema_item.title %}
-    <h1 class="section__title saturn">{{section.schema_item.title}}</h1>
+    <h1 class="section__title saturn"><span>{{section.schema_item.number}}. </span>{{section.schema_item.title}}</h1>
```
ajmaddaford: Section number is optional, so we shouldn't output the span if it's not in the schema.

ajmaddaford (changes requested):

ajmaddaford (approved):

hamishtaplin (approved):
LGTM

Issue #551, opened 2016-11-08
Submitted at time missing on thank you page


ajmaddaford:
### Expected behaviour

When a survey is completed and the thank you page is displayed it should contain the text "Submitted on {{date}}"

### Actual behaviour

The text "Submitted on {{date}}" is not displayed on the summary screen

### Steps to reproduce the behaviour

- Complete any survey and submit
- View the thank you page

### Screenshot
![screen shot 2016-11-08 at 13 27 07](https://cloud.githubusercontent.com/assets/1754822/20100914/5fe576c2-a5b7-11e6-9e7f-48f8ae24426d.png)

collisdigital:
Still occurring on pre-prod

collisdigital:
Remove the supporting code that isn't needed anymore (date of submission).

PR #619, opened 2016-12-02
Eq 754 update submitter

benjefferies:
* Added a new version of the answer converter to support repeating
answering and groups.
* Added data_version property in schema to support different converters
* Add assertpy assertion library

[Trello card](https://trello.com/c/CRpuib4U/754-8-update-submitter-to-handle-latest-features)

### How to review 
Unit tests should pass.

### DIFF
--- a/app/main/views/questionnaire.py
+++ b/app/main/views/questionnaire.py
@@ -1,10 +1,11 @@
 import logging
 
 from app.authentication.session_manager import session_manager
-from app.globals import get_answer_store, get_answers, get_completed_blocks, get_metadata, get_questionnaire_store
+from app.globals import get_answer_store, get_completed_blocks, get_metadata, get_questionnaire_store
 from app.helpers.schema_helper import SchemaHelper
 from app.questionnaire.navigator import Navigator
 from app.questionnaire.questionnaire_manager import get_questionnaire_manager
+from app.submitter.converter import convert_answers
 from app.submitter.submitter import SubmitterFactory
 from app.templating.introduction_context import get_introduction_context
 from app.templating.metadata_context import build_metadata_context
@@ -215,8 +216,13 @@ def submit_answers(eq_id, form_type, collection_id):
     is_valid, invalid_location = q_manager.validate_all_answers()
 
     if is_valid:
+        metadata = get_metadata(current_user)
+        answer_store = get_answer_store(current_user)
+        navigator = Navigator(g.schema_json, metadata, answer_store)
         submitter = SubmitterFactory.get_submitter()
-        submitter.send_answers(get_metadata(current_user), g.schema, get_answers(current_user))
+        message = convert_answers(metadata, g.schema, answer_store, navigator.get_routing_path())
+        submitter.send_answers(message)
+        logger.info("Responses submitted tx_id=%s", metadata["tx_id"])
         return redirect_to_thank_you(eq_id, form_type, collection_id)
     else:
         return redirect(block_url(eq_id, form_type, collection_id,
--- a/app/submitter/converter.py
+++ b/app/submitter/converter.py
@@ -1,134 +1,157 @@
-import json
 import logging
+from collections import OrderedDict
 
 from datetime import datetime, timezone
 
 from app import settings
-from app.questionnaire_state.state_repeating_answer_question import extract_answer_instance_id
-
 
 logger = logging.getLogger(__name__)
 
 
-class SubmitterConstants(object):
+class DataVersionError(Exception):
+    def __init__(self, version):
+        self.version = version
 
-    PARADATA_KEY = "paradata"
-
-    METADATA_KEY = "metadata"
-
-    RU_REF_KEY = "ru_ref"
-
-    USER_ID_KEY = "user_id"
-
-    SUBMITTED_AT_KEY = "submitted_at"
-
-    COLLECTION_KEY = "collection"
-
-    PERIOD_KEY = "period"
-
-    EXERCISE_SID_KEY = "exercise_sid"
-
-    INSTRUMENT_KEY = "instrument_id"
-
-    SURVEY_ID_KEY = "survey_id"
-
-    ORIGIN_KEY = "origin"
-
-    VERSION_KEY = "version"
-
-    TYPE_KEY = "type"
-
-    DATA_KEY = "data"
-
-    TYPE = "uk.gov.ons.edc.eq:surveyresponse"
-
-    VERSION = "0.0.1"
-
-    ORIGIN = "uk.gov.ons.edc.eq"
-
-    TX_ID = "tx_id"
+    def __str__(self):
+        return 'Data version {} not supported'.format(self.version)
 
 
-class Converter(object):
+def convert_answers(metadata, questionnaire, answer_store, routing_path):
+    """
+    Create the JSON answer format for down stream processing
 
-    @staticmethod
-    def prepare_answers(metadata_store, questionnaire, answers):
-        """
-        Create the JSON answer format for down stream processing
+    :param metadata: metadata for the questionnaire
+    :param questionnaire: the questionnaire schema
+    :param answer_store: the users answers
+    :param routing_path: the path followed by the user when answering the questionnaire
+    :return: a JSON object in the following format:
+      {
+        "tx_id": "0f534ffc-9442-414c-b39f-a756b4adc6cb",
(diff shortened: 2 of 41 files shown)


tests/app/submitter/test_converter.py
```diff
@@ -93,24 +68,54 @@ def test_prepare_answers(self):
...
 
-            self.assertEquals(EXPECTED_RESPONSE["type"], answer_object["type"])
-            self.assertEquals(EXPECTED_RESPONSE["version"], answer_object["version"])
-            self.assertEquals(EXPECTED_RESPONSE["origin"], answer_object["origin"])
-            self.assertEquals(EXPECTED_RESPONSE["survey_id"], answer_object["survey_id"])
-            self.assertEquals(EXPECTED_RESPONSE["collection"], answer_object["collection"])
-            self.assertEquals(EXPECTED_RESPONSE["metadata"], answer_object["metadata"])
-            self.assertEquals(EXPECTED_RESPONSE["paradata"], answer_object["paradata"])
-            self.assertEquals(EXPECTED_RESPONSE["data"], answer_object["data"])
+            assert_that(answer_object).has_type("uk.gov.ons.edc.eq:surveyresponse") \
```
benjefferies: I have used a assertion libary called [assertpy](https://github.com/ActivisionGameScience/assertpy) here. It's worth taking a look at the API and get a general feeling of if it's worth using as it has some good utilities and a fluent DSL

jonnyshaw89:
Is it worth pulling this fix in? https://github.com/ONSdigital/eq-survey-runner/issues/551

benjefferies:
It's probably worth doing separately as it's out of scope and has no overlap

benjefferies:
We should really have an integration test to check that messages are successfully put on the messaging queue but this is difficult to do with our current test set up

jonnyshaw89:
I just saw you had a commit message of "Add submitted_at test" so thought you were working in that area but if not thats fine.

app/submitter/converter.py
```diff
@@ -1,134 +1,148 @@
...
+               'type': 'uk.gov.ons.edc.eq:surveyresponse',
+               'origin': 'uk.gov.ons.edc.eq',
+               'survey_id': survey_id,
+               'submitted_at': submitted_at.isoformat(),
+               'collection': _build_collection(metadata),
+               'metadata': _build_metadata(metadata),
+    }
+    if questionnaire.data_version == '0.0.2':
+        payload['data'] = convert_answers_to_census_data(answer_store, routing_path)
+        payload['version'] = '0.0.2'
```
jonnyshaw89: can we not just set this above in the payload to the questionnaire.data_version, minor change, I'm just being picky.

jonnyshaw89 (changes requested):

app/submitter/converter.py
benjefferies: Good point!

app/submitter/converter.py
```diff
@@ -1,134 +1,147 @@
...
+               'tx_id': metadata['tx_id'],
+               'type': 'uk.gov.ons.edc.eq:surveyresponse',
+               'version': questionnaire.data_version,
+               'origin': 'uk.gov.ons.edc.eq',
+               'survey_id': survey_id,
+               'submitted_at': submitted_at.isoformat(),
+               'collection': _build_collection(metadata),
+               'metadata': _build_metadata(metadata),
+    }
+    if questionnaire.data_version == '0.0.2':
```
jonnyshaw89, line 59: do we potentially want this the other way around so it defaults to the latest version?

jonnyshaw89 (changes requested):

app/submitter/converter.py
benjefferies: Shouldn't matter it's a required property and needs to be one of the enum values either 0.0.1 or 0.0.2

benjefferies: Not sure if we want to enforce those two in the code e.g. if version == '0.0.2 elif version == '0.0.1'

jonnyshaw89: yeh, maybe the else makes it implicit rather than explicit. could potentially do something like this. http://stackoverflow.com/questions/60208/replacements-for-switch-statement-in-python

benjefferies: Yeah could do something like this https://github.com/ONSdigital/eq-survey-runner/blob/master/app/questionnaire/navigator.py#L83 but the arguments change so it probably doesn't apply very well

requirements_for_test.txt
```diff
@@ -5,3 +5,4 @@ coverage==4.2b1
 pytest-cov==2.3.0
 jsonschema==2.5.1
 pylint==1.6.4
+assertpy==0.10
```
iwootten: assertpy appears to offer us little more than we would get by using basic python assertions with pytest. Additionally, the project appears to have a low bus factor and very few regular releases. I'd be wary of basing unit tests on it when we already have the option of the very well supported pytest in our requirements already. Maybe we would be better off learning the specifics of that over the unittest package. What do we get over basic assertions as described in this thread here, other than a DSL? 

https://www.reddit.com/r/Python/comments/3nebhy/assertpy_dead_simple_assertion_framework_for_unit/

app/submitter/converter.py
jonnyshaw89: hmm no. maybe we just need to make the else explicit and error if the version doesn't match what we expect?

iwootten (changes requested):

app/submitter/converter.py
```diff
@@ -1,134 +1,147 @@
...
+          "user_id": "789473423",
+          "ru_ref": "432423423423"
+        },
+        "data": {}
+      }
+    """
+    survey_id = questionnaire.survey_id
+    submitted_at = datetime.now(timezone.utc)
+    payload = {
+               'tx_id': metadata['tx_id'],
```
iwootten: The indentation here is a bit odd.

requirements_for_test.txt
benjefferies: The pros for me were the DSL made it more readable and the API has some useful functions https://github.com/ActivisionGameScience/assertpy

app/main/views/questionnaire.py
```diff
@@ -215,8 +215,9 @@ def submit_answers(eq_id, form_type, collection_id):
     is_valid, invalid_location = q_manager.validate_all_answers()
 
     if is_valid:
+        navigator = Navigator(g.schema_json, get_metadata(current_user), get_answer_store(current_user))
         submitter = SubmitterFactory.get_submitter()
-        submitter.send_answers(get_metadata(current_user), g.schema, get_answers(current_user))
+        submitter.send_answers(get_metadata(current_user), g.schema, get_answer_store(current_user), navigator.get_routing_path())
```
ajmaddaford: I think it would be cleaner to convert the data and then call send_answers:
```
payload = convert_answers(get_metadata(current_user), g.schema, get_answer_store(current_user), navigator.get_routing_path())
submitter.send_answers(payload)
```

app/submitter/converter.py
benjefferies: Fixed

app/submitter/converter.py
benjefferies: Done?

ajmaddaford (changes requested):

app/main/views/questionnaire.py

jonnyshaw89 (approved):

app/main/views/questionnaire.py
```diff
@@ -215,8 +216,12 @@ def submit_answers(eq_id, form_type, collection_id):
     is_valid, invalid_location = q_manager.validate_all_answers()
 
     if is_valid:
+        metadata = get_metadata(current_user)
+        navigator = Navigator(g.schema_json, metadata, get_answer_store(current_user))
```
iwootten: For conciseness, would it be better to assign answer_store to a var and pass to each method?

app/submitter/converter.py
```diff
@@ -1,134 +1,157 @@
...
 from datetime import datetime, timezone
 
 from app import settings
-from app.questionnaire_state.state_repeating_answer_question import extract_answer_instance_id
-
 
 logger = logging.getLogger(__name__)
 
 
-class SubmitterConstants(object):
```
iwootten, line 13: Glad this has finally disappeared!

iwootten (approved):

ajmaddaford (approved):

Issue #621, opened 2016-12-05
Census: When no response is selected for 7. Are you a schoolchild or student? gives Error 404


yoganandkunche:
### Expected behaviour
When no response is selected and [Save and continue] must navigate to Question 9. What is your country of birth?
### Actual behaviour
Giving Error 404
### Steps to reproduce the behaviour
A branch is created with a selenium test to reproduce the issue.

https://github.com/ONSdigital/eq-survey-runner/tree/are-you-a-schoochild-issue

yarn run test_functional -- --spec=census/issues/are-you-a-schoolchild   

#### Browser
All

yoganandkunche:
Re-test of this Defect Passed QA.

PR #624, opened 2016-12-05
Are you a schoolchild issue

jonnyshaw89:
Fixes https://github.com/ONSdigital/eq-survey-runner/issues/621

### DIFF
--- a/app/data/census_household.json
+++ b/app/data/census_household.json
@@ -1379,22 +1379,17 @@
                     "routing_rules": [
                         {
                             "goto": {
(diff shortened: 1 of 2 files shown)


benjefferies (approved):

benjefferies:
Should we add some logic to the validator test you created to ensure that the routing rule is referencing something that is actually and id?

jonnyshaw89:
@benjefferies like this? https://github.com/ONSdigital/eq-survey-runner/blob/master/tests/app/data/test_schema_validation.py#L58

PR #635, opened 2016-12-05
Update dev-page.html

alexmorris:
Basic tidy up of the launch page and a simple string randomiser for spawning new surveys.

### How to review 
Run the launch page - check clicking the new refresh icon creates random string, check can launch new survey. If ok then we're good to go.


iwootten:
This looks great, but...the shortened dropdowns make it difficult to read which schema is selected.

screen shot 2016-12-05 at 19 07 55

app/dev_mode/templates/dev-page.html
```diff
@@ -1,84 +1,180 @@
...
+  font-size: 1rem;
+}
+
+</style>
 
 <div class="container">
   <div class="panel u-mt-l">
     <div role="main" class="panel__body" id="main" tabindex="-1">
+
+      <h1>eQ Pre-production</h1>
```
iwootten: We'd probably want to allow this to be configurable to show current environment, rather than hardcoded to preprod.

alexmorris: good call - do we have a variable for this?

alexmorris: re widths - i didnt have the live data to see the changes - i'll extend that select out

app/dev_mode/templates/dev-page.html
```diff
@@ -1,84 +1,180 @@
...
+  width:32px;
+  height:auto;
+  display:block;
+  position:absolute;
+  top: -6px;
+  right: 4px;
+  cursor:pointer;
+}
+
+.qa-btn-submit-dev {
```
hamishtaplin: Use a separate class here as the `qa-` ones are for tests and might change.

hamishtaplin (dismissed):

Issue #634, opened 2016-12-05
Census: Navigation from Question '28 Last week were you:' when no option is selected is not correct


yoganandkunche:
### Expected behaviour
When no option is selected On question 28 '28 Last week were you:' it must navigate to question 29.
### Actual behaviour
Skipping the next 5 questions and navigating to section complete screen

### Steps to reproduce the behaviour
A branch is created with a selenium test to reproduce the issue.

https://github.com/ONSdigital/eq-survey-runner/tree/routing-last-week-were-you

yarn run test_functional -- --spec=census/issues/routing-last-week-were-you

#### Browser
All

PR #643, opened 2016-12-06
Fixing Issue: Routing last week were you

jonnyshaw89:
Fixes https://github.com/ONSdigital/eq-survey-runner/issues/634

### DIFF
--- a/app/questionnaire/navigator.py
+++ b/app/questionnaire/navigator.py
@@ -16,10 +16,18 @@ def evaluate_rule(when, answer_value):
     match_value = when['value']
     condition = when['condition']
 
+    answer_to_test = str(answer_value)
+
+    if isinstance(answer_value, list):
+        if len(answer_value) == 1:
+            answer_to_test = answer_value[0]
+        else:
+            return False
+
     # Evaluate the condition on the routing rule
-    if condition == 'equals' and match_value == str(answer_value):
+    if condition == 'equals' and match_value == answer_to_test:
         return True
-    elif condition == 'not equals' and match_value != str(answer_value):
+    elif condition == 'not equals' and match_value != answer_to_test:
         return True
     return False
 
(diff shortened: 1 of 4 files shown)


benjefferies:
@jonnyshaw89 tests are failing on this branch is that because it is still WIP?

jonnyshaw89:
@benjefferies what failing tests? 😉

app/questionnaire/navigator.py
```diff
@@ -16,10 +16,15 @@ def evaluate_rule(when, answer_value):
     match_value = when['value']
     condition = when['condition']
 
+    answer_to_test = answer_value
+
+    if isinstance(answer_value, list) and len(answer_value) == 1:
```
benjefferies: If there are more than one answer should we not test for any of them to match the goto rule?

tests/functional/spec/census/issues/routing-last-week-were-you.spec.js
```diff
@@ -0,0 +1,158 @@
...
+import Volunteering from '../../../pages/surveys/census/household/volunteering.page.js'
+import EmploymentType from '../../../pages/surveys/census/household/employment-type.page.js'
+import Jobseeker from '../../../pages/surveys/census/household/jobseeker.page.js'
+
+describe('ArriveInUk', function () {
+  it('Given I am answering question 28.	Last week were you.., When I dont not select any response, Then I am routed to 29. Were you actively looking for...?', function () {
+      openAndStartCensusQuestionnaire('census_household.json')
+
+      // who-lives-here
+      PermanentOrFamilyHome
```
benjefferies, line 46: I think this is harder to read than what is was when it was chained on one line. Views?

benjefferies (changes requested):

app/questionnaire/navigator.py
jonnyshaw89: Then you get into the realm of if multiple match in different directions what do you do. this is the MVP for getting through census.

benjefferies: Cool as long as we have had the discussion and came to the conclusion

benjefferies (approved):

tests/functional/spec/census/issues/routing-last-week-were-you.spec.js
jonnyshaw89: I agree, But given these issue tests are going to be short lived I'm not going to loose a lot of sleep over it.
However this was the agreed approach from the last discussion. It will be one to raise in the next testing review?

ajmaddaford (approved):

PR #631, opened 2016-12-05
EQ 622: Household Composition Summary

iwootten:
### What is the context of this PR?
Adds the household composition summary page, showing a list of users entered and giving the option to return to edit the list.

In order to support this I've added conditional routing to a previous location. Allows us to support the routing of users to the household composition screen in census, based on the answer to a question. This is handled by adding a route to the routing path and removing the users current answer to this question (which would otherwise prevent the location path being calculated by navigator). This also fixes #629.

### How to review 
There are unit tests in navigator and integration tests to cover new behaviour. I've had to update current composition selenium tests to work with the new block.

PR #646, opened 2016-12-06
Allow introduction block to be configurable

iwootten:
Allowed introduction to be configured within the navigator, as required for the census schemas. Additionally removed the census_theme.json and rolled existing selenium tests against it into census-individual. Also removed the intro block as mandatory from the validation schema.

### How to review 
Check tests pass. Any improvements welcomed.


app/questionnaire/navigator.py
```diff
@@ -101,6 +101,11 @@ def __init__(self, survey_json, metadata=None, answer_store=None):
         self.metadata = metadata or {}
         self.survey_json = survey_json
 
+        self.preceeding_path = []
```
jonnyshaw89, line 112: I wonder if an approach like the interstitial blocks may work? i.e., define the introduction as a block? that would mean no custom logic here? also it could allow us to have multiple introductions pages for example. just thinking possibly more flexibility and a cleaner schema?

iwootten: Mmm, yes - I've thought this in the past. An introduction is unique from a block in that it's structure is different and the content is rendered using a different template.

benjefferies: We should be able to add a template or something optionally?

iwootten: Think I'll defer to @ajmaddaford on this as he had some thoughts on why interstitials should be handled differently.

tests/app/questionnaire/test_navigator.py
```diff
@@ -34,6 +34,24 @@ def test_previous_block(self):
         navigator = Navigator(survey)
         self.assertEqual(navigator.get_previous_location(current_block_id=current_block_id), previous_block)
 
+    def test_introduction_in_non_census_location_path(self):
```
benjefferies: test name could make more sense, test_introduction_in_path_when_in_schema?

tests/functional/helpers.js
```diff
@@ -18,7 +18,6 @@ export const openAndStartCensusQuestionnaire = (schema, sexualIdentity = false,
   }
 
   devPage.submit()
-  landingPage.getStarted()
```
benjefferies, line 21: Could just rename the method to startCensusQuestionnaire now

tests/functional/spec/census-individual.spec.js
```diff
@@ -188,4 +188,22 @@ describe('Census Individual', function () {
...
     })
 
+    it('Given the census theme is selected, the help should not be visible', function() {
+        // Given
+        openAndStartCensusQuestionnaire('census_individual.json')
+        // Then
+        expect(browser.isVisible('.js-help-body')).to.be.false
+    })
+
+    it('Given the census theme is selected, and I click the "help and support" button the help should be visible', function() {
```
benjefferies: Would be good to stick to given when then for consistency when it applies (although it was copy paste)

app/questionnaire/navigator.py
ajmaddaford: I think we need to be able to distinguish which pages are part of the question flow, and which sit outside of it (introduction, summary). For this it could be clearer for them to remain implemented outside of the groups and blocks. For now, this change makes sense, and we can review the wider options in the new year.

ajmaddaford (approved):

jonnyshaw89 (approved):

tests/functional/helpers.js
iwootten: Have done so in #631, where I think you also made this comment

jonnyshaw89:
👍

PR #657, opened 2016-12-08
Load a different variant of a schema based on language flag

jonnyshaw89:
Added functionality to load a different variant of a schema based on the language flag

### How to review 
Use the language flag on the dev page to load schemas in different languages

### DIFF
--- a/app/data/test_language.json
+++ b/app/data/test_language.json
@@ -1,0 +1,31 @@
+{
+    "mime_type": "application/json/ons/eq",
+    "questionnaire_id": "0",
+    "schema_version": "0.0.1",
+    "data_version": "0.0.1",
+    "survey_id": "0",
+    "title": "English Langauage Survey",
+    "theme": "default",
+    "description": "A questionnaire to demonstrate english langauge",
+    "eq_id": "0",
+    "groups": [
+        {
+            "id": "english-group",
+            "title": "",
+            "blocks": [
+                {
+                    "id": "english-block",
+                    "sections": [
+                        {
+                            "title": "English language schema",
+                            "id": "english-section",
+                            "questions": [
+                            ]
+                        }
+                    ],
+                    "title": ""
+                }
+            ]
+        }
+    ]
+}
--- a/app/data/test_language_cy.json
+++ b/app/data/test_language_cy.json
@@ -1,0 +1,31 @@
+{
+    "mime_type": "application/json/ons/eq",
+    "questionnaire_id": "0",
+    "schema_version": "0.0.1",
+    "data_version": "0.0.1",
+    "survey_id": "0",
+    "title": "Welsh Langauage Survey",
+    "theme": "default",
+    "description": "A questionnaire to demonstrate welsh langauge",
+    "eq_id": "0",
+    "groups": [
+        {
+            "id": "welsh-group",
+            "title": "",
+            "blocks": [
+                {
+                    "id": "welsh-block",
+                    "sections": [
+                        {
+                            "title": "Holiadur Cymraeg",
+                            "id": "welsh-section",
+                            "questions": [
+                            ]
+                        }
+                    ],
+                    "title": ""
+                }
+            ]
+        }
+    ]
+}
--- a/app/dev_mode/templates/dev-page.html
+++ b/app/dev_mode/templates/dev-page.html
@@ -156,6 +156,14 @@ select {
              </div>
 
             <div class="field-container">
+              <label for="language_code">Language</label>
+                <select id="language_code" name="language_code" class="qa-language-code">
+                  <option name="en" value="en">en</option>
+                  <option name="cy" value="cy">cy</option>
+                </select>
+             </div>
+
+            <div class="field-container">
               <label for="sexual_identity">Sexual Identity Flag</label>
               <input id="sexual_identity" type="checkbox" name="sexual_identity" class="qa-sexual-identity" value="true">
             </div>
--- a/app/dev_mode/views.py
+++ b/app/dev_mode/views.py
@@ -38,11 +38,26 @@ def dev_mode():
         return_by = form.get("return_by")
         employment_date = form.get("employment_date")
         region_code = form.get("region_code")
+        language_code = form.get("language_code")
         sexual_identity = form.get("sexual_identity")
         variant_flags = {"sexual_identity": sexual_identity}
-        payload = create_payload(user, exp_time, eq_id, period_str, period_id, form_type, collection_exercise_sid,
-                                 ref_p_start_date, ref_p_end_date, ru_ref, ru_name, trad_as, return_by, employment_date,
-                                 region_code, variant_flags)
+        payload = create_payload(user=user,
+                                 exp_time=exp_time,
+                                 eq_id=eq_id,
+                                 period_str=period_str,
+                                 period_id=period_id,
+                                 form_type=form_type,
+                                 collection_exercise_sid=collection_exercise_sid,
+                                 ref_p_start_date=ref_p_start_date,
+                                 ref_p_end_date=ref_p_end_date,
+                                 ru_ref=ru_ref,
+                                 ru_name=ru_name,
+                                 trad_as=trad_as,
+                                 return_by=return_by,
+                                 employment_date=employment_date,
+                                 region_code=region_code,
+                                 language_code=language_code,
+                                 variant_flags=variant_flags)
         return redirect("/session?token=" + generate_token(payload).decode())
 
     return render_template("dev-page.html", user=os.getenv('USER', 'UNKNOWN'), available_schemas=available_schemas())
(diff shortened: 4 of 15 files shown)


app/data/test_language.json
```diff
@@ -0,0 +1,45 @@
+{
+    "mime_type": "application/json/ons/eq",
+    "questionnaire_id": "0",
+    "schema_version": "0.0.1",
+    "data_version": "0.0.1",
+    "survey_id": "0",
+    "title": "English Langauage Survey",
+    "theme": "default",
+    "description": "A questionnaire to demonstrate english langauge",
+    "introduction": {
```
benjefferies: We can remove this now

app/data/test_language.json
```diff
@@ -0,0 +1,45 @@
...
+        "description": "",
+        "information_to_provide": []
+    },
+    "eq_id": "0",
+    "groups": [
+        {
+            "blocks": [
+                {
+                    "display": {
```
benjefferies: This could be removed?

app/data/test_language.json
```diff
@@ -0,0 +1,45 @@
...
+                        "properties": {}
+                    },
+                    "id": "f22b1ba4-d15f-48b8-a1f3-db62b6f34cc0",
```
benjefferies: Can we change the uuids?

app/dev_mode/views.py
```diff
@@ -74,7 +75,7 @@ def extract_eq_id_and_form_type(schema_name):
 
 
 def create_payload(user, exp_time, eq_id, period_str, period_id, form_type, collection_exercise_sid, ref_p_start_date,
```
benjefferies: Can we change the method parameters to be the form so it doesn't have the be the world record holder for method parameters

app/schema_loader/schema_loader.py
```diff
@@ -15,19 +15,29 @@ class SchemaNotFound(Exception):
...
     """
-    logging.debug("About to load schema for eq-id %s and form type %s", eq_id, form_type)
+    logging.debug("About to load schema for eq-id %s and form type %s and language %s", eq_id, form_type, language_code)
+
+    filename_format = "{}.json"
+
+    if form_type and form_type != "-1":
+        filename_format = "{}_" + filename_format
+
+    if language_code and language_code != "en":
```
benjefferies, line 34: Do we need this can't line 29 just be `filename_format = "{}_{}.json"`

app/schema_loader/schema_loader.py
```diff
@@ -15,19 +15,29 @@ class SchemaNotFound(Exception):
 logger = logging.getLogger(__name__)
 
 
-def load_schema(eq_id, form_type):
+def load_schema(eq_id, form_type, language_code='en'):
```
benjefferies, line 18: should 'en' be a default why not None?

benjefferies (changes requested):

app/dev_mode/views.py
jonnyshaw89: This is used for integration tests and we should really be sticking to the pattern of passing in whats needed as opposed to a wrapping object

app/schema_loader/schema_loader.py
jonnyshaw89: The story defined that the default should be en, i was going along the explicit over implicit rule

app/schema_loader/schema_loader.py
jonnyshaw89: no because none may match making it `{}.json` or both of these conditions may match making it `{}_{}_{}.json`

app/dev_mode/views.py
samiwel: Would passing `**kwargs` help make this more manageable?

jonnyshaw89: what about "be explicit not implicit"?

app/dev_mode/templates/dev-page.html
```diff
@@ -156,6 +156,14 @@
              </div>
 
             <div class="field-container">
+              <label for="region_code">Language</label>
```
samiwel: Minor, and only affecting the Dev page, but label is for the wrong element, should be `for="language_code"`.

app/dev_mode/views.py
```diff
@@ -38,11 +38,26 @@ def dev_mode():
         return_by = form.get("return_by")
         employment_date = form.get("employment_date")
         region_code = form.get("region_code")
+        language_code = form.get("language_code")
```
samiwel, line 41: Is there a reason why the language_code is not part of the variant flags?

app/main/views/root.py
```diff
@@ -76,7 +76,7 @@ def login():
         logger.error("Missing EQ id %s or form type %s in JWT", eq_id, form_type)
         raise NotFound
 
-    json, _ = get_schema()
+    json, _ = get_schema(metadata)
 
     navigator = Navigator(json, get_metadata(current_user), get_answer_store(current_user))
```
samiwel: Can we re-use `metadata` here rather than fetching it again?

app/dev_mode/templates/dev-page.html
jonnyshaw89: sorted

app/schema_loader/schema_loader.py
```diff
@@ -15,19 +15,29 @@ class SchemaNotFound(Exception):
...
+    if language_code and language_code != "en":
+        filename_format = "{}_" + filename_format
+
     # form type is mandatory JWT claim, but it isn't needed when getting schemas from the schema bucket
     # in that case default it to -1 and ignore it
     if form_type and form_type != "-1":
-        return load_schema_file("{}_{}.json".format(eq_id, form_type))
+        return load_schema_file(filename_format.format(eq_id, form_type, language_code))
 
     return load_s3_schema_file("{}.json".format(eq_id))
```
samiwel, line 42: Do we need to do anything when loading schemas from S3 bucket if language_code is specified?

samiwel (changes requested):

app/dev_mode/views.py
jonnyshaw89: Im not 100% sure on the reasoning but here is the PR https://github.com/ONSdigital/ons-schema-definitions/pull/12

app/main/views/root.py
jonnyshaw89: good spot

app/schema_loader/schema_loader.py
jonnyshaw89: Im not sure we use this logic at all so I've not changed its behaviour

app/dev_mode/views.py
samiwel: Fair enough, from the example in that PR, it looks like variant flags might be reserved for boolean values.

app/schema_loader/schema_loader.py
samiwel: Okay, I guess fix it when/if we need to in future.

app/data/test_language_cy.json
```diff
@@ -0,0 +1,31 @@
...
+    "groups": [
+        {
+            "id": "welsh-group",
+            "title": "",
+            "blocks": [
+                {
+                    "id": "welsh-block",
+                    "sections": [
+                        {
+                            "title": "English language schema",
```
samiwel: `"title": "Holiadur Cymraeg",`

samiwel:
Looks good in general. Wonder if it's worth just putting together a couple of integration/functional tests just to cover the language switching functionality (using the test schemas you've already created - not necessarily the census schemas).

app/dev_mode/views.py
collisdigital: We agreed that flags would only be booleans not strings or dicts/lists between the products and language may not be a variant in future (i.e it's the same schema/variant but in a different language)

collisdigital: The longer term plan wasn't to duplicate the schema like this but to hold the strings separately for translations and use the schema for structure (future target)

app/dev_mode/views.py
```diff
@@ -38,11 +38,26 @@ def dev_mode():
...
         sexual_identity = form.get("sexual_identity")
         variant_flags = {"sexual_identity": sexual_identity}
-        payload = create_payload(user, exp_time, eq_id, period_str, period_id, form_type, collection_exercise_sid,
-                                 ref_p_start_date, ref_p_end_date, ru_ref, ru_name, trad_as, return_by, employment_date,
-                                 region_code, variant_flags)
+        payload = create_payload(user=user,
```
benjefferies, line 44: This looks a lot clearer as explicitly named parameters, maybe this is all it needed!

app/utilities/schema.py
```diff
@@ -36,10 +33,10 @@ def load_and_parse_schema(eq_id, form_type):
...
 
-    json_schema = load_schema(eq_id, form_type)
+    json_schema = load_schema(eq_id, form_type, language_code)
     if json_schema:
         parser = SchemaParserFactory.create_parser(json_schema)
         schema = parser.parse()
         return json_schema, schema
     else:
-        return None
+        return None, None
```
benjefferies, line 42: Not in the scope of this change but load_schema will fail and raise either FileNotFoundError or botocore.exceptions.ClientError if it fails to find a schema, it will be caught logged and then None returned. I wonder if we should change that going forward to just raise the error log it and reraise it?

app/schema_loader/schema_loader.py
benjefferies: Probably one to remove when we get time, I'll create a trello card https://trello.com/c/w2XeaQvA/296-remove-logic-to-load-schemas-from-s3

app/utilities/schema.py
jonnyshaw89: the error gets raised in get_schema()

benjefferies (approved):

samiwel (approved):

PR #658, opened 2016-12-08
Eq 780 fix census household repeating sections

benjefferies:
Household member details should be asked for all people added in the household composition question
Visitor details page should be asked for the number of visitors answered

### How to review 
Should look like https://docs.google.com/presentation/d/1zWbLY1InINyAWNlgOU-ndHtThaOBXF4YsM6e3d0cLEw/edit#slide=id.g136caf6c9d_0_0
Selenium tests should pass
Manually test

### DIFF
--- a/app/data/census_household.json
+++ b/app/data/census_household.json
@@ -901,6 +901,14 @@
         {
             "id": "household-member",
             "title": "Household Member Details",
+            "routing_rules": [
+                {
+                    "repeat": {
+                        "type": "answer_count",
+                        "answer_id": "household-full-name"
+                    }
+                }
+            ],
             "blocks": [
                 {
                     "type": "interstitial",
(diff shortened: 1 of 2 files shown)


app/data/census_household.json
```diff
@@ -3613,7 +3621,7 @@
             ]
         },
         {
-            "id": "visitors",
+            "id": "visitors-count",
```
jonnyshaw89: maybe we could call this number-of-visitors?
as this will make part of the url

jonnyshaw89 (changes requested):

app/data/census_household.json
benjefferies: Done

jonnyshaw89 (approved):

samiwel (approved):
Looks good to me 👍

PR #660, opened 2016-12-08
Eq 781 mandatory first name

benjefferies:
* Configuring first name on household composition question to be mandatory
* Fixing generate pages script
* Adding logic to validate we are submitting to the correct page

### How to review 
Selenium tests should pass

### DIFF
--- a/tests/functional/generate_pages.py
+++ b/tests/functional/generate_pages.py
@@ -69,6 +69,46 @@ CHECKBOX_RADIO_CLICKER = r"""  click{optionName}() {
 
 """
 
+REPEATING_ANSWER_SETTER = r"""  set{answerName}(value, index = 0) {
+    var field = '{answerId}'
+    if (index > 0) {
+      field = field + '_' + index
+    }
+    browser.setValue('[name="' + field + '"]', value)
+    return this
+  }
+
+"""
+
+REPEATING_ANSWER_GETTER = r"""  get{answerName}(index) {
+    var field = '{answerId}'
+    if (index > 0) {
+      field = field + '_' + index
+    }
+    browser.element('[name="' + field + '"]').getValue()
+    return this
+  }
+
+"""
+
+REPEATING_ANSWER_ADD_REMOVE = r"""  addPerson() {
+    browser.click('button[name="action[add_answer]"]')
+    return this
+  }
+
+  removePerson(index) {
+    browser.click('button[value="' + index + '"]')
+    return this
+  }
+
+"""
+
+CONSTRUCTOR = r"""  constructor() {
+    super('{block_id}')
+  }
+
+"""
+
 FOOTER = r"""}
 
 export default new {pageName}Page()
@@ -88,8 +128,8 @@ def process_options(answer_id, options, page_spec):
         page_spec.write(CHECKBOX_RADIO_CLICKER.replace("{optionName}", generate_camel_case_from_id(answer_id) + option_name).replace("{optionId}", option_id))
 
 
-def process_answer(answer, page_spec):
-    if answer['type'] == 'Radio' or answer['type'] == 'Checkbox':
+def process_answer(question_type, answer, page_spec):
+    if answer['type'] == 'Radio' or answer['type'] == 'Checkbox' or answer['type'] == 'Relationship':
         process_options(answer['id'], answer['options'], page_spec)
 
     elif answer['type'] == 'Date':
@@ -103,8 +143,12 @@ def process_answer(answer, page_spec):
     elif (answer['type'] == 'TextField' or answer['type'] == 'Integer' or
             answer['type'] == 'PositiveInteger' or answer['type'] == 'TextArea'):
         answer_name = generate_camel_case_from_id(answer['id'])
-        page_spec.write(ANSWER_SETTER.replace("{answerName}", answer_name).replace("{answerId}", answer['id']))
-        page_spec.write(ANSWER_GETTER.replace("{answerName}", answer_name).replace("{answerId}", answer['id']))
+        if question_type == 'RepeatingAnswer':
+            page_spec.write(REPEATING_ANSWER_SETTER.replace("{answerName}", answer_name).replace("{answerId}", answer['id']))
+            page_spec.write(REPEATING_ANSWER_GETTER.replace("{answerName}", answer_name).replace("{answerId}", answer['id']))
+        else:
+            page_spec.write(ANSWER_SETTER.replace("{answerName}", answer_name).replace("{answerId}", answer['id']))
+            page_spec.write(ANSWER_GETTER.replace("{answerName}", answer_name).replace("{answerId}", answer['id']))
 
     else:
         raise Exception('Answer type [%s] not configured' % answer['type'])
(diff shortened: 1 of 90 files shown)


app/data/census_individual.json
```diff
@@ -34,23 +34,23 @@
                                             "label": "First Name",
                                             "mandatory": false,
```
jonnyshaw89: Should this one also be mandatory?

tests/functional/generate_pages.py
```diff
@@ -90,10 +131,15 @@ def process_answer(answer, page_spec):
         answer_name = generate_camel_case_from_id(answer['id'])
         page_spec.write(_write_month_year_date_answer(answer_name, answer['id']))
 
-    elif answer['type'] == 'TextField' or answer['type'] == 'Integer' or answer['type'] == 'Textarea':
-        answer_name = generate_camel_case_from_id(answer['id'])
-        page_spec.write(ANSWER_SETTER.replace("{answerName}", answer_name).replace("{answerId}", answer['id']))
-        page_spec.write(ANSWER_GETTER.replace("{answerName}", answer_name).replace("{answerId}", answer['id']))
+    elif answer['type'] == 'TextField' or answer['type'] == 'Integer' or answer['type'] == 'TextArea':
+        if question_type == 'RepeatingAnswer':
+            answer_name = generate_camel_case_from_id(answer['id'])
```
jonnyshaw89: define this outside of the if so you don't also have to define it in the else

tests/functional/spec/census-household.spec.js
```diff
@@ -83,7 +83,7 @@ describe('Census Household', function () {
 
         // who-lives-here
         PermanentOrFamilyHome.clickPermanentOrFamilyHomeAnswerYes().submit()
-        HouseholdComposition.setPersonName(0, 'John Smith').submit()
+        HouseholdComposition.setFirstName(0, 'John').submit()
```
jonnyshaw89: do we have any scenario where we enter middle and last name?

jonnyshaw89 (changes requested):

tests/functional/spec/census-household.spec.js
benjefferies: Don't think we need to. We need to test the scenario that the middle/last is optional. Maybe that would be good as an explicit test. As for the second test I am going to delete, it's not really a selenium test, it is unit tested

tests/functional/pages/surveys/census/household/address-type.page.js
```diff
@@ -2,6 +2,11 @@ import MultipleChoiceWithOtherPage from '../../multiple-choice.page'
 
 class AddressTypePage extends MultipleChoiceWithOtherPage {
 
+  isOpen() {
```
jonnyshaw89: Can this not move up the the question page class and then just provide the blockId in each file?

tests/functional/spec/census-household.spec.js
samiwel: The Jina filter I created should display middle and last names appropriately if only middle and/or last were provided. But the fact that we've added mandatory validation here will help to ensure that we always have something, which is good.

samiwel:
I noticed that the validation messages appears correctly when the Save and continue button is pressed, but when I the add new person button is pressed the error was not displayed yet I was prevented from adding additional household members (presumably due to failed validation).

Does it make sense to ignore the validation check when adding new people? So that users can add as many new people as they like, but fail the validation as soon as they hit save and continue? From a UX point of view, I guess that is more consistent with the way that screen behaves when JS is enabled on that page.

benjefferies:
I think your right Sam, from a UX views they should act the same. I have removed validation on the add/remove action

tests/integration/questionnaire/test_questionnaire_household_composition.py
```diff
@@ -0,0 +1,29 @@
...
+        post_data = {
+            "action[add_answer]": ""
+        }
+
+        # When
+        resp = self.client.post(self.household_composition_url, data=post_data)
+
+        # Then
+        self.assertEqual(200, resp.status_code)
+        self.assertNotIn('This field is mandatory.', resp.get_data(True))
```
benjefferies: If we returned a status code of 400 for invalid answers we wouldn't need to check the body making testing end points nicer. I know in REST apis status codes are ubiquitous but not sure how it fits in outside REST i.e. will some browsers intervene (probably IE)

jonnyshaw89:
@benjefferies @samiwel i think the removing of validation isn't a good thing. if i click add lots of times, when go previous and say no one lives here, then go forward i get repeated around lots of groups of people with empty names.

app/main/views/questionnaire.py
```diff
@@ -237,18 +237,19 @@ def post_household_composition(eq_id, form_type, collection_id, group_id):
         'group_id': group_id,
         'group_instance': 0,
     }
+    if 'action[add_answer]' in request.form or 'action[remove_answer]' in request.form:
+        questionnaire_manager.build_state(this_block, request.form)
```
jonnyshaw89: if my request.form contained answers for other blocks would they be applied without being validated. I'm a little worried that we have exposed our answer store to be open to abuse by not validating.

tests/functional/generate_pages.py
```diff
@@ -46,6 +46,13 @@
 
 """
 
+ANSWER_SELECT = r"""  set{answerName}(value) {
```
jonnyshaw89: ANSWER_SELECT -> ANSWER_SELECTOR?

jonnyshaw89 (changes requested):

jonnyshaw89:
If you add a person with no name, click previous then go forwards again. you can then not ever remove them.

tests/functional/pages/surveys/question.page.js
```diff
@@ -1,10 +1,23 @@
...
+  }
+
   getAlertText() {
     return browser.element('[data-qa="error-body"]').getText()
   }
 
   submit() {
+    if (this.pageName != undefined && !this.isOpen()) {
+    const url = browser.url().value
+        throw new Error(`Expected to be on ${this.pageName} page but was on ${browser.getUrl()}`)
```
jonnyshaw89: Kudos

jonnyshaw89:
I have created an issue (https://github.com/ONSdigital/eq-survey-runner/issues/727) for the scenario i found above. But given that that existed before this PR I'm happy.

jonnyshaw89 (approved):

samiwel (approved):
Looks good to me 👍

tests/functional/spec/census-household.spec.js
```diff
@@ -387,6 +387,25 @@ describe('Census Household', function () {
...
+        startCensusQuestionnaire('census_household.json', false, 'GB-WLS')
+
+        // who-lives-here
+        PermanentOrFamilyHome.clickPermanentOrFamilyHomeAnswerYes().submit()
+        HouseholdComposition.submit()
+
+        expect(HouseholdComposition.getErrorMsg()).to.contain('Please enter a name or remove the person to continue')
+    })
+
+    it('Given I enter a first name but no middle or surname, When I save and continue, Then I should see errors that first name is mandatory.', function () {
```
ajmaddaford: Should this say no errors?

ajmaddaford (changes requested):

tests/functional/spec/census-household.spec.js
benjefferies: Your sharp-eyed, always catching me out on these!

ajmaddaford (approved):

PR #661, opened 2016-12-08
Navigator Location refactor

iwootten:
Updates the Navigator to use a Location class for representation of group/instance/block locations and for lists of them to be returned on requesting paths. 

Includes work from #631.

### How to review 
Take a look at the approach, see if you approve.

### DIFF
--- a/app/questionnaire/location.py
+++ b/app/questionnaire/location.py
@@ -1,0 +1,15 @@
+class Location(object):
+
+    def __init__(self, group_id, group_instance, block_id):
+
+        self.group_id = group_id
+        self.group_instance = group_instance
+        self.block_id = block_id
+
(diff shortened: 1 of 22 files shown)


app/questionnaire/location.py
```diff
@@ -0,0 +1,15 @@
+class Location(object):
+
+    def __init__(self, group_id, group_instance, block_id):
+
+        self.group_id = group_id
+        self.group_instance = group_instance
+        self.block_id = block_id
+
+    def __eq__(self, other):
```
collisdigital, line 15: Usually best to also define `__hash__` when defining `__eq__` e.g. http://stackoverflow.com/questions/1608842/types-that-define-eq-are-unhashable

PR #662, opened 2016-12-08
Add suites to functional tests and default to just core suite

ajmaddaford:
The Census tests take a long time to complete and are holding up builds. Two suites have been created - core and census. The 'core' suite contains the core functionality tests and the 'census' suite contains all of the extended Census tests (issues, routing etc).

By default run_tests.sh now just runs the core tests. To run all tests from the command line you can still run 'yarn test_functional', or to run just the extended Census tests you can use `yarn test_functional -- --suite=census`.

As Travis uses run_tests.sh, it will only run the core tests.

### How to review 
- Run `./scripts/run_tests.sh`, verify it passes and only core tests are run
- Run `yarn test_functional -- --suite=census`, verify only the extended census tests are run

### DIFF
--- a/scripts/run_tests.sh
+++ b/scripts/run_tests.sh
@@ -40,5 +40,18 @@ display_result $? 2 "Unit tests"
 
 # Run front end tests
 npm config set python python2.7
-yarn test
-display_result $? 1 "Front end tests"
+
+yarn test_unit_no_watch
+display_result $? 1 "Front end unit tests"
+
+if [ -z "$EQ_FUNCTIONAL_TEST_SUITES" ]; then
+    if [ "${TRAVIS_BRANCH}" = "master" ]; then
+        export EQ_FUNCTIONAL_TEST_SUITES="core,census"
+    else
+        export EQ_FUNCTIONAL_TEST_SUITES="core"
+    fi
+fi
+echo "Running front end functional tests [${EQ_FUNCTIONAL_TEST_SUITES}]"
+yarn test_functional -- --suite ${EQ_FUNCTIONAL_TEST_SUITES}
+
+display_result $? 1 "Front end functional tests"
--- a/README.md
+++ b/README.md
@@ -90,6 +90,7 @@ The functional tests can be executed with a couple of options.
 
 `--env=preprod` Will run the test against preprod.
 `--spec=mci` Will run a single test spec (`mci.spec.js`) instead of the entire suite.
+`--suite=core` Will run a suite of tests (core) instead of the entire suite.
 
 These options can be combined with `test_functional` or `test_functional_sauce`, eg:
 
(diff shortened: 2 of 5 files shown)


benjefferies:
I wonder if we could do something via travis environmental variables to tigger different suites. We could use the branch to do this by the looks of it.

```
suit="core"
IF [[ $TRAVIS_PULL_REQUEST_BRANCH == *"census"* ]] THEN
    suit = "$suit, census" 
FI

yarn test_functional -- --suite $suit
```

https://docs.travis-ci.com/user/environment-variables/#Default-Environment-Variables

jonnyshaw89 (changes requested):
Would it be worth adding a condition to the .travis file so the census tests get run on master? that way its non blocking but still runs?

ajmaddaford:
@benjefferies I don't think any of our census feature branches tend to include the word census though. Also doesn't help with reducing build times as everything we're working on at the moment is census related.
@jonnyshaw89 Good idea, I'll take a look into that

benjefferies:
@ajmaddaford travis has an environmental variable for that `TRAVIS_BRANCH`

benjefferies (changes requested):
We should move the census household/individual etc into the census directory

ajmaddaford:
@jonnyshaw89 @benjefferies I've updated the run_tests script to run all tests for developers and on master
@benjefferies we want a census test to run on every pull request, so those census household/individual tests should stay there (we do need to change them so it does just one run through the whole census though)

scripts/run_tests.sh
```diff
@@ -40,7 +40,17 @@ display_result $? 2 "Unit tests"
 
 # Run front end tests
 npm config set python python2.7
+
 yarn test_unit_no_watch
 display_result $? 1 "Front end unit tests"
-yarn test_functional -- --suite core
+
+if [ "$TRAVIS" ]; then
+    if [ "${TRAVIS_BRANCH}" = "master" ]; then
```
jonnyshaw89, line 48: is it possible to do a not equals? then you wouldn't have to define `yarn test_functional` twice

jonnyshaw89 (approved):
Im happy with this. i guess we won't be sure it works until its in master.

scripts/run_tests.sh
```diff
@@ -40,5 +40,15 @@ display_result $? 2 "Unit tests"
 
 # Run front end tests
 npm config set python python2.7
-yarn test
-display_result $? 1 "Front end tests"
+
+yarn test_unit_no_watch
+display_result $? 1 "Front end unit tests"
+
+SUITE="core"
```
benjefferies: It's good to use lowercase variables in bash scripts to avoid possible clashes with env variables

ajmaddaford: I've changed the script to be configurable via an external env var now

scripts/run_tests.sh
ajmaddaford: Re-worked the script

benjefferies (approved):

jonnyshaw89:
👍 LGTM

Issue #684, opened 2016-12-09
Census: Navigation from Question '28 Last week were you:' was incorrect when more than 2 options out of the top 6 options are selected


yoganandkunche:
### Expected behaviour

Must skip the next 5 questions

### Actual behaviour
No Questions are skipped and application simply navigates to next question which is inccorect.
**Note:** Next five questions are skipped when only one option from the top 6 option(s) are selected

### Steps to reproduce the behaviour
Launch Census_household survey
Navigate to Question 26. Last week were you:
And select any two or more options from the top 6 options and click save and continue

#### Browser
All

yoganandkunche:
@jonnyshaw89 Re-tested the issue #684 one scenarios seems to be not working. Lets discuss on Monday

yoganandkunche:
One scenario is failing. When respondent selects any of the first 6 checkboxes along with the 7 the checkbox it should still skip the next 5 questions which it is not doing. 
demonstrated the scenario to @jonnyshaw89 

I am reopening the issue and also moved the relevant card on Trello to development

yoganandkunche:
Defect passed retest from release on 15th Dec 2016

PR #683, opened 2016-12-09
Updated census household & communal establishment schemas to match slides

jonnyshaw89:
### What is the context of this PR?
Updating the census schemas to match the defined content in the slides

Fixes #665
Fixes #680
Fixes #682 
Fixes #684
Fixes #685
Fixes #698
Fixes #700
Fixes #701

### How to review 
Check that content match slides

PR #687, opened 2016-12-10
Routing on one of multiple selected checkbox answers

jonnyshaw89:
Added a contains condition to routing rules.
When a routing rule is specified on a checkbox, the routing rule checks if a value is one of the selected answers.

Fixes https://github.com/ONSdigital/eq-survey-runner/issues/684

### How to review 
On the census-household questionnaire test the routing of the employment-type question


collisdigital:
Also needed on the which passport screen I think

collisdigital:
This also fixes some issues reported by yogi, can we add them to the description?

collisdigital:
https://github.com/ONSdigital/eq-survey-runner/issues/684

jonnyshaw89:
@collisdigital i can't add this to the passports question as the question it routes to is added in here https://github.com/ONSdigital/eq-survey-runner/pull/683 so need this to go in then i can merge into the other branch

yoganandkunche:
@jonnyshaw89 Re-tested the issue #684 one scenarios seems to be not working. Lets discuss on Monday

jonnyshaw89:
@yoganandkunche i have corrected the routing on the none of the above option

yoganandkunche:
@jonnyshaw89 As discussed selecting the 7 option along with any/all the 1-6 options must skip the next questions

yoganandkunche:
Not merged into master yet. So will be testing it in pre-prod tomorrow as I will be busy with e2e testing today

app/questionnaire/rules.py
```diff
@@ -8,19 +8,15 @@ def evaluate_rule(when, answer_value):
...
     condition = when['condition']
 
-    if isinstance(answer_value, list):
-        if len(answer_value) == 1:
-            answer_to_test = answer_value[0]
-        else:
-            return False
-    else:
-        answer_to_test = str(answer_value)
+    answer_to_test = str(answer_value) if 'equals' in condition else answer_value
```
iwootten: 'equals' in condition 
or

    condition == 'equals'?

iwootten (changes requested):

app/questionnaire/rules.py
iwootten: Ah is this converting everything with 'equals' in the title to a string? If so, then you probably should be explicit about it:

     if condition in ['equals', 'not equals']

iwootten (changes requested):

app/questionnaire/rules.py
samiwel: I agree with Ian's comment here. It's wasn't immediately obvious that you're matching `equals` and `not equals` conditions. Probably better to make that explicit.

samiwel (commented):
Looks good to me.
Do any of the census schemas have `contains` routing where an input field is displayed upon selecting Other?  I didn't see any examples of that in this PR. Are these all the cases?

samiwel (approved):

iwootten (approved):

PR #691, opened 2016-12-11
Fixes #667 other value not retained throughout census survey.

### LINKED ISSUE #667: Census: Text entered in other fields is not retained through out the survey
### Expected behaviour
Text entered in any of the other(Please specify) fields must be retained

### Actual behaviour
Texted entered in Other (please specify) are not retained when respondent navigates backwards

Launch Census Survey
Navigate to any radio or Checkbox field with Other option
select other and enter text in the field and click save and continue
Click [Previous]

samiwel:
Fixes #667 other value not retained throughout census survey.

Turns out there were two issues:

1. Case sensitive match for 'other'. In the census schemas the value to be matched is 'Other'. I changed this to perform a case insensitive match when checking to see if an other value was posted.

2. When restoring the Other value an assumption was being made that the other option in the radio had the label 'Other'. This has now been changed to get the label from the schema option.

### How to review 
Re-test the issue #667.
Observe that the text entered in the 'other' text field in radio options is retained.
All tests should pass.

### DIFF
--- a/app/questionnaire_state/state_answer.py
+++ b/app/questionnaire_state/state_answer.py
@@ -1,6 +1,8 @@
 from app.data_model.answer_store import Answer
 from app.questionnaire_state.state_item import StateItem
 
+from werkzeug.datastructures import MultiDict
+
 
 class StateAnswer(StateItem):
     def __init__(self, id, schema_item):
@@ -29,16 +31,27 @@ class StateAnswer(StateItem):
         # Get radio options from the schema.
         for option in self.schema_item.options:
             # If the radio answer has an Other option...
-            is_other_radio_option = option['label'] == 'Other' and 'other' in option
+            if 'other' in option:
 
-            if is_other_radio_option:
+                other_option = option
+
                 schema_options = [option['value'] for option in self.schema_item.options]
 
                 if self.input not in schema_options:
                     # Input is not a pre-defined answer so it must be the user-entered value for other
-                    self.input = 'other'
-                    self.value = 'other'
+                    self.input = option['value']
+                    self.value = option['value']
                     self.other = self.schema_item.get_user_input(user_input)
+
+                elif self.input == other_option['value'] and isinstance(user_input, MultiDict):
+                    # User selected other but typed in a value which is already in schema
+                    other_input_field = user_input.getlist(self.id)[-1:]
+                    if other_input_field:
+                        value = other_input_field[0]
+                        if value in schema_options:
+                            self.input = value
+                            self.value = value
+                            self.other = None
 
     def get_answers(self):
         return [self]
--- a/app/schema/widgets/multiple_choice_widget.py
+++ b/app/schema/widgets/multiple_choice_widget.py
@@ -62,8 +62,10 @@ class MultipleChoiceWidget(Widget, metaclass=ABCMeta):
         Compare the posted_data with the options in the schema, if there is a value which doesn't match
         the options it must be the other value
         """
-        if posted_data and 'other' in posted_data:
+        if posted_data and 'other' in (value.lower() for value in posted_data):
             for answer in posted_data:
-                if answer and not answer.isspace() and not any(option['value'] == answer for option in options) and answer is not 'other':
+                answer_in_options = any(option['value'] == answer for option in options)
+                if answer and not answer.isspace() and not answer_in_options and answer.lower() != 'other':
                     return answer
+
         return None
--- a/tests/app/questionnaire_state/test_state_answer.py
+++ b/tests/app/questionnaire_state/test_state_answer.py
@@ -1,0 +1,89 @@
+import unittest
+
+from werkzeug.datastructures import MultiDict
+
+from app.questionnaire_state.state_answer import StateAnswer
+from app.schema.answer import Answer
+from app.schema.widgets.text_widget import TextWidget
+
+
+class TestStateAnswer(unittest.TestCase):
+
+    def setUp(self):
+        self.answer_schema = Answer('multiple-choice-with-other')
+        self.answer_schema.widget = TextWidget(self.answer_schema.id)
+        self.answer_schema.options = [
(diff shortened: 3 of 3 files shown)


samiwel:
I'm going to add a `not-ready-to-merge` label to this PR. Feel free to start reviewing because it should fix the raised defect. But there's a few additional scenarios I want to test, namely:

1. Other field for checkbox questions, and
2. When the Other option is not the last option.

Both of these scenarios are found in various census test questions. So if there's any additional fixes to be made in these areas, it makes sense to fix them under this PR.

collisdigital:
@samiwel we agreed with Census that 'Other' with please specify will always be the last option for the test as we don't support it anywhere else. What Q's is this on?

samiwel:
@collisdigital As I was going through the individual questionnaire yesterday, I noticed it. I made a note of it on the Trello card. So it may just be a case of swapping the options around in the schema then.

LJBabbage:
There is code in checkbox_answer.py which already has the capacity to find the right 'other' value, regardless of each position, however as @collisdigital mentioned above it was decided to put it last meaning it de-scoped itself

samiwel:
I tested checkboxes with other fields in the individual census schema on this branch and it seemed to behave correctly. Removing the `not-ready-to-merge` label.

app/questionnaire_state/state_answer.py
```diff
@@ -23,21 +23,19 @@ def update_state(self, user_input):
         self.value = self.schema_item.get_user_input(user_input)
         self.other = self.schema_item.get_other_value(user_input)
         if self.schema_item.type == 'Radio' and self.input:
-            self._restore_other_value(user_input)
+            self.restore_other_value(user_input)
 
-    def _restore_other_value(self, user_input):
+    def restore_other_value(self, user_input):
```
LJBabbage: Why has this been changed to public?

app/schema/widgets/multiple_choice_widget.py
```diff
@@ -62,8 +62,9 @@ def find_other_value(posted_data, options):
         Compare the posted_data with the options in the schema, if there is a value which doesn't match
         the options it must be the other value
         """
-        if posted_data and 'other' in posted_data:
-            for answer in posted_data:
-                if answer and not answer.isspace() and not any(option['value'] == answer for option in options) and answer is not 'other':
+        if posted_data and 'other' in (value.lower() for value in posted_data):
+            for answer in filter(None, [value.strip() for value in posted_data]):
```
LJBabbage: Not a fan of this double for loop filter to just get rid of spaces, prefer the other way

LJBabbage (changes requested):

app/schema/widgets/multiple_choice_widget.py
```diff
@@ -62,8 +62,9 @@ def find_other_value(posted_data, options):
         Compare the posted_data with the options in the schema, if there is a value which doesn't match
         the options it must be the other value
         """
-        if posted_data and 'other' in posted_data:
-            for answer in posted_data:
-                if answer and not answer.isspace() and not any(option['value'] == answer for option in options) and answer is not 'other':
+        if posted_data and 'other' in (value.lower() for value in posted_data):
```
LJBabbage, line 65: Doesn't quite seem right to make everything lower case to just check the other value isn't Other

LJBabbage (changes requested):

app/schema/widgets/multiple_choice_widget.py
samiwel: Is there a better way to perform a case insensitive match?

app/questionnaire_state/state_answer.py
samiwel: It was changed so that I could unit test its behaviour, but you're right, it doesn't need to be public. I'll revert this change.

app/schema/widgets/multiple_choice_widget.py
samiwel: The other way doesn't get rid of the spaces. Maybe that's valid, but even if not it should arguably be done in `check_user_input` in the widget instead. I'll put this back the way it was for now.

jonnyshaw89:
@samiwel The only issue i can find with this is that if you enter a value in the text box which is the same as one of the options to click then the value is not retained? Im not sure how that sits in terms of requirements though?

samiwel:
@jonnyshaw89 Thanks for letting me know about that. Should be fixed by my latest commit.

app/questionnaire_state/state_answer.py
```diff
@@ -38,6 +41,16 @@ def _restore_other_value(self, user_input):
                     self.value = option['value']
                     self.other = self.schema_item.get_user_input(user_input)
 
+                elif self.input == other_option['value'] and hasattr(user_input, 'getlist'):
```
jonnyshaw89: is the hasattr checking if its a MultiDict?

jonnyshaw89 (changes requested):

jonnyshaw89:
@samiwel Just re-tested. it now selects the value i typed, which i guess is acceptable behaviour

app/questionnaire_state/state_answer.py
samiwel: Yeah, unfortunately the logic gets called on GET and POST. When POST-ing an `ImmutableMultiDict` gets passed in. This whole `other` solution needs to be re-visited, I feel.
At the moment we're sort of storing an answer inside an answer and that causes quite a lot of unnecessary complexity - perhaps we can change this when we bring in the WTF-Forms solution.

jonnyshaw89 (approved):

LJBabbage (approved):

jonnyshaw89:
👍 @samiwel rebase and squash?

yoganandkunche:
Defect passed re-test. This defect is tested across all the census schemas and works as expected.

PR #693, opened 2016-12-11
Fixes #692: Refactor build_path in navigator.py to be iterative, not …

### LINKED ISSUE #692: More than 26 householders causes stack overflow
### Expected behaviour
- You can add more than 26 household members

### Actual behaviour
- Recursive function call for 26 people causes a stack overflow (finally an issue with stack overflow in the title; http://stackoverflow.com/questions/tagged/python)

- Add 27 people to household composition

### Technical information
- Caused by build_path in navigator.py being recursive, exhausts Python stack when repeats gets too long.

#### Browser
- Chrome

#### Operating System
- macOS Sierra

### Screenshot
- N/A

collisdigital:
Fixed #692 by changing build_path in navigator.py to be iterative and not recursive (avoiding stack exhaustion). 

Also changed repeating answer question to use shallow copy as deep copy is massive performance bottleneck (30 seconds to finish for 26 people).

Note there is more refactoring that should be done in build_path (such as break into some smaller functions) but whilst all other PRs are in-flight this is the minimal change to allow easier merging; once all PRs are merged this can be refactored from master.

More is also more work needed on performance later, first step is to allow build_path to exit early (e.g. don't go all the way to end of schema if you only need next block), later may be to convert this and calling code to an iterator to avoid building full path up-front and exit when done etc. Will add tech debt card to cover these items.

### How to review 
- Check for any issues in the change from recursive to iterative
- Flag up any problems with moving to a shallow copy for the widget state

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,6 @@ cache:
   - $HOME/.yarn-cache
 install:
 - . $HOME/.nvm/nvm.sh
-- export NVM_NODEJS_ORG_MIRROR=https://npm.taobao.org/mirrors/node
 - nvm install v6.9.1
 - pip install -U pip wheel
 - pip install --require-hashes -r requirements.txt
--- a/app/questionnaire/navigator.py
+++ b/app/questionnaire/navigator.py
@@ -43,55 +43,73 @@ class Navigator:
                         b["group_id"] == location['group_id'] and b['group_instance'] == location['group_instance'])
         return None
 
-    def build_path(self, blocks, this_location, path):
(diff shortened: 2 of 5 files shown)


collisdigital:
Full selenium test suite run locally and pass

samiwel (approved):
Looks good to me. There's a few refactoring that could be done, as you mentioned in the PR, but overall a good change I think it is easier to reason about than the recursive call.

benjefferies (approved):

benjefferies:
👍

PR #628, opened 2016-12-05
Navigation support with repeating groups

LJBabbage:
### What is the context of this PR?
To add in navigation for the user. Each group appears in the navigation unless it has a repeating rule which hasn't been completed then it is hidden. If a group is repeating and the feeding question has been answered it then repeats the required amount. Navigation is always linear, unless you reach the end, and it then goes to the last completed block. The previous link is not present on the first block of each group.

Fixes #686

PR #694, opened 2016-12-11
Fixes #686 Previous Link on the first question gives error.

### LINKED ISSUE #686: Census: Previous Link on the first question gives error
### Expected behaviour
No previous link on first question.

### Actual behaviour
Previous Link exists and when clicked gives error 404

Launch Census survey from RH
Click previous link on first question

### Screenshot
screen shot 2016-12-10 at 09 43 04

samiwel:
Fixes #686 by not displaying the Previous link if it's the first block in the survey.

### How to review 
Observe that the Previous link is not displayed on the first block of any survey.

### DIFF
--- a/app/templates/questionnaire.html
+++ b/app/templates/questionnaire.html
@@ -63,10 +63,11 @@
 
       <button class="btn btn--primary qa-btn-submit venus" type="submit" name="action[save_continue]">{% block submit_button_text %}Save and continue{% endblock %}</button>
       <br/>
+      {% if state.group_instance > 0 or not block.first_in_schema() %}
       <div class="u-mt-xs">
         <a id="bottom-previous" class="mars" href="{{previous_location}}">Previous</a>
       </div>
-
+      {% endif %}
     </form>
   </div>
 </div>
--- a/app/schema/block.py
+++ b/app/schema/block.py
@@ -26,3 +26,8 @@ class Block(Item):
 
     def get_state_class(self):
         return StateBlock
+
+    def first_in_schema(self):
+        first_block_in_group_blocks = self.container.blocks.index(self) == 0
+        first_group_in_questionnaire_groups = self.container.container.groups.index(self.container) == 0
+        return first_block_in_group_blocks and first_group_in_questionnaire_groups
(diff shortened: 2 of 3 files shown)


app/templates/questionnaire.html
```diff
@@ -63,10 +63,11 @@ <h1 class="alert__title venus">This page has {{all_errors|length}} errors</h1>
 
       <button class="btn btn--primary qa-btn-submit venus" type="submit" name="action[save_continue]">{% block submit_button_text %}Save and continue{% endblock %}</button>
       <br/>
+      {% if state.group_instance > 0 or not block.first_in_schema() %}
```
collisdigital, line 66: why the `state.group_instance > 0` ?

samiwel: For repeating groups the block could still be the first in the schema so we need this check to allow going backwards in repeating groups.

collisdigital: Except the navigation PR hides the previous link for the first block in a group; so I guess this will conflict (or merge out) when that lands?

samiwel: There may be some overlap here, not sure I'll need to check with Lewis/Hamish tomorrow.
If the navigation PR fixes this defect already then it should reference the issue.

collisdigital: Doh! Suspect that is the case...

iwootten, line 66: I don't think adding this sort of logic to the schemas is the way to go for resolving this issue. The previous location is calculated before rendering the template and presumably should be none for the first pages on census schemas, so isn't it easier just to check it isn't None?

iwootten (changes requested):

app/templates/questionnaire.html
samiwel: Hi Ian, I think this defect is already fixed by PR #628 so there's probably no need to make this change after all.

LJBabbage:
Hold fire on this one, the navigation branch will indeed fix this. Its done in the backend so there is no logic in the front end

iwootten:
Closing this given seems like nav is going to fix original issue #686.

PR #695, opened 2016-12-11
Fixes #690 Quotes in household name causes error

### LINKED ISSUE #690: Entering double quote or backslash in household composition breaks household summary
### Expected behaviour
You can enter any value into any name field and there is no error

### Actual behaviour
Entering a double quote " or backslash \ into a name field causes a 500 error on the household summary screen.

- Start Census Household
- Get to name entry screen
- Enter a double quote or backslash into any name field
- Save and continue
- See 500 error screen.

#### Browser
- Chrome

#### Operating System
-  macOS Sierra

### Screenshot
- N/A

samiwel:
Fixes #690.
Render function gets called both with dict and string in different places.
When called with string (substituting piping values) need to escape characters to ensure valid JSON.

### How to review 
Re-test the defect.
Ensure can enter quotes or backslashes for household member names.
Names should be displayed appropriately.

### DIFF
--- a/tests/functional/config.js
+++ b/tests/functional/config.js
@@ -24,7 +24,11 @@ let config = {
   sync: true,
   connectionRetryTimeout: 5000,
   connectionRetryCount: 3,
-  capabilities: [chrome],
+  capabilities: [{
+    name: 'Chrome (local)',
+    browserName: 'chrome',
+    maxInstances: 1
+  }],
   framework: 'mocha',
   reporters: ['spec'],
   mochaOpts: {
--- a/tests/functional/config.js
+++ b/tests/functional/config.js
@@ -24,11 +24,7 @@ let config = {
   sync: true,
   connectionRetryTimeout: 5000,
   connectionRetryCount: 3,
-  capabilities: [{
-    name: 'Chrome (local)',
-    browserName: 'chrome',
-    maxInstances: 1
-  }],
+  capabilities: [chrome],
   framework: 'mocha',
   reporters: ['spec'],
   mochaOpts: {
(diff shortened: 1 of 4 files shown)


collisdigital:
Putting `<Chris` in the first name causes the name to be lost, is it related?

collisdigital:
Actually putting `<` anywhere in the string loses all content after it..

samiwel:
I think they should be separate issues. One is to escape JSON when piping. The other is to bleach input values to escape HTML. I started looking at bleaching the input and got loads of failing tests.

jonnyshaw89:
I think this may be related to https://github.com/ONSdigital/eq-survey-runner/issues/699 arguably there should be no quotes in the name if the data is being bleached correctly?

samiwel:
@jonnyshaw89 Not sure about that, what if you had an apostrophe in the name e.g. O'Neil?

jonnyshaw89:
but that should be stored in the json as "O\\'Neil" which then wouldn't error when being parsed by the template_renderer?

jonnyshaw89 (approved):

tests/functional/config.js
```diff
@@ -24,7 +24,11 @@ let config = {
   sync: true,
   connectionRetryTimeout: 5000,
   connectionRetryCount: 3,
-  capabilities: [chrome],
```
benjefferies, line 27: Do we need to change the default here?

samiwel: @benjefferies Not sure, I just had to merge this, did I pick the wrong change?

benjefferies: Yes looks like a merge issue

samiwel: Okay, I'll revert this change, probably best to stick with what's in master already.

benjefferies (approved):

PR #696, opened 2016-12-12
Add Household composition real time (client side) changes

ajmaddaford:

Enables the "realtime" adding/removing of household members in the household composition question.

In order to guarantee there will be no side effects for future dependent groups (relationships, household member details) we remove all answers for all household members on POST of the household composition screen. This is not the greatest user experience, but future stories will improve on this and only wipe what is necessary.

This has modified the functionality of the remove method in the answer store to remove any answers identified by filter criteria. The original remove method has been renamed to remove_answer.

### How to review 

1. Visit `test_repeating_household`
2. Complete household question, adding several members
3. Edit from summary screen, ensuring edited values are carried through.


benjefferies (approved):

app/data_model/answer_store.py
```diff
@@ -195,27 +196,37 @@ def map(self, group_id=None, block_id=None, answer_id=None, group_instance=None,
...
+        for answer in self.filter(group_id, block_id, answer_id, group_instance, answer_instance):
             answer_id = answer['answer_id']
             answer_id += "_" + str(answer['answer_instance']) if answer['answer_instance'] > 0 else ''
 
             result[answer_id] = answer['value']
 
         return OrderedDict(sorted(result.items(), key=lambda t: natural_order(t[0])))
 
-    def remove(self, answer):
+    def remove_answer(self, answer):
```
samiwel, line 207: Just wondering if we still need both remove methods. Perhaps we do, but do you think the remove_answer method of questionnaire manager can/should be updated to call the remove method (below) instead?

ajmaddaford: Yeah, @iwootten has just mentioned it as well. We need this merged as it's holding up other pull requests, so I'll create a card for the potential refactoring.

samiwel: Okay yeah happy with changes otherwise, we can re-work this separately.

samiwel (approved):

iwootten (approved):

PR #711, opened 2016-12-13
Fixes #705 - routing on question 1 of CE schema wrong

### LINKED ISSUE #705: Census_Communal: Navigation is not correct when user select No from Question 1 under Usual residents section
### Expected behaviour
When option no is selected must navigate to Question 3 under Section: **Completion Preferences**

3. If tomorrow was Census Day, how would you (or a member of staff) complete the Census establishment questionnaire?

### Actual behaviour
Navigating to question 3 within the same section.

Launch Census_Houshold survey
Navigate to Question 1 under Usual residents section
Select 'No' andc click Save and continue
#### Browser
All

collisdigital:
Fixes #705 - Routing bug in CE schema

### How to review 
Check fix against issue and that Selenium test is valid.

### DIFF
--- a/tests/functional/spec/census-communal.spec.js
+++ b/tests/functional/spec/census-communal.spec.js
@@ -54,5 +54,20 @@ describe('Example Test', function() {
       expect(ThankYou.isOpen()).to.be.true
   })
 
+  it('Given I am completing the Communal Establishment questionnaire, When I answer No to the Usual Residents question, Then I should be routed to Completion Preferences', function() {
+
+      // Given I am completing the Communal Establishment questionnaire,
+      openQuestionnaire('census_communal.json')
+
+      EstablishmentType.clickEstablishmentTypeAnswerHotel().submit()
+      BedSpaces.setBedSpacesAnswer(2).submit()
+
+      // When I answer No to the Usual Residents question
+      UsualResidents.clickUsualResidentsAnswerNo().submit()
+
+      // Then I should be routed to Completion Preferences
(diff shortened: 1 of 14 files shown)


tests/functional/spec/census-communal.spec.js
```diff
@@ -54,5 +54,27 @@ describe('Example Test', function() {
...
+      openQuestionnaire('census_communal.json')
+
+      EstablishmentType.clickEstablishmentTypeAnswerHotel().submit()
+      BedSpaces.setBedSpacesAnswer(2).submit()
+
+      // When I answer No to the Usual Residents question
+      UsualResidents.clickUsualResidentsAnswerNo().submit()
+
+      // Then I should be routed to Completion Preferences
+      CompletionPreferenceEstablishment.clickCompletionPreferenceEstablishmentAnswerPaper().submit()
```
jonnyshaw89: @iwootten raised a good point yesterday, in that these tests don't need to reach the end of the survey. you only need to test that you get to the CompletionPreferenceEstablishment page. this would save us lots of build time.

jonnyshaw89 (changes requested):

tests/functional/spec/census-communal.spec.js
benjefferies: When PR #660 is merged the selenium pages have a isOpen function which we can used to solve this. Some of the other tests have already been updated to do this

benjefferies (approved):

jonnyshaw89 (approved):

jonnyshaw89:
👍 LGTM

PR #716, opened 2016-12-13
Run selenium suites in parallel

jonnyshaw89:
Allow the selenium suites to run in parallel so we are able to run all of the tests with no added build time.
Also removing travis workaround as fix is now in.
https://docs.travis-ci.com/user/build-environment-updates/2016-12-06/#2016-12-06

### DIFF
--- a/tests/functional/config.js
+++ b/tests/functional/config.js
@@ -10,7 +10,7 @@ let config = {
   coloredLogs: true,
   screenshotPath: paths.test.errorShots,
   baseUrl: process.env.BASEURL,
-  waitforTimeout: 500,
+  waitforTimeout: 2000,
   updateJob: true,
   specs: [`${paths.test.wdioSpec}/**/*.spec.js`],
   suites: {
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,50 +1,48 @@
 language: python
 sudo: required
 dist: trusty
-# Workaround for https://github.com/travis-ci/travis-ci/issues/6928#issuecomment-264227708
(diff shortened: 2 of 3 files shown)


tests/functional/config.js
```diff
@@ -7,6 +7,7 @@ let config = {
   services: ['selenium-standalone'],
   // Level of logging verbosity: silent | verbose | command | data | result | error
   logLevel: 'error',
+  bail: 1,
```
benjefferies: Not sure if we should do this, I find it quite useful to know the full state of the build on travis

tests/functional/config.js
```diff
@@ -46,7 +47,7 @@ if (process.env.TRAVIS === 'true') {
   config = {
     ...config,
     logLevel: 'silent',
-    capabilities: [chrome]
+    capabilities: [firefox]
```
benjefferies: Why firefox?

jonnyshaw89: note the label "not-ready-to-merge"
don't waste your time looking at this yet.

tests/functional/config.js
jonnyshaw89: doesn't work anyway ;)

collisdigital:
Anyway to split tests into 4 (5 mins each?)

jonnyshaw89:
@collisdigital the more we split this out the more we will hit the concurrent build limit.

benjefferies (approved):

iwootten (approved):
👍

PR #730, opened 2016-12-14
EQ 807: Modify sexual id routing to use both metadata and answer response

### LINKED ISSUE #715: Census_Household_Sexual_ID(true): Is not displaying the Sexual_ID question in the survey
### Expected behaviour
Must ask respondent the following question "Which of the following options best describes how you think of " while filling the questionnaire

### Actual behaviour
Not displaying 'Which of the following options best describes how you think of ' question for the respondent

Launch Census Household schema
Navigate to ethnic group and answers the ethnic questions and click [Save and continue]

### Technical information

Claim details passed on from RH are as followes
{:collection_exercise_sid=>"0", :eq_id=>"census", :exp=>1481632766, :form_type=>"household", :iat=>1481629166, :language=>"en", :period_id=>"", :period_str=>"", :ref_p_start_date=>"2000-01-01", :ref_p_end_date=>"2000-01-01", :region_code=>"GB-ENG", :ru_name=>"", :ru_ref=>"1000000000002215", :return_by=>"2000-01-01", :tx_id=>"4962db4b-4392-4149-a56d-1be69acea202", :user_id=>"", :variant_flags=>{:sexual_identity=>true}}

iwootten:
This change updates the routing around sexual identity question to be informed by both the metadata flag coming from respondent home and the answer to the over 16 question.

As part of this it has been necessary to update all when rules to supply conditions as lists and all behaviour that uses them to evaluate every one of the conditions supplied.

Additionally, we were incorrectly converting the sexual identity flag supplied by respondent home to a string value, so I have had to modify the routing rules/evaluation to work against a boolean value. This could be better in that we should really not convert the type at all and modify all existing rules in schemas to use an appropriate typed value. I felt this would muddy this pull request and so we should raise it as technical debt to resolve later.

I've also got rid of the SchemaCondition/When classes and just used the rule evaluation from dicts throughout.

It's also been necessary to simplify _build_path slightly as our tests started failing due to the computational complexity of it being too high!

Fixes #715

### How to review 
Check that there is appropriate test coverage and that all tests pass.


app/questionnaire/rules.py
```diff
@@ -30,20 +30,25 @@ def evaluate_goto(goto_rule, metadata, answer_store, group_instance):
...
     if 'when' in goto_rule.keys():
-        when = goto_rule['when']
+        all_conditions_met = True
 
-        if 'id' in when:
-            answer_index = when['id']
-            filtered = answer_store.filter(answer_id=answer_index, group_instance=group_instance)
-            if len(filtered) == 1:
-                return evaluate_rule(when, filtered[0]['value'])
+        for when_rule in goto_rule['when']:
```
jonnyshaw89, line 33: curious of we can use this function to evaluate skip conditions? as the logic seems pretty similar? maybe thats not for this PR thou?

app/templating/summary/question.py
```diff
@@ -15,8 +15,12 @@ def __init__(self, question_schema, answers):
 
     def is_skipped(self, all_answers):
```
jonnyshaw89, line 16: are we evaluating skip_conditions in 2 places?

tests/functional/pages/surveys/census/household/sexual-identity.page.js
```diff
@@ -24,6 +24,10 @@ class SexualIdentityPage extends MultipleChoiceWithOtherPage {
     return this
   }
 
+  isOpen() {
```
jonnyshaw89: This logic should now exist in QuestionPage on master?

app/schema/when.py
```diff
@@ -1,8 +0,0 @@
-
```
jonnyshaw89, line 1: Good

jonnyshaw89 (changes requested):

tests/functional/pages/surveys/census/household/sexual-identity.page.js
iwootten: You mean 7 minutes after I opened this PR?

app/questionnaire/navigator.py
```diff
@@ -77,29 +76,24 @@ def build_path(self, blocks, this_location):
...
-                        is_meta_rule = SchemaHelper.is_goto_meta_rule(rule)
+                for rule in [r for r in block['routing_rules'] if SchemaHelper.is_goto_rule(r)]:
+                    should_goto = evaluate_goto(rule['goto'], self.metadata, self.answer_store, this_location['group_instance'])
+
+                    if should_goto:
                         next_location = this_location.copy()
                         next_location['block_id'] = rule['goto']['id']
 
                         next_block_index = self._block_index_for_location(blocks, next_location)
+                        next_precedes_current = next_block_index is not None and block_index > next_block_index
```
benjefferies: Just the extraction of this variable makes it so clear

app/schema/when.py
iwootten: Thanks pal

app/questionnaire/rules.py
iwootten: I did try this, but the group_id/instance arguments mean we'd break it into two submethods to evaluate.

app/templating/summary/question.py
iwootten: Looks like it :)

app/questionnaire/rules.py
jonnyshaw89: Yeh i guess. thats fine then

tests/functional/pages/surveys/census/household/sexual-identity.page.js
jonnyshaw89: Was just giving heads up. ;)

app/schema/when.py
jonnyshaw89: Your very welcome.

jonnyshaw89:
@iwootten looks like your missing some lines of coverage

tests/functional/spec/census-household.spec.js
```diff
@@ -82,7 +82,7 @@ const expect = chai.expect
 
 describe('Census Household', function () {
 
-    it('Given Respondent Home has identified the respondent should have the Household Questionnaire without the sexual id question, When I complete the EQ, Then I should not be asked the sexual id question', function () {
+    it('Given Respondent Home has not sent the sexual identity flag, When I complete the EQ, Then I should not be asked the sexual id question', function () {
```
benjefferies: not set*

tests/functional/spec/census-household.spec.js
```diff
@@ -96,18 +96,57 @@ describe('Census Household', function () {
...
+        InEducation.clickInEducationAnswerNo().submit()
+        CountryOfBirth.clickCountryOfBirthEnglandAnswerEngland().submit()
+        Carer.clickCarerAnswerNo().submit()
+        NationalIdentity.clickNationalIdentityEnglandAnswerEnglish().submit()
+        EthnicGroup.clickEthnicGroupEnglandAnswerWhite().submit()
+        WhiteEthnicGroup.clickWhiteEthnicGroupEnglandAnswerEnglishWelshScottishNorthernIrishBritish().submit()
 
-        // visitors
-        completeVisitorSection()
+        expect(SexualIdentity.isOpen()).to.be.false
```
benjefferies: For these it's worth adding a message otherwise it's not very clear in the logs when it fails like https://github.com/ONSdigital/eq-survey-runner/pull/660/files#diff-d1d932da779856122e7ba86f29c1e5c6R67

benjefferies (changes requested):
Looking good!

app/data/test_metadata_routing.json
```diff
@@ -43,11 +43,13 @@
...
-                                "when": {
-                                    "meta" : "variant_flags.flag_1",
-                                    "condition": "equals",
-                                    "value":"true"
-                                }
+                                "when": [
+                                    {
+                                        "meta" : "variant_flags.flag_1",
+                                        "condition": "equals",
+                                        "value":"true"
```
ajmaddaford: Should this be `true` rather than `"true"`?

jonnyshaw89 (approved):

tests/app/templating/summary/test_question.py
```diff
@@ -243,7 +243,7 @@ def test_question_should_be_skipped(self):
         # Given
         answers = {'answer_1': 'skip me'}
         answer_schema = {'id': 'answer_1', 'title': '', 'type': '', 'label': ''}
-        skip_condition = {'when': {'id': 'answer_1', 'condition': 'equals', 'value': 'skip me'}}
+        skip_condition = {'when': [{'id': 'answer_1', 'condition': 'equals', 'value': 'skip me'}]}
```
ajmaddaford, line 246: Minor point, and possibly one for discussion rather than changing it in this pull request, but I think the earlier tests (e.g. test_conditional_display_skips_when_not_equals) that split the test data dicts over multiple lines are far easier to read. Maybe a +1 or -1 on this comment from people to gauge opinion 🙂

app/questionnaire/navigator.py
ajmaddaford: I agree, but I think that it would read better using `next_block_index < block_index`

ajmaddaford (changes requested):

tests/app/templating/summary/test_question.py
iwootten: Fair enough, I'm leaving this for now - the whole set is like this and I'm not about to upset the coverage gods any more.

tests/functional/spec/census-household.spec.js
iwootten: Well, I think sent makes equal sense here, but whatever.

iwootten:
@ajmaddaford @jonnyshaw89 @benjefferies Have sorted all these points. Ended up rebasing in order to kick Travis into action as it wasn't picking up commits and triggering builds. Sorry if that makes it confusing to see resolutions.

app/data/census_household.json
```diff
@@ -2182,121 +2248,150 @@
...
+                                "when": [
+                                    {
+                                        "id": "over-16-answer",
+                                        "condition": "equals",
+                                        "value": "Yes"
+                                    },
+                                    {
+                                        "meta" : "variant_flags.sexual_identity",
+                                        "condition": "equals",
+                                        "value": true
```
collisdigital, line 2418: If this change is happening it might be worth discussing if having a condition like "is true" and omitting the "value" statement would be better, a bit like the various assert calls in unit tests (assertTrue, assertEqual etc.) just a thought.

app/helpers/schema_helper.py
```diff
@@ -64,4 +64,10 @@ def is_goto_rule(rule):
 
     @staticmethod
     def is_goto_meta_rule(rule):
```
collisdigital: This function name doesn't really match what it's doing now, more like `is_goto_rule_containing_meta_condition`, which probably raises the question of if the calling code needs to behave differently if the rule can contains a meta and answer condition (maybe not?)

app/parser/v0_0_1/schema_parser.py
```diff
@@ -279,21 +275,6 @@ def _parse_question(self, schema, questionnaire):
 
         return question
 
-    @staticmethod
-    def _parse_skip_condition(skip_condition_schema):
```
collisdigital, line 283: :+1:

app/questionnaire/navigator.py
```diff
@@ -59,10 +59,9 @@ def build_path(self, blocks, this_location):
         """
         path = []
         block_index = 0
-        blocks_len = len(blocks)
 
         # Keep going unless we've hit the last block
-        while block_index < blocks_len:
+        while block_index < len(blocks):
```
collisdigital: caching the blocks len was a bit of optimisation; although practically it may be insignificant I guess and makes the code harder to read? I put together a noddy example:

```python
import time

def test1(blocks):
    count = 0
    while count < len(blocks):
      count += 1

def test2(blocks):
    count = 0
    blocks_len = len(blocks)
    while count < blocks_len:
      count += 1

data = [1]*100

start = time.time()
for x in range(100000):
  test1(data)
end = time.time()
print("Time taken len(blocks): {}".format(end-start))

start = time.time()
for x in range(100000):
  test2(data)
end = time.time()
print("Time taken blocks_len: {}".format(end-start))

```

Output:

```
Time taken len(blocks): 0.973261833190918
Time taken blocks_len: 0.560056209564209
```

app/questionnaire/navigator.py
```diff
@@ -77,29 +76,24 @@ def build_path(self, blocks, this_location):
             # If routing rules exist then a rule must match (i.e. default goto)
             if 'routing_rules' in block and len(block['routing_rules']) > 0:
                 original_this_location = this_location.copy()
-                for rule in block['routing_rules']:
-                    if SchemaHelper.is_goto_rule(rule) and evaluate_goto(rule['goto'], self.metadata, self.answer_store, this_location['group_instance']):
-                        is_meta_rule = SchemaHelper.is_goto_meta_rule(rule)
+                for rule in [r for r in block['routing_rules'] if SchemaHelper.is_goto_rule(r)]:
```
collisdigital: Consider using `filter` for this kind of thing; maybe just opinion but I find it easier to read as it removes the temporary variables `r` e.g.

```python
for rule in filter(SchemaHelper.is_goto_rule, block['routing_rules']):
```

app/questionnaire/navigator.py
```diff
@@ -77,29 +76,24 @@ def build_path(self, blocks, this_location):
...
+                            self._remove_rule_answers(rule, this_location)
 
                         this_location = next_location
                         break
 
                 # If we haven't changed location based on routing rules then we can't progress any further
-                if (this_location['block_id'] == original_this_location['block_id'] and
-                        this_location['group_id'] == original_this_location['group_id'] and
-                        this_location['group_instance'] == original_this_location['group_instance']):
+                if this_location == original_this_location:
```
collisdigital, line 88: :+1: I think Andrew and I were having a brain freeze when we wrote that!

app/questionnaire/navigator.py
```diff
@@ -116,6 +110,17 @@ def build_path(self, blocks, this_location):
 
         return path
 
+    def _remove_rule_answers(self, rule, location):
+        # We're jumping backwards, so need to delete all answers from which
+        # route is derived
+        if not SchemaHelper.is_goto_meta_rule(rule):
```
collisdigital: Looking at this (and my previous comment about the name/behaviour of it changing) I think this could be a problem; if the rule contains both a meta and answer conditions then it will skip removing the answers that weren't meta conditions.; I think it should always loop of all conditions and do the check for a meta rule against each condition.

app/questionnaire/navigator.py
```diff
@@ -116,6 +110,17 @@ def build_path(self, blocks, this_location):
...
+            for condition in rule['goto']['when']:
+                self.answer_store.remove(answer_id=condition['id'],
+                                         answer_instance=0,
```
collisdigital: I assume we don't support routing based on anything other than the first answer_instance right now? If that changes in future this could fail... not sure what you can do now about that though, just highlighting it.

app/questionnaire/rules.py
```diff
@@ -30,20 +30,25 @@ def evaluate_goto(goto_rule, metadata, answer_store, group_instance):
...
+            condition_met = False
 
-        elif 'meta' in when:
-            key = when['meta']
-            value = get_metadata_value(metadata, key)
-            return evaluate_rule(when, value)
+            if 'id' in when_rule:
+                answer_index = when_rule['id']
+                filtered = answer_store.filter(answer_id=answer_index, group_instance=group_instance)
+                if len(filtered) == 1:
```
collisdigital: Should this raise an exception or at least log an error as an `else` to this; not sure what is expected if more than one answer comes back (can that happen for multiple answer_instances?) but currently it will silently return False which could be a nasty one to debug. Another solution could be to get answer_store.filter to take an `expected_match_count=1` and have it do the log/exception so anyone calling filter doesn't need to deal with it.

app/questionnaire/rules.py
```diff
@@ -30,20 +30,25 @@ def evaluate_goto(goto_rule, metadata, answer_store, group_instance):
...
+                    condition_met = evaluate_rule(when_rule, filtered[0]['value'])
 
-        return False
+            elif 'meta' in when_rule:
+                key = when_rule['meta']
+                value = get_metadata_value(metadata, key)
+                condition_met = evaluate_rule(when_rule, value)
+
+            all_conditions_met &= condition_met
```
collisdigital: Since the `when` statements are AND conditions, there's no need to do this, you can just return False (or break the loop) as soon as any condition doesn't match i.e. there's no need to go through every `when`.

app/questionnaire_state/state_item.py
```diff
@@ -40,6 +44,23 @@ def get_errors(self):
...
+        # Process any conditional display rules
+        if self.schema_item:
+
+            if hasattr(self.schema_item, 'skip_condition') and self.schema_item.skip_condition:
+                rule = self.schema_item.skip_condition
+                all_conditions_met = True
+
+                for when in rule['when']:
+                    value = answers.get(when['id']) if 'id' in when else get_metadata_value(metadata, when['meta'])
+                    all_conditions_met &= evaluate_rule(when, value)
```
collisdigital: Similar to above; no need to go through every `when` and OR these together; just break on the first False.

app/templating/summary/question.py
```diff
@@ -15,8 +15,12 @@ def __init__(self, question_schema, answers):
...
         if self.skip_condition is not None:
-            answer = all_answers.get(self.skip_condition['when']['id'])
-            return evaluate_rule(self.skip_condition['when'], answer)
+            all_conditions_met = True
+
+            for when_rule in self.skip_condition['when']:
+                answer = all_answers.get(when_rule['id'])
+                all_conditions_met &= evaluate_rule(when_rule, answer)
```
collisdigital: Same as before...

tests/functional/spec/census-household.spec.js
```diff
@@ -133,32 +171,42 @@ describe('Census Household', function () {
...
+        EthnicGroup.clickEthnicGroupEnglandAnswerWhite().submit()
+        WhiteEthnicGroup.clickWhiteEthnicGroupEnglandAnswerEnglishWelshScottishNorthernIrishBritish().submit()
 
-        // visitors
-        completeVisitorSection()
+        expect(SexualIdentity.isOpen()).to.be.equal(false, "Did not expect sexual identity page to be open")
+    })
 
-        Confirmation.submit()
+    it('Given Respondent Home has set the sexual identity flag and I have chosen over 16, When I complete the EQ, Then I should be asked the sexual id question', function () {
```
collisdigital, line 181: :+1: great example of Trello Card to BDD tests.

app/helpers/schema_helper.py
collisdigital: Update: it could be a problem, see later comment on calling code...

app/questionnaire/navigator.py
collisdigital: If these `@staticmethod`s in SchemaHelper were just top-level methods on the schema_helper.py this could be even easier to read:

```Python
for rule in filter(is_goto_rule, block['routing_rules']):
```

app/questionnaire/rules.py
benjefferies: I think there are a few places in the navigator where we are in an erroneous state either where the schema may have been misconfigured or something we can't cope with has happened. Might be worth looking at the navigator as a whole to see if we should be raising errors in certain states

app/data/census_household.json
jonnyshaw89: then you would need "is true" and "is false". i think making the value optional is going down a bad path. the more caveats that are added with regards to the schema structure the more logic will need to be included in the author and the harder the validation of a schema will become.

collisdigital: KISS 👍

app/questionnaire/navigator.py
iwootten: Eh really? I would think that length was only ever evaluated once. Happy to revert if it does save time..

collisdigital: I guess Python can't determine if blocks is going to change within the loop so needs to re-check each time; maybe another argument to use an iterator in future (we have a tech debt card around this)

app/helpers/schema_helper.py
iwootten: Doesn't goto_meta_rule imply that it's a goto rule containing a meta condition? Don't see the point in making this so wordy.

app/questionnaire/rules.py
iwootten: I don't think this is the responsibility of the navigator. We shouldn't have schemas that contain invalid routing rules, we should have better validation.

app/helpers/schema_helper.py
collisdigital: Except it no longer is, the rule can contain multiple when's so it could be a meta and non-meta at the same time now...

app/questionnaire/rules.py
collisdigital: That's fine as long as when/if we do implementing routing based on multiple answers this gets picked up, it's easy to forget stuff like this and without any output indicating otherwise it can go undetected for a long time until someone finds a bug in production. Possibly a perfect case for an `assert` statement actually.

iwootten: I've updated this to log a warning and added a trello card to the backlog: https://trello.com/c/mXL1a8JD/297-update-navigator-to-handle-multiple-filtered-answers

app/helpers/schema_helper.py
iwootten: I disagree, the rule uses metadata in it's evaluation always, despite how many conditions there are.

app/questionnaire/rules.py
```diff
@@ -38,6 +43,8 @@ def evaluate_goto(goto_rule, metadata, answer_store, group_instance):
                 filtered = answer_store.filter(answer_id=answer_index, group_instance=group_instance)
                 if len(filtered) == 1:
                     condition_met = evaluate_rule(when_rule, filtered[0]['value'])
+                elif len(filtered) > 1:
+                    logger.warning("Condition will not met: Multiple (%d) answers found" % len(filtered))
```
collisdigital: Having thought about it; an assert here (as well) would be more useful for future bugs (asserting the current system behaviour) - https://wiki.python.org/moin/UsingAssertionsEffectively

app/questionnaire/rules.py
```diff
@@ -44,7 +44,7 @@ def evaluate_goto(goto_rule, metadata, answer_store, group_instance):
                 if len(filtered) == 1:
                     condition_met = evaluate_rule(when_rule, filtered[0]['value'])
                 elif len(filtered) > 1:
-                    logger.warning("Condition will not met: Multiple (%d) answers found" % len(filtered))
+                    logger.warning("Condition will not met: Multiple ({:d}) answers found".format(len(filtered)))
```
collisdigital: I think log formats should be comma based

`logger.warning("Condition will not met: Multiple (%d) answers found", len(filtered))`

See https://docs.python.org/3.5/library/logging.html#logging.Logger.debug

app/questionnaire/rules.py
```diff
@@ -46,9 +44,8 @@ def evaluate_goto(goto_rule, metadata, answer_store, group_instance):
                 value = get_metadata_value(metadata, key)
                 condition_met = evaluate_rule(when_rule, value)
 
-            all_conditions_met &= condition_met
-
-        return all_conditions_met
+            if not condition_met:
```
jonnyshaw89: why even have this variable? why not just return as opposed to setting it?

jonnyshaw89 (changes requested):

jonnyshaw89 (approved):

app/questionnaire/rules.py
```diff
@@ -90,16 +90,6 @@ def is_goto_rule(rule):
     return 'goto' in rule and 'when' in rule['goto'].keys() or 'id' in rule['goto'].keys()
 
 
-def is_goto_meta_rule(rule):
```
collisdigital: Well that solved that issue 😁

app/questionnaire/rules.py
```diff
@@ -1,3 +1,8 @@
+import logging
+
+logger = logging.getLogger(__name__)
```
collisdigital: No longer needed?

collisdigital (approved):

benjefferies (approved):

ajmaddaford (approved):

PR #731, opened 2016-12-14
Add pause to allow animation to complete

benjefferies:
Add a 300ms pause to allow the animation of the relationship answer to collapse.

### How to review 
Selenium tests should stop intermittently failing.

### DIFF
--- a/tests/functional/generate_pages.py
+++ b/tests/functional/generate_pages.py
@@ -69,6 +69,13 @@ CHECKBOX_RADIO_CLICKER = r"""  click{optionName}() {
 
 """
 
+RELATIONSHIP_RADIO_CLICKER = r"""  click{optionName}() {
+    browser.element('[id="{optionId}"]').click().pause(300)
+    return this
+  }
+
+"""
+
 REPEATING_ANSWER_SETTER = r"""  set{answerName}(value, index = 0) {
     var field = '{answerId}'
     if (index > 0) {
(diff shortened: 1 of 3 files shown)


jonnyshaw89:
lets run this a few times In travis

hamishtaplin:
@hamishtaplin You're Travis!

tests/functional/generate_pages.py
```diff
@@ -69,6 +69,13 @@
 
 """
 
+RELATIONSHIP_RADIO_CLICKER = r"""  click{optionName}() {
+    browser.element('[id="{optionId}"]').click().pause(300)
```
samiwel, line 73: Rather than an arbitrary pause duration, is there some way we could wait until something in the DOM changes? e.g. a CSS class gets added? e.g. `browser.waitForExist(selector...`

hamishtaplin: There's nothing in the DOM to indicate when a transition has completed, only started. I did consider this but we'd have to manage the lifecycle of all transitions (listening for events, dealing with cross-browser problems that are a result) and then adding classes for just the tests. 

It's adding a lot of code for something that just makes tests a bit cleaner.

benjefferies: We change the element to have an "animating" class or something while it is animated but the 300ms isn't as arbitrary as you think as it's the actual time in the animation in the css https://github.com/ONSdigital/eq-survey-runner/blob/master/app/assets/styles/partials/components/_relationship.scss#L38

samiwel: Fair enough, just if we change the duration of the animation (or change the animation) we'll have to go back and revisit all these.

benjefferies: Your right, it's the risk. I think it's low risk enough and the chances it needs to change are low enough to hold fire as it will require some effort from @hamishtaplin

hamishtaplin: Agreed but the overhead of listening for `transitionend` events isn't worth it—it's not very reliable and requires a library to solve the cross-browser issues. Also, if we removed the transition or changed it to an `animation` (different CSS property) we'd also have to update the JS.

It would also break in older browsers that don't use animations (the class would never appear so the test wouldn't work)

jonnyshaw89 (approved):

samiwel (approved):

jonnyshaw89:
👍 LGTM

PR #735, opened 2016-12-15
Fix repeating group bug for census

### LINKED ISSUE #727: When clicking previous after adding household members, the reporting groups still repeat
### Expected behaviour
Given i add household members then go previous and change my answer to no one lives here.
When i navigate to the repeating household members group
Then there should be no household members to repeat.

### Actual behaviour
It repeats over the household members that i changed my mind on.

benjefferies:

Repeating answers will repeat if the answer which determines the repeat
was originally answered but if the user then changes their mind to go
down a path where the answer which determines the repeat will no longer
appear on. The expected result is for those repeating answers to not be
repeated. The actual result is that the repeating answers are repeated.

The scenario this occurs under is:
Given I have added a person in my household, When I change my mind and
they should be visitors, Then I do not have to answer household member
details

The short term fix for this is to remove the answer which determines the
repeat when the user goes down the other path which is consistent with
what we do when a user changes their answer which determines the number
of repeats.

The ideal fix going forward would be to check if the answer which
determines the repeating answers is on the path that the user is being
followed.

[trello](https://trello.com/c/vnjf9dR9/813-eq-survey-runner-727-when-clicking-previous-after-adding-household-members-the-reporting-groups-still-repeat-issue)
[trello for the real fix](https://trello.com/c/QXkI2J6i/298-repeating-groups-should-only-be-evaluated-if-the-answer-which-determines-the-repeating-groups-in-on-the-path)

### How to review 
All selenium tests should still pass and also the BDD test `Given I have added a person in my household, When I change my mind they should be visitors, Then I do not have to answer household details`

Fixes #727

### DIFF
--- a/app/main/views/questionnaire.py
+++ b/app/main/views/questionnaire.py
@@ -247,14 +247,7 @@ def post_household_composition(eq_id, form_type, collection_id, group_id):
     }
 
     if 'action[save_continue]' in request.form:
-        answer_store.remove(group_id=group_id, block_id='household-composition')
-
-        questionnaire_store = get_questionnaire_store(current_user.user_id, current_user.user_ik)
-        for answer in SchemaHelper.get_answers_that_repeat_in_block(g.schema_json, 'household-composition'):
-            groups_to_delete = SchemaHelper.get_groups_that_repeat_with_answer_id(g.schema_json, answer['id'])
-            for group in groups_to_delete:
-                answer_store.remove(group_id=group['id'])
-                questionnaire_store.completed_blocks[:] = [b for b in questionnaire_store.completed_blocks if b.get('group_id') != group['id']]
+        _remove_repeating_on_household_answers(answer_store, group_id)
 
     valid = questionnaire_manager.process_incoming_answers(this_block, request.form)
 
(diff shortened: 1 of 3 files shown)


app/main/views/questionnaire.py
```diff
@@ -276,6 +269,24 @@ def post_household_composition(eq_id, form_type, collection_id, group_id):
     return redirect(location_url(eq_id, form_type, collection_id, next_location))
 
 
+@questionnaire_blueprint.route('<group_id>/<int:group_instance>/permanent-or-family-home', methods=["POST"])
+@login_required
+def post_everyone_at_address_confirmation(eq_id, form_type, collection_id, group_id, group_instance):
+    _remove_repeating_on_household_answers(get_answer_store(current_user), group_id)
```
jonnyshaw89: Do we only want to do this if the answer is no?
If i navigate backwards and then forwards without changing the answer then the values are deleted.

jonnyshaw89 (changes requested):

app/main/views/questionnaire.py
benjefferies: Good point

app/main/views/questionnaire.py
```diff
@@ -272,8 +272,19 @@ def post_household_composition(eq_id, form_type, collection_id, group_id):
...
+    next_location = navigator.get_next_location(current_group_id=group_id, current_iteration=group_instance,
+                                                current_block_id='permanent-or-family-home')
+    if next_location['block_id'] != 'household-composition':
+        _remove_repeating_on_household_answers(get_answer_store(current_user), group_id)
+        this_block = {
+            'group_id': group_id,
+            'group_instance': group_instance,
+            'block_id': 'permanent-or-family-home',
+        }
+        get_questionnaire_manager(g.schema, g.schema_json).update_questionnaire_store(this_block)
```
jonnyshaw89: why this?

jonnyshaw89 (approved):

jonnyshaw89:
@benjefferies the flaw with selenium tests is that they don't count towards coverage

samiwel:
But integration tests do ;)

ajmaddaford (approved):

benjefferies:
Ready for merge once builds pass

PR #738, opened 2016-12-15
Add navigation capability

ajmaddaford:
Adds navigation capability into survey runner, tested and working for Census use case. Each group appears in the navigation unless it has a repeating rule which hasn't been completed then it is hidden. If a group is repeating and the feeding question has been answered it then repeats the required amount. Navigation when clicking 'Save and Continue' is always linear, unless you reach the end, and it then goes to the last completed block. The previous link is not present on the first block of each group.

### How to review 
- Run survey runner and load the Census household survey
- Use the navigation and verify it works as expected
- When adding people in the household composition, they should appear in the navigation
- When answering > 0 visitors, the 'Visitors' link should appear in the navigation
- When a navigation section is completed, a tick should appear next to the navigation link

Work was completed by @LJBabbage in #628 but the original branch was in rebase hell.


jonnyshaw89 (changes requested):
Only Joking

jonnyshaw89 (approved):

iwootten (approved):
Hurry up!

benjefferies (approved):

collisdigital (approved):
Go go go

Issue #699, opened 2016-12-12
Answers are not bleached before being stored in answer_store


jonnyshaw89:
### Expected behaviour
When i post malicious content to the application then this should be sanitised before being stored and displayed on subsequent pages.

### Actual behaviour
The malicious content is being stored in the answer-store and displayed on subsequent pages

### Steps to reproduce the behaviour
On the census_household.json In the persons name enter `<img src="https://www.google.co.uk/images/branding/googlelogo/2x/googlelogo_color_120x44dp.png"/>`

samiwel:
@jonnyshaw89 Just checked and it appears that at least some answers are being bleached before storing in the answer store. This happens during validation in schema. The `bleach.clean` call is what causes everything after the '<' when entering a household name to be lost (see #695). Not sure if there's a way of telling the clean method not so strip content, but rather encode the value?

samiwel:
Looks promising... https://docs.python.org/3/library/html.html#html.escape The second argument allows escaping of quotation marks.

collisdigital:
Closing this as we bleach values on input from widgets.

PR #742, opened 2016-12-15
Bleach values entering AnswerStore

jonnyshaw89:
Move the bleaching of input to when the value is entered into the AnswerStore. This should then catch all input.

Fixes https://github.com/ONSdigital/eq-survey-runner/issues/699

### DIFF
--- a/app/data_model/answer_store.py
+++ b/app/data_model/answer_store.py
@@ -71,8 +71,8 @@ class AnswerStore(object):
 
         if isinstance(answer.value, list):
             bleached_values = []
-            for index, answer_from_list in enumerate(answer.value):
-                bleached_values.append(self._bleach(answer.value[index]))
+            for answer_from_list in answer.value:
+                bleached_values.append(self._bleach(answer_from_list))
 
             return bleached_values
         else:
--- a/app/data_model/answer_store.py
+++ b/app/data_model/answer_store.py
@@ -2,6 +2,8 @@ import re
 
 from collections import OrderedDict
 
+import bleach
+
 
 class Answer(object):
     def __init__(self, group_id=None, block_id=None, answer_id=None, value=None, group_instance=0, answer_instance=0):
@@ -63,6 +65,26 @@ class AnswerStore(object):
         if not isinstance(answer, Answer):
             raise TypeError("Method only supports Answer argument type")
 
+    def _sanitise(self, answer):
+        if answer.value is None:
+            return answer.value
+
+        if isinstance(answer.value, list):
+            bleached_values = []
+            for index, answer_from_list in enumerate(answer.value):
+                bleached_values.append(self._bleach(answer.value[index]))
+
+            return bleached_values
+        else:
+            return self._bleach(answer.value)
+
+    @staticmethod
+    def _bleach(value):
+        if isinstance(value, int):
+            return value
+
+        return bleach.clean(value)
+
     def add(self, answer):
         """
         Add a new answer into the answer store.
@@ -70,6 +92,7 @@ class AnswerStore(object):
         :param answer: A dict of flattened answer details.
         """
         self._validate(answer)
+        answer.value = self._sanitise(answer)
 
         answer_to_add = answer.__dict__.copy()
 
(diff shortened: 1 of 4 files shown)


tests/app/data_model/test_answer_store.py
```diff
@@ -624,3 +624,18 @@ def test_remove_all_answers(self):
...
+            answer_instance=1,
+            group_id="5",
+            group_instance=1,
+            value='<img src="https://www.google.co.uk/images/branding/googlelogo/2x/googlelogo_color_120x44dp.png"/>',
+        )
+
+        self.store.add(answer_1)
+
+        self.assertNotEqual(self.store.get(answer_1), '<img src="https://www.google.co.uk/images/branding/googlelogo/2x/googlelogo_color_120x44dp.png"/>')
+        self.assertFalse('<' in self.store.get(answer_1))
```
benjefferies: self.assertNotIn would give a better failure message

app/data_model/answer_store.py
```diff
@@ -63,13 +65,34 @@ def _validate(answer):
         if not isinstance(answer, Answer):
             raise TypeError("Method only supports Answer argument type")
 
+    def _sanitise(self, answer):
+        if answer.value is None:
+            return answer.value
+
+        if isinstance(answer.value, list):
+            bleached_values = []
+            for index, answer_from_list in enumerate(answer.value):
```
collisdigital: why enumerate? can't you just `answer_from_list in answer.value` ?

tests/app/data_model/test_answer_store.py
```diff
@@ -624,3 +624,18 @@ def test_remove_all_answers(self):
 
         self.store.remove()
         self.assertEqual(self.store.map(), {})
+
+    def test_answer_store_bleaches_answers(self):
```
benjefferies, line 628: We should probably have a unit test for bleaching list and ints

samiwel:
I've just noticed that when entering `<person name>` into the household composition entry screen that when the answer gets piped into the group title it displays the encoded values `&lt;person name&gt;`. Do you think we should be marking group titles as `safe` or should the unescaping happen in the household member name filter?

The behaviour for entering `person<name` is the same as before, but maybe that is correct/expected and bleach is just doing its thing.

app/data_model/answer_store.py
jonnyshaw89: Ah yes i can change to this now. I had a different approach originally.

benjefferies (changes requested):

app/schema/answer.py
```diff
@@ -58,14 +56,12 @@ def check_user_input(user_input):
 
     def get_typed_value(self, raw_input):
 
-        user_input = bleach.clean(raw_input)
```
iwootten, line 61: Isn't the fix here not to remove the current bleaching, but to validate all input before storing it? Moving the bleaching to retrieval seems like the wrong way to go, couldn't we have executed something malicious on storing it?

iwootten (changes requested):

app/schema/answer.py
benjefferies: Good point it does make sense to sanitise on entering the survey runner

collisdigital: Really it should be bleached immediately after being received from the POST before anything else happens; you don't know what effect dirty data can have on any of the system, even validation code etc.

samiwel: I think the values are being bleached when they get validated on POST. That's the reason we observe text being cut off in household composition member names if you leave a half open tag in there e.g. 'Pers<on Name' only 'Pers' gets stored.

collisdigital: In reality we should probably show an error to the user if they put a `<` in their name (like non-numeric input in numeric fields). I'm thinking WTForms stuff (regex acceptance list for text input?)

ajmaddaford: That might be ok for people's names, but if we ever have inputs for something like a company name blocking '<' or '>' probably isn't what we want to do e.g. MORE TH>N.

benjefferies: It's interesting that bleach actually removes parts of the answer instead of encoding

jonnyshaw89:
We probably need to do a more in depth review of how we handle data within the application. what is allowed ini and what isn't.

PR #744, opened 2016-12-16
Split routing tests

benjefferies:
When a test fails it takes a long time to run all of the routing rules
to see what has broken. It is also more difficult to maintain. It will
only get larger and more unmaintainable if we don't split them out

### How to review 
Check that no selenium tests have gone missing in the refactor
All selenium tests should still pass


tests/functional/spec/census/routing/private-response.spec.js
```diff
@@ -0,0 +1,181 @@
...
+        completeVisitorSection()
+        VisitorsCompleted.submit()
+
+        Confirmation.submit()
+
+        // Thank You
+        expect(ThankYou.isOpen()).to.be.true
+    })
+
+    function completeHouseholdAndAccommodation() {
```
benjefferies: I will refactor these to be importable modules

scripts/run_tests.sh
```diff
@@ -32,21 +32,25 @@ function display_result {
...
 
-py.test --cov=app --cov-report xml $@ $1
-display_result $? 2 "Unit tests"
+if [ -z "$EQ_FUNCTIONAL_TEST_SUITES" ]; then
+    export EQ_FUNCTIONAL_TEST_SUITES="core"
+fi
 
-# Run front end tests
-npm config set python python2.7
+if [ "$EQ_FUNCTIONAL_TEST_SUITES" = "core" ]; then
```
jonnyshaw89: If a unit test fails then the other builds will run all of the functional tests? surely this breaks the fail fast rule?

jonnyshaw89 (changes requested):

scripts/run_tests.sh
benjefferies: I guess you still get the fail fast indication from the other build, the disadvantage is that you don't free up your travis resource I guess. I am just playing at the moment to see how much time it saves us to see if the cost is worth the benefit

benjefferies: The results are:


Running all tests on both matrix:
https://travis-ci.org/ONSdigital/eq-survey-runner/builds/184486318 - 14 min 37 sec
https://travis-ci.org/ONSdigital/eq-survey-runner/builds/184489761 - 16 min 17 sec

Only running all tests on core:
https://travis-ci.org/ONSdigital/eq-survey-runner/builds/184491190 - 11 min 17 sec

benjefferies: We could fail the census branch if the core branch fails to free up resources, that said we can't do that until travis get their act together - https://github.com/travis-ci/travis-ci/issues/2062

jonnyshaw89: i can't see much difference in the build times.

benjefferies: I've updated my comment above to include the times

ajmaddaford:
I think splitting out the routing tests and changing the behaviour of run_tests.sh should be two different pull requests. Splitting the routing tests could go in now, whereas changing the behaviour of the script will involve more discussion.

benjefferies:
Removed run_tests change. Note that the build time will increase as a consequence of these changes

jonnyshaw89 (approved):

tests/functional/spec/census/routing/who-lives-here.spec.js
```diff
@@ -0,0 +1,271 @@
...
+    OvernightVisitors.setOvernightVisitorsAnswer(1).submit()
+    HouseholdRelationships.clickHouseholdRelationshipsAnswerHusbandOrWife().submit()
+    WhoLivesHereCompleted.submit()
+
+    // household-and-accommodation
+    completeHouseholdAndAccommodation()
+
+    // household-member
+    completeHouseholdDetails()
+    completeHouseholdDetails()
```
iwootten: Is this meant to be filled in twice?

benjefferies: It's other one of those implicit checks I can change it while I'm the area

tests/functional/spec/census/routing/carer.spec.js
```diff
@@ -0,0 +1,176 @@
...
+    startCensusQuestionnaire('census_household.json', true)
+    PermanentOrFamilyHome.clickPermanentOrFamilyHomeAnswerYes().submit()
+    HouseholdComposition.setFirstName('John').addPerson().setFirstName('Jane', 1).submit()
+    EveryoneAtAddressConfirmation.clickEveryoneAtAddressConfirmationAnswerYes().submit()
+    OvernightVisitors.setOvernightVisitorsAnswer(0).submit()
+    HouseholdRelationships.clickHouseholdRelationshipsAnswerHusbandOrWife().submit()
+    WhoLivesHereCompleted.submit()
+
+    // household-and-accommodation
+    TypeOfAccommodation.clickTypeOfAccommodationAnswerWholeHouseOrBungalow().submit()
```
benjefferies, line 82: I think it will be useful to use the helper methods in the complete-sections.js. I won't do it in this PR as it will muddy the water

ajmaddaford (approved):

PR #753, opened 2016-12-16
Adding skip_conditions for groups

### LINKED ISSUE #740: None or zero Visitors answer still shows Visitor section complete interstitial
### Expected behaviour
- Visitors complete interstitial only shown when there are more than 0 visitors.

### Actual behaviour
- Visitors complete interstitial is shown even when there are 0 visitors or skipping forward and not answering the number of visitors question.

- Start household questionnaire
- Skip the Household and Accommodation via navigation
- Complete Household and Accommodation section
- Observe that the Visitors complete interstitial is shown incorrectly.

#### Browser
Chrome

#### Operating System
macOS

### Screenshot
N/A

benjefferies:
We need to be able to skip the visitors completed interstitial page in
the census household questionnaire. It lives in a separate group to the
visitor questions therefore it won't be skipped by the repeat rule.

This change implements skip_condition on a group to solve that.

Fixes #740 and Fixes
https://trello.com/c/wGkp3T52/291-remove-need-to-update-state-in-navigator


app/data/test_skip_condition_group.json
```diff
@@ -0,0 +1,118 @@
...
+                            ]
+                        }
+                    ]
+                }
+            ]
+        },
+        {
+            "id": "should-skip-group",
+            "title": "This question may or may not be skipped",
+            "skip_condition": {
```
jonnyshaw89, line 64: do you need to add this to schema-v1.json?

app/questionnaire/navigator.py
```diff
@@ -199,8 +191,12 @@ def get_blocks(self):
         blocks = []
 
         for group_index, group in enumerate(SchemaHelper.get_groups(self.survey_json)):
-            no_of_repeats = 1
+            if 'skip_condition' in group:
```
jonnyshaw89, line 194: if we are putting this on groups maybe we want to support blocks too? would have made some of the census routing rule a bit simpler?

jonnyshaw89 (changes requested):

app/data/test_skip_condition_group.json
benjefferies: Yep, still a bit WIP but wanted to get some views on how we are implementing this. On the whole it works but it needs tidy ups like your suggestion

app/questionnaire/navigator.py
benjefferies: Might make sense one day but let's do it in a feature driven manor then just add it now. Looks at the S3 bucket stuff, isn't used.

app/data/test_skip_condition_group.json
```diff
@@ -0,0 +1,118 @@
...
+    "questionnaire_id": "0",
+    "schema_version": "0.0.1",
+    "data_version": "0.0.1",
+    "survey_id": "0",
+    "title": "Skip group",
+    "theme": "default",
+    "description": "A questionnaire to skipping a group.",
+    "eq_id": "0",
+    "messages": {
+      "INTEGER_TOO_LARGE": "Number is too large",
```
collisdigital: Do we actually support these at the top-level like this, would have saved copy/pasting them against each answer in the Census schema!

app/questionnaire/rules.py
```diff
@@ -69,6 +69,20 @@ def evaluate_repeat(repeat_rule, answer_store):
         return repeat_function(filtered)
 
 
+def evaluate_skip_condition(skip_condition, answer_store):
```
collisdigital: Pretty sure there is already at least one if not more of these identical blocks of code about parsing/evaluating when rules, can it be consolidated?

benjefferies: There are now 3, it should be possible to consolidate. I started doing it initially and it was a bit of a can of worms so planned on coming back to it later so thanks for the reminder 👍

app/data/test_skip_condition_group.json
benjefferies: Will take a look to see if we can remove, we need to spend some time making sure the schema is correct and introduce validation so we wouldn't have to cross reference against the code

benjefferies: Done

app/data/test_skip_condition_group.json
benjefferies: Done

app/questionnaire/rules.py
iwootten: I think this method name should be more generic. Perhaps something like evaluate_when_list, given it doesn't just apply to skip_conditions, it's equally applicable to the when attribute of any rule.

benjefferies: Done

app/data/schema/schema-v1.json
```diff
@@ -176,12 +188,16 @@
                               }
                             },
                             "skip_condition": {
```
jonnyshaw89, line 190: is this correct? arn't skip conditions on sections?

tests/functional/pages/surveys/skip_condition_group/do-you-want-to-skip.page.js
```diff
@@ -0,0 +1,23 @@
+// >>> WARNING THIS PAGE WAS AUTO-GENERATED ON 2016-12-19 12:17:40.003901 - DO NOT EDIT!!! <<<
```
jonnyshaw89: these still have timestamps, you can regenerate without

jonnyshaw89 (changes requested):

app/data/schema/schema-v1.json
benjefferies: Who knows they appear on both in the state items!

benjefferies: Looks to be on questions

jonnyshaw89 (approved):

app/questionnaire/rules.py
```diff
@@ -69,6 +53,35 @@ def evaluate_repeat(repeat_rule, answer_store):
...
+    """
+    for when_rule in when_rules:
+        if 'id' in when_rule:
+            answer_index = when_rule['id']
+            filtered = answer_store.filter(answer_id=answer_index, group_instance=group_instance)
+
+            assert len(filtered) <= 1, "Condition will not met: Multiple ({:d}) answers found".format(len(filtered))
+
+            if len(filtered) == 1 and not evaluate_rule(when_rule, filtered[0]['value']):
+                return evaluate_rule(when_rule, filtered[0]['value'])
```
iwootten: Can't you just return False, given this is part of the if?

iwootten (changes requested):

app/questionnaire/rules.py
iwootten: In fact, think that's what it originally did right?

benjefferies: Fixed and rebase to speed up process of merging

iwootten (approved):

PR #756, opened 2016-12-16
Fixes #749 answers retaining values after changing household.

### LINKED ISSUE #749: 'Is this everyone for whom this is their permanent or family home?' question preselects answer
### Expected behaviour

Not all answers should retain their selected response when returned to. 

On the question 'Is this everyone for whom this is their permanent or family home?', upon answering 'No, I need to add another person', subsequently adding another person and then navigating back to the question, 'No...' answer should be not selected.

### Actual behaviour

The 'No, I need to add another person' answer is preselected—this is the least likely option that a user would select here and clicking 'Save and continue' without changing their 'answer' takes them backwards.

- start 'household_composition' questionnaire
- complete up to adding people, add one or more
- progress to 'Is this everyone for whom...' question
- select 'No..' option
- save and continue
- 'No...' option is already selected

### Technical information

N/A

### Screenshot

screenshot 2016-12-16 12 13 05

samiwel:
Fixes #749.

### How to review 
Retest the defect. Going back after entering household members and adding an additional member should clear the next block's answer.

### DIFF
--- a/app/helpers/schema_helper.py
+++ b/app/helpers/schema_helper.py
@@ -63,3 +63,34 @@ class SchemaHelper(object):
             repeating_rule = cls.get_repeating_rule(group)
             if repeating_rule and repeating_rule['answer_id'] == answer_id:
                 yield group
+
+    @classmethod
+    def get_blocks_that_route_back_to_block(cls, completed_blocks, survey_json, block_id):
+        # We're only concerned with blocks completed 'after' the block we're searching for
+        if block_id not in [b['block_id'] for b in completed_blocks]:
+            return []
+
+        blocks_that_route_back = []
+        for completed_block in reversed(completed_blocks):
+            block = cls.get_block(survey_json, completed_block['block_id'])
+
+            # Go back through the completed answers until we reach the block we're searching for.
+            if block['id'] == block_id:
+                break
+
+            if 'routing_rules' in block:
+                for routing_rule in block['routing_rules']:
+                    if 'goto' in routing_rule and routing_rule['goto']['id'] == block_id:
+                        blocks_that_route_back.append(block)
+                        break
+
+        return blocks_that_route_back
+
+    @classmethod
+    def get_answers_that_route_back_to_block(cls, blocks_that_route_back, survey_json):
+        for block_that_routes_back in blocks_that_route_back:
+            block = cls.get_block(survey_json, block_that_routes_back['id'])
+            for section in (s for s in block['sections'] if block['sections']):
+                for question in (q for q in section['questions'] if section['questions']):
+                    for answer in (a for a in question['answers'] if question['answers']):
+                        yield answer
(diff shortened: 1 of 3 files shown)


app/helpers/schema_helper.py
```diff
@@ -63,3 +63,34 @@ def get_groups_that_repeat_with_answer_id(cls, survey_json, answer_id):
             repeating_rule = cls.get_repeating_rule(group)
             if repeating_rule and repeating_rule['answer_id'] == answer_id:
                 yield group
+
+    @classmethod
+    def get_blocks_that_route_back_to_block(cls, completed_blocks, survey_json, block_id):
```
benjefferies, line 68: We should have test coverage of this

app/helpers/schema_helper.py
```diff
@@ -63,3 +63,34 @@ def get_groups_that_repeat_with_answer_id(cls, survey_json, answer_id):
...
+            if 'routing_rules' in block:
+                for routing_rule in block['routing_rules']:
+                    if 'goto' in routing_rule and routing_rule['goto']['id'] == block_id:
+                        blocks_that_route_back.append(block)
+                        break
+
+        return blocks_that_route_back
+
+    @classmethod
+    def get_answers_that_route_back_to_block(cls, blocks_that_route_back, survey_json):
```
benjefferies, line 90: We should have test coverage of this

app/helpers/schema_helper.py
```diff
@@ -63,3 +63,34 @@ def get_groups_that_repeat_with_answer_id(cls, survey_json, answer_id):
...
+        for block_that_routes_back in blocks_that_route_back:
+            block = cls.get_block(survey_json, block_that_routes_back['id'])
+            for section in (s for s in block['sections'] if block['sections']):
+                for question in (q for q in section['questions'] if section['questions']):
```
benjefferies, line 94: Should we refactor get_questions() out now we do it in a couple of places

benjefferies (changes requested):
We should have an integration tests to confirm it works - something like https://github.com/ONSdigital/eq-survey-runner/blob/master/tests/integration/questionnaire/test_questionnaire_remove_repeat_answers.py

app/helpers/schema_helper.py
samiwel: Yep, still in progress 👍

benjefferies: Sorry missed the label

samiwel: No worries, good to know about those integration tests anyway.

samiwel:
@ajmaddaford Could use another pair of eyes to understand why my integration test is not behaving as expected. I noticed different behaviour running the test versus running the application in the Navigator when building the path. I think it's because the answer store in my integration test is being mocked therefore has no real answers to determine the repeats. Yet, the block I'm expecting to find _IS_ in the list, so not sure if there's a problem in the logic of `Navigator::build_path()` or not.

samiwel:
Closing this PR since the defect is resolved by another PR (#775)

PR #759, opened 2016-12-16
Add Location class

iwootten:

Updates survey runner to use a Location class for representation of group/instance/block locations and updates navigator to return lists of them when requesting paths. 

Allows us to generate urls from locations when metadata is known, which negates the need for jinja filter to create navigation links.

Updated version of https://github.com/ONSdigital/eq-survey-runner/pull/661, based on huge amount of work gone in of late.

### How to review 
Take a look at the approach, see if you approve.

### DIFF
--- a/app/data_model/answer_store.py
+++ b/app/data_model/answer_store.py
@@ -151,7 +151,7 @@ class AnswerStore(object):
 
         return len(self.filter(answer.group_id, answer.block_id, answer.answer_id, answer.group_instance, answer.answer_instance))
 
-    def filter(self, group_id=None, block_id=None, answer_id=None, group_instance=None, answer_instance=None):
+    def filter(self, group_id=None, block_id=None, answer_id=None, group_instance=None, answer_instance=None, location=None):
         """
         Find all answers in the answer store for a given set of filter parameter matches.
         If no filter parameters are passed it returns a copy of the list of all answers.
@@ -171,6 +171,11 @@ class AnswerStore(object):
             "answer_instance": answer_instance,
             "group_instance": group_instance,
         }
+        if location:
+            filter_vars['group_id'] = location.group_id
+            filter_vars['group_instance'] = location.group_instance
+            filter_vars['block_id'] = location.block_id
+
         for answer in self.answers:
             matches = True
             for k, v in filter_vars.items():
@@ -214,7 +219,7 @@ class AnswerStore(object):
         if index is not None:
             del self.answers[index]
 
-    def remove(self, group_id=None, block_id=None, answer_id=None, group_instance=None, answer_instance=None):
+    def remove(self, group_id=None, block_id=None, answer_id=None, group_instance=None, answer_instance=None, location=None):
         """
         Removes answer(s) from the answer store.
 
@@ -223,8 +228,10 @@ class AnswerStore(object):
         :param group_id:
         :param answer_instance:
         :param group_instance:
+        :param location:
         """
-        for answer in self.filter(group_id, block_id, answer_id, group_instance, answer_instance):
+
+        for answer in self.filter(group_id, block_id, answer_id, group_instance, answer_instance, location):
             self.answers.remove(answer)
 
 
--- a/app/data_model/answer_store.py
+++ b/app/data_model/answer_store.py
@@ -156,12 +156,13 @@ class AnswerStore(object):
         Find all answers in the answer store for a given set of filter parameter matches.
         If no filter parameters are passed it returns a copy of the list of all answers.
 
-        :param answer_id:
-        :param block_id:
-        :param group_id:
-        :param answer_instance:
-        :param group_instance:
-        :return:
+        :param answer_id: The answer id to filter results by
+        :param block_id: The block id to filter results by
+        :param group_id: The group id to filter results by
+        :param answer_instance: The answer instance to filter results by
+        :param group_instance: The group instance to filter results by
+        :param location: The location to filter results by (takes precedence over group_id, group_instance and block_id)
+        :return: Return a list of answers which satisfy the filter criteria
         """
         filtered = []
         filter_vars = {
@@ -172,6 +173,9 @@ class AnswerStore(object):
             "group_instance": group_instance,
         }
         if location:
+            assert not (group_id or group_instance or block_id), \
+                "Expected either a location object or one or more of group_id, group_instance, block_id params"
+
             filter_vars['group_id'] = location.group_id
             filter_vars['group_instance'] = location.group_instance
             filter_vars['block_id'] = location.block_id
@@ -223,12 +227,12 @@ class AnswerStore(object):
         """
         Removes answer(s) from the answer store.
 
-        :param answer_id:
-        :param block_id:
-        :param group_id:
-        :param answer_instance:
-        :param group_instance:
-        :param location:
+        :param answer_id: The answer id to filter results to remove
+        :param block_id: The block id to filter results to remove
+        :param group_id: The group id to filter results to remove
+        :param answer_instance: The answer instance to filter results to remove
+        :param group_instance: The group instance to filter results to remove
+        :param location: The location to filter results to remove (takes precedence over group_id, group_instance and block_id)
         """
 
         for answer in self.filter(group_id, block_id, answer_id, group_instance, answer_instance, location):
--- a/app/questionnaire/location.py
+++ b/app/questionnaire/location.py
@@ -1,0 +1,74 @@
+from flask import url_for
+
+
+class Location(object):
+
+    def __init__(self, group_id, group_instance, block_id):
+
+        self.group_id = group_id
+        self.group_instance = group_instance
+        self.block_id = block_id
+
+    def __eq__(self, other):
+        """
+        Check to see if two locations are equal.
+        Two answers are considered to be equal if their dictionary representations equal one another.
+
+        :param other: An answer to compare
+        :return: True if both answers match, otherwise False.
+        """
+        return isinstance(other, Location) and self.__dict__ == other.__dict__
+
+    def __hash__(self):
+        return hash(self.__dict__.values())
+
+    def __str__(self):
+        """
+        String representation of the location, handy for debug messages
+
+        :return:
+        """
+        return "{}/{}/{}".format(self.group_id, self.group_instance, self.block_id)
+
+    def is_interstitial(self):
+        return self.block_id in ['introduction', 'summary', 'thank-you']
+
+    def url(self, metadata):
+        """
+        Return the survey runner url that this location represents
+
+        :param metadata:
+        :return:
+        """
+        eq_id = metadata["eq_id"]
+        collection_id = metadata["collection_exercise_sid"]
+        form_type = metadata["form_type"]
+
+        if self.is_interstitial():
+            if self.block_id == 'summary':
+                return url_for('questionnaire.get_summary',
+                               eq_id=eq_id,
+                               form_type=form_type,
+                               collection_id=collection_id)
+            elif self.block_id == 'introduction':
+                return url_for('questionnaire.get_introduction',
+                               eq_id=eq_id,
+                               form_type=form_type,
+                               collection_id=collection_id)
+            elif self.block_id == 'confirmation':
+                return url_for('questionnaire.get_confirmation',
+                               eq_id=eq_id,
+                               form_type=form_type,
+                               collection_id=collection_id)
+            elif self.block_id == 'thank-you':
+                return url_for('questionnaire.get_thank_you',
+                               eq_id=eq_id,
+                               form_type=form_type,
+                               collection_id=collection_id)
+        return url_for('questionnaire.get_block',
+                       eq_id=eq_id,
+                       form_type=form_type,
+                       collection_id=collection_id,
+                       group_id=self.group_id,
+                       group_instance=self.group_instance,
+                       block_id=self.block_id)
--- a/app/questionnaire/navigator.py
+++ b/app/questionnaire/navigator.py
@@ -1,9 +1,11 @@
+import copy
 import logging
 
 from collections import defaultdict
 
 from app.data_model.answer_store import AnswerStore
 from app.helpers.schema_helper import SchemaHelper
+from app.questionnaire.location import Location
 from app.questionnaire.rules import evaluate_goto, evaluate_repeat, evaluate_skip_condition, is_goto_rule
 
 
@@ -24,20 +26,15 @@ class Navigator:
         if SchemaHelper.has_introduction(self.survey_json):
             self.preceeding_path = self.PRECEEDING_INTERSTITIAL_PATH
 
-    @classmethod
-    def is_interstitial_block(cls, block_id):
-        return block_id in cls.PRECEEDING_INTERSTITIAL_PATH or block_id in cls.CLOSING_INTERSTITIAL_PATH
-
-    @classmethod
-    def _block_index_for_location(cls, blocks, location):
+    @staticmethod
+    def _block_index_for_location(blocks, location):
         try:
-            if not cls.is_interstitial_block(location['block_id']):
-                return next(index for (index, b) in enumerate(blocks) if b["block"]["id"] == location['block_id'] and
-                            b["group_id"] == location['group_id'] and b['group_instance'] == location['group_instance'])
+            if not location.is_interstitial():
+                return next(index for (index, b) in enumerate(blocks) if b["block"]["id"] == location.block_id and
+                            b["group_id"] == location.group_id and b['group_instance'] == location.group_instance)
         except StopIteration:
             logger.error("Navigation failure looking for %s", location)
             raise
-
         return None
 
     def build_path(self, blocks, this_location):
@@ -45,8 +42,8 @@ class Navigator:
         Visits all the blocks from a location forwards and returns path
         taken given a list of answers.
 
-        :param blocks: A list containing all blocks in the survey
-        :param this_location: The location to start navigating from
+        :param blocks: A list containing all block content in the survey
+        :param this_location: The location to visit, represented as a dict
         :return: A list of locations followed through the survey
         """
         path = []
@@ -55,10 +52,10 @@ class Navigator:
 
         # Keep going unless we've hit the last block
         while block_index < blocks_len:
-            if this_location['block_id'] in self.CLOSING_INTERSTITIAL_PATH:
+            if this_location.block_id in self.CLOSING_INTERSTITIAL_PATH:
                 return path
 
-            block_index = self._block_index_for_location(blocks, this_location)
+            block_index = Navigator._block_index_for_location(blocks, this_location)
             if block_index is None:
                 logger.error('build_path: _block_index_for_location %s is None (invalid location)', this_location)
                 return path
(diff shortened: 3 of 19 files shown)


app/main/views/questionnaire.py
```diff
@@ -240,33 +218,31 @@ def post_household_composition(eq_id, form_type, collection_id, group_id):
     questionnaire_manager = get_questionnaire_manager(g.schema, g.schema_json)
     answer_store = get_answer_store(current_user)
 
-    this_block = {
-        'block_id': 'household-composition',
-        'group_id': group_id,
-        'group_instance': 0,
-    }
+    this_location = Location(group_id, 0, 'household-composition')
```
collisdigital: does lose some readability by not having the name arguments, what about (skipping obvious named group_id):

```python
this_location = Location(
    group_id, 
    group_instance=0,
    block_id='household-composition')
```

Just for the cases when the parameters being passed in aren't matching names to the function args.

app/questionnaire/location.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def __eq__(self, other):
+        """
+        Check to see if two locations are equal.
+        Two answers are considered to be equal if their dictionary representations equal one another.
+
+        :param other: An answer to compare
+        :return: True if both answers match, otherwise False.
+        """
+        return self.__dict__ == other.__dict__
```
benjefferies: why self.__dict__ over vars(self)?

app/questionnaire/location.py
```diff
@@ -0,0 +1,71 @@
...
+
+class Location(object):
+
+    def __init__(self, group_id, group_instance, block_id):
+
+        self.group_id = group_id
+        self.group_instance = group_instance
+        self.block_id = block_id
+
+    def __eq__(self, other):
```
collisdigital, line 12: don't forget `__hash__`

app/questionnaire/navigator.py
```diff
@@ -54,7 +51,7 @@ def build_path(self, blocks, this_location):
 
         # Keep going unless we've hit the last block
         while block_index < blocks_len:
-            if this_location['block_id'] in self.CLOSING_INTERSTITIAL_PATH:
+            if this_location.block_id in self.CLOSING_INTERSTITIAL_PATH:
```
collisdigital, line 55: maybe this should also be in the Location class as well (if something's opening/closing interstitial?)

app/templates/partials/navigation.html
```diff
@@ -1,12 +1,12 @@
...
-    <li class="nav__item  pluto {{ 'nav__item--completed' if group.completed }} {{ 'nav__item--current' if group.highlight }}">
-      {% if group.current_location %}
-        <div>{{group.link_name}}</div>
+  {% for nav_item in navigation %}
+    <li class="nav__item  pluto {{ 'nav__item--completed' if nav_item.completed }} {{ 'nav__item--current' if nav_item.highlight }}">
+      {% if nav_item.current_location %}
+        <div>{{nav_item.link_name}}</div>
       {% else %}
-        <a class="nav__link" href="{{meta.survey|format_url(group.group_id, group.instance, group.block_id)}}">{{group.link_name}}</a>
+        <a class="nav__link" href="{{nav_item.link_location.url(meta.survey)}}">{{nav_item.link_name}}</a>
```
benjefferies: Would it not be better to expand this when we build the navigation context?

app/questionnaire/navigator.py
```diff
@@ -67,13 +64,13 @@ def build_path(self, blocks, this_location):
...
-                original_this_location = this_location.copy()
+                original_this_location = copy.copy(this_location)
                 for rule in filter(is_goto_rule, block['routing_rules']):
-                    should_goto = evaluate_goto(rule['goto'], self.metadata, self.answer_store, this_location['group_instance'])
+                    should_goto = evaluate_goto(rule['goto'], self.metadata, self.answer_store, this_location.group_instance)
 
                     if should_goto:
-                        next_location = this_location.copy()
-                        next_location['block_id'] = rule['goto']['id']
+                        next_location = copy.copy(this_location)
```
collisdigital, line 73: We can remove this copy now and just modify `this_location` directly as you can use `original_this_location` on line 79 to do the remove, this eliminates the need for `next_location`

tests/app/questionnaire/test_navigator.py
```diff
@@ -48,19 +59,39 @@ def test_introduction_not_in_path_when_not_in_schema(self):
...
     def test_next_with_conditional_path(self):
         survey = load_schema_file("0_star_wars.json")
 
         expected_path = [
-            {"block_id": "f22b1ba4-d15f-48b8-a1f3-db62b6f34cc0"},
-            {"block_id": "96682325-47ab-41e4-a56e-8315a19ffe2a"},
-            {"block_id": "cd3b74d1-b687-4051-9634-a8f9ce10a27d"},
-            {"block_id": "an3b74d1-b687-4051-9634-a8f9ce10ard"},
-            {"block_id": "846f8514-fed2-4bd7-8fb2-4b5fcb1622b1"}
+            Location(
```
benjefferies: Is it not clear enough if we one line the variables, it's a bit noisy otherwise

app/questionnaire/location.py
```diff
@@ -0,0 +1,71 @@
...
+        String representation of the location, handy for debug messages
+
+        :return:
+        """
+        return "{}/{}/{}".format(self.group_id, self.group_instance, self.block_id)
+
+    def is_interstitial(self):
+        return self.block_id in ['introduction', 'summary', 'thank-you']
+
+    def url(self, metadata):
```
benjefferies, line 36: We should probably have a unit test for this

app/questionnaire/navigator.py
```diff
@@ -90,11 +87,9 @@ def build_path(self, blocks, this_location):
 
             # No routing rules, so if this isn't the last block, step forward a block
             elif block_index < len(blocks) - 1:
-                this_location = {
-                    "block_id": blocks[block_index + 1]['block']['id'],
-                    "group_id": blocks[block_index + 1]['group_id'],
-                    "group_instance": blocks[block_index + 1]['group_instance'],
-                }
+                this_location = Location(blocks[block_index + 1]['group_id'],
```
collisdigital, line 91: What about `Location.from_schema_block(blocks[block_index+1])`

^^ as a static method on `Location` which returns a Location object.

benjefferies (changes requested):

app/questionnaire/navigator.py
```diff
@@ -110,9 +105,9 @@ def _remove_rule_answers(self, goto_rule, location):
                 if 'meta' not in condition.keys():
                     self.answer_store.remove(answer_id=condition['id'],
                                              answer_instance=0,
-                                             block_id=location['block_id'],
-                                             group_id=location['group_id'],
-                                             group_instance=location['group_instance'])
+                                             block_id=location.block_id,
```
collisdigital: Update answer_store to take a Location object?

app/questionnaire/location.py
iwootten: Why not?

benjefferies: Isn't vars() the pythonic way?

tests/app/questionnaire/test_navigator.py
iwootten: Guh, could've asked for these in my other PR, there's 301 of them! See you tomorrow.

app/questionnaire/navigator.py
```diff
@@ -266,7 +238,7 @@ def get_latest_location(self, completed_blocks=None):
         """
         location_path = self.get_location_path()
         if completed_blocks:
-            incomplete_blocks = [item for item in location_path if item not in completed_blocks]
+            incomplete_blocks = [item for item in location_path if item.__dict__ not in completed_blocks]
```
collisdigital: You could allow `Location.__eq__` to check if `other` is a dict and do a dict comparison so you can do `Location == {...}` ?

app/questionnaire/location.py
iwootten: Is it? Never heard of vars and a quick google seems to suggest using __dict__ is the preferred approach.

app/questionnaire/location.py
```diff
@@ -0,0 +1,71 @@
+from flask import url_for
+
+
+class Location(object):
+
+    def __init__(self, group_id, group_instance, block_id):
```
collisdigital, line 6: any value in defaulting these e.g. group_instance = 0 ?

app/questionnaire/location.py
collisdigital: http://stackoverflow.com/questions/21297203/use-dict-or-vars ?

app/questionnaire/navigator.py
collisdigital: I guess that's actually just a `from_dict(...)` method, except the 'block' bit, humph

app/questionnaire/location.py
benjefferies: From [Fluent Python](https://www.amazon.co.uk/Fluent-Python-Luciano-Ramalho/dp/1491946008)

How Special Methods Are Used
The first thing to know about special methods is that they are meant to be called by the Python interpreter, and not by you. You don’t write `my_object.__len__()`. You write len(my_object) and, if my_object is an instance of a user-defined class, then Python calls the `__len__` instance method you implemented.

tests/app/questionnaire/test_navigator.py
collisdigital: The problem is the GUID ids, not the multi-line? Sticking these on one line would be hard to follow I think

app/main/views/questionnaire.py
benjefferies: It's a preference thing but I think it's beneficial to keep Location(...) one line as long as it fits but named params make sense where applicable

app/questionnaire/navigator.py
benjefferies: You could have something like
```
def changing(group_id=None, block_id=None, group_instance=None):
    return Answer(group_id or self.group_id, block_id or self.block_id, group_instance or self.group_instance)
```

app/questionnaire/location.py
iwootten: Sorry, arbitrary links on the net do nothing for me: http://stackoverflow.com/questions/1227121/compare-object-instances-for-equality-by-their-attributes-in-python

collisdigital: https://www.google.co.uk/trends/explore?q=Python%20__dict__,Python%20vars ? 😁

collisdigital: I have no opinion on it, just supplying random web content as you suggest.

tests/app/questionnaire/test_navigator.py
iwootten: Too late, taken ages and removed them as requested. I tend to agree that we're going to know the order of the args and the real problem is the guids need updating.

app/main/views/questionnaire.py
iwootten: Too late now.

app/questionnaire/location.py
iwootten: Think it better being explicit. What would .url() do in such a case?

app/questionnaire/location.py
iwootten: ok

app/questionnaire/location.py
iwootten: Yes, as I was yesterday :)

app/questionnaire/navigator.py
iwootten: Really? If I make this change it fails all over the place.

app/questionnaire/location.py
```diff
@@ -0,0 +1,77 @@
...
+        """
+        Check to see if two locations are equal.
+        Two answers are considered to be equal if their dictionary representations equal one another.
+
+        :param other: An answer to compare
+        :return: True if both answers match, otherwise False.
+        """
+        return self.__dict__ == other.__dict__
+
+    def __ne__(self, other):
```
collisdigital: This is automatically defined by Python to be not '__eq__' so no need to add it

app/questionnaire/location.py
jonnyshaw89: I always favour consistency...... however github won't allow me to post the links

jonnyshaw89: __dict__ used 5 times in our code
vars(self) not used in our code

app/questionnaire/navigator.py
collisdigital: Not sure then, best leave it!

app/questionnaire/questionnaire_manager.py
```diff
@@ -60,11 +60,11 @@ def update_questionnaire_store(self, location):
         # Store answers in QuestionnaireStore
         questionnaire_store = get_questionnaire_store(current_user.user_id, current_user.user_ik)
 
-        for answer in self.get_state_answers(location['block_id']):
+        for answer in self.get_state_answers(location.block_id):
             questionnaire_store.answer_store.add_or_update(answer.flatten())
 
-        if location not in questionnaire_store.completed_blocks:
-            questionnaire_store.completed_blocks.append(location)
+        if location.__dict__ not in questionnaire_store.completed_blocks:
```
collisdigital: Same as above comment about `__eq__`

collisdigital:
Sorry Ian, I had started these review comments last week and didn't press the Submit button!

app/questionnaire/navigator.py
iwootten: We'd deliberately be masking a type conversion, which seems wrong...

app/main/views/questionnaire.py
```diff
@@ -123,28 +113,24 @@ def post_interstitial(eq_id, form_type, collection_id, block_id):
...
 
-    valid_location = this_block in navigator.get_location_path()
-    q_manager.process_incoming_answers(this_block, request.form)
+    valid_location = this_location in navigator.get_location_path()
+    q_manager.process_incoming_answers(this_location, request.form)
 
     # Don't care if data is valid because there isn't any for interstitial
     if not valid_location:
-        return _render_template(q_manager.state, this_block['group_id'], this_block['group_instance'], block_id, template='questionnaire')
+        return _render_template(q_manager.state, this_location.group_id, this_location.group_instance, block_id, template='questionnaire')
```
ajmaddaford: Can we change _render_template to take a location parameter? This would look much neater throughout as `_render_template(q_manager.state, this_location, template='questionnaire'`

app/questionnaire/location.py
```diff
@@ -0,0 +1,74 @@
...
+    def __str__(self):
+        """
+        String representation of the location, handy for debug messages
+
+        :return:
+        """
+        return "{}/{}/{}".format(self.group_id, self.group_instance, self.block_id)
+
+    def is_interstitial(self):
+        return self.block_id in ['introduction', 'summary', 'thank-you']
```
ajmaddaford, line 34: I think we need a better name for these, as we can have blocks of type 'interstitial' in a survey.

app/questionnaire/questionnaire_manager.py
ajmaddaford: Maybe we should think about completed blocks being a list of Location objects, and when we serialise/deserialise in QuestionnaireStore we convert it to/from a dict.

app/submitter/converter.py
```diff
@@ -128,7 +128,9 @@ def convert_answers_to_census_data(answer_store, routing_path):
     """
     data = []
     for location in routing_path:
-        answers_in_block = answer_store.filter(group_id=location['group_id'], group_instance=location['group_instance'], block_id=location['block_id'])
+        answers_in_block = answer_store.filter(group_id=location.group_id,
+                                               group_instance=location.group_instance,
+                                               block_id=location.block_id)
```
ajmaddaford: Can this be replaced by `answer_store.filter(location=location)`?

app/templating/summary_context.py
```diff
@@ -21,7 +21,7 @@ def build_summary_rendering_context(schema_json, answer_store, metadata):
     sections = []
     for group in schema_json['groups']:
         for block in group['blocks']:
-            if block['id'] in [b['block_id'] for b in path]:
+            if block['id'] in [l.block_id for l in path]:
```
ajmaddaford: I think using `location` rather than `l` would be clearer here

ajmaddaford (dismissed):

tests/app/questionnaire/test_location.py
```diff
@@ -0,0 +1,26 @@
+import unittest
+
+from app import create_app
+
+from app.questionnaire.location import Location
+
+
+class TestLocation(unittest.TestCase):
+
+    def setUp(self):
```
benjefferies, line 10: This setUp shouldn't be needed

app/data_model/answer_store.py
```diff
@@ -151,7 +151,7 @@ def count(self, answer):
 
         return len(self.filter(answer.group_id, answer.block_id, answer.answer_id, answer.group_instance, answer.answer_instance))
 
-    def filter(self, group_id=None, block_id=None, answer_id=None, group_instance=None, answer_instance=None):
+    def filter(self, group_id=None, block_id=None, answer_id=None, group_instance=None, answer_instance=None, location=None):
         """
         Find all answers in the answer store for a given set of filter parameter matches.
         If no filter parameters are passed it returns a copy of the list of all answers.
```
benjefferies, line 157: Shall we remove the :param and :return: documentation if we aren't going to add detail otherwise it's just boilerplate noise

app/data_model/answer_store.py
```diff
@@ -223,8 +228,10 @@ def remove(self, group_id=None, block_id=None, answer_id=None, group_instance=No
         :param group_id:
         :param answer_instance:
         :param group_instance:
+        :param location:
```
benjefferies: Shall we remove the :param and :return: documentation if we aren't going to add detail otherwise it's just boilerplate noise

app/questionnaire/location.py
benjefferies: Done

app/questionnaire/navigator.py
```diff
@@ -170,21 +159,13 @@ def get_location_path(self, group_id=None, group_instance=0):
         can_reach_summary = self.can_reach_summary(routing_path)
 
         # Make sure we don't update original
```
benjefferies: I think this comment is now stale that we don't keep state of the routing_path

benjefferies (changes requested):
Think there are a few tidy ups we could apply before merging back. It's hard to know what has been actioned/discussed offline etc. It would be good to see a comment to let us know what is happening with those. In general this is a good tidy up

tests/app/questionnaire/test_location.py
iwootten: It is, as the app context is necessary to allow the url generation.

app/questionnaire/navigator.py
iwootten: Not sure what is being asked here, either solution seems to be more complex than what is in place already.

app/questionnaire/location.py
iwootten: Agreed offline with @ajmaddaford for this to be looked as part of making interstitials proper blocks.

app/questionnaire/navigator.py
collisdigital: yeah no clean way to deal with it; one to revisit in future

app/questionnaire/navigator.py
iwootten: Agreed offline with @ajmaddaford for this to be looked as part of making interstitials proper blocks.

app/questionnaire/navigator.py
collisdigital: True, although this is implicitly stating that a location `__dict__` matches a block dict structure; i.e. all code using this method of comparison is making that assumption rather than Location object itself doing so. Not sure what's better, leave it to your discretion.

app/data_model/answer_store.py
```diff
@@ -151,17 +151,18 @@ def count(self, answer):
...
-        :param group_id:
-        :param answer_instance:
-        :param group_instance:
-        :return:
+        :param answer_id: The answer id to filter results by
+        :param block_id: The block id to filter results by
+        :param group_id: The group id to filter results by
+        :param answer_instance: The answer instance to filter results by
+        :param group_instance: The group instance to filter results by
+        :param location: The location to filter results by (takes precedence over group_id, group_instance and block_id)
```
benjefferies, line 164: 👍

benjefferies (approved):
Looks good to me!

app/data_model/answer_store.py
```diff
@@ -171,6 +172,11 @@ def filter(self, group_id=None, block_id=None, answer_id=None, group_instance=No
             "answer_instance": answer_instance,
             "group_instance": group_instance,
         }
+        if location:
```
collisdigital, line 175: maybe an assert/warning/exception when location  and the individual settings are both set, since its overwriting them here and the caller needs to decide what's being used.

collisdigital (approved):
Reviewed and Approved

app/data_model/answer_store.py
iwootten: Added an assert here

PR #760, opened 2016-12-16
Stop passing the full schema into the templates

jonnyshaw89:
We are currently passing the full schema through to the templates.
This isn't needed as we are only using the title from it.

### DIFF
--- a/app/templates/partials/topbar.html
+++ b/app/templates/partials/topbar.html
@@ -2,10 +2,10 @@
   <div class="bar__inner container">
     <div class="badge badge--amber u-mr-xs">BETA</div>
     {% if is_hero %}
-      <h1 class="bar__title jupiter">{% if meta %}{{ schema['title'] }}{% endif %}</h1>
+      <h1 class="bar__title jupiter">{{ schema_title }}</h1>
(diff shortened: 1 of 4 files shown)


app/templates/partials/topbar.html
```diff
@@ -2,10 +2,10 @@
   <div class="bar__inner container">
     <div class="badge badge--amber u-mr-xs">BETA</div>
     {% if is_hero %}
-      <h1 class="bar__title jupiter">{% if meta %}{{ schema['title'] }}{% endif %}</h1>
+      <h1 class="bar__title jupiter">{% if meta %}{{ schema_title }}{% endif %}</h1>
```
iwootten: Why the meta check?

iwootten (changes requested):

benjefferies (approved):

app/templates/partials/topbar.html
jonnyshaw89: why was it checking it before? ill remove that

iwootten: No idea..?

iwootten (approved):

PR #766, opened 2016-12-19
Fix the routing on other box value for Census schemas

### LINKED ISSUE #748: Census_Household: Navigation incorrect - Individual section Q 5 when user selects - Yes an address outside the UK
### Expected behaviour
Must navigate to Q 6 What is that address?

### Actual behaviour
Navigate to Q 7 which is incorrect

Launch Census_Household
Navigate to individual section 
for Question 5 select Yes an address outside the UK and enter any country name
Click Save and continue

jonnyshaw89:
Have fixed the routing on questions where the other option takes a value.

Have also updated the selenium page models to include a setter for the other box value

Fixes https://github.com/ONSdigital/eq-survey-runner/issues/748

### DIFF
--- a/app/data/census_household.json
+++ b/app/data/census_household.json
@@ -1480,6 +1480,18 @@
                         },
                         {
                             "goto": {
+                                "id": "address-type",
+                                "when": [
+                                    {
+                                        "id": "another-address-answer",
+                                        "condition": "not equals",
+                                        "value": "Other"
+                                    }
+                                ]
+                            }
+                        },
+                        {
+                            "goto": {
                                 "id": "in-education"
                             }
                         }
--- a/app/data/census_individual.json
+++ b/app/data/census_individual.json
@@ -276,6 +276,18 @@
                         },
                         {
                             "goto": {
+                                "id": "address-type",
+                                "when": [
+                                    {
+                                        "id": "another-address-answer",
+                                        "condition": "not equals",
+                                        "value": "Other"
+                                    }
+                                ]
+                            }
+                        },
+                        {
+                            "goto": {
                                 "id": "in-education"
                             }
                         }
--- a/tests/functional/spec/census/routing/another-address.spec.js
+++ b/tests/functional/spec/census/routing/another-address.spec.js
@@ -94,7 +94,7 @@ describe('Another address routing Scenarios', function () {
     expect(OtherAddress.isOpen()).to.equal(true, 'Expecting go to other address')
   })
 
-  it('Given I am answering question 5 in the individual detail section -Do you stay at another address for more than 30 days a year?, When I select -Yes, an address outside the UK- and enter text in other field as response, Then I am routed to 6. What is that address?', function () {
+  it('Given I am answering question 5 in the individual detail section -Do you stay at another address for more than 30 days a year?, When I select -Yes, an address outside the UK, Then I am routed to 6. What is that address?', function () {
     startCensusQuestionnaire('census_household.json', true)
     PermanentOrFamilyHome.clickPermanentOrFamilyHomeAnswerYes().submit()
     HouseholdComposition.setFirstName('John').addPerson().setFirstName('Jane', 1).submit()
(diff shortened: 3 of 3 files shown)


tests/functional/pages/surveys/census/communal/establishment-type.page.js
```diff
@@ -33,6 +33,11 @@ class EstablishmentTypePage extends MultipleChoiceWithOtherPage {
     return this
   }
 
+  setEstablishmentTypeAnswerOtherOtherText(value) {
```
iwootten: *OtherOtherText doesn't seem like the best method name?

tests/functional/pages/surveys/census/communal/further-contact.page.js
```diff
@@ -1,4 +1,4 @@
-// >>> WARNING THIS PAGE WAS AUTO-GENERATED ON 2016-12-14 14:19:14.086947 - DO NOT EDIT!!! <<<
+// >>> WARNING THIS PAGE WAS AUTO-GENERATED - DO NOT EDIT!!! <<<
```
iwootten, line 1: Do we really need to update all these comments as part of this?

iwootten (changes requested):

tests/functional/pages/surveys/census/communal/further-contact.page.js
jonnyshaw89: Its really annoying having the time stamps in them as everytime you regenerate the pages arn't identical if they haven't changed.

tests/functional/pages/surveys/census/communal/establishment-type.page.js
jonnyshaw89: it uses the name of the radio box suffixed with OtherText if the radio option isn't Other would it still make sense?

tests/functional/pages/surveys/census/communal/further-contact.page.js
iwootten: On the flipside, it's also really annoying to review a PR where the change is muddled in amongst hundreds of trivial touches like this. Probably should have been the subject of its own PR.

tests/functional/pages/surveys/census/communal/establishment-type.page.js
iwootten: I guess merely adding a suffix probably isn't the best option then? Surely Other and OtherText would be a more sensible choice?

tests/functional/pages/surveys/census/communal/further-contact.page.js
jonnyshaw89: Yeh i agree, which should mean this would be the last?

tests/functional/pages/surveys/census/communal/establishment-type.page.js
jonnyshaw89: But if the text isn't Other it would read something like `setEstablishmentTypeAnswerHotelText`?

tests/functional/pages/surveys/census/communal/further-contact.page.js
collisdigital: Options:
- don't commit the files if nothing has changed (i.e. just the timestamp)
- don't have them in Git at all (generate on build time?)
- remove the timestamp?

tests/functional/pages/surveys/census/communal/establishment-type.page.js
jonnyshaw89: @iwootten i was getting confused thinking that the name of the function was generated from the label. apparently its from the answer value so i have updated this.

tests/functional/pages/surveys/census/communal/further-contact.page.js
ajmaddaford: Can we create another pull request for the addition of the of the other functionality and the re-generation of the page specs?

jonnyshaw89: Have raised https://github.com/ONSdigital/eq-survey-runner/pull/768

benjefferies (approved):
Looks good, can you rebase incase I have missed something amongst the noise

iwootten (approved):
Ah, that's better

PR #767, opened 2016-12-19
Add flask debug toolbar extension

jonnyshaw89:
Adds a toolbar to make debugging flask and development of templates easier.

### How to review 
Set EQ_ENABLE_FLASK_DEBUG_TOOLBAR to true and check that the flask toolbar is displayed on screen when running the app.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -3,10 +3,8 @@ import logging
 import os
 import sys
 from datetime import timedelta
-from logging.handlers import RotatingFileHandler
 
 from app import settings
-from app.analytics.custom_google_analytics import CustomGoogleAnalytics
 from app.authentication.authenticator import Authenticator
 from app.authentication.cookie_session import SHA256SecureCookieSessionInterface
 from app.data_model.database import db_session
@@ -16,8 +14,6 @@ from app.submitter.submitter import SubmitterFactory
 from flask import Flask
 from flask import url_for
 
-from flask_analytics import Analytics
-
 from flask_babel import Babel
 
 from flask_login import LoginManager
@@ -25,8 +21,6 @@ from flask_login import LoginManager
 from flask_themes2 import Themes
 
 from flaskext.markdown import Markdown
-
-from healthcheck import HealthCheck
 
 from splunk_handler import SplunkHandler
 
@@ -175,6 +169,8 @@ def setup_profiling(application):
 
 def setup_analytics(application):
     # Setup analytics
+    from flask_analytics import Analytics
+    from app.analytics.custom_google_analytics import CustomGoogleAnalytics
 
     Analytics.provider_map['google_analytics'] = CustomGoogleAnalytics
     Analytics(application)
@@ -183,6 +179,9 @@ def setup_analytics(application):
 
 
 def configure_logging(application):
+
+    from logging.handlers import RotatingFileHandler
+
     # set up some sane logging, as opposed to what flask does by default
     log_format = "[%(asctime)s] %(levelname)s [%(name)s.%(funcName)s:%(lineno)d] %(message)s"
     levels = {
@@ -224,6 +223,7 @@ def configure_logging(application):
 
 
 def setup_splunk_logging():
+
     splunk_handler = SplunkHandler(host=settings.EQ_SPLUNK_HOST,
                                    port=settings.EQ_SPLUNK_PORT,
                                    username=settings.EQ_SPLUNK_USERNAME,
@@ -234,6 +234,7 @@ def setup_splunk_logging():
 
 
 def setup_cloud_watch_logging(application):
+
     # filter out botocore messages, we don't wish to log these
     class NoBotocoreFilter(logging.Filter):
 
@@ -259,6 +260,12 @@ def start_dev_mode(application):
     application.debug = True
     # Not in dev mode, so use secure_session_cookies
     application.config['SESSION_COOKIE_SECURE'] = False
+
+    if settings.EQ_ENABLE_FLASK_DEBUG_TOOLBAR:
+        from flask_debugtoolbar import DebugToolbarExtension
+        DebugToolbarExtension(application)
+        application.config['DEBUG_TB_PROFILER_ENABLED'] = True
+        application.config['DEBUG_TB_INTERCEPT_REDIRECTS'] = False
 
 
 def add_blueprints(application):
@@ -294,6 +301,7 @@ def setup_babel(application):
 
 
 def add_health_check(application, headers):
+    from healthcheck import HealthCheck
     application.healthcheck = HealthCheck(
         application, '/healthcheck', success_headers=headers, failed_headers=headers)
     application.healthcheck.add_check(rabbitmq_available)
--- a/app/authentication/authenticator.py
+++ b/app/authentication/authenticator.py
@@ -29,7 +29,7 @@ class Authenticator(object):
             questionnaire_store = get_questionnaire_store(user.user_id, user.user_ik)
             metadata = questionnaire_store.metadata
 
-            logger.info("Session token exists for tx_id=%s", metadata["tx_id"])
+            logger.debug("Session token exists for tx_id=%s", metadata["tx_id"])
 
             return user
         else:
--- a/requirements.txt
+++ b/requirements.txt
@@ -259,3 +259,7 @@ flake8-tuple==0.2.12 \
 pycodestyle==2.0.0 \
     --hash=sha256:2ce83f2046f5ab85c652ceceddfbde7a64a909900989b4b43e92b10b743d0ce5 \
     --hash=sha256:37f0420b14630b0eaaf452978f3a6ea4816d787c3e6dcbba6fb255030adae2e7
+flask-debugtoolbar==0.10.0 \
+    --hash=sha256:d10cb2f58c9cc40e8bb8eb7e086cf3a5c18a7d00dab4a1561fe91e3cea92faf3
+Blinker==1.4 \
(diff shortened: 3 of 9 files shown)


app/__init__.py
```diff
@@ -135,6 +136,11 @@ def override_url_for():  # pylint: disable=unused-variable
         add_health_check(application, headers)
         start_dev_mode(application)
 
+    if settings.EQ_FLASK_DEBUG:
+        DebugToolbarExtension(application)
+        application.config['DEBUG_TB_PROFILER_ENABLED'] = True
+        application.config['DEBUG_TB_INTERCEPT_REDIRECTS'] = False
```
collisdigital: Maybe pull this one out as an env var too so we can enable it locally?

app/__init__.py
```diff
@@ -292,6 +298,7 @@ def setup_babel(application):
     application.babel = Babel(application)
     application.babel.localeselector(get_locale)
     application.jinja_env.add_extension('jinja2.ext.i18n')
+    application.jinja_env.cache = {}
```
collisdigital: did you notice any improvement from this in the end? I wonder if it would still be effective when running behind Apache WSGI

requirements.txt
```diff
@@ -259,3 +259,7 @@ flake8-tuple==0.2.12 \
 pycodestyle==2.0.0 \
     --hash=sha256:2ce83f2046f5ab85c652ceceddfbde7a64a909900989b4b43e92b10b743d0ce5 \
     --hash=sha256:37f0420b14630b0eaaf452978f3a6ea4816d787c3e6dcbba6fb255030adae2e7
+flask-debugtoolbar==0.10.0 \
```
collisdigital, line 262: Should these be in a different requirements.txt (we don't we have a prod vs. pre-prod/dev list yet) lots of stuff going into a prod build that isn't needed potentially.

jonnyshaw89: I don't think thats a good idea, essentially then you would be running different code in prod

app/__init__.py
jonnyshaw89: im not too sure if its even effective here. but I'm not sure why it would behave any differently on Apache?

collisdigital: Because of the way the app is *not* launched in Apache, it uses the WSGI hooks via application.py (not `__main__`) so not sure exactly how it loads the app on each request (caches may get wiped/not persisted) someone who understands WSGI better than me would probably know...

jonnyshaw89: What i mean is I don't think it works locally either. it doesn't persist the cache across requests. il going to remove this as the default is a cache of 400 objects. which should be ample, however it works

app/__init__.py
jonnyshaw89: I think its probably easier to use the request interception in chrome than switch this on and off

requirements_for_test.txt
```diff
@@ -5,3 +5,4 @@ coverage==4.2b1
 pytest-cov==2.3.0
 jsonschema==2.5.1
 pylint==1.6.4
+flask-debugtoolbar==0.10.0
```
benjefferies, line 8: Do we need this in both?

tests/integration/test_application_variables.py
```diff
@@ -0,0 +1,33 @@
+from app import settings
+from tests.integration.create_token import create_token
+from tests.integration.integration_test_case import IntegrationTestCase
+
+
+class TestFlaskToolbar(IntegrationTestCase):
```
benjefferies: Probably worth having a test that it's not shown when the flag is false

benjefferies (changes requested):

app/__init__.py
```diff
@@ -134,6 +124,12 @@ def override_url_for():  # pylint: disable=unused-variable
         add_health_check(application, headers)
         start_dev_mode(application)
 
+    if settings.EQ_FLASK_DEBUG:
```
ajmaddaford: I think we should rename the environment variable to `EQ_ENABLE_FLASK_DEBUG_TOOLBAR` and move the logic into the `start_dev_mode` method.

ajmaddaford (changes requested):

requirements.txt
ajmaddaford: I think putting it in requirements.txt is consistent with the other dev/build requirements like flake8. Maybe we could raise a tech debt card to review development/build dependencies in requirements.txt?

requirements_for_test.txt
jonnyshaw89: Apparently so. i took it out and it failed my tests locally.

requirements.txt
jonnyshaw89: raised https://trello.com/c/tdisphoa/339-review-development-build-dependencies-in-requirements-txt

app/authentication/authenticator.py
```diff
@@ -29,7 +29,7 @@ def check_session():
             questionnaire_store = get_questionnaire_store(user.user_id, user.user_ik)
             metadata = questionnaire_store.metadata
 
-            logger.info("Session token exists for tx_id=%s", metadata["tx_id"])
+            logger.debug("Session token exists for tx_id=%s", metadata["tx_id"])
```
benjefferies, line 32: 👍

tests/integration/questionnaire/test_questionnaire_interstitial.py
```diff
@@ -24,7 +24,7 @@ def test_interstitial_page_button_text_is_continue(self):
         resp_url, resp = self.postRedirectGet(resp_url, post_data)
         content = resp.get_data(True)
 
-        self.assertRegex("/Continue$", content)
+        self.assertTrue("Continue" in content)
```
benjefferies: The prior gives a better error if this fails otherwise we'll get expected true was false not "expecting … to contain Continue"

tests/integration/test_application_variables.py
```diff
@@ -0,0 +1,37 @@
...
+    def setUp(self):
+        settings.EQ_ENABLE_FLASK_DEBUG_TOOLBAR = True
+        settings.EQ_UA_ID = 'TestId'
+        IntegrationTestCase.setUp(self)
+
+    def tearDown(self):
+        settings.EQ_ENABLE_FLASK_DEBUG_TOOLBAR = False
+        IntegrationTestCase.tearDown(self)
+
+    def test_flask_toolbar_is_displayed(self):
```
benjefferies, line 17: Should we have the opposite test of this?

benjefferies (changes requested):

tests/integration/test_application_variables.py
jonnyshaw89: Thats a lot of overhead to write a whole separate test to check that something isn't there?

benjefferies: Or maybe we should have a unit test at least to test that when it is false we don't configure it?

jonnyshaw89: Isn't that the same thing?

benjefferies: I wasn't sure what time of overhead you meant as to write the other test is surely a copy and paste with inverted conditions?

tests/integration/test_application_variables.py
```diff
@@ -0,0 +1,37 @@
...
+
+        token = create_token('star_wars', '0')
+        resp = self.client.get('/session?token=' + token.decode(), follow_redirects=True)
+
+        self.assertEquals(resp.status_code, 200)
+
+        content = resp.get_data(True)
+
+        self.assertTrue("flDebugToolbarHandle" in content)
```
benjefferies: Using self.assertRegex would be preferable for failing test output

tests/integration/test_application_variables.py
```diff
@@ -0,0 +1,37 @@
...
+    def test_google_analytics_code_is_present(self):
+
+        token = create_token('star_wars', '0')
+        resp = self.client.get('/session?token=' + token.decode(), follow_redirects=True)
+
+        self.assertEquals(resp.status_code, 200)
+
+        content = resp.get_data(True)
+
+        self.assertTrue("GoogleAnalyticsObject" in content)
```
benjefferies: Using self.assertRegex would be preferable for failing test output

tests/integration/test_application_variables.py
jonnyshaw89: I have added a negative test for this.

tests/integration/test_application_variables.py
jonnyshaw89: is it preferable to print out the full html of the page that the content isn't in?
assertRegex gives
`AssertionError: Regex didn't match: '/Continue$' not found in '<!doctype html>\n<!-.....`

benjefferies: Hmm forgot it's the whole document. I think your right in this case. I wonder if using a dom library may be nicer then the regex but that's for another time.

benjefferies (approved):

tests/integration/questionnaire/test_questionnaire_interstitial.py
```diff
@@ -24,7 +24,7 @@ def test_interstitial_page_button_text_is_continue(self):
         resp_url, resp = self.postRedirectGet(resp_url, post_data)
         content = resp.get_data(True)
 
-        self.assertTrue("Continue" in content, "Interstitial page does not contain Continue as button text")
+        self.assertRegex(content, "Continue")
```
collisdigital: There is also an AssertContains I believe that avoids the need to use a regex...

ajmaddaford (approved):

PR #768, opened 2016-12-19
Re-generate page models with other option text

jonnyshaw89:
Re-generated the selenium page models to include other option text box and remove timestamp

### DIFF
--- a/tests/functional/generate_pages.py
+++ b/tests/functional/generate_pages.py
@@ -32,7 +32,7 @@ describe('Example Test', function() {
 
 """
 
-HEADER = r"""// >>> WARNING THIS PAGE WAS AUTO-GENERATED ON {generated} - DO NOT EDIT!!! <<<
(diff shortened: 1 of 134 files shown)


ajmaddaford (approved):

tests/functional/generate_pages.py
```diff
@@ -69,6 +69,13 @@
 
 """
 
+MULTIPLE_CHOICE_OTHER = r"""  set{answerName}(value) {
```
benjefferies, line 72: Do we need the extra template as it's just a setter right?

benjefferies (changes requested):

tests/functional/generate_pages.py
jonnyshaw89: Yes as the name is not unique so the existing ANSWER_SETTER won't work

benjefferies: Missed it referenced the id

benjefferies (approved):

PR #770, opened 2016-12-19
Removing unused fields from state e.g. skip_condition

benjefferies:
skip_condition is defined on various state items yet it is not ever referenced. Having state which isn't ever used is maintenance overhead and confusing. There are other fields which are defined by not referenced.

### How to review 
All tests should pass

### DIFF
--- a/app/data/schema/schema-v1.json
+++ b/app/data/schema/schema-v1.json
@@ -243,19 +243,6 @@
                                   },
                                   "alias": {
                                     "type": "string"
-                                  },
-                                  "display": {
-                                    "type": "object",
-                                    "properties": {
-                                      "properties": {
-                                        "type": "object",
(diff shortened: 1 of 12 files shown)


jonnyshaw89:
Nice. Just a few codecov misses.

benjefferies:
Deleting more code makes the codecov pass now @jonnyshaw89

ajmaddaford (approved):
Nice!

LJBabbage (changes requested):
We use display for 2 colums at answer level, so we can't remove it without approval from the designers

jonnyshaw89:
@hamishtaplin can you confirm what @LJBabbage is saying? do you guys use this?

LJBabbage (approved):

PR #771, opened 2016-12-19
Remove unnecessary views folder

iwootten:

Remove the main folder and move the views subdirectory up one level. Rename 'root' routes to 'main'.

### How to review 

Check if tests continue to pass.


app/views/main.py
```diff
@@ -22,21 +22,21 @@
 logger = logging.getLogger(__name__)
 
 
-root_blueprint = Blueprint('root', __name__)
```
jonnyshaw89, line 25: is main not almost as generic a name as root?

iwootten: Give me a more appropriate name and I'll change it. My concern with the name "root" was it implies some level of access "main" doesn't.

app/views/main.py
```diff
@@ -22,21 +22,21 @@
...
 
-@root_blueprint.after_request
+@main_blueprint.after_request
 def add_cache_control(response):
     response.cache_control.no_cache = True
     return response
 
 
-@root_blueprint.route('/', methods=['GET'])
+@main_blueprint.route('/', methods=['GET'])
```
benjefferies: This isn't actually required

app/views/main.py
benjefferies: I thought that root was referring to the root of the url path rather than access

app/views/main.py
```diff
@@ -22,21 +22,21 @@
...
 def root():
     raise NotFound
 
 
-@root_blueprint.route('/information/<message_identifier>', methods=['GET'])
+@main_blueprint.route('/information/<message_identifier>', methods=['GET'])
```
benjefferies: I think we can delete this. Instead of redirecting the to a different URL we could raise a MultipleSurveyError and handle it in errors.py. That would leave the only route as the `/session` route. That might solve our blueprint naming issue as we could call it authentication/session_blueprint

app/views/questionnaire.py
```diff
@@ -43,7 +43,7 @@ def check_survey_state():
     values = request.view_args
 
     if not _same_survey(values['eq_id'], values['form_type'], values['collection_id']):
```
benjefferies: If we throw and exception we could rename to _check_same_survey and we would not longer need the condition

benjefferies (changes requested):

iwootten:
@benjefferies @jonnyshaw89 - I'm just removing and renaming a file chaps, not looking for feedback on anything else.

benjefferies (approved):

jonnyshaw89 (approved):

jonnyshaw89:
👍

PR #775, opened 2016-12-20
Save state on navigator answer_store mutation

### LINKED ISSUE #749: 'Is this everyone for whom this is their permanent or family home?' question preselects answer

iwootten:
Save the answerstore on navigator mutation of it. 

Fixes #749

### How to review 
See if issue is fixed, tests pass.

### DIFF
--- a/app/questionnaire/questionnaire_manager.py
+++ b/app/questionnaire/questionnaire_manager.py
@@ -66,8 +66,6 @@ class QuestionnaireManager(object):
         if location not in questionnaire_store.completed_blocks:
             questionnaire_store.completed_blocks.append(location)
 
-        questionnaire_store.save()
-
     def process_incoming_answers(self, location, post_data):
         logger.debug("Processing post data for %s", location)
 
--- a/app/data_model/questionnaire_store.py
+++ b/app/data_model/questionnaire_store.py
@@ -1,3 +1,4 @@
+import json
 import logging
 
 from app.data_model.answer_store import AnswerStore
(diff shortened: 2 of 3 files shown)


app/questionnaire/questionnaire_manager.py
```diff
@@ -56,7 +56,7 @@ def validate_all_answers(self):
 
         return True, None
 
-    def update_questionnaire_store(self, location):
+    def update_questionnaire_store(self, location, save=True):
```
benjefferies: Maybe a save method would be more explicit than a flag which modifies the behaviour

benjefferies: Maybe a save method would be more explicit than a flag which modifies the behaviour

benjefferies (changes requested):

app/questionnaire/questionnaire_manager.py
iwootten: Maybe, but then we'd be calling save twice - once as part of this update method and once in the new save method.

benjefferies: Internet reference from a [great book](https://www.amazon.co.uk/Clean-Code-Handbook-Software-Craftsmanship/dp/0132350882)
Flag arguments are ugly. Passing a boolean into a function is a truly terrible practice. It immediately complicates the signature of the method, loudly proclaiming that this function does more than one thing. It does one thing if the flag is true and another if the flag is false!

iwootten: Yes, it's ugly. I understand. I take my orders from above. Talk to @ajmaddaford

benjefferies: This self organising thing is great 👍

benjefferies (approved):

benjefferies (changes requested):
Reverting approval as we are awaiting updated implementation

samiwel (approved):
Agree that there's probably a more elegant solution. But this PR does fix the issue at hand and the change is fairly minimal. So as Ben suggested maybe we create a Tech Debt card to look at it again some time.

app/main/views/questionnaire.py
```diff
@@ -49,6 +49,18 @@ def check_survey_state():
 @questionnaire_blueprint.after_request
 def add_cache_control(response):
     response.cache_control.no_cache = True
+
+    return response
+
+
+@questionnaire_blueprint.after_request
+def save_questionnaire_store(response):
+    if not current_user.is_anonymous:
```
benjefferies: I think we wouldn't need this change if we annotate with @login_required

app/main/views/questionnaire.py
```diff
@@ -49,6 +49,18 @@ def check_survey_state():
 @questionnaire_blueprint.after_request
 def add_cache_control(response):
     response.cache_control.no_cache = True
+
+    return response
+
+
+@questionnaire_blueprint.after_request
+def save_questionnaire_store(response):
```
jonnyshaw89: Nice and tidy :)

jonnyshaw89 (approved):

benjefferies (approved):

app/main/views/questionnaire.py
iwootten: Can't get this approach to work - possibly different to the check_survey_state example because it's after the request?

PR #776, opened 2016-12-20
Refactor tidy up main blueprint

benjefferies:
    Tidying up blueprint
    
    * Removing unnecessary `/` route handling
    * Throwing exception for multiple surveys
    * Renaming session

### How to review 
Test should pass and agree fits best practices


LJBabbage:
It was intended that it was a multi purpose message page, however this has made it into just the 2 surveys open message, I am not sure we want to do that.

app/views/session.py
```diff
@@ -13,40 +12,23 @@
...
 
 
-@main_blueprint.after_request
+@session_blueprint.after_request
 def add_cache_control(response):
     response.cache_control.no_cache = True
     return response
 
 
-@main_blueprint.route('/', methods=['GET'])
```
jonnyshaw89, line 33: What happens now if you navigate to /?

benjefferies: The same, 404

collisdigital: / should be 200; apache relies on it for health checks (unless that's changed in Elastic Beanstalk)

benjefferies: It returns 404 currently in master, @collisdigital you may be thinking about the `/healthcheck` end point

collisdigital: I was thinking of `/` but the problem may have been resolved since (this was 11 months ago!) where EB was reporting a healthcheck failure from Apache because Apache defaults to checking if `/` returns 200; like I say that may have been resolved so if it isn't causing a problem don't worry.

app/templates/multiple_survey.html
```diff
@@ -0,0 +1,16 @@
+{% extends 'layouts/_twocol.html' %}
+
+{% block page_title %}Information{% endblock %}
+
+{% block main %}
+  <div>
+    <h1 class="saturn">
+      {% block system_message_title %}Information{% endblock %}
```
ajmaddaford: Now that this page has it's own template, it seems unnecessary to make the heading configurable.

tests/integration/mci/test_information_page.py
```diff
@@ -5,19 +5,6 @@
 
 class TestInformationPage(IntegrationTestCase):
```
ajmaddaford: As there is no longer an `information` page, we should rename this class something like `MultipleSurveyErrorPage`

ajmaddaford (changes requested):

tests/integration/mci/test_information_page.py
benjefferies: Done

app/templates/multiple_survey.html
benjefferies: Done

app/views/session.py
ajmaddaford: Elastic Beanstalk health check is on `/status`, current pre-prod deployment returns a 404 for `/`.

ajmaddaford (approved):

iwootten (approved):

jonnyshaw89 (approved):

PR #784, opened 2016-12-20
Fixes #782. Value of relationships changed from Welsh to English.

### LINKED ISSUE #782: Census_Household_Wales_Lang: Wales text changes to english upon selection
### Expected behaviour
Must display the relation in Welsh

### Actual behaviour
Is displaying the relationship in English please find the attached screenshot

Launch Census_Household survey in Welsh Language 
Navigate to Relationship screen 
and pick a relation
### Screenshot
welsh language

samiwel:
Fixes #782.

Relationships page displays the value entered in the schema.
Since no routing is based on these values, should be able to change these to Welsh equivalent in the schema.

### How to review 
Re-test the issue.

### DIFF
--- a/app/assets/js/app/modules/relationships.js
+++ b/app/assets/js/app/modules/relationships.js
@@ -41,11 +41,14 @@ class HouseholdRelationship extends EventEmitter {
 
     let selection = this.el.querySelector('input:checked')
     if (selection) {
-      this.setRelationship(selection.value)
+      this.setRelationship(selection.id)
     }
   }
 
-  setRelationship(relationship) {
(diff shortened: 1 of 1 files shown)


collisdigital:
Can't we use the labels in the frontend ?

ajmaddaford:
I don't think we can do this, as the value we save will be in Welsh for this question, whereas for every other multi-choice question the English value will be saved. 
@hamishtaplin is there a straightforward way for the JavaScript to use the label in the 'someone is related to someone else' text?

hamishtaplin (approved):

PR #788, opened 2016-12-20
Adding logic to support unanswered questions

benjefferies:
An unanswered answer can occur when you skip over an answer using
navigation. We can set the unanswered rule to handle skip conditions or
routing rules.

### How to review 
Tests pass and solution meets requirement

### DIFF
--- a/app/data/schema/schema-v1.json
+++ b/app/data/schema/schema-v1.json
@@ -17,7 +17,8 @@
                     "enum": [
                         "equals",
                         "not equals",
-                        "contains"
+                        "contains",
+                        "not set"
                     ]
                 },
                 "value": {
@@ -25,8 +26,7 @@
                 }
             },
             "required": [
-                "condition",
-                "value"
+                "condition"
             ],
             "oneOf": [
                 {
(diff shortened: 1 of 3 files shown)


jonnyshaw89 (approved):

app/data/schema/schema-v1.json
```diff
@@ -17,7 +17,8 @@
                     "enum": [
                         "equals",
                         "not equals",
-                        "contains"
+                        "contains",
+                        "unanswered"
```
collisdigital: Would "not set" be better? Just thinking if it's a metadata rule then it might not be an answer that's being checked against.

collisdigital: Or "empty"?

benjefferies: Not sure about empty as should that include an empty string which it doesn't currently, we should probably not sure empty strings is the fix for that.
For not set we could define it as that but I guess it's just a synonym really and we should choose what our favourite [synonym](http://www.thesaurus.com/browse/unanswered) is

iwootten (approved):

PR #793, opened 2016-12-20
Removing logic to load schemas from s3

benjefferies:
The app fails to start up if you don't have the EQ_SCHEMA_BUCKET.
We don't currently utilise s3 for loading schemas therefore we can
remove the logic until the time it is needed.

### How to review 
All tests should still pass.

### DIFF
--- a/app/schema_loader/schema_loader.py
+++ b/app/schema_loader/schema_loader.py
@@ -4,10 +4,6 @@ import os
 
 from app import settings
 
-import boto3
-
-import botocore
-
 
 class SchemaNotFound(Exception):
     pass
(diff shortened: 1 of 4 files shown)


app/schema_loader/schema_loader.py
```diff
@@ -4,10 +4,6 @@
 
 from app import settings
 
-import boto3
```
jonnyshaw89, line 7: are these still defined in requirements?

jonnyshaw89 (changes requested):

app/schema_loader/schema_loader.py
benjefferies: I did initially remove but wanted to check today to see if it was still required. I overheard we were using boto for pushing logs to cloudwatch so it might still be required

iwootten: It's only imported in this file/test, so I'd think not?

iwootten: Actually, watchtower (which pushes to cloudwatch) uses botocore too, so there is a dependency there.

benjefferies: Also looks like it imports boto3

jonnyshaw89 (approved):

iwootten (approved):

PR #805, opened 2016-12-21
Add Navigation class

iwootten:

Depends on the addition of the location class in https://github.com/ONSdigital/eq-survey-runner/pull/759.

Move the navigation methods from current navigator to a new Navigation class. Both frontend sidebar navigation and path navigation can be generated independently from one another and this split allows for separation of concerns. Refactor current methods to make behaviour clearer.
    
Rename the current Navigator to PathFinder to prevent confusion and migrate tests.

### How to review 
Check if tests pass and you agree with approach.

### DIFF
--- a/app/questionnaire/navigation.py
+++ b/app/questionnaire/navigation.py
@@ -1,0 +1,170 @@
+import copy
+import logging
+
+from collections import defaultdict
+
+from app.helpers.schema_helper import SchemaHelper
+from app.questionnaire.location import Location
+from app.questionnaire.rules import evaluate_repeat
+
+logger = logging.getLogger(__name__)
+
+
+class Navigation(object):
+    """
+    Reads navigation config from the schema and returns collections of dicts that describe
+    completion status of each group
+    """
+    def __init__(self, survey_json, answer_store, metadata=None, completed_blocks=None):
+        self.survey_json = survey_json
+        self.metadata = metadata or {}
+        self.answer_store = answer_store
+        self.completed_blocks = completed_blocks or []
+
+    def build_navigation(self, current_group_id, current_group_instance):
+        """
+        Build navigation based on the current group/instance selected
+
+        :param current_group_id:
+        :param current_group_instance:
+        :return:
+        """
+        if 'navigation' not in self.survey_json:
+            return None
+
+        navigation = []
+
+        for group in filter(lambda x: 'hide_in_navigation' not in x, self.survey_json['groups']):
+            first_location = Location(group['id'], 0, group['blocks'][0]['id'])
+
+            logger.debug("Building frontend navigation for group %s", group)
+
+            repeating_rule = SchemaHelper.get_repeat_rule(group)
+
+            if repeating_rule:
+                navigation += self._build_repeating_navigation(repeating_rule, group, current_group_id,
+                                                               current_group_instance)
+            else:
+                navigation.append(self._build_single_navigation(group, current_group_id, first_location))
+
+        return navigation
+
+    def _build_repeating_navigation(self, repeating_rule, group, current_group_id, current_group_instance):
+        first_location = Location(group['id'], 0, group['blocks'][0]['id'])
+        no_of_repeats = evaluate_repeat(repeating_rule, self.answer_store)
+
+        repeating_nav = []
+
+        if repeating_rule['type'] == 'answer_count':
+            completed_id = group['completed_id'] if 'completed_id' in group else group['blocks'][-1]['id']
+            is_current_group = group['id'] == current_group_id
+            link_names = self._generate_link_names(repeating_rule)
+
+            repeating_nav = self._generate_repeated_items(link_names, first_location, completed_id, no_of_repeats,
+                                                          is_current_group, current_group_instance)
+
+        elif no_of_repeats > 0:
+            item = self._build_single_navigation(group, current_group_id, first_location)
+            repeating_nav.append(item)
+
+        return repeating_nav
+
+    def _build_single_navigation(self, group, current_group_id, first_location):
+        completed_id = group['completed_id'] if 'completed_id' in group else group['blocks'][-1]['id']
+        default_highlight = self._should_highlight(group, current_group_id)
+        is_completed = self._is_complete_non_repeating(completed_id)
+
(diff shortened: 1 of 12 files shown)


app/questionnaire/navigation.py
```diff
@@ -0,0 +1,151 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class Navigation(object):
+    """
+    Reads navigation config from the schema and returns collections of dicts that describe
+    completion status of each group
+    """
+    def __init__(self, survey_json, metadata=None, answer_store=None, completed_blocks=None):
```
benjefferies: answer_store should be a mandatory argument. When would we ever use navigation without taking into account the users answers similarly for completed blocks. We are just making more scenarios we should have unit tests for by making things optional just because we can

app/questionnaire/navigation.py
```diff
@@ -0,0 +1,151 @@
...
+            is_current_group = group['id'] == current_group_id
+
+            default_highlight = self._should_highlight(group, current_group_id)
+
+            logger.debug("Building frontend navigation for group %s", group)
+
+            repeating_rule = SchemaHelper.get_repeat_rule(group)
+            completed_id = group['completed_id'] if 'completed_id' in group else group['blocks'][-1]['id']
+
+            if repeating_rule:
```
benjefferies, line 44: This method would be more readable if we broke it out into a private method _append_repeating_navigation or navigation += _build_repeating_navigation

app/questionnaire/navigation.py
```diff
@@ -0,0 +1,151 @@
...
+        return navigation
+
+    @staticmethod
+    def _should_highlight(group, current_group_id):
+        is_current_group = group['id'] == current_group_id
+        current_group_in_highlight = 'highlight_when' in group and current_group_id in group['highlight_when']
+
+        return (current_group_in_highlight and not is_current_group) or is_current_group
+
+    def _is_complete(self, group_instance, completed_id, repeating=False):
```
benjefferies: It is not clear what the repeating boolean flag is for as it's not documented. We are dealing with repeating groups [here](https://github.com/ONSdigital/eq-survey-runner/pull/805/files#diff-58141777c4c9157703a13f01a64832b5R60) yet repeating is set to false

app/questionnaire/path_finder.py
```diff
@@ -55,7 +52,7 @@ def build_path(self, blocks, this_location):
             if this_location.block_id in self.CLOSING_INTERSTITIAL_PATH:
                 return path
 
-            block_index = Navigator._block_index_for_location(blocks, this_location)
+            block_index = PathFinder._block_index_for_location(blocks, this_location)
```
benjefferies, line 55: _block_index_for_location shouldn't have a private marker on it anymore I guess

benjefferies (changes requested):
Great effort in being proactive and refactoring this. It was needed and should have been caught earlier in PR!

app/questionnaire/navigation.py
iwootten: Originally it was, but I wanted to preserve the argument order that Navigator (now PathFinder) has and weirdly metadata comes before it there. Given metadata is optional and a kwarg can't precede an arg, hence it's ended up this way.

Additionally, originally the tests were modifying the content of the store without ever having created one, so this was a solution for that. I've already refactored the tests, so may well be able to resolve this now.

app/questionnaire/path_finder.py
iwootten: It should as I don't believe anything uses it other than PathFinder, honestly I think this test should be removed as it's checking a private method.

app/questionnaire/navigation.py
iwootten: This is a flag to decide whether the item being marked as complete is actually a repeating item. Maybe therefore this should actually be two methods making the check rather than one, seeing as how we should be fluent about this and all :)

app/questionnaire/navigation.py
iwootten: Unsure of what you're referring to here?

benjefferies: Smaller methods are easier to follow than larger methods. Line 51 to 64 being refactoring into a method called _build_repeating_navigation

and then we'd have something like 

```
    def build_navigation(self, current_group_id, current_group_instance):
        """
        Build navigation based on the current group/instance selected

        :param current_group_id:
        :param current_group_instance:
        :return:
        """
        if 'navigation' not in self.survey_json:
            return None

        navigation = []

        for group in filter(lambda x: 'hide_in_navigation' not in x, self.survey_json['groups']):
            first_location = Location(group['id'], 0, group['blocks'][0]['id'])
            is_current_group = group['id'] == current_group_id

            default_highlight = self._should_highlight(group, current_group_id)

            logger.debug("Building frontend navigation for group %s", group)

            completed_id = group['completed_id'] if 'completed_id' in group else group['blocks'][-1]['id']

            if repeating_rule:
                navigation += _build_repeating_navigation(...)

        return navigation

    def _build_repeating_navigation(...):
        # repeating logic

File: app/questionnaire/navigation.py (hunk 1 above)
Comment (jonnyshaw89) [reply app/questionnaire/navigation.py]: Deja Vu of a previous discussion here.

File: app/questionnaire/navigation.py [hunk 5]
```diff
@@ -0,0 +1,170 @@
...
+
+            logger.debug("Building frontend navigation for group %s", group)
+
+            repeating_rule = SchemaHelper.get_repeat_rule(group)
+
+            if repeating_rule:
+                navigation += self._build_repeating_navigation(repeating_rule, group, current_group_id,
+                                                               current_group_instance)
+            else:
+                navigation.append(self._build_single_navigation(group, current_group_id, first_location))
```
benjefferies, line 48: Not a major one but we are appending differently, might be good to be consistent here

benjefferies (approved):

app/questionnaire/navigation.py
iwootten: This can't be consistent as one returns a list the other returns a single item as a dict. += tries to merge the dict as a result and wipes it.

It is possible to cast the single item to a list and then add it as a list using +=, but that approach involves an unnecessary cast.

jonnyshaw89 (approved):

PR #806, opened 2016-12-22
Add equality check in case of hash collision

benjefferies:
[trello card](https://trello.com/c/lzYc3wLd/335-questionnaire-store-has-changed-could-result-in-hash-collisions)
Because we were relying on hash values to check equality we could
appear to have the same value but the actual values differ if we had a
hash collision.

Scenario
A user changes/adds new answers which has a hash calculated to the same
value as their last answer state was. They choose to save and continue,
on checking whether the answers have changed, the hash values are the
same meaning we class it as not being changed and therefore we do not
save the answers.

We now check the hash and then the equality if the hash appears to be
equal.

### How to review 
The tests pass


app/data_model/questionnaire_store.py
```diff
@@ -49,9 +50,8 @@ def get_json(self):
         return json.dumps(self.get_data())
 
     def has_changed(self):
-        current_hash = hash(self.get_json())
-
-        return current_hash != self.initial_hash
+        data = self.get_json()
+        return hash(data) != self.hash_on_load or data != self.data_on_load
```
collisdigital: Is doing the hash needed then, is it any faster?

benjefferies: I can test it but probably not worth the effort. I imagine it will be marginally quicker but as we check if it has_changed once per request it's probably not needed

app/data_model/questionnaire_store.py
```diff
@@ -36,7 +35,7 @@ def __init__(self, user_id, user_ik):
             if 'COMPLETED_BLOCKS' in data:
                 self.completed_blocks = data['COMPLETED_BLOCKS']
 
-        self.initial_hash = hash(self.get_json())
+        self.data_on_load = self.get_json()
```
collisdigital: Change Line 27 to return the raw JSON string then do the json.loads() in here, call it once for the 'live' data to edit and once for the data_on_load; then on has_changed check the two Python dicts

collisdigital: In addition, since what this is doing is serialising and de-serialising the database raw data, create two functions with those names and do the json.loads and dumps in there (and merge in the get_data call?)

collisdigital:
I haven't reviewed but can you confirm the user_ik is only stored in the user's cookies on their browser and never in the database? It must be provided on every request from the user.

app/data_model/questionnaire_store.py
```diff
@@ -2,30 +2,21 @@
...
-        else:
-            raise ValueError("No user_id or user_ik found in session")
-
-        self.storage = get_storage()
-
-        if self.storage.has_data(self.user_id):
-            logger.debug("User %s has previous data loading", user_id)
-            data = self.storage.get(self.user_id, self.user_ik)
+        if self.storage.has_questionnaire_state():
+            self.json_data = self.storage.get_user_data()
```
collisdigital: My idea was to make 'initial_data = json.loads(json_data)' and then compare the Python dicts before saving rather than raw strings. And to wrap the loads and dumps calls in 'serialise' and 'deserialise' functions to make it clear what's happening. This would save json.dumps each time the check for changes occurs and also avoid doing comparisons on original JSON string and the newly serialised one (which I wasn't convinced would always serialise to identical strings even if the actual data was the same ~ e.g. The order of JSON objects in the string output)

benjefferies: Have a look at https://gist.github.com/benjefferies/18f71640503794837913c9e895230465#file-questionnaire_store-py-L30, feel free to change to what your suggesting

app/data_model/questionnaire_store.py.orig
```diff
@@ -0,0 +1,61 @@
...
+
+            if 'METADATA' in data:
+                self.metadata = data['METADATA']
+
+            if 'ANSWERS' in data:
+                self.answer_store.answers = data['ANSWERS']
+
+            if 'COMPLETED_BLOCKS' in data:
+                for location_dict in data['COMPLETED_BLOCKS']:
+                    location = Location(
```
collisdigital: Could do something like this (making Location just a wrapper around a dict):

```python
class Location:
    def __init__(..., wrap_dict=None):
         if not wrap_dict:
             self._location = { 'group_id': ..., 'block_id': ...}
        else:
             self._location = wrap_dict

    @property
    def group_id(self):
        return self._location.get('group_id')

    @group_id.setter
    def group_id(self, value):
        self._location['group_id'] = value
....
```

Mean this code could be:

```python
self.completed_blocks = [Location(wrap_dict=d) for d in data['COMPLETED_BLOCKS']]
```

This would also mean that each Location object isn't holding copies of the values but actual references to original dict passed in, it's just wrapping it (you may not agree this is better)

Just an idea...

collisdigital: Of course you could just leave Location as-is but allow init from a dict like above to make it easier to create one

app/data_model/questionnaire_store.py.orig
```diff
@@ -0,0 +1,61 @@
+import json
+import logging
+
+from app.data_model.answer_store import AnswerStore
+<<<<<<< HEAD
```
iwootten: Guessing this file wasn't meant to be committed?

app/storage/encrypted_questionnaire_database_storage.py
```diff
@@ -23,28 +22,29 @@ def generate_key(user_id, user_ik, pepper=settings.EQ_SERVER_SIDE_STORAGE_ENCRYP
     return to_bytes(cek)
 
 
-class EncryptedStorage(DatabaseStorage):
+class EncryptedQuestionnaireDatabaseStorage(QuestionnaireDatabaseStorage):
```
iwootten: I actually think the original class names were better. Could be QuestionnaireStorage and EncryptedQuestionnaireStorage and change QuestionnaireStore to something else?

app/storage/questionnaire_state_database_storage.py
```diff
@@ -0,0 +1,57 @@
+import logging
+
+from app.data_model.database import QuestionnaireState, commit_or_rollback
+from app.data_model.database import db_session
+
+logger = logging.getLogger(__name__)
+
+
+class QuestionnaireDatabaseStorage:
```
iwootten: Seems like the class name is different to the file here.

tests/app/storage/test_encrypted_storage.py
```diff
@@ -1,11 +1,19 @@
-from app.storage.encrypted_storage import EncryptedStorage, generate_key
+from app.storage.encrypted_questionnaire_database_storage import EncryptedQuestionnaireDatabaseStorage, generate_key
 
 
 import unittest
 
 
 class TestEncryptedStorage(unittest.TestCase):
```
iwootten: This test class no longer is named correctly against the class it's testing.

app/data_model/questionnaire_store.py
```diff
@@ -2,62 +2,54 @@
...
+    def delete_questionnaire_state(self):
+        self._storage.delete_questionnaire_state()
+        self._set_data(data={})
+        self._initial_data = {}
 
-    def save(self):
-        data = self.get_data()
-        logger.debug("Saving user data %s for user id %s", data, self.user_id)
-        self.storage.store(data=data, user_id=self.user_id, user_ik=self.user_ik)
+    def add_or_update_questionnaire_state(self):
```
iwootten: Why not just 'save' or 'add_or_update'?

app/data_model/questionnaire_store.py
```diff
@@ -2,62 +2,54 @@
...
     def has_changed(self):
-        current_hash = hash(self.get_json())
-
-        return current_hash != self.initial_hash
+        return self._initial_data != self._get_data()
 
-    def delete(self):
-        logger.debug("Deleting questionnaire data for %s", self.user_id)
-        self.storage.delete(self.user_id)
+    def delete_questionnaire_state(self):
```
iwootten: Why the name change?

app/storage/questionnaire_state_database_storage.py
```diff
@@ -0,0 +1,57 @@
...
+
+        with commit_or_rollback(db_session):
+            db_session.add(questionnaire_state)
+
+    def get_user_data(self):
+        logger.debug("Loading questionnaire state for user %s", self.user_id)
+        questionnaire_state = QuestionnaireState.query.filter(QuestionnaireState.user_id == self.user_id).first()
+        return questionnaire_state.get_data()
+
+    def _get_questionnaire_state(self):
```
iwootten: Surely we've qualified the type of data by the class we're using, do we really need the questionnaire_state() suffixed to each method?

iwootten (changes requested):

app/data_model/questionnaire_store.py.orig
benjefferies: Done

app/data_model/questionnaire_store.py
benjefferies: For explicitness initially but can be reverted

app/data_model/questionnaire_store.py
benjefferies: I think add_or_update is more explicit

app/data_model/questionnaire_store.py.orig
benjefferies: Should we just treat location as an object instead of a object and a dict

app/storage/encrypted_questionnaire_database_storage.py
benjefferies: Done

app/storage/questionnaire_state_database_storage.py
benjefferies: Done

app/storage/questionnaire_state_database_storage.py
benjefferies: Possibly not was just aiming to be more explicit

tests/app/storage/test_encrypted_storage.py
benjefferies: Done

benjefferies:
@collisdigital the user_ik is not stored in the database. It is only used to generate the encryption key

iwootten (approved):

app/storage/questionnaire_storage.py
```diff
@@ -0,0 +1,50 @@
+import logging
+
+from app.data_model.database import QuestionnaireState, commit_or_rollback
+from app.data_model.database import db_session
+
+logger = logging.getLogger(__name__)
+
+
+class QuestionnaireStorage:
```
jonnyshaw89, line 9: Are we storing the Questionnaire?

jonnyshaw89 (changes requested):

app/storage/questionnaire_storage.py
benjefferies: Depends what you define the questionnaire as? What's your suggestion?

jonnyshaw89: To me the name implies storing the questions rather than the values in the answers?

benjefferies: It stores quite a lot to be fair, it stores the users metadata, completed block and answers. If you can think of an apt name for that I am happy to change

jonnyshaw89: I suppose given that we have QuestionnaireStore holding the data then this name is consistent.

jonnyshaw89 (approved):

PR #810, opened 2016-12-30
Add q_codes into checkbox options and update converter

LJBabbage:
To add q_codes into the options of checkboxes so the converter can map the form data to the right q_code before sending to SDX. Needed for UKIS

### How to review 
Using test_checkbox.json, put in debug statement in converter.py just above the data return, to expose the data

i.e
```
  if option:
      if option['value'] == 'other':
           data[option['q_code']] = CheckboxHelper.find_other_value(value, item.options) or 'other'
      else:
          data[option['q_code']] = user_answer
  
  logger.error(data)
  return data
```

Run the application and fill out the form as normal, check your console to see what data is being produced. It should map your answers to the correct q_codes for the first questions checkboxes, if you use 'other' it will map the answer you provided to the q_code of other.

### DIFF
--- a/app/data/test_checkbox.json
+++ b/app/data/test_checkbox.json
@@ -41,34 +41,40 @@
                                             "options": [
                                                 {
                                                     "label": "Cheese",
-                                                    "value": "Cheese"
+                                                    "value": "Cheese",
+                                                    "q_code": 1
                                                 },
                                                 {
                                                     "label": "Ham",
-                                                    "value": "Ham"
+                                                    "value": "Ham",
+                                                    "q_code": 2
                                                 },
                                                 {
                                                     "label": "Pineapple",
-                                                    "value": "Pineapple"
+                                                    "value": "Pineapple",
+                                                    "q_code": 3
                                                 },
                                                 {
                                                     "label": "Tuna",
-                                                    "value": "Tuna"
+                                                    "value": "Tuna",
+                                                    "q_code": 4
                                                 },
                                                 {
                                                     "label": "Pepperoni",
-                                                    "value": "Pepperoni"
+                                                    "value": "Pepperoni",
+                                                    "q_code": 5
                                                 },
                                                 {
                                                     "label": "Other",
                                                     "value": "other",
+                                                    "q_code": 6,
                                                     "description": "Choose any other topping",
                                                     "other": {
                                                       "label": "Please specify other"
                                                     }
                                                 }
                                             ],
-                                            "q_code": "20",
+                                            "q_code": "",
                                             "type": "Checkbox",
                                             "validation": {
                                                 "messages": {}
--- a/app/submitter/converter.py
+++ b/app/submitter/converter.py
@@ -4,6 +4,8 @@ from collections import OrderedDict
 from datetime import datetime, timezone
 
 from app import settings
+from app.helpers.multiple_choice_helper import MultipleChoiceHelper
+from app.schema.answers.checkbox_answer import CheckboxAnswer
 
 logger = logging.getLogger(__name__)
 
@@ -19,7 +21,6 @@ class DataVersionError(Exception):
 def convert_answers(metadata, questionnaire, answer_store, routing_path):
     """
     Create the JSON answer format for down stream processing
-
     :param metadata: metadata for the questionnaire
     :param questionnaire: the questionnaire schema
     :param answer_store: the users answers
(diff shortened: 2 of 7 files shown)


tests/app/submitter/test_converter.py
```diff
@@ -235,6 +236,136 @@ def test_converter_raises_runtime_error_for_unsupported_version(self):
...
+                                {
+                                    "label": "Other",
+                                    "value": "other",
+                                    "q_code": "4",
+                                    "description": "Choose any other flavour",
+                                    "other": {
+                                      "label": "Please specify other"
+                                    }
+                                }
+                                            ]
```
benjefferies: Formatting looks odds

tests/app/submitter/test_converter.py
```diff
@@ -235,6 +236,136 @@ def test_converter_raises_runtime_error_for_unsupported_version(self):
...
```
benjefferies: Formatting looks odd

app/data/test_checkbox.json
```diff
@@ -41,34 +41,40 @@
                                             "options": [
                                                 {
                                                     "label": "Cheese",
-                                                    "value": "Cheese"
+                                                    "value": "Cheese",
+                                                    "q_code": 1
```
benjefferies, line 45: We should update the schema to contain q_codes in options

benjefferies (changes requested):

app/submitter/converter.py
```diff
@@ -82,14 +83,30 @@ def convert_answers_to_data(answer_store, questionnaire_schema):
     for answer in answer_store.answers:
         item = questionnaire_schema.get_item_by_id(answer['answer_id'])
         value = answer['value']
+
         if item is not None and value is not None:
-            if item.code not in data:
-                data[item.code] = _encode_value(value)
+            if not isinstance(item, CheckboxAnswer) or any('q_code' not in option for option in item.options):
```
iwootten, line 88: Might be better to reverse the cases and remove the 'not's here?

app/submitter/converter.py
```diff
@@ -82,14 +83,30 @@ def convert_answers_to_data(answer_store, questionnaire_schema):
...
+
+                if item.code not in data:
```
iwootten: I think this if/else would be better broken into it's own method returning the data for a particular item.code. Maybe something like _get_answer_data?

app/submitter/converter.py
```diff
@@ -82,14 +83,30 @@ def convert_answers_to_data(answer_store, questionnaire_schema):
...
+                    if not isinstance(data[item.code], list):
+                        list_answers = [data[item.code]]
+                        data[item.code] = list_answers
+                    data[item.code].append(_encode_value(value))
             else:
-                if not isinstance(data[item.code], list):
-                    list_answers = [data[item.code]]
-                    data[item.code] = list_answers
-                data[item.code].append(_encode_value(value))
+                for user_answer in value:
```
iwootten: Similarly, I think this whole for loop would be better broken into it's own method which returns checkbox data for a item.code. Like _get_checkbox_answer_data?

Breaking them both out should make it very clear what's going on here.

iwootten (changes requested):

app/submitter/converter.py
LJBabbage: I did that initially, however I changed it back for a couple of reason. 1. most surveys won't use q_codes in their checkboxes and therefore I felt this was more efficient 2. I had concerns over a missed q_code in a checkbox when the others had them. I originally had a safety net in place to catch this, however the other box turned to non mandatory made a mess of this, I therefore decided to make sure every option had a q_code and reverse the condition

app/submitter/converter.py
LJBabbage: Agree for readability, however I don't think I can get rid of the fact I would be passing the array, modifying it and passing it back, which I want to avoid. The ordered dict (data) has to sit outside the for loop for **answer_store.answers**, and the problem is there are checks in the later code to make sure the data doesn't already exist **item.code not in data**, and if it does exist what type it is **isinstance(data[item.code], list)**

app/data/test_checkbox.json
LJBabbage: Done, but obviously its not in this file but in schema-v1.json

app/submitter/converter.py
iwootten: Unless I've misunderstood, there's no need to modify the original list, you just need to inspect it and return a variable based on it's contents. Something like:

    def _get_answer_data(original_data, item_code):
        if item_code in original_data:
            data_to_return = original_data[item_code]
            if not isinstance(data_to_return, list):
                 data_to_return = [data_to_return]
            data_to_return.append(_encode_value(value))

            return data_to_return
        return _encode_value(value)


    data[item.code] = _get_answer_data(data, item.code)

app/submitter/converter.py
```diff
@@ -82,17 +83,47 @@ def convert_answers_to_data(answer_store, questionnaire_schema):
...
+        data_to_return = original_data[item_code]
+        if not isinstance(data_to_return, list):
+            data_to_return = [data_to_return]
+        data_to_return.append(_encode_value(value))
+
+        return data_to_return
+    return _encode_value(value)
+
+
+def _convert_checkboxes_with_q_codes_to_data(item, value):
```
iwootten: This method and the associated var within it seem like they have very verbose names now. I wonder if _get_checkbox_answer_data() might be more in line with the other method?

benjefferies (approved):

app/submitter/converter.py
LJBabbage: It is a bit of a mouth full, but it is doing exactly what it says. I felt it would be slightly misleading if we used _get_checkbox_answer_data() as it would suggest all checkboxes use this, they don't its only the ones with q_codes in their options (UKIS)

iwootten: Fair enough, but _convert on the method here implies a conversion of the original data, whereas it's just populating a new ordered dict. 

Are we using any attribute on item within this method other than options? Could this be qualified further through better naming of args - _get_options_data(options_with_qcode, value) for example.

LJBabbage: It can only happen for checkboxes due to multiple answers being selected, everything else maps directly to the q_code in the answer and doesn't need a q_code in the options. Although I think the original name is clearer I will change it to _get_checkbox_answer_data(options_with_qcode, value) and the var to checkbox_answer_data

iwootten (approved):

LJBabbage:
Squashed rebased and ready to merge

PR #814, opened 2017-01-04
Adding missing properties to json schema

benjefferies:
JSON schema is the source of truth when configuring a questionnaire so
this should be up to date.

* Added additionalProperties false configuration to highlight when new
  properties haven't been added.
* Removing empty optional configuration
* Removing unused eq_id

### How to review 
Tests pass
check schema is update to date

### DIFF
--- a/app/data/0_rogue_one.json
+++ b/app/data/0_rogue_one.json
@@ -10,7 +10,6 @@
     "introduction": {
         "description": ""
     },
-    "eq_id": "1",
     "groups": [
         {
             "blocks": [
@@ -59,12 +58,10 @@
                                     "description": "",
                                     "id": "4bfa0229-0ce8-4190-817b-535f55ad2817",
                                     "title": "",
-                                    "type": "General",
-                                    "validation": []
+                                    "type": "General"
                                 }
                             ],
-                            "title": "Who do you want to know more about?",
-                            "validation": []
+                            "title": "Who do you want to know more about?"
                         }
                     ],
                     "routing_rules":[
@@ -121,8 +118,7 @@
 
                         }
                     ],
-                    "title": "",
-                    "validation": []
+                    "title": ""
                 },
                 {
                     "type": "questionnaire",
@@ -159,12 +155,10 @@
                                     "description": "",
                                     "id": "574e5d5e-36b4-44ea-b8f6-09d744552241",
                                     "title": "Do you like this page?",
-                                    "type": "General",
-                                    "validation": []
+                                    "type": "General"
                                 }
                             ],
-                            "title": "{{answers.character}}",
-                            "validation": []
+                            "title": "{{answers.character}}"
                         }
                     ],
                     "routing_rules":[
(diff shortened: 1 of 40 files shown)


iwootten (changes requested):
Doesn't removing eq_id from the schema negate the need for it in the url?

benjefferies:
@iwootten yes that's correct but we do not use it for this at the moment. I think the bigger question we need to look at whether we need have the eq_id at all but this will mean discussing amongst the wider teams

jonnyshaw89:
@benjefferies you say we need to discuss wether or not we need it at all. but isn't removing it from the schema basically removing it completely? or does it also come in metadata?

benjefferies:
@jonnyshaw89 It's is passed in the URL and metadata I believe

jonnyshaw89:
@benjefferies the only place i can see it used not from the url is in check_survey_state. but that could quite easily be grabbed from the metadata instead. so ill 2nd @iwootten on thinking we should remove it from the URL too.

benjefferies:
@jonnyshaw89 @iwootten https://trello.com/c/5wDJxaAi/338-remove-eq-id-from-url for the discussion above.

collisdigital:
eq-id was supposed to replace form_type and survey code; i.e. each questionnaire would have a unique id; but this hasn't happened yet. There is a limitation with the current form_type survey code approach in that multiple questionnaires can have the same combination (and some do!) However, there is also the concept of versions of the same questionnaire (i.e. if we update the schema month on month, is it a new id or the same id with a new version).

iwootten:
@collisdigital if that's the case, then it sounds like we should remove form type and survey code from the url. There doesn't seem to be a need for them all, unless they have been kept for ux purposes.

collisdigital:
@iwootten aren't they used to know if the respondent has multiple surveys open at the moment?

benjefferies:
We have two places where we are defining a eq_id:
1. The first part of the schema file name **census**_household.json
2. As a property in the schema

The property is never referenced in the schema for that reason I removed it in this PR.

As for the URL structure we shouldn't need the eq_id, form_type or survey code as this is stored in the users session. The only place we make use of any of the eq_id, form_type_survey code via the url is to check that the user is opening multiple surveys. I think this can be solved at a lower level when the users are initially authenticated or fix the root of the problem

collisdigital:
@benjefferies the problem was users who had a tab/windows open after starting a survey then went back and launched a new tab/window for a different survey or period (e.g. March and April); if they then return to the original tab the survey (March) they have open is no longer valid but could submit answers for the wrong survey (April) (if the answer id values were the same, which they can be/are across periods).

benjefferies:
@collisdigital I guess that's because we are storing one record of data per user and that data isn't being qualified for us to be able to determine the context of the user data. Should we be looking at fixing that rather than limiting the number of open surveys?

iwootten (approved):

iwootten:
Think I'm happy for this to go forward, we don't use the id in the schema, so it need not be defined. Whether the url design needs to change is another matter.

jonnyshaw89 (approved):

PR #817, opened 2017-01-04
Removing eq id from url

benjefferies:
The eq_id is passed in as the payload of the session log in and is it
stored against the users session therefore we do not have to pass it on
every request.

[trello](https://trello.com/c/5wDJxaAi/338-remove-eq-id-from-url)

### How to review 
Tests pass
No downstream impact on other systems


app/views/questionnaire.py
```diff
@@ -270,42 +271,35 @@ def _delete_user_data():
...
                        collection_id=collection_id)
 
 
-def _check_same_survey(eq_id, form_type, collection_id):
-    metadata = get_metadata(current_user)
-    current_survey = eq_id + form_type + collection_id
-    metadata_survey = metadata["eq_id"] + metadata["form_type"] + metadata["collection_exercise_sid"]
-    if current_survey != metadata_survey:
+def _check_same_survey(metadata, form_type, collection_id):
+    if form_type != metadata["form_type"] or collection_id != metadata["collection_exercise_sid"]:
```
jonnyshaw89, line 302: I like

jonnyshaw89 (changes requested):

jonnyshaw89:
Pretty sure i made more comments here?

benjefferies:
I was a bit confused when I saw I like and then a changes requested. I haven't noticed any other comments

jonnyshaw89:
I think i was questioning if form_type was also still needed in the url?

collisdigital:
form_type identifies the questionnaire so I'd say yes

jonnyshaw89:
@collisdigital but as with eq_id this is supplied in the metadata. so is purely in the url for show?

collisdigital:
I thought it was used to check if they'd launched a second survey; I think I'll leave it to you as I probably don't have the latest understanding of how that works.

benjefferies:
Closing as it has been moved to the backlog

PR #819, opened 2017-01-05
Backfilling jinja filter tests

benjefferies:
[trello](https://trello.com/c/csuYulm1/99-tests-for-jinja-filters)
Adding missing tests for jinja filters

### How to review 
All tests pass

### DIFF
--- a/app/jinja_filters.py
+++ b/app/jinja_filters.py
@@ -21,9 +21,11 @@ def format_percentage(value):
 
 @evalcontextfilter
 @blueprint.app_template_filter()
-def nl2br(context, value):
-    _paragraph_re = re.compile(r'(?:\r\n|\r|\n){2,}')
-    result = u'\n\n'.join(u'<p>%s</p>' % p.replace('\n', '<br>\n') for p in _paragraph_re.split(escape(value)))
+def format_multilined_string(context, value):
+    escaped_value = escape(value)
+    new_line_regex = r'(?:\r\n|\r|\n)+'
+    value_with_line_break_tag = re.sub(new_line_regex, '<br>', escaped_value)
+    result = '<p>{}</p>'.format(value_with_line_break_tag)
 
     if context.autoescape:
         result = Markup(result)
--- a/tests/app/test_jinja_filters.py
+++ b/tests/app/test_jinja_filters.py
@@ -1,7 +1,9 @@
 from datetime import datetime
 from unittest import TestCase
 
-from app.jinja_filters import format_date, format_percentage
+from mock import Mock
+
+from app.jinja_filters import format_date, format_currency, format_multilined_string, format_percentage
 from app.jinja_filters import format_str_as_date_range
 from app.jinja_filters import format_str_as_date
 from app.jinja_filters import format_str_as_month_year_date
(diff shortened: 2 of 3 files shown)


tests/app/test_jinja_filters.py
```diff
@@ -10,6 +12,37 @@
...
+    def test_new_line_to_break(self):
+        # Given
+        new_line = 'this is on a\nnew line'
+        context = Mock()
+        context.autoescape = False
+
+        # When
+        format_value = nl2br(context, new_line)
+
+        self.assertEquals(format_value, '<p>this is on a<br>\nnew line</p>')
```
jonnyshaw89: The name of this filter implies that is replaces the \n with a <br>?
But this test checks that there is a <br> before the \n

tests/app/test_jinja_filters.py
```diff
@@ -10,6 +12,37 @@
...
+    def test_new_line_to_break_auto_escape(self):
+        # Given
+        new_line = '<'
+        context = Mock()
+        context.autoescape = True
+
+        # When
+        format_value = nl2br(context, new_line)
+
+        self.assertEquals(str(format_value), '<p>&lt;</p>')
```
jonnyshaw89, line 77: the adding of <p> tags also seems a bit implied

jonnyshaw89 (changes requested):

tests/app/test_jinja_filters.py
benjefferies: I don't think it has any impact but is confusing so will change the implementation

tests/app/test_jinja_filters.py
benjefferies: Is it clearer with the new method name?

jonnyshaw89: Much

jonnyshaw89 (approved):

app/jinja_filters.py
```diff
@@ -21,9 +21,9 @@ def format_percentage(value):
 
 @evalcontextfilter
 @blueprint.app_template_filter()
-def nl2br(context, value):
+def format_multilined_string(context, value):
     _paragraph_re = re.compile(r'(?:\r\n|\r|\n){2,}')
-    result = u'\n\n'.join(u'<p>%s</p>' % p.replace('\n', '<br>\n') for p in _paragraph_re.split(escape(value)))
+    result = u'\n\n'.join(u'<p>%s</p>' % p.replace('\n', '<br>') for p in _paragraph_re.split(escape(value)))
```
samiwel: It might be correct, but why are we joining with `u'\n\n'`? Isn't this method replacing \n with `<br>`?

jonnyshaw89: i think it maintains the line breaks in the source code but also added the br so it appears correct when rendered?

jonnyshaw89: So maybe my previous comment about removing the \n is incorrect?
Depends if we care how the source looks?

benjefferies: As it's generated source then we probably don't care so I'll remove

app/jinja_filters.py
```diff
@@ -21,9 +21,9 @@ def format_percentage(value):
 
 @evalcontextfilter
 @blueprint.app_template_filter()
-def nl2br(context, value):
+def format_multilined_string(context, value):
     _paragraph_re = re.compile(r'(?:\r\n|\r|\n){2,}')
-    result = u'\n\n'.join(u'<p>%s</p>' % p.replace('\n', '<br>\n') for p in _paragraph_re.split(escape(value)))
+    result = u''.join(u'<p>%s</p>' % p.replace('\n', '<br>') for p in _paragraph_re.split(escape(value)))
```
jonnyshaw89: can probably remove the join completely?

jonnyshaw89 (changes requested):

app/jinja_filters.py
benjefferies: I have changed this a lot now as I think there was a bug in the code

jonnyshaw89 (approved):

samiwel (approved):
LGTM

PR #820, opened 2017-01-05
Enabled Gzip compression on Elastic Beanstalk

jonnyshaw89:
Have enabled Gzip compression on Elastic Beanstalk
**Before**
screen shot 2017-01-05 at 12 16 14
**After**
screen shot 2017-01-05 at 13 07 28

### DIFF
--- a/.ebextensions/custom-apache.config
+++ b/.ebextensions/custom-apache.config
@@ -1,3 +1,5 @@
 container_commands:
-  add_appache_conf:
+  add_apache_conf:
     command: "cp app-httpd.conf /etc/httpd/conf.d"
+  add_apache_mod_deflate:
+    command: "cp .ebextensions/enable_mod_deflate.conf /etc/httpd/conf.d/enable_mod_deflate.conf"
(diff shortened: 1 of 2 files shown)


hamishtaplin (changes requested):
We should add webfonts to the list as per this answer on SO http://stackoverflow.com/questions/14005758/gzip-compression-htaccess-deflate

jonnyshaw89:
@hamishtaplin from a quick google we shouldn't compress the web fonts are they are already compressed. In the same way we don't for images.

hamishtaplin:
@jonnyshaw89 Got any links? It's standard practise to compress webfonts (maybe not WOFF) as you can see in the (widely used) HTML5 boilerplate .htaccess: https://github.com/h5bp/html5-boilerplate/blob/master/dist/.htaccess#L756

jonnyshaw89:
@hamishtaplin https://www.tunetheweb.com/performance/gzip/
"but don't compress content which is already compressed including images and woff and woff2 fonts"

benjefferies (approved):

ajmaddaford:
@jonnyshaw89 The link you posted mentions compressing fonts that aren't woff or woff2

jonnyshaw89:
@ajmaddaford i had a chat with @hamishtaplin and as we arn't using these formats and our fonts are going onto a CDN i didn't see the point in adding. But i can if you see a benefit?

hamishtaplin (approved):

PR #822, opened 2017-01-06
Add save and come back functionality

LJBabbage:
Adding the save and come back functionality on all surveys. Mandatory validation should be skipped if save and come back is clicked, however validation errors (i.e. text in a number) should be caught. A block should not be marked as completed if save and come back is clicked, meaning the user gets returned to the same block on authentication.

### How to review 
1. Record your collection ID and user name, complete half of a survey without any potential validation errors and click close and come back (make sure data is on the page), confirm it goes to the correct page(sign-out.html) Now re-authenticate with the same details and make sure it goes to the last place with the data you entered.
2. Repeat the above but leave a mandatory question (1_0102 can be used on the first 2 questions)blank before clicking save and come back , make sure it goes to the sign out page
3. Repeat the above but now generate a validation error, like text in a number field. Make sure a validation error is generated and must be removed/corrected before save and come back later works.
4. Once save and continue has been used make sure you can't get back into the survey without re-authenticating, you should get a 401. You could do this by using back after seeing the sign-out page and trying to refresh.

### DIFF
--- a/app/questionnaire/questionnaire_manager.py
+++ b/app/questionnaire/questionnaire_manager.py
@@ -28,11 +28,10 @@ class QuestionnaireManager(object):
         self._schema = schema
         self.block_state = None
 
-    def validate(self, location, post_data):
+    def validate(self, location, post_data, skip_mandatory_validation=False):
 
         self.build_block_state(location, post_data)
-
-        return location.is_interstitial() or self.block_state.schema_item.validate(self.block_state)
+        return location.is_interstitial() or self.block_state.schema_item.validate(self.block_state, skip_mandatory_validation)
 
     def validate_all_answers(self):
 
@@ -49,14 +48,22 @@ class QuestionnaireManager(object):
         return True, None
 
     def update_questionnaire_store(self, location):
-        # Store answers in QuestionnaireStore
-        questionnaire_store = get_questionnaire_store(current_user.user_id, current_user.user_ik)
-
-        for answer in self.get_state_answers(location.block_id):
-            questionnaire_store.answer_store.add_or_update(answer.flatten())
+        questionnaire_store = self._add_update_answer_store(location)
 
         if location not in questionnaire_store.completed_blocks:
             questionnaire_store.completed_blocks.append(location)
+
+    def update_questionnaire_store_save_sign_out(self, location):
+        questionnaire_store = self._add_update_answer_store(location)
+
+        if location in questionnaire_store.completed_blocks:
+            questionnaire_store.completed_blocks.remove(location)
+
+    def _add_update_answer_store(self, location):
+        questionnaire_store = get_questionnaire_store(current_user.user_id, current_user.user_ik)
+        for answer in self.get_state_answers(location.block_id):
+            questionnaire_store.answer_store.add_or_update(answer.flatten())
+        return questionnaire_store
 
     def process_incoming_answers(self, location, post_data):
         logger.debug("Processing post data for %s", location)
--- a/app/schema/questions/date_range_question.py
+++ b/app/schema/questions/date_range_question.py
@@ -7,22 +7,25 @@ class DateRangeQuestion(Question):
     def __init__(self):
         super().__init__()
 
-    def validate(self, state):
-        is_valid = super().validate(state)
+    def validate(self, state, skip_mandatory_validation):
+
+        is_valid = super().validate(state, skip_mandatory_validation)
         if is_valid:
             state.errors = []
             state.is_valid = True
-            from_date = datetime.strptime(state.children[0].value, "%d/%m/%Y")
-            to_date = datetime.strptime(state.children[1].value, "%d/%m/%Y")
 
-            if to_date == from_date:
-                state.is_valid = False
-                state.errors.append(self.questionnaire.get_error_message("INVALID_DATE_RANGE_TO_FROM_SAME", self.id))
-                return False
+            if len(state.children) == 2 and state.children[0].value and state.children[1].value:
+                from_date = datetime.strptime(state.children[0].value, "%d/%m/%Y")
+                to_date = datetime.strptime(state.children[1].value, "%d/%m/%Y")
 
-            if to_date < from_date:
-                state.is_valid = False
-                state.errors.append(self.questionnaire.get_error_message("INVALID_DATE_RANGE_TO_BEFORE_FROM", self.id))
-                return False
+                if to_date == from_date:
+                    state.is_valid = False
+                    state.errors.append(self.questionnaire.get_error_message("INVALID_DATE_RANGE_TO_FROM_SAME", self.id))
+                    return False
+
+                if to_date < from_date:
+                    state.is_valid = False
+                    state.errors.append(self.questionnaire.get_error_message("INVALID_DATE_RANGE_TO_BEFORE_FROM", self.id))
+                    return False
 
         return is_valid
--- a/app/templates/questionnaire.html
+++ b/app/templates/questionnaire.html
@@ -19,7 +19,7 @@
 
     {% if navigation %}
       <div class="page__menubtn">
-        <button class="btn btn--link btn--menu js-menu-btn" data-close-label="{{_('Hide sections')}}" type="button" id="menu-btn">{{_('View sections')}}</button>
+        <button class="btn btn--link pluto btn--menu js-menu-btn" data-close-label="{{_('Hide sections')}}" type="button" id="menu-btn">{{_('View sections')}}</button>
       </div>
     {% endif %}
   </div>
@@ -79,7 +79,9 @@
       </div>
 
       <button class="btn btn--primary btn--lg qa-btn-submit venus" type="submit" name="action[save_continue]">{% block submit_button_text %}Save and continue{% endblock %}</button>
-
+      <div>
+        <button class="btn btn--link mars" type="submit" name="action[save_sign_out]">{% block save_sign_out_button_text %}Save and complete later{% endblock %}</button>
+      </div>
     </form>
 
 {% endblock main %}
--- a/app/views/questionnaire.py
+++ b/app/views/questionnaire.py
@@ -83,7 +83,7 @@ def get_block(eq_id, form_type, collection_id, group_id, group_instance, block_i
     current_location = Location(group_id, group_instance, block_id)
 
     _render_schema(current_location)
-    return _render_template(current_location, q_manager.block_state, template)
+    return _build_template(current_location, q_manager.block_state, template)
 
 
 @questionnaire_blueprint.route('<group_id>/<int:group_instance>/<block_id>', methods=["POST"])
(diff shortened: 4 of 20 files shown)


app/views/questionnaire.py
```diff
@@ -92,6 +92,13 @@ def post_block(eq_id, form_type, collection_id, group_id, group_instance, block_
     q_manager = get_questionnaire_manager(g.schema, g.schema_json)
 
     this_location = Location(group_id, group_instance, block_id)
+    if 'action[save_come_back]' in request.form:
```
jonnyshaw89: Should we not make this its own endpoint as we have with submit_answers?

app/questionnaire/questionnaire_manager.py
```diff
@@ -31,13 +31,13 @@ def __init__(self, schema, json=None):
         self._schema = schema
         self.state = None
 
-    def validate(self, location, post_data):
+    def validate(self, location, post_data, skip_mandatory_validation=False):
```
jonnyshaw89, line 31: Im not sure i like the approach of passing a boolean down the stack to change the behaviour of the validation at a lower level

app/schema/questions/date_range_question.py
```diff
@@ -7,22 +7,25 @@ class DateRangeQuestion(Question):
...
-                state.is_valid = False
-                state.errors.append(self.questionnaire.get_error_message("INVALID_DATE_RANGE_TO_FROM_SAME", self.id))
-                return False
-
-            if to_date < from_date:
-                state.is_valid = False
-                state.errors.append(self.questionnaire.get_error_message("INVALID_DATE_RANGE_TO_BEFORE_FROM", self.id))
-                return False
+
+            if state.children[0].value and state.children[1].value:
```
jonnyshaw89: do we need to check that state.children has a length of 2 instead? as `state.children[1].value` will error if there is no index of 1 whereas checking the existence of the indexes won't.

jonnyshaw89 (changes requested):

yoganandkunche:
Currently in QA
an issue has been raised #827

yoganandkunche:
@LJBabbage Census on Visitors pages gives Error 500 when user comes back and removes the visitors and click 'Save and Come back later'

app/questionnaire/questionnaire_manager.py
benjefferies: It's know what your saying it feels nasty but it hard to say it's worth tackling because wtforms being on the horizon? I guess we need to know what is happening with that before we commit to the effort to refactor

app/questionnaire/questionnaire_manager.py
```diff
@@ -56,14 +56,17 @@ def validate_all_answers(self):
 
         return True, None
 
-    def update_questionnaire_store(self, location):
+    def update_questionnaire_store(self, location, save_come_back=False):
```
benjefferies: I wonder if it would be nicer to refactor this out into a method called save_and_come_back instead of the boolean logic?

app/views/questionnaire.py
benjefferies: I agree, it would be good if we could do this however I remember @hamishtaplin suggesting posting a form to a different URL is a html 5 thing

tests/functional/spec/save-come-back.spec.js
```diff
@@ -0,0 +1,118 @@
...
+import retailTurnoverPage from '../pages/surveys/rsi/0102/retail-turnover.page'
+import internetSalesPage from '../pages/surveys/rsi/0102/internet-sales.page'
+import changeInRetailTurnover from '../pages/surveys/rsi/0102/changes-in-retail-turnover.page'
+import SummaryPage from '../pages/summary.page'
+
+const expect = chai.expect
+
+describe('RSI - Save and restore test', function() {
+
+  it('Given I am completing survey 0102, when I select save and come back later, then I am redirected to sign out page, showing a message that my survey has been saved and I cant continue without re-authenticating', function() {
```
benjefferies: Minor point, if this was split out into two tests it might read nicer:
- Given I am completing survey 0102, when I select save and come back later, then I am redirected to sign out page, showing a message that my survey has been saved
- Given I have saved to continue later, when I go back to the questionnaire, then I need to authenticate

tests/functional/spec/save-come-back.spec.js
```diff
@@ -0,0 +1,118 @@
...
+
+    // Given I am completing survey 0102
+    const collectionId = getRandomString(5)
+    startQuestionnaire('1_0102.json', 'test_user', collectionId)
```
benjefferies: 'test_user' and collectionId don't need to be defined as they are optional

tests/functional/spec/save-come-back.spec.js
```diff
@@ -0,0 +1,118 @@
...
+    // Then I must correct my response before my questionnaire can be saved and I am signed out
+    expect(retailTurnoverPage.getAlertText()).to.contain('Please only enter whole numbers into the field.')
+
+    retailTurnoverPage.setRetailTurnover('2000')
+      .saveComeBack()
+
+    expect(signOut.isOpen(), 'Sign out page should be open').to.be.true
+    })
+
+  it('Given I am completing completing survey 0102 and I have not completed mandatory questions, when I select save and come back later, then the missing mandatory questions will be ignored and I will be redirected to the exit page', function() {
```
benjefferies: This is covered in the integration test `test_save_come_back_with_household_question`

tests/functional/spec/save-come-back.spec.js
```diff
@@ -0,0 +1,118 @@
...
+    // When I select save and come back later
+    retailTurnoverPage.saveComeBack()
+
+    // Then I am redirected to sign out page, showing a message that my survey has been saved and I cant continue without re-authenticating
+    expect(signOut.isOpen(), 'sign out page should be open').to.be.true
+    browser.url(url)
+    expect(browser.getSource()).to.contain('Error 401')
+    })
+
+    it('Given I am completing survey 0102 and have outstanding validation failures, when I select save and come back later, then I must correct my response before my questionnaire can be saved and I am signed out', function() {
```
benjefferies: This is covered by `test_save_come_back_with_non_mandatory_validation_error`

tests/functional/spec/save-come-back.spec.js
```diff
@@ -0,0 +1,118 @@
...
+      .submit()
+
+    // When I select save and come back later
+    retailTurnoverPage.saveComeBack()
+
+    // Then the missing mandatory questions will be ignored and I will be redirected to the exit page
+    expect(signOut.isOpen(), 'sign out page should be open').to.be.true
+    })
+
+  it('Given whilst completing survey 0102 I have opted to save and come back later, when I return to the questionnaire after re-authenticating, then I am returned to the page I was on when I selected save and come back later and can then complete the survey', function() {
```
benjefferies: This is covered by `test_save_come_back_complete_a_block_then_revist_it_and_use_save_and_come_back`

tests/integration/questionnaire/test_questionnaire_save_come_back.py
```diff
@@ -0,0 +1,109 @@
...
+            "action[save_come_back]": "Save and come back later"
+        }
+        resp = self.client.post(block_one_url, data=post_data, follow_redirects=False)
+        self.assertEquals(resp.status_code, 200)
+        # Then
+        # we are presented with an error message
+        content = resp.get_data(True)
+        self.assertRegexpMatches(content, 'Please only enter whole numbers into the field.')
+
+    def test_save_come_back_complete_a_block_then_revist_it_and_use_save_and_come_back(self):
```
benjefferies: Should this test be name `test_save_come_back_complete_a_block_then_revist_it` minus `_and_use_save_and_come_back`

benjefferies (changes requested):

tests/functional/spec/save-come-back.spec.js
LJBabbage: This is a copy paste error, one of the tests logged back in and therefore I needed to hold the collectionId and user, I will remove the ones that don't log back in

tests/functional/spec/save-come-back.spec.js
LJBabbage: The household question has a bespoke view, and therefore its own call to save and come back, we will need both to cover it

app/views/questionnaire.py
LJBabbage: What Ben said and its the current pattern we are having to use for post_household_composition as well

tests/functional/spec/save-come-back.spec.js
LJBabbage: Kept it as one to avoid duplication, but made it a lot shorter and easier to read

tests/functional/spec/save-come-back.spec.js
LJBabbage: Slimed down the Selenium test to 2 major ones, which addresses the points below

app/questionnaire/questionnaire_manager.py
LJBabbage: Not sure what else we can do till WTF time as at the top level its just an array of messages with no validation type associated, so I can't tell if they are mandatory failures

LJBabbage:
Issues raised from QA have been fixed

benjefferies (approved):

app/views/questionnaire.py
```diff
@@ -195,6 +197,14 @@ def get_thank_you(eq_id, form_type, collection_id):
     return thank_you_page
 
 
+@questionnaire_blueprint.route('sign-out', methods=["GET"])
```
ajmaddaford: I think this URL should be `signed-out`, as the page isn't for signing out, it's the page we display once a user has 'signed out'. This would also apply to the template name.

app/templates/questionnaire.html
```diff
@@ -79,7 +79,9 @@ <h1 class="panel__title venus">This page has {{all_errors|length}} errors</h1>
       </div>
 
       <button class="btn btn--primary btn--lg qa-btn-submit venus" type="submit" name="action[save_continue]">{% block submit_button_text %}Save and continue{% endblock %}</button>
-
+      <div>
+        <button class="btn btn--link mars" type="submit" name="action[save_come_back]">{% block save_come_back_button_text %}Save and come back later{% endblock %}</button>
```
ajmaddaford: Any reason this text is overridable? Also, the text should be 'Save and complete later'

app/views/questionnaire.py
```diff
@@ -238,6 +248,9 @@ def post_household_composition(eq_id, form_type, collection_id, group_id):
         questionnaire_manager.remove_answer(this_location, answer_store, index_to_remove)
         return get_block(eq_id, form_type, collection_id, group_id, 0, 'household-composition')
 
+    elif 'action[save_come_back]' in request.form:
+        return _save_come_back(collection_id, eq_id, form_type, questionnaire_manager, this_location)
```
ajmaddaford: I think we should replace all references to `save_come_back` with `save_sign_out` as that is the functionality we are implementing, it's up to the user if they come back later or not.

ajmaddaford (changes requested):

app/questionnaire/questionnaire_manager.py
jonnyshaw89: ok lets review after WTForms

jonnyshaw89 (approved):

ajmaddaford (approved):

yoganandkunche:
@LJBabbage Save and come back later currently giving Error 500 (tested on 0112 & census household schemas)

yoganandkunche:
QA Done on this PR. Happy to sign off from QA perspective.

PR #823, opened 2017-01-06
Caching schema between requests

jonnyshaw89:
Possible implementation of caching of survey json after the 1st request.

### How to review 
Feedback on approach please.


benjefferies (changes requested):
Have we ran all the tests against the application in EQ_DEV_MODE=False?

jonnyshaw89:
@benjefferies I have run all the functional tests with caching enabled.

app/__init__.py
```diff
@@ -127,6 +131,9 @@ def override_url_for():  # pylint: disable=unused-variable
         # TODO fix health check so it no longer sends message to queue
         add_health_check(application, headers)
         start_dev_mode(application)
+        cache.init_app(application, config={'CACHE_TYPE': 'null'})  # Doesnt cache
```
benjefferies: cache.init_app(application) will default to the non-caching config

tests/integration/test_schema_cache.py
```diff
@@ -0,0 +1,43 @@
...
+        with self.application.app_context():
+
+            self.assertEqual(len(cache.cache._cache), 0)
+
+            token = create_token('star_wars', '0')
+            resp = self.client.get('/session?token=' + token.decode(), follow_redirects=True)
+
+            self.assertEquals(resp.status_code, 200)
+
+            self.assertNotEquals(cache.cache._cache['app.utilities.schema.load_and_parse_schema_memver'], None)
```
benjefferies: `cache.cache._cache['app.utilities.schema.load_and_parse_schema_memver']` could be changed to use the public method cache.get('app.utilities.schema.load_and_parse_schema_memver')

tests/integration/test_schema_cache.py
```diff
@@ -0,0 +1,43 @@
...
+from tests.integration.create_token import create_token
+from tests.integration.integration_test_case import IntegrationTestCase
+
+
+class TestApplicationVariables(IntegrationTestCase):
+
+    def setUp(self):
+        settings.EQ_DEV_MODE = False
+
+        for key_name, dev_location in settings._KEYS.items():
```
benjefferies: Would we be able to apply this to all of the integration tests to run closer to production?

benjefferies (changes requested):

benjefferies (approved):

app/__init__.py
```diff
@@ -123,6 +127,11 @@ def override_url_for():  # pylint: disable=unused-variable
 
     login_manager.init_app(application)
 
+    if settings.EQ_ENABLE_CACHE:
```
benjefferies, line 130: 👍

tests/integration/test_schema_cache.py
```diff
@@ -0,0 +1,26 @@
...
+from tests.integration.create_token import create_token
+from tests.integration.integration_test_case import IntegrationTestCase
+
+
+class TestApplicationVariables(IntegrationTestCase):
+
+    def test_schema_is_cached(self):
+        with self.application.app_context():
+
+            self.assertEqual(len(cache.cache._cache), 0)
```
iwootten, line 11: Hmm, this is a bit weird. Can we import from app.cache or 'as' another name to save drilling down this far?

tests/integration/test_schema_cache_off.py
```diff
@@ -0,0 +1,22 @@
...
+        IntegrationTestCase.setUp(self)
+
+    def tearDown(self):
+        settings.EQ_ENABLE_CACHE = True
+        IntegrationTestCase.tearDown(self)
+
+    def test_schema_is_cached(self):
+        with self.application.app_context():
+
+            self.assertTrue(isinstance(cache.cache, NullCache));
```
iwootten: and again here?

iwootten (changes requested):

tests/integration/test_schema_cache.py
jonnyshaw89: Im not sure what you mean? can you give an example?

iwootten: from app.cache import cache
or
    
    from app.cache import cache as app_cache

jonnyshaw89: ah ok. but if i did that then i would still need to higher level import for the next assertion on the cache on line 20 `cache.get`

iwootten: Oh, didn't spy that, fair enough.

iwootten (approved):

benjefferies:
Are there any outstanding questions to be answered or can this be merged?

jonnyshaw89:
The only open question it how much of a performance gain we get. but we can't run that at the moment but i don't see a reason that should block merge?

benjefferies:
Discussed with @ajmaddaford and it's probably worth running a performance test locally before merging. This will save the overhead of rolling back if it's not better

jonnyshaw89:
In progress

jonnyshaw89:
Results from test with 100 users.
Average survey completion time
Before 693ms
After 607ms

PR #832, opened 2017-01-10
Convert from GUID  to human readable Id for live surveys.

samiwel:
All live surveys have their GUIDs replaced with human readable id values.
**Note: a separate tech debt card has been created to convert all of the test schemas. That will be covered by another PR.**
Schema validator regex added to enforce format with description guidance.

I've replaced the Id in the following schema, which I think covers all of the 'Live' surveys.
  - 1_0102.json
  - 1_0112.json
  - 1_0203.json
  - 1_0205.json
  - 1_0213.json
  - 1_0215.json

The census schemas followed this naming convention from the beginning as far as I can tell.

### How to review 
Please check the live schemas to double check that I haven't missed any GUID identifiers.
Please also bring to my attention if I have missed any 'Live' schemas.
All tests should pass. Survey runner should behave normally.


collisdigital:
Can't see the schema validator changes in this PR

app/data/schema/schema-v1.json
```diff
@@ -2,6 +2,10 @@
   "$schema": "http://json-schema.org/draft-04/schema#",
   "type": "object",
   "definitions": {
+    "id": {
+        "type": "string",
+        "pattern": "^[a-z0-9][a-z0-9\\-]*[a-z0-9]$"
```
jonnyshaw89, line 8: Do we need some tests against this regex to check allowed and disallowed values?

jonnyshaw89: Also the TestSchemaValidation validates all schemas so given the test schemas haven't been updated i would expect the test to fail?

jonnyshaw89 (changes requested):

app/data/schema/schema-v1.json
samiwel: The GUIDs satisfy this regex so this won't break on account of the test schemas. I think the main purpose of adding this is to ensure that nobody adds any stupid identifiers in the future.
One change I think should be made, once all the test schemas have been converted, is to drop the 0-9 from the first match. This will help to enforce that Ids don't start with a number. But GUIDs will break that rule at the moment.

I tested the RegExp using an online tool, but if it's helpful for documentation I can also write a few tests.

jonnyshaw89 (approved):

tests/app/data/test_schema_id_regex.py
```diff
@@ -0,0 +1,133 @@
...
+        json_to_validate = create_schema_with_id('not_a_valid_id')
+
+        # When
+        errors = self.validate_json_against_schema(json_to_validate, self.schema)
+
+        # Then
+        self.assertEqual(len(errors), 1)
+
+    @staticmethod
+    def validate_json_against_schema(json_to_validate, schema):
```
iwootten: QC seems to be complaining that this doesn't appear with test_ in the name. Given it's static, is it worth moving it outside the class as with create_schema_with_id?

iwootten (changes requested):

tests/app/data/test_schema_id_regex.py
samiwel: Thanks, hadn't spotted that QC issue.

app/data/schema/schema-v1.json
benjefferies: Maybe add a description to say that it must be a alphanumerical hyphen separated string?

iwootten (approved):

app/data/schema/schema-v1.json
samiwel: @benjefferies I've added a description for the type definition in the latest squashed commit. Feel free to take a look.

PR #834, opened 2017-01-11
Debt 835 pylint

benjefferies:
Codacy uses pylint for linting which gives differing results to the flake8 which is ran locally and on builds therefore we don't know about some of the linting issues until it's passed through codacy which can mean waiting another iteration of the build.

The changes I have made is to use pylint instead of flake8.

I have also applied pylint to the test code.

### How to review 
Tests continue to pass


scripts/run_tests.sh
```diff
@@ -32,8 +32,14 @@ function display_result {
   fi
 }
 
-flake8 --max-complexity 10 --count
```
collisdigital, line 39: this was also doing a McCabe complexity analysis, does pylint have the same plugin?

benjefferies: [py-lint plugin](https://pypi.python.org/pypi/pylint-mccabe) Yeah looks like it although I didn't think about adding it so thanks for pointing it out

app/__init__.py
```diff
@@ -124,7 +118,6 @@ def override_url_for():  # pylint: disable=unused-variable
     login_manager.init_app(application)
 
     if settings.EQ_DEV_MODE:
-        # TODO fix health check so it no longer sends message to queue
```
benjefferies, line 114: https://trello.com/c/5WpLqNYy/364-health-check-rabbitmq-without-putting-a-message-on-the-queue

app/cryptography/jwt_encoder.py
```diff
@@ -27,28 +28,22 @@ def __init__(self):
...
         """
         if jwe_protected_header is None:
             jwe_protected_header = self._jwe_protected_header()
-        if cek is None:
-            cek = self.cek
-        if iv is None:
-            iv = self.iv
         if encrypted_key is None:
-            encrypted_key = self._encrypted_key(cek)
+            encrypted_key = self._encrypted_key(self.cek)
```
iwootten, line 45: Don't these variables need to be named as what is passed into the method?

iwootten (changes requested):
I feel like this change possibly should have been carried out across multiple PR's. At the moment, it's pretty difficult to see which rules are being tackled in each file.

Has a decision been made on whether this is in addition to or instead of our current flake8 rules?

app/cryptography/jwt_encoder.py
benjefferies: Not sure what this means

iwootten: Ah, I was under the impression the cek and iv variables had been renamed in the method description here, but seems they have been removed. Do we not use these args then?

benjefferies: We used them as args only for tests, we could have just set self.cek like we do now to change it for testing hence I they were removed

benjefferies:
I believe we have decided to go with pylint unless I am wrong @ajmaddaford?

iwootten:
Without this PR identifying why we would use PyLint over Flake8, I'm not sure I can approve this. Switching over to it simply because one of our two quality tools uses it seems a bit premature. It would also presumably either mean switching off quantified code or hitting different errors with that check too?

Additionally, I don't see anything in the original story about disabling current flake8 checks?

benjefferies:
Flake8 gives us linting and cyclomatic complexity evaluation, pylint also applies these checks.
Flake8 doesn't apply strict pep 8 styling check it uses it's owned styling [inspired by pep 8](http://flake8.pycqa.org/en/latest/manpage.html) checks. Pylint follows pep 8 more closely.
A benefit of sticking to pep 8 styling guide is that it conforms with the generally accepted styling guide which other tools such as pycharm uses.
Non of the static code analysis tools that we use apply flake8 style checks but codacy does apply pep 8 checks. This can be frustrating and can waste time if the builds pass but it fails linting because the tools we use don't match the style.
quantified doesn't check lint using flake8 so we can still run it

ajmaddaford:
On using pylint rather than flake8, pylint appears to be stricter than flake8 - on master, running flake8 on the tests - `flake8 tests` - generates no issues, running pylint - `pylint --rcfile=.pylintrc ./tests` - generates lots of valid issues.

Codacy can be configured to use the .pylintrc from the repo so we would get consistent checks locally and via status checks on pull requests.

Before we can merge this we need to ensure we consider the functionality provided by the flake8 plugins - `flake8-print, pep8-naming, flake8-blind-except, flake8-commas, flake8-debugger, flake8-mock, flake8-tuple` - these need to be reviewed to determine if pylint covers what they do, or they are no longer required.

I also ran pep8 and this generates style issues not picked up by pylint e.g.
```
./token_generator.py:13:13: E126 continuation line over-indented for hanging indent
./app/settings.py:74:45: E241 multiple spaces after ':'
./app/settings.py:75:45: E241 multiple spaces after ':'
...
./app/libs/utils.py:17:35: E226 missing whitespace around arithmetic operator
./app/submitter/converter.py:120:42: E126 continuation line over-indented for hanging indent
./app/templating/metadata_context.py:45:22: E241 multiple spaces after ':'
./app/validation/type_validator_factory.py:34:13: E123 closing bracket does not match indentation of opening bracket's line
...
```

I think we need pylint to pick these up, or also run pep8. As this pull request is already quite large it would probably be best to tackle those issues in another pull request.

collisdigital:
Note that we already have a .pylintrc file; so the things pylint is currently missing may be because of the configuration in there set to ignore them.

benjefferies:
I think we will need to review what is ignored in the .pylintrc if we go ahead with but I'll raise a card to do that as this PR is a bit unwieldy as it is

benjefferies:
√ flake8-commas - https://github.com/taozle/pylint-comma
~ flake8-print - Checks for print statements in form `print 'test'` but not `print('test')` 
√ pep8-naming - currently disabled in .pylintrc as invalid-name
√ flake8-blind-except - Caught as bare-except
x flake8-debugger
x flake8-mock
√ flake8-tuple - Caught as superfluous-parens

requirements_for_test.txt
```diff
@@ -6,3 +6,5 @@ pytest-cov==2.3.0
 jsonschema==2.5.1
 pylint==1.6.4
 flask-debugtoolbar==0.10.0
+pylint-mccabe==0.1.3
+git+https://github.com/taozle/pylint-comma.git
```
ajmaddaford: I think the check covered by pylint-comma is done in flake8-tuple

scripts/run_tests.sh
```diff
@@ -39,6 +39,15 @@ function display_result {
...
 display_result $? 1 "Code style check"
 
+pylint --rcfile=.pylintrc -j 0 ./app ./tests
+# pylint bit encodes the exit code to allow you to figure out which category has failed.
+# https://docs.pylint.org/en/1.6.0/run.html#exit-codes
+# http://stackoverflow.com/questions/6626351/how-to-extract-bits-from-return-code-number-in-bash
+# We want to fail if there are any errors or fatal errors so we use 3
+errorcode=$?
+(( res = $errorcode & 3 ))
+display_result $res "Code style check"
```
ajmaddaford: It would be better to have a different message for the flake8 and pylint results that mentioned flake8/pylint

app/cryptography/jwt_encoder.py
```diff
@@ -27,22 +29,28 @@ def __init__(self):
     def encode(self, payload):
         return jwt.encode(payload, self.private_key, algorithm="RS256", headers={'kid': KID, 'typ': 'jwt'})
 
-    def encrypt_token(self, text, jwe_protected_header=None, encrypted_key=None, tag=None):
+    def encrypt_token(self, text, jwe_protected_header=None, cek=None, iv=None, encrypted_key=None, tag=None):
```
ajmaddaford: Removing `cek` and `iv` should be done in another pull request, unless it's required by the pylint fixes?

ajmaddaford (changes requested):

ajmaddaford:
As we can't cover all of the checks done by flake8 with pylint plugins, we're going to use both.

app/cryptography/jwt_encoder.py
benjefferies: Something mess up here because the change was originally done on this branch then pulled out into another then rebased but it took this branches version over master

ajmaddaford (approved):

iwootten (approved):

PR #835, opened 2017-01-11
WTForms spike

iwootten:

This is a first pass of replacement of state in survey-runner with the WTForms library. The main views have been updated to be constructed using WTForms components only and templates have been updated to use the data coming from WTForms for rendering. Most of the integration tests fail right now, mainly due to them no longer making sense in a WTForms implementation and the length of time required to test all scenarios. The unit tests obviously still pass but we'd need to update the to test equivalent WTForms elements.

WTForms is able to support all our custom question types through the use of custom field enclosures. The Household Relationship/Composition and forms using dates all use this approach. It's worth noting the use of metaclasses to generate forms, which will therefore only describe the instance of the class used to construct the form. Where we require dynamic fields and validation, it is necessary to define methods that construct these metaclasses. It's worth [reading up](https://blog.ionelmc.ro/2015/02/09/understanding-python-metaclasses/) on this concept if you're not familiar with it.

This PR removes the need for what we have called widgets, though the template structure has mostly been left in place as a future step for removal. Additionally all the schema/state folders have been left in place due to so many tests requiring updates if we were to opt to use it. Everything in questionnaire_state and schema folders could likely be removed by this change.

I've had to update many of the census functional test pages due to the zero-based indexing that wtforms uses for multiple choice elements.

Piping has been completely ignored as part of this work, but could be achieved in part by passing a piped version of the schema to jinja to render rather than the schema itself.

#### Parent/Child Answers

This PR introduces the concept of parent/child answers in order to support the use an additional input where multiple choice questions require 'other' options to be specified as text input fields.

Story questions have been answered in confluence here: https://digitaleq.atlassian.net/wiki/display/EQ/WTForms

### How to review 

High level feedback welcome


benjefferies (commented):
It has less impact than I expected. It looks like a positive move where we can still support all of the widgets we currently use but have less custom logic.
It looks like a refactor is needed before we get it in and some tests to cover some of the wtforms

ajmaddaford:
Closing as the spike has been reviewed and we've agreed a way forward

PR #836, opened 2017-01-12
Add RU ref into email subject on introduction page.

samiwel:
This PR adds the reporting unit reference into the subject of the email when the email link is clicked on the survey introduction page.

### How to review 
Open up any survey that has an introduction page.
When the email link is clicked, observe that the subject line has the RU reference number inline.

### DIFF
--- a/tests/integration/introduction/test_introduction.py
+++ b/tests/integration/introduction/test_introduction.py
@@ -1,0 +1,23 @@
+from tests.integration.integration_test_case import IntegrationTestCase
+from tests.integration.create_token import create_token
+
+
+class TestIntroduction(IntegrationTestCase):
+
+    def test_mail_link_contains_ru_ref_in_subject(self):
+        # Given
+        response = self.start_new_survey()
+
+        # When
+        content = response.get_data(True)
+
+        # Then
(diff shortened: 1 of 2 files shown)


tests/integration/introduction/test_introduction.py
```diff
@@ -0,0 +1,33 @@
+from tests.integration.integration_test_case import IntegrationTestCase
+from tests.integration.create_token import create_token
+
+
+class TestIntroduction(IntegrationTestCase):
+
+    def test_introduction_has_email_link(self):
```
iwootten: Isn't this test a subset of the one below?

iwootten (changes requested):

tests/integration/introduction/test_introduction.py
samiwel: Yeah I guess you're right. I can remove it - I initially added it as a sanity check to make sure the survey I chose for the tests had an introduction page. Some of the surveys (e.g. census) don't have an introduction page with an email link.

jonnyshaw89 (approved):

iwootten (approved):

PR #837, opened 2017-01-12
Remove previous links on thank you pages

### LINKED ISSUE #828: All Live Surveys: Remove 'previous' link from successful submission screen
### Expected behaviour
'Previous' link must not be displayed on 'successful submission screen' as it will only trigger an error.
### Actual behaviour
Previous link is displayed
Launch any eQ survey
Complete the survey and submit
#### Browser
All Browsers
#### Operating System
All
### Screenshot
screen shot 2017-01-10 at 09 53 32

iwootten:
Removed previous link from thank you pages and added tests to check where previous links appear.

Fixes #828 

### How to review 
Check tests still pass.

### DIFF
--- a/app/views/questionnaire.py
+++ b/app/views/questionnaire.py
@@ -326,7 +326,10 @@ def _build_template(current_location, context=None, template=None):
     previous_location = path_finder.get_previous_location(current_location)
 
     previous_url = None
-    if previous_location is not None and not SchemaHelper.is_first_block_id_for_group(g.schema_json, current_location.group_id, current_location.block_id):
+
+    is_first_block_for_group = SchemaHelper.is_first_block_id_for_group(g.schema_json, current_location.group_id, current_location.block_id)
+
+    if previous_location is not None and not is_first_block_for_group and not current_location.block_id == 'thank-you':
         previous_url = previous_location.url(metadata)
 
     return _render_template(context, current_location.block_id, front_end_navigation, metadata_context, previous_url, template)
--- a/tests/integration/questionnaire/test_questionnaire_previous_link.py
+++ b/tests/integration/questionnaire/test_questionnaire_previous_link.py
@@ -1,0 +1,76 @@
+from tests.integration.create_token import create_token
+from tests.integration.integration_test_case import IntegrationTestCase
+
+
+class TestQuestionnairePreviousLink(IntegrationTestCase):
+
+    def test_previous_link_doesnt_appear_on_introduction(self):
+        # Given
+        token = create_token('final_confirmation', 'test')
+        resp = self.client.get('/session?token=' + token.decode(), follow_redirects=False)
+
+        # When we open the introduction
+        content = resp.get_data(True)
+
+        # Then previous link does not appear
+        self.assertNotIn('Previous', content)
+
+    def test_previous_link_appears_on_page_following_introduction(self):
+        base_url = '/questionnaire/test/final_confirmation/789/'
+
+        # Given
+        token = create_token('final_confirmation', 'test')
+        self.client.get('/session?token=' + token.decode(), follow_redirects=False)
+
+        # When
+        # We proceed through the questionnaire
+        post_data = {
+            'action[start_questionnaire]': 'Start Questionnaire'
+        }
+
+        block_one_url, resp = self.postRedirectGet(base_url + 'introduction', post_data)
+
+        content = resp.get_data(True)
+        self.assertNotRegexpMatches(content, 'Previous')
+
+        post_data = {
+            "ca3ce3a3-ae44-4e30-8f85-5b6a7a2fb23c": " Bacon",
+            "action[save_continue]": "Save &amp; Continue"
+        }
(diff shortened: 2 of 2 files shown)


tests/integration/questionnaire/test_questionnaire_previous_link.py
```diff
@@ -0,0 +1,75 @@
...
+    def test_previous_link_doesnt_appear_on_introduction(self):
+        # Given
+        token = create_token('final_confirmation', 'test')
+        resp = self.client.get('/session?token=' + token.decode(), follow_redirects=False)
+
+        # When we open the introduction
+        content = resp.get_data(True)
+
+        # Then previous link does not appear
+        self.assertNotRegexpMatches(content, 'Previous')
```
benjefferies: I think we should use assertNotIn when we are checking for contains otherwise it's a bit misleading

tests/integration/questionnaire/test_questionnaire_previous_link.py
```diff
@@ -0,0 +1,75 @@
...
+    def test_previous_link_appears_on_all_non_start_end_pages(self):
+        base_url = '/questionnaire/test/final_confirmation/789/'
+
+        # Given
+        token = create_token('final_confirmation', 'test')
+        self.client.get('/session?token=' + token.decode(), follow_redirects=False)
+
+        # When
+        # We proceed through the questionnaire
+        post_data = {
```
benjefferies, line 27: Do we need to post to the introduction page or can we just post the first page

benjefferies (changes requested):

tests/integration/questionnaire/test_questionnaire_previous_link.py
jonnyshaw89: If the integration tests don't post in the same pages as the user would can we be sure that the state of the system will be the same?

benjefferies: I guess either way it should behave the same because the user in theory could jump around by changing the URL

app/views/questionnaire.py
```diff
@@ -304,7 +304,10 @@ def _render_template(current_location, context=None, template=None):
     previous_location = path_finder.get_previous_location(current_location)
 
     previous_url = None
-    if previous_location is not None and not SchemaHelper.is_first_block_id_for_group(g.schema_json, current_location.group_id, current_location.block_id):
+
+    is_first_block_for_group = SchemaHelper.is_first_block_id_for_group(g.schema_json, current_location.group_id, current_location.block_id)
+
+    if previous_location is not None and not is_first_block_for_group and not current_location.block_id == 'thank-you':
```
benjefferies, line 332: A more general observation, I wonder if we should move the previous location logic into the navigation class after the previous location is a part of navigation even if it's not rendered in the same way

tests/integration/questionnaire/test_questionnaire_previous_link.py
samiwel: I think `assertNotRegexpMatches` has been deprecated and should be replaced with `assertNotRegex`.

tests/integration/questionnaire/test_questionnaire_previous_link.py
```diff
@@ -0,0 +1,75 @@
...
+        token = create_token('final_confirmation', 'test')
+        resp = self.client.get('/session?token=' + token.decode(), follow_redirects=False)
+
+        # When we open the introduction
+        content = resp.get_data(True)
+
+        # Then previous link does not appear
+        self.assertNotRegexpMatches(content, 'Previous')
+
+    def test_previous_link_appears_on_all_non_start_end_pages(self):
```
samiwel: What is a non-start end page? I wonder if there is any benefit of splitting this out into two tests for clarity? e.g. `test_previous_link_not_visible_on_page_immediately_after_introduction` and `test_previous_link_visible_on_subsequent_question_pages` or similar.

iwootten: A page that isn't at the start or end? :)

I've renamed this for clarity.

app/views/questionnaire.py
iwootten: That's a fair enough observation - the wtforms spike makes some changes in this regard too whereby the current location is also passed to jinja. Think tackling it is outside the scope of this issue though.

jonnyshaw89 (approved):

yoganandkunche:
Defect passed retest on branch

tests/integration/questionnaire/test_questionnaire_previous_link.py
iwootten: Yeah, the form is posting back - so I'm not sure why we'd change the integration test to do something else.

benjefferies (approved):

tests/integration/questionnaire/test_questionnaire_previous_link.py
benjefferies: I follow what's going on now, got misled by the prior test name

PR #844, opened 2017-01-16
Add validation messages to UKIS survey.

samiwel:
Added appropriate validation messages to the UKIS schema as per the descriptions in the Google Slides document.

### How to review 
Please check to ensure that validation messages are as per the Google Slides and let me know if I've missed any.

### DIFF
--- a/app/data/1_0001.json
+++ b/app/data/1_0001.json
@@ -413,14 +413,7 @@
                                             "mandatory": false,
                                             "options": [],
                                             "q_code": "1410",
-                                            "type": "Currency",
-                                            "validation": {
-                                                "messages" : {
-                                                    "INTEGER_TOO_LARGE": "The maximum value allowed is 9999999999. Please correct your answer.",
-                                                    "NOT_INTEGER": "The value you have entered is invalid. Please enter a numeric value.",
-                                                    "NEGATIVE_INTEGER": "The value cannot be negative. Please correct your answer."
-                                                }
-                                            }
+                                            "type": "Currency"
                                         }
                                     ],
                                     "description": "",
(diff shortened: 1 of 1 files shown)


yoganandkunche:
Issue/Observation: When a number bigger then 10 digits is entered the error message must be  **Expected**: 'The maximum value allowed is 9999999999. Please correct your answer.'
**Actual:** This number is too big.

yoganandkunche:
Same as above with % fields. @Jenny to update the spec file for the correct validation message.

yoganandkunche:
Same issue as above for Question: 13. General Information
13.2. How long has it taken you to complete this questionnaire?

collisdigital:
With the agreement on how £1000s will be handled I am concerned that the 9 digit limit will prevent legitimate data; can wbe verify that this limit should still be applied, it was originally because of downstream processing limitations; if SDX is doing a transform on this data we may need to add 3 more digits....

samiwel:
Received following confirmation from Jenny:
> DCM have advised not to add the additional 3 digits. Sally who has lots of business knowledge has said it would not be needed.
>> @jeffinerp So to be clear, are we saying that validation messages should be consistent as per all other surveys?
>>> Yes :0)

So I have implemented validation messages as per all other surveys (i.e. 'The maximum value allowed is 9999999999. Please correct your answer.')
On Percentage fields I have customised this message to say 100.

LJBabbage (approved):

app/data/1_0001.json
```diff
@@ -4859,7 +4890,12 @@
                                             "mandatory": false,
                                             "options": [],
                                             "q_code": "2620",
-                                            "type": "Percentage"
+                                            "type": "Percentage",
+                                            "validation": {
+                                                "messages" : {
+                                                    "INTEGER_TOO_LARGE": "The maximum value allowed is 100. Please correct your answer."
```
benjefferies, line 4896: Can this not be made a global message?

samiwel: It is defined as a global message, it's being overridden for Percentage fields so that the value reads 100 instead of 9999999999. I'm not sure if it's possible to define messages at a Question level?

samiwel: Note: due to key sorting, the global messages are down the bottom of the file.

samiwel: @benjefferies I tested moving the validation message overrides to the Question level, rather than having them on each percentage answer. But the messages didn't get resolved correctly - it just displayed the global messages. Perhaps its worth creating a new card to investigate if specifying validation overrides on the Question makes sense and to fix that. But in terms of this particular card, I think this is the least amount of duplication I am able to achieve.

benjefferies (approved):

PR #846, opened 2017-01-17
Use structlog for structured logging

benjefferies:
Using existing logging set up wrapped by structlog. This means logging
should not change apart from the logging structure. Logging to splunk
and to file will still behave the same.

Applying SDC logging standards to include informative data in all
logging. Logging is output in json format. Standard logging ids:

eq_id: eq id
form_type: Form typee
ce_id: Collection exercise sid
tx_id: Transaction id

[Review logging](https://digitaleq.atlassian.net/wiki/display/SDC/Logging)

### How to review 
Evaluate whether logging changes are appropriate.
All test should pass.


app/authentication/authenticator.py
```diff
@@ -1,5 +1,8 @@
 import logging
```
samiwel: I think this import is unused.

app/authentication/jwt_decoder.py
```diff
@@ -1,6 +1,8 @@
 import json
 import logging
```
samiwel: Same as above.

app/authentication/authenticator.py
benjefferies: I was going to wait for the pylint changes before sorting imports out as there will be lots of conflicts

app/authentication/session_storage.py
```diff
@@ -42,23 +41,21 @@ def has_user_id(self):
...
     def clear(self):
         """
         Removes a user id from the session
         """
-        logger.debug("SessionManager remove_user_id() - session %s", session)
         if EQ_SESSION_ID in session:
             eq_session_id = session[EQ_SESSION_ID]
             eq_session = self._get_user_session(eq_session_id)
-            logger.debug("About to delete entry from eq_session table %s", eq_session)
+            logger.debug("Deleting session from eq_session table", session=eq_session)
```
samiwel: Good job on the log messaging changes 👍

app/authentication/authenticator.py
samiwel: Ah okay, fair enough.

app/__init__.py
```diff
@@ -41,43 +45,24 @@
...
-        logging.error('Cannot connect to Message Server')
+        logger.error('Cannot connect to Message Server')
         return False, "rabbit mq unavailable"
 
 
 def git_revision():
     return True, settings.EQ_GIT_REF
 
 
-login_manager = LoginManager()
```
ajmaddaford: I can see why this was moved, but it should be done in a separate pull request.

ajmaddaford (changes requested):

app/__init__.py
```diff
@@ -4,6 +4,13 @@
...
 
+from flask_login import LoginManager
+from structlog import configure
+from structlog import get_logger
+from structlog.processors import JSONRenderer
+from structlog.stdlib import LoggerFactory
+from structlog.threadlocal import wrap_dict
+
 from app import settings
 from app.authentication.authenticator import Authenticator
```
jonnyshaw89: This references a deleted class which causes `ImportError: cannot import name 'Authenticator'`

jonnyshaw89 (changes requested):
I also don't seem to be able to start the app because of an error 
`The file /Users/jonshaw/code/eq-survey-runner/key='/Users/jonshaw/code/eq-survey-runner/jwt-test-keys/sdc-user-authentication-signing-rrm-private-key.pem' does not exist.`
However that file does exist

app/__init__.py
benjefferies: Fixed

benjefferies:
This branch is failing on codecov failure in areas that have been changed due to logging changes but no logic added. Can we ignore the failure in this case or should I backfill tests? If we want to action the tests I suggest we do it on another branch as a backfill highlighted missing tests debt ticket

app/questionnaire/path_finder.py
```diff
@@ -30,7 +31,7 @@ def _block_index_for_location(blocks, location):
                 return next(index for (index, b) in enumerate(blocks) if b["block"]["id"] == location.block_id and
                             b["group_id"] == location.group_id and b['group_instance'] == location.group_instance)
         except StopIteration:
-            logger.error("Navigation failure looking for %s", location)
+            logger.error("Could not get index", location=location)
```
ajmaddaford: This will log out a location object, which may not be what we want?

app/questionnaire/path_finder.py
```diff
@@ -54,7 +55,7 @@ def build_path(self, blocks, this_location):
 
             block_index = PathFinder._block_index_for_location(blocks, this_location)
             if block_index is None:
-                logger.error('build_path: _block_index_for_location %s is None (invalid location)', this_location)
+                logger.error('Block index is None (invalid location)', location=this_location)
```
ajmaddaford: This will log out a location object, which may not be what we want?

app/questionnaire/questionnaire_manager.py
```diff
@@ -42,7 +44,7 @@ def validate_all_answers(self):
             is_valid = self.validate(location, answers)
 
             if not is_valid:
-                logger.debug("Failed validation with current location %s", str(location))
+                logger.debug("Failed validation", location=location)
```
ajmaddaford: This will log out a location object, which may not be what we want?

app/questionnaire/questionnaire_manager.py
```diff
@@ -66,7 +68,7 @@ def _add_update_answer_store(self, location):
         return questionnaire_store
 
     def process_incoming_answers(self, location, post_data):
-        logger.debug("Processing post data for %s", location)
+        logger.debug("Processing post data", location=location)
```
ajmaddaford: This will log out a location object, which may not be what we want?

app/schema_loader/schema_loader.py
```diff
@@ -48,7 +49,7 @@ def load_schema_file(schema_file, language_code='en'):
         if language_code != 'en':
             return load_schema_file(schema_file, 'en')
         else:
-            logging.error("No schema file exists %s", schema_file)
+            logger.error("No schema file exists", schema_name=schema_file)
```
ajmaddaford: I think using `filename` rather than `schema_name` would be clearer

app/settings.py
```diff
@@ -20,7 +20,7 @@ def parse_mode(string):
 
 def get_key(_key_name):
     if _key_name:
-        logger.debug("Opening file %s", _key_name)
+        logger.debug("Reading key from file", key=_key_name)
```
ajmaddaford: I think renaming `key` to `filename` would make this clearer

app/authentication/authenticator.py
```diff
@@ -10,77 +10,74 @@
...
+    session_storage.store_user_ik(user_ik)
+
+    # store the meta data
+    metadata = parse_metadata(token)
+
+    questionnaire_store = get_questionnaire_store(user_id, user_ik)
+    questionnaire_store.metadata = metadata
+    questionnaire_store.add_or_update()
+
+    logger.bind(tx_id=metadata["tx_id"])
```
ajmaddaford: Can we bind this immediately after parsing the metadata?

app/submitter/submitter.py
```diff
@@ -90,7 +91,7 @@ def send_message(self, message, queue):
         :return: a boolean value indicating if it was successful
         """
         message_as_string = str(message)
-        logger.info("Sending messaging " + message_as_string)
+        logger.info("Sending message to Rabbit MQ", message=str(message_as_string))
```
ajmaddaford: message is converted to a string in the line above

app/views/questionnaire.py
```diff
@@ -144,8 +143,9 @@ def post_interstitial(eq_id, form_type, collection_id, block_id):
         raise NotFound
 
     metadata = get_metadata(current_user)
-    logger.info("Redirecting user to next location %s with tx_id=%s", str(next_location), metadata["tx_id"])
-    return redirect(next_location.url(metadata))
+    next_location_url = next_location.url(metadata)
+    logger.debug("Redirecting user", redirect=next_location_url)
```
ajmaddaford: It would be good to start working towards consistency in logging messages. I think a message of `Redirecting` and `url=next_location_url` is as clear and could be used anywhere we redirect.

scripts/run_tests.sh
```diff
@@ -36,8 +36,8 @@ function display_result {
   fi
 }
 
-flake8 --max-complexity 10 --count
-display_result $? 1 "Code style check"
+# flake8 --max-complexity 10 --count
+# display_result $? 1 "Code style check"
```
ajmaddaford: Not sure this should be commented out

ajmaddaford (changes requested):

application.py
```diff
@@ -1,18 +1,92 @@
...
+    """
+    Add the service name to the event dict.
+    """
+    event_dict['service'] = 'eq-survey-runner'
+    return event_dict
+
+
+def start():
+    from app import create_app
+    application = create_app()
```
collisdigital: Is this going to work behind Apache? It expects an application at the top level for the wsgi module to call...

benjefferies: Not sure will investigate. Another reason to be running the app in development in the same way as we do in production

jonnyshaw89:
@benjefferies is there any way we can disable the access logs locally?
eg.
`127.0.0.1 - - [27/Jan/2017 10:46:21] "GET /s/fonts/lato/latolatin-regular-webfont.woff2 HTTP/1.1" 200 -`

I find them quite busy and make it hard to see other logging

benjefferies:
@jonnyshaw89 if you set EQ_WERKZEUG_LOG_LEVEL=WARNING it will silence them

README.md
```diff
@@ -201,6 +201,7 @@ EQ_ENABLE_FLASK_DEBUG_TOOLBAR - Enable the flask debug toolbar
 EQ_ENABLE_CACHE - Enable caching of the schema
 EQ_MAX_HTTP_POST_CONTENT_LENGTH - The maximum http post content length that the system wil accept
 EQ_MAX_NUM_REPEATS - The maximum number of repeats the system will allow
+EQ_DEVELOPER_LOGGING - Enable developer style logging described here http://structlog.readthedocs.io/en/stable/development.html
```
jonnyshaw89, line 204: when i enable this i get the following error
```
  File "application.py", line 112, in <module>
    configure_logging()
  File "application.py", line 99, in configure_logging
    renderer_processor = ConsoleRenderer() if EQ_DEVELOPER_LOGGING else JSONRenderer()
  File "/Users/jonshaw/jon/lib/python3.5/site-packages/structlog/dev.py", line 68, in __init__
    package="colorama"
SystemError: ConsoleRenderer requires the colorama package installed.  If you want to use the helpers from structlog.dev, it is strongly recommended to install structlog using `pip install structlog[dev]`.
```

jonnyshaw89 (changes requested):

README.md
benjefferies: Good spot that's missing from the requirements. A reinstall of requirements.txt should sort this

app/questionnaire/path_finder.py
```diff
@@ -54,7 +55,7 @@ def build_path(self, blocks, this_location):
 
             block_index = PathFinder._block_index_for_location(blocks, this_location)
             if block_index is None:
-                logger.error('build_path: _block_index_for_location %s is None (invalid location)', this_location)
+                logger.error('block index is none (invalid location)', **this_location)
```
ajmaddaford: This will break, it should be `**this_location.__dict__`

app/questionnaire/navigation.py
```diff
@@ -36,16 +37,18 @@ def build_navigation(self, current_group_id, current_group_instance):
...
-            logger.debug("Building frontend navigation for group %s", group['id'])
+            logger.debug("building frontend navigation", group_id=group['id'])
 
-            repeating_rule = SchemaHelper.get_repeat_rule(group)
             skip_group = self._should_skip_group(current_group_instance, group)
 
             if not skip_group:
+                repeating_rule = SchemaHelper.get_repeat_rule(group)
                 if repeating_rule:
+                    logger.debug("building repeating navigation", group_id=group['id'])
```
jonnyshaw89, line 47: Are these debug logs delivering much value as the give no context as to what is actually building. e.g. which repeating rule?

jonnyshaw89 (changes requested):

app/questionnaire/navigation.py
benjefferies: You can only have one repeating rule per group at the moment hence it gives enough information to be able to find the rule. Rules don't have any identifier so the alternative would be to log the whole this which seems unnecessary

requirements.txt
```diff
@@ -284,3 +284,6 @@ Blinker==1.4 \
     --hash=sha256:471aee25f3992bd325afa3772f1063dbdbbca947a041b8b89466dc00d606f8b6
 Flask-Cache==0.13.1 \
     --hash=sha256:90126ca9bc063854ef8ee276e95d38b2b4ec8e45fd77d5751d37971ee27c7ef4
+structlog==16.1.0 \
+    --hash=sha256:a8cc45c208c6b251031ec223940552eccd97dbdb322474e0c4ec9e50f76a225a
+structlog[dev]==16.1.0
```
jonnyshaw89: Doesn't work after re-install. also you are missing the hash.

You can generate these with `hashin` followed by the dependency

jonnyshaw89 (changes requested):

application.py
```diff
@@ -1,12 +1,116 @@
...
+                record.exc_text = self.formatException(record.exc_info)
+        if record.exc_text:
+            if s[-1:] != "\n":
+                s += "\n"
+            s = s + record.exc_text
+        if record.stack_info:
+            if s[-1:] != "\n":
+                s += "\n"
+            s = s + self.formatStack(record.stack_info)
+        return s
```
ajmaddaford: Can we call the `logging.Formatter` `format` method if `exception_key not in record.message` rather than copy the method in?

ajmaddaford (changes requested):

app/questionnaire/path_finder.py
benjefferies: Good spot, fixed

application.py
benjefferies: Removed this logic as it isn't required if we want to log the exception. We should add it to `logger.error(exc_info=e)`. We shouldn't use `log.exception` but we should produce the stack info and add it to `logger.error` so it will play nice with json messages

requirements.txt
benjefferies: Done, the library installed is actually colorama and structlog[dev] is just an alias to add that dependency on

README.md
benjefferies: Fixed

app/__init__.py
```diff
@@ -28,22 +28,17 @@
...
-
 cache = Cache()
+logger = get_logger()
 
 
 def rabbitmq_available():
     submitter = SubmitterFactory.get_submitter()
     if submitter.send_test():
-        logging.info('RabbitMQ Healthtest OK')
+        logger.info('rabbitmq healthtest ok')
```
iwootten, line 38: What's the reasoning behind removing uppercasing?

benjefferies: It's suggested event messages are lowercase in https://digitaleq.atlassian.net/wiki/pages/viewpage.action?spaceKey=SDC&title=Logging

app/questionnaire/questionnaire_manager.py
```diff
@@ -43,7 +42,7 @@ def validate_all_answers(self):
             is_valid = self.validate(location, answers)
 
             if not is_valid:
-                logger.debug("Failed validation with current location %s", str(location))
+                logger.debug("failed validation", **location.__dict__)
```
iwootten, line 45: Might be better to define "__str__" on the location rather than calling this internal method?

iwootten (changes requested):

app/questionnaire/questionnaire_manager.py
benjefferies: The only thing with that is we'd have a location in the message which looks like 
```
{
  "event": "failed validation"
  "location": "group/block/answer"
}
```
vs
```
{
"event": "failed validation"
"group": "group_id',
"block": "block_id",
"answer": "answer_id"
}
```

The latter might make it easier to query in the logs? i.e. find all log messages for group X

iwootten: oky doky

iwootten (approved):

.pylintrc
```diff
@@ -20,7 +20,7 @@ persistent=yes
 
 # List of plugins (as comma separated values of python modules names) to load,
 # usually to register additional checkers.
-load-plugins=pylint_mccabe
+load-plugins=pylint_mccabe,pylint_structlog_exc_info
```
ajmaddaford: I think `pylint_structlog_exc_info` also needs to be removed from here

ajmaddaford (changes requested):

.pylintrc
benjefferies: 👍  Done

ajmaddaford (approved):

jonnyshaw89 (approved):

PR #848, opened 2017-01-17
Add test that checks for duplicate keys in schemas.

samiwel:
Add a test which checks the schema files for any duplicate keys.
This PR addresses a tech debt task 367.

### How to review 
Please review the test in case there are any problems with the logic.
Otherwise observe that all schemas are checked by this test to ensure that no duplicate Id are present.

### DIFF
--- a/tests/app/data/test_schema_validation.py
+++ b/tests/app/data/test_schema_validation.py
@@ -52,12 +52,10 @@ class TestSchemaValidation(unittest.TestCase):
         # Certain keys need to be ignored to avoid false positives.
         ignored_keys = ['routing_rules', 'skip_condition']
 
-        logger.info('Checking schema files for duplicate keys.')
         for schema_file in schema_files:
             unique_id = []
             with open(os.path.join(settings.EQ_SCHEMA_DIRECTORY, schema_file), encoding="utf8") as file:
                 schema_json = json.load(file)
-                logger.info('Checking %s...' % schema_file)
                 for id_value in self._parse_id_values(schema_json, ignored_keys):
                     if id_value in unique_id:
                         self.fail('Duplicate Id found. schema: %s, id: %s' % (schema_file, id_value))
--- a/tests/app/data/test_schema_validation.py
+++ b/tests/app/data/test_schema_validation.py
@@ -46,6 +46,35 @@ class TestSchemaValidation(unittest.TestCase):
 
             self.fail("{} Schema Validation Errors.".format(len(errors)))
 
+    def test_no_duplicate_ids_in_schema(self):
+        schema_files = schema_loader.available_schemas()
+        ignored_keys = ['routing_rules', 'skip_condition']
+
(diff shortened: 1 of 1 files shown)


tests/app/data/test_schema_validation.py
```diff
@@ -46,6 +46,35 @@ def test_schema(self):
 
             self.fail("{} Schema Validation Errors.".format(len(errors)))
 
+    def test_no_duplicate_ids_in_schema(self):
+        schema_files = schema_loader.available_schemas()
+        ignored_keys = ['routing_rules', 'skip_condition']
+
+        print('Checking schema files for duplicate keys.')
```
benjefferies: Let's use the logger

tests/app/data/test_schema_validation.py
```diff
@@ -46,6 +46,35 @@ def test_schema(self):
...
+        for schema_file in schema_files:
+            unique_id = []
+            with open(os.path.join(settings.EQ_SCHEMA_DIRECTORY, schema_file), encoding="utf8") as file:
+                schema_json = json.load(file)
+                print('Checking %s...' % schema_file)
```
benjefferies: Let's use the logger

tests/app/data/test_schema_validation.py
```diff
@@ -46,6 +46,35 @@ def test_schema(self):
...
+                for id_value in self._parse_id_values(schema_json, ignored_keys):
+                    if id_value in unique_id:
+                        self.fail('Duplicate Id found. schema: %s, id: %s' % (schema_file, id_value))
+                    else:
+                        unique_id.append(id_value)
+
+    def _parse_id_values(self, schema_json, ignored_keys):
+        for k, v in schema_json.items():
+            if k == 'id':
+                yield schema_json[k]
```
benjefferies: yield v?

tests/app/data/test_schema_validation.py
```diff
@@ -46,6 +46,35 @@ def test_schema(self):
...
+            elif k in ignored_keys:
```
benjefferies, line 69: I guess it doesn't matter much if we just evaluate all keys. The performance impact is going to be unnoticeable?

benjefferies (changes requested):

tests/app/data/test_schema_validation.py
samiwel: It isn't a performance optimisation, we genuinely need to skip routing_rules and skip_conditions or we'll get false-positives in terms of duplicate Ids.

benjefferies: Oh yeah doh! Might would worth giving them a name to reflect that. That's a tough ask so might be worth a comment to state they contain id references which are likely to contain duplicates

benjefferies (approved):

LJBabbage (approved):

PR #850, opened 2017-01-18
Removing splunk integration

benjefferies:
As a part of reviewing the logging we don't use the splunk integration hence we are removing it.

Related to [structured logging](https://github.com/ONSdigital/eq-survey-runner/pull/846)

### How to review 
Tests pass


LJBabbage (approved):

samiwel:
Will this not be needed going forward? Or are the splunk team only concerned with logging SDX traffic?
Or is splunk just slurping everything in from Syslog?

samiwel (approved):

benjefferies:
To clarify we don't actually integrate with splunk via this integration or at all yet. I think the plan is to log to syslog or sysout and something collects and forwards the logs on to splunk

PR #854, opened 2017-01-19
Eq 847 remove jquery

bennpearson:
Removed jQuery because it is only used for one pattern so have replaced this with javascript and used polyfills.js to add support to older browsers.  

### How to review 
Check focus on radio and checkbox inputs.
Also check browsers that don't support the following methods:
- https://developer.mozilla.org/en-US/docs/Web/API/Element/closest
- https://developer.mozilla.org/en/docs/Web/API/Element/classList

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -26,6 +26,7 @@
     "consistent-return": 0,
     "no-underscore-dangle": 0,
     "indent": [2, 2, {"SwitchCase": 1}],
-    "space-before-blocks": [2, "always"]
+    "space-before-blocks": [2, "always"],
+    "no-var": [1]
   }
 }
--- a/app/assets/js/app/modules/focus-styles.js
+++ b/app/assets/js/app/modules/focus-styles.js
@@ -1,4 +1,5 @@
 import domready from './domready'
+import {getNextSiblings, getPreviousSiblings} from '../helpers/selectors'
 
 const focusableClass = 'js-focusable'
 const focusableBoxClass = 'js-focusable-box'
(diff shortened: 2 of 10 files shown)


app/assets/js/app/modules/focus-styles.js
```diff
@@ -7,23 +7,49 @@ const checkedClass = 'is-checked'
...
-        // uncheck siblings
-        $closestFocusableBox.siblings(`.${focusableBoxClass}`).removeClass(checkedClass)
+  const checkedElements = Array.from(document.getElementsByClassName(focusableClass))
+  for (let checkedElement of checkedElements) {
+    if (checkedElement.checked === true) {
+      checkedElement.closest(`.${focusableBoxClass}`).classList.add(checkedClass)
+    }
+  }
+  function getNextSiblings(e, filter) {
+    var siblings = []
```
hamishtaplin: don't use `var`—use `let` or `const` where appropriate

might be worth adding a linting rule for this too: http://eslint.org/docs/rules/no-var

app/assets/js/app/modules/focus-styles.js
```diff
@@ -7,23 +7,49 @@ const checkedClass = 'is-checked'
...
-      if ($this.attr('type') === 'radio') {
-        // uncheck siblings
-        $closestFocusableBox.siblings(`.${focusableBoxClass}`).removeClass(checkedClass)
+  const checkedElements = Array.from(document.getElementsByClassName(focusableClass))
+  for (let checkedElement of checkedElements) {
+    if (checkedElement.checked === true) {
+      checkedElement.closest(`.${focusableBoxClass}`).classList.add(checkedClass)
+    }
+  }
+  function getNextSiblings(e, filter) {
```
hamishtaplin: move these into a helper called selectors and then `import`

hamishtaplin (changes requested):

hamishtaplin:
Still need to add [linting rule](http://eslint.org/docs/rules/no-var) to discourage use of `var` through a warning.

hamishtaplin (approved):

.eslintrc
```diff
@@ -26,6 +26,7 @@
     "consistent-return": 0,
     "no-underscore-dangle": 0,
     "indent": [2, 2, {"SwitchCase": 1}],
-    "space-before-blocks": [2, "always"]
+    "space-before-blocks": [2, "always"],
+    "no-var": [1]
```
benjefferies, line 30: Is this something we want to turn off globally or just ignore on the case by case scenario?

benjefferies (changes requested):

.eslintrc
hamishtaplin: @benjefferies There's no reason to use `var` unless you specifically need to take advantage of the way it's scoped—which is highly unlikely.

benjefferies (approved):

.eslintrc
benjefferies: Though this was disabling it as I thought it was enabled by default

PR #855, opened 2017-01-20
eq-848 optimise JS bundle

hamishtaplin:

Previously, the whole of lodash and babel-runtime were being bundled and served to users. This PR implements rollupify—a browserify plugin based on Rollup that uses "tree-shaking" to remove unused code.

#### Before

bundle.min.js: 173kb (48kb gzip'd)

#### After

bundle.min.js: 100kb (37kb gzip'd)

### How to review 

- `yarn compile` to compile
- check all functional tests still pass

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,3 @@
 {
-  "presets": ["es2015", "stage-2"]
+  "presets": ["es2015", "stage-2"],
 }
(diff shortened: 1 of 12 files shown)


bennpearson (approved):

benjefferies (commented):
Is there a measurable benefit to this optimisation?

hamishtaplin:
@benjefferies Less kb sent to users is always a good thing. It's difficult to measure accurately as we don't implement gzip locally but it should make a difference on poor connections.

benjefferies:
Cool makes sense seeing the stats

PR #857, opened 2017-01-20
Eq 857 update Travis configuration to use Python 3.4

samiwel:
Changed Travis configuration to use the Python 3.4 interpreter so that it more closely resembles what's available in Elastic Beanstalk (3.4.3).
Updated some tests that used features not available in the 3.4 standard libraries.

### How to review 
All tests should pass and Travis build should succeed.

### DIFF
--- a/tests/app/data/test_schema_validation.py
+++ b/tests/app/data/test_schema_validation.py
@@ -1,7 +1,10 @@
+
+from json import load
+
+import json
 import logging
 import os
 import unittest
-from json import load, JSONDecodeError
 
 import jsonschema
 from jsonschema import validate
@@ -25,18 +28,12 @@ class TestSchemaValidation(unittest.TestCase):
         schema = load(schema_file)
 
         for file in files:
+            json_file = open(os.path.join(settings.EQ_SCHEMA_DIRECTORY, file), encoding="utf8")
+            json_to_validate = json.load(json_file)
 
-            try:
-                json_file = open(os.path.join(settings.EQ_SCHEMA_DIRECTORY, file), encoding="utf8")
-                json_to_validate = load(json_file)
-
-                errors.extend(self.validate_json_against_schema(file, json_to_validate, schema))
-
-                errors.extend(self.validate_schema_contains_valid_routing_rules(file, json_to_validate))
-
-                errors.extend(self.validate_routing_rules_has_default_if_not_all_answers_routed(file, json_to_validate))
-            except JSONDecodeError as e:
-                errors.append("JSON Decode Error! Could not decode [{}]. Error [{}]".format(file, e))
+            errors.extend(self.validate_json_against_schema(file, json_to_validate, schema))
+            errors.extend(self.validate_schema_contains_valid_routing_rules(file, json_to_validate))
+            errors.extend(self.validate_routing_rules_has_default_if_not_all_answers_routed(file, json_to_validate))
 
         if errors:
             for error in errors:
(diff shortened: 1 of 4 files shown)


tests/app/data/test_schema_validation.py
```diff
@@ -83,8 +75,6 @@ def validate_json_against_schema(file, json_to_validate, schema):
             return []
         except jsonschema.exceptions.ValidationError as e:
             return ["Schema Validation Error! File [{}] does not validate against schema. Error [{}]".format(file, e)]
-        except JSONDecodeError as e:
```
jonnyshaw89, line 84: should we not at least add a default catch?
otherwise we will only ever return the 1st error?

jonnyshaw89 (changes requested):

tests/app/data/test_schema_validation.py
samiwel: I guess I could and it might lead to a cleaner error handling, but won't an unhandled exception fail the test with a stack trace anyway?

jonnyshaw89: it will but the functionality before displayed the errors with all the schemas on the 1st run. if all the schemas are broken this will currently only show you the 1st

samiwel: In fact, I'd argue that if something else fails here (other than schema validation) then we're unlikely to be able to recover from it, so not sure what benefits a catch-all exception would provide.

samiwel: Okay, I see, yeah in that case I guess it makes sense. I'll add a handler to trap the error so that at least other schemas get parsed under the same test.

jonnyshaw89 (approved):

ajmaddaford (approved):

samiwel:
Just a heads up, once merged developers should update their local development environments to use Python 3.4. This will help to avoid introducing new features that won't be supported when built on Travis or run in AWS. This can be achieved quite easily using Python virtual environments.

tests/app/data/test_schema_validation.py
benjefferies: I find that the output at the moment in time doesn't really allow me to resolve more than one schema at a time if it fails because it's too verbose

tests/integration/create_token.py
```diff
@@ -23,11 +23,11 @@ def create_token(form_type_id, eq_id, start_date=None, end_date=None, employment
...
+    payload_vars['form_type'] = form_type_id
+    payload_vars['region_code'] = region_code
+    payload_vars['language_code'] = language_code
 
-    payload = create_payload(**payload_vars,
-                             eq_id=eq_id,
-                             form_type=form_type_id,
-                             region_code=region_code,
-                             language_code=language_code)
+    payload = create_payload(**payload_vars)
```
benjefferies, line 31: 👍

tests/integration/questionnaire/test_questionnaire_previous_link.py
```diff
@@ -31,7 +31,7 @@ def test_previous_link_appears_on_page_following_introduction(self):
         block_one_url, resp = self.postRedirectGet(base_url + 'introduction', post_data)
 
         content = resp.get_data(True)
-        self.assertNotRegexpMatches(content, 'Previous')
+        self.assertNotRegex(content, 'Previous')
```
benjefferies, line 34: Also could be assertNotIn which is easier to read

benjefferies (approved):
Looks good minor point that doesn't need to be fixed

PR #867, opened 2017-01-23
Change converter to use routing path.

### LINKED ISSUE #858: When I change my route all answers are being submitted
### Expected behaviour
When I change my route after answering a question (i.e go back and choose a different route) the submitter should not send any of the old routes data.
### Actual behaviour
All data is being sent regardless of the route
1. Expose the message sent by the submitter, a simple logger message in send_answers will do it
2. Using test_conditional_routing, select Yes for the first question, fill out the next question so you are on the summary, now go back to the first question and change your answer to no i prefer tea, complete the survey, see that both responses for the 2nd questions are being passed

samiwel:
Fixes #858.

### How to review 
Please test the defect as described and observe only the answers from the route the user takes are output.
All existing tests should pass and the structure of the output should still be qcode -> answer value as before for any surveys with data version 0.0.1.

### DIFF
--- a/app/submitter/converter.py
+++ b/app/submitter/converter.py
@@ -60,7 +60,7 @@ def convert_answers(metadata, questionnaire, answer_store, routing_path):
     if questionnaire.data_version == '0.0.2':
         payload['data'] = convert_answers_to_census_data(answer_store, routing_path)
     elif questionnaire.data_version == '0.0.1':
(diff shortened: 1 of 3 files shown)


iwootten:
This looks good to me, but seems to be failing pylint checks.

iwootten (approved):

jonnyshaw89 (approved):

PR #869, opened 2017-01-24
Limit number of repeating blocks

### LINKED ISSUE #860: No limit on repeating sections count
### Expected behaviour
- There is an upper limit on the number of repeats in a repeating section

### Actual behaviour
- There is no limit, you can repeat 1,000,000 times if wanted

- Use census_household.json
- Get to Question 4
- Answer 10000000 for visitors
- Break into code and see that code is looping in `get_location_path`

### Technical information
Problem here:

```python
            no_of_repeats = 1

            if repeating_rule:
                no_of_repeats = evaluate_repeat(repeating_rule, self.answer_store)

            for i in range(0, no_of_repeats):  # << Big numbers here are bad news!
                blocks.extend([{
                    "group_id": group['id'],
                    "group_instance": i,
                    "block": block,
                } for block in group['blocks']])
        return blocks
```

jonnyshaw89:
This PR fixes https://github.com/ONSdigital/eq-survey-runner/issues/860 by adding a limit to the number of times a block can repeat.
A limit of 50 was chosen after analysis to find the level at which the system started to noticeably degrade in performance

### How to review 
Launch census household survey.
Enter a large number in number of visitors question.
Observe that the system performance is not dramatically effected.

### DIFF
--- a/app/questionnaire/rules.py
+++ b/app/questionnaire/rules.py
@@ -1,3 +1,9 @@
+import logging
+from app import settings
+
+logger = logging.getLogger(__name__)
+
+
 def evaluate_rule(when, answer_value):
     """
     Determine whether a rule will be satisfied based on a given answer
@@ -52,7 +58,13 @@ def evaluate_repeat(repeat_rule, answer_store):
         repeat_index = repeat_rule['answer_id']
         filtered = answer_store.filter(answer_id=repeat_index)
         repeat_function = repeat_functions[repeat_rule['type']]
-        return repeat_function(filtered)
+        no_of_repeats = repeat_function(filtered)
+
+        if no_of_repeats > settings.EQ_MAX_NUM_REPEATS:
+            logger.warning('Excessive number of repeats found [%s] capping at [%s]', no_of_repeats, settings.EQ_MAX_NUM_REPEATS)
+            no_of_repeats = settings.EQ_MAX_NUM_REPEATS
+
+        return no_of_repeats
 
 
 def _get_answer_value():
--- a/app/settings.py
+++ b/app/settings.py
@@ -29,7 +29,10 @@ def get_key(_key_name):
 
 EQ_MINIMIZE_ASSETS = parse_mode(os.getenv('EQ_MINIMIZE_ASSETS', 'False'))
 # max request payload size in bytes
-EQ_MAX_HTTP_POST_CONTENT_LENGTH = 65536
+EQ_MAX_HTTP_POST_CONTENT_LENGTH = os.getenv('EQ_MAX_HTTP_POST_CONTENT_LENGTH', 65536)
+
+# max number of repeats from a rule
+EQ_MAX_NUM_REPEATS = os.getenv('EQ_MAX_NUM_REPEATS', 50)
 
 EQ_PROFILING = parse_mode(os.getenv('EQ_PROFILING', 'False'))
 
--- a/README.md
+++ b/README.md
@@ -199,6 +199,8 @@ SAUCE_ACCESS_KEY - Sauce Labs private key
 EQ_DEV_MODE - Enable dev mode
 EQ_ENABLE_FLASK_DEBUG_TOOLBAR - Enable the flask debug toolbar
 EQ_ENABLE_CACHE - Enable caching of the schema
+EQ_MAX_HTTP_POST_CONTENT_LENGTH - The maximum http post content length that the system wil accept
+EQ_MAX_NUM_REPEATS - The maximum number of repeats the system will allow
 ```
 ## Loading schemas from S3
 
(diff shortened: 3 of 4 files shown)


app/questionnaire/path_finder.py
```diff
@@ -180,6 +180,10 @@ def get_blocks(self):
             if repeating_rule:
                 no_of_repeats = evaluate_repeat(repeating_rule, self.answer_store)
 
+            if no_of_repeats > 50:
```
LJBabbage: Can we have this in a settings file please (the count on the number of repeats allowed)

samiwel:
Looks good, just wondering if there should be a test for the case when someone adds more than 50 answers to assert that only 50 are created?

app/questionnaire/path_finder.py
```diff
@@ -180,6 +180,10 @@ def get_blocks(self):
...
+                logger.error('Excessive number of repeats found [%s] capping at 50', no_of_repeats)
+                no_of_repeats = 50
```
LJBabbage: Not sure this is what we want, if they have entered over 50, only repeat for 50. Ideally on a repeating feeding answer we should show an error if over 50 to make sure they just haven't made a mistake before this

LJBabbage (changes requested):

app/questionnaire/path_finder.py
samiwel: I think the original household limit was supposed to be 24 - should it be the same for visitors?

jonnyshaw89:
@samiwel the way the tests currently work there would be nothing to prevent me entering more than 50 answers. But the question will never be shown on the frontend. This test shows that the block is never repeated more than 50 times so the question stops being asked.

I is probably worth mentioning that this is a temporary fix which will stop the system crashing while we look into a solution on how the routing path is generated.

app/questionnaire/path_finder.py
LJBabbage: Should it also not be done in the evaluate_repeat function?

jonnyshaw89: This limit is designed to be a system limit not a census limit on household members or visitors.

app/questionnaire/path_finder.py
jonnyshaw89: As previously mentioned this is just enforcing a system limit to prevent crashing. The validation that you are talking about is a survey specific validation.

app/questionnaire/path_finder.py
benjefferies: In my opinion for now a hard coded limit would be okay. When we have the scenario where we need to have varying number of repeats based on question we should look at making it configurable.

It might be good to localise this logic to the evaluate_repeat function

app/questionnaire/path_finder.py
```diff
@@ -180,6 +180,10 @@ def get_blocks(self):
             if repeating_rule:
                 no_of_repeats = evaluate_repeat(repeating_rule, self.answer_store)
 
+            if no_of_repeats > 50:
+                logger.error('Excessive number of repeats found [%s] capping at 50', no_of_repeats)
```
benjefferies: This seems more like a warn level log to me

benjefferies (changes requested):

app/questionnaire/path_finder.py
jonnyshaw89: localising the logic to evaluate_repeat wont fix the performance impact as this function will still build up the massive arrays multiple times per request.

app/questionnaire/rules.py
```diff
@@ -52,7 +57,13 @@ def evaluate_repeat(repeat_rule, answer_store):
         repeat_index = repeat_rule['answer_id']
         filtered = answer_store.filter(answer_id=repeat_index)
         repeat_function = repeat_functions[repeat_rule['type']]
-        return repeat_function(filtered)
+        no_of_repeats = repeat_function(filtered)
+
+        if no_of_repeats > 50:
```
collisdigital: Would be good if this was configurable, ENV var at least so we can change the behaviour easily if it causes a problem as this feels like a system-wide imposed limit (like the maximum size of a POST data allowed etc.)

app/settings.py
```diff
@@ -31,6 +31,9 @@ def get_key(_key_name):
 # max request payload size in bytes
 EQ_MAX_HTTP_POST_CONTENT_LENGTH = 65536
 
+# max number of repeats from a rule
+EQ_MAX_NUM_REPEATS = 50
```
collisdigital: Update README

ajmaddaford (approved):

LJBabbage (approved):

benjefferies (approved):

PR #871, opened 2017-01-24
Spike flush data

LJBabbage:
To get a first look at flushing data from a survey. The spike was initially undertaken to investigate whether it was possible, which it is. However, how it should work is still up for discussion.

As long as we have the eq_id, form_type, collection_exercise_sid and ru_ref we can recreate the user_id and ik meaning we can get the data out of the database for any user, without needing any session stored keys. 

At the moment I have put a flag on the JWT and dealt with it in session.py, however I also experimented with making an end point and that was just as effective. A middle ground is probably the best solution, with a flag in the JWT passing to an endpoint to do the flushing.
The JWT seems to be the way to go as it offers authentication but the rest is up for debate

N.B You can see it working by using /dev and ticking the flush data checkbox. I have left a logger statement in the code on purpose so the flushed data can be seen for this spike

### How to review 
No review required, just general comments and idea's how it should work

### DIFF
--- a/app/parser/metadata_parser.py
+++ b/app/parser/metadata_parser.py
@@ -59,6 +59,7 @@ metadata_fields = {
     "language_code": MetadataField(mandatory=False),
     "tx_id": MetadataField(mandatory=False, validator=uuid_4_parser, generator=id_generator),
     "variant_flags": MetadataField(mandatory=False),
+    "flush_data": MetadataField(mandatory=False),
 }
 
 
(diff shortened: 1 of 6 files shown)


collisdigital:
As discussed; we'd also benefit from he ability to not just flush but also clear a respondents data using the same/similar mechanism (e.g if their data is broken in some way) so we should think about that as part of the discussion/approach.

collisdigital:
Andrew's idea was the add to the JWT a list of endpoints that the authentication provides permissions to access and add a /clear /flush etc. Endpoint that can only be accessed if the JWT gave permission to prevent regular users just hitting them. Combined with this the JWT could contain an 'action' to take e.g. go to /flush.

ajmaddaford:
Spike has proven what it needed to, a card has been created for the implementation of this.

PR #872, opened 2017-01-24
eq-861 Answer Label Incorrect Semantics

hamishtaplin:

Fixes #872 by replacing the semantically incorrect `label` element with a `legend`.

- Changed logic in question template so that, if a fieldset/legend is used the legend is hidden rather than replacing the `h2` on the question title.
- Added `field__legend` styling for new legend element.
- Legend falls back to repeating question title if no "answer label" is supplied. Also visually hidden in that case.
- Improved semantics of UKIS survey by moving "Select all that apply" out of answer label (and therefore legend) and into question description wherever possible.

### How to review 

- start UKIS survey
- Check that semantics are correct for as many questions as possible. IE. a `legend` is used where there are fields with multiple `input` (ie radios/checkboxes)

### DIFF
--- a/app/assets/styles/partials/components/_field.scss
+++ b/app/assets/styles/partials/components/_field.scss
@@ -2,6 +2,17 @@
   background: none;
 }
 
+.field__legend {
+  font-size: 1rem;
+  font-weight: $font-weight-bold;
+  margin-bottom: 0.5rem;
+}
+
+.field__description {
+  line-height: 1.3;
+  display: block;
+}
+
 .field__item {
   margin: 0 0.5rem 0.5rem 0;
   position: relative;
--- a/app/assets/styles/partials/components/_time.scss
+++ b/app/assets/styles/partials/components/_time.scss
@@ -1,0 +1,25 @@
+.time {
+  position: relative;
+}
+
+.time__visual {
+  position: relative;
+  overflow: visible !important;
+  margin: 3rem auto;
+  display: block;
+}
+
+.time__text {
+  position: absolute;
+  top: 0;
+  right: 0;
+  bottom: 0;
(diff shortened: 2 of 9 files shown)


hamishtaplin:
The main issue with this is that it now creates an unwanted split between question title and the legend by having the guidance inbetween in some places.

screenshot 2017-01-31 13 57 41

collisdigital:
@hamishtaplin what affect does this have an all other questionnaires?

hamishtaplin:
@collisdigital no visible affect as far as I can tell but it does warrant some testing. Existing surveys don't use the answer label so the (hidden) legend just inherits the question title, as it did anyway.

collisdigital:
What testing and who would need to do it?

hamishtaplin:
@collisdigital Testing the other surveys to make sure they are not negatively affected. I'm pretty sure they aren't having been through Census but it warrants a fresh pair of eyes. Anyone can do it.

collisdigital:
Perfect, when can you start?

hamishtaplin:
@collisdigital I've already taken a look, someone else needs to double-check things. Especially the UKIS survey as I'm not as familiar with it.

yoganandkunche:
@hamishtaplin @collisdigital Tested all the surveys (MCI, RSI, Census & UKIS) by going through all the survey pages nothing seems to be broken visually or content wise. Also did test using voice over with different question times and work fine. However I guess we are going to have another set of DAC testing to prove the accessibility side of it.

bennpearson (approved):

hamishtaplin:
As agreed with Will, I moved any guidance that is splitting the question in two (as above) to the hidden guidance underneath answers.

PR #873, opened 2017-01-24
Safari scroll issue

### LINKED ISSUE #862: Scroll doesn't work using Safari
### Expected behaviour
Scrolling down the page should work in safari

### Actual behaviour
Scrolling doesn't work in safari

- Use Safari browser
- Select UKIS 1_0001.json schema
- Try scrolling down the page

#### Browser
Safari 10.0.1

bennpearson:
Fixes #862 Scroll issue in safari 

### How to review 

- Use Safari
- Select UKIS 1_0001.json schema
- Try scrolling down the page

### DIFF
--- a/app/assets/styles/partials/objects/_page.scss
+++ b/app/assets/styles/partials/objects/_page.scss
@@ -4,6 +4,7 @@
   flex-direction: column;
   width: 100%;
   overflow-x: hidden;
+  overflow-y: visible;
 }
 
 .page__header {
@@ -47,8 +48,12 @@
 
 .page__content {
   flex: 1 0 auto;
+  // NOTE: padding-bottom: 1px fixes safari scroll issue
+  // Issue number #862
+  padding-bottom: 1px;
   position: relative;
   overflow-x: hidden;
+  overflow-y: visible;
 }
 
 .page__nav.nav {


app/assets/styles/partials/objects/_page.scss
```diff
@@ -47,8 +48,10 @@
 
 .page__content {
   flex: 1 0 auto;
+  padding-bottom: 1px;
```
hamishtaplin, line 53: Can we at least add a comment in the code with a link to this bug so we know why this seemingly arbitrary bit of padding his here?

hamishtaplin (changes requested):

hamishtaplin (approved):

ajmaddaford:
@bennpearson I don't get this issue on master, I'm using Safari on a Mac
image

hamishtaplin:
@ajmaddaford It's a definite issue, I can reproduce—it happens on the introduction page.

ajmaddaford:
@hamishtaplin I believe it's an issue, but I couldn't test that this fixes it! I do get an additional scroll bar without this fix, and it fixes that.

ajmaddaford (approved):

PR #874, opened 2017-01-25
Fixes 'Enter' key deletes person on household composition screen.

### LINKED ISSUE #859: Using 'Enter' key on Household Composition removes people
### Expected behaviour
- People aren't removed when hitting enter

### Actual behaviour
- People are removed
- Behaviour is inconsistent, sometimes Person 2 is removed regardless of who is being edited
- Sometimes hitting enter submits the form

- Add a load of people
- Select Person 4's last name and hit enter, see who gets removed
- Select a different person and different field, see what happens
- Save and Continue and then go back
- Form behaves differently now that people are pre-rendered (not JS created).

#### Browser
Chrome

#### Operating System
macOS Sierra

### Screenshot
N/A

samiwel:
Fixes #859 .

### How to review 
Re-test the defect, confirm that person is not deleted when Enter key is pressed.
The enter key will now behave as a 'Save and continue' operation, it will submit the form and navigate to the next page.

Front end devs:
Please can you check to ensure that my solution is in the correct place and that it adheres to front end best practices etc. 
Happy to learn the best way to do things.

### DIFF
--- a/app/templates/partials/questions/repeatinganswer.html
+++ b/app/templates/partials/questions/repeatinganswer.html
@@ -20,7 +20,7 @@
           {%- if loop.index == 1 -%}
              <span class="js-household-action">{{ _('You') }}</span>
           {%- else -%}
-            <button class="btn btn--link pluto js-btn-remove" name="action[remove_answer]" value="{{ loop.index0 }}" type="submit">{{ _('Remove') }}
+            <button class="btn btn--link pluto js-btn-remove" name="action[remove_answer]" value="{{ loop.index0 }}" type="button">{{ _('Remove') }}
               <span class="u-vh">{{ _('Person') }} {{loop.index0}}</span>
             </button>
           {%- endif -%}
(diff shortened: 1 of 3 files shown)


hamishtaplin:
We shouldn't be hijacking default keyboard behaviour here. If we remove the `type=submit` attribute from the `button` it won't submit the form.

hamishtaplin (changes requested):
We shouldn't be hijacking default keyboard behaviour here. If we remove the `type=submit` attribute from the `button` it won't submit the form.

samiwel:
@hamishtaplin does the call to `preventDefault` remove the `type=submit` from the `button`? I expected this to only override the Enter key press of the text `input` fields. Would it be better to narrow down the selector to ensure that only text inputs are included?

hamishtaplin:
@samiwel I don't think you need to use JS at all here, just remove the `type=submit` attribute from [this line in the template](https://github.com/ONSdigital/eq-survey-runner/blob/master/app/templates/partials/questions/repeatinganswer.html#L23):

samiwel:
@hamishtaplin Ah! Good shout. Thanks for pointing that out. I've changed the remove button to `type=button`. If I remove it completely it defaults with a submit action (which removes the person).

hamishtaplin (approved):

ajmaddaford (approved):

PR #875, opened 2017-01-25
Eq 863 highlight styles

### LINKED ISSUE #863: Highlight style of emphasis should be limited to question title
### Expected behaviour

- The "highlight" style applied to `em` should only be used in question title, with bold being used elsewhere.

### Actual behaviour

- Highlight style is used throughout

- open UKIS `1_0001.json`
- highlight is used throughout in text that is not question title

### Screenshot

screenshot 2017-01-23 09 51 33

bennpearson:
Fixes #863  
Removed highlight background colour from all elements except `.question__title` 

### How to review 

- Select 1_0001.json schema
- Check highlight background colour only exists on `.question__title`

### DIFF
--- a/app/assets/styles/partials/base/_typography.scss
+++ b/app/assets/styles/partials/base/_typography.scss
@@ -43,5 +43,6 @@ li {
 }
 
 em {
-  @extend .highlight;
+  font-style: normal;
+  font-weight: 600;
 }
--- a/app/assets/styles/partials/base/_typography.scss
+++ b/app/assets/styles/partials/base/_typography.scss
@@ -44,5 +44,5 @@ li {
 
 em {
   font-style: normal;
-  font-weight: 600;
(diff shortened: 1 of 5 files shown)


hamishtaplin (changes requested):
This will break the use of `.highlight` as a standalone class as you've removed the background colour from it.

app/assets/styles/partials/base/_typography.scss
```diff
@@ -43,5 +43,6 @@ li {
 }
 
 em {
-  @extend .highlight;
+  font-style: normal;
+  font-weight: 600;
```
hamishtaplin: use `$font-weight-bold` here instead of 600

hamishtaplin (changes requested):

hamishtaplin (approved):

PR #876, opened 2017-01-25
Changed nav background colour

### LINKED ISSUE #864: Wrong colour used in nav on business survey themes
### Expected behaviour

- the colour used to highlight the current navigation item on business surveys should be very light blue

### Actual behaviour

- the colour is a mid blue

- start UKIS survey
- navigate to first question

### Screenshot

screenshot 2017-01-23 11 22 36

bennpearson:
Fixes #864 

### How to review 
Use both census and ukis schemas to check background colour of navigation

### DIFF
--- a/app/assets/styles/partials/components/_nav.scss
+++ b/app/assets/styles/partials/components/_nav.scss
@@ -51,7 +51,7 @@
     }
   }
   .nav__item--current {
-    background-color: lighten($color-secondary, 64%);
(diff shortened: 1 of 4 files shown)


hamishtaplin (changes requested):
The background colour of `.panel--info` should also use the slightly darker blue of the nav (`#f5f6ff`). Might be worth checking to see if anything else can use it too.

hamishtaplin (approved):

PR #877, opened 2017-01-26
Make BEIS and DoF logos visible for NIR respondents.

samiwel:
BEIS and the NI department of finance have requested that the NI department of finance and BEIS logos are present on the NI UKIS questionnaire landing page.

This PR ensures that both logos are visible when UKIS survey is launched with a Northern Ireland Region code.

Here is what Northern Ireland respondents should see:
image

Here is what respondents from other regions should see:
image

### How to review 
The UKIS landing page has both the NI and BEIS logo present when launched in the Northern Ireland region.
The NI logo is not present in non-Northern Ireland regions (i.e. just the BEIS logo is shown)


collisdigital:
Can you attach a screenshot. Note the BEIS logo has changed (see Trello).

samiwel:
Sure, I'll attach a screenshot once I've replaced the logo.

app/themes/ukis/templates/introduction.html
```diff
@@ -12,9 +12,10 @@
 {% block business_details %}
   <div class="u-mb-m">
     {% if meta.survey.region_code == 'GB-NIR' %}
+      <img src="{{ theme_static('logo-beis.png') }}" alt="Department for Business, Energy and Industrial Strategy" width="200" />
```
collisdigital: If you swap the order of the IF statement you could always have the BEIS logo and only add the NI logo inside the IF; no need for an ELSE.

jonnyshaw89 (approved):

ajmaddaford:
The `gy` on the end of 'Strategy' in the BEIS logo is slightly cut off for me
image

hamishtaplin:
Can we get this closer to the design? There should be a margin between the logos and attention paid to how they stack on mobile. Also, the proportions of the logos are off—the BEIS one is too big and the file needs to be scaled down so we're not shipping a 900px wide image only to display it at 220px.

hamishtaplin (changes requested):
Can we get this closer to the design? There should be a margin between the logos and attention paid to how they stack on mobile. Also, the proportions of the logos are off—the BEIS one is too big and the file needs to be scaled down so we're not shipping a 900px wide image only to display it at 200px. Ideally it would be re-sized to 400px wide so it scales nicely for @2x display.

samiwel:
image

I've updated the styling so that it more closely resembles the design.
I've also reduced the size of the image to 400px and removed the inline CSS.

On mobile it stacks like this:

image

hamishtaplin (approved):

PR #885, opened 2017-01-27
EQ-872: Update to use WTForms

iwootten:
Detach the current implementation of state/questionnaire_manager and replace with WTForms equivalents. Create custom forms for Census Household Compositions/Relationship and add the concept of parent/child answers to schemas to allow for "other" options to be correctly represented.

Update functional tests as per new wtforms radio option indexes and integration tests with consistent non-zero padded date values (due to wtforms validation of data that is not in a select).

This still includes the state and schema classes, mainly because our parsing depends on them, so the parser utils class will need updating to though exceptions without them being available. I'm working on an update to remove the widget classes, though this depends on state not referencing them.

Some further refactoring:

- A class for generating metaclasses: Move metaclass creation into their own classes, rather than top-level methods. This was fine at first, but due to the inheritance of top level messages, it's necessary to pass them around right down to a field level. 

### How to review 
High level design feedback, ideas on further approach to removing schema/widgets/state. 

*This currently touches 379 files at the time of writing, so I'd prefer we focused on critical improvements rather than 'nice to haves'. It's becoming very difficult to rebase, due to files getting added back in from master. We can iterate on the change once it's in.*


app/data/0_star_wars.json
```diff
@@ -167,7 +167,7 @@
                                     ],
                                     "description": "",
                                     "id": "680f2ff9-d5a5-4057-b1cd-9fde2660b244",
-                                    "title": "A wise choice young Yedi. Pick your hero",
+                                    "title": "A wise choice young Jedi. Pick your hero",
```
jonnyshaw89, line 170: One for @collisdigital to review

app/data/census_household.json
```diff
@@ -1429,12 +1429,17 @@
                                                 {
                                                     "label": "Yes, an address outside the UK (please specify)",
                                                     "value": "Other",
-                                                    "other": {
-                                                        "label": "Please enter country"
-                                                    }
+                                                    "child_answer_id": "another-address-answer-other"
```
benjefferies, line 1432: Should we need to define the child and parent?

app/forms/date_form.py
```diff
@@ -0,0 +1,136 @@
...
+
+    :param error_messages: The messages during validation
+    :param answer: The answer on which to base this form
+    :param to_field_data: The data coming from the
+    :param validate_range: Whether the dateform should add a daterange validator
+    :return: The generated DateForm metaclass
+    """
+    class DateForm(Form):
+
+        MONTH_CHOICES = [('', 'Select month')] + [(str(x), calendar.month_name[x]) for x in range(1, 13)]
```
benjefferies: Would be useful to refactor this out with a comment or as a method as it's confusing code

app/forms/date_form.py
```diff
@@ -0,0 +1,136 @@
...
+        month = SelectField(choices=MONTH_CHOICES, default='')
+        year = StringField()
+
+    validate_with = [validators.optional()]
+
+    if not error_messages:
+        error_messages = {}
+
+    if answer['mandatory'] is True:
+        if 'validation' in answer and 'messages' in answer['validation'] \
```
benjefferies: Is the check for a mandatory message required. Surely it has to be defined or we fall back to the default. This applies to all of the checks to see if validation messages exist

app/data/schema/schema-v1.json
```diff
@@ -328,6 +328,9 @@
                                   "id": {
                                     "$ref": "#/definitions/id"
                                   },
+                                  "parent_answer_id": {
```
benjefferies, line 335: Are we missing child id?

app/templates/partials/answers/currency.html
```diff
@@ -1,3 +1,28 @@
...
-  {{ answer.schema_item.widget.render(answer)|safe}}
+    {% set label = {
+      'for': answer.id,
+      'text': answer.label
+    } %}
+
+    {% include 'partials/forms/label.html' %}
+
+    {% set raw_data = form[answer['id']].raw_data[0] if form[answer['id']].raw_data is not none else '' %}
+    {% set form_data = form[answer['id']].data if form[answer.id].data is not none else '' %}
```
benjefferies, line 10: As discussed maybe the `default` field could be used to avoid `X if None else ''` statements

app/forms/date_form.py
benjefferies: Failing that a generic way of adding default error messages would be good

app/forms/date_form.py
```diff
@@ -0,0 +1,136 @@
...
+    field_to = FormField(
+        get_date_form(answer=answer_to, error_messages=error_messages),
+        label=answer_to['label'] if 'label' in answer_to else '',
+        description=answer_to['guidance'] if 'guidance' in answer_to else '',
+    )
+
+    return field_from, field_to
+
+
+def get_date_data(form_data, answer_id):
```
benjefferies, line 121: Some tests for this would be good

app/forms/fields.py
```diff
@@ -0,0 +1,149 @@
...
+        "MonthYearDate": get_date_field,
+        "Currency": get_integer_field,
+        "Integer": get_integer_field,
+        "PositiveInteger": get_integer_field,
+        "Percentage": get_integer_field,
+        "TextArea": get_text_area_field,
+        "TextField": get_string_field,
+    }[answer['type']](answer, label, guidance, error_messages)
+
+    if field is None:
```
benjefferies, line 28: I guess this should never happen but if it does will it fall over later down the line? If yes would we be better of raising a meaningful error here or at least log at a higher level than info?

app/forms/date_form.py
```diff
@@ -0,0 +1,136 @@
...
+
+from wtforms import Form, FormField, SelectField, StringField
+from wtforms import validators
+
+from app.validation.validators import DateCheck, DateRangeCheck, DateRequired, MonthYearCheck
+
+logger = logging.getLogger(__name__)
+
+
+def get_date_form(answer=None, to_field_data=None, validate_range=False, error_messages=None):
```
benjefferies, line 12: It's worth looking to see if we can remove the methods by having an __init__ method which takes the relevant state. It makes it harder to understand what's going on when we have a method returning a class definition just to share state. It appears possible via the docs

app/forms/fields.py
```diff
@@ -0,0 +1,149 @@
...
+
+
+def build_choices(options):
+    choices = []
+    for option in options:
+        choices.append((option['value'], option['label']))
+    return choices
+
+
+def get_validators(answer, error_messages):
```
benjefferies: We should rename this to get_mandatory_validator

app/forms/fields.py
```diff
@@ -0,0 +1,149 @@
...
+    return validate_with
+
+
+def get_string_field(answer, label, guidance, error_messages):
+    validate_with = get_validators(answer, error_messages)
+
+    return StringField(
+        label=label,
+        description=guidance,
+        widget=TextArea(),
```
benjefferies: We aren't using any of the WTform widgets but the existing widgets so this won't be doing anything and can be removed

app/forms/fields.py
```diff
@@ -0,0 +1,149 @@
...
+
+        validate_with = [
+            validators.InputRequired(
+                message=mandatory_message,
+            ),
+        ]
+    return validate_with
+
+
+def get_string_field(answer, label, guidance, error_messages):
```
benjefferies, line 59: Should probably have tests for each of the fields

app/forms/fields.py
```diff
@@ -0,0 +1,149 @@
...
+
+def get_text_area_field(answer, label, guidance, error_messages):
+    validate_with = get_validators(answer, error_messages)
+
+    return TextAreaField(
+        label=label,
+        description=guidance,
+        widget=TextArea(),
+        validators=validate_with,
+        filters=[lambda x: x if x else None],
```
benjefferies: Should probably have a test to check this filter

app/forms/household_composition_form.py
```diff
@@ -0,0 +1,112 @@
...
+            error_message = error_messages['MANDATORY']
+            if 'validation' in answer and 'messages' in answer['validation'] \
+                    and 'MANDATORY' in answer['validation']['messages']:
+                error_message = answer['validation']['messages']['MANDATORY']
+
+            validate_with = [validators.InputRequired(
+                message=error_message,
+            )]
+
+        # Have to be set this way given hyphenated names
```
benjefferies, line 28: Is this comment the actual reason? I thought it was because we have a dynamic number of fields?

app/forms/household_composition_form.py
```diff
@@ -0,0 +1,112 @@
...
+        # are considered invalid in python
+        field = StringField(validators=validate_with)
+
+        setattr(NameForm, answer['id'], field)
+
+    return NameForm
+
+
+def serialise_composition_answers(location, data):
```
benjefferies, line 37: Does this need to be a separate method outside of HouseHoldCompositionForm.serialise?

app/forms/household_composition_form.py
```diff
@@ -0,0 +1,112 @@
...
+                location=location,
+                answer_id=answer_id,
+                answer_instance=index,
+                value=person_data[answer_id],
+            )
+            answers.append(answer)
+    return answers
+
+
+def deserialise_composition_answers(answers):
```
benjefferies, line 51: This makes sense to belong to the HouseHoldCompositionForm class with serialise

app/forms/household_composition_form.py
```diff
@@ -0,0 +1,112 @@
...
+    household = {}
+    answer_instance = 0
+
+    while any((a for a in answers if a['answer_instance'] == answer_instance)):
```
benjefferies: Could this not be simplified by sorting by answer_instance and then iterating over them? or I am wildly out in guessing what this block is trying to do

benjefferies (changes requested):
Think we're missing a lot of unit tests on the new code. I think the code coverage will still be high because the integration tests hit everything but we should have unit tests to cover scenarios.

On a wider note I don't think we should count coverage from the integration tests

app/data/census_household.json
iwootten: How else do you know if an answer has a parent, from an answers json? We could iterate the entire schema and build a map, but seems like quite an overhead when inspecting each answer.

app/forms/household_composition_form.py
iwootten: That's another reason too, yes.

app/forms/household_composition_form.py
iwootten: Yes, due to the complexity of the generate method.

app/forms/household_composition_form.py
iwootten: Same comment as above

app/data/census_household.json
benjefferies: Having to define a two way relationship seems like a configuration overhead with room for error. At the very least we'd need some validation to check this but going back to my point it seems like a configuration overhead if we know the mapping already

app/forms/household_composition_form.py
benjefferies: What is happening with the hyphens here I can't tie the two things together?

app/forms/household_composition_form.py
benjefferies: If we removed the methods and referenced classes wouldn't that reduce the complexity?

app/forms/date_form.py
iwootten: I've tried this and the short answer is it's a metaclass so behaves differently to standard class init and doesn't work.

benjefferies: Oh, did you try it with `__new__`?

app/forms/household_composition_form.py
iwootten: What do you mean by "referenced classes"?

app/forms/household_composition_form.py
iwootten: ids are comprised of strings which include hyphens, which are invalid variable names in python.

app/forms/household_composition_form.py
benjefferies: If we removed the methods and instead of invoking the method we create an instance of the form passing in what will be state like they do here http://wtforms.readthedocs.io/en/latest/forms.html

app/forms/date_form.py
iwootten: Yes, I think so. We don't want to change the validate_with validators unless it's mandatory....

benjefferies (changes requested):
From a high level this seems to be the right approach. I think there is outstanding refactoring and tidying up required and a deeper review to follow.

app/forms/date_form.py
iwootten: Done

app/forms/fields.py
iwootten: Done

app/forms/household_composition_form.py
iwootten: The error structure isn't customisable and depends on the form structure, so you couldn't manipulate it's structure through using form state. The forms state is populated using the data parameter supplied to the generate method.

app/forms/date_form.py
collisdigital: Might be worth considering a utility to do this kind of thing; like https://github.com/ONSdigital/eq-google-slides-to-schema/blob/master/utils.py

Usage: 

`if get_nested_dict_value(answer, 'validation', 'messages', 'MANADAOTRY'):`

app/forms/fields.py
```diff
@@ -0,0 +1,148 @@
...
+from app.forms.date_form import get_date_form, get_month_year_form
+from app.validation.validators import IntegerCheck, PositiveIntegerCheck, NumberRange
+from structlog import get_logger
+
+
+logger = get_logger()
+
+
+def get_field(answer, label, error_messages):
+    guidance = answer['guidance'] if 'guidance' in answer else ''
```
collisdigital: Or:

`guidance = answer.get('guidance', '')`

app/forms/fields.py
```diff
@@ -0,0 +1,148 @@
...
+    mandatory_or_optional = get_mandatory_validator(answer, answer_errors)
+    validate_with = mandatory_or_optional + [
+        IntegerCheck(messages=answer_errors),
+        NumberRange(max=9999999999, messages=answer_errors),
+    ]
+
+    if answer['type'] == 'Currency' or answer['type'] == 'PositiveInteger':
+        validate_with += [PositiveIntegerCheck(messages=answer_errors)]
+    elif answer['type'] == 'Percentage':
+        validate_with = mandatory_or_optional + [
```
collisdigital: This overrides the assignment on line 130, making those lines a bit redundant if it's a Percentage type ; re-ordering this if to be beforehand or similar re-jig would make this cleaner.

app/forms/household_composition_form.py
```diff
@@ -0,0 +1,112 @@
...
+                    ordered_errors += map_field_errors(field.errors, index)
+            return ordered_errors
+
+        def answer_errors(self, input_id):
+            return [error[1] for error in self.map_errors() if input_id == error[0]]
+
+        def remove_person(self, index_to_remove):
+            popped = []
+
+            while index_to_remove != len(self.household.data):
```
collisdigital: If the index to remove was ever more than the len  of the data  this could potentially loop forever; worth being defensive here somehow to protect against that? Maybe an assert.

app/forms/household_composition_form.py
```diff
@@ -0,0 +1,112 @@
...
+
+        def serialise(self, location):
+            """
+            Returns a list of answers representing the form data
+            :param location: The location to associate the form data with
+            :return:
+            """
+            return serialise_composition_answers(location, self.household.data)
+
+    return HouseHoldCompositionForm(MultiDict(data), meta={'csrf': False})
```
collisdigital, line 109: Are we ok with csrf being disabled?

collisdigital: If so can we add a comment explaining

app/forms/household_relationship_form.py
```diff
@@ -0,0 +1,93 @@
...
+def generate_relationship_form(block_json, number_of_entries, data, error_messages):
+    class HouseHoldRelationshipForm(FlaskForm):
+        def map_errors(self):
+            ordered_errors = []
+
+            if len(self.errors) > 0:
+                for answer_id, error_list in self.errors.items():
+                    for errors in error_list:
+                        for error in errors:
+                            ordered_errors.append((answer_id, error))
```
collisdigital, line 95: Consider yield instead to avoid building the whole list.

app/forms/household_relationship_form.py
```diff
@@ -0,0 +1,93 @@
...
+
+            return ordered_errors
+
+        def answer_errors(self, input_id):
+            return [error[1] for error in self.map_errors() if input_id == error[0]]
+
+    answer = SchemaHelper.get_first_answer_for_block(block_json)
+    guidance = answer['guidance'] if 'guidance' in answer else ''
```
collisdigital: Or 

`answer.get('guidance', '')`

app/forms/household_relationship_form.py
```diff
@@ -0,0 +1,93 @@
...
+        choices=build_choices(answer['options']),
+        widget=ListWidget(),
+        option_widget=RadioInput(),
+        validators=get_mandatory_validator(answer, error_messages),
+    ), min_entries=number_of_entries)
+
+    setattr(HouseHoldRelationshipForm, answer['id'], field)
+
+    if data:
+        form = HouseHoldRelationshipForm(MultiDict(data), meta={'csrf': False})
```
collisdigital, line 122: Same comments about csrf

app/forms/questionnaire_form.py
```diff
@@ -0,0 +1,100 @@
...
+
+def get_answer_fields(question, data, error_messages):
+    answer_fields = {}
+    for answer in question['answers']:
+        if 'parent_answer_id' in answer and answer['parent_answer_id'] in data and \
+                data[answer['parent_answer_id']] == 'Other':
+            answer['mandatory'] = \
+                next(a['mandatory'] for a in question['answers'] if a['id'] == answer['parent_answer_id'])
+
+        name = answer['label'] if 'label' in answer else question['title']
```
collisdigital: Consider:

`name= answer.get('label') or question.get('title')`

Or

`name = answer.get('label') or question['title']`

app/forms/questionnaire_form.py
```diff
@@ -0,0 +1,100 @@
...
+
+    class QuestionnaireForm(FlaskForm):
+        def map_errors(self):
+            ordered_errors = []
+
+            answer_json_list = SchemaHelper.get_answers_for_block(block_json)
+
+            for answer_json in answer_json_list:
+                if answer_json['id'] in self.errors:
+                    ordered_errors += map_subfield_errors(self.errors, answer_json['id'])
```
collisdigital: Consider replacing this and line 77 with yields.

app/helpers/form_helper.py
```diff
@@ -0,0 +1,109 @@
+from app.helpers.schema_helper import SchemaHelper
+from app.forms.household_composition_form import generate_household_composition_form, deserialise_composition_answers
+from app.forms.household_relationship_form import build_relationship_choices, generate_relationship_form
+from app.forms.questionnaire_form import generate_form
+
+
+def get_form_for_location(block_json, location, answer_store, error_messages, disable_mandatory=False):
```
collisdigital: This feels like multiple functions in one ; might be worth splitting each of the if clauses into a separate function since each block of code behaves so differently; the flag for disable mandatory also looks like another toggle to the function which could be handled by the caller to reduce the complexity/variants through this code.

app/helpers/schema_helper.py
```diff
@@ -95,3 +188,9 @@ def get_last_location(cls, survey_json):
             group_instance=0,
             block_id=cls.get_last_block_id(survey_json),
         )
+
+    @classmethod
+    def get_block_for_location(cls, survey_json, location):
+        group = cls.get_group(survey_json, location.group_id)
+
+        return next(b for b in group['blocks'] if b["id"] == location.block_id)
```
collisdigital: Single quotes for dictionary keys

app/templates/partials/answers/currency.html
collisdigital: Or .get() if it works inside templates.

app/validation/validators.py
```diff
@@ -0,0 +1,129 @@
+from datetime import datetime
+
+from app.validation.error_messages import error_messages
+
+from wtforms import validators
+
+
+class IntegerCheck(object):
```
collisdigital, line 8: Doesn't WTForms provide these kinds of validators out the box; possibly my not understanding how this integrates with what we already have.

app/views/questionnaire.py
```diff
@@ -69,50 +72,112 @@ def save_questionnaire_store(response):
...
+
+    error_messages = SchemaHelper.get_messages(g.schema_json)
+
+    if 'action[save_continue]' in request.form:
+        _remove_repeating_on_household_answers(answer_store, group_id)
+
+    form, _ = post_form_for_location(block, current_location, answer_store, request.form, error_messages)
+
+    if 'action[add_answer]' in request.form:
+        form, _ = post_form_for_location(block, current_location, answer_store, request.form, error_messages, disable_mandatory=True)
```
collisdigital: Seems a bit wasteful to do line 154 then overwrite it here again just to set the flag, any way to remove that situation from occurring?

app/views/questionnaire.py
```diff
@@ -69,50 +72,112 @@ def save_questionnaire_store(response):
...
+
+    current_location = Location(group_id, 0, 'household-composition')
+    block = _render_schema(current_location)
+
+    error_messages = SchemaHelper.get_messages(g.schema_json)
+
+    if 'action[save_continue]' in request.form:
+        _remove_repeating_on_household_answers(answer_store, group_id)
+
+    form, _ = post_form_for_location(block, current_location, answer_store, request.form, error_messages)
```
collisdigital: See comments below; this isn't the default for all scenarios to warrant always doing it here before deciding what is needed.

app/views/questionnaire.py
```diff
@@ -69,50 +72,112 @@ def save_questionnaire_store(response):
...
+    if 'action[add_answer]' in request.form:
+        form, _ = post_form_for_location(block, current_location, answer_store, request.form, error_messages, disable_mandatory=True)
+
+        form.household.append_entry()
+    elif 'action[remove_answer]' in request.form:
+        index_to_remove = int(request.form.get('action[remove_answer]'))
+
+        form.remove_person(index_to_remove)
+    elif 'action[save_sign_out]' in request.form:
+        form, _ = post_form_for_location(block, current_location, answer_store, request.form,
```
collisdigital: Again here overriding the previous assignment.

collisdigital (commented):
Agree with Ben, it might be worth disabling the component tests to get a coverage report showing the gaps, at least to make it visible for the new code to check.

collisdigital:
keep going, it's looking good!

app/views/questionnaire.py
iwootten: It doesn't 'just' set a flag, it reconstructs the whole form with mandatory field errors ignored. Once the form is created, it can't be altered. I can't see a way of getting round this I'm afraid, given the default path depends on a non-mandatory form being created.

app/views/questionnaire.py
iwootten: Same as above

app/forms/date_form.py
iwootten: I've tried multiple approaches to getting this working including __new__, __call__ and __init__. To be clear, what's being passed in is not state (i.e the data that will populate a form), but the form structure.

iwootten:
@benjefferies @collisdigital I'm confused at why we've chosen to review this without taking into account the component tests? I've just spent the best part of two weeks getting them straight and now I'm told they should be ignored? That seems like it should really have been declared somewhat earlier. Are we doing the same for all other PRs?

app/forms/household_composition_form.py
iwootten: We don't have support for it at the moment so this is no different?

app/forms/household_relationship_form.py
iwootten: We want the full list.

app/validation/validators.py
iwootten: Yes, but we can't override their default error, so need to create our own versions of them

app/forms/household_relationship_form.py
iwootten: As above.

benjefferies:
@iwootten We are refactoring a significant amount of the code, removing lots of old code which the integration tests should catch any regression but we are introducing lots of new code. Unit tests will give us quick feedback when things aren't behaving correctly and it also describes the intended behaviour of a function at a granular level and would be useful for when we are doing refactoring on a smaller scale.

If we don't write unit tests on the new code we are introducing now then when?

app/forms/household_composition_form.py
collisdigital: ok, raised a card to review CSRF in general: https://trello.com/c/ZppQHbWg/392-review-use-of-csrf-and-whether-it-is-enabled-disabled-etc

app/forms/household_relationship_form.py
collisdigital: Looks like it's use on line 71 to find the first match, although may be use elsewhere I accept.

app/validation/validators.py
collisdigital: fair enough

jonnyshaw89:
@iwootten On the whole i think this change is really good. It will be a huge improvement to survey running and reduce the amount of custom logic in the system.
However I'm going to now start a look into the code in finer detail.
Good Job 👍

app/forms/fields.py
iwootten: I'm going to leave this as is, as it means each of the fields are associated with appropriately defined widgets despite whether we're using them for rendering.

benjefferies: It seems wrong to leave code in which isn't doing anything?

app/views/questionnaire.py
collisdigital: What I mean is that this line is the same as 154 except that flag is True not False; it also means that if it goes down this path the code executed on 154 was a waste as it isn't needed/the form is completely replaced here so there was no need to construct it twice. So for example, it could have a local disable_mandatory flag that is set True/False based on the conditional logic and the call to post_for_for_location only called once when needed or put 154 inside an if statement or similar refactor.

collisdigital:
I didn't mean the component tests aren't useful, they're very useful and should be there and pass; it was more that some classes/functions benefit from isolated unit tests and it is hard to see what does/doesn't have them when the component tests mask the coverage; keeping in mind the testing pyramid e.g. https://jfiaffe.files.wordpress.com/2014/09/tests-pyramid.png

app/forms/fields.py
benjefferies: The reason I initially raised this is because I thought the wrong widget was being used as StringField should be a TextField if we were using it so I wasted time testing it and figuring out how it was still working to find out it's not being used at all

app/forms/household_relationship_form.py
iwootten: That's returning the last element of a tuple, for every tuple in the list.

app/helpers/schema_helper.py
iwootten: Really? They're the same thing...? Why are they not being enforced in our linting checks? There's other places in this class and the rest of the codebase that are using them interchangeably.

benjefferies: This would satisfy my OCD https://github.com/zheller/flake8-quotes/

app/helpers/form_helper.py
jonnyshaw89: We have had discussions on this before with the disable_mandatory flag. that flag gets passed through many functions to change a nested behaviour.

app/helpers/form_helper.py
```diff
@@ -0,0 +1,109 @@
...
+
+    :param disable_mandatory: Make mandatory answers optional
+    :param error_messages: The default error messages to use within the form
+    :param block_json: The block json
+    :param location: The location which this form is for
+    :param answer_store: The current answer store
+    :return: form, template_args A tuple containing the form for this location and any additional template arguments
+    """
+    if disable_mandatory:
+        block_json = disable_mandatory_answers(block_json)
```
jonnyshaw89, line 19: Modification of incoming variables isn't generally considered best practice.

app/helpers/form_helper.py
```diff
@@ -0,0 +1,109 @@
...
+
+    :param disable_mandatory: Make mandatory answers optional
+    :param error_messages: The default error messages to use within the form
+    :param block_json: The block json
+    :param location: The location which this form is for
+    :param answer_store: The current answer store
+    :param request_form: form, template_args A tuple containing the form for this location and any additional template arguments
+    """
+    if disable_mandatory:
+        block_json = disable_mandatory_answers(block_json)
```
jonnyshaw89, line 70: same comment as above

app/parser/v0_0_1/schema_parser.py
```diff
@@ -81,190 +37,67 @@ def get_parser_version(self):
...
-
-        if "introduction" in self._schema.keys():
-            questionnaire.introduction = self._parse_introduction(self._schema['introduction'])
-
         if "groups" in self._schema.keys():
             for group_schema in self._schema['groups']:
-                questionnaire.add_group(self._parse_group(group_schema, questionnaire))
+                self._parse_group(group_schema)
         else:
             raise SchemaParserException('Questionnaire must contain at least one group')
```
jonnyshaw89, line 47: have we got a trello raised to remove the rest of this?

tests/functional/pages/surveys/census/communal/completion-preference-establishment.page.js
```diff
@@ -9,12 +9,12 @@ class CompletionPreferenceEstablishmentPage extends MultipleChoiceWithOtherPage
   }
 
   clickCompletionPreferenceEstablishmentAnswerOnline() {
-    browser.element('[id="completion-preference-establishment-answer-1"]').click()
+    browser.element('[id="completion-preference-establishment-answer-0"]').click()
```
jonnyshaw89, line 12: we should probably create a card to investigate generating these on the fly to keep @benjefferies happy :)

app/helpers/form_helper.py
iwootten: So every += we do is incorrect? The problem here is there's no neat way of disabling mandatory answers, without cascading down through all the field logic. This is a neater solution in my opinion.

app/forms/household_relationship_form.py
collisdigital: Oh yes! sorry.

app/validation/validators.py
collisdigital: Thought about this some more, can we not derive from the WTForms validator and just override/catch/re-throw the exception with our message to re-use as much of the existing validator as possible?

app/forms/household_relationship_form.py
```diff
@@ -71,12 +71,10 @@ def answer_errors(self, input_id):
             return [error[1] for error in self.map_errors() if input_id == error[0]]
 
     answer = SchemaHelper.get_first_answer_for_block(block_json)
-    guidance = answer['guidance'] if 'guidance' in answer else ''
-    label = answer['label'] if 'label' in answer else ''
 
     field = FieldList(SelectField(
-        label=label,
-        description=guidance,
+        label=answer.get('guidance'),
```
collisdigital: Note that you'll get None returned by default now where before it was `''`; is that an intentional change, if not add the default argument to `get`.

app/forms/household_relationship_form.py
```diff
@@ -0,0 +1,91 @@
...
+            return ordered_errors
+
+        def answer_errors(self, input_id):
+            return [error[1] for error in self.map_errors() if input_id == error[0]]
+
+    answer = SchemaHelper.get_first_answer_for_block(block_json)
+
+    field = FieldList(SelectField(
+        label=answer.get('guidance'),
+        description=answer.get('label'),
```
collisdigital: I commented on this but GitHub decided it was out of date for some reason; in the change to use `.get()` the default of `''` was removed, is that intentional? This will return None now if they aren't found.

app/parser/v0_0_1/schema_parser.py
iwootten: No, you're welcome to open one.

app/forms/household_relationship_form.py
iwootten: Doesn't seem like it was necessary in the first place, so probably better we don't fallback.

app/data/census_household.json
iwootten: You're right, potentially manual errors could occur from this. But this two way relationship is however used in multiple places, such as checking where error messages should be mapped and which answer ids are valid for a form. I can't see a cleaner route to doing this.

app/forms/fields.py
iwootten: Done

app/validation/validators.py
iwootten: The only validator that WTForms supplies that we've created our own version of is the number range - but that's because we want to throw a different error for each limit. There is no Integer validator in wtforms. As we'd need to get right into the call in order to change the message for each limit, this is the simplest solution.

app/helpers/form_helper.py
benjefferies: I guess the alternative is this method doesn't know anything about disabling mandatory answers and the questionnaire.py logic does it when we save/sign out

app/helpers/form_helper.py
benjefferies: I agree, the 2nd return being optional doesn't seem ideal

app/templates/partials/answers/currency.html
iwootten: Having made default changes quite a few of the integration tests fail. Given I know what a nested nightmare the templates are already, and that they should be refactored over and above this anyway, I'm not going to make this change in this PR as I think it'll be a time suck.

app/helpers/form_helper.py
```diff
@@ -0,0 +1,109 @@
...
+    answer_json_list = SchemaHelper.get_answers_for_block(block_json)
+
+    # Deserialise all dates from the store
+    date_answer_ids = [a['id'] for a in answer_json_list if a['type'] == "Date" or a['type'] == 'MonthYearDate']
+
+    for date_answer_id in date_answer_ids:
+        if date_answer_id in mapped_answers:
+            substrings = mapped_answers[date_answer_id].split("/")
+
+            if len(substrings) == 3:
```
benjefferies, line 102: Would be more explicit and more readable to check for the types here rather than the number of splits

app/parser/v0_0_1/schema_parser.py
benjefferies: Is it difficult to remove or can we remove it now? It doesn't seem to add anything

app/templates/partials/answer.html
```diff
@@ -1,17 +1,17 @@
...
-<div class="answer answer--{{answer.schema_item.type|lower}} {{'js-has-errors' if invalid}}" id="{{answer.schema_item.widget.id}}">
+<div class="answer answer--{{answer.type|lower}} {{'js-has-errors' if invalid}}" id="{{answer.id}}">
 
   {% if render_guidance != False %}
 
   {%- set answer_guidance %}
-    {% if answer.schema_item.guidance %}
+    {% if answer.guidance %}
       {% with answer_guidance = {
-        'id': answer.schema_item.id,
```
benjefferies, line 12: Looks much nicer removing the schema_item

app/templates/partials/answers/textarea.html
```diff
@@ -1,3 +1,25 @@
 <div class="field">
-  {{ answer.schema_item.widget.render(answer)|safe}}
+    {% set label = {
+      'for': answer.id,
+      'text': answer.label
+    } %}
+
+    {% include 'partials/forms/label.html' %}
+
+    {% set input = content.form[answer['id']] %}
```
benjefferies: Guess this can just be form for consistency assuming it's already been set?

app/templates/partials/answers/textfield.html
```diff
@@ -1,3 +1,18 @@
+{% set input = form[answer.id] if input is undefined else input %}
```
benjefferies, line 1: Should this be `input is None`?

app/templates/partials/questions/repeatinganswer.html
```diff
@@ -28,14 +26,23 @@ <h3 class="neptune">{{ _('Person ') }}<span class="js-household-loopindex">{{ lo
         </h3>
         <fieldset class="question__fieldset">
           <legend class="question__legend u-vh">{{ _('Person ') }} <span class="js-household-loopindex">{{ loop.index }}</span></legend>
-          {% for answer in composite_answer %}
+          {% for input in name_form %}
             {% with render_guidance = False %}
+              {% set schema_item = question.answers[loop.index0] %}
```
benjefferies, line 31: Would this be better defined as answer or input for consistency?

app/templating/template_renderer.py
```diff
@@ -26,36 +26,5 @@ def render(self, renderable, **context):
         result = rendered if isinstance(renderable, dict) else json.dumps(rendered)
         return json.loads(result)
 
-    def render_schema_items(self, schema_item, context):
```
benjefferies, line 29: 👍

benjefferies (dismissed):

app/parser/v0_0_1/schema_parser.py
iwootten: Yes, it's difficult. I've attempted to do it by making all arrays have a minimum of 1 item in json schema. The resultant problem becomes that interstitials currently define empty lists for answers/questions and all the schemas have empty arrays to suppress the exceptions raised by this schema parser. I'd rather save it for another PR, given this is simply getting WTForms in.

app/templates/partials/questions/repeatinganswer.html
iwootten: I think an input is more appropriate here, as it's the WTForms input which doesn't have the same name as the answer in the schema.

app/validation/validators.py
collisdigital: Got it 👍

collisdigital (approved):
Given the size of this PR, the fact that it's already a big improvement over what was there before and the rebase hell Ian is hitting I'm going for approve and get a team effort to do some testing once deployed to pre-prod. I'd also propose that any outstanding comments are directly related work is immediately prioritised after this merges (not just for Ian to do I'll note)

app/helpers/form_helper.py
```diff
@@ -0,0 +1,109 @@
...
+    if location.block_id == 'household-composition':
+        answers = answer_store.filter(location=location)
+
+        data = deserialise_composition_answers(answers)
+
+        return generate_household_composition_form(block_json, data, error_messages), None
+    elif location.block_id in ['relationships', 'household-relationships']:
+        relationships = next((a['value'] for a in answer_store.answers if a['answer_id'] == 'who-is-related'), None)
+        choices = build_relationship_choices(answer_store, location.group_instance)
+        form = generate_relationship_form(block_json, len(choices), {'who-is-related': relationships}, error_messages)
```
ajmaddaford: Previous answers to the Census Household relationships questions are not re-populated if you return the screens after submitting them. It looks like it's because the answer id is `household-relationships-answer` rather than `who-is-related`

app/data/schema/schema-v1.json
iwootten: Yes, I've added it now and tests to confirm they correctly match one another.

ajmaddaford (changes requested):

app/templates/partials/answers/textfield.html
iwootten: No, specifically the input variable doesn't exist in one case.

app/forms/fields.py
```diff
@@ -0,0 +1,144 @@
...
+            get_date_form(answer, error_messages=error_messages),
+            label=label,
+            description=guidance,
+        )
+
+
+def get_select_field(answer, label, guidance, error_messages):
+    validate_with = get_mandatory_validator(answer, error_messages)
+
+    if answer['type'] == 'Checkbox':
```
ajmaddaford: Any reason not to have `get_select_field` and `get_select_multiple_field` to avoid the if statement (as the mapping is already done in the get_field field mapping above)?

app/forms/fields.py
```diff
@@ -0,0 +1,144 @@
...
+        label=label,
+        description=guidance,
+        validators=validate_with,
+        filters=[lambda x: x if x else None],
+    )
+
+
+def get_date_field(answer, label, guidance, error_messages):
+
+    if answer['type'] == 'MonthYearDate':
```
ajmaddaford: Any reason not to have `get_date_field` and `get_month_year_field` to avoid the if statement (as the mapping is already done in the get_field field mapping above)?

yoganandkunche:
WTForms 1st round of QA done on branch. Tested with all the existing surveys (0102,0112,0203,0205,0213,0215, 0001 & Census Household,individual and communal) 
Please find below are the observations:

**WTForms Issues/Observations:**
1. Displaying all the validation errors upon validation failure. (Across all the surveys)
2. Surveys 0112 & 0215 on the employees screen enter a special character/alphabets and click save and continue gives all the validations errors like above issues and also clears the data entered (doesn’t clear the data in pre-prod)
3. Issue 2 also present in Census Communal, Individual & household. (basically for any numeric field other than £ )
4. Issue 2 also exist in UKIS survey (for any numeric field other than £)
5. Census Individual / household survey on the date of birth question when only a month/year or Month & Year are selected and save and continue is clicked application is not displaying an validation error (in pre-prod it does)
6. UKIS Enter value higher than 100/special chars/alphabets in % field is displaying an error message but clearing the data (doesn’t clear the data in pre-prod)
7. **UKIS: Entering valid value in % field and navigating to next question and coming back clears the data. (P1)**

app/helpers/schema_helper.py
```diff
@@ -58,21 +69,103 @@ def get_skip_condition(group):
...
+        aliases = {}
+        for block in cls.get_blocks(survey_json):
+            for question in cls.get_questions_for_block(block):
+                for answer in question['answers']:
+                    if 'alias' in answer and answer['alias'] not in aliases:
+                        aliases[answer['alias']] = {
+                            'answer_id': answer['id'],
+                            'repeats': answer['type'] == 'Checkbox' or question['type'] == 'RepeatingAnswer',
+                        }
+                    elif 'alias' in answer and aliases[answer['alias']] != answer['id']:
```
ajmaddaford: This won't work as expected, `aliases[answer['alias']]` will be a dict `{'answer_id': ..., 'repeats': ...}` not an answer id. Also, I don't think we need to check if the answer ids match, all answer ids are unique and so if we find a duplicate alias it will be attached to a different answer id - we can simply error if a duplicate alias is found. 

Aside from this, we should consider making this something we check in the schema tests rather than at runtime.

app/templates/questionnaire.html
```diff
@@ -16,24 +13,19 @@
 
 {% block main %}
 
-    {% if all_errors %}
+    {% if form and form.errors %}
 
     <div class="panel panel--error u-mb-s" data-qa="error">
       <div class="panel__header">
-        <h1 class="panel__title venus">This page has {{all_errors|length}} errors</h1>
+        <h1 class="panel__title venus">This page has {{form.map_errors()|length}} errors</h1>
```
ajmaddaford: It would be better to assign form.map_errors to a template variable to avoid calling it twice

ajmaddaford (changes requested):

app/helpers/form_helper.py
```diff
@@ -0,0 +1,115 @@
+from app.helpers.schema_helper import SchemaHelper
+from app.forms.household_composition_form import generate_household_composition_form, deserialise_composition_answers
+from app.forms.household_relationship_form import build_relationship_choices, deserialise_relationship_answers, generate_relationship_form
+from app.forms.questionnaire_form import generate_form
+
+
+def get_form_for_location(block_json, location, answer_store, error_messages, disable_mandatory=False):
```
ajmaddaford, line 7: Is there a reason for passing error messages and disabling mandatory fields on get_form?

ajmaddaford (changes requested):

ajmaddaford (approved):

yoganandkunche:
5.Census Individual / household survey on the date of birth question when only a month/year or Month & Year are selected and save and continue is clicked application is not displaying an validation error (in pre-prod it does)  - **Still failing**

yoganandkunche:
**Issue:** When other selected does not open up the text field for Census Individual & Communal (works fine for Census Household).

collisdigital:
I'm content for the date issue can be resolved after this merges. This does need squashing though.

PR #887, opened 2017-01-27
Adding tests for submitter

benjefferies:
Backfilling tests for submitter

### How to review 
Tests should pass

### DIFF
--- a/tests/app/submitter/test_submitter.py
+++ b/tests/app/submitter/test_submitter.py
@@ -1,0 +1,75 @@
+from unittest import TestCase
+
+from mock import patch, Mock, call
+from pika.exceptions import AMQPError
+
+from app import settings
+from app.submitter.submitter import RabbitMQSubmitter
+
+
+class TestSubmitter(TestCase):
+
+    def test_when_fail_to_connect_to_queue_then_published_false(self):
+        # Given
+        submitter = RabbitMQSubmitter()
+
+        # When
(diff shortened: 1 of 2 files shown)


tests/app/submitter/test_submitter.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def test_send_message_fails(self):
+        # Given
+        submitter = RabbitMQSubmitter()
+
+        # When
+        published = submitter.send_message(message={}, queue='test_queue')
+
+        # Then
+        self.assertFalse(published, 'send_message should fail to publish message')
```
jonnyshaw89: Why should it fail? test doesn't explain why. so if this test fails i wouldn't know what was wrong

jonnyshaw89, line 20: Why should it fail? test doesn't explain why. so if this test fails i wouldn't know what was wrong

jonnyshaw89 (changes requested):

tests/app/submitter/test_submitter.py
benjefferies: Could you take a look again and let me know if you approve?

jonnyshaw89: Beautiful

jonnyshaw89 (approved):

ajmaddaford (approved):

Issue #884, opened 2017-01-27
Census: Performance is dropped when 50 household people are added


yoganandkunche:
### Expected behaviour
Submission must be quick

### Actual behaviour
10 sec wait time for every submit

**Note: The performance improves as I come towards the last person in the group.**

### Steps to reproduce the behaviour
Launch a census household survey
Add more than 50 people in household

Issue #883, opened 2017-01-27
Census: Relationship screen asks to add relationships information for more than 50 people


yoganandkunche:
### Expected behaviour
Must display only 50 people (hard coded at the moment) in the navigation bar and Only ask relationship details for those 50 people

### Actual behaviour
Asking relationship information for all the people added.

### Steps to reproduce the behaviour
Launch Census_Household survey
Add 52 people to the household and click save and continue

### Screenshot
<img width="934" alt="screen shot 2017-01-27 at 09 30 20" src="https://cloud.githubusercontent.com/assets/21335236/22366931/ec65da8a-e476-11e6-9a24-06ad108889c8.png">
<img width="588" alt="screen shot 2017-01-27 at 09 30 48" src="https://cloud.githubusercontent.com/assets/21335236/22366944/f3ea2c3e-e476-11e6-87d6-ebf71edb66d8.png">

collisdigital:
Re-opened as this is still occurring on pre-prod... (maybe regressed after WTForms?)

collisdigital:
Downgraded the priority of this as the main problem is resolved and there is a less serious problem remaining (fix in progress)

PR #897, opened 2017-01-30
add repeat limit to household relationships

jonnyshaw89:
Fixes https://github.com/ONSdigital/eq-survey-runner/issues/883
Fixes https://github.com/ONSdigital/eq-survey-runner/issues/884

### How to review 
Test adding over 25 household members and observe that household relationship questions are only asked for first 25.

### DIFF
--- a/app/data_model/answer_store.py
+++ b/app/data_model/answer_store.py
@@ -151,7 +151,7 @@ class AnswerStore(object):
 
         return len(self.filter(answer.group_id, answer.block_id, answer.answer_id, answer.group_instance, answer.answer_instance))
 
-    def filter(self, group_id=None, block_id=None, answer_id=None, group_instance=None, answer_instance=None, location=None):
+    def filter(self, group_id=None, block_id=None, answer_id=None, group_instance=None, answer_instance=None, location=None, limit=None):
         """
         Find all answers in the answer store for a given set of filter parameter matches.
         If no filter parameters are passed it returns a copy of the list of all answers.
@@ -162,6 +162,7 @@ class AnswerStore(object):
         :param answer_instance: The answer instance to filter results by
         :param group_instance: The group instance to filter results by
         :param location: The location to filter results by (takes precedence over group_id, group_instance and block_id)
+        :param limit: Limit the number of answers returned
         :return: Return a list of answers which satisfy the filter criteria
         """
         filtered = []
@@ -187,6 +188,8 @@ class AnswerStore(object):
                     matches = matches and answer[k] == v
             if matches:
                 filtered.append(answer)
+                if limit and len(filtered) == limit:
+                    break
         return filtered
 
     def clear(self):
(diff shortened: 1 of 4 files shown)


app/settings.py
```diff
@@ -32,7 +32,7 @@ def get_key(_key_name):
 EQ_MAX_HTTP_POST_CONTENT_LENGTH = os.getenv('EQ_MAX_HTTP_POST_CONTENT_LENGTH', 65536)
 
 # max number of repeats from a rule
-EQ_MAX_NUM_REPEATS = os.getenv('EQ_MAX_NUM_REPEATS', 50)
+EQ_MAX_NUM_REPEATS = os.getenv('EQ_MAX_NUM_REPEATS', 25)
```
collisdigital, line 35: While we didn't plan to change this with this PR, I'm happy for this to be 25 by default as a sensible number for now.

app/questionnaire_state/relationship_state_question.py
```diff
@@ -15,8 +16,8 @@ def build_repeating_state(self, user_input):
         template_answer = self.answers.pop()
         group_instance = template_answer.group_instance
 
-        first_name_answers = get_answer_store(current_user).filter(answer_id='first-name')
-        last_name_answers = get_answer_store(current_user).filter(answer_id='last-name')
+        first_name_answers = get_answer_store(current_user).filter(answer_id='first-name')[0:settings.EQ_MAX_NUM_REPEATS]
```
iwootten: This sort of chaining of methods followed by an array index seems pretty dirty to me. Can we break it into multiple lines, return the answerstore, then names, then choose the subset? 

What might be better is to add a limit flag to answerstore filtering, to end early, rather than looping through and fetching everything if we're going to throw anything above an index of 25 away.

iwootten (changes requested):

app/questionnaire_state/relationship_state_question.py
benjefferies: 👍 We should probably have some tests around this too

benjefferies (changes requested):

iwootten (approved):

app/data_model/answer_store.py
```diff
@@ -187,6 +188,8 @@ def filter(self, group_id=None, block_id=None, answer_id=None, group_instance=No
                     matches = matches and answer[k] == v
             if matches:
                 filtered.append(answer)
+                if limit and len(filtered) == limit:
+                    break
```
benjefferies, line 192: We should `return filtered` here instead of breaking

benjefferies (changes requested):

app/data_model/answer_store.py
jonnyshaw89: isn't it good practice to only have 1 return from a method?

benjefferies: Not sure I always prefer to be explicit in what I want to happen i.e. return the list when it's met 25 but maybe a preference thing here so I'm happy to let go

jonnyshaw89: ⛄️

benjefferies (approved):
Snowman has convinced me to let it go... let it go...

PR #898, opened 2017-01-30
Add total field and test schema for breakdown of percentage values.

samiwel:
This change implements the totalling percentages question pattern for the UKIS survey.

### How to review 
Ensure that assets are compiled using `yarn compile`
Launch the `test_total_breakdown.json` schema.
Ensure that the acceptance criteria is met as per the Trello card.


app/assets/js/app/modules/totalliser.js
```diff
@@ -0,0 +1,71 @@
...
+domready(() => {
+  const breakdownFields = document.querySelectorAll('.answer:not(.calculated) input')
+  const totalField = document.querySelectorAll('.calculated input')[0]
+
+  const onAnswerChanged = () => {
+    let newTotal = 0
+    let count = breakdownFields.length
+
+    let i
+    for (i = 0; i < count; i++) {
```
hamishtaplin: no need to declare `i` and `count` separately here as you can just do this:

```js
for (let i = 0; i < breakdownFields.length; i++) {
```

app/assets/js/app/modules/totalliser.js
```diff
@@ -0,0 +1,71 @@
...
+      // good
+      removeHighlight()
+    } else if (value < 100 && value >= 0) {
+      // normal
+      highlight()
+    }
+  }
+
+  forEach(breakdownFields, (value) => {
+    value.addEventListener('change', onAnswerChanged)
```
hamishtaplin: It's often better to utilise event bubbling on the parent to catch and then filter events than add a bunch of them as event listeners have perf implications—less is better.

app/assets/styles/partials/components/_answer.scss
```diff
@@ -12,3 +12,13 @@
 .answer__guidance {
   display: block;
 }
+
+.calculated {
```
hamishtaplin: Don't use arbitrary class names like this, follow the [BEM naming convention](https://csswizardry.com/2013/01/mindbemding-getting-your-head-round-bem-syntax/).

This would probably be something like `answer--calculated`.

app/assets/js/app/modules/totalliser.js
```diff
@@ -0,0 +1,71 @@
+import domready from './domready'
+import {forEach} from 'lodash'
+import getTransitionEndEvent from '../helpers/transitionend'
+
+const transitionendEvent = getTransitionEndEvent()
+
+domready(() => {
+  const breakdownFields = document.querySelectorAll('.answer:not(.calculated) input')
```
hamishtaplin: It's better to explicitly add hooks to the DOM rather than derive them from the classes used for styling and the structure of the DOM. We use the convention of prefixing classes with `js-` for javascript hooks.

app/assets/js/app/modules/totalliser.js
```diff
@@ -0,0 +1,71 @@
...
+  const totalField = document.querySelectorAll('.calculated input')[0]
```
hamishtaplin: `document.querySelector` returns the first matched DOM element.

However, this should use explicit classes rather than relying on structure of DOM.

app/assets/styles/partials/components/_answer.scss
```diff
@@ -12,3 +12,13 @@
...
+    padding-top: 1rem;
+    border-top: 3pt solid #999;
+}
+
+.calculated .highlighted {
```
hamishtaplin: If being highlighted is a state use `.is-highlighted` to indicate styling is statefull. 

However, this probably belongs as part of the styling of `inputs` in general, rather than being tied to answers as it's potentially reusable. It's also used to indicate an error so something like `has-error` is more appropriate.

app/assets/styles/partials/components/_answer.scss
```diff
@@ -12,3 +12,13 @@
 .answer__guidance {
   display: block;
 }
+
+.calculated {
+    padding-top: 1rem;
+    border-top: 3pt solid #999;
```
hamishtaplin: Use `$color-borders` instead of arbitrary color. Also, use `px` not `pt`.

app/data/test_total_breakdown.json
```diff
@@ -0,0 +1,131 @@
...
+                                            "calc_method": {
+                                                "name": "sum",
+                                                "values": [
+                                                    "percentage-1",
+                                                    "percentage-2",
+                                                    "percentage-3",
+                                                    "percentage-4"
+                                                ]
+                                            },
+                                            "description": "<strong>The total percentages should be 100%</strong>",
```
hamishtaplin: Not sure we should be using `description` for error messages. This is a new feature that needs thinking about carefully as it has accessibility implications.

hamishtaplin (changes requested):

app/assets/js/app/modules/totalliser.js
samiwel: @hamishtaplin Can you give an example of what you mean by this? Or point me to a link where I can read up about it?

hamishtaplin: You should be able to add the event listener to a parent element of the fields as events bubble upwards through the DOM.

https://javascript.info/tutorial/bubbling-and-capturing

app/assets/js/app/modules/totalliser.js
```diff
@@ -0,0 +1,75 @@
+import domready from './domready'
+import {forEach} from 'lodash'
+import getTransitionEndEvent from '../helpers/transitionend'
+
+const transitionendEvent = getTransitionEndEvent()
+
+domready(() => {
+  const breakdownFields = document.querySelectorAll('input:not(.js-input--calculated)')
```
hamishtaplin: No need to to use BEM for this name, it doesn't really make sense.

app/assets/js/app/modules/totalliser.js
```diff
@@ -0,0 +1,75 @@
...
+
+  const removeHighlight = () => {
+    if (transitionendEvent) {
+      totalField.addEventListener(transitionendEvent, onTransitionHighlightEnd, true)
+    } else {
+      window.setTimeout(onTransitionHighlightEnd, 500)
+    }
+    totalField.classList.remove('input--has-error')
+  }
+
```
hamishtaplin: both of these functions can be combined into a `toggleHighlight` function as they are largely the same apart from the add/remove aspect.

hamishtaplin (changes requested):

app/assets/js/app/modules/totalliser.js
samiwel: @hamishtaplin I may be totally wrong here, but my understanding after reading that article is that we could add a single event listener to a surrounding element, but is that really what we want? I'm only interested in when any of the breakdown input fields change. Since the total field is also an Answer, if the change listener was applied to a surrounding element, it would also apply to the Total field, which is not what we want. Are the performance gains significant enough to introduce a new element (e.g. a DIV) around the breakdown fields?

app/assets/js/app/modules/totalliser.js
samiwel: Combining these two functions into a `toggleHighlight` function will not have the desired effect. Having two separate functions allows better control of highlighting in the change event listener. But I take your point about the duplication and have refactored out the common logic into a separate function.

app/assets/js/app/modules/totalliser.js
samiwel: I've updated the name.

app/assets/js/app/modules/totalliser.js
hamishtaplin: @samiwel It's not a huge deal but you can filter out change events on the total field itself via the event object, eg.

```js
onAnswerChanged(e) {
   if (e.target === totalField) { 
      return
   }

   // update total
}
```

That's not the exact implementation but you get the idea.

samiwel:
I'll squash the commits once I have two green reviews.

app/assets/js/app/modules/totalliser.js
```diff
@@ -0,0 +1,73 @@
+import domready from './domready'
+import getTransitionEndEvent from '../helpers/transitionend'
+
+const transitionendEvent = getTransitionEndEvent()
+
+domready(() => {
+  const breakdownFields = document.querySelectorAll('input:not(.js-input-calculated)')
+  const totalField = document.querySelector('.js-input-calculated')
+  const fieldSet = document.querySelector('.question__fieldset')
```
hamishtaplin: Please only use specific 'js-' classes as selectors in js, never classes used for styling.

hamishtaplin (changes requested):

samiwel:
Used BrowserStack free trial and tested on a few different browsers including IE, Android and iOS devices.

hamishtaplin (approved):

jonnyshaw89:
@samiwel looks good to me but ill hold the approval until the conflicts are resolved as there might be a bit of re-work

samiwel:
@jonnyshaw89 Happy for me to rebase and squash?

jonnyshaw89 (approved):

PR #901, opened 2017-01-31
Return to Household Composition - Retain Data

LJBabbage:
To retain the repeating answers on Census if there are no changes on the who lives here question. This is done by comparing the answer_store to the request.form dict. If there are no changes then the data for all the repeats should be kept

### How to review 
Using Census_household.json
1. Navigate to the question "2. List the names of everyone who lives here" 
2. Enter 2 people in the boxes below
3. Fill out the rest of the section, including the relationship between the people entered
4. Jump to one of the people on the right and fill out some of their questions
5. Using the link on the right go to "Who lives here"
6. Without changing anything go through that section, making sure anything has been retained
7. Same for the person you filled in 4
8. Go back to "Who lives here" and now change something in regards to the people you have entered
9. Go forward and make sure just the relationship grid is wiped
10. Also check the details you entered in 4 have been wiped.

### DIFF
--- a/app/views/questionnaire.py
+++ b/app/views/questionnaire.py
@@ -16,6 +16,7 @@ from app.questionnaire.location import Location
 from app.questionnaire.navigation import Navigation
 from app.questionnaire.path_finder import PathFinder
 from app.questionnaire.questionnaire_manager import get_questionnaire_manager
+from app.questionnaire_state.state_repeating_answer_question import extract_answer_instance_id
 from app.submitter.converter import convert_answers
 from app.submitter.submitter import SubmitterFactory
 from app.templating.introduction_context import get_introduction_context
@@ -234,7 +235,7 @@ def post_household_composition(eq_id, form_type, collection_id, group_id):
 
     this_location = Location(group_id, group_instance=0, block_id='household-composition')
 
-    if 'action[save_continue]' in request.form:
+    if _household_answers_changed(answer_store):
         _remove_repeating_on_household_answers(answer_store, group_id)
 
     valid = questionnaire_manager.process_incoming_answers(this_location, post_data=request.form)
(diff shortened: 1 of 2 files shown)


app/views/questionnaire.py
```diff
@@ -281,6 +283,20 @@ def _save_sign_out(collection_id, eq_id, form_type, q_manager, this_location):
         return redirect(url_for('.get_sign_out', eq_id=eq_id, form_type=form_type, collection_id=collection_id))
 
 
+def _compare_and_remove_repeating_on_household_if_answers_changed(answer_store, group_id):
+
+    household_answer_store = answer_store.filter(block_id="household-composition")
```
benjefferies: household_answer_store suggests it's something other than an array, just household_answers would be more meaningful

app/views/questionnaire.py
```diff
@@ -281,6 +283,20 @@ def _save_sign_out(collection_id, eq_id, form_type, q_manager, this_location):
         return redirect(url_for('.get_sign_out', eq_id=eq_id, form_type=form_type, collection_id=collection_id))
 
 
+def _compare_and_remove_repeating_on_household_if_answers_changed(answer_store, group_id):
```
benjefferies: It would be nicer if this just compared and returned whether it was true or not otherwise I think it's doing too much. Something like

```
if not _same_answers(…)
    _remove_repeating_on_household_answers(answer_store, group_id)
```

tests/integration/household_composition/test_repeating_household.py
```diff
@@ -92,6 +96,46 @@ def test_change_repeat_answer_removes_all_answers_for_repeated_groups(self):
         self.assertRegex(content, 'What is their age')
         self.assertNotRegex(content, '9992')
 
+    def test_no_change_repeat_answer_keeps_repeated_groups(self):
+
+        # Given I add some people
+        household_composition_page = self.first_page
+        form_data = MultiDict()
```
benjefferies: If you named this household_form_data and the other age_form_data we don't need to duplicate the form data and it's more obvious we are posting the same format data

benjefferies (changes requested):

app/views/questionnaire.py
LJBabbage: I thought the same initially but then decided to avoid the extra logic

app/views/questionnaire.py
LJBabbage: True

jonnyshaw89 (approved):

benjefferies (approved):

benjefferies:
Needs squash and rebasing

LJBabbage:
Squashed and rebased

yoganandkunche:
Retested on Master and works as expected.

PR #902, opened 2017-01-31
Added skip conditions to blocks

jonnyshaw89:
Have added the skip condition to blocks
Have also added a not contains rule

### How to review 
Run through the new test schema 'test_skip_condition_block.json' and observe that the block is skipped.

### DIFF
--- a/app/questionnaire/rules.py
+++ b/app/questionnaire/rules.py
@@ -23,6 +23,8 @@ def evaluate_rule(when, answer_value):
         return True
     elif condition == 'contains' and isinstance(answer_to_test, list) and match_value in answer_to_test:
         return True
+    elif condition == 'not contains' and isinstance(answer_to_test, list) and match_value not in answer_to_test:
+        return True
     elif condition == 'not set':
         return answer_to_test is None
     return False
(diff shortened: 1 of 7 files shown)


benjefferies (commented):
Are we going to follow up the frontend navigation on another card?

app/questionnaire/rules.py
```diff
@@ -23,6 +23,8 @@ def evaluate_rule(when, answer_value):
         return True
     elif condition == 'contains' and isinstance(answer_to_test, list) and match_value in answer_to_test:
         return True
+    elif condition == 'not contains' and isinstance(answer_to_test, list) and match_value not in answer_to_test:
```
benjefferies, line 26: We should have some tests for 'not contains' conditions

jonnyshaw89: Like this one? https://github.com/ONSdigital/eq-survey-runner/pull/902/files#diff-c0704fcf066dc43ec2f73117f8ca828eR110

benjefferies: err yeah

benjefferies (approved):

ajmaddaford (approved):

PR #905, opened 2017-01-31
ID is displayed as N/A for all Surveys

### LINKED ISSUE #891: ID is displayed as N/A for all Surveys
### Expected behaviour
Survey ID must be displayed on the top of the page when survey is launched

### Actual behaviour
Survey ID is displayed as 'N/A' which is not correct

Launch any eQ Survey
id_issue

LJBabbage:
Fixes #891 Missing Survey ID in header

### How to review 
Using any default theme survey, make sure the survey_id is in the header and not N/A

### DIFF
--- a/app/templates/partials/header.html
+++ b/app/templates/partials/header.html
@@ -13,7 +13,7 @@
       {% if meta %}
       <dl class="info__list">
         <dt class="info__dt first"><abbr title="Survey identification code">ID:</abbr></dt>
-        <dd class="info__dd">{{ meta.survey.survey_code or 'N/A'}}</dd>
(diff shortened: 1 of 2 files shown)


samiwel (approved):

benjefferies:
I wonder if it's worth having selenium tests to catch regression. What do you think?

LJBabbage:
A good idea, however we need to be aware that our Selenium suite is becoming wieldy and we need to work out exactly what we want to test.

benjefferies:
I suppose it doesn't need to be an integration test and could be an IT

jonnyshaw89 (approved):

PR #907, opened 2017-01-31
fixing schema validation tests

jonnyshaw89:
Fixing test_schema_validation and adding a negative test to check if we break the test again

### How to review 
Check that the tests pass

### DIFF
--- a/tests/app/data/test_schema_validation.py
+++ b/tests/app/data/test_schema_validation.py
@@ -13,7 +13,20 @@ logger = getLogger()
 
 class TestSchemaValidation(unittest.TestCase):
 
-    def test_schema(self):
+    def test_invalid_schema(self):
+        schema_file = open(os.path.join(settings.EQ_SCHEMA_DIRECTORY, "schema/schema-v1.json"), encoding="utf8")
+        schema = load(schema_file)
+
+        file = "test_invalid_routing.json"
+
+        json_file = open(os.path.join(os.path.dirname(os.path.realpath(__file__)), file), encoding="utf8")
+        json_to_validate = load(json_file)
+
+        errors = self.validate_schema(schema, file, json_to_validate)
+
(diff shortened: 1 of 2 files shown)


tests/app/data/test_schema_validation.py
```diff
@@ -13,7 +13,20 @@
...
 class TestSchemaValidation(unittest.TestCase):
 
-    def test_schema(self):
+    def test_invalid_schema(self):
+        schema_file = open(os.path.join(settings.EQ_SCHEMA_DIRECTORY, "schema/schema-v1.json"), encoding="utf8")
+        schema = load(schema_file)
+
+        file = "test_invalid_routing.json"
+
+        json_file = open(os.path.join(settings.EQ_SCHEMA_DIRECTORY + '/../../tests/app/data/', file), encoding="utf8")
```
benjefferies: Not sure I like the filesystem traversal but not sure what else to suggest

benjefferies (approved):

tests/app/data/test_schema_validation.py
jonnyshaw89: yeh its the only way i could make it with by running it from any directory e.g. running single test and in run_tests.sh

jonnyshaw89: However, Travis doesn't like it.

benjefferies: Other thing I thought about was having the schema in the same directory as the test and find it based on the test_schema_validation.py location

jonnyshaw89: My bad. was in wrong directory

jonnyshaw89: how would you do that?

benjefferies: Oh fair enough. Think you can do something like this `os.path.abspath(__file__)`

ajmaddaford (approved):

PR #912, opened 2017-02-01
MWSS gross pay piping

jonnyshaw89:
Added the piping for gross pay into next question

### How to review 
Enter a gross pay into each section and see that appropriate value is piped into the question text

### DIFF
--- a/app/data/mwss_4.json
+++ b/app/data/mwss_4.json
@@ -163,7 +163,8 @@
                                             "mandatory": true,
                                             "options": [],
                                             "q_code": "50",
-                                            "type": "Currency"
+                                            "type": "Currency",
+                                            "alias": "weekly_total_gross"
                                         }
                                     ],
                                     "description": "",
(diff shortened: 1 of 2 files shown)


app/data/mwss_4.json
```diff
@@ -242,7 +243,7 @@
                                     "description": "",
                                     "id": "weekly-pay-question-6",
                                     "number": "2.3",
-                                    "title": "Of the [TOTAL gross PAY] paid to <em>weekly paid</em> employees in the <em>last week of December 2016</em>, what amount was:",
+                                    "title": "Of the {{answers.weekly_total_gross}} paid to <em>weekly paid</em> employees in the <em>last week of December 2016</em>, what amount was:",
```
ajmaddaford: I think there should be a pound sign in this text, it currently looks like this:
image

Same for all other occurrences.

ajmaddaford (changes requested):

app/data/mwss_4.json
jonnyshaw89: @ajmaddaford i have updated this.

ajmaddaford (approved):

ajmaddaford:
I've added currency formatting for the piped gross pay values and pushed the rebased changes

LJBabbage (changes requested):
ignore

LJBabbage (approved):

PR #916, opened 2017-02-03
Fix remove person with JS disabled

### LINKED ISSUE #889: Census: Household Composition: JS Disabled Cannot Remove Person
### Expected behaviour
When JS is disabled I can still remove people

### Actual behaviour
When JS is disable I cannot remove a person

- Disable JS (e.g. Chrome->Settings->Advanced->Content Settings->Do not allow any site to run JS)
- Navigate to Household Composition
- Add 3 people
- Try and remove Person 2
- Observe the link (Remove) does nothing

#### Browser
Chrome or Lynx (Terminal)

#### Operating System
Windows / macOS Sierra

### Screenshot
image

jonnyshaw89:
Fixes https://github.com/ONSdigital/eq-survey-runner/issues/889
Also adds assertions to functional tests to actually test functionality

Test with and without JS to ensure that functionality fixes issue

### DIFF
--- a/app/assets/js/app/modules/household.js
+++ b/app/assets/js/app/modules/household.js
@@ -14,6 +14,7 @@ class HouseholdMember extends EventEmitter {
   bindToExisting(node) {
     this.node = node
     this.removeBtn = this.node.querySelector('.js-btn-remove')
+    this.removeBtn.setAttribute('type', 'button')
     this.bindToDOM()
   }
 
--- a/app/templates/partials/questions/repeatinganswer.html
+++ b/app/templates/partials/questions/repeatinganswer.html
@@ -18,7 +18,7 @@
           {%- if loop.index == 1 -%}
              <span class="js-household-action">{{ _('You') }}</span>
           {%- else -%}
-            <button class="btn btn--link pluto js-btn-remove" name="action[remove_answer]" value="{{ loop.index0 }}" type="button">{{ _('Remove') }}
+            <button class="btn btn--link pluto js-btn-remove" name="action[remove_answer]" value="{{ loop.index0 }}" type="submit">{{ _('Remove') }}
               <span class="u-vh">{{ _('Person') }} {{loop.index0}}</span>
             </button>
           {%- endif -%}
--- a/tests/functional/capabilities.js
+++ b/tests/functional/capabilities.js
@@ -13,6 +13,14 @@ export const chrome = {
   ...defaultCapabilities
 }
 
+export const chromeNoJS = {
+  ...chrome,
+  name: 'Chrome (No JavaScript) 48 | OS X 10.11',
+  chromeOptions: {
+    prefs: {'profile.managed_default_content_settings.javascript': 2}
+  }
+}
+
 export const firefox = {
   name: 'Firefox 43 | OS X 10.11',
   browserName: 'firefox',
--- a/tests/functional/config.js
+++ b/tests/functional/config.js
@@ -1,5 +1,5 @@
 import {paths} from '../../gulp/paths'
-import {chrome, firefox, ie11, ie10, ie9, ie8} from './capabilities'
+import {chrome, chromeNoJS, firefox, ie11, ie10, ie9, ie8} from './capabilities'
 
 var argv = require('yargs').argv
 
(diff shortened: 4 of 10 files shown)


hamishtaplin (commented):
I think a simpler solution would be to swap the `type` attribute of the remove buttons to `button` at runtime so they don't submit the form when enter is pressed but will still work when interacted with by the user. This is consistent with how the 'add person' button works currently.

jonnyshaw89:
@hamishtaplin Yes that is a lot simpler! 👍

app/assets/js/app/modules/household.js
```diff
@@ -30,9 +30,9 @@ class HouseholdMember extends EventEmitter {
       }
     }
 
-    this.inputs = this.node.querySelectorAll('input')
+    this.inputs = this.node.querySelectorAll('.js-btn-remove')
```
hamishtaplin: `this.inputs` is used elsewhere, this will break it.

in the `bindToExisting` method, there's already a reference to the removeBtn—just do the attribute thing there

hamishtaplin (changes requested):

app/templates/partials/questions/repeatinganswer.html
```diff
@@ -20,7 +20,7 @@ <h3 class="neptune">{{ _('Person ') }}<span class="js-household-loopindex">{{ lo
           {%- if loop.index == 1 -%}
              <span class="js-household-action">{{ _('You') }}</span>
           {%- else -%}
-            <button class="btn btn--link pluto js-btn-remove" name="action[remove_answer]" value="{{ loop.index0 }}" type="button">{{ _('Remove') }}
+            <button class="btn btn--link pluto js-btn-remove" name="action[remove_answer]" value="{{ loop.index0 }}" type="submit">{{ _('Remove') }}
```
samiwel, line 21: For non-JS enabled browsers, this is what was causing people (specifically Person 2) to be removed when users press ENTER key. So just to be aware that the defect will be re-introduced for non-JS browsers with this change.

samiwel: But I understand that without this change you can't remove people for non-JS - which is probably more important.

jonnyshaw89: @samiwel When i test this with javascript enabled it navigates me to the next page on pressing enter.

jonnyshaw89:
@samiwel @hamishtaplin Ready for review.

app/templates/partials/questions/repeatinganswer.html
hamishtaplin: @samiwel @jonnyshaw89 I believe the non-js version should stick to the browser defaults (ie. the first button with `type=submit` submits the form when 'enter' is pressed). with JS, we remove that attribute and handle those actions only on click—meaning the browser will do exactly as the DOM describes.

samiwel (approved):

app/templates/questionnaire.html
```diff
@@ -63,7 +63,7 @@ <h1 class="panel__title venus">This page has {{all_errors|length}} errors</h1>
     {% endif %}
 
     <form action="{{form_action|default('')}}" class="form qa-questionnaire-form" role="form" method="POST" novalidate>
-
+      <input type="submit" style="display: none" name="action[save_continue]" />
```
hamishtaplin: I don't think we should be introducing this, especially to all questionnaires.

The non-js version (which nobody will use) behaves exactly as it's supposed to, this adds the risk of unwanted side-effects.

hamishtaplin (changes requested):

app/templates/questionnaire.html
jonnyshaw89: @hamishtaplin i have removed this now

hamishtaplin (approved):

tests/functional/capabilities.js
```diff
@@ -13,6 +13,17 @@ export const chrome = {
   ...defaultCapabilities
 }
 
+export const chromeNoJS = {
```
hamishtaplin, line 16: It's cleaner to do this:

```js
export const chromeNoJS = {
  ...chrome,
  name: 'Chrome (No JavaScript) 48 | OS X 10.11',
  chromeOptions: {
    prefs: {'profile.managed_default_content_settings.javascript': 2}
  }
}
```

hamishtaplin (changes requested):

hamishtaplin (approved):

PR #917, opened 2017-02-03
eq-635 timeout

hamishtaplin:

[Trello card](https://trello.com/c/Wx4tvl1f/635-timeout-facility-13)

- Added Dialog component (using [a11ydialog](http://edenspiekermann.github.io/a11y-dialog/#introduction) under the hood)
- Added Fetch component (for AJAX requests)
- Created Timeout component that displays a dialog once the threshold of the session timeout has been reached (default 120 seconds) allowing user to continue with survey or sign out.

### How to review 

- set env var `EQ_SESSION_TIMEOUT_SECONDS` to `160`
- start app
- begin a survey
- wait 10 seconds for dialog
- test 'continue survey' button resets timeout
---
- start again
- confirm 'save and sign out' button works
---
- start again
- confirm session times out


app/views/questionnaire.py
```diff
@@ -491,6 +508,17 @@ def _build_template(current_location, context=None, template=None):
 def _render_template(context, block_id, front_end_navigation=None, metadata_context=None, current_location=None, previous_url=None, template=None):
     theme = g.schema_json.get('theme', None)
     logger.debug("theme selected", theme=theme)
+    session_timeout = settings.EQ_SESSION_TIMEOUT - settings.EQ_SESSION_TIMEOUT_GRACE_PERIOD
```
benjefferies: Isn't this the session_timeout_warning?

app/views/questionnaire.py
```diff
@@ -49,6 +51,7 @@ def check_survey_state():
                  ce_id=values['collection_id'], method=request.method, url_path=request.full_path)
 
     _check_same_survey(values['eq_id'], values['form_type'], values['collection_id'])
+    session.permanent = True
```
benjefferies: Should we also set PERMANENT_SESSION_LIFETIME to EQ_SESSION_TIMEOUT?

app/settings.py
```diff
@@ -44,7 +44,8 @@ def get_key(_key_name):
 EQ_GIT_REF = os.getenv('EQ_GIT_REF', 'unknown')
 EQ_NEW_RELIC_CONFIG_FILE = os.getenv('EQ_NEW_RELIC_CONFIG_FILE', './newrelic.ini')
 EQ_SCHEMA_DIRECTORY = os.getenv('EQ_SCHEMA_DIRECTORY', 'app/data')
-EQ_SESSION_TIMEOUT = int(os.getenv('EQ_SESSION_TIMEOUT', '28800'))
+EQ_SESSION_TIMEOUT = int(os.getenv('EQ_SESSION_TIMEOUT', '2700'))
```
collisdigital: I recommend adding _SECONDS to the name so it's clear this is in seconds, when dealing with times like this it's always hard to know if its milliseconds or not. Same below.

bennpearson (changes requested):
Timeout doesn't appear on the summary screen

jonnyshaw89:
@hamishtaplin its also missing on introduction page

yoganandkunche:
Observations:
1. Getting a timeout pop up on a session time out screen.
2. Hitting [esc] have the same effect as [Continue Survey] not sure if we want to fix it or leave it as a feature.
3. Pop up always displays ‘UK Innovation survey’ irrespective of the survey that we are currently filling.

Rest looks good. The button colour on the pop differs between census and other survey which is really good.

jonnyshaw89:
@hamishtaplin should i put the same fix to the session timeout screen as i have to the thank you page?

bennpearson (approved):

LJBabbage:
We use "Save and sign out" on the time out screen but  "Save and complete later" on the survey screen, they are doing the same thing, should it not be consistent?

LJBabbage (changes requested):
We use "Save and sign out" on the time out screen but "Save and complete later" on the survey screen, they are doing the same thing, should it not be consistent?

hamishtaplin:
@LJBabbage We're changing the label of the 'save and complete later' button anyway so probably doesn't matter.

app/views/errors.py
```diff
@@ -31,7 +31,7 @@ def add_cache_control(response):
 @errors_blueprint.app_errorhandler(NoTokenException)
 def unauthorized(error=None):
     log_exception(error)
-    return _render_error_page(401)
+    return render_theme_template('default', 'session-expired.html'), 401
```
LJBabbage, line 34: Not a fan of this, so we are showing the session-expired page for all 401 token exceptions. No token/unauthorised we are putting up the message "To help protect your information we have signed you out" and the page title is "Save Successful"

jonnyshaw89: If you can find a way to distinguish the 2 in flask them I'm open to feedback. However the only reason someone would hit the 401 would be if they were up to no good, so showing them the Session expired page is probably ok.

LJBabbage: I don't have the answer. The analytics will be fun if anyone asks us how many people have tried to hack the token and we have to say that includes timeouts (chalk and cheese together)

yoganandkunche:
Retested on branch and works as expected (issues now resolved)

jonnyshaw89:
@LJBabbage All on you :)

app/views/errors.py
hamishtaplin: @LJBabbage @jonnyshaw89 Could we pass something to that page if someone actually times out vs no token/unauthorised to distinguish between the two? Even if it's just a querystring.

jonnyshaw89: @hamishtaplin not very easily. i spent a day looking at this and as soon as flask know that the session has expired it removes any trace of it.

LJBabbage (approved):

PR #918, opened 2017-02-04
Update builds and instructions to align with AWS ElasticBeanstalk

collisdigital:
https://trello.com/c/74T6bv1m

This PR updates our builds (local/Docker) to align more closely with Amazon Elastic Beanstalk to match our deployment model. 

I have added a .python-version file with the version set to 3.4.2. In order to easily install the correct/exact version of Python I found pyenv useful (it is also what generates the .python-version file) so have updated the README to refer to this tool (https://github.com/yyuu/pyenv).

I also updated the Docker setup to be based on the Amazon AMI base image, install Apache and Python 3.4.2 and run the app behind mod_wsgi. It's not identical to AWS EB, but much closer than we had before.

The Dockerfile is now based on a base image we've added via https://github.com/ONSdigital/docker-aws-apache-wsgi/tree/python-3.4.2 which is auto-built by https://hub.docker.com/ as part of the onsdigital organisation (https://hub.docker.com/r/onsdigital/docker-aws-apache-wsgi/) this adds the basic AWS image, Apache, WSGI and Node/Yarn dependencies.

In addition, the Docker container has been updated to use some of the newer environment variables (e.g. developer logging) and removed the mounted volume as this stops the Docker image from re-building correctly when the file system changes. This change should make Docker re-run yarn compile if the filesystem changes (i.e. with `docker-compose build`) without having to re-build the whole container.

Note I removed the S3 instructions/environment vars from the README/config as it is no longer applicable since S3 support was removed a while ago.

Note I successfully got PyCharm to use the pyenv virtualenv by adding the virtualenv binary to the interpreters setup.

### How to review 
- Try using pyenv to install the correct Python version and setting a virtualenv by following the README
- Try launching via Docker to verify it works correctly


Dockerfile
```diff
@@ -22,10 +20,18 @@ ADD requirements.txt /code/requirements.txt
...
+ENV PYENV_ROOT /code/.pyenv
+ENV PATH $PYENV_ROOT/shims:$PYENV_ROOT/bin:$PATH
 
 ADD . /code
 
+RUN pyenv install
+RUN pyenv rehash
+
+RUN pip install -U pip setuptools wheel
+RUN pip install -U -I -r requirements.txt
```
samiwel: Combining these lines would help reduce the number of steps and size of the final image. A new image layer gets created with each RUN command. They can be combined with && and \ to break at the end of a line.

samiwel (changes requested):

Dockerfile
collisdigital: Will collapse these down

collisdigital: Done.

samiwel (approved):

samiwel:
It is possible to reduce the number of steps in the Docker build even further by re-arranging to something like this. But I just tested and it only reduces the image size by about 30MB in this case, so probably not worth optimising any further.

```
FROM ubuntu:14.04

ENV RUNTIME_PACKAGES="postgresql python-psycopg2" BUILD_PACKAGES="curl build-essential libpq-dev ca-certificates libffi-dev libssl-dev git" PYENV_ROOT="/code/.pyenv"

ADD . /code

WORKDIR /code

EXPOSE 5000

RUN apt-key adv --fetch-keys http://dl.yarnpkg.com/debian/pubkey.gpg \
    && echo "deb http://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list \
		&& apt-get update && apt-get install -y $RUNTIME_PACKAGES $BUILD_PACKAGES \
		&& curl -sL https://deb.nodesource.com/setup_6.x | bash - \
		&& apt-get install -y nodejs yarn \
		&& yarn global add gulp-cli \
		&& git clone git://github.com/yyuu/pyenv.git .pyenv

ENV PATH="$PYENV_ROOT/shims:$PYENV_ROOT/bin:$PATH"
RUN pyenv install 3.4.2 && pyenv rehash && pyenv local 3.4.2 && pip install -U pip setuptools wheel

ENTRYPOINT pip install -r requirements.txt && yarn compile && python application.py runserver
```

collisdigital:
that does also make the file harder to read and most likely harder to identify when a step fails; so think it should stay as-is.

Dockerfile
```diff
@@ -22,10 +20,12 @@ ADD requirements.txt /code/requirements.txt
...
+ENV PYENV_ROOT /code/.pyenv
+ENV PATH $PYENV_ROOT/shims:$PYENV_ROOT/bin:$PATH
 
 ADD . /code
 
-RUN yarn compile
+RUN pyenv install &&  pyenv rehash && pip install -U pip setuptools wheel
 
-ENTRYPOINT python3 application.py runserver
+ENTRYPOINT pip install -r requirements.txt && yarn compile && python application.py runserver
```
jonnyshaw89: Why have we moved this to ENTRYPOINT?

jonnyshaw89 (changes requested):

Dockerfile
collisdigital: so that you don't need to re-build the whole container whenever requirements.txt or yarn compile is needed; restarting the container will do these if it's part of the entry point; it's fast as it only does something if there are changes so maybe adds 5 seconds to the startup time vs. 15 minutes to do a container rebuild

jonnyshaw89: This is kind of an anti-pattern for docker in that the images should be built with all the dependencies and just run. I guess the question is if this image is intended for development or for allowing other people to easily run the system.
The other issue this brings is that failures in the image happen at runtime not compile time as we won't know the image doesn't work until its run.
eg. this image currently fails for me with `ImportError: No module named 'cryptography'` and as our docker compose specifies `restart: always` then it loops around re-attempting infinatly.

collisdigital: We're currently using Docker for dev only; which is why the container imports the current directory into /code (so you get live filesystem updates). I think if we wanted to use Docker for builds we should re-think this.

Cryptography is failing because of the last merge, need to fix that. Because we aren't using ./scripts/run_app.sh it isn't installing the Git version of that package. We should discuss that as well.

Dockerfile
```diff
@@ -1,31 +1,32 @@
-FROM ubuntu:14.04
+FROM onsdigital/docker-aws-apache-wsgi:python-3.4.2
 
-ENV RUNTIME_PACKAGES="python3 postgresql python-psycopg2"
-ENV BUILD_PACKAGES="curl build-essential python3-dev libpq-dev ca-certificates libffi-dev libssl-dev git"
+# Install eq-survey-runner Python packages
+WORKDIR /tmp/eq-survey-runner
```
jonnyshaw89: if you made this directory more generic e.g. `/tmp/application` then it could move into base image

Dockerfile
```diff
@@ -1,31 +1,32 @@
...
+ADD requirements.txt /tmp/eq-survey-runner/requirements.txt
+RUN pip install -r requirements.txt
```
jonnyshaw89: You could look at adding an ONBUILD in the base image to specify this command?

Dockerfile
```diff
@@ -1,31 +1,32 @@
...
+    pip install -e git+https://github.com/reaperhulk/cryptography.git@password-cb#egg=cryptography
 
+# Add remaining eq-survey-runner files and compile frontend
+ADD . /tmp/eq-survey-runner
 RUN yarn compile
 
-ENTRYPOINT python3 application.py runserver
+# Because eq-survey-runner creates 'eq.log' in the current directory
+# it needs write permission as the apache user
+RUN chown apache:apache .
```
jonnyshaw89, line 13: this can probably be an ONBUILD in the base image also.

Dockerfile
```diff
@@ -1,31 +1,32 @@
...
+
+# Start Apache with mod_wsgi on port 80
+EXPOSE 80
```
jonnyshaw89: These 3 command could also move to the base image?

jonnyshaw89 (changes requested):

Dockerfile
collisdigital: Will update

Dockerfile
```diff
@@ -1,31 +1,32 @@
-FROM ubuntu:14.04
+FROM onsdigital/docker-aws-apache-wsgi:python-3.4.2
 
-ENV RUNTIME_PACKAGES="python3 postgresql python-psycopg2"
-ENV BUILD_PACKAGES="curl build-essential python3-dev libpq-dev ca-certificates libffi-dev libssl-dev git"
+# Install eq-survey-runner Python packages
+WORKDIR /tmp/eq-survey-runner
+ADD requirements.txt /tmp/eq-survey-runner/requirements.txt
```
benjefferies: Is there any benefit adding the requirements separately to the rest of the resources?

jonnyshaw89: It improves the caching and build times after the 1st build. but i believe this is going anyway.

benjefferies: Oh yeah, I see your point.

collisdigital:
All comments addressed, ready for re-review

jonnyshaw89 (approved):

PR #926, opened 2017-02-08
Implement a flush mechanism to push unfinished surveys downstream

LJBabbage:
Add a /flush endpoint to survey runner that authenticates via the JWT, identifies which survey response the flush is required for (collection_exercise_sid, eq_id, form_type, ru_ref), and attempts to flush the requested survey response.

N.B at the moment its purely returning a status 200 if successful

### How to review 
1. Put a logger message in flush.py to expose the message being sent to RabbitMQ, just before submitter.send_answers(message) line 60
1. Open a survey as normal, taking note of the collection_exercise_sid, eq_id, form_type, ru_ref. Close the survey without submitting
2. Go to /dev_flush and put in the details above and hit flush, it should return a status 200
3. Test for incorrect and missing details and double flushing, recording the results

### DIFF
--- a/app/templates/dev-flush-page.html
+++ b/app/templates/dev-flush-page.html
@@ -1,0 +1,28 @@
+{% set _theme="default" %}
+{% extends 'dev-page.html' %}
+{% block page_title %}EQ Development Flush Data{% endblock %}
+
+{% block devpage %}
+
+<div class="field-container">
+  <label for="schema">Schemas</label>
+    <select id="schema" name="schema" class="qa-select-schema">
+      {% for id in available_schemas %}
+      <option name="{{id}}" value="{{id}}">{{id}}</option>
+      {% endfor %}
+  </select>
+</div>
+
+<div class="field-container">
+  <label for="collection_exercise_sid">Collection Exercise SID</label>
+  <span><input id="collection_exercise_sid" name="collection_exercise_sid" type="text" class="qa-collection-sid">
+  <img class="refresh" data-refresh="collection_exercise_sid"></span>
+</div>
+
+<div class="field-container">
+  <label for="ru_ref">RU Ref</label>
+  <span><input id="ru_ref" name="ru_ref" type="text" class="qa-ru-ref">
+  <img class="refresh" data-refresh="ru_ref"></span>
+</div>
+
+{% endblock devpage %}
--- a/app/views/dev_mode.py
+++ b/app/views/dev_mode.py
@@ -54,9 +54,36 @@ def dev_mode():
                                  region_code=region_code,
                                  language_code=language_code,
                                  variant_flags=variant_flags)
+
         return redirect("/session?token=" + generate_token(payload).decode())
 
     return render_template("dev-page.html", user=os.getenv('USER', 'UNKNOWN'), available_schemas=available_schemas())
+
+
+@dev_mode_blueprint.route('/dev/flush', methods=['POST'])
+def dev_flush_mode_post():
+    form = request.form
+    schema = form.get("schema")
+    eq_id, form_type = extract_eq_id_and_form_type(schema)
+    collection_exercise_sid = form.get("collection_exercise_sid")
+    ru_ref = form.get("ru_ref")
+
+    payload = {
+        'iat': time.time(),
+        'exp': time.time() + 1000,
+        "eq_id": eq_id,
+        "form_type": form_type,
+        "collection_exercise_sid": collection_exercise_sid,
+        "ru_ref": ru_ref,
+        "roles": ["flusher"],
+    }
+
+    return redirect("/flush?token=" + generate_token(payload).decode())
+
+
+@dev_mode_blueprint.route('/dev/flush', methods=['GET'])
+def dev_flush_mode_get():
+    return render_template("dev-flush-page.html", available_schemas=available_schemas())
 
 
 def extract_eq_id_and_form_type(schema_name):
--- a/app/views/flush.py
+++ b/app/views/flush.py
@@ -1,0 +1,62 @@
+from app.globals import get_answer_store, get_metadata, get_questionnaire_store
+from app.submitter.submitter import SubmitterFactory
+from app.submitter.converter import convert_answers
+from app.utilities.schema import get_schema
+
+from app.authentication.user import User
+from app.authentication.user_id_generator import UserIDGenerator
+from app.authentication.authenticator import jwt_decrypt
+
+from app.questionnaire.path_finder import PathFinder
+
+from flask import Blueprint, Response, request, session
+
+flush_blueprint = Blueprint('flush', __name__)
+
+
+@flush_blueprint.route('/flush', methods=['GET'])
+def flush_data():
+
+    if session:
+        session.clear()
+
+    encrypted_token = request.args.get('token')
+
+    if encrypted_token is None:
+        return Response(status=403)
+
+    decrypted_token = jwt_decrypt(encrypted_token)
+
+    roles = decrypted_token.get('roles')
+
+    if roles and 'flusher' in roles:
+        user = _get_user(decrypted_token)
+        if _submit_data(user):
+            return Response(status=200)
+        else:
+            return Response(status=404)
+    else:
+        return Response(status=403)
+
+
+def _submit_data(user):
+
+    answer_store = get_answer_store(user)
+
+    if answer_store.answers:
+        metadata = get_metadata(user)
+        schema = get_schema(metadata)
+        routing_path = PathFinder(schema, answer_store, metadata).get_routing_path()
+        submitter = SubmitterFactory.get_submitter()
+        message = convert_answers(metadata, schema, answer_store, routing_path)
+        submitter.send_answers(message)
+        get_questionnaire_store(user.user_id, user.user_ik).delete()
+        return True
+    else:
+        return False
+
+
+def _get_user(decrypted_token):
+    user_id = UserIDGenerator.generate_id(decrypted_token)
+    user_ik = UserIDGenerator.generate_ik(decrypted_token)
+    return User(user_id, user_ik)
--- a/tests/integration/test_flush_data.py
+++ b/tests/integration/test_flush_data.py
@@ -1,0 +1,95 @@
+import time
+
+from tests.integration.create_token import create_token
+from tests.integration.integration_test_case import IntegrationTestCase
+from tests.integration.mci import mci_test_urls
+from app.views.dev_mode import generate_token
+
+
+class TestFlushData(IntegrationTestCase):
+
+    def setUp(self):
+        super().setUp()
+
+        token = create_token('0205', '1')
+        self.client.get('/session?token=' + token.decode(), follow_redirects=True)
+
+        post_data = {
+            'action[start_questionnaire]': 'Start Questionnaire'
+        }
+        resp = self.client.post(mci_test_urls.MCI_0205_INTRODUCTION, data=post_data, follow_redirects=False)
+        block_one_url = resp.location
+        self.client.get(block_one_url, follow_redirects=False)
+
+        form_data = {
+            "period-from-day": "01",
+            "period-from-month": "4",
+            "period-from-year": "2016",
+            "period-to-day": "30",
+            "period-to-month": "4",
+            "period-to-year": "2016",
+            "total-retail-turnover": "100000",
+            "action[save_continue]": "Save &amp; Continue"
+        }
+
+        self.client.post(block_one_url, data=form_data, follow_redirects=False)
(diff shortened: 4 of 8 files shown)


app/authentication/no_survey_data_to_flush.py
```diff
@@ -0,0 +1,7 @@
+class NoSurveyDataToFlush(Exception):
```
iwootten: This should probably be in a file/class appended with 'Exception/_exception'

app/authentication/flush_permission_denied.py
```diff
@@ -0,0 +1,8 @@
+
+class FlushPermissionDenied(Exception):
```
iwootten: This should probably be in a file/class appended with 'Exception/_exception'

Also seems to be an empty line at the top here.

app/authentication/map_role_to_permissions.py
```diff
@@ -0,0 +1,12 @@
+from app import settings
+
+
+def map_role_to_permissions(roles):
```
iwootten: This seems like something that would be better suited to living within a authentication helper.

app/authentication/no_survey_data_to_flush.py
iwootten: This exception could be named something more generic, such as SurveyDataException, in order to prevent further exceptions from being created and attached to the same view.

app/authentication/flush_permission_denied.py
iwootten: This exception could be named something more generic, such as PermissionDeniedException, in order to prevent further exceptions from being created and attached to the same view.

iwootten (changes requested):

app/settings.py
```diff
@@ -99,3 +99,5 @@ def get_key(_key_name):
 
 # JWT configurations
 EQ_JWT_LEEWAY_IN_SECONDS = 120
+
+EQ_ROLE_PERMISSIONS = {'admin': ['flush']}
```
jonnyshaw89: Isn't admin a bit of a generic term?
should we name the role more appropriately to the permissions enclosed in it?

app/templates/dev-flush-page.html
```diff
@@ -0,0 +1,118 @@
+{% set _theme="default" %}
+{% extends 'layouts/_base.html' %} {% block page_title %}EQ Development Flush Data{% endblock %} {% block content %}
+
+<!-- this page needs a total rethink, but this tidies things up a little for now -->
```
jonnyshaw89: is it worth doing this now? could it not use the dev-page just with a different button at the bottom?

app/views/dev_mode.py
```diff
@@ -54,11 +54,36 @@ def dev_mode():
...
+@dev_mode_blueprint.route('/dev_flush', methods=['GET', 'POST'])
+def dev_flush_mode():
+    if request.method == "POST":
+        form = request.form
+        schema = form.get("schema")
+        eq_id, form_type = extract_eq_id_and_form_type(schema)
+        collection_exercise_sid = form.get("collection_exercise_sid")
+        ru_ref = form.get("ru_ref")
+
+        payload = {
```
jonnyshaw89: Can create_payload() not be reused?

app/views/questionnaire.py
```diff
@@ -294,6 +294,7 @@ def submit_answers(eq_id, form_type, collection_id):
         path_finder = PathFinder(g.schema_json, answer_store, metadata)
         submitter = SubmitterFactory.get_submitter()
         message = convert_answers(metadata, g.schema_json, answer_store, path_finder.get_routing_path())
+        message['completed'] = True
```
jonnyshaw89: could this boolean be defaulted on convert_answers? that way if this step isn't done after then it will default.

jonnyshaw89 (changes requested):

app/views/flush.py
```diff
@@ -0,0 +1,74 @@
...
+        routing_path = PathFinder(schema, answer_store, metadata).get_routing_path()
+        submitter = SubmitterFactory.get_submitter()
+        message = convert_answers(metadata, schema, answer_store, routing_path)
+        message['completed'] = False
+        submitter.send_answers(message)
+    else:
+        raise NoSurveyDataToFlush("There are no answers to flush")
+
+
+def _get_user(decrypted_token):
```
jonnyshaw89: These functions could be shared with the code in authenticator.py to reduce duplication

jonnyshaw89 (changes requested):

app/templates/dev-flush-page.html
benjefferies: Think this is just a copy and paste carry over

benjefferies: It's probably worth refactoring the style out

app/views/dev_mode.py
```diff
@@ -54,11 +54,36 @@ def dev_mode():
...
+            "eq_id": eq_id,
+            "form_type": form_type,
+            "collection_exercise_sid": collection_exercise_sid,
+            "ru_ref": ru_ref,
+            "roles": ["admin"],
+        }
+
+        return redirect("/flush?token=" + generate_token(payload).decode())
+
+    return render_template("dev-flush-page.html", user=os.getenv('USER', 'UNKNOWN'), available_schemas=available_schemas())
```
benjefferies: The user context appears not to be used

app/views/dev_mode.py
```diff
@@ -54,11 +54,36 @@ def dev_mode():
                                  region_code=region_code,
                                  language_code=language_code,
                                  variant_flags=variant_flags)
+
         return redirect("/session?token=" + generate_token(payload).decode())
 
     return render_template("dev-page.html", user=os.getenv('USER', 'UNKNOWN'), available_schemas=available_schemas())
 
 
+@dev_mode_blueprint.route('/dev_flush', methods=['GET', 'POST'])
```
benjefferies: Think they can be two separate end points. This would remove conditional logic

app/views/flush.py
```diff
@@ -0,0 +1,74 @@
...
+from flask import Blueprint, Response, request, session
+
+flush_blueprint = Blueprint('flush', __name__)
+
+
+@flush_blueprint.route('/flush', methods=['GET'])
+def flush_data():
+
+    if session:
+        session.clear()
```
benjefferies, line 21: Why do we need to clear the session?

app/settings.py
benjefferies: I wonder if we need a nested structure with permissions at this point, would a flusher role suffice?

app/authentication/map_role_to_permissions.py
benjefferies: Not sure if we need roles and permissions at this moment in time since we only have one role with one permission. If we simplify that this can be removed.

tests/integration/test_flush_data.py
```diff
@@ -0,0 +1,95 @@
...
+
+        # Once the data has been flushed it is wiped. It can't be flushed again and should return 404 no data on second flush
+        resp = self.client.get('/flush?token=' + generate_token(self.get_payload()).decode(), follow_redirects=True)
+        self.assertEqual(resp.status_code, 404)
+
+    def test_no_token_passed_to_flush(self):
+        resp = self.client.get('/flush', follow_redirects=False)
+        self.assertEqual(resp.status_code, 401)
+
+    def test_in_valid_token_passed_to_flush(self):
```
benjefferies: typo it should be `test_invalid_token_passed_to_flush`

benjefferies (changes requested):

app/authentication/map_role_to_permissions.py
LJBabbage: Andrew requested this, rather then just having flush

app/settings.py
LJBabbage: As above

app/templates/dev-flush-page.html
LJBabbage: The flush process needs just 4 parameters to work, I didn't want the overhead of another 15 variables being passed around. I am all for reuse but this gave me a clean end point without the need for logic in the post. 

As this will be used stand alone, I wanted to make the process as realistic a possible,  respondent home will only pass the 4 parameters to make it work, therefore the dev_flush tool should do the same

app/views/dev_mode.py
LJBabbage: As above, the create_payload is not realistic as the token would never have the extra variables.

app/views/flush.py
LJBabbage: It confuses the application if you are already login in, as it's gone through an authentication process, then you are asking for it to do again. There is no logically reason why you would be completing a survey and wanting to flush at the same time, so a clean session is the best way to go

iwootten (approved):

app/settings.py
ajmaddaford: I realise I suggested having roles and permissions, but I've thought about it some more and agree with the comments made. I think for now we should simplify this and go with a `flusher` role; we can expand on this with permissions should we need them at a later date.

app/views/flush.py
ajmaddaford: Is it possible to write a test for this?

app/views/dev_mode.py
```diff
@@ -54,11 +54,38 @@ def dev_mode():
                                  region_code=region_code,
                                  language_code=language_code,
                                  variant_flags=variant_flags)
+
         return redirect("/session?token=" + generate_token(payload).decode())
 
     return render_template("dev-page.html", user=os.getenv('USER', 'UNKNOWN'), available_schemas=available_schemas())
 
 
+@dev_mode_blueprint.route('/dev_flush', methods=['POST'])
```
ajmaddaford: Can we make this `/dev/flush`?

app/views/dev_mode.py
```diff
@@ -54,11 +54,38 @@ def dev_mode():
...
+        "form_type": form_type,
+        "collection_exercise_sid": collection_exercise_sid,
+        "ru_ref": ru_ref,
+        "roles": ["admin"],
+    }
+
+    return redirect("/flush?token=" + generate_token(payload).decode())
+
+
+@dev_mode_blueprint.route('/dev_flush', methods=['GET'])
```
ajmaddaford: Can we make this `/dev/flush`?

ajmaddaford (changes requested):

app/views/questionnaire.py
LJBabbage: Done, but in reverse, i.e set to false until complete

app/views/flush.py
LJBabbage: Inadvertently its already does,as the tests in the setup populate the data, meaning there is a session, which needs clearing. Its how I noticed the problem originally.

ajmaddaford:
This can't be merged until the `completed` flag has been added to SDX - see https://github.com/ONSdigital/sdx-validate/issues/28

app/views/flush.py
```diff
@@ -0,0 +1,66 @@
...
+
+    if answer_store.answers:
+        metadata = get_metadata(user)
+        schema = get_schema(metadata)
+        routing_path = PathFinder(schema, answer_store, metadata).get_routing_path()
+        submitter = SubmitterFactory.get_submitter()
+        message = convert_answers(metadata, schema, answer_store, routing_path)
+        submitter.send_answers(message)
+    else:
+        raise SurveyDataException("There are no answers to flush")
```
benjefferies: Do we want to raise a SurveyDataException here, this will cause us to return a 404 web page which doesn't seem right for an API designed for a system

app/views/flush.py
```diff
@@ -0,0 +1,66 @@
...
+
+    roles = decrypted_token.get('roles')
+
+    if roles:
+        if 'flusher' in roles:
+            user = _get_user(decrypted_token)
+            _submit_data(user)
+            get_questionnaire_store(user.user_id, user.user_ik).delete()
+        else:
+            raise PermissionDeniedException("Your role doesn't have permission to flush data")
```
benjefferies: Do we want to raise this exception, it will cause a web page to be returned which isn't ideal for an automated system. Why not just return Response(status=403)

benjefferies (changes requested):

app/views/flush.py
LJBabbage: See acceptance criteria

app/views/flush.py
LJBabbage: See acceptance criteria

benjefferies (approved):

jonnyshaw89 (approved):

ajmaddaford (approved):

Issue #922, opened 2017-02-08
Non Mandatory Date field does not work as expected


yoganandkunche:
### Expected behaviour
Must throw an validation error when only month/year or month & year selected as the date entered is invalid. 

### Actual behaviour
Does not throw any validation errors

### Steps to reproduce the behaviour
Launch Census Individual/Household survey
navigate to date of birth field

- 1 Select only month and click [Save and continue]
 or
- 2 Select Month and Year click [Save and continue]

iwootten:
Fixed in #928

PR #928, opened 2017-02-08
Add an OptionalForm validator

iwootten:
This adds an OptionalForm validator, which can be applied to any single field on a form to correctly halt validation if all of the fields on the form are empty. 

We use this currently only with DateForms, but could be used on any formfield which has multiple subfields that is only deemed empty when all fields are empty.

Also fixed a regression around optional month fields being able to change their answers back to empty (#930).

Fixes issues #922, #930  

### How to review 
Check it out, using Census/Individual see if you can proceed without entering a day in the date of birth field.

### DIFF
--- a/app/validation/validators.py
+++ b/app/validation/validators.py
@@ -3,6 +3,7 @@ from datetime import datetime
 from app.validation.error_messages import error_messages
 
 from wtforms import validators
+from wtforms.compat import string_types
 
 
 class IntegerCheck(object):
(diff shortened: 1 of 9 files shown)


collisdigital:
Do we need a Selenium test that exercises the date question pattern in the way that the issue failed as well as the component test?

yoganandkunche:
@ian: a minor issue not related to this fix though.. I think it regressed (may be worth fixing in the same PR). The responded must be able to select the 'Select month' value even after selecting any month from the drop down for a non mandatory date field.

collisdigital:
@yoganandkunche but only if the question is optional, otherwise they shouldn't be able to re-select 'Select month'; at least that's what I believe the behaviour is

yoganandkunche:
@collisdigital : Yes thats right.! However I am raising a separate defect as suggested by @iwootten

app/validation/validators.py
```diff
@@ -47,6 +48,28 @@ def __call__(self, form, field):
...
+    field_flags = ('optional',)
+
+    def __call__(self, form, field):
+        empty_form = True
+
+        for formfield in form:
+            # By default we'll receive empty arrays for values not posted, so need to allow empty lists
+            empty_field = hasattr(formfield, 'raw_data') and \
+                          (True if len(formfield.raw_data) == 0 else
+                           isinstance(formfield.raw_data[0], string_types) and not formfield.raw_data[0])
```
ajmaddaford: Can we break this down a bit so it's clearer?

ajmaddaford (changes requested):

iwootten:
I've added the fix for #930 into this, given the tests depended on the fix for #922 being in place.

jonnyshaw89 (approved):

ajmaddaford (approved):

PR #935, opened 2017-02-09
Fixes #923 other text value retained even when other not selected.

### LINKED ISSUE #923: Other text field retains data even the user changes the option
### Expected behaviour
Must wipe the data entered in other text field before when user changes the option.

### Actual behaviour
Retains previously entered text.

Launch any survey which has an other text field option
select other and enter data and click save and continue
navigate back using previous change the option and click save and continue
navigate back and select the other option again

samiwel:
Fixes #923.

### How to review 
All automated tests should pass.
Open either `test_checkbox.json` and `test_radio.json` schemas.
Select the `Other` option and enter a value into the text field.
Save and continue.
Go back.
Select a different option.
Save and continue.
Go back.
Select the `Other` option again.
Confirm that the previous value is not in the Other text field.

Should also work when selecting multiple options (checkbox).

### DIFF
--- a/app/helpers/form_helper.py
+++ b/app/helpers/form_helper.py
@@ -1,3 +1,5 @@
+from werkzeug.datastructures import MultiDict
+
 from app.helpers.schema_helper import SchemaHelper
 from app.forms.household_composition_form import generate_household_composition_form, deserialise_composition_answers
 from app.forms.household_relationship_form import build_relationship_choices, deserialise_relationship_answers, generate_relationship_form
@@ -77,7 +79,8 @@ def post_form_for_location(block_json, location, answer_store, request_form, err
 
         return form, {'relation_instances': choices}
     else:
-        return generate_form(block_json, request_form, error_messages), None
+        data = clear_other_text_field(request_form, SchemaHelper.get_questions_for_block(block_json))
+        return generate_form(block_json, data, error_messages), None
 
 
 def disable_mandatory_answers(block_json):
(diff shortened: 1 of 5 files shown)


app/forms/questionnaire_form.py
```diff
@@ -93,8 +93,23 @@ def answer_errors(self, input_id):
...
+        clear_other_text_field(block_json, data, form_data)
+
+        form = QuestionnaireForm(form_data, meta={'csrf': False})
     else:
         form = QuestionnaireForm(meta={'csrf': False})
 
     return form
+
+
+def clear_other_text_field(block_json, data, form_data):
```
iwootten: At the moment, the form_helper has some deserialisation logic around dates within it. I would think that this clearing of the other text field would be better placed alongside as the data wouldn't be manipulated as part of the form logic then.

app/forms/questionnaire_form.py
```diff
@@ -93,8 +93,23 @@ def answer_errors(self, input_id):
...
+    for question in SchemaHelper.get_questions_for_block(block_json):
+        for answer in question['answers']:
+            if 'parent_answer_id' in answer and \
+                    answer['parent_answer_id'] in data and \
+                    'Other' not in form_data.getlist(answer['parent_answer_id']) and \
+                    form_data.get(answer['id']) is not None and \
```
iwootten: I think these last two conditions can simply be expressed as:

    and form_data.get(answer['id'])

iwootten (changes requested):

app/forms/questionnaire_form.py
iwootten: Actually, this clear could possibly be achieved within the get_answer_fields method?

jonnyshaw89 (approved):

iwootten (approved):

samiwel:
Squashed and rebased ready to be merged.

PR #940, opened 2017-02-13
Fix #931: Override validate and add date range validation at form level

### LINKED ISSUE #931: Displays invalid validation error message on the date field.
### Expected behaviour
Must not display error when the date is 28/feb/2019 which is within the range.

### Actual behaviour
Displaying wrong error message for a incorrect field (Please find attached) 

Launch a survey with date range (ex: 0102)
Enter 28/Feb/2019 for From
Enter 31/Sept/2020 for To
and click save and continue.

### Additional information
Tested in Pre-prod and it fails (something caused from WTForms). Work fine in Prod
#### Browser
Chrome
### Screenshot
screen shot 2017-02-09 at 11 01 39

iwootten:
Add form level validation to the QuestionnaireForm by overriding it's validate method in order to validate DateRanges at a question level and display errors appropriately. Additionally, this moves the metaclass definition for QuestionnaireForm out of the generate_form method and reduces the methods complexity.

Removes the need for current DateRange validation as part of the date_form.

Fixes #931.

### How to review 

Confirm if unit tests pass, whether you agree with approach.

### DIFF
--- a/app/forms/questionnaire_form.py
+++ b/app/forms/questionnaire_form.py
@@ -1,26 +1,74 @@
 import logging
 
+from wtforms import validators
+
 from flask_wtf import FlaskForm
 
-from app.forms.date_form import get_date_data, get_date_range_fields
 from app.forms.fields import get_field
 from app.helpers.schema_helper import SchemaHelper
+from app.validation.validators import DateRangeCheck
 
 from werkzeug.datastructures import MultiDict
 
 logger = logging.getLogger(__name__)
 
 
-def get_date_range_field(question, data, error_messages):
-    to_field_id = question['answers'][1]['id']
-    to_field_data = get_date_data(data, to_field_id)
+class QuestionnaireForm(FlaskForm):
 
-    from_field, to_field = get_date_range_fields(question, to_field_data, error_messages)
+    def __init__(self, block_json, formdata=None, **kwargs):
+        self.block_json = block_json
+        self.question_errors = {}
 
-    return {
-        question['answers'][0]['id']: from_field,
-        question['answers'][1]['id']: to_field,
-    }
+        if formdata:
+            super(QuestionnaireForm, self).__init__(formdata=formdata, **kwargs)
+        else:
+            super(QuestionnaireForm, self).__init__(**kwargs)
+
+    def validate(self):
+        """
+        Validate this form as usual and check for any form-level date-range validation errors
+        :return:
+        """
+        valid_form = True
+        valid_fields = FlaskForm.validate(self)
+
+        for question_id, period_from_id, period_to_id in self.date_ranges:
+            if period_from_id not in self.errors and period_to_id not in self.errors:
+                period_from = getattr(self, period_from_id)
+                period_to = getattr(self, period_to_id)
+                validator = DateRangeCheck()
+
+                # Check every field on each form has populated data
+                populated = all([f.data for f in period_from] + [f.data for f in period_to])
+
+                if populated:
+                    try:
+                        validator(self, period_from, period_to)
+                    except validators.ValidationError as e:
+                        self.question_errors[question_id] = str(e)
+                        valid_form = False
+
+        return valid_fields and valid_form
+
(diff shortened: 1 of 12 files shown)


ajmaddaford (approved):

yoganandkunche:
@iwootten Tested on the branch the actual defect #931 Works as expected. However when the  'period to' date is before the 'period from' date. the error [link](url)  is missing. Please find the attached screen shot 
screen shot 2017-02-14 at 15 05 47

app/forms/questionnaire_form.py
```diff
@@ -1,26 +1,74 @@
...
+
+                # Check every field on each form has populated data
+                populated = all([f.data for f in period_from] + [f.data for f in period_to])
+
+                if populated:
+                    try:
+                        validator(self, period_from, period_to)
+                    except validators.ValidationError as e:
+                        self.question_errors[question_id] = str(e)
+                        valid_form = False
```
benjefferies, line 49: We could just have valid_fields and just set that to false. I am not too sure what the difference between them is.

app/validation/validators.py
```diff
@@ -122,26 +122,24 @@ def __call__(self, form, field):
...
+        to_date = datetime.strptime(to_date_str, "%d/%m/%Y")
 
-            date_diff = to_date - from_date
+        date_diff = to_date - from_date
 
-            if date_diff.total_seconds() == 0:
-                raise validators.ValidationError(self.messages['INVALID_DATE_RANGE_TO_FROM_SAME'])
-            elif date_diff.total_seconds() < 0:
-                raise validators.ValidationError(self.messages['INVALID_DATE_RANGE_TO_BEFORE_FROM'])
+        if date_diff.total_seconds() == 0:
```
benjefferies: Would this read better as from_data == to_date?

app/validation/validators.py
```diff
@@ -122,26 +122,24 @@ def __call__(self, form, field):
...
+            raise validators.ValidationError(self.messages['INVALID_DATE_RANGE_TO_FROM_SAME'])
+        elif date_diff.total_seconds() < 0:
```
benjefferies: Would this read better from_date > to_date?

benjefferies (changes requested):

app/forms/questionnaire_form.py
iwootten: A form can be invalid even if all the fields are valid. Marking a field as invalid isn't technically true if it fails form validation.

app/validation/validators.py
iwootten: So comparing two different instances of the same date? This wouldn't work.

benjefferies: Why not? Doesn't datetime implement \_eq\_ and \_ne\_ to compare the values rather than the object references?

iwootten: Nope, my bad, I checked using different instances of now() which doesn't work. I guess because now is slightly different between the time it takes to create each instance.

Honestly, I'd rather not change anything over the issue though, this is as it was before.

app/forms/questionnaire_form.py
benjefferies: You can still mark it as invalid with the one boolean and isn't it validating two fields rather than the whole form?

app/validation/validators.py
benjefferies: Ok I'll try and remember to make the change if I am in the area in the future I guess

app/forms/questionnaire_form.py
iwootten: I think that approach masks the individual checks that are actually taking place. There's two levels of validity, one at field level and one at the form. We're being explicit about what the type of failure is.

This method validates the whole form as otherwise no other errors for the same block would be raised. The fact that only date_ranges are checked at a form level makes no difference, it still informs whether that form is valid.

benjefferies (approved):

yoganandkunche:
@iwootten Retested on branch and works as expected.

PR #947, opened 2017-02-14
Fixing regression caused by WTForms

jonnyshaw89:
Re-implementing feature in https://github.com/ONSdigital/eq-survey-runner/pull/897 that was regressed in https://github.com/ONSdigital/eq-survey-runner/pull/885

### How to review 
Re-test https://github.com/ONSdigital/eq-survey-runner/issues/883

### DIFF
--- a/app/forms/household_relationship_form.py
+++ b/app/forms/household_relationship_form.py
@@ -1,6 +1,7 @@
 from flask_wtf import FlaskForm
 from wtforms import FieldList, RadioField
 
+from app import settings
 from app.forms.fields import build_choices, get_mandatory_validator
 from app.data_model.answer_store import Answer
 from app.helpers.schema_helper import SchemaHelper
@@ -17,16 +18,17 @@ def build_relationship_choices(answer_store, group_instance):
     :param group_instance: The instance of the group being iterated over
     :return:
     """
-    household_answers = answer_store.filter(block_id='household-composition')
+    household_first_name_answers = answer_store.filter(block_id='household-composition', answer_id='first-name', limit=settings.EQ_MAX_NUM_REPEATS)
+    household_last_name_answers = answer_store.filter(block_id='household-composition', answer_id='last-name', limit=settings.EQ_MAX_NUM_REPEATS)
 
     first_names = []
     last_names = []
 
-    for answer in household_answers:
-        if answer['answer_id'] == 'first-name':
-            first_names.append(answer['value'])
-        if answer['answer_id'] == 'last-name':
-            last_names.append(answer['value'])
+    for answer in household_first_name_answers:
+        first_names.append(answer['value'])
+
+    for answer in household_last_name_answers:
+        last_names.append(answer['value'])
 
     household_members = []
 
--- a/tests/app/forms/test_household_relationship_form.py
+++ b/tests/app/forms/test_household_relationship_form.py
@@ -1,6 +1,7 @@
 import unittest
 
 from app import create_app
+from app import settings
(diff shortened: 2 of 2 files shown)


app/forms/household_relationship_form.py
```diff
@@ -17,7 +18,7 @@ def build_relationship_choices(answer_store, group_instance):
     :param group_instance: The instance of the group being iterated over
     :return:
     """
-    household_answers = answer_store.filter(block_id='household-composition')
+    household_answers = answer_store.filter(block_id='household-composition', limit=settings.EQ_MAX_NUM_REPEATS)
```
benjefferies: Maybe a test to prevent regressions?

benjefferies (changes requested):

app/forms/household_relationship_form.py
```diff
@@ -17,7 +18,7 @@ def build_relationship_choices(answer_store, group_instance):
     :param group_instance: The instance of the group being iterated over
     :return:
     """
-    household_answers = answer_store.filter(block_id='household-composition')
+    household_answers = answer_store.filter(block_id='household-composition', limit=settings.EQ_MAX_NUM_REPEATS * 2)
```
iwootten: Why x 2?

iwootten (changes requested):

app/forms/household_relationship_form.py
jonnyshaw89: because it should have an answer for first-name and an answer for last-name

iwootten: Won't it also return middle names? Maybe it would be better to query each answer_id individually?

collisdigital: Last name isn't mandatory by the way

app/forms/household_relationship_form.py
```diff
@@ -17,16 +18,17 @@ def build_relationship_choices(answer_store, group_instance):
     :param group_instance: The instance of the group being iterated over
     :return:
     """
-    household_answers = answer_store.filter(block_id='household-composition')
+    household_first_name_answers = answer_store.filter(block_id='household-composition', answer_id='first-name', limit=settings.EQ_MAX_NUM_REPEATS)
```
iwootten, line 21: I guess the problem with this approach is that we can't guarantee that the last_names each match with the first_name as last_name is not mandatory.

iwootten (changes requested):

app/forms/household_relationship_form.py
jonnyshaw89: That problem comes with the zip? would the answer-instance maintain the gaps?

iwootten: Ah yes, but I think if I remember correctly we are storing blank answers for this particular form where there's nothing entered (the tests all seem to assume this), so probably not a problem.

collisdigital: Assuming that change thats in another PR doesn't stop it storing blank entries, the trim thing?

iwootten: Well, that will just strip it down to the empty string '' and is only for mandatory answers...so won't affect middle or last names. Middle and last names are stripped by the optional validator, but don't have to be populated and store blank entries. You can see some existing tests which haven't been altered in my branch for such behaviour here: https://github.com/ONSdigital/eq-survey-runner/blob/master/tests/app/forms/test_household_composition.py#L62 

I should note this behaviour is as it was prior to the WTForms merge.

iwootten (approved):

benjefferies (approved):

Issue #946, opened 2017-02-14
Introduction description must not be displayed for UKIS survey


yoganandkunche:
### Expected behaviour
No Introduction description must be displayed.

### Actual behaviour
Displaying introduction description for UKIS survey (PFA)

### Steps to reproduce the behaviour
Launch UKIS survey

### Screenshot
<img width="1039" alt="screen shot 2017-02-14 at 15 40 13" src="https://cloud.githubusercontent.com/assets/21335236/22936016/20dd5c30-f2cc-11e6-8312-6212f6dc60cc.png">

yoganandkunche:
Defect fixed with PR #948

PR #938, opened 2017-02-13
Fixes #909 missing description on introduction page.

samiwel:
### What is the context of this PR?
Fixes #909.

### How to review 
Re-test the defect.
The description should be displayed on the introduction page.
All existing tests should pass.

PR #948, opened 2017-02-14
Removes the ukis landing page introduction description

yoganandkunche:
Removes the introduction description from the UKIS landing page.

### How to review 
Retest Defect #946 

Launch UKIS survey and make sure that introduction description is not present on UKIS landing page.

### DIFF
--- a/app/data/1_0001.json
+++ b/app/data/1_0001.json
@@ -5774,7 +5774,7 @@
     "theme": "ukis",
     "legal_basis": "Voluntary",
     "introduction": {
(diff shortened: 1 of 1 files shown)


samiwel (approved):
A recent change(#938) re-introduced the description section in the introduction pages. The UKIS schema needed to be updated to remove this description as it was not in the spec.

This change looks fine to me. I don't think we need specific tests for this change since we already have integration tests for showing/hiding the descriptio in `test_introduction.py`.

Issue #941, opened 2017-02-13
Census: Entering space in First Name field on household create a blank household person


yoganandkunche:
### Expected behaviour
Must not allow white spaces in Name field 
### Actual behaviour
Allowing white spaces in first name field and creates a placeholder in the navigation. (Screen shot)

### Steps to reproduce the behaviour
Launch Census Household schema
Navigate to Add person screen
and hit space in First name field for the third person
click [Save and continue]

### Screenshot
<img width="1138" alt="screen shot 2017-02-13 at 16 13 36" src="https://cloud.githubusercontent.com/assets/21335236/22891581/c389b4b2-f207-11e6-84a0-c24f76a11842.png">

PR #949, opened 2017-02-14
Fixes #933 none displayed in navigation after save and complete later.

### LINKED ISSUE #933: Using save and complete later on Household Composition allows None person
### Expected behaviour
- Save and complete later doesn't 'save' an empty entry

### Actual behaviour
- Save and complete later adds 'None' to the household composition if an empty person is added before Save and complete later is used.

- Add a few people to household composition
- Add another person but leave all field blank
- Use save and complete later
- Sign back in
- See that 'None' is listed as a person in the navigation

#### Browser
Chrome

#### Operating System
macOS Sierra

### Screenshot
image

samiwel:
Fixes #933.

This is related to, but not the same as #941.
In the case that the Navigation menu cannot construct the link name (because a person with a blank name has been stored into the answer store), this PR removes answers for the household members from the answer store if no value was provided.

### How to review 
Launch the `census_household.json` survey and make a note of the collection exercise Id.
Navigate to the household composition screen.
Add members to the household but do not enter any names.
Save and complete later.
Return to the survey using the same collection exercise Id.
Ensure that None is not displayed in the Navigation menu.
Enter the person name (first name or surname) and save and complete later.
Go back into the survey and ensure the names are retained.

_Note: When you partially complete the household member names (by either entering a first name or a surname) the answers will be saved. Entering only a middle name will be treated as an empty answer and the value won't be retained. The reason for this is because the Navigation depends on the first name and last name for displaying a navigation entry. Entering only a middle name would cause `None` to be displayed on the navigation and hence would not resolve the defect._

### DIFF
--- a/tests/app/views/test_questionnaire.py
+++ b/tests/app/views/test_questionnaire.py
@@ -8,7 +8,7 @@ from app.data_model.answer_store import Answer
 from app.data_model.questionnaire_store import QuestionnaireStore
 from app.schema_loader.schema_loader import load_schema_file
 from app.questionnaire.location import Location
-from app.views.questionnaire import update_questionnaire_store_with_answer_data, update_questionnaire_store_with_form_data
+from app.views.questionnaire import update_questionnaire_store_with_answer_data, update_questionnaire_store_with_form_data, remove_empty_household_members_from_answer_store
 
 from flask import g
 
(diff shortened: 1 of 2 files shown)


app/questionnaire/navigation.py
```diff
@@ -141,7 +141,7 @@ def _generate_repeated_items(self, link_names, first_location, completed_id, no_
             location.group_instance = i
             is_current_group_instance = is_current_group and i == current_group_instance
 
-            if link_names:
+            if link_names and link_names.get(i):
```
jonnyshaw89: Is this not just sticking a plaster rather than fixing the issue? why do we have a repeat if it doesn't have a name?

samiwel: Yeah, that's being being fixed in (941). In theory, you should never reach a situation where this fix is required. But actually in the case where the Navigation can't construct a name for the link, something is probably wrong, so this change just ensures that an invalid link isn't displayed in the navigation.

LJBabbage: Problem is although this would hide it in the navigation on the right, you would still get it if you went through the survey.

samiwel: Hmm good point, so actually maybe there's a question to be asked whether we need this fix at all. I guess the real solution here is to make sure that any answers that generate a navigation entry (repeated group) need to be sanitised/validated properly to stop that from happening?

samiwel: Or since this only appears to be an issue when hitting `save and complete later` on the household question for the moment, we could just deal with this specific case and delete any answers that have empty values for this question when `save and complete later` is pressed?

tests/app/views/test_questionnaire.py
```diff
@@ -8,7 +8,7 @@
 from app.data_model.questionnaire_store import QuestionnaireStore
 from app.schema_loader.schema_loader import load_schema_file
 from app.questionnaire.location import Location
-from app.views.questionnaire import update_questionnaire_store_with_answer_data, update_questionnaire_store_with_form_data
+from app.views.questionnaire import update_questionnaire_store_with_answer_data, update_questionnaire_store_with_form_data, _remove_empty_household_members_from_answer_store
```
iwootten: If this really is to be considered a private method, I don't think we should be testing it outside of the view using it.

tests/app/views/test_questionnaire.py
```diff
@@ -146,3 +146,256 @@ def test_update_questionnaire_store_with_answer_data(self):
...
+                group_instance=0,
+                block_id='household-composition',
+                answer_id='last-name',
+                answer_instance=0,
+                value=''
+            )
+        ]
+
+        answers = []
+        answers.extend(unanswered)
```
iwootten: Could remove the copy here and simply iterate over unanswered in each of the following loops?

iwootten (changes requested):

tests/app/views/test_questionnaire.py
samiwel: Valid point. I was just developing test-first to make sure the various scenarios were covered.

samiwel:
Is there any reason why only the first name and the last name are displayed in the navigation menu?

I don't particularly like the fact that middle name is having to be treated as a special case. Since only first name and surnames are displayed in the Navigation, this fix currently removes answer store entries that do not have either a first name or a last name.

The problem with this is that if a user adds a household member, and enters only a middle name value, before saving and completing later on, then the answer will be lost (since the navigation would display `None` in this case).

I think it would be better to display the full name in the navigation and then we wouldn't have this weird edge case with the middle name. But welcome thoughts on this.

collisdigital:
First-name is mandatory so you can't have only a middle-name. We aren't going to display the full-name in navigation.

samiwel:
Okay, fair enough, then I guess this fix just needs review/approval.

iwootten (approved):

jonnyshaw89 (approved):

jonnyshaw89:
@samiwel can you rebase?

PR #950, opened 2017-02-14
Fix #941: Add a response required validator

### LINKED ISSUE #941: Census: Entering space in First Name field on household create a blank household person

iwootten:

Adds a response required validator that validates that input was provided for a field. This is a copy of the InputRequired validator provided by wtforms, which checks that form-input data was provided, but additionally adds a kwarg to strip whitespace, as is already available on the Optional() validator wtforms provides. Oddly, stripping whitespace is not  an option for DataRequired or InputRequired validators in wtforms.

Fixes #941.

### How to review 
See if you agree with approach, tests pass.

### DIFF
--- a/app/validation/validators.py
+++ b/app/validation/validators.py
@@ -19,6 +19,32 @@ class IntegerCheck(object):
             raise validators.StopValidation(self.message)
 
 
+class ResponseRequired(object):
+    """
+    Validates that input was provided for this field. This is a copy of the
+    InputRequired validator provided by wtforms, which checks that form-input data
+    was provided, but additionally adds a kwarg to strip whitespace, as is available
+    on the Optional() validator wtforms provides. Oddly, stripping whitespace is not
+    an option for DataRequired or InputRequired validators in wtforms.
+    """
+    field_flags = ('required', )
+
+    def __init__(self, message=None, strip_whitespace=True):
+        if not message:
+            message = error_messages['MANDATORY']
+        self.message = message
+
+        if strip_whitespace:
+            self.string_check = lambda s: s.strip()
+        else:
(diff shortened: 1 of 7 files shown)


app/validation/validators.py
```diff
@@ -19,6 +19,32 @@ def __call__(self, form, field):
...
+        self.message = message
+
+        if strip_whitespace:
+            self.string_check = lambda s: s.strip()
+        else:
+            self.string_check = lambda s: s
+
+    def __call__(self, form, field):
+        if not field.raw_data or not field.raw_data[0] or not self.string_check(field.raw_data[0]):
+            field.errors[:] = []
```
samiwel, line 44: I understand the approach you've taken and it seems reasonable, but can you explain why are we making a copy of `field.errors` (by splicing) and assigning it to an empty list?

iwootten: This was the approach taken in the original InputRequired validator this is based on, but all it's doing is completely emptying the existing list of errors for the field, because it's about to call StopValidation.

samiwel: Ah I see, it was my misunderstanding of the behaviour of slicing. The slice does a shallow copy, so makes sense that assigning to the sliced list would empty it. The notation just confused me for a moment there!

samiwel (approved):

LJBabbage (approved):

PR #952, opened 2017-02-15
Eq 912b relationship question accessibility

bennpearson:
DAC accessibility changes for the relationships question

### How to review 
- Choose census_household.json 
- Navigate to the relationships question 
- Check tags and attributes
- Check voiceover
- run tests


bennpearson:
@LJBabbage Can you add the letter to the number please (5a, 5b, 5c)

iwootten:
This change will require generate_pages to be updated to correctly support changes in selenium tests.

tests/functional/pages/surveys/census/household/household-relationships.page.js
```diff
@@ -9,67 +9,7 @@ class HouseholdRelationshipsPage extends MultipleChoiceWithOtherPage {
...
-  }
-
-  clickHouseholdRelationshipsAnswerRelationOther(instance = 0) {
-    browser.element('[id="household-relationships-answer-' + instance + '-11"]').click().pause(300)
-    return this
-  }
-
-  clickHouseholdRelationshipsAnswerUnrelatedIncludingFosterChild(instance = 0) {
-    browser.element('[id="household-relationships-answer-' + instance + '-12"]').click().pause(300)
+    browser.selectByValue('#household-relationships-answer-0', 'Husband or wife')
```
iwootten: This will only ever change the first select on the page. There could be multiple, so we need to use the instance value to update the correct select.

tests/functional/spec/household-relationship.spec.js
```diff
@@ -10,63 +10,6 @@ describe('Household relationship', function() {
 
   var schema = 'test_relationship_household.json'
 
-  it('Given I edit a relationship, when closing that row, I should be then able to edit the next unanswered relationship', function() {
```
iwootten, line 13: These tests need to be updated so they test the same behaviour with the new selects.

tests/functional/pages/surveys/census/household/household-relationships.page.js
iwootten: I was under the impression that selecting by value was a bad thing, as if the schema value changes it impacts the tests. We don't appear to be using it elsewhere? The existing tests use the radio id rather than a string.

iwootten (changes requested):

app/templates/partials/questions/relationship.html
```diff
@@ -1,26 +1,25 @@
 {% import 'macros/helpers.html' as helpers %}
+<div class="question">
+  <fieldset class="question__fieldset">
+    <legend class="question__legend">
```
hamishtaplin: The title should be in a `h2` and also hidden `legend` field. Take a look at the other question templates for the structure.

app/templates/partials/widgets/relationship_widget.html
```diff
@@ -1,50 +1,28 @@
...
+        {% set label = {
+          "class": "label venus",
+          "for": select_option.id
+        } %}
+        <label {{label|xmlattr}}>
+          {{current_person}} is {{other_person}}'s
+        </label>
+        {% set select = {
+          "class": "input input--select",
+          "name": 'select',
```
hamishtaplin: Is this name (select) correct? Usually it's something a bit less generic.

tests/functional/pages/surveys/census/household/household-relationships.page.js
hamishtaplin: @iwootten I agree. Also, aren't these supposed to be autogenerated anyway?

hamishtaplin (changes requested):

tests/functional/pages/surveys/census/household/household-relationships.page.js
iwootten: @hamish Yes, this is as I mentioned to do with generate_pages.py, I can take a look at that if it makes things easier.

tests/functional/generate_pages.py
```diff
@@ -76,8 +76,8 @@
 
 """
 
-RELATIONSHIP_RADIO_CLICKER = r"""  click{optionName}(instance = 0) {
-    browser.selectByValue('[id="{answerId}' + instance + '"]', value)
+RELATIONSHIP_RADIO_CLICKER = r"""  click{optionName}(instance = 0, value) {
```
collisdigital: Is it valid to have a default first argument but not second?

collisdigital:
Make sure this gets squashed before merging

app/templates/partials/questions/relationship.html
```diff
@@ -1,26 +1,28 @@
...
+  <h2 class="question__title neptune">
+    {%- if question_number -%}
+      <span class="question__number">{{question_number}}{{current_location.group_instance|format_number_to_alphabetic_letter}}. </span>
+    {% endif %}
+    {{question_title|safe}}
+  </h2>
+
+  {{question_description|safe}}
+  <fieldset>
+    <legend class="field__legend mars u-vh">
```
hamishtaplin: This doesn't need the `.mars` class as it's hidden.

app/templates/partials/questions/relationship.html
```diff
@@ -1,26 +1,28 @@
...
       {%- if question_number -%}
-        <span class="question__number">{{question_number}}. </span>
+        <span class="question__number">{{question_number}}{{current_location.group_instance|format_number_to_alphabetic_letter}}. </span>
```
hamishtaplin: I would remove the number from the `legend` to be consistent with other templates.

hamishtaplin (changes requested):

hamishtaplin (approved):

benjefferies (changes requested):
I can't complete a survey with relationship answers in them now.

Steps to reproduce.

1. Open `test_relationship_household` schema
1. Add 2 people
1. Select the relationship
1. Submit the answers.

Expected.
I see the thank you page

Actual
Redirected to relationship question

tests/functional/spec/household-relationship.spec.js
bennpearson: `edit` has been removed so think this is ok now

tests/functional/pages/surveys/relationship/household-relationship.page.js
```diff
@@ -2,33 +2,13 @@ import QuestionPage from '../question.page'
...
-  getRelationshipLabelAt(index) {
-    const elementId = browser.elements('[data-qa="relationship-title"]').value[index]
-    return browser.elementIdText(elementId.ELEMENT).value
-  }
-
   setHusbandOrWifeRelationship(index, relationshipIndex) {
-    const id = this.buildRelationshipAnswerId(index, 0)
-    browser.waitForExist(id)
-    browser.element(id).click().pause(300)
+    browser.selectByValue('#who-is-related-0', 'Husband or wife')
```
iwootten: Shouldn't these methods also be updated to be using the instance, value arguments as per the other relationship page?

tests/functional/spec/household-relationship.spec.js
iwootten: Fair enough, but only one of these tests relates to editing details. The other two test behaviour that should be be updated.

iwootten (changes requested):

tests/functional/generate_pages.py
```diff
@@ -76,8 +76,8 @@
 
 """
 
-RELATIONSHIP_RADIO_CLICKER = r"""  click{optionName}(instance = 0) {
-    browser.element('[id="{answerId}-' + instance + '-{optionIndex}"]').click().pause(300)
+RELATIONSHIP_RADIO_CLICKER = r"""  click{optionName}(instance = 0, value) {
```
benjefferies: Should be RELATIONSHIP_DROPDOWN_CLICKER

tests/functional/pages/surveys/census/household/household-relationships.page.js
```diff
@@ -8,68 +8,68 @@ class HouseholdRelationshipsPage extends MultipleChoiceWithOtherPage {
...
-  clickHouseholdRelationshipsAnswerHusbandOrWife(instance = 0) {
-    browser.element('[id="household-relationships-answer-' + instance + '-0"]').click().pause(300)
+  clickHouseholdRelationshipsAnswerHusbandOrWife(instance = 0, value) {
+    browser.selectByValue('[id="household-relationships-answer-' + instance + '"]', value)
     return this
   }
 
-  clickHouseholdRelationshipsAnswerSameSexCivilPartner(instance = 0) {
-    browser.element('[id="household-relationships-answer-' + instance + '-1"]').click().pause(300)
+  clickHouseholdRelationshipsAnswerSameSexCivilPartner(instance = 0, value) {
```
benjefferies: It seems wrong the have lots of `clickHouseholdRelationshipsAnswerXXXXX` and the pass the value in to select. I means all of the other `clickHouseholdRelationshipsAnswerXXXXX` methods are redundant. We should do one of the other.

1) Have one method where the value is passed in or
2) Have one method per answer type.

To be consistent I'd suggest the latter.

tests/functional/spec/household-relationship.spec.js
```diff
@@ -10,63 +10,6 @@ describe('Household relationship', function() {
...
-    HouseholdRelationshipPage
-      .setHusbandOrWifeRelationship(0)
-      .toggleEditCloseButton(0)
-    browser.pause(100)
-    HouseholdRelationshipPage.toggleEditCloseButton(0)
-
-    expect(browser.isVisible('[data-relationship-name="John Doe"]')).to.be.true
-  });
-
-  it('Given I have added Joe Bloggs, Jane Doe and John Doe to my household when I set Joe Bloggs relationships then I should see the questions `Joe Bloggs is the husband or wife of Jane Doe` and `Joe Bloggs is the son or daughter of John Doe`', function() {
```
benjefferies, line 33: This test should not be delete but updated

tests/functional/spec/household-relationship.spec.js
```diff
@@ -10,63 +10,6 @@ describe('Household relationship', function() {
...
-    // When
-    HouseholdRelationshipPage.setHusbandOrWifeRelationship(0)
-      .setSonOrDaughterRelationship(1)
-
-    // Then
-    expect(HouseholdRelationshipPage.getRelationshipLabelAt(0)).to.have.string('Joe Bloggs is the husband or wife of Jane Doe')
-    expect(HouseholdRelationshipPage.getRelationshipLabelAt(1)).to.have.string('Joe Bloggs is the son or daughter of John Doe');
-  })
-
-  it('Given I have answered how Joe Bloggs is related Jane Doe and John Doe when I go to the next relationship question then I should see the questions `Jane Doe is the … of John Doe`', function() {
```
benjefferies, line 52: This test should not be delete but updated

benjefferies (changes requested):

tests/functional/spec/household-relationship.spec.js
benjefferies: I stand corrected. The label isn't piped anymore.

tests/functional/spec/household-relationship.spec.js
benjefferies: I stand corrected. The label isn't piped anymore.

app/jinja_filters.py
```diff
@@ -71,3 +73,9 @@ def format_household_summary(names):
 
         return person_list
     return ''
+
+
+@blueprint.app_template_filter()
+def format_number_to_alphabetic_letter(number):
+    if int(number) < 26:
```
benjefferies: Should have a test for this and also need to check for `int(number) >= 0`

benjefferies (changes requested):

benjefferies (approved):

iwootten (approved):

LJBabbage (approved):

iwootten:
@bennpearson - Seems like there's still a few points outstanding on the HouseholdRelationshipPage, to do with method/arguments etc.

tests/functional/generate_pages.py
```diff
@@ -148,7 +148,10 @@ def process_relationship_options(answer_id, options, template, page_spec):
         option_index = "{index}".format(index=index)
         template_with_index = template.replace("{optionIndex}", option_index)
         template_with_answer_id = template_with_index.replace("{answerId}", answer_id)
-        page_spec.write(template_with_answer_id.replace("{optionName}", generate_camel_case_from_id(answer_id) + option_name))
+        template_with_option_value = template_with_answer_id.replace("{optionValue}", option['value'])
+        template_with_option_name = template_with_option_value.replace("{optionName}", generate_camel_case_from_id(answer_id) + option_name)
+        print(template_with_option_name)
```
benjefferies: Print statement should be removed

benjefferies (changes requested):

tests/functional/pages/surveys/relationship/household-relationship.page.js
iwootten: This page has yet to be updated as per the other...?

iwootten (changes requested):

iwootten (approved):

benjefferies (approved):

PR #954, opened 2017-02-15
Remove view sections button from intro and thank-you pages.

### LINKED ISSUE #921: Mobile: Displays view section link on the UKIS landing page
### Expected behaviour
Must not display 'View Sections' link on the UKIS landing page or on the final successful submission screen

### Actual behaviour
Displaying View section link on the UKIS landing page and it does not do anything when clicked

**Note:** **Does not display navigation/view section link on landing page when launched from a laptop/ tablets**

Launch UKIS survey on a mobile

#### Operating System
Iphone, Samsung galaxy
### Screenshot
screen shot 2017-02-07 at 14 06 23

samiwel:
Fixes #921.

### How to review 
Re-test the defect.
Verify that the view sections link is no longer displayed on both the introduction and successful submission pages for UKIS.
All existing tests and checks should pass.

### DIFF
--- a/tests/functional/spec/navigation.spec.js
+++ b/tests/functional/spec/navigation.spec.js
@@ -1,9 +1,14 @@
 import chai from 'chai'
+import landingPage from '../pages/landing.page'
+import PercentagePage from '../pages/surveys/percentage/percentage.page'
+import SummaryPage from '../pages/summary.page'
 import {
   startQuestionnaire,
+  openQuestionnaire,
   setMobileViewport,
   openMobileNavigation,
-  closeMobileNavigation
+  closeMobileNavigation,
+  isViewSectionsVisible
 } from '../helpers'
 
 const expect = chai.expect
@@ -26,5 +31,30 @@ describe('Navigation', function() {
     navIsNotVisible = closeMobileNavigation()
     // Then
     expect(navIsNotVisible).to.be.true
+  })
+
+  it('Given survey launched on mobile device, When on Introduction page, Then view sections link should not be displayed.', function() {
+    // Given
+    setMobileViewport()
+
+    // When
+    openQuestionnaire('test_percentage.json')
+
+    // Then
+    expect(isViewSectionsVisible()).to.be.false
+  })
+
+  it('Given survey launched on mobile device, When on Thank-You page, Then view sections link should not be displayed.', function() {
+    // Given
+    setMobileViewport()
+    startQuestionnaire('test_percentage.json')
+
+    // When
+    expect(isViewSectionsVisible()).to.be.true
+    PercentagePage.submit()
+    SummaryPage.submit()
+
+    // Then
+    expect(isViewSectionsVisible()).to.be.false
   })
 })
--- a/app/data/test_percentage.json
+++ b/app/data/test_percentage.json
@@ -17,6 +17,7 @@
       "NEGATIVE_INTEGER": "Number cannot be less than zero",
       "NOT_INTEGER": "Please enter an integer"
     },
+    "navigation" : true,
     "groups": [
         {
             "blocks": [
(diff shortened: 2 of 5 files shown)


app/templates/thank-you.html
```diff
@@ -2,6 +2,8 @@
 
 {% block page_title %}Submission Successful{% endblock %}
 
+{% block subheader %}{% endblock subheader %}
```
hamishtaplin: This will remove the 'previous' link too as it's in that block.

There's a value `navigation` that is exposed to `layouts/_base.html` that is supposed to switch the nav on/off. Can this be used instead?

hamishtaplin (changes requested):

app/templates/thank-you.html
samiwel: Can certainly look at it, but what is the expectation for the previous link on the introduction and thank you pages?

app/templates/thank-you.html
```diff
@@ -2,6 +2,8 @@
 
 {% block page_title %}Submission Successful{% endblock %}
 
+{% set navigation = false %}
```
iwootten: Given each of these templates have their own get endpoints, it might make sense to set the navigation flag as part of the context on the backend?

iwootten (changes requested):

app/templates/thank-you.html
LJBabbage: Makes sense, you then wouldn't need it in the lower templates

tests/functional/pages/surveys/question.page.js
```diff
@@ -33,6 +33,11 @@ class QuestionPage {
     return browser.click('a[id="top-previous"]')
   }
 
+  next() {
```
jonnyshaw89: What does this give us that submit() doesn't?

tests/functional/spec/navigation.spec.js
```diff
@@ -1,30 +1,61 @@
...
+  closeMobileNavigation,
+  isViewSectionsVisible
 } from '../helpers'
 
 const expect = chai.expect
 
 describe('Navigation', function() {
-  it('Given a page with navigation, a user on mobile should be able to access it via the associated button', function() {
-    let navIsVisible
+//  it('Given a page with navigation, a user on mobile should be able to access it via the associated button', function() {
```
jonnyshaw89: Tests didn't work?

jonnyshaw89 (changes requested):

tests/functional/pages/surveys/question.page.js
samiwel: Submit verifies that you're on the correct page. The next allows me to bypass this check. The alternative would be to import 40 page objects - which is overkill for this test. All I care about is navigating through to the thank you page so this is the simplest way of doing that.

tests/functional/spec/navigation.spec.js
samiwel: Oops, these shouldn't have been commented out. I'll revert this change.

tests/functional/pages/surveys/question.page.js
jonnyshaw89: The reason that was added was because there were issues where the page it was submitting wasn't the one it thought it was. hence we got the errors in travis. i.e.. it would click the submit multiple times on a page and only change page once.

samiwel: Yeah, that's fair enough. But in this case, I'm not concerned with routing. I just want to get to the thank you page so the test just keeps on hitting continue until it reaches thank you. I'd argue that having to import > 40 different files just to do that wouldn't help with readability of the test.

jonnyshaw89: Could you load a shorter survey? seems like a lot of overhead just to get to the thank-you page.

LJBabbage: Although the fix is for UKIS, it is valid for all surveys, so a shorter one would make sense. It also is something we would want going forward, so we should keep it out of the UKIS suite, as we might not always run UKIS in the future

samiwel: Used the shortest survey I could find 👍

jonnyshaw89 (approved):

LJBabbage (approved):

hamishtaplin (approved):

iwootten (approved):

PR #955, opened 2017-02-15
Resolve issues preventing Docker images working

collisdigital:
Currently when using `docker-compose up` the Postgres Docker image may not be ready when the app starts up. Due to the way wsgi does its importing at startup if the app dies at startup (which it does if it can't connect to the database) it cannot recover without restarting the docker image; this change introduces a retry mechanism for connecting to and setting-up the database with a configurable pause between attempts.

In addition, running the tests in the Docker container fail because of a few issues, including currently having to run Selenium tests which require Chrome etc. I have parameterized run_tests.sh to support running python and/or functional tests with -p and -f flags (by default it behaves as before and runs all tests). In addition I added a run_tests_python.sh and run_tests_function.sh as shortcuts which can be used if overriding the ENTRYPOINT on the Docker container on the command-line to just run the python tests for example.

### Commit Summary
- Add retry mechanism for setup of database - this improves the startup process when using docker-compose as the Postgres image may not be ready before the app
- Allow Python and Functional tests to be run independently
- Filter out .json files for schema validation to avoid file load errors - was occurring inside Docker
- Move Git clone to /tmp and install test packages - Caused pylint inside Docker to check python cryptography and fail
- Fix test_submitter not failing when RabbitMQ actually exists - Which it does inside the Docker container
- Make tests work even when ENV flags are different - Docker ENV flags didn't match the unit test
- Split unit test into separate Dockerfile and add instructions to run
- After various attempts to update Travis to use the Docker builds; some improvements have been added although using Docker hasn't been done as it slows the build down by 4-5 minutes (but it did work).

### How to review 
You should be able to run `docker-compose up` and see the database connects successfully after failing at least once (it always fails once for me); the app should work without having to restart docker.

Try running just the python unit tests in the container:

```
docker run --entrypoint=/tmp/application/scripts/run_tests_python.sh eqsurveyrunner_eq-survey-runner 
```


.travis.yml
```diff
@@ -1,50 +1,37 @@
-language: python
+language: node_js
+node_js: 6
 sudo: required
```
benjefferies: We might be able to do container builds now?

.travis.yml
```diff
@@ -1,50 +1,37 @@
...
 before_script:
-    - yarn compile
-    - ./scripts/run_app.sh
+    - docker-compose --version
+    - docker-compose -d up
+    - docker --version
+    - docker tag eqsurveyrunner_eq-survey-runner onsdigital/eq-survey-runner
+    - docker build -t onsdigital/eq-survey-runner-unit-tests -f Dockerfile.test .
+    - docker run onsdigital/eq-survey-runner-unit-tests
+    - yarn
```
benjefferies: Do we need to run yarn? Won't this run in the container?

.travis.yml
```diff
@@ -1,50 +1,37 @@
...
-
 before_script:
-    - yarn compile
-    - ./scripts/run_app.sh
+    - docker-compose --version
+    - docker-compose -d up
+    - docker --version
+    - docker tag eqsurveyrunner_eq-survey-runner onsdigital/eq-survey-runner
+    - docker build -t onsdigital/eq-survey-runner-unit-tests -f Dockerfile.test .
+    - docker run onsdigital/eq-survey-runner-unit-tests
```
benjefferies: Should we do this on script?

scripts/run_tests.sh
```diff
@@ -3,25 +3,33 @@
 # Run project tests
 #
 # NOTE: This script expects to be run from the project root with
-# ./scripts/run_tests.sh
-
-# Use default environment vars for localhost if not already set
-export EQ_CLOUDWATCH_LOGGING=False
+# ./scripts/run_tests.sh [-u] [-f]
```
benjefferies: Looks like we are adding a lot of logic to support option parameters. Why not split the two bits of logic into `run_tests_functional.sh` and `run_tests_unit.sh` where this run_tests.sh wraps both?

benjefferies (changes requested):

.travis.yml
collisdigital: It's needed to run the Selenium tests; did discuss with Hamish about splitting the testing part out from the app part, they're currently all one thing.

scripts/run_tests.sh
collisdigital: good idea

collisdigital: done

.travis.yml
```diff
@@ -1,46 +1,48 @@
...
+    - pyenv global $(cat .python-version)
     - pip install -U pip wheel
     - pip install --require-hashes -r requirements.txt
-    - pip install -e git+https://github.com/ONSdigital/cryptography.git@4a90c254278231d7defeac304a3cfd752e96e786#egg=cryptography
     - pip install -r requirements_for_test.txt
-
-before_script:
-    - yarn compile
-    - ./scripts/run_app.sh
+    - pushd /tmp
```
benjefferies, line 34: Cool, didn't know about pushd/popd

jonnyshaw89 (approved):

benjefferies (approved):

.travis.yml
collisdigital: yeah useful; standard *nix commands

PR #956, opened 2017-02-16
Fix previous link relationship question

LJBabbage:
The household relationship grid questions in census needs to have a previous button, both on the first block (returns to overnight-visitors) and any repeating blocks (instance before). Although not ideal after talking to Andrew we decided to hard code it for now

### How to review 
With Census, fill in multiple people (>2) on the who lives here page, navigate to relationship grid, making sure the first block and all repeating blocks have a previous working button

### DIFF
--- a/app/questionnaire/path_finder.py
+++ b/app/questionnaire/path_finder.py
@@ -221,6 +221,15 @@ class PathFinder:
         :return: The previous location as a dict
         :return:
         """
+
+        if current_location.group_id == 'who-lives-here-relationship':
+            return self._relationship_previous_location(current_location.group_instance)
+
+        is_first_block_for_group = SchemaHelper.is_first_block_id_for_group(self.survey_json, current_location.group_id, current_location.block_id)
+
+        if is_first_block_for_group or current_location.block_id == 'thank-you':
+            return None
+
         location_path = self.get_location_path(current_location.group_id, current_location.group_instance)
         current_location_index = PathFinder._get_current_location_index(location_path, current_location)
 
@@ -243,3 +252,11 @@ class PathFinder:
                 return incomplete_blocks[0]
 
         return location_path[0]
+
+    @staticmethod
+    def _relationship_previous_location(current_group_instance):
+        if current_group_instance == 0:
+            previous_location = Location('who-lives-here', 0, 'overnight-visitors')
+        else:
+            previous_location = Location('who-lives-here-relationship', current_group_instance - 1, 'household-relationships')
+        return previous_location
--- a/app/views/questionnaire.py
+++ b/app/views/questionnaire.py
@@ -521,9 +521,7 @@ def _build_template(current_location, context=None, template=None, display_navig
 
     previous_url = None
 
-    is_first_block_for_group = SchemaHelper.is_first_block_id_for_group(g.schema_json, current_location.group_id, current_location.block_id)
-
-    if previous_location is not None and not is_first_block_for_group and not current_location.block_id == 'thank-you':
+    if previous_location is not None:
         previous_url = previous_location.url(metadata)
 
     if user_logout:
--- a/tests/app/questionnaire/test_path_finder.py
+++ b/tests/app/questionnaire/test_path_finder.py
@@ -208,6 +208,7 @@ class TestPathFinder(unittest.TestCase):  # pylint: disable=too-many-public-meth
 
         self.assertEqual(expected_next_location, next_location)
 
+    @pytest.mark.xfail(reason="The first group should go back to the introduction, however we block previous on the first block in a group", strict=True)
     def test_get_previous_location_introduction(self):
         survey = load_schema_file("0_star_wars.json")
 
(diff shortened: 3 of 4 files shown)


app/views/questionnaire.py
```diff
@@ -491,10 +491,20 @@ def _build_template(current_location, context=None, template=None):
 
     if previous_location is not None and not is_first_block_for_group and not current_location.block_id == 'thank-you':
         previous_url = previous_location.url(metadata)
+    elif current_location.group_id == 'who-lives-here-relationship':
+        previous_url = _relationship_grid_previous_url(current_location.group_instance).url(metadata)
 
     return _render_template(context, current_location.block_id, front_end_navigation, metadata_context, current_location, previous_url, template)
 
 
+def _relationship_grid_previous_url(current_instance):
```
iwootten: I don't think we need to have "_grid" in the method name here, we don't use it elsewhere and there's only 1 type of relationship page. 

It might also make sense to clarify the type of instance that is passed in eg. current_group_instance, given we have other types of instances in the system.

tests/integration/questionnaire/test_questionnaire_previous_link.py
```diff
@@ -74,3 +74,43 @@ def test_previous_link_doesnt_appear_on_thank_you(self):
         self.assertTrue(final_url.endswith('thank-you'))
         self.assertNotIn('Previous', content)
 
+    def test_previous_link_on_relationship_grid(self):
```
iwootten: Similarly on the "_grid" method naming here.

iwootten (changes requested):

app/views/questionnaire.py
benjefferies: It might be nice to have one method to build the previous url. I think it would make the behemoth _build_template method nicer too

iwootten (approved):

app/views/questionnaire.py
```diff
@@ -491,10 +491,20 @@ def _build_template(current_location, context=None, template=None):
 
     if previous_location is not None and not is_first_block_for_group and not current_location.block_id == 'thank-you':
         previous_url = previous_location.url(metadata)
+    elif current_location.group_id == 'who-lives-here-relationship':
```
benjefferies: Is this the right place to fix this issue? Shouldn't the previous location generated by the path_finder get the previous group instance?

benjefferies (changes requested):

app/views/questionnaire.py
LJBabbage: Will need to take is_first_block_for_group or current_location.block_id == 'thank-you' with me to do that, although I don't see a problem with that, potentially they should be down there anyway

benjefferies:
Not sure if the change fully reflects what I mean. Shouldn't the current - 1 of get_location_path path be the previous group instance?

LJBabbage:
We just can't do that, that was our original problem and why we never had previous on the first block of a group. If they jump around the survey using navigation the previous group (-1) might not be the group they were on before.

LJBabbage:
Updated. One test was removed as I realised it was already covered. The other one was misleading. The path_finder.get_previous_location was returning a value which was then being hidden by the view, so it didn't reflect what was actually happening. We don't put a link to the introduction from any block on the front-end so why assert we do.

app/questionnaire/path_finder.py
```diff
@@ -243,3 +252,11 @@ def get_latest_location(self, completed_blocks=None):
                 return incomplete_blocks[0]
 
         return location_path[0]
+
+    @staticmethod
+    def _relationship_previous_url(current_group_instance):
```
benjefferies: Should be updated to reflect it's not a url

app/questionnaire/path_finder.py
```diff
@@ -243,3 +252,11 @@ def get_latest_location(self, completed_blocks=None):
...
+        if current_group_instance == 0:
+            previous_url = Location('who-lives-here', 0, 'overnight-visitors')
```
benjefferies: Should be previous_location

tests/app/questionnaire/test_path_finder.py
```diff
@@ -217,7 +217,7 @@ def test_get_previous_location_introduction(self):
 
         previous_location = path_finder.get_previous_location(current_location=first_location)
 
-        self.assertEqual('introduction', previous_location.block_id)
+        self.assertEqual(None, previous_location)
```
benjefferies: Why has this test changed? Looks unrelated

tests/app/questionnaire/test_path_finder.py
```diff
@@ -519,18 +519,6 @@ def test_repeating_groups_no_of_answers_minus_one(self):
 
         self.assertEqual(expected_path, path_finder.get_routing_path())
 
-    def test_repeating_groups_previous_location_introduction(self):
```
benjefferies, line 522: I think this test is still valid?

benjefferies (changes requested):

tests/app/questionnaire/test_path_finder.py
LJBabbage: See note

tests/app/questionnaire/test_path_finder.py
LJBabbage: See note

benjefferies (approved):

PR #915, opened 2017-02-03
885 generate translated schemas at runtime

benjefferies:
### What is the context of this PR?
* Add shell script to translate schemas
* Run schema translate on run_app.sh
* Removing checked in translated schemas
* Add full run through of welsh individual schema

### How to review 
- [x] Checkout branch and run `./scripts/run_app.sh` and observe generated schemas
- [x] Travis builds and tests pass
- [x] Can deploy to AWS environment

PR #957, opened 2017-02-16
Introduction and summary as blocks

benjefferies:
Removing hard coded logic around introduction and summary pages to allow
them to behave like blocks.


app/data/0_rogue_one.json
```diff
@@ -166,7 +176,7 @@
                         {
                             "goto": {
                                 "location": "block",
-                                "id" : "5e633f04-073a-44c0-a9da-a7cc2116b937"
+                                "id" : "rouge-one-group-last-block"
```
iwootten: sp. rogue not rouge

app/data/0_rogue_one.json
```diff
@@ -8,13 +8,23 @@
...
-        "description": ""
-    },
     "groups": [
         {
             "blocks": [
                 {
+                    "type": "introduction",
+                    "id": "introduction",
+                    "title": "Introduction",
+                    "sections": [
```
iwootten: Seems like the addition of empty lists is a workaround for our parser which incorrectly assumes there must be 1 of everything. This is throughout all our other schemas too. Seems like this is what travis is complaining about too.

tests/integration/star_wars/star_wars_test_urls.py
```diff
@@ -1,11 +1,11 @@
 STAR_WARS_BASE = '/questionnaire/0/star_wars/789/'
-STAR_WARS_INTRODUCTION = STAR_WARS_BASE + 'introduction'
+STAR_WARS_INTRODUCTION = STAR_WARS_BASE + '14ba4707-321d-441d-8d21-b8367366e766/0/introduction'
```
iwootten, line 2: Would it make sense to update some of these ids to be more readable?

iwootten (dismissed):

iwootten:
Has this accidentally ended up with https://github.com/ONSdigital/eq-survey-runner/pull/915 in it? Seems to have quite a high number of deletions.

benjefferies:
Good spot. I deleted them because they were failing because they need to be updated to be the same as the main census schemas but the will be generated. I was hoping no one would notice and they would get merged out. It would also mean @ajmaddaford not raising anymore issues with #915 and actually approving 👍

app/data/0_rogue_one.json
```diff
@@ -364,8 +366,21 @@
...
-            "id": "82504ab4-de5d-41fd-9a86-9ad5ea483296",
+            "id": "rogue-one-group",
             "title": ""
+        },
+        {
+            "blocks": [
+                {
+                    "type": "summary",
+                    "id": "summary",
+                    "title": ""
```
LJBabbage: Blocks according to the schema definition don't need titles, we might as well remove all the summary titles if they are going to be blank

app/data/1_0102.json
```diff
@@ -8,18 +8,21 @@
...
-          "reasons for changes to figures"
-        ]
-    },
     "groups": [
         {
             "blocks": [
                 {
+                    "type": "introduction",
+                    "id": "introduction",
+                    "title": "",
```
LJBabbage: Same as the summary titles

app/data/mwss_4.json
```diff
@@ -1,18 +1,21 @@
...
-        ]
-    },
     "groups": [
         {
             "blocks": [
                 {
+                    "type": "introduction",
+                    "id": "introduction",
+                    "title": "",
+                    "description": "",
```
LJBabbage: Same again, but for description

app/helpers/schema_helper.py
```diff
@@ -35,7 +35,7 @@ def is_first_block_id_for_group(cls, survey_json, group_id, block_id):
 
     @staticmethod
     def get_last_block_id(survey_json):
-        return survey_json['groups'][0]['blocks'][-1]['id']
+        return survey_json['groups'][-1]['blocks'][-1]['id']
```
LJBabbage, line 34: Sure there is a good reason, why have we gone from the last block of the first group to the last block of the last group?

app/helpers/schema_helper.py
```diff
@@ -129,23 +129,27 @@ def get_aliases(cls, survey_json):
...
+    def get_answers_by_id_for_block(cls, block_json):
         answers = {}
-        for section_json in block_json['sections']:
+        for section_json in cls._get_sections_in_block(block_json):
             for question_json in section_json['questions']:
                 for answer_json in question_json['answers']:
                     answers[answer_json['id']] = answer_json
         return answers
 
+    @staticmethod
```
LJBabbage, line 137: Not sure this adds much, also makes it a little inconsistent as you could say we should then do it for section_json['questions'] and question_json['answers'] as well

app/questionnaire/path_finder.py
```diff
@@ -154,17 +140,11 @@ def get_location_path(self, group_id=None, group_instance=0):
...
-        location_path += routing_path
 
-        if can_reach_summary:
-            for block_id in PathFinder.CLOSING_INTERSTITIAL_PATH:
-                location_path.append(Location(SchemaHelper.get_last_group_id(self.survey_json), 0, block_id))
+        if self.can_reach_summary(routing_path):
+            routing_path.append(Location(SchemaHelper.get_last_group_id(self.survey_json), 0, 'thank-you'))
 
-        return location_path
+        return routing_path
```
LJBabbage: The function is called get_location_path and is returning routing path, one needs to be changed

app/templating/introduction_context.py
```diff
@@ -1,15 +1,18 @@
+from app.helpers.schema_helper import SchemaHelper
+
+
 def get_introduction_context(schema_json):
     """
     :param schema_json: the schema for the current questionnaire
     :return: introduction context
     """
-    introduction = schema_json['introduction']
+    intro_block = next(block for block in SchemaHelper.get_blocks(schema_json) if block['type'] == 'introduction')
```
LJBabbage: Not a fan of abbreviations as we use introduction everywhere else

app/views/questionnaire.py
```diff
@@ -196,51 +205,16 @@ def post_household_composition(eq_id, form_type, collection_id, group_id):
...
-    next_location_url = next_location.url(metadata)
-    logger.debug("redirecting", url=next_location_url)
-    return redirect(next_location_url)
-
-
-@questionnaire_blueprint.route('summary', methods=["GET"])
+@questionnaire_blueprint.route('<group_id>/0/summary', methods=["GET"])
 @login_required
-def get_summary(eq_id, form_type, collection_id):  # pylint: disable=unused-argument
+def get_summary(eq_id, form_type, collection_id, group_id):  # pylint: disable=unused-argument
```
LJBabbage: Added group_id?

app/views/questionnaire.py
```diff
@@ -296,9 +253,9 @@ def get_session_expired(eq_id, form_type, collection_id):  # pylint: disable=unu
     return _render_template({}, block_id='session-expired')
 
 
-@questionnaire_blueprint.route('submit-answers', methods=["POST"])
+@questionnaire_blueprint.route('<group_id>/0/submit-answers', methods=["POST"])
 @login_required
-def submit_answers(eq_id, form_type, collection_id):
+def submit_answers(eq_id, form_type, collection_id, group_id):
```
LJBabbage: Same as above

LJBabbage (changes requested):

app/views/questionnaire.py
```diff
@@ -530,6 +485,15 @@ def _build_template(current_location, context=None, template=None, display_navig
     return _render_template(context, current_location.block_id, front_end_navigation, metadata_context, current_location, previous_url, template)
 
 
+def _get_front_end_navigation(answer_store, completed_blocks, current_location, metadata):
+    if not current_location.is_interstitial():
```
LJBabbage: This is really confusing and needs changing, The function looks like

def is_interstitial(self):
        return self.block_id in ['introduction', 'summary', 'thank-you']

It would be very easy to call it thinking it covers all interstitial, however we use the term interstitial in an another manner to mean blocks inbetween other blocks with no questions on. When I saw it, I thought we wouldn't have navigation on the "this section is finished, up next is" blocks

LJBabbage (changes requested):

app/helpers/schema_helper.py
benjefferies: It's used by get_last_location which get's the last group id. The test schema it was working on initially only had one group but I added another so it highlighted the flaw.

app/data/0_rogue_one.json
benjefferies: Good spot, thought it was required when I first looked!

app/views/questionnaire.py
benjefferies: Yeah this needs a bit of a rework. All of questionnaire.py does before this goes in I think.

LJBabbage: Seems like a double check as well as we do
```if block_json['type'] == 'questionnaire' or block_json['type'] == 'interstitial' else None```
further down. We should be able to get rid of one

**Review by LJBabbage — CHANGES_REQUESTED** [review]

File: app/data/1_0102.json (hunk 5 above)
Comment (benjefferies) [reply app/data/1_0102.json]: Done

File: app/data/mwss_4.json (hunk 6 above)
Comment (benjefferies) [reply app/data/mwss_4.json]: Done

File: app/data/0_rogue_one.json (hunk 4 above)
Comment (benjefferies) [reply app/data/0_rogue_one.json]: Done

File: app/helpers/schema_helper.py (hunk 8 above)
Comment (benjefferies) [reply app/helpers/schema_helper.py:137]: I have changed it so sections are not mandatory anymore whereas questions are

File: app/templating/introduction_context.py (hunk 10 above)
Comment (benjefferies) [reply app/templating/introduction_context.py]: Me too so not sure why I did it! I might be able to delete all of this with some refactoring anyway

**Comment by benjefferies** [comment]
@LJBabbage I think I have addressed everything now. Feel free to review again when you can.

File: app/questionnaire/path_finder.py [hunk 14]
```diff
@@ -207,7 +172,10 @@ def get_next_location(self, current_location):
         :param current_location:
         :return: The next location as a dict
         """
-        location_path = self.get_location_path(current_location.group_id, current_location.group_instance)
+        location_path = self.get_routing_path(current_location.group_id, current_location.group_instance)
+        if self.can_reach_summary(location_path):
+            location_path.append(Location(SchemaHelper.get_last_group_id(self.survey_json), 0, 'thank-you'))
```
iwootten: Can the thankyou location be created on init? Doesn't seem like something that should only be accessible by calling this method.

benjefferies: It's the only place now using it so don't think it needs to be more global at this moment. I have refactored this out for clarity though.

tests/integration/star_wars/star_wars_test_urls.py
benjefferies: Bit the bullet and sorted them all out under https://github.com/ONSdigital/eq-survey-runner/pull/987/files

LJBabbage (approved):

app/questionnaire/path_finder.py
```diff
@@ -207,7 +172,11 @@ def get_next_location(self, current_location):
         :param current_location:
         :return: The next location as a dict
         """
-        location_path = self.get_location_path(current_location.group_id, current_location.group_instance)
+        location_path = self.get_routing_path(current_location.group_id, current_location.group_instance)
+        if self.can_reach_summary(location_path):
+            thank_you_location = Location(SchemaHelper.get_last_group_id(self.survey_json), 0, 'thank-you')
+            location_path.append(thank_you_location)
```
jonnyshaw89: Do we need thank-you on the location_path. chatting so @ajmaddaford maybe the submit answers should just be coded to redirect there on success?

jonnyshaw89 (changes requested):

app/questionnaire/path_finder.py
benjefferies: Done

jonnyshaw89:
@benjefferies i think path_finder.can_reach_summary is now redundant?

jonnyshaw89 (approved):

jonnyshaw89:
@benjefferies squishy rebaseiness

tests/app/data/test_invalid_routing.json
```diff
@@ -8,12 +8,12 @@
     "description": "",
     "theme": "default",
     "legal_basis": "StatisticsOfTradeAct",
-    "introduction": {
+    "Introduction": {
```
ajmaddaford: The case of the introduction property is wrong in a few places. Also might not be needed at all in some places.

tests/app/data/test_schema_validation.py
```diff
@@ -163,9 +168,9 @@ def validate_routing_rules_has_default_if_not_all_answers_routed(file, json_to_v
         errors = []
 
         for block in SchemaHelper.get_blocks(json_to_validate):
-            for section in (s for s in block['sections'] if 'sections' in block):
-                for question in (q for q in section['questions'] if 'questions' in section):
-                    for answer in (a for a in question['answers'] if 'answers' in question):
+            for section in block.get('sections', []):
+                for question in section.get('questions', []):
+                    for answer in question.get('answers', []):
```
ajmaddaford, line 173: Nice, this is much clearer.

app/questionnaire/path_finder.py
```diff
@@ -207,7 +146,8 @@ def get_next_location(self, current_location):
         :param current_location:
         :return: The next location as a dict
         """
-        location_path = self.get_location_path(current_location.group_id, current_location.group_instance)
+        location_path = self.get_routing_path(current_location.group_id, current_location.group_instance)
```
ajmaddaford: I think we can rename all occurrences of `location_path` with `routing_path` now that we no longer have the concept of a location path.

app/templating/summary_context.py
```diff
@@ -17,8 +17,8 @@ def build_summary_rendering_context(schema_json, answer_store, metadata):
     sections = []
     for group in schema_json['groups']:
         for block in group['blocks']:
-            if block['id'] in [location.block_id for location in path]:
-                if "type" not in block or block['type'] != "interstitial":
+            if block['id'] in [location.block_id for location in path] and block['type'] == 'Questionnaire':
+                if "type" not in block or block['type'] != "Interstitial":
```
ajmaddaford: I don't think this if statement is needed any more (only `Questionnaire` blocks will get through to this line)

app/views/questionnaire.py
```diff
@@ -75,212 +73,122 @@ def save_questionnaire_store(response):
...
-        raise NotFound
+    is_valid_location = valid_group and current_location in path_finder.get_routing_path(group_id, group_instance)
+    if not is_valid_location:
+        return _redirect_to_latest_location(collection_id, eq_id, form_type, path_finder)
 
     block = _render_schema(current_location)
+    block_type = block['type']
+    is_skipping_to_end = block_type in ['Summary', 'Confirmation'] and current_location != path_finder.get_latest_location(get_completed_blocks(current_user))
+    if is_skipping_to_end:
+        return _redirect_to_latest_location(collection_id, eq_id, form_type, path_finder)
```
ajmaddaford: `_redirect_to_latest_location` calls `get_latest_location` again meaning that compared to the previous code there is an extra call to generate the routing path. We should pass the latest location to `_redirect_to_latest_location` rather than path finder.

ajmaddaford (changes requested):

app/views/questionnaire.py
```diff
@@ -75,212 +73,122 @@ def save_questionnaire_store(response):
...
-    return _build_template(current_location=current_location, display_navigation=False, user_logout=True)
+    theme = g.schema_json.get('theme', None)
+    logger.debug("theme selected", theme=theme)
+    metadata = get_metadata(current_user)
+    metadata_context = build_metadata_context(metadata)
+    thank_you_template = render_theme_template(theme, template_name='thank-you.html',
+                                               meta=metadata_context,
+                                               schema_title=g.schema_json['title'],
+                                               legal_basis=g.schema_json['legal_basis'],
+                                               survey_id=g.schema_json['survey_id'])
```
ajmaddaford: `analytics_ua_id=settings.EQ_UA_ID` is not being set in the new calls to `render_theme_template` so we'll be getting no analytics for these pages

ajmaddaford (changes requested):

ajmaddaford (approved):

PR #959, opened 2017-02-17
Modify converter to correctly lookup child_answer values

### LINKED ISSUE #958: Converter incorrectly handles 'other' input data for non-census schemas
### Expected behaviour

Surveys containing checkboxes with child_answer input fields are able to be submitted successfully.

### Actual behaviour

Checkboxes with child_answer input fields using a schema with data['version'] = '0.0.1' fail on submission.

Proceed through the test_checkbox.json schema and proceed through the questionnaire and attempt to submit.

#### Browser

Chrome Version 56.0.2924.87

#### Operating System

Mac 10.12.2

### Screenshot

screen shot 2017-02-17 at 13 57 10

iwootten:

Updates the downstream converter to correctly lookup child_answer values when the data version of the schema is 0.0.1 and updates the tests to correctly represent answer_store contents.

Only visible when attempting to submit the answers from test_checkbox survey.

Fixes #958 

### How to review 

See if you agree with the fix, feedback on approach.


jonnyshaw89 (approved):

app/submitter/converter.py
```diff
@@ -120,11 +119,12 @@ def _get_checkbox_answer_data(checkboxes_with_qcode, value):
         option = next((option for option in checkboxes_with_qcode['options'] if option['value'] == user_answer), None)
 
         if option:
-            if option['value'].lower() == 'other':
+            if 'child_answer_id' in option:
+                filtered = answer_store.filter(answer_id=option['child_answer_id'], limit=1)
```
benjefferies, line 123: This would behave incorrectly if we had a repeating answer but then so would the whole rest of it I believe. Something to note if we ever get something with data version 0.0.1 with repeating

benjefferies (approved):

LJBabbage:
The summary screen is not looking good when using Other, The specify other is always shown, even if I don't select 'Other'
image However I reckon this isn't connected to this PR and an issue needs raising, will talk to Jenny. Bar that everything else is fine

LJBabbage:
Issue raised, so this can be rebased and merged

PR #961, opened 2017-02-17
Add implicit waits to fix intermittent tests

benjefferies:
Checking if this fixes the intermittency

### How to review 

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,6 +249,9 @@ EQ_ENABLE_CACHE - Enable caching of the schema
 EQ_MAX_HTTP_POST_CONTENT_LENGTH - The maximum http post content length that the system wil accept
 EQ_MAX_NUM_REPEATS - The maximum number of repeats the system will allow
 EQ_DEVELOPER_LOGGING - Enable developer style logging described here http://structlog.readthedocs.io/en/stable/development.html
+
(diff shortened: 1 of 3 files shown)


benjefferies:
This appears to be failing tests in navigation.spec.js tests are below.

```
[chrome 48.0 OS X 10.11 #0-25]       1) Given survey launched on mobile device, When on Introduction page, Then view sections link should not be displayed.
[chrome 48.0 OS X 10.11 #0-25]       2) Given survey launched on mobile device, When on Thank-You page, Then view sections link should not be displayed.
```

To rerun tests `yarn test_functional -- --spec=navigation`.

The reason it is failing is because the tests can't find the `View sections` link on the page which is correct but it should return `false` to state element not found instead it errors.

I am putting this on hold to pick up refactoring of intro/summary page so feel free to pick up or I will when I finish the other task?

samiwel:
I added the two tests you mentioned above, so I'm happy to take a look.

samiwel:
@benjefferies Fixed the failing tests by resetting the implicit wait time for tests that check for the absence of an element on the screen. It gets the tests passing again for now, but I guess there's an argument to be made for changing the test logic in the longer term. It's a tricky one, because I think asserting that something is not displayed is a valid test. I guess maybe something like `browser.elements(selector).length === 0` could be used instead.

PR #962, opened 2017-02-17
Add browserstack integration - WIP

collisdigital:
WIP

### How to review 
WIP


hamishtaplin (changes requested):
I'm sure this is on the cards but the config should be tidied up so it just overrides the bits needed when using BS

PR #967, opened 2017-02-20
Add question and answers types for Time input

LJBabbage:
Add New question type for time input

### How to review 
Using test_time_input.json check that it work as expected. Minutes can't be >59 and hours can not be >999, both should be numbers


app/validation/validators.py
```diff
@@ -167,3 +167,31 @@ def __call__(self, form, from_field, to_field):
             raise validators.ValidationError(self.messages['INVALID_DATE_RANGE_TO_FROM_SAME'])
         elif from_date > to_date:
             raise validators.ValidationError(self.messages['INVALID_DATE_RANGE_TO_BEFORE_FROM'])
+
+
+class TimeInputCheck(object):
```
iwootten, line 172: We already have a NumberRange validator that can be used for each of the hour and minute fields on any (including custom) forms. Either error message can be overridden.

app/validation/validators.py
```diff
@@ -167,3 +167,31 @@ def __call__(self, form, from_field, to_field):
...
+    def __init__(self, messages=None):
+        if not messages:
+            messages = error_messages
+        self.messages = messages
+
+    def __call__(self, form, field):
+        logger.error(self.messages)
```
iwootten, line 179: Should this be here?

app/validation/validators.py
```diff
@@ -167,3 +167,31 @@ def __call__(self, form, from_field, to_field):
...
+        if form.minute.data and int(form.minute.data) > 59:
+            raise validators.ValidationError(self.messages['INVALID_MINUTES'])
+        if form.hour.data and int(form.hour.data) > 999:
+            raise validators.ValidationError(self.messages['INVALID_HOURS'])
+
+
+class TimeInputRequired(object):
```
iwootten, line 186: I wonder if this and the DateRequired should be consolidated into a single "FormResponseRequired" validator or similar, which checks that some data was entered at a form level (DateForm or TimeForm) in this case. Otherwise these type of validators will be added for each new form data structure that is added to the system.

app/validation/validators.py
```diff
@@ -167,3 +167,31 @@ def __call__(self, form, from_field, to_field):
...
+
+class TimeInputCheck(object):
+    def __init__(self, messages=None):
+        if not messages:
+            messages = error_messages
+        self.messages = messages
+
+    def __call__(self, form, field):
+        logger.error(self.messages)
+        if form.minute.data and int(form.minute.data) > 59:
```
iwootten, line 180: This won't catch the case if the value is negative.

app/templates/partials/answers/Minute.html
```diff
@@ -0,0 +1,3 @@
+{% with input_type = 'mins' %}
+{% include 'partials/forms/input-type.html' %}
```
iwootten, line 2: 'input-type' seems like quite a vague name to give this partial given we already have one named 'input'. Isn't it really an integer-field?

app/jinja_filters.py
```diff
@@ -71,3 +71,11 @@ def format_household_summary(names):
 
         return person_list
     return ''
+
+
+@blueprint.app_template_filter()
+def format_time_input(value):
+    time_input = value.split(":")
+    hours = time_input[0] + " hours " if time_input[0] else ''
+    minutes = time_input[1] + " minutes " if time_input[1] else ''
```
iwootten, line 80: The mockup says "mins", should it be formatted the same on the summary?

iwootten (changes requested):

PR #970, opened 2017-02-22
Make session timeout configurable via the schema

ajmaddaford:
The session timeout was previously made configurable by using session storage, but this won't work when using phantomjs for testing. This change also allows the timeout to be configured per survey if required.

As part of this pull request the grace period has been changed so that it is added to the server side session rather than subtracted from the timeout period sent to the client. This is so that when the timeout is defined in the schema, that is the timeout the client experiences before being timed out. 

### How to review 
- Check out and use the `test_timeout.json` survey. It should show the prompt after 5 seconds and timeout after 10 seconds.
- Run the `timeout.spec.js` test


app/__init__.py
```diff
@@ -202,7 +202,7 @@ def add_blueprints(application):
 def setup_secure_cookies(application):
     application.secret_key = settings.EQ_SECRET_KEY
     application.permanent_session_lifetime = timedelta(
-        seconds=settings.EQ_SESSION_TIMEOUT_SECONDS)
+        seconds=settings.EQ_SESSION_TIMEOUT_SECONDS + settings.EQ_SESSION_TIMEOUT_GRACE_PERIOD_SECONDS)
```
jonnyshaw89, line 205: Im not sure how the server side timeout gets set in the schema?

jonnyshaw89 (changes requested):

app/__init__.py
benjefferies: https://github.com/ONSdigital/eq-survey-runner/pull/970/files#diff-9b6edc9d34e32d43abe13a513b21e492R536

jonnyshaw89: That sets the time that gets passed to the front end but doesn't modify the server-side timeout of the session

benjefferies: Good point, I wonder if this can be set on the session object

ajmaddaford: Good spot! It doesn't, and we can't set this via the schema property as this is set on application start for all surveys. The result of this is that the client timeout can be set lower than the server and the client would appear timed out, but the session wouldn't be removed until the server timeout is reached. At the moment this is only used in the test timeout schema, and we should review the implications of this before using it in an actual survey.

At the moment a survey can specify a client timeout greater than the server timeout, which would result in the prompt not being displayed and the user getting a timed out page on the first request after the server timeout. I'll change this so that we always set the client timeout to the same as the server timeout if it's greater than the server timeout.

collisdigital: ... Making sure that the client always times-out with a buffer (30 seconds?) before the server to account for delays etc. i.e. the grace period.

jonnyshaw89 (approved):

hamishtaplin (approved):

PR #971, opened 2017-02-22
Wrap with quotes to stop splitting

benjefferies:
Generating schemas will fail if there are spaces in the path.

Depends on https://github.com/ONSdigital/eq-translations/pull/8

### How to review 
1. Clone repository into a directory with spaces in `git clone https://github.com/ONSdigital/eq-survey-runner.git eq\ survey\ runner`
1. Run  `./scripts/translate_schemas.sh`


scripts/run_app.sh
```diff
@@ -24,7 +24,7 @@ fi
 
 # Output the current git revision
 if [ -z "$EQ_GIT_REF" ]; then
```
jonnyshaw89, line 20: do we still need any of this? it was used to determine if the app had changed but no is only displayed on the status page?

jonnyshaw89 (changes requested):

scripts/run_app.sh
collisdigital: There was also discussion about it being somewhere in the log output at startup for Splunk, is it logged anywhere?

benjefferies: Yeah it's logged out on start up

scripts/run_app.sh
```diff
@@ -40,12 +40,12 @@ fi
 echo "Environment variables in use:"
 env | grep EQ_
 
-$DIR/build.sh
+"${DIR}"/build.sh
 
-url="`python token_generator.py`"
+url=$(python token_generator.py)
```
jonnyshaw89: Can we get rid of this yet? no one likes it :)

benjefferies: I'm happy to get rid of it. Everyone goes to the dev page now surely?

jonnyshaw89: Yeh, probably just launch the dev page

benjefferies: Done

jonnyshaw89 (approved):

ajmaddaford (approved):

PR #975, opened 2017-02-22
Added support for New Relic

jonnyshaw89:
This adds support for New Relic.

### How to review 
When the NEW_RELIC_LICENSE_KEY is supplied then data should be shown in New Relic

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -256,3 +256,5 @@ WTForms==2.1 \
     --hash=sha256:ffdf10bd1fa565b8233380cb77a304cd36fd55c73023e91d4b803c96bc11d46f
 Flask_WTF==0.14.2 \
     --hash=sha256:d9a9e366b32dcbb98ef17228e76be15702cd2600675668bca23f63a7947fd5ac
+newrelic==2.78.0.57 \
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -256,3 +256,5 @@ WTForms==2.1 \
     --hash=sha256:ffdf10bd1fa565b8233380cb77a304cd36fd55c73023e91d4b803c96bc11d46f
 Flask_WTF==0.14.2 \
     --hash=sha256:d9a9e366b32dcbb98ef17228e76be15702cd2600675668bca23f63a7947fd5ac
+newrelic==2.78.0.57 \
```
collisdigital, line 259: should this only be installed for test? i.e. move to the other reqs file?

jonnyshaw89: I don't think it works like that. the other file is only used for unit / integration tests?
This in theory isn't a test dependency as we want it in the code all the time.

ajmaddaford (approved):

requirements.txt
collisdigital: I guess there's no simple way of dynamically installing dependencies based on your build options...

jonnyshaw89: I guess then the argument would be that the tested build isn't the build you run in production?

LJBabbage (approved):

PR #976, opened 2017-02-22
Added resource to error logs

jonnyshaw89:
When errors are logged they don't contain enough context to be useful.
This PR adds the resource that was being requested when the error occurred.
We were seeing many 404 errors but the logs did not show what resource was being requested.

### How to review 
Hit an invalid url and observe that the resource you requested is in the logged error

### DIFF
--- a/app/views/errors.py
+++ b/app/views/errors.py
@@ -30,46 +30,46 @@ def add_cache_control(response):
 @errors_blueprint.app_errorhandler(401)
 @errors_blueprint.app_errorhandler(NoTokenException)
 def unauthorized(error=None):
-    log_exception(error)
+    log_exception(error, 401)
     return render_theme_template('default', 'session-expired.html'), 401
 
 
 @errors_blueprint.app_errorhandler(InvalidTokenException)
 def forbidden(error=None):
-    log_exception(error)
+    log_exception(error, 403)
     return _render_error_page(403)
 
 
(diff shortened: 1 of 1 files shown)


benjefferies (approved):

benjefferies (changes requested):
Should probably add to the else statement too?

app/views/errors.py
```diff
@@ -30,46 +30,46 @@ def add_cache_control(response):
...
 def service_unavailable(error=None):
-    log_exception(error)
+    log_exception(error, 503)
     return _render_error_page(503)
 
 
 @errors_blueprint.app_errorhandler(MultipleSurveyError)
 def multiple_survey_error(error=None):
-    log_exception(error)
+    log_exception(error, 200)
```
benjefferies, line 58: Seems weird to return 200 here, maybe a 409 would be more applicable

jonnyshaw89: This PR isn't really about changing the status that is returned. Its more adding disability of what is returned. @ajmaddaford what do you think about this?

ajmaddaford: Agreed, this pull request is to add the url and status code to the log output, if we want to change the behaviour we should do so in another pull request.

ajmaddaford (approved):

benjefferies (approved):

Issue #977, opened 2017-02-22
Google analytics events


bennpearson:
### Expected behaviour
Google analytics events should work as expected

### Actual behaviour
Google analytics events behaving erratically

### Steps to reproduce the behaviour
Log into google analytics and check events happen when using 'help and support', 'previous' and 'guidance' links. Also check errors are still working

collisdigital:
Discussed with Benn; this is fixed as of #978

PR #978, opened 2017-02-22
google analytics issue fix

bennpearson:
Fixs #977 Google analytics events were behaving erratically so this fixes the issue.

### How to review 
- in terminal `set -x EQ_UA_ID UA-56892037-7`
- Select schema `census_household.json`
- navigate to question that doesn't have a label and click on `guidance`, `help and support`, `previous`.
- cause an error on the page
- check google analytics


app/templates/layouts/_base.html
```diff
@@ -33,7 +33,18 @@
     <![endif]-->
     {% endblock styles -%}
     {% endblock head -%}
-    {{ analytics }}
+    <!-- Google Analytics -->
+    <script>
```
collisdigital, line 37: I wonder if we should put this into a separate template and include it here to keep it separate?

collisdigital (changes requested):
You also need to remove everything under `app/analytics/` and the setup_analytics() function in `app/__init__.py` and calls to it.

app/assets/js/app/modules/analytics.js
```diff
@@ -8,7 +8,9 @@ export default function initAnalytics() {
   }
 
   if (typeof window.ga !== 'undefined') {
-    trackEvent = window.ga
+    trackEvent = function(evt, data) {
```
hamishtaplin: If we're gonna change this then:

```javascript
trackEvent = (evt, data) => window.ga(evt, data)
```

is more concise

hamishtaplin (changes requested):

hamishtaplin (approved):

collisdigital (commented):
Looks good to me, needs the team to review it

collisdigital (approved):

PR #979, opened 2017-02-23
Implement HTML5 elements

hamishtaplin:

[Trello card](https://trello.com/c/8P8PTEqR/846-spike-investigate-use-of-html5-tags)

#### Outcome of spike

Following the [lead of gov.uk](https://gds.blog.gov.uk/2012/02/10/using-html5-for-gov-uk/):

- html5 shim loaded for browsers <= ie8 (this was already implemented)
- no special attention paid to legacy browsers without JS

This PR adds some semantic HTML5 tags where appropriate.

### How to review 

- Review Census and UKIS surveys in a couple of browsers (including ie8) to check no negative effects of introducing html5 tags (very unlikely as they don't have any CSS by default)
- Review semantics of application making any suggestions for improvement

### DIFF
--- a/app/templates/layouts/_base.html
+++ b/app/templates/layouts/_base.html
@@ -13,7 +13,7 @@
     <meta content="width=device-width, initial-scale=1" name="viewport">
     <link rel="shortcut icon" href="{{ url_for('static', filename='img/favicon.ico') }}">
     <!--[if lt IE 9]>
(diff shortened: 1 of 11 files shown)


bennpearson (approved):

collisdigital (commented):
Demo'd this to Emma and she's fine with IE8 (no JS) issues this causes.

PR #980, opened 2017-02-23
improved reliability of timeout test

hamishtaplin:

Increased timeout period


benjefferies (approved):
You were right @ajmaddaford

PR #983, opened 2017-02-23
Reduce complexity of adding new relic

jonnyshaw89:
Fixing the build by reducing the complexity of the create_app method

### How to review 
Does the build pass?

### DIFF
--- a/README.md
+++ b/README.md
@@ -250,6 +250,7 @@ EQ_MAX_HTTP_POST_CONTENT_LENGTH - The maximum http post content length that the 
 EQ_MAX_NUM_REPEATS - The maximum number of repeats the system will allow
 EQ_DEVELOPER_LOGGING - Enable developer style logging described here http://structlog.readthedocs.io/en/stable/development.html
 
+EQ_NEW_RELIC_ENABLED - Enable New Relic monitoring
 NEW_RELIC_LICENSE_KEY - Enable new relic monitoring by supplying a New Relic licence key
 NEW_RELIC_APP_NAME - The name to display for the application in New Relic
 ```
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -17,6 +17,7 @@ from app.authentication.authenticator import login_manager
 from app.authentication.cookie_session import SHA256SecureCookieSessionInterface
 from app.data_model.database import db_session
 from app.libs.utils import get_locale
+from app.new_relic import setup_newrelic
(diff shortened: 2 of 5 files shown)


ajmaddaford (approved):

benjefferies (approved):
I think create_app can be tidied up a lot more and we should look at doing it properly at some point but I understand this is a blocking issue for now

README.md
```diff
@@ -250,6 +250,7 @@ EQ_MAX_HTTP_POST_CONTENT_LENGTH - The maximum http post content length that the
 EQ_MAX_NUM_REPEATS - The maximum number of repeats the system will allow
 EQ_DEVELOPER_LOGGING - Enable developer style logging described here http://structlog.readthedocs.io/en/stable/development.html
 
+EQ_NEW_RELIC_ENABLED - Enable New Relic monitoring
 NEW_RELIC_LICENSE_KEY - Enable new relic monitoring by supplying a New Relic licence key
 NEW_RELIC_APP_NAME - The name to display for the application in New Relic
```
benjefferies, line 255: Is this something that needs to be configurable?

tests/integration/test_application_variables.py
```diff
@@ -1,3 +1,5 @@
+from mock import patch
```
benjefferies: Unused import?

benjefferies (changes requested):

README.md
jonnyshaw89: What do you mean?

benjefferies: We could remove NEW_RELIC_APP_NAME  and do

```
def setup_newrelic():
    if settings.EQ_NEW_RELIC_ENABLED:
        import newrelic.agent
        newrelic.agent.initialize()
        newrelic.agent.register_application(name='eq-survey-runner')
```

jonnyshaw89: We could but what if we wanted to log both pre-prod and prod to the same New Relic account?

benjefferies: I guess that's what NEW_RELIC_ENVIRONMENT is for according to https://docs.newrelic.com/docs/agents/python-agent/customization-extension/python-management-api#initialize

jonnyshaw89: They are much the same but this just means we don't need to add the ini file

benjefferies: Fair enough if they are. I didn't know if new relic had a view where you could see all of the same apps in their different environments or something.

benjefferies (approved):

PR #990, opened 2017-02-27
Limit repeats for minus_one rule

benjefferies:
Repeating relationship question should have a maximum number of 24
repeats if limit is 25 and relationships needed to be answers reduces by
one per page.

Relationship group uses answer_count_minus_one repeat rule. It is
consistent for minus_one rule to apply to the maximum repeats.

### How to review 
1. Load census_household schema
1. Add 26 people to household composition on who-lives-here/0/household-composition
1. Go through relationship questions until one relationship left to answer (who-lives-here-relationship/23/household-relationships)
1. Click next

Expected
Next screen is `You have successfully completed the ‘Who lives here?’ section`

Fixes https://github.com/ONSdigital/eq-survey-runner/issues/883

### DIFF
--- a/app/questionnaire/rules.py
+++ b/app/questionnaire/rules.py
@@ -52,9 +52,9 @@ def evaluate_repeat(repeat_rule, answer_store):
     :return: The number of times to repeat
     """
     repeat_functions = {
-        'answer_value': _get_answer_value(),
+        'answer_value': _get_answer_value,
         'answer_count': len,
-        'answer_count_minus_one': _get_answer_count_minus_one(),
+        'answer_count_minus_one': _get_answer_count_or_repeat_limit_minus_one,
     }
     if 'answer_id' in repeat_rule and 'type' in repeat_rule:
         repeat_index = repeat_rule['answer_id']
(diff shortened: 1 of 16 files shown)


app/questionnaire/rules.py
```diff
@@ -69,13 +69,14 @@ def evaluate_repeat(repeat_rule, answer_store):
...
-def _get_answer_value():
-    return lambda filtered_answers: int(
-        filtered_answers[0]['value'] if len(filtered_answers) == 1 and filtered_answers[0]['value'] else 0)
+def _get_answer_value(filtered_answers):
+    return int(filtered_answers[0]['value'] if len(filtered_answers) == 1 and filtered_answers[0]['value'] else 0)
 
 
-def _get_answer_count_minus_one():
-    return lambda filtered_answers: len(filtered_answers) - 1 if len(filtered_answers) > 0 else 0
+def _get_answer_count_minus_one(filtered_answers):
```
LJBabbage: Probably needs a name change now, as it won't always be answer_count_minus_one

tests/integration/household_composition/test_repeating_relationship.py
```diff
@@ -0,0 +1,37 @@
...
+    def setUp(self):
+        super(TestRepeatingRelationship, self).setUp()
+
+        self.token = create_token('household', 'census')
+        resp = self.client.get('/session?token=' + self.token.decode(), follow_redirects=False)
+        is_home = {'permanent-or-family-home-answer': 'Yes'}
+        resp = self.client.post(resp.location, data=is_home, follow_redirects=False)
+        self.household_composition_location = resp.location
+
+    def test_should_ask_twenty_four_relationships_when_twenty_five_max_limit_and_twenty_six_people_added(self):
```
LJBabbage, line 16: Perhaps remove the numbers here and capture it in the when and then, otherwise it is a real mouth full and hard to follow. something like test_relationships_count_when_max_people_reached

LJBabbage (changes requested):

LJBabbage:
Tested with an adjusted setting 
EQ_MAX_NUM_REPEATS = os.getenv('EQ_MAX_NUM_REPEATS', 3)
But works fine at this level

tests/integration/household_composition/test_repeating_relationship.py
benjefferies: I understand the concern. The only thing is this is a bit more expressive in that you know what the test is suppose to do when the max is reached. I am happy to change if everyone thinks it's OTT

app/questionnaire/rules.py
benjefferies: Done

jonnyshaw89 (approved):

LJBabbage (approved):

PR #992, opened 2017-02-27
Add question and answers types for Time input

LJBabbage:
Add New question type for time input

### How to review 
Using test_time_input.json check that it work as expected. Minutes can't be >59 and hours can not be >999, both should be numbers

N.B this is a reopening of a PR, the commit history has been changed, so you will need to delete your local branch if you have already opened this branch before


app/assets/styles/partials/components/_question.scss
```diff
@@ -103,3 +103,16 @@
     }
   }
 }
+
+.question--timeinput {
+  // .question__answer {
```
iwootten: This commented code should really be removed.

app/validation/validators.py
```diff
@@ -167,3 +150,47 @@ def __call__(self, form, from_field, to_field):
...
+    field_flags = ('required', )
+
+    def __init__(self, message=None):
+        if not message:
+            message = error_messages['MANDATORY']
+        self.message = message
+
+    def __call__(self, form, field):
+
+        if hasattr(form, 'day'):
```
iwootten, line 165: I think this would be more concise if it iterated over all the fields on the current form and checked each had data. The default iterator for a form should allow this.

    form_has_data = False

    for field in form:
        form_has_data |= bool(field.data)

    if not form_has_data:
            

That would mean we could just apply to any form without caring about it's structure.

app/validation/validators.py
```diff
@@ -167,3 +150,47 @@ def __call__(self, form, from_field, to_field):
...
+                raise validators.StopValidation(self.message)
+        elif hasattr(form, 'month'):
+            if not form.month.data and not form.year.data:
+                raise validators.StopValidation(self.message)
+        elif hasattr(form, 'hours') and hasattr(form, 'mins'):
+            if not form.hours.data and not form.mins.data:
+                raise validators.StopValidation(self.message)
+
+
+class TimeInputCheck(object):
```
iwootten, line 176: Can the Integer/NumberRange validators within this TimeInputCheck validator not just be applied to the form fields upon TimeInputForm, removing the need for this validator?

app/views/questionnaire.py
```diff
@@ -444,6 +431,21 @@ def update_questionnaire_store_with_form_data(questionnaire_store, location, ans
         questionnaire_store.completed_blocks.append(location)
 
 
+def _format_answer_values(answer_value):
+
+    is_day_month_year = 'day' in answer_value and 'month' in answer_value and 'year' in answer_value
+    is_month_year = 'day' not in answer_value and 'year' in answer_value and 'month' in answer_value
+    is_time_input = 'hours' in answer_value or 'mins' in answer_value
```
iwootten, line 355: Should this be 'and'? The TimeInputForm is only ever created with both.

app/validation/validators.py
```diff
@@ -66,7 +66,7 @@ def __init__(self, min=None, max=None, messages=None):
         self.messages = messages
 
     def __call__(self, form, field):
-        data = field.data
+        data = int(field.data)
```
iwootten, line 69: I don't think there should be a need to cast this as NumberRanges should only be applied to IntegerFields.

app/validation/validators.py
```diff
@@ -167,3 +150,47 @@ def __call__(self, form, from_field, to_field):
...
+        self.message = message
+        self.message['NEGATIVE_INTEGER'] = self.message['INTEGER_TOO_LARGE'] = self.message['INVALID_TIME_INPUT']
+
+    def __call__(self, form, field):
+
+        integer_check = IntegerCheck(self.message['INVALID_TIME_INPUT'])
+
+        if form.hours.data:
+            integer_check(form, form.hours)
+            hours_number_range = NumberRange(0, 999, messages=self.message)
```
iwootten: NumberRange takes a dictionary of errors as it's argument and it appears this is only sending a single error.

iwootten (dismissed):

app/validation/validators.py
LJBabbage: I ended up have to use StringFields in the time_input_form. Mainly because of Mandatory means only one field is needed out of the 2 (hours  or minutes). If I used IntergerField and I only give it minutes, it throws an error because of the None. I looked at your CustomIntegerField but that cleans the data if its not a valid integer to none, which wouldn't work for me later. So I used StringField and therefore the cast

iwootten: I think that is because the current TimeInputCheck isn't using the validation chain and is instead calling validators manually, which means the validation chain isn't stopped after failing a cast to integer as part of the IntegerCheck. If the validators are moved to the form, this cast shouldn't be necessary.

app/validation/validators.py
collisdigital: or use the any() built-in to do the what Ian suggests, but will exit as soon as any field value evaluates to True without evaluating the whole list (I don't think the cast to bool is needed this way as well):

```python
if not any(field.data for field in form):
```

app/templates/partials/answers/hours.html
```diff
@@ -0,0 +1,3 @@
+{% with input_type = 'hours' %}
```
iwootten: Is there a reason why there are hours.html, mins.html and timeinput.html. Don't we just need the timeinput.html?

iwootten:
I don't think there's an easy solution to the validation here. It's required at both the form and field level. The way we have done form-level validation elsewhere is to override validate on the form and handle the creation of appropriate errors manually.

I've pushed up a branch overriding validate on the timeinput form here: [eq-783-time-input-pattern-2](https://github.com/ONSdigital/eq-survey-runner/tree/eq-783-time-input-pattern-2). It's far from complete, but goes some way to demonstrate how it could be approached. questionnaire_form.py also uses this approach for date range validation.

benjefferies:
This one looks to be failing the build.

app/data/test_time_input.json
```diff
@@ -0,0 +1,77 @@
...
+                            "id": "time-input-section",
+                            "questions": [
+                                {
+                                    "answers": [
+                                        {
+                                            "guidance": "",
+                                            "id": "time-input-mandatory-answer",
+                                            "label": "Time Input Mandatory",
+                                            "mandatory": true,
+                                            "options": [],
```
benjefferies, line 29: Options is mandatory so this could be removed

app/validation/validators.py
benjefferies: Ians point makes sense with the rename and would reduce complexity

app/views/questionnaire.py
```diff
@@ -361,6 +348,21 @@ def update_questionnaire_store_with_form_data(questionnaire_store, location, ans
...
+
+    if is_day_month_year and answer_value['day'] and answer_value['month']:
+        return "{:02d}/{:02d}/{}".format(int(answer_value['day']), int(answer_value['month']), answer_value['year'])
```
benjefferies, line 358: Bit late in the day but why aren't we using ISO 8601 dates? I wonder if we should use ISO 8601 duration?

app/validation/validators.py
benjefferies: I agree with this point. I don't see what TimeInputCheck is giving us over the other validators

app/views/questionnaire.py
```diff
@@ -328,21 +328,8 @@ def update_questionnaire_store_with_form_data(questionnaire_store, location, ans
...
-                    date_str = "{:02d}/{:02d}/{}".format(
-                        int(answer_value['day']),
-                        int(answer_value['month']),
-                        answer_value['year'],
-                    )
-                    answer = Answer(answer_id=answer_id, value=date_str, location=location)
-                elif is_month_year and answer_value['month']:
-                    date_str = "{:02d}/{}".format(int(answer_value['month']), answer_value['year'])
-                    answer = Answer(answer_id=answer_id, value=date_str, location=location)
+                answer = Answer(answer_id=answer_id, value=_format_answer_values(answer_value), location=location)
```
benjefferies, line 332: Nice refactor

tests/app/forms/test_fields.py
```diff
@@ -291,3 +291,23 @@ def test_percentage_field(self):
...
+            "type": "TimeInput",
+            "validation": {
+                "messages": {
+                     "INVALID_TIME_INPUT": "The time entered is not valid. Please correct your answer"
+                }
+            }
+        }
+
+        unbound_field = get_field(time_input_json, time_input_json['label'], error_messages)
+        self.assertTrue(unbound_field.field_class == FormField)
```
benjefferies, line 311: assertEqual would give better debug here and I wonder if we can check the form_class type here?

tests/app/forms/test_time_input.py
```diff
@@ -0,0 +1,39 @@
+import unittest
+
+from app import create_app
+from app.forms.time_input_form import get_time_input_form
+from app.schema_loader.schema_loader import load_schema_file
+from app.helpers.schema_helper import SchemaHelper
+
+
+class TestTimeInputForm(unittest.TestCase):
+    def setUp(self):
```
benjefferies, line 10: setUp isn't needed

tests/app/validation/test_time_input_validator.py
```diff
@@ -0,0 +1,60 @@
...
+        validator = TimeInputCheck()
+
+        mock_form = Mock()
+        mock_form.hours.raw_data = [1000]
+        mock_form.hours.data = 1000
+        mock_form.mins.data = None
+
+        mock_field = Mock()
+
+        with self.assertRaises(ValidationError) as ite:
```
benjefferies, line 20: do we need `as ite` it might come up as unused variable by pylint

tests/app/validation/test_time_input_validator.py
```diff
@@ -0,0 +1,60 @@
...
+            validator(mock_form, mock_field)
+
+    def test_time_input_validator_invalid_mins_range(self):
```
benjefferies, line 23: Do we want to test the valid scenarios?

benjefferies (changes requested):
Looks pretty good generally. Nice refactoring

Issue #991, opened 2017-02-27
Timeout not saving the answers 


yoganandkunche:
### Expected behaviour
When a survey times out all the answers entered till that point must be saved. 

### Actual behaviour
No answers are saved when the session times out 

### Steps to reproduce the behaviour
Launch any survey in pre-prod
answer 2 questions and on the 3rd question allow the session to time out 
launch the same survey with the same JWT details 

#### Browser
Chrome

collisdigital:
Fixed by #993

PR #993, opened 2017-02-27
Session timeout should save answers

benjefferies:
Session timeout should save answers when user clicks save and sign out. 
If a user times out before clicking save and sign out all answers but the latest page they had timed out on should be saved

### How to review 
Verify timeout.spec.js tests the scenario correctly

### DIFF
--- a/app/views/questionnaire.py
+++ b/app/views/questionnaire.py
@@ -192,6 +192,7 @@ def get_sign_out(eq_id, form_type, collection_id):  # pylint: disable=unused-arg
                                                 analytics_ua_id=settings.EQ_UA_ID,
                                                 survey_id=g.schema_json['survey_id'])
     session_storage.clear()
+    logout_user()
     return signed_out_template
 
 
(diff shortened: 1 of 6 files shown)


app/views/questionnaire.py
```diff
@@ -292,7 +292,6 @@ def get_timeout_continue(eq_id, form_type, collection_id):  # pylint: disable=un
 @questionnaire_blueprint.route('session-expired', methods=["GET"])
 @login_required
 def get_session_expired(eq_id, form_type, collection_id):  # pylint: disable=unused-argument
-    _delete_user_data()
```
jonnyshaw89, line 207: you should replace this with logout_user() to kill the flask session

jonnyshaw89 (changes requested):

app/views/questionnaire.py
benjefferies: Good spot, I'll also add an IT for it

app/views/questionnaire.py
```diff
@@ -292,6 +292,7 @@ def get_timeout_continue(eq_id, form_type, collection_id):  # pylint: disable=un
 @questionnaire_blueprint.route('session-expired', methods=["GET"])
 @login_required
 def get_session_expired(eq_id, form_type, collection_id):  # pylint: disable=unused-argument
+    session_storage.clear()
```
jonnyshaw89: i still think we need to call logout_user if only just to stop a login attempt when they reload?

LJBabbage (approved):

LJBabbage:
Stick a "fixes" in the PR text so it closes the issue when its merged

jonnyshaw89 (approved):

collisdigital:
Would be good if this was in for the demo tomorrow!

benjefferies:
Tell travis that @collisdigital. Fix one intermittent timeout test another one fails.

benjefferies:
I lied. It was broken from the rebase. test_timeout.json now needs a summary block.

collisdigital:
👍

collisdigital:
Fixes #991

PR #994, opened 2017-02-27
Reduce Template Logic

iwootten:
Removes some of the nasty template logic surrounding integers/other options and moves it to backend at a form level.

### How to review 
See if the existing tests continue to pass..

### DIFF
--- a/app/forms/questionnaire_form.py
+++ b/app/forms/questionnaire_form.py
@@ -18,6 +18,7 @@ class QuestionnaireForm(FlaskForm):
     def __init__(self, block_json, formdata=None, **kwargs):
         self.block_json = block_json
         self.question_errors = {}
+        self.options_with_children = {}
 
         if formdata:
             super(QuestionnaireForm, self).__init__(formdata=formdata, **kwargs)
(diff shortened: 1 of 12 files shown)


app/forms/questionnaire_form.py
```diff
@@ -70,6 +71,26 @@ def map_errors(self):
     def answer_errors(self, input_id):
         return [error[1] for error in self.map_errors() if input_id == error[0]]
 
+    def get_data(self, answer_id):
+        attr = getattr(self, answer_id)
+        return attr.raw_data[0] if attr.raw_data else ''
+
+    def option_has_other(self, answer_id, option_index):
+        if not self.options_with_children:
+            self.options_with_children = SchemaHelper.get_parent_options_for_block(self.block_json)
```
benjefferies, line 80: Do we need to get all the options for just this answer_id?

benjefferies (changes requested):
Looks better

app/forms/questionnaire_form.py
iwootten: Nope, but we will as part of the request for the block, which this call will cache.

jonnyshaw89 (approved):

benjefferies (approved):

PR #996, opened 2017-02-28
Pylint Fixes

iwootten:
Resolve various minor pylint warnings, in order to reduce superfluous output.

### How to review 
See if tests continue to pass.

### DIFF
--- a/app/forms/fields.py
+++ b/app/forms/fields.py
@@ -124,7 +124,7 @@ class CustomIntegerField(IntegerField):
     further validation steps by using a separate IntegerCheck validator
     """
     def __init__(self, **kwargs):
-
+        self.data = None
         super(CustomIntegerField, self).__init__(**kwargs)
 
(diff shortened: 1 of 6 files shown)


app/forms/fields.py
```diff
@@ -124,15 +124,15 @@ class CustomIntegerField(IntegerField):
...
+        self.data = None
         super(CustomIntegerField, self).__init__(**kwargs)
 
     def process_formdata(self, valuelist):
         if valuelist:
             try:
                 self.data = int(valuelist[0])
             except ValueError:
-                self.data = None
+                pass
```
collisdigital, line 135: Is it possible that self.data could contain its previous value if it isn't reset?

iwootten: No, it's always a new instance on every field.

benjefferies (approved):

jonnyshaw89 (approved):

PR #998, opened 2017-02-28
Dont count before getting database resource

ajmaddaford:
Removes doing a count query before a select for database resources, as we can determine from the return of the  select if the resource exists. In all instances if the count returned a positive number the select was done anyway. 

### How to review 
- Run the application and verify it works as expected (logging in, saving data, retreiving previous answers)
- Run the unit tests and functional tests and verify they all pass


benjefferies (approved):

jonnyshaw89 (approved):

jonnyshaw89:
@ajmaddaford rebase?

PR #999, opened 2017-02-28
Dont revalidate every page before submission of survey data

jonnyshaw89:
Removed the re-validation of pages before submission. Instead it checks that every page in the routing path is complete. Once a page is complete it cannot be made invalid as we don't save invalid data.

### How to review 
Validate that data is submitted to queue when submitting the last page in the survey (summary or final_confirmation)

### DIFF
--- a/app/views/questionnaire.py
+++ b/app/views/questionnaire.py
@@ -219,9 +219,9 @@ def get_session_expired(eq_id, form_type, collection_id):  # pylint: disable=unu
 def submit_answers(eq_id, form_type, collection_id):
     metadata = get_metadata(current_user)
     answer_store = get_answer_store(current_user)
-    is_valid, invalid_location = validate_all_answers(answer_store, metadata)
+    is_completed, invalid_location = is_survey_completed(answer_store, metadata)
 
-    if is_valid:
+    if is_completed:
         path_finder = PathFinder(g.schema_json, answer_store, metadata)
         submitter = SubmitterFactory.get_submitter()
         message = convert_answers(metadata, g.schema_json, answer_store, path_finder.get_routing_path())
@@ -240,17 +240,16 @@ def post_everyone_at_address_confirmation(eq_id, form_type, collection_id, group
     return post_block(eq_id, form_type, collection_id, group_id, group_instance, 'permanent-or-family-home')
 
 
-def validate_all_answers(answer_store, metadata):
+def is_survey_completed(answer_store, metadata):
     path_finder = PathFinder(g.schema_json, answer_store, metadata)
-    error_messages = SchemaHelper.get_messages(g.schema_json)
+    completed_blocks = get_completed_blocks(current_user)
 
     for location in path_finder.get_routing_path():
-        block_json = _render_schema(location)
-        form, _ = get_form_for_location(block_json, location, answer_store, error_messages)
+        if location.block_id in ['thank-you', 'summary', 'confirmation']:
+            continue
 
-        if not form.validate():
-            logger.debug("Failed validation", location=str(location))
-            return False, location
+        if location not in completed_blocks:
+                return False, location
 
     return True, None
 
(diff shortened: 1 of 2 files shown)


app/views/questionnaire.py
```diff
@@ -161,6 +165,10 @@ def post_household_composition(eq_id, form_type, collection_id, group_id):
         metadata = get_metadata(current_user)
         path_finder = PathFinder(g.schema_json, get_answer_store(current_user), metadata)
         next_location = path_finder.get_next_location(current_location=current_location)
+
```
benjefferies: Probably don't need this here do we or is it just to satisfy tests?

benjefferies (changes requested):

benjefferies (approved):

app/views/questionnaire.py
```diff
@@ -121,6 +121,10 @@ def post_block(eq_id, form_type, collection_id, group_id, group_instance, block_
     else:
         _update_questionnaire_store(current_location, form)
         next_location = path_finder.get_next_location(current_location=current_location)
+
+        if next_location is None:
```
ajmaddaford: I think we should check explicitly for a user action e.g. `if 'action[submit_answers]' in request.form` rather than rely on next_location being None.

app/views/questionnaire.py
```diff
@@ -241,16 +241,13 @@ def post_everyone_at_address_confirmation(eq_id, form_type, collection_id, group
 
 
 def validate_all_answers(answer_store, metadata):
```
ajmaddaford: As this no longer validates all answers, I think we should rename this method something like `is_survey_completed`

ajmaddaford (changes requested):

app/views/questionnaire.py
jonnyshaw89: What if the last page in the survey is an additional interstitial and not the summary or final-confimation?

benjefferies: Can we not delete the action[submit_answers] thing and rely on this check. Seems like unnecessary information if it's the last block just submit?

jonnyshaw89: maybe path_finder.get_next_location could also return a flag to indicate that there is no next location as it has got to then end?

benjefferies: I think if None is returned that's a valid reason of saying there is no next location therefore you're at the end. That said we might want to do something in the case where the current_location isn't a valid location

jonnyshaw89: if the current location isn't valid. 404?

benjefferies: https://github.com/ONSdigital/eq-survey-runner/pull/999/files#diff-9b6edc9d34e32d43abe13a513b21e492R107

jonnyshaw89: ok so the next location can only be None if you are at the end?

benjefferies: In theory yes

samiwel: Late to the party, but I agree with Ben here. I think if `next_location is None` is clear enough. Implies that there is no next location, so we must be at the end.

jonnyshaw89: this conversation can be carried on here. https://github.com/ONSdigital/eq-survey-runner/pull/1003

ajmaddaford (approved):

app/views/questionnaire.py
```diff
@@ -240,17 +240,16 @@ def post_everyone_at_address_confirmation(eq_id, form_type, collection_id, group
...
-def validate_all_answers(answer_store, metadata):
+def is_survey_completed(answer_store, metadata):
     path_finder = PathFinder(g.schema_json, answer_store, metadata)
-    error_messages = SchemaHelper.get_messages(g.schema_json)
+    completed_blocks = get_completed_blocks(current_user)
 
     for location in path_finder.get_routing_path():
-        block_json = _render_schema(location)
-        form, _ = get_form_for_location(block_json, location, answer_store, error_messages)
+        if location.block_id in ['thank-you', 'summary', 'confirmation']:
```
collisdigital, line 248: Will this still work since we moved Intro and Summary etc into Blocks and gave them specific types. Suggest this should be checking the block type not id now?

collisdigital: Have we confirmed all current surveys use this id values for these block types?

PR #1003, opened 2017-03-01
Submit survey at end of routing path

jonnyshaw89:
When the end of the routing path is reached submit the survey.

### How to review 
Does the survey still submit data when expected

### DIFF
--- a/app/data/1_0001.json
+++ b/app/data/1_0001.json
@@ -4954,10 +4954,6 @@
                         "title": "You are now ready to submit this survey"
                     }],
                     "type": "Confirmation"
-                },
-                {
-                    "type": "Summary",
-                    "id": "summary"
                 }
             ],
             "hide_in_navigation": true,
--- a/app/templates/confirmation.html
+++ b/app/templates/confirmation.html
@@ -1,7 +1,5 @@
 {% extends 'questionnaire.html' %}
 
-{% set form_action = url_prefix + "/submit-answers" %}
-
 {% block page_title %}Confirmation{% endblock %}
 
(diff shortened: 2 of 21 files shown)


benjefferies (approved):

samiwel (approved):
LGTM but needs rebasing.

jonnyshaw89:
I think @ajmaddaford has some views on this so will hold it until he has reviewed.

benjefferies:
Rebase?

ajmaddaford:
In this approach I have concerns over implicitly submitting a users answers when we decide the end of the survey (routing path) is reached. I think this should only be done when the user has performed an action to submit their answers e.g. clicking a 'submit answers' button. After discussing with Jon we've come up with a hybrid solution - make the submit button on any page we determine to be at the end of the routing path a 'submit answers' button and look for this in the POST request.

ajmaddaford (changes requested):
Needs adapting to the solution we discussed.

ajmaddaford:
After further discussions with @benjefferies the new proposal is that we submit the survey when there is no next location **and** the current location block is of type `Summary` or `Confirmation`. This will give us confidence the user is at the end of the survey and allows us to make the default action on POST to be 'Save and continue' or 'Submit answers'. This also means that the `action[x]` buttons will only be needed for additional submit buttons on a page (e.g. adding/removing household members in the Census or 'Save and complete later').

ajmaddaford (approved):

jonnyshaw89:
@ajmaddaford didn't want me to squash?

PR #1005, opened 2017-03-01
Create QBS schema

LJBabbage:
Create QBS schema, according to https://docs.google.com/presentation/d/1kZYAfeMcH_H_iAjl3kgFC3Ijr03B5kQ10sULZYyIaow/edit#slide=id.g1ba6eee5f3_0_82. Also change ref_end_date to be non-mandatory 

### How to review 
Open up 2_0001.json and make sure the survey works as expected and is according to the slides. Note the Introduction page at this stage will not have the base for completion section working and ru_name is always used

N.B to test the single date you will need to remove ref_p_end_date from the dev page

### DIFF
--- a/app/data/2_0001.json
+++ b/app/data/2_0001.json
@@ -1,0 +1,110 @@
+{
+    "data_version": "0.0.1",
+    "description": "Quarterly Business Survey",
+    "groups": [
+        {
+            "blocks": [
+                {
+                    "type": "Introduction",
+                    "id": "introduction",
+                    "information_to_provide": [
+                      "Employee numbers"
+                    ]
+                },
+                {
+                    "id": "number-of-employees",
+                    "sections": [
+                        {
+                            "id": "number-of-employees-section",
+                            "questions": [
+                                {
+                                    "answers": [
+                                        {
+                                            "id": "number-of-employees-male-more-30-hours",
+                                            "label": "Number of male employees working more than 30 hours per week",
+                                            "mandatory": false,
+                                            "q_code": "51",
+                                            "type": "PositiveInteger"
+                                        },
+                                        {
+                                            "id": "number-of-employees-male-less-30-hours",
+                                            "label": "Number of male employees working 30 hours or less per week",
+                                            "mandatory": false,
+                                            "q_code": "52",
+                                            "type": "PositiveInteger"
+                                        },
+                                        {
+                                            "id": "number-of-employees-female-more-30-hours",
+                                            "label": "Number of female employees working more than 30 hours per week",
+                                            "mandatory": false,
+                                            "q_code": "53",
+                                            "type": "PositiveInteger"
+                                        },
+                                        {
+                                            "id": "number-of-employees-female-less-30-hours",
+                                            "label": "Number of female employees working 30 hours or less per week",
+                                            "mandatory": false,
+                                            "q_code": "54",
+                                            "type": "PositiveInteger"
+                                        },
+                                        {
+                                            "id": "number-of-employees-total",
+                                            "label": "Total number of employees",
+                                            "mandatory": true,
+                                            "q_code": "50",
+                                            "type": "PositiveInteger"
+                                        }
+                                    ],
+                                    "description": "An employee is anyone aged 16 years or over that your organisation directly pays from its payroll(s), in return for carrying out a full-time or part-time job or being on a training scheme.",
+                                    "guidance": [
+                                        {
+                                            "list": [
+                                                "all workers paid directly from this business’s payroll(s)",
+                                                "those temporarily absent but still being paid, for example on maternity leave"
+                                            ],
+                                            "title": "Include:"
+                                        },
+                                        {
+                                            "list": [
+                                                "agency workers paid directly from the agency payroll",
+                                                "voluntary workers",
+                                                "former employees only receiving a pension",
+                                                "self-employed workers",
+                                                "working owners who are not paid via PAYE"
+                                            ],
+                                            "title": "Exclude:"
+                                        }
+                                    ],
+                                    "id": "number-of-employees-question",
+                                    "title": "On {{exercise.start_date|format_date}} what was the number of employees for {{respondent.ru_name}}?",
+                                    "type": "General"
+                                }
+                            ],
+                            "title": "Quarterly Business Survey"
+                        }
+                    ],
+                    "type": "Questionnaire"
+                },
+                {
+                    "type": "Summary",
+                    "id": "summary"
+                }
+            ],
+            "id": "qbs",
+            "title": "Quarterly Business Survey"
+        }
+    ],
+    "legal_basis": "StatisticsOfTradeAct",
+    "messages": {
+		"MANDATORY": "Please provide a value, even if your value is 0.",
+		"NOT_INTEGER": "Please only enter whole numbers into the field.",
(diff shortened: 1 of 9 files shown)


app/data/2_0001.json
```diff
@@ -0,0 +1,109 @@
+{
```
benjefferies, line 1: What does the 2 signify in the filename if nothing would quarterly be more descriptive?

app/data/2_0001.json
```diff
@@ -0,0 +1,109 @@
...
+                                    ],
+                                    "id": "number-of-employees-question",
+                                    "title": "On {{exercise.start_date|format_date}}  what was the number of employees for {{respondent.ru_name}}?",
+                                    "type": "General"
+                                }
+                            ],
+                            "title": "Quarterly Business Survey"
+                        }
+                    ],
+                    "type": "Questionnaire"
```
benjefferies, line 86: Are we missing the description

An employee is anyone aged 16 years or over that your organisation directly pays from its payroll(s), in return for carrying out a full-time or part-time job or being on a training scheme.

tests/app/parser/test_metadata_parser.py
```diff
@@ -245,23 +245,6 @@ def test_is_valid_fails_invalid_return_by(self):
             parse_metadata(jwt)
         self.assertIn("incorrect data in token", ite.exception.value)
 
-    def test_is_valid_fails_missing_ref_p_end_date(self):
```
collisdigital, line 248: Keep this but make it test_is_valid_succeeds_missing_ref_p_end_date ?

collisdigital (approved):

app/templates/introduction.html
```diff
@@ -42,7 +42,7 @@ <h2 class="saturn col-10@l">{{legal_text}}</h2>
           <dl class="dl">
             <dt class="dl__title mercury">PERIOD</dt>
             <dd class="dl__data mars">
-              <span>{{ meta.survey.start_date|format_date }} to {{ meta.survey.end_date|format_date}}
+              <span>{{ meta.survey.start_date|format_date }}{% if meta.survey.end_date %} to {{ meta.survey.end_date|format_date}}{% endif %}
```
benjefferies: Seems like a jinja filter would be better for this. It could be tested and more logic removed from template

tests/app/parser/test_metadata_parser.py
LJBabbage: Done, GH has ignored it so the comment hasn't been minimised

tests/integration/introduction/test_introduction.py
```diff
@@ -39,3 +40,58 @@ def test_intro_description_not_displayed(self):
...
+
+        # When on the introduction page
+        content = resp.get_data(True)
+
+        # Then one date is shown and there is no 'to'
+        self.assertIn('1 May 2016', content)
+        self.assertNotIn('1 May 2016 to', content)
+
+    @staticmethod
+    def _payload_vars_without_ref_end_date():
```
benjefferies: A note going forward I have made create_token easier to be able to these kinds of things. However, it's not in master https://github.com/ONSdigital/eq-survey-runner/pull/1009/files#diff-d7db136e32b2be1e61f07882bcdf8031R25

tests/integration/introduction/test_introduction.py
```diff
@@ -39,3 +40,58 @@ def test_intro_description_not_displayed(self):
 
         # Then description should not be displayed
         self.assertNotIn('qa-intro-description', content)
+
+    def test_ref_start_and_end_date(self):
```
benjefferies: If we use a jinja filter to achieve the `$date to $date` logic we could unit test this logic instead of integration test

benjefferies (changes requested):

collisdigital:
Checked content and q_codes and I painstakingly tested all the routes. All good.

app/data/2_0001.json
LJBabbage: Its the eq id. For this one its 2. Don't know why

collisdigital: It is the id that RRM Classic provides; '2' is agreed because the form type (0001) clashes with UKIS

collisdigital: We would prefer better names but are tied to the existing system for now; we can review when we switch to RAS.

app/templates/introduction.html
LJBabbage: Original it was planned to do it in the template rather then a filter, will check with @ajmaddaford

app/data/2_0001.json
benjefferies: Fair enough if it's already agreed across systems. Let's push for some meaningful names going forward this means we don't need mappings and it's readable

app/templates/introduction.html
ajmaddaford: @benjefferies raises a good point below about it being unit testable as a jinja filter, I think we should do as he suggests. It's also worth looking at the `format_str_as_date_range` filter that is used to generate a from/to date range for the summary screen to see if it could be used.

tests/integration/introduction/test_introduction.py
```diff
@@ -1,5 +1,6 @@
 from tests.integration.integration_test_case import IntegrationTestCase
 from tests.integration.create_token import create_token
+from app.views.dev_mode import generate_token, create_payload
```
benjefferies: Has this import accidentally slipped in?

app/templates/introduction.html
```diff
@@ -42,7 +42,7 @@ <h2 class="saturn col-10@l">{{legal_text}}</h2>
           <dl class="dl">
             <dt class="dl__title mercury">PERIOD</dt>
             <dd class="dl__data mars">
-              <span>{{ meta.survey.start_date|format_date }} to {{ meta.survey.end_date|format_date}}
+              <span>{{ meta.survey.start_date|format_date(meta.survey.end_date)}}</span>
```
benjefferies: This looks really weird but works

benjefferies (changes requested):

app/templates/introduction.html
collisdigital: ugly

yoganandkunche:
Tested on branch looks good apart from the two known things...  piping of 'Trading as ' and basis of completion on landing page.

tests/integration/introduction/test_introduction.py
LJBabbage: It was needed for the first test, forgot to remove it, when I removed the test

benjefferies (approved):

benjefferies:
Rebase/squash?

app/templates/introduction.html
```diff
@@ -42,7 +42,7 @@ <h2 class="saturn col-10@l">{{legal_text}}</h2>
           <dl class="dl">
             <dt class="dl__title mercury">PERIOD</dt>
             <dd class="dl__data mars">
-              <span>{{ meta.survey.start_date|format_date }} to {{ meta.survey.end_date|format_date}}
+              <span>{{ format_start_end_date(meta.survey.start_date, meta.survey.end_date)}}</span>
```
collisdigital, line 45: image

ajmaddaford (approved):

benjefferies (approved):

PR #1006, opened 2017-03-01
Linting json on run_tests.sh

benjefferies:
Automatically format schema json

### How to review 


gulpfile.babel.js
```diff
@@ -167,5 +169,26 @@ gulp.task('test', [
...
+  gulp.src(['./app/data/*.json'])
+    .pipe(prettify())
+    .pipe(diff())
+    .pipe(diff.reporter({
+      quiet: false,
+      fail: true
+    }))
+})
+
+gulp.task('format', () => {
```
ajmaddaford: I think the task should better reflect what it does and be consistent with the lint task, `beautify:json` maybe?

ajmaddaford (changes requested):

gulpfile.babel.js
benjefferies: Yeah sure I agree but why beautify over format?

ajmaddaford: Just because it uses https://www.npmjs.com/package/js-beautify, I suppose if the only thing it is going to do is reformat the JSON, then format is fine.

benjefferies: Fair enough think it just formats `This little beautifier will reformat and reindent bookmarklets, ugly JavaScript, unpack scripts packed`. Think format reflects what it is doing making it more self explanatory.

ajmaddaford (approved):

benjefferies:
To be clear this doesn't format on build this just lints the json. If the json fails linting `yarn format` can be ran to format.

LJBabbage (approved):
I followed the "how to review" it to the letter, described it to a bloke in the kitchen and showed him some screens shots, he was happy so :+1: from me

PR #1008, opened 2017-03-02
Fixes navigation missing on confirmation page.

### LINKED ISSUE #1002: Navigation missing on the final submission screen
### Expected behaviour
Navigation links on the right must be displayed on the final submission screen

### Actual behaviour
Navigation links are missing on the final submission screen. (Please find attached)

Launch UKIS survey (on Master)
complete the survey and navigate to submission screen 

### Screenshot
screen shot 2017-02-28 at 15 55 49

samiwel:
Fixes #1002 navigation missing on the confirmation screen.
I also spotted and fixed another issue where if you set navigation to false in the schema, it still displays the navigation as part of this same PR.

### How to review 
Retest the defect.
On the confirmation page (i.e. the page just before the final submission), the navigation menu should be displayed.

### DIFF
--- a/app/questionnaire/navigation.py
+++ b/app/questionnaire/navigation.py
@@ -29,7 +29,7 @@ class Navigation(object):
         :param current_group_instance:
         :return:
         """
-        if 'navigation' not in self.survey_json:
+        if self.survey_json.get('navigation', False) is False:
             return None
 
         navigation = []
--- a/app/views/questionnaire.py
+++ b/app/views/questionnaire.py
@@ -445,7 +445,7 @@ def _get_front_end_navigation(answer_store, current_location, metadata):
     completed_blocks = get_completed_blocks(current_user)
     navigation = Navigation(g.schema_json, answer_store, metadata, completed_blocks)
     block_json = SchemaHelper.get_block_for_location(g.schema_json, current_location)
-    if block_json is not None and (block_json['type'] == 'Questionnaire' or block_json['type'] == 'Interstitial'):
+    if block_json is not None and block_json['type'] in ('Questionnaire', 'Interstitial', 'Confirmation'):
         return navigation.build_navigation(current_location.group_id, current_location.group_instance)
     else:
         return None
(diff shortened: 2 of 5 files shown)


app/questionnaire/navigation.py
```diff
@@ -29,7 +29,7 @@ def build_navigation(self, current_group_id, current_group_instance):
         :param current_group_instance:
         :return:
         """
-        if 'navigation' not in self.survey_json:
+        if 'navigation' not in self.survey_json or self.survey_json['navigation'] is False:
```
collisdigital: or:

```python
if self.survey_json.get('navigation', False) is False:
```

yoganandkunche:
Tested it on branch works as expected

ajmaddaford (approved):

app/views/questionnaire.py
```diff
@@ -445,7 +445,7 @@ def _get_front_end_navigation(answer_store, current_location, metadata):
     completed_blocks = get_completed_blocks(current_user)
     navigation = Navigation(g.schema_json, answer_store, metadata, completed_blocks)
     block_json = SchemaHelper.get_block_for_location(g.schema_json, current_location)
-    if block_json is not None and (block_json['type'] == 'Questionnaire' or block_json['type'] == 'Interstitial'):
+    if block_json is not None and block_json['type'] in ('Questionnaire', 'Interstitial', 'Confirmation'):
```
benjefferies, line 448: I don't think we need to check if the `block is not None` as would have been validated before to check the user is on a valid block/location

benjefferies (approved):
Looks good. Comment is suggestion and not a blocker

app/views/questionnaire.py
samiwel: I'll leave this as-is for now, if that's okay? It's how it was previously and I don't think there's any harm in leaving it in here since `block_json` is assigned in the line just before and `get_block_for_location` has the potential to return `None`.

benjefferies: No worries

PR #1009, opened 2017-03-03
Track jti claim to prevent replay attacks

benjefferies:
JWT rfc defines a token which can be used to prevent repeat attacks.
e.g. login using the same token twice by using the same JWT token before
the JWT token has expiried.
The token is defined here: [jti](https://tools.ietf.org/html/rfc7519#section-4.1.7)

If the jti token is present in the JWT payload we store the jti token in
a postgres table called used_token.
If the same token is seen more than once the authentication request is
rejected as invalid.
If the jti token is not present we do not protect against repeat
attacks. This puts the responsibility on the upstream system.

The tokens are stored in a postgres table and old tokens can be cleared
out by referencing the timestamp column. i.e. `where timestamp > '1
month old'`

##### Developers can walk through the integration test
test_login.test_login_with_token_twice_is_unauthorised_when_same_jti_provided
test_login.test_login_with_token_twice_is_authorised_when_no_jti

##### Testers can manually test using chrome
1. Open the dev page
1. Open chrome console (cmd + option + i)
1. Click on chrome console network tab
1. filter on `session`
1. Right click line and click `Open in new tab` to repeat log in
image

### DIFF
--- a/app/authentication/authenticator.py
+++ b/app/authentication/authenticator.py
@@ -9,7 +9,7 @@ from app.authentication.session_storage import session_storage
 from app.authentication.user import User
 from app.authentication.user_id_generator import UserIDGenerator
 from app.globals import get_questionnaire_store
-from app.parser.metadata_parser import is_valid_metadata, parse_metadata
+from app.parser.metadata_parser import is_valid_metadata
 
 logger = get_logger()
 
@@ -47,39 +47,24 @@ def load_user():
         return None
 
 
-def jwt_login(request):
+def store_session(metadata):
     """
-    Login using a JWT token, this must be an encrypted JWT.
-    :param request: The flask request
+    Store new session and metadata
+    :param metadata: metadata parsed from jwt token
     """
     # clear the session entry in the database
     session_storage.clear()
     # also clear the secure cookie data
     session.clear()
 
-    encrypted_token = request.args.get('token')
-
-    if encrypted_token is None:
-        raise NoTokenException("Please provide a token")
-
-    decrypted_token = jwt_decrypt(encrypted_token)
-
-    # once we've decrypted the token correct
-    # check we have the required user data
-    _check_user_data(decrypted_token)
-
     # get the hashed user id for eq
-    user_id = UserIDGenerator.generate_id(decrypted_token)
-    user_ik = UserIDGenerator.generate_ik(decrypted_token)
+    user_id = UserIDGenerator.generate_id(metadata)
+    user_ik = UserIDGenerator.generate_ik(metadata)
 
     # store the user id in the session
     session_storage.store_user_id(user_id)
     # store the user ik in the cookie
     session_storage.store_user_ik(user_ik)
-
-    # store the meta data
-    metadata = parse_metadata(decrypted_token)
-    logger.bind(tx_id=metadata["tx_id"])
 
     questionnaire_store = get_questionnaire_store(user_id, user_ik)
     questionnaire_store.metadata = metadata
@@ -87,13 +72,16 @@ def jwt_login(request):
     logger.info("user authenticated")
 
 
-def jwt_decrypt(encrypted_token):
+def decrypt_token(encrypted_token):
+    logger.debug("decrypting token")
+    if encrypted_token is None:
+        raise NoTokenException("Please provide a token")
     decoder = JWTDecryptor()
     decrypted_token = decoder.decrypt_jwt_token(encrypted_token)
-    return decrypted_token
 
-
-def _check_user_data(token):
-    valid, field = is_valid_metadata(token)
+    valid, field = is_valid_metadata(decrypted_token)
     if not valid:
         raise InvalidTokenException("Missing value {}".format(field))
+
+    logger.debug("token decrypted")
+    return decrypted_token
--- a/app/authentication/jti_claim_storage.py
+++ b/app/authentication/jti_claim_storage.py
@@ -1,0 +1,51 @@
+from uuid import UUID
+
+from sqlalchemy.exc import IntegrityError
+from structlog import get_logger
+
+from app.data_model.database import UsedJtiClaim, commit_or_rollback, db_session
+
+logger = get_logger()
+
+
+class JtiTokenUsed(Exception):
+
+    def __init__(self, jti_claim):
+        super(JtiTokenUsed, self).__init__()
+        self.jti_claim = jti_claim
+
+    def __str__(self, *args, **kwargs):
+        return "jti claim '{jti_claim}' has already been used".format(jti_claim=self.jti_claim)
+
+
+def _is_valid(jti_claim):
+    try:
+        UUID(jti_claim, version=4)
+    except ValueError:
+        return False
+    return True
+
+
+def use_jti_claim(jti_claim):
+    """
+    Use a jti claim
+    :param jti_claim: jti claim to mark as used.
+    :raises ValueError: when jti_claim is None.
+    :raises TypeError: when jti_claim is not a valid uuid4.
+    :raises JtiTokenUsed: when jti_claim has already been used.
+    """
+    if jti_claim is None:
+        raise ValueError
+    if not _is_valid(jti_claim):
+        logger.info("jti claim is invalid", jti_claim=jti_claim)
+        raise TypeError
+
+    try:
+        with commit_or_rollback(db_session):
+            jti = UsedJtiClaim(jti_claim)
+            # pylint: disable=maybe-no-member
+            # db_session has an add function but it is wrapped in a session_scope which confuses pylint
+            db_session.add(jti)
+    except IntegrityError as e:
+        logger.error("jti claim has already been used", jti_claim=jti_claim)
+        raise JtiTokenUsed(jti_claim) from e
--- a/app/data_model/database.py
+++ b/app/data_model/database.py
@@ -60,6 +60,19 @@ class EQSession(base):
         return "<EQSession('%s', '%s', '%s')>" % (self.eq_session_id, self.user_id, self.timestamp)
 
 
+class UsedJtiClaim(base):
+    __tablename__ = "used_jti_claim"
+    jti_claim = Column("jti_claim", String, primary_key=True)
+    used_at = Column("used_at", DateTime)
+
+    def __init__(self, jti_claim):
+        self.jti_claim = jti_claim
+        self.used_at = datetime.datetime.now()
+
+    def __repr__(self):
+        return "<UsedJtiClaim('%s', '%s')>" % (self.jti_claim, self.used_at)
+
+
 def _create_session_and_engine():
     logger.info("setting up database...")
     logger.debug("creating database engine")
--- a/app/views/session.py
+++ b/app/views/session.py
@@ -4,10 +4,12 @@ from flask import request
 from flask import session
 from flask_login import current_user
 from structlog import get_logger
-from werkzeug.exceptions import NotFound
+from werkzeug.exceptions import NotFound, Unauthorized
 
-from app.authentication import authenticator
-from app.globals import get_answer_store, get_completed_blocks, get_metadata
+from app.authentication.authenticator import store_session, decrypt_token
+from app.authentication.jti_claim_storage import use_jti_claim, JtiTokenUsed
+from app.globals import get_answer_store, get_completed_blocks
+from app.parser.metadata_parser import parse_metadata
 from app.questionnaire.path_finder import PathFinder
 from app.utilities.schema import get_schema
 
@@ -31,24 +33,31 @@ def login():
     :return: a 302 redirect to the next location for the user
     """
     logger.new()
-    # logging in again clears any session state
-    if session:
-        session.clear()
-
-    logger.debug("attempting token authentication")
-    authenticator.jwt_login(request)
-    logger.debug("token authenticated - linking to session")
-
-    metadata = get_metadata(current_user)
-
+    decrypted_token = decrypt_token(request.args.get('token'))
+    metadata = parse_metadata(decrypted_token)
     eq_id = metadata["eq_id"]
     form_type = metadata["form_type"]
-
-    logger.bind(eq_id=eq_id, form_type=form_type)
+    tx_id = metadata["tx_id"]
+    logger.bind(eq_id=eq_id, form_type=form_type, tx_id=tx_id)
 
     if not eq_id or not form_type:
         logger.error("missing eq id or form type in jwt")
         raise NotFound
+
+    # logging in again clears any session state
+    if session:
+        session.clear()
+
+    jti_claim = metadata.get('jti')
+    if jti_claim is None:
+        logger.debug('jti claim not provided')
+    else:
+        try:
+            use_jti_claim(jti_claim)
+        except JtiTokenUsed as e:
+            raise Unauthorized from e
+
+    store_session(metadata)
 
     json = get_schema(metadata)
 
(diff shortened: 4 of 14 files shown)


app/authentication/authenticator.py
```diff
@@ -87,13 +76,17 @@ def jwt_login(request):
...
+    if encrypted_token is None:
+        raise NoTokenException("Please provide a token")
     decoder = JWTDecryptor()
     decrypted_token = decoder.decrypt_jwt_token(encrypted_token)
-    return decrypted_token
-
-
-def _check_user_data(token):
-    valid, field = is_valid_metadata(token)
+    # once we've decrypted the token correct
```
collisdigital: sentence a bit broken here

app/authentication/used_token_storage.py
```diff
@@ -0,0 +1,21 @@
...
+    try:
+        with commit_or_rollback(db_session):
+            jti = UsedToken(token)
+            # pylint: disable=maybe-no-member
+            # db_session has a add function but it is wrapped in a session_scope which confuses pylint
+            db_session.add(jti)
+    except IntegrityError:
+        return False
+
+    return True
```
collisdigital: Might be safer to return False by default and only return True after line 17? Maybe I'm just being paranoid... see later comments about this too

app/data_model/database.py
```diff
@@ -60,6 +60,19 @@ def __repr__(self):
         return "<EQSession('%s', '%s', '%s')>" % (self.eq_session_id, self.user_id, self.timestamp)
 
 
+class UsedToken(base):
+    __tablename__ = "used_token"
+    token = Column("token", String, primary_key=True)
```
collisdigital: Maybe this should be called 'jwt_jti' to match the value that's being stored and for variable names in this class?

app/views/session.py
```diff
@@ -35,15 +36,17 @@ def login():
     if session:
         session.clear()
 
-    logger.debug("attempting token authentication")
-    authenticator.jwt_login(request)
-    logger.debug("token authenticated - linking to session")
+    token = decrypt_token(request.args.get('token'))
+    jti_token = token.get('jti')
+    if jti_token is not None and not use_token(jti_token):
```
collisdigital: Maybe allow use_token to take None and return False if so to simplify this and make use_token more resilient? Should we also check the content of the jti is correct/valid (the right size/length or format or similar?)

tests/integration/create_token.py
```diff
@@ -1,7 +1,11 @@
+from uuid import uuid4
+
+import time
+
 from app.views.dev_mode import generate_token, create_payload
 
 PAYLOAD = {
-    'user': "mci-integration-test",
+    'user_id': "mci-integration-test",
```
collisdigital: be consistent with quotes, should be single quotes everywhere really (wonder if there is a lint rule for this?)

app/data_model/database.py
```diff
@@ -60,6 +60,19 @@ def __repr__(self):
...
+    timestamp = Column("timestamp", DateTime)
```
collisdigital: what's this a timestamp of 'created/seen/used' ? consider naming to reflect usage.

app/views/session.py
benjefferies: Seems like the validation stuff is something the PyJWT should check for as it's in the JWT spec but not sure if it does. Not sure if us checking it's a valid uuid makes it anymore secure. If they chose to use some other globally unique uuid would that be less secure or bad (other than not implementing by the spec)

benjefferies: Will push None check into use_token

tests/integration/create_token.py
benjefferies: Yeah I looked into this before. There is either a flake8 plugin or pylint plugin. We should look into that for consistency but maybe on another ticket as it wasn't introduced here and we can fix quotes across the board

app/authentication/used_token_storage.py
benjefferies: Hmm not sure it makes it more safer but tests would! Oops!

app/data_model/database.py
benjefferies: Does it matter that it's a jwt_jti? I guess it's just a unique token at this point?

app/data_model/database.py
benjefferies: Good point will change to used

app/authentication/authenticator.py
benjefferies: Done

app/data_model/database.py
collisdigital: except we're adhering to our ONS schema spec, so it should be a jti claim not just any value.

tests/integration/create_token.py
collisdigital: agreed

app/views/session.py
```diff
@@ -35,15 +36,15 @@ def login():
     if session:
         session.clear()
 
-    logger.debug("attempting token authentication")
-    authenticator.jwt_login(request)
-    logger.debug("token authenticated - linking to session")
+    token = decrypt_token(request.args.get('token'))
+    if not use_token(token.get('jti')):
```
collisdigital: Continuing previous discussion here. The RFC doesn't specify a format for the 'jti' just that it is unique; we (ONS) have defined the actual contents of it to be a GUID so should validate that what we received is correct as anything else suggests the sender hasn't implemented the spec correctly which is either bad implementation (could be broken; may fail or cause incorrect results ~ false positives etc.) or an indication of a compromised system. As a minimum we should check the length of the jti matches that of a a guid as defined in the schema def.

app/authentication/authenticator.py
```diff
@@ -47,27 +47,16 @@ def load_user():
         return None
 
 
-def jwt_login(request):
+def store_session(decrypted_token):
     """
     Login using a JWT token, this must be an encrypted JWT.
```
collisdigital: This is now wrong (encrypted)

app/authentication/used_token_storage.py
```diff
@@ -0,0 +1,28 @@
+from sqlalchemy.exc import IntegrityError
+from structlog import get_logger
+
+from app.data_model.database import UsedToken, commit_or_rollback, db_session
+
+logger = get_logger()
+
+
+def use_token(token):
```
collisdigital: Is it the token or the jti that's being passed in here; better to be explicit.

collisdigital (commented):
not sure this is clear enough about what values are being used based on the names of things, when it comes to security explicit is better than implicit (Zen of Python too)

app/data_model/database.py
benjefferies: Will change, went with this approach after brief discussion with @ajmaddaford yesterday

app/views/session.py
benjefferies: I've validated it's a uuid4. Not sure of the reason it had to be a uuid in the spec though but to late for that.

app/authentication/used_token_storage.py
benjefferies: Done

app/authentication/authenticator.py
benjefferies: Done

app/views/session.py
collisdigital: Due to needing a globally unique identifier that isn't duplicated across multiple servers (negligible probability)

app/authentication/jti_claim_storage.py
```diff
@@ -0,0 +1,39 @@
...
+        return False
+    try:
+        with commit_or_rollback(db_session):
+            jti = UsedJtiClaim(jti_claim)
+            # pylint: disable=maybe-no-member
+            # db_session has a add function but it is wrapped in a session_scope which confuses pylint
+            db_session.add(jti)
+            return True
+    except IntegrityError:
+        logger.error("jti token has already been used")
```
collisdigital: nit picking now; `jti token` -- `jti claim`

collisdigital (commented):
Looking good

app/views/session.py
benjefferies: Yup understand that. The jwt docs specify that without specifying the implementation details.

app/authentication/jti_claim_storage.py
benjefferies: Done

app/views/session.py
collisdigital: In my experience not specifying these things explicitly normally results in 'best case' poor implementation; 'worst case' broken/wrong implementations.

tests/integration/create_token.py
```diff
@@ -13,21 +17,19 @@
...
-    payload_vars['employment_date'] = employment_date or payload_vars['employment_date']
     payload_vars['eq_id'] = eq_id
     payload_vars['form_type'] = form_type_id
-    payload_vars['region_code'] = region_code
-    payload_vars['language_code'] = language_code
-
-    payload = create_payload(**payload_vars)
+    payload_vars['jti'] = str(uuid4())
+    payload_vars['iat'] = time.time()
+    payload_vars['exp'] = time.time() + float(3600)  # one hour from now
```
samiwel: `payload_vars['exp'] = payload_vars['iat'] + float(3600)`?

benjefferies: Done

collisdigital (commented):
Tested this locally; works as described.

app/authentication/jti_claim_storage.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def use_jti_claim(jti_claim):
+    """
+    Use a jti claim if it is supplied
+    :param jti_claim: jti claim to mark as used. Must be a valid uuid 4.
+    :return: True if the token was used or wasn't supplied and False if the token is invalid or has already been used.
+    """
+    if jti_claim is None:
+        return True
```
ajmaddaford: I think the check for `None` should be checked before calling `use_jti_claim`, at the moment it's not clear 
that the JTI is optional and also is misleading to say `use_jti_claim` is true.

app/authentication/jti_claim_storage.py
```diff
@@ -0,0 +1,39 @@
...
+    if not _is_valid(jti_claim):
+        return False
```
ajmaddaford: We should log a message if the jti is invalid

app/authentication/jti_claim_storage.py
```diff
@@ -0,0 +1,39 @@
...
+    try:
+        with commit_or_rollback(db_session):
+            jti = UsedJtiClaim(jti_claim)
+            # pylint: disable=maybe-no-member
+            # db_session has a add function but it is wrapped in a session_scope which confuses pylint
```
ajmaddaford: minor typo 'a add' -> 'an add'

app/authentication/jti_claim_storage.py
```diff
@@ -0,0 +1,39 @@
...
+            db_session.add(jti)
+            return True
+    except IntegrityError:
+        logger.error("jti claim has already been used")
```
ajmaddaford: Is it worth logging the jti? or do we have the tx_id at this point?

app/data_model/database.py
```diff
@@ -60,6 +60,19 @@ def __repr__(self):
         return "<EQSession('%s', '%s', '%s')>" % (self.eq_session_id, self.user_id, self.timestamp)
 
 
+class UsedJtiClaim(base):
+    __tablename__ = "used_jti_claim"
+    jti_claim = Column("jti_claim", String, primary_key=True)
+    used = Column("used", DateTime)
```
ajmaddaford: I think `used_at` would make it clearer that this is a time

ajmaddaford (changes requested):

app/authentication/jti_claim_storage.py
benjefferies: Was how I initially implemented and I also think it's not as clear. Can change back if @collisdigital is happy with that

app/data_model/database.py
benjefferies: Good point done

app/authentication/jti_claim_storage.py
collisdigital: If you change that then this should probably raise a ValueError if jti_claim is None

app/authentication/jti_claim_storage.py
benjefferies: I've done both of these. Added jti_claim and refactored to bind the tx_id

app/authentication/jti_claim_storage.py
benjefferies: Done

app/authentication/jti_claim_storage.py
benjefferies: Done

app/authentication/jti_claim_storage.py
```diff
@@ -0,0 +1,50 @@
...
+    return True
+
+
+def use_jti_claim(jti_claim):
+    """
+    Use a jti claim
+    :param jti_claim: jti claim to mark as used.
+    :raises ValueError: when jti_claim is None.
+    :raises TypeError: when jti_claim is not a valid uuid4.
+    :raises IntegrityError: when jti_claim has already been used.
```
collisdigital: Doc is now wrong

app/authentication/jti_claim_storage.py
```diff
@@ -0,0 +1,50 @@
...
+    """
+    if jti_claim is None:
+        raise ValueError
+    elif not _is_valid(jti_claim):
```
collisdigital: No need to have this as elif now as the else can't execute if the ValueError is raised; might be easier to read as a series of `if` statements. Up to you.

app/views/session.py
```diff
@@ -31,25 +33,32 @@ def login():
...
+    decrypted_token = decrypt_token(request.args.get('token'))
+    metadata = parse_metadata(decrypted_token)
     eq_id = metadata["eq_id"]
     form_type = metadata["form_type"]
-
-    logger.bind(eq_id=eq_id, form_type=form_type)
+    tx_id = metadata["tx_id"]
+    logger.bind(eq_id=eq_id, form_type=form_type, tx_id=tx_id)
 
     if not eq_id or not form_type:
```
collisdigital, line 43: shouldn't this be before the binding above?

benjefferies: Could be but then need to break out the tx_id to bind before. Don't think it matters much so did it before

app/authentication/jti_claim_storage.py
benjefferies: Preference thing I guess. I if/elif to highlight only one of those scenarios will happen

app/authentication/jti_claim_storage.py
benjefferies: Nice spot

app/views/session.py
collisdigital: yep good point.

app/authentication/jti_claim_storage.py
collisdigital: yeah agree

ajmaddaford: I agree with Chris' first comment on this. I think it should be an `if` rather than an `elif`. As an `elif` when I read it, it makes me glance back to evaluate when it wouldn't get there.

ajmaddaford (approved):

app/authentication/jti_claim_storage.py
benjefferies: Fair enough will change

app/authentication/jti_claim_storage.py
benjefferies: Done

collisdigital (approved):

PR #1010, opened 2017-03-03
Enable trading as name to be piped into questions

benjefferies:
Create a way of dynamically populated trad_as or ru_name from the metadata. See [trello](https://trello.com/c/JLIbPCgI/975-enable-trading-as-name-to-be-piped-into-questions)

### How to review 
- Open qbs with trad_as and ru_name
Expect to see trad_as populated

- Open qbs with only ru_name
Expect to see ru_name populated

- Open mwss with trad_as and ru_name
Expect to see trad_as populated

- Open mwss with only ru_name
Expect to see ru_name populated

### DIFF
--- a/app/templating/schema_context.py
+++ b/app/templating/schema_context.py
@@ -54,4 +54,5 @@ def _build_respondent(metadata):
     return {
         "ru_name": metadata["ru_name"],
         "trad_as": metadata["trad_as"],
+        "trad_as_or_ru_name": metadata["trad_as"] or metadata["ru_name"],
     }
--- a/gulpfile.babel.js
+++ b/gulpfile.babel.js
@@ -181,6 +181,10 @@ gulp.task('lint:json', () => {
       quiet: false,
       fail: true
     }))
+    .on('error', (err) => {
+      gutil.log('Linting failed try running `yarn format`')
+      throw err
+    })
 })
 
 gulp.task('format:json', () => {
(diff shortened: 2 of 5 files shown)


app/templating/schema_context.py
```diff
@@ -54,4 +54,5 @@ def _build_respondent(metadata):
     return {
         "ru_name": metadata["ru_name"],
         "trad_as": metadata["trad_as"],
+        "display_name": metadata["trad_as"] or metadata["ru_name"],
```
LJBabbage: I reckon we use trad_as_or_ru_name, just makes it clear

LJBabbage (changes requested):

app/templating/schema_context.py
benjefferies: Done

LJBabbage (approved):

gulpfile.babel.js
```diff
@@ -181,6 +181,10 @@ gulp.task('lint:json', () => {
       quiet: false,
       fail: true
     }))
+    .on('error', (err) => {
```
jonnyshaw89, line 184: Shouldn't this be a separate PR?
I don't really care but its a comment you often raise :)

benjefferies: Yes but didn't want to create a ticket for it 😆. Do you want me to?

jonnyshaw89: No. I was just being annoying.

jonnyshaw89 (approved):

PR #1011, opened 2017-03-06
Added csrf token to all form pages

jonnyshaw89:
Have enabled the CSRF token on all WTForms pages.

### How to review 
Check that all forms still work correctly.
Modify the 'csrf_token' in the hidden field in the HTML and check that the posted data isn't saved.

### DIFF
--- a/app/views/questionnaire.py
+++ b/app/views/questionnaire.py
@@ -271,7 +271,12 @@ def _save_sign_out(collection_id, eq_id, form_type, this_location, form):
 
 def _household_answers_changed(answer_store):
     household_answers = answer_store.filter(block_id="household-composition")
-    if len(household_answers) != len(request.form)-1:
+    stripped_form = request.form.copy()
+    del stripped_form['csrf_token']
+    remove = [k for k in stripped_form if 'action[' in k]
+    for k in remove:
+        del stripped_form[k]
+    if len(household_answers) != len(stripped_form):
         return True
     for answer in request.form:
         answer_id, answer_index = extract_answer_id_and_instance(answer)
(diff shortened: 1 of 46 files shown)


app/views/questionnaire.py
```diff
@@ -271,7 +271,7 @@ def _save_sign_out(collection_id, eq_id, form_type, this_location, form):
 
 def _household_answers_changed(answer_store):
     household_answers = answer_store.filter(block_id="household-composition")
-    if len(household_answers) != len(request.form)-1:
+    if len(household_answers) != len(request.form)-2:
```
benjefferies: Not a fan of comments but neither I am of random numbers. It might be worth adding a comment to make it clear why we -2 as it wasn't obvious initially.

tests/app/forms/test_household_relationship_form.py
```diff
@@ -74,53 +74,61 @@ def test_build_relationship_choices(self):
...
+            }, error_messages=error_messages)
 
-        message = "Not a valid choice"
+            form.validate()
+            mapped_errors = form.map_errors()
 
-        self.assertTrue(self._error_exists(answer['id'], message, mapped_errors))
+            message = "Not a valid choice"
+
+            print(mapped_errors)
```
benjefferies: print can be removed

benjefferies (changes requested):

app/views/questionnaire.py
jonnyshaw89: Yeh i might make this more explicit by copying the dict and removing the expected answers so the lengths will be equal?

benjefferies: Yeah that would make it a lot more obvious

tests/app/forms/test_household_relationship_form.py
jonnyshaw89: Yes, since i found where the Welsh was coming from.

benjefferies (approved):

tests/integration/integration_test_case.py
```diff
@@ -44,3 +45,20 @@ def postRedirectGet(self, url, post_data):
         resp = self.client.get(resp_url, follow_redirects=False)
         self.assertEqual(resp.status_code, 200)
         return resp_url, resp
+
+    def extract_csrf_token(self, html):
+        m = re.search('<input id="csrf_token" name="csrf_token" type="hidden" value="(.+?)">', html)
```
LJBabbage: Just m?

LJBabbage (changes requested):

tests/integration/integration_test_case.py
jonnyshaw89: mmmmmmmmm

LJBabbage (approved):

PR #1012, opened 2017-03-06
Fix session_storage flow issues and add unit test coverage

collisdigital:
An issue was identified with the flow of the code in the `session_storage.py` file where code was meant to be inside an `if` but was actually outside.

As part of this it became clear the whole file was lacking test coverage; so I have added additional coverage and used Flask app context to give access to the actual flask.session as the mocking/patching was getting pretty involved.

Finally; in writing the unit tests it became clear the `store_used_id` was nit behaving as expected. It looks up a session in the database and ignores the user_id that was being passed in, or assuming that if an existing session matched it was for the same user_id value (but not validating that). The desired behaviour is that any new login replaces an older one and the new user_id is taken; so I have updated `store_user_id` to reflect this (it now always overwrite the session with the new user_id).

### How to review 
- Review the code changes and code coverage

### DIFF
--- a/app/authentication/session_storage.py
+++ b/app/authentication/session_storage.py
@@ -15,24 +15,16 @@ logger = get_logger()
 
 class SessionStorage:
 
-    def store_user_id(self, user_id):
+    @staticmethod
+    def store_user_id(user_id):
         """
-        Store a user's id for retrieval later
-        :param user_id: the user id
+        Create a new eq_session_id and associate it with the user_id specified
         """
-        if EQ_SESSION_ID not in session:
-            eq_session_id = str(uuid4())
-            logger.debug("creating new eq session id", session_id=eq_session_id)
-            session[EQ_SESSION_ID] = eq_session_id
-            eq_session = EQSession(eq_session_id, user_id)
-        else:
-            eq_session_id = session[EQ_SESSION_ID]
-            logger.debug("got session id from session", session_id=eq_session_id)
-            eq_session = self._get_user_session(eq_session_id)
-            if eq_session is not None:
-                logger.debug("got session from database", eq_session_id=eq_session.eq_session_id, user_id=eq_session.user_id,
-                             timestamp=eq_session.timestamp.isoformat())
+        eq_session_id = str(uuid4())
+        session[EQ_SESSION_ID] = eq_session_id
+        eq_session = EQSession(eq_session_id, user_id)
 
+        logger.debug("Adding eq_session to database", eq_session_id=eq_session_id, user_id=user_id)
         with commit_or_rollback(db_session):
             # pylint: disable=maybe-no-member
             # session has a add function but it is wrapped in a session_scope which confuses pylint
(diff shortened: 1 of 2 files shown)


app/authentication/session_storage.py
```diff
@@ -15,24 +15,16 @@
 
 class SessionStorage:
 
-    def store_user_id(self, user_id):
+    @staticmethod
+    def store_user_id(user_id):
         """
-        Store a user's id for retrieval later
-        :param user_id: the user id
+        Create a new eq_session_id and associate it with the user_id specified
```
benjefferies, line 21: Much better, much simpler

tests/app/authentication/test_database_session_manager.py
```diff
@@ -1,49 +1,124 @@
...
+            self.assertFalse(EQ_SESSION_ID in flask.session)
+
+            # Store a user id for the first time
+            self.session_storage.store_user_id('1')
+
+            # When we store the user_id we generate a GUID and use that
+            # as the session_id mapped to the user_id in the database
+            self.assertTrue(EQ_SESSION_ID in flask.session)
+            self.assertIsNotNone(flask.session[EQ_SESSION_ID])
+            self.assertTrue(self.db_session.add.called)
```
benjefferies: Is this assertion necessary with the call count check below?

tests/app/authentication/test_database_session_manager.py
```diff
@@ -1,49 +1,124 @@
...
+        self.addCleanup(self.db_session_patcher.stop)
+        self.db_session = self.db_session_patcher.start()
+
+        # Note each test will need to create a test_request_context
+        # in order to access the Flask session object
+
+    def test_store_new_user_id_then_replace(self):
+        with self.flask_app.test_request_context('/status'):
+            # Initially there is no session id present
+            self.assertFalse(EQ_SESSION_ID in flask.session)
```
benjefferies: Aren't we testing the test is doing the right thing in set up here?

tests/app/authentication/test_database_session_manager.py
```diff
@@ -1,49 +1,124 @@
...
+            eq_session_id = flask.session[EQ_SESSION_ID]
+
+            # Try replacing the session with a new one
+            self.session_storage.store_user_id('2')
+
+            # Session GUID should have changed
+            self.assertEqual(self.db_session.add.call_count, 2)
+            self.assertNotEqual(eq_session_id, flask.session[EQ_SESSION_ID])
+
+    def test_clear(self):
```
benjefferies: This test seems over complicated and should probably be 3 tests. Each test should fail for it's own reason. This makes it easier to diagnose and fix the problem.
- `test_should_not_clear_with_no_session_data`
- `test_should_not_clear_with_no_session_data`
- `test_should_clear_with_session_and_data`

benjefferies (changes requested):

tests/app/authentication/test_database_session_manager.py
collisdigital: I wasn't sure what the default state of flask.session is on start-up and you can't access it until you have a request context; this could be removed now that we've verified the default behaviour if we're confident it won't ever change?

tests/app/authentication/test_database_session_manager.py
collisdigital: will remove

tests/app/authentication/test_database_session_manager.py
collisdigital: Will do

tests/app/authentication/test_database_session_manager.py
collisdigital: Decide to remove it

app/authentication/session_storage.py
collisdigital: Thanks

collisdigital:
@benjefferies comments addressed

benjefferies (approved):

tests/app/authentication/test_database_session_manager.py
```diff
@@ -1,49 +1,122 @@
...
+            # Store the user_id to lookup later
+            self.session_storage.store_user_id(user_id)
 
-    @staticmethod
-    def test_clear_rollback():
-        # Given
-        session_storage = SessionStorage()
+            # Mocking database session lookup, replace internal function on SessionStorage
+            eq_session = EQSession('123', user_id)
+            self.session_storage._get_user_session = lambda eq_session_id: eq_session
```
ajmaddaford, line 113: Any reason for having `eq_session_id` here rather than `_` as in the earlier test?

ajmaddaford (changes requested):

tests/app/authentication/test_database_session_manager.py
collisdigital: Made them all eq_session_id

ajmaddaford (approved):

PR #1015, opened 2017-03-07
Resolve Issue #1013 None radio values don't work as expected

### LINKED ISSUE #1013: UKIS: None options are getting preselected.
### Expected behaviour
'None' options must not get pre-selected.

### Actual behaviour
All the none options are getting pre-selected

Launch an UKIS survey.
Navigate to Question 9.9 
### Screenshot
screen shot 2017-03-06 at 18 23 13

collisdigital:
Fixes #1013

WTForms defaults to use a coerce function on Radio (Selector) fields of `str()`; this will convert Python `NoneType` to `'None'` string which means WTForms cannot tell the difference between the user not choosing any option `data=None` and choosing an option that happens to have a value of `value='None'` in HTML (i.e. `data='None'`).

This PR changes the coerce function for radio fields to be aware of `NoneType` in Python to avoid converting it to the string `'None'` to avoid this clash. This fixes #1013 which is caused by a Survey that contains a Radio field with a `'None'` option.

I have updated the Selenium tests and test_radio.json to exercise a number of scenarios around this (choosing None, changing the answer, checking the summary screen etc.)

I also added a None option to test_checkbox.json - but it doesn't appear affected by the same issue (presumably because it processes lists `[]` not just a single value).

### How to review 
- Run Selenium tests (or rely on Travis pass)
- Run some manual testing with radio fields 
- Check issue #1013 is resolved.

### DIFF
--- a/app/forms/fields.py
+++ b/app/forms/fields.py
@@ -104,14 +104,31 @@ def get_select_multiple_field(answer, label, guidance, error_messages):
     )
 
 
+def _coerce_str_unless_none(value):
+    """
+    Coerces a value using str() unless that value is None
+    :param value: Any value that can be coerced using str() or None
+    :return: str(value) or None if value is None
+    """
+    return str(value) if value is not None else None
+
+
 def get_select_field(answer, label, guidance, error_messages):
     validate_with = get_mandatory_validator(answer, error_messages)
+
+    # We use a custom coerce function to avoid a defect where Python NoneType
+    # is coerced to the string 'None' which clashes with legitimate Radio field
+    # values of 'None'; i.e. there is no way to differentiate between the user
+    # not providing an answer and them selecting the 'None' option otherwise.
+    # https://github.com/ONSdigital/eq-survey-runner/issues/1013
+    # See related WTForms PR: https://github.com/wtforms/wtforms/pull/288
 
     return SelectField(
         label=label,
         description=guidance,
         choices=build_choices(answer['options']),
         validators=validate_with,
+        coerce=_coerce_str_unless_none,
     )
 
 
--- a/tests/app/forms/test_fields.py
+++ b/tests/app/forms/test_fields.py
@@ -2,7 +2,7 @@ import unittest
 
 from wtforms import validators, StringField, TextAreaField, FormField, SelectField, SelectMultipleField
 
-from app.forms.fields import CustomIntegerField, get_field, get_mandatory_validator
+from app.forms.fields import CustomIntegerField, get_field, get_mandatory_validator, _coerce_str_unless_none
 from app.validation.error_messages import error_messages
 from app.validation.validators import ResponseRequired
 
@@ -153,9 +153,19 @@ class TestFields(unittest.TestCase):
         expected_choices = [(option['label'], option['value']) for option in radio_json['options']]
 
         self.assertTrue(unbound_field.field_class == SelectField)
+        self.assertTrue(unbound_field.kwargs['coerce'], _coerce_str_unless_none)
         self.assertEquals(unbound_field.kwargs['label'], radio_json['label'])
         self.assertEquals(unbound_field.kwargs['description'], radio_json['guidance'])
         self.assertEquals(unbound_field.kwargs['choices'], expected_choices)
+
+    def test__coerce_str_unless_none(self):
+        # pylint: disable=protected-access
+        self.assertEquals(_coerce_str_unless_none(1), '1')
+        self.assertEquals(_coerce_str_unless_none('bob'), 'bob')
+        self.assertEquals(_coerce_str_unless_none(12323245), '12323245')
+        self.assertEquals(_coerce_str_unless_none('9887766'), '9887766')
+        self.assertEquals(_coerce_str_unless_none('None'), 'None')
+        self.assertEquals(_coerce_str_unless_none(None), None)
 
     def test_checkbox_field(self):
         checkbox_json = {
--- a/tests/app/forms/test_questionnaire_form.py
+++ b/tests/app/forms/test_questionnaire_form.py
@@ -106,6 +106,69 @@ class TestQuestionnaireForm(unittest.TestCase):
         self.assertEqual(form.data, expected_form_data)
         self.assertEqual(form.question_errors['reporting-period-question'], expected_message)
 
+    def test_form_radio_options_no_selection(self):
+        # Given I have not selected a radio option, When I load a form,
+        # Then no radio options should be selected by default
+        survey = load_schema_file("test_radio.json")
+        block_json = SchemaHelper.get_block(survey, "radio-mandatory")
+        error_messages = SchemaHelper.get_messages(survey)
+
+        # Given
+        data = {}
+
+        # When
+        form = generate_form(block_json, data, error_messages)
+
+        # Then
+        for option in form['radio-mandatory-answer']:
+            self.assertFalse(option.checked)
+
+    def test_form_radio_options_select_none(self):
+        # Given I have selected the None radio option, When I load a form,
+        # Then the None radio option should be selected And no other options should be selected.
+        survey = load_schema_file("test_radio.json")
+        block_json = SchemaHelper.get_block(survey, "radio-mandatory")
+        error_messages = SchemaHelper.get_messages(survey)
+
+        # Given
+        data = {
+            'radio-mandatory-answer': 'None'
+        }
+
+        # When
+        form = generate_form(block_json, data, error_messages)
+
+        # Then
+        options = [option for option in form['radio-mandatory-answer']]
+
+        self.assertTrue(options[0].checked)  # None
+
+        for option in options[1:]:
+            self.assertFalse(option.checked)  # All others
+
+    def test_form_radio_options_select_none(self):
+        # Given I have selected the Bacon radio option, When I load a form,
+        # Then the Bacon radio option should be selected And no other options should be selected.
+        survey = load_schema_file("test_radio.json")
+        block_json = SchemaHelper.get_block(survey, "radio-mandatory")
+        error_messages = SchemaHelper.get_messages(survey)
+
+        # Given
+        data = {
+            'radio-mandatory-answer': 'Bacon'
+        }
+
+        # When
+        form = generate_form(block_json, data, error_messages)
+
+        # Then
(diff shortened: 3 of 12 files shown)


app/forms/fields.py
```diff
@@ -104,14 +104,30 @@ def get_select_multiple_field(answer, label, guidance, error_messages):
...
+    :param value: Any value that can be coerced using str() or None
+    :return: str(value) or None if value is None
+    '''
+    return str(value) if value is not None else None
+
+
 def get_select_field(answer, label, guidance, error_messages):
     validate_with = get_mandatory_validator(answer, error_messages)
 
+    # We use a custom coerce function to avoid a defect where Python NoneType
```
benjefferies, line 119: Looks like this is a stab at fixing https://github.com/wtforms/wtforms/pull/288. Might be worth referencing it incase it gets merged and this can be undone

tests/app/forms/test_fields.py
```diff
@@ -153,10 +153,20 @@ def test_radio_field(self):
...
         self.assertEquals(unbound_field.kwargs['choices'], expected_choices)
 
+    def test__coerce_str_unless_none(self):
+        # pylint: disable=protected-access
+        self.assertEquals(_coerce_str_unless_none(1), '1')
+        self.assertEquals(_coerce_str_unless_none('bob'), 'bob')
+        self.assertEquals(_coerce_str_unless_none(12323245), '12323245')
+        self.assertEquals(_coerce_str_unless_none('9887766'), '9887766')
+        self.assertEquals(_coerce_str_unless_none(None), None)
+        self.assertNotEquals(_coerce_str_unless_none(None), 'None')
```
benjefferies: This assertion is redundant from the assertion above

tests/app/forms/test_fields.py
```diff
@@ -153,10 +153,20 @@ def test_radio_field(self):
...
+        self.assertTrue(unbound_field.kwargs['coerce'], _coerce_str_unless_none)
         self.assertEquals(unbound_field.kwargs['label'], radio_json['label'])
         self.assertEquals(unbound_field.kwargs['description'], radio_json['guidance'])
         self.assertEquals(unbound_field.kwargs['choices'], expected_choices)
 
+    def test__coerce_str_unless_none(self):
+        # pylint: disable=protected-access
+        self.assertEquals(_coerce_str_unless_none(1), '1')
+        self.assertEquals(_coerce_str_unless_none('bob'), 'bob')
+        self.assertEquals(_coerce_str_unless_none(12323245), '12323245')
```
benjefferies, line 165: Is there value in having this assertion as wel as `self.assertEquals(_coerce_str_unless_none(1), '1')`

tests/functional/generate_pages.py
```diff
@@ -29,7 +29,7 @@
 
 HEADER = r"""// >>> WARNING THIS PAGE WAS AUTO-GENERATED - DO NOT EDIT!!! <<<
 
-import {basePage} from '../../{basePageFile}'
+import {basePage} from '../{basePageFile}'
```
benjefferies, line 32: Good effort updating this

tests/functional/generate_pages.py
```diff
@@ -153,7 +160,7 @@ def process_answer(question_type, answer, page_spec):
         logger.debug("\t\tSkipping Child Answer: %s", answer['id'])
         return
     elif answer['type'] == 'Radio' or answer['type'] == 'Checkbox':
-        process_options(answer['id'], answer['options'], CHECKBOX_RADIO_CLICKER, page_spec)
+        process_options(answer['id'], answer['options'], CHECKBOX_RADIO_CLICKER, CHECKBOX_RADIO_IS_SELECTED, page_spec)
```
benjefferies, line 165: generate_camel_case_from_id(answer_id) ?

tests/functional/spec/radio.spec.js
```diff
@@ -53,22 +53,94 @@ describe('Radio button with "other" option', function() {
...
     RadioMandatoryPage.clickRadioMandatoryAnswerBacon().submit()
-    RadioNonMandatory.clickTopPrevious()
+    RadioNonMandatoryPage.clickTopPrevious()
 
     //Then
     RadioMandatoryPage.clickOther()
     expect(RadioMandatoryPage.getOtherInputField()).to.equal('')
   })
 
+  it('Given I have previously selected the None option and saved, When I return to the screen, Then my choice is remembered.', function() {
```
benjefferies, line 74: I wonder if it's easy enough to do this as a unit test with wtforms

tests/functional/spec/radio.spec.js
```diff
@@ -53,22 +53,94 @@ describe('Radio button with "other" option', function() {
...
+
+    // When I return to the screen
+    RadioNonMandatoryPage.clickTopPrevious()
+
+    // Then my choice is remembered (Eggs)
+    expect(RadioMandatoryPage.RadioMandatoryAnswerNoneIsSelected()).to.be.false
+    expect(RadioMandatoryPage.RadioMandatoryAnswerBaconIsSelected()).to.be.false
+    expect(RadioMandatoryPage.RadioMandatoryAnswerEggsIsSelected()).to.be.true
+    expect(RadioMandatoryPage.RadioMandatoryAnswerSausageIsSelected()).to.be.false
+    expect(RadioMandatoryPage.RadioMandatoryAnswerOtherIsSelected()).to.be.false
```
benjefferies, line 115: Does all of the checks below this add anything?

benjefferies (changes requested):

tests/functional/spec/radio.spec.js
collisdigital: Prior to this fix it wasn't possible to change your answer from any option to the None option; so this verifies that you can change to None and then change to something else; corner cases but were real issues

app/forms/fields.py
collisdigital: Good idea

tests/app/forms/test_fields.py
collisdigital: you never know; do multi-digit fields get treated differently?

tests/functional/spec/radio.spec.js
benjefferies: Fair enough. Does it warrant being a selenium test when the bug is in wtforms?

tests/functional/generate_pages.py
collisdigital: Looks like answer_id is used on line 139 directly as well but I'll make some improvements to this as there is some duplication

tests/functional/spec/radio.spec.js
collisdigital: Possibly; I didn't try to be fair

app/forms/fields.py
collisdigital: Will do

collisdigital:
@benjefferies update force pushed (trying to keep this ready to merge)

app/forms/fields.py
collisdigital: Done

tests/app/forms/test_fields.py
```diff
@@ -153,10 +153,21 @@ def test_radio_field(self):
...
+        self.assertEquals(_coerce_str_unless_none('9887766'), '9887766')
+        self.assertEquals(_coerce_str_unless_none('None'), 'None')
```
collisdigital, line 167: @ajmaddaford added 'None' check

tests/functional/generate_pages.py
collisdigital: Although we should probably sort out the generation of the SummaryPage now it's a block; I updated to stop the script exploding on them but it produces an empty Summary for now.

tests/functional/generate_pages.py
collisdigital: Pulled answer_name up to top of this function and used here also set option_name once in the process_options() instead of working out the name multiple times.

tests/functional/spec/radio.spec.js
collisdigital: We could do with a set of common scenarios for all question types; such as checking default value on first page load; submitting and returning to screen to see value persisted, changing the value, removing an answer etc. At the moment coverage of these typical use cases is patchy. If WTForms does change this issue could get re-introduced so perhaps safer to test our usage of it until we have sufficient use cases covered generically?

tests/app/forms/test_fields.py
```diff
@@ -153,10 +153,21 @@ def test_radio_field(self):
...
+        self.assertEquals(_coerce_str_unless_none(None), None)
+        self.assertNotEquals(_coerce_str_unless_none(None), 'None')
```
LJBabbage: Do we really need both? implicitly if something equals one thing, it won't equal another. We really don't need to test a string isn't equal to None do we?

self.assertEquals(_coerce_str_unless_none(None), None)
self.assertNotEquals(_coerce_str_unless_none(None), 'None')

LJBabbage (changes requested):

tests/app/forms/test_fields.py
benjefferies: +1. We're testing pythons assertion library more than our code here

LJBabbage (approved):

collisdigital:
@benjefferies force pushed a WTForms form test

tests/app/forms/test_questionnaire_form.py
```diff
@@ -106,6 +106,69 @@ def test_date_range_to_precedes_from_raises_question_error(self):
         self.assertEqual(form.data, expected_form_data)
         self.assertEqual(form.question_errors['reporting-period-question'], expected_message)
 
+    def test_form_radio_options_no_selection(self):
```
benjefferies, line 109: Nice, really easy to read what these tests are doing/testing

tests/functional/generate_pages.py
benjefferies: Probably should regenerate page models now I guess?

benjefferies (changes requested):

benjefferies (approved):

georgschoelly:
A related pull request was just merged in wtforms, this might solve the issue: https://github.com/wtforms/wtforms/pull/288

nealedj:
Ah excellent. We'll look at updating wtforms and removing our workaround.

PR #1019, opened 2017-03-07
Don't output qcodes for empty strings

ajmaddaford:
Fixes the converter to not add q_codes for empty strings to the downstream format

### How to review 
- Add some debugging to output the downstream payload before encryption
- Complete a 0205 survey with no comments and verify q_code 146 is not in the payload
- Complete a 0205 survey with a comment and verify q_code 146 is in the payload

### DIFF
--- a/tests/app/submitter/test_converter.py
+++ b/tests/app/submitter/test_converter.py
@@ -399,6 +399,55 @@ class TestConverter(SurveyRunnerTestCase):
             self.assertEqual(answer_object['data']['1'], 'Ready salted')
             self.assertEqual(answer_object['data']['4'], 'Bacon')
 
+    def test_converter_q_codes_for_empty_strings(self):
+        with self.application.test_request_context():
+            routing_path = [Location(group_id='favourite-food', group_instance=0, block_id='crisps')]
+            answers = [create_answer('crisps-answer', '', group_id='favourite-food', block_id='crisps')]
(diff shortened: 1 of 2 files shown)


benjefferies (approved):

tests/app/submitter/test_converter.py
```diff
@@ -399,6 +399,55 @@ def test_converter_checkboxes_with_q_codes_and_other_value(self):
             self.assertEqual(answer_object['data']['1'], 'Ready salted')
             self.assertEqual(answer_object['data']['4'], 'Bacon')
 
+    def test_converter_q_codes_for_empty_strings(self):
+        with self.application.test_request_context():
+            routing_path = [Location(group_id='favourite-food', group_instance=0, block_id='crisps')]
+            answers = [create_answer('crisps-answer', '', group_id='favourite-food', block_id='crisps')]
+            answers += [create_answer('other-crisps-answer', 'Ready salted', group_id='favourite-food', block_id='crisps')]
+
+            questionnaire = {
```
benjefferies, line 408: Would it be nicer to load test_textarea.json?

collisdigital (approved):
Checked locally; works!

tests/app/submitter/test_converter.py
ajmaddaford: Maybe, it was done to be consistent with what was already there. I think given we need to spend some time increasing and improving the downstream test coverage, we should review all of the tests then.

PR #1020, opened 2017-03-07
Change question q_code 120/220 to TextField so respondent can enter number or %

collisdigital:
https://trello.com/c/xFLWeChE/820-production-questionnaire-mwss

MWSS failed PO approval because for question q_codes 120/220x the respondent needs to be able to enter a % sign to indicate if the value is a number or percentage. This PR changes the field to a TextField in the absence of a better field/solution.

### How to review 
- Check MWSS; can you enter "5%" into the question on screen 2.10/3.10/4.10 etc.

image

### DIFF
--- a/app/data/1_0005.json
+++ b/app/data/1_0005.json
@@ -548,7 +548,7 @@
                                     "label": "What was the percentage or number of weekly paid employees who received this new pay rate?",
                                     "mandatory": false,
                                     "q_code": "120",
-                                    "type": "PositiveInteger"
+                                    "type": "TextField"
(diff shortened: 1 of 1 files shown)


yoganandkunche:
@collisdigital q_120 field is now allowing all the spl chars and alphabets and negative values.

collisdigital:
Emma says this field only goes into the image not downstream so that's not a problem

LJBabbage (approved):
Ugly

samiwel (approved):

PR #1021, opened 2017-03-08
Eq 881 confirmation in navigation

samiwel:
Adds a link in the navigation menu once all sections are completed.
For surveys with Summary pages it will display a "Summary" link.
For surveys with a Confirmation page it will display a "Submit answers" link.

### How to review 
Test the functionality against the acceptance criteria in the Trello card.
Ensure that the correct link is displayed for surveys with/without Summary/Confirmation pages respectively.

### DIFF
--- a/app/questionnaire/navigation.py
+++ b/app/questionnaire/navigation.py
@@ -15,11 +15,12 @@ class Navigation(object):
     Reads navigation config from the schema and returns collections of dicts that describe
     completion status of each group
     """
-    def __init__(self, survey_json, answer_store, metadata=None, completed_blocks=None):
+    def __init__(self, survey_json, answer_store, metadata=None, completed_blocks=None, routing_path=None):
         self.survey_json = survey_json
         self.metadata = metadata or {}
         self.answer_store = answer_store
         self.completed_blocks = completed_blocks or []
+        self.routing_path = routing_path
 
     def build_navigation(self, current_group_id, current_group_instance):
         """
@@ -34,12 +35,17 @@ class Navigation(object):
 
         navigation = []
 
+        last_group_id = SchemaHelper.get_last_group_id(self.survey_json)
+
         for group in filter(lambda x: 'hide_in_navigation' not in x, self.survey_json['groups']):
             first_location = Location(group['id'], 0, group['blocks'][0]['id'])
 
             logger.debug("building frontend navigation", group_id=group['id'])
 
             skip_group = self._should_skip_group(current_group_instance, group)
+            last_block_in_current_group = group['blocks'][-1]
+            is_summary_or_confirmation = last_block_in_current_group and \
+                last_block_in_current_group['type'] in ('Summary', 'Confirmation')
 
             if not skip_group:
                 repeating_rule = SchemaHelper.get_repeat_rule(group)
@@ -47,9 +53,14 @@ class Navigation(object):
                     logger.debug("building repeating navigation", group_id=group['id'])
                     navigation.extend(self._build_repeating_navigation(repeating_rule, group, current_group_id,
                                                                        current_group_instance))
-                else:
+                elif last_group_id == group['id'] and is_summary_or_confirmation and self.routing_path:
+                    logger.debug("building navigation", group_id=group['id'])
+                    if len(self.completed_blocks) > 0 and set(self.routing_path[:-1]).issubset(self.completed_blocks):
+                        navigation.append(self._build_single_navigation(group, current_group_id, first_location))
+                elif last_group_id != group['id'] and not is_summary_or_confirmation:
                     logger.debug("building navigation", group_id=group['id'])
                     navigation.append(self._build_single_navigation(group, current_group_id, first_location))
+
         return navigation
 
     def _should_skip_group(self, current_group_instance, group):
--- a/tests/app/questionnaire/test_navigation.py
+++ b/tests/app/questionnaire/test_navigation.py
@@ -7,6 +7,7 @@ from app.questionnaire.navigation import Navigation
 from app.schema_loader.schema_loader import load_schema_file
 
 
+# pylint: disable=R0904,C0302
 class TestNavigation(unittest.TestCase):
 
     @staticmethod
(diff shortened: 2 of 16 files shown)


tests/app/questionnaire/test_navigation.py
```diff
@@ -662,3 +662,60 @@ def test_build_navigation_returns_navigation_when_schema_navigation_is_true(self
 
         # Then
         self.assertIsNotNone(nav_menu)
+
+    def test_build_navigation_confirmation_link_hidden_when_uncompleted_sections(self):
```
benjefferies: Should probably be `test_build_navigation_summary_link_hidden_when_uncompleted_sections`. Same for test below

tests/functional/spec/navigation.spec.js
```diff
@@ -72,4 +75,73 @@ describe('Navigation', function() {
...
+    openQuestionnaire('test_navigation.json')
+
+    // When
+    Navigation.completeAllSections().clickSummaryLink()
+
+    // Then
+    expect(NavigationSummaryPage.isOpen()).is.true
+  })
+
+  it('Given a survey which has a "Confirmation" page, When I have not completed all the sections, Then I will not see a link called "Submit answers" in the navigation', function() {
```
benjefferies: Could the confirmation type block scenarios be unit tested in test_navigation rather than selenium? Not sure the high level test is worth the test execution cost and extra schema etc

app/questionnaire/navigation.py
```diff
@@ -50,8 +50,20 @@ def build_navigation(self, current_group_id, current_group_instance):
                 else:
                     logger.debug("building navigation", group_id=group['id'])
                     navigation.append(self._build_single_navigation(group, current_group_id, first_location))
+
+        if self._all_sections_completed(navigation):
```
benjefferies: Instead of adding on a navigation item could we hide the final navigation (Summary or Confirmation) unless all sections are complete?

benjefferies (changes requested):

LJBabbage:
As discussed, if an author uses hide_in_navigation which isn't associated with a repeating group, this can show the summary/confirmation page before the hidden group is completed. As agreed a failing test is to be written and a note added to the navigation card to tackle this.
Currently none of our surveys use this, but it is a feasible implementation going forward

tests/functional/spec/navigation.spec.js
samiwel: Yeah, it probably could. I will re-write the selenium tests as integration tests.

benjefferies: Could probably even be unit tests in test_navigation.py

app/questionnaire/navigation.py
samiwel: Had a quick look into this. At the moment, there is no concept of toggling the visibility of items in the Navigation. Items are either added or not. Whilst I agree that it would probably be more efficient to build the navigation item and toggle its visibility rather than adding the summary/confirmation link to the end, I think it should be looked at in another PR, since it's a larger bit of work to re-work the navigation to work based on the routing path.

tests/app/questionnaire/test_navigation.py
samiwel: I will rename this test and also add some more tests as mentioned below.

samiwel:
Selenium test is currently failing due to browser being in mobile viewport and navigation needs to be toggled to be displayed. (Not sure why since I am not setting mobile viewport for those tests....) but anyway I'm going to re-write some of those tests as ~~integration~~ unit tests.

tests/functional/spec/navigation.spec.js
samiwel: Yeah, sorry I meant unit tests :) Strictly speaking though, unit tests shouldn't be interacting with a schema file.

benjefferies: True but for convenience sake

app/questionnaire/navigation.py
LJBabbage: How we are originally used 'hide in navigation' was tied to repeating groups, the summary/confirmation made use of this. I think we need to be careful if we start adding everything then hiding somethings

app/questionnaire/navigation.py
```diff
@@ -179,3 +191,11 @@ def _generate_link_names(self, repeating_rule):
             link_names[link_name] = " ".join(link_names[link_name])
 
         return link_names
+
+    @staticmethod
+    def _all_sections_completed(navigation):
```
benjefferies: We shouldn't rely on nav_items being completed to indicate all sections are complete incase there are groups that are hidden and not complete. We should check all blocks in the routing path are in completed blocks

app/questionnaire/navigation.py
benjefferies: Just to be clear I don't mean we should add then hide things. I mean only add Summary or Confirmation type group/block if they have not got `"hide_in_navigation": true` *and* the questionnaire has be completed i.e. all blocks are visited

benjefferies (changes requested):

app/questionnaire/navigation.py
samiwel: @benjefferies If I understand correctly, this could be achieved by moving all confirmation/summary blocks into their own group at the end of every schema that we currently have and removing the `hide_in_navigation` property from these blocks.

Then some additional logic would be required in `build_navigation` to check that all blocks (including any hidden groups) on the routing path are in `completed_blocks`?

benjefferies: Correct @samiwel

app/questionnaire/navigation.py
LJBabbage: Agreed with Andrew and Chris this will be tackled in the overall navigation improvements. Its what my comment was about. I have added the information on the Trello card so this PR can progress without it, it won't affect anything in flight at present

tests/app/questionnaire/test_navigation.py
```diff
@@ -719,3 +749,201 @@ def test_build_navigation_confirmation_link_visible_when_all_sections_complete(s
...
+            'link_name': 'Summary',
+            'highlight': False,
+            'repeating': False,
+            'completed': False,
+            'link_url': Location('summary-group', 0, 'summary').url(metadata)
+        }
+
+        self.assertNotIn(confirmation_link, navigation.build_navigation('property-details', 0))
+
+    @pytest.mark.xfail(reason="Summary should not display since there is a hidden group that isn't completed")
```
benjefferies: Good use of highlighting a known bug

samiwel: This should be fixed with my latest changes. I've removed the xFail annotation, will be pushing up changes soon.

app/questionnaire/navigation.py
```diff
@@ -34,22 +35,32 @@ def build_navigation(self, current_group_id, current_group_instance):
...
             if not skip_group:
                 repeating_rule = SchemaHelper.get_repeat_rule(group)
                 if repeating_rule:
                     logger.debug("building repeating navigation", group_id=group['id'])
                     navigation.extend(self._build_repeating_navigation(repeating_rule, group, current_group_id,
                                                                        current_group_instance))
-                else:
+                elif last_group_id == group['id'] and is_summary_or_confirmation and self.routing_path:
+                    logger.debug("building navigation", group_id=group['id'])
+                    if set(self.routing_path[:-1]).issubset(self.completed_blocks) and len(self.completed_blocks) > 0:
```
ajmaddaford: I think it would read better if the length check on completed_blocks was done first

collisdigital:
Still an issue with this; on UKIS with the conditional display blocks; if you complete all the blocks after the conditional ones and are currently on one of those blocks you get  the Submit answers link; if you navigate back to before an incomplete conditional display block it disappears.

## Section Before:
image

## Section After:
image

samiwel:
The latest commit should fix the above issue, but since it has to evaluate the full routing path on every request, we need to be careful (and possibly test) that this doesn't introduce a performance issue for surveys where there are a large number of repeating groups.

tests/app/data/test_schema_validation.py
```diff
@@ -41,6 +41,7 @@ def test_schemas(self):
         schema = load(schema_file)
 
         for file in files:
+            print(file)
```
benjefferies: We probably don't want this debug anymore

tests/app/questionnaire/test_navigation.py
```diff
@@ -662,3 +663,363 @@ def test_build_navigation_returns_navigation_when_schema_navigation_is_true(self
...
+
+        confirmation_link = {
+            'link_name': 'Summary',
+            'highlight': False,
+            'repeating': False,
+            'completed': False,
+            'link_url': Location('summary-group', 0, 'summary').url(metadata)
+        }
+
+        self.assertNotIn(confirmation_link, navigation.build_navigation('property-details', 0))
```
benjefferies: Is it also worth checking the number of links in the navigation for all of these `not in` tests

tests/app/questionnaire/test_navigation.py
```diff
@@ -662,3 +663,363 @@ def test_build_navigation_returns_navigation_when_schema_navigation_is_true(self
...
+            'link_name': 'Summary',
+            'highlight': False,
+            'repeating': False,
+            'completed': False,
+            'link_url': Location('summary-group', 0, 'summary').url(metadata)
+        }
+
+        self.assertIn(confirmation_link, navigation.build_navigation('property-details', 0))
+
+    def test_build_navigation_submit_answers_link_not_visible_for_survey_with_summary(self):
```
benjefferies: I question whether this test make sense now we are basing it on the blocks that have been configured. I guess what we are testing here is that a made up block cannot be added into the navigation?

tests/app/questionnaire/test_navigation.py
```diff
@@ -662,3 +663,363 @@ def test_build_navigation_returns_navigation_when_schema_navigation_is_true(self
...
+            'link_name': 'Submit answers',
+            'highlight': False,
+            'repeating': False,
+            'completed': False,
+            'link_url': Location('confirmation-group', 0, 'confirmation').url(metadata)
+        }
+
+        self.assertIn(confirmation_link, navigation.build_navigation('property-details', 0))
+
+    def test_build_navigation_summary_link_not_visible_for_survey_with_confirmation(self):
```
benjefferies: I question whether this test make sense now we are basing it on the blocks that have been configured. I guess what we are testing here is that a made up block cannot be added into the navigation?

tests/integration/questionnaire/test_questionnaire_final_confirmation.py
```diff
@@ -16,7 +16,7 @@ def test_final_confirmation_asked_at_end_of_questionnaire(self):
         post_data = {
             'action[start_questionnaire]': 'Start Questionnaire'
         }
-        resp = self.get_and_post_with_csrf_token(base_url + '14ba4707-321d-441d-8d21-b8367366e766/0/introduction', data=post_data, follow_redirects=False)
+        resp = self.get_and_post_with_csrf_token(base_url + 'final-confirmation/0/introduction', data=post_data, follow_redirects=False)
```
benjefferies, line 19: How was this ever working?

benjefferies (changes requested):

tests/app/data/test_schema_validation.py
samiwel: Well spotted.

tests/integration/questionnaire/test_questionnaire_final_confirmation.py
samiwel: Spotted it whilst fixing some of the failing tests. If I remember, it was working because the response code was still 302 so it passed the assert even through the group was incorrect.

tests/app/questionnaire/test_navigation.py
samiwel: I added this test as it was part of the acceptance criteria in the Trello card that `Submit answers` and `Summary` are mutually exclusive. One should be shown, depending on the schema, but not both. Happy to remove this test if you feel it doesn't add any value. It basically just prevents you from doing something like adding a Summay and a Confirmation page.. which actually might not make sense going forward and will eventually be enforced by an authoring tool, I'd imagine.

tests/app/questionnaire/test_navigation.py
samiwel: Same as below.

tests/app/questionnaire/test_navigation.py
samiwel: I've added some assertions to ensure the number of links are correct.

tests/app/data/test_schema_validation.py
```diff
@@ -86,6 +86,7 @@ def test_no_duplicate_ids_in_schema(self):
     def test_child_answers_define_parent(self):
         for schema_file in self.all_schema_files():
             with open(schema_file, encoding="utf8") as file:
+                print(schema_file)
```
ajmaddaford: This print statement needs to be removed

ajmaddaford (changes requested):

app/questionnaire/navigation.py
```diff
@@ -34,22 +35,32 @@ def build_navigation(self, current_group_id, current_group_instance):
...
             if not skip_group:
                 repeating_rule = SchemaHelper.get_repeat_rule(group)
                 if repeating_rule:
                     logger.debug("building repeating navigation", group_id=group['id'])
                     navigation.extend(self._build_repeating_navigation(repeating_rule, group, current_group_id,
                                                                        current_group_instance))
-                else:
+                elif last_group_id == group['id'] and is_summary_or_confirmation and self.routing_path:
+                    logger.debug("building navigation", group_id=group['id'])
+                    if len(self.completed_blocks) > 0 and set(self.routing_path[:-1]).issubset(self.completed_blocks):
```
benjefferies, line 58: Shouldn't we check the length of routing_path or is that a safe operation on empty arrays?

tests/integration/questionnaire/test_questionnaire_final_confirmation.py
benjefferies: 👍

benjefferies (changes requested):

app/questionnaire/navigation.py
samiwel: For an empty routing path `self.routing_path[:-1]` == `[]`. So we'd be checking if `completed_blocks` is a subset of `[]` where we have more than zero completed blocks. So this should be safe.

collisdigital:
Rebase and squash?

benjefferies (approved):

ajmaddaford (approved):

PR #1025, opened 2017-03-09
Add support for dumping data by authorised users

collisdigital:
- Added a /dump route Blueprint that allows the answer store or submission data to be dumped as JSON at any time.
- This requires a user to be logged-in and their metadata to contain a 'role' of 'dumper'.
- The /dev page has been updated to always enable the 'dumper' role.
- The 'role' metadata has been persisted to the user's metadata.
- A `role_required` function decorator has been added to `app.authentication.roles` that allows a function to require a role in the user's metadata or will raise a Flask Forbidden exception.

### How to test?
- Start a questionnaire and answer some questions.
- To dump the answer store go to: /dump/answers
- To dump the submission data go to: /dump/submission
- Check that an invalid dump gives an error e.g. /dump/nothing
- Save and complete later and verify you cannot get to /dump/answers or submission anymore (session expired)

- If you try dumping when there are no answer (e.g. on landing page) you'll get a
bad request because the answer store is empty.

### To review:
- The naming of things, is dump correct, what about the role?
- Are we happy to have the dump blueprint in Production?
- Are the URLs meaningful/standard?
- Is the output useful?

### To-do:
- [x] Add tests
- [x] code review
- [x] rebase and squash

### DIFF
--- a/app/views/dump.py
+++ b/app/views/dump.py
@@ -1,0 +1,33 @@
+from flask_login import current_user
+from flask_login import login_required
+from flask import Blueprint, jsonify
+
+
+from app.authentication.roles import role_required
+from app.globals import get_answer_store, get_metadata
+from app.submitter.converter import convert_answers
+from app.utilities.schema import get_schema
+from app.questionnaire.path_finder import PathFinder
+
+
+dump_blueprint = Blueprint('dump', __name__)
+
+
+@dump_blueprint.route('/dump/answers', methods=['GET'])
+@login_required
+@role_required('dumper')
+def dump_answers():
+    response = {'answers': get_answer_store(current_user).answers or []}
+    return jsonify(response), 200
+
+
+@dump_blueprint.route('/dump/submission', methods=['GET'])
+@login_required
+@role_required('dumper')
+def dump_submission():
+    answer_store = get_answer_store(current_user)
+    metadata = get_metadata(current_user)
+    schema = get_schema(metadata)
+    routing_path = PathFinder(schema, answer_store, metadata).get_routing_path()
+    response = {'submission': convert_answers(metadata, schema, answer_store, routing_path)}
+    return jsonify(response), 200
--- a/tests/app/views/test_dump.py
+++ b/tests/app/views/test_dump.py
@@ -1,0 +1,211 @@
+import json
+from tests.integration.integration_test_case import IntegrationTestCase
+
+
+class TestDumpAnswers(IntegrationTestCase):
+
+    def test_dump_answers_not_authenticated(self):
+        # Given I am not an authenticated user
+        # When I attempt to dump the answer store
+        response = self.client.get('/dump/answers')
+
+        # Then I receive a 401 Unauthorised response code
+        self.assertEqual(401, response.status_code)
+
+    def test_dump_answers_authenticated_missing_role(self):
+        # Given I am an authenticated user who has launched a survey
+        # but does not have the 'dumper' role in my metadata
+        self.launchSurvey('test', 'radio')
+
+        # When I attempt to dump the answer store
+        response = self.client.get('/dump/answers')
+
+        # Then I receive a 403 Forbidden response code
+        self.assertEqual(403, response.status_code)
+
+        # And the response data contains Forbidden
+        actual = response.get_data(True)
+        self.assertIn('Forbidden', actual)
+
+    def test_dump_answers_authenticated_with_role_no_answers(self):
+        # Given I am an authenticated user who has launched a survey
+        # and does have the 'dumper' role in my metadata
+        _, response = self.launchSurvey('test', 'radio', roles=['dumper'])
+
+        # When I haven't submitted any answers
+        # And I attempt to dump the answer store
+        response = self.client.get('/dump/answers')
+
+        # Then I get a 200 OK response
+        self.assertEqual(200, response.status_code)
+
+        # And the JSON response contains an empty array
+        actual = json.loads(response.get_data(True))
+        expected = {'answers': []}
+
+        self.assertDictEqual(actual, expected)
+
+    def test_dump_answers_authenticated_with_role_with_answers(self):
+        # Given I am an authenticated user who has launched a survey
+        # and does have the 'dumper' role in my metadata
+        url, response = self.launchSurvey('test', 'radio', roles=['dumper'])
+
+        # When I submit an answer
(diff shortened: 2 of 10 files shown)


app/views/dump.py
```diff
@@ -0,0 +1,52 @@
...
+from app.utilities.schema import get_schema
+from app.questionnaire.path_finder import PathFinder
+
+
+dump_blueprint = Blueprint('dump', __name__)
+
+
+@dump_blueprint.route('/dump/<store>', methods=['GET'])
+@login_required
+def dump_data(store):
```
collisdigital: Should this be split into multiple routes?

LJBabbage: Would get rid of some of the logic which is always a win, and you then could put in in a role checker which could be used for the flusher as well

app/views/dump.py
```diff
@@ -0,0 +1,52 @@
...
+    response = {'status': 'Unknown', 'message': ''}
+    metadata = get_metadata(current_user)
+
+    if 'dumper' in metadata.get('roles', []):
+        answer_store = get_answer_store(current_user)
+
+        if answer_store.answers:
+            if store == 'answers':
+                response['status'] = 'OK'
```
benjefferies: Response seems unnecessary since we get a message and the http code

app/views/dump.py
benjefferies: I'd agree with different routes

app/views/dump.py
```diff
@@ -0,0 +1,52 @@
...
+@dump_blueprint.route('/dump/<store>', methods=['GET'])
+@login_required
+def dump_data(store):
+    response = {'status': 'Unknown', 'message': ''}
+    metadata = get_metadata(current_user)
+
+    if 'dumper' in metadata.get('roles', []):
+        answer_store = get_answer_store(current_user)
+
+        if answer_store.answers:
```
benjefferies: Is the answer_store.answers check necessary?

tests/integration/create_token.py
```diff
@@ -1,32 +1,33 @@
...
+    'employment_date': '1983-06-02',
+    'variant_flags': None,
+    'region_code': 'GB-ENG',
+    'language_code': 'en',
+    'exp_time': 3600  # one hour from now
 }
 
 
-def create_token(form_type_id, eq_id, start_date=None, end_date=None, employment_date=None, region_code=None, language_code=None):
+def create_token(form_type_id, eq_id, **metadata_kwargs):
```
benjefferies: Did a similar thing for jti claim branch, now that's in it's worth rebasing and resolving conflicts

benjefferies (changes requested):

app/views/dump.py
collisdigital: Agreed, will remove

tests/integration/create_token.py
collisdigital: Ah great minds, will merge it.

app/views/dump.py
collisdigital: Nope, tidying this all up as part of the route split

app/views/dump.py
collisdigital: On it

app/views/dump.py
benjefferies: Also we might want to submit answers for a survey which is completely optional (all questions optional)

app/views/dump.py
```diff
@@ -0,0 +1,33 @@
...
+from app.questionnaire.path_finder import PathFinder
+
+
+dump_blueprint = Blueprint('dump', __name__)
+
+
+@dump_blueprint.route('/dump/answers', methods=['GET'])
+@login_required
+@role_required('dumper')
+def dump_answers():
```
benjefferies, line 19: Much cleaner

app/views/dump.py
collisdigital: Done

app/views/dump.py
collisdigital: Done

app/views/dump.py
collisdigital: Done

tests/integration/create_token.py
collisdigital: Done

tests/app/views/test_dump.py
```diff
@@ -0,0 +1,211 @@
...
+
+        # When I attempt to dump the answer store
+        response = self.client.get('/dump/answers')
+
+        # Then I receive a 403 Forbidden response code
+        self.assertEqual(403, response.status_code)
+
+        # And the response data contains Forbidden
+        actual = response.get_data(True)
+        self.assertIn('Forbidden', actual)
```
benjefferies, line 28: Probably will fail the test now

tests/integration/create_token.py
```diff
@@ -4,25 +4,26 @@
...
+    'variant_flags': None,
+    'region_code': 'GB-ENG',
+    'language_code': 'en',
+    'roles': [],
 }
 
 
 def create_token(form_type_id, eq_id, **extra_payload):
-    payload_vars = PAYLOAD
+    payload_vars = PAYLOAD.copy()
```
benjefferies, line 26: Good spot

app/views/dump.py
```diff
@@ -0,0 +1,33 @@
...
+from app.utilities.schema import get_schema
+from app.questionnaire.path_finder import PathFinder
+
+
+dump_blueprint = Blueprint('dump', __name__)
+
+
+@dump_blueprint.route('/dump/answers', methods=['GET'])
+@login_required
+@role_required('dumper')
```
benjefferies, line 18: Worth adding to flush now or later?

benjefferies (approved):

app/views/dump.py
collisdigital: Not easy to do; flush doesn't use the normal login route or user/metadata; we should consider whether to update that later I think

tests/app/views/test_dump.py
collisdigital: Passes 👍

LJBabbage (approved):
Sure you have enough tests?

collisdigital:
@LJBabbage it was a pain getting the decorator right; tests mostly helped me understand how it should/did behave in these various scenarios as it's functions returning functions retuning functions wrapping functions! Could possible remove some of the tests now it's working; but do we want to?

PR #1026, opened 2017-03-09
Page titles populated with questionnaire data

benjefferies:
As a respondent, I want the page titles to be descriptive, so that I can
easily tell what the page content is with and without assistive
technologies.

* Improve use of web analytics data as will be able to better recognise
  pages
* Increase reach by improving accessibility

Changing the page titles so they reflect the acceptance criteria in the [trello card](https://trello.com/c/b3IQKkea/840-page-titles-for-blocks)

### How to review 
1. Open up the census_household.json survey
1. Answer questionnaire checking page title is correct and … replaces sensitive information.
1. Check history for page titles


app/views/questionnaire.py
```diff
@@ -457,6 +463,26 @@ def _get_front_end_navigation(answer_store, current_location, metadata):
         return None
 
 
+def get_page_title_for_block(schema_json, current_location):
```
LJBabbage: To me it feels like this should be passing a block and not current_location. Obviously we would then need a helper like get_group_for_block if we don't have something like that already.

LJBabbage (changes requested):

app/views/questionnaire.py
```diff
@@ -173,12 +173,14 @@ def get_thank_you(eq_id, form_type, collection_id):  # pylint: disable=unused-ar
     logger.debug("theme selected", theme=theme)
     metadata = get_metadata(current_user)
     metadata_context = build_metadata_context(metadata)
+    page_title = 'We\'ve received your answers - {survey_title}'.format(survey_title=g.schema_json['title'])
```
LJBabbage: Welsh? I do wonder if these messages should be extracted so they can be updated easier if needed. Not sure

LJBabbage (changes requested):

tests/app/templating/test_template_renderer.py
```diff
@@ -91,3 +91,31 @@ def test_render_nested_templatable_property(self):
         }
 
         self.assertEqual(rendered, expected)
+
+    def test_safe_content_should_be_none_safe(self):
```
LJBabbage: Not a clue when just reading the function name

tests/app/templating/test_template_renderer.py
```diff
@@ -91,3 +91,31 @@ def test_render_nested_templatable_property(self):
...
+        content = None
+
+        safe_content = TemplateRenderer.safe_content(content)
+
+        self.assertEqual(safe_content, None)
+
+    def test_empty_should_be_return_empty(self):
```
LJBabbage: Not much better

app/views/questionnaire.py
benjefferies: Not sure the benefit we will get by passing the block instead of the current_location?

app/views/questionnaire.py
benjefferies: Yeah this is a good point, maybe I should put this in the template and wrap with _() method

tests/app/views/test_questionnaire.py
```diff
@@ -397,3 +397,63 @@ def test_remove_empty_household_members_middle_name_only_not_stored(self):
...
+        g.schema_json = load_schema_file("test_final_confirmation.json")
+
+        # When
+        page_title = get_page_title_for_block(g.schema_json, Location('final-confirmation', 0, 'introduction'))
+
+        # Then
+        self.assertEqual(page_title, 'Final confirmation to submit')
+
+    def test_page_title_should_be_group_title_then_survey_title_when_type_interstitial(self):
+        # Given
```
LJBabbage: Would prefer 'and' rather than 'then' in the function name, confused me as I thought it was a test pattern (Given, when, then) also it would be better to use schema_title as that is what the function is using

tests/app/views/test_questionnaire.py
```diff
@@ -397,3 +397,63 @@ def test_remove_empty_household_members_middle_name_only_not_stored(self):
...
+        g.schema_json = load_schema_file("test_interstitial_page.json")
+
+        # When
+        page_title = get_page_title_for_block(g.schema_json, Location('favourite-foods', 0, 'breakfast-interstitial'))
+
+        # Then
+        self.assertEqual(page_title, 'Favourite food - Interstitial Pages')
+
+    def test_page_title_should_be_question_then_survey_title_when_type_questionnaire(self):
+        # Given
```
LJBabbage: As above

tests/app/views/test_questionnaire.py
```diff
@@ -397,3 +397,63 @@ def test_remove_empty_household_members_middle_name_only_not_stored(self):
...
+        g.schema_json = load_schema_file("test_final_confirmation.json")
+
+        # When
+        page_title = get_page_title_for_block(g.schema_json, Location('final-confirmation', 0, 'breakfast'))
+
+        # Then
+        self.assertEqual(page_title, 'What is your favourite breakfast food - Final confirmation to submit')
+
+    def test_page_title_should_contain_ellipsis_when_type_questionnaire_with_piping(self):
+        # Given
```
LJBabbage: You used Jinja filter above, would prefer consistancey

LJBabbage (changes requested):

app/views/questionnaire.py
LJBabbage: As agreed the name change works as well

app/templates/confirmation.html
```diff
@@ -1,5 +1,5 @@
 {% extends 'questionnaire.html' %}
 
-{% block page_title %}Confirmation{% endblock %}
+{% block page_title %}{{page_title}}{% endblock %}
```
hamishtaplin: I think the idea behind having a block for the page title was that it could be set in the templates as we had no proper system of deriving the page title. Now we do, this can just be set in the `_base` template as the default value of that block (so it can be overridden at template level if needs-be)

hamishtaplin (changes requested):

app/templates/confirmation.html
benjefferies: Not sure it's worth it. I think most if not all pages have different page_titles now. You may have put this comment before I refactored based on @LJBabbage review

hamishtaplin (approved):

app/views/questionnaire.py
```diff
@@ -457,6 +460,20 @@ def _get_front_end_navigation(answer_store, current_location, metadata):
         return None
 
 
+def get_page_title_for_location(schema_json, current_location):
+    block = SchemaHelper.get_block_for_location(schema_json, current_location)
+    if block['type'] == 'Interstitial':
+        group = SchemaHelper.get_group(schema_json, current_location.group_id)
+        page_title = '{group_title} - {schema_title}'.format(group_title=group['title'], schema_title=schema_json['title'])
```
collisdigital: Or is this survey title rather than schema title?

app/views/questionnaire.py
```diff
@@ -457,6 +460,20 @@ def _get_front_end_navigation(answer_store, current_location, metadata):
...
+    elif block['type'] == 'Questionnaire':
+        first_question = SchemaHelper.get_questions_for_block(block)[0]
```
collisdigital: Maybe a get_first_question_for_block?

app/views/questionnaire.py
```diff
@@ -457,6 +460,20 @@ def _get_front_end_navigation(answer_store, current_location, metadata):
...
+        page_title = '{question_title} - {schema_title}'.format(question_title=first_question['title'], schema_title=schema_json['title'])
+    else:
+        page_title = schema_json['title']
+
+    return TemplateRenderer.safe_content(page_title)
```
collisdigital: I wonder if we need something just in case page_title is an empty string, a fallback like "ONS Survey"?

requirements_for_test.txt
```diff
@@ -7,3 +7,4 @@ jsonschema==2.5.1
 pylint==1.6.4
 flask-debugtoolbar==0.10.0
 pylint-mccabe==0.1.3
+-e git+https://github.com/kuc2477/extraction.git@8595eb187c575cb7bcf13efa2bc0b8c4f86e908b#egg=extraction
```
collisdigital, line 10: This seems to have been committed below?

tests/app/templating/test_template_renderer.py
```diff
@@ -91,3 +91,31 @@ def test_render_nested_templatable_property(self):
...
+
+    def test_given_empty_string_safe_content_should_return_empty_string(self):
+        content = ''
+
+        safe_content = TemplateRenderer.safe_content(content)
+
+        self.assertEqual(safe_content, '')
+
+    def test_should_replace_jinja_template_variable(self):
+        content = 'Your answer is {{ answers.your_answer}}'
```
collisdigital, line 110: Is it also valid to have a single { for jinja?

app/views/questionnaire.py
collisdigital: Or maybe that should always be on the end " - ONS" I mean?

collisdigital: Will ping @madebymoon

app/views/questionnaire.py
benjefferies: Done

app/templates/signed-out.html
```diff
@@ -1,6 +1,6 @@
 {% extends theme('layouts/_twocol.html') %}
 
-{% block page_title %}Save Successful{% endblock %}
```
collisdigital, line 3: If {{survey_title}} is /always/ meant to be here is there a way to set this in the base template so it cannot be missed? so that you just define the title to the left of the survey title in derived pages?

collisdigital: Where there isn't a survey title the base page could fallback to a safe default using a filter/if?

app/views/questionnaire.py
benjefferies: As it's the only place we get the first question does it need to be refactored out if get_questions_for_block is a list generator?

app/views/questionnaire.py
benjefferies: survey title is required but there is nothing stopping up from setting an empty string. Maybe we should enforce a proper survey title?

app/templates/signed-out.html
benjefferies: The default is 'eQ - Survey Runner' in _base.html. Might need updating to...?

tests/app/templating/test_template_renderer.py
benjefferies: Tried it and it didn't substitute anything

app/views/questionnaire.py
collisdigital: much better

app/views/questionnaire.py
collisdigital: change the schemav1.json to enforce a minimum title length (suggest 3 e.g. RSI)?

tests/app/templating/test_template_renderer.py
collisdigital: Worth checking some other string combinations that are valid e.g. pipe symbol is used often as well as brackets e.g. {{something|length}} or {{something(else, a, b)}} just to cover a few scenarios from our current schemas (maybe worth pull some of these out of UKIS, Census etc.

app/templates/signed-out.html
collisdigital: to... "ONS Survey" ?

collisdigital:
This may be a defect/issue to be raised outside of this PR or could be related; but Chrome's history is all off-by-one when looking at Page Titles vs. URLs e.g. the Page Title for the URL in the history is for the wrong page???

collisdigital:
Hmm or behaviour is inconsistent; can't seem to get the date of birth question to keep the right Page Title in Chrome's history: http://localhost:5000/questionnaire/census/household/789/household-member/0/date-of-birth as soon as the page is submitted the Page Title in history for the date of birth URL changes to the page title for the next question, so the history contains two identically named items for two different URLs , WTF?

app/templates/signed-out.html
benjefferies: Done

tests/app/templating/test_template_renderer.py
benjefferies: Done

requirements_for_test.txt
benjefferies: Done

benjefferies:
History on Chrome: When a page is loaded you get 2 history entries which look like it's for that page, then when you move to the next page one of those entries change to the now current page and you also get a second entry.
History on Safari: When going through the questionnaire you will only see a history entry for the current/last page visited.

All seems a bit strange to me but the page title always appears to be correct

collisdigital:
@hamishtaplin any comment on the weirdness of browser history?
@benjefferies only that the page titles in my Chrome history don't match for some entries; i.e. I click the history entry and it takes me to a different page than the title. I wouldn't spend any time on this unless there something obvious

tests/integration/questionnaire/test_questionnaire_page_titles.py
```diff
@@ -0,0 +1,116 @@
...
+
+class TestQuestionnairePageTitles(IntegrationTestCase):
+
+    def test_should_have_question_in_page_title_when_loading_introduction(self):
+        # Given
+        token = create_token('final_confirmation', 'test')
+        resp = self.client.get('/session?token=' + token.decode(), follow_redirects=True)
+
+        # When
+        extraction = Extractor().extract(resp.get_data(True))
```
collisdigital, line 15: Consider moving this into a helper in the base class like `get_page_title(response)` so we can start to build more convenience functions in there? We should think about the naming in the base class now if it's getting bigger; originally it followed the unittest module naming conventions and style.

collisdigital: Or perhaps `assertPageTitleEqual(response, 'title')` ?

collisdigital:
Looking good; needs a squash and rebase now, just consider whether to move the page title checks into the IntegrationTestCase class.

LJBabbage (approved):

Issue #1007, opened 2017-03-02
Apache log rotation causes high response times under load


ajmaddaford:
### Expected behaviour
When under load response times are not affected by Apache rotating logs. The graph below shows response time spikes on the hour under load.

![eq_response_times](https://cloud.githubusercontent.com/assets/1754822/23506146/a8252686-ff3e-11e6-9446-b07bee206b24.png)

### Actual behaviour
Response times spike on the hour when log rotation occurs.

### Steps to reproduce the behaviour
Load the service over an extended period and monitor the response times.

### Technical information
Survey runner running on Elastic Beanstalk

PR #1028, opened 2017-03-09
Prevent apache restarts on logrotate

jonnyshaw89:
Changed log rotate to not rotate which prevents apache process being restarted to switch to a new file.

### How to review 
Run in AWS and confirm that Logs are still being delivered to CloudWatch and that on the hour the apache process is not restarted.

Fixes https://github.com/ONSdigital/eq-survey-runner/issues/1007

### DIFF
--- a/.ebextensions/deploy.config
+++ b/.ebextensions/deploy.config
@@ -7,3 +7,8 @@ packages:
 commands:
     01-upgrade-pip:
         command: "/opt/python/run/venv/bin/pip  install --upgrade pip"
+
+files:
+    "/etc/logrotate.elasticbeanstalk.hourly/logrotate.elasticbeanstalk.httpd.conf":
+        content: |
+           # The content of this file has been removed to prevent log rotation


.ebextensions/cloudwatchlogs-apache/cwl-setup.config
```diff
@@ -1,138 +0,0 @@
-########################################################################
```
collisdigital: hmm, had we added this ourselves? I guess EB now does this itself instead?

jonnyshaw89: It made no difference when I took it out

.ebextensions/deploy.config
```diff
@@ -10,3 +10,22 @@ commands:
...
+           /var/log/httpd/* {
+                size 10M
+                missingok
+                notifempty
+                rotate 5
+                sharedscripts
+                compress
+                dateext
+                dateformat -%s
+                copytruncate
```
benjefferies: I take it we are happy with the small chance of losing logs here?

ajmaddaford (approved):

.ebextensions/cloudwatchlogs-apache/cwl-setup.config
benjefferies: The docs around this is very wooly but I have tested and still getting application logs after deploying on this branch

.ebextensions/deploy.config
ajmaddaford: No, I think we need to take another look at this. I have separated out the log format change from this in #1029

ajmaddaford (changes requested):
We need to look at alternatives that don't lose log lines

ajmaddaford:
Some options we should discuss:
- Apache piped logs (see http://httpd.apache.org/docs/2.2/logs.html#piped and http://tech.zarmory.com/2015/06/aws-elastic-beantalk-is-amateur.html)
- Change log rotation to happen daily
- Do nothing, and under high load accept slow requests on the hour when log rotation is happening

.ebextensions/deploy.config
```diff
@@ -7,3 +7,11 @@ packages:
 commands:
     01-upgrade-pip:
         command: "/opt/python/run/venv/bin/pip  install --upgrade pip"
+
+files:
+    "/etc/logrotate.elasticbeanstalk.hourly/logrotate.elasticbeanstalk.httpd.conf":
+        mode: "000644"
```
benjefferies: What is the default file permissions/owner of this file, do we actually need to configure it?

benjefferies (changes requested):

.ebextensions/deploy.config
jonnyshaw89: These were set to the same as was already on the instance. I've removed as they are not needed.

benjefferies (approved):

ajmaddaford (approved):

PR #1034, opened 2017-03-13
Simplify integration tests by enhancing IntegrationTestCase

collisdigital:
Make IntegrationTestCase smarter so that the integration tests themselves are simpler.

**For High Level Review**

- Integration tests can now interact with Flask more like a browser with get/post operations.
- get/post operations follow redirects and cache the results for asserting against
- Added a series of `assert` functions that can be used to verify the previous get/post e.g. `assertInPage` to check for content in the HTML.
- Major regex find/replace to update tests to use new methods then cleaned them up a bit.
- Changed tests not to depend on EQ_DEV_MODE being set (test sets it on setUp) if needed
- Removed settings.EQ_SERVER_SIDE_STORAGE_DATABASE_URL = "sqlite://" as it had no effect due to import order; however setting it outside of the integration tests improves performance by about 30% (memory db rather than file) and also avoids trashing your local SQLite database. Should consider setting this in run_test_unit.sh (I've enabled for Travis to test)?
- Moved MyMockSubmitter into IntegrationTestCase and removed the old derived class; any component test that submitted data was failing unless the RabbitMQ settings were disabled; so this allows all tests to submit and the data submitted to be investigated.
        

### How to review 
- Check IntegrationTestCase to start with and see the new methods
- Look at some of the integration test cases to see the difference
- Is the approach useful/better?
- Is the naming sensible?
- Pros/Cons?


collisdigital:
Closing as Travis is broked

PR #1035, opened 2017-03-13
Simplify integration tests by enhancing IntegrationTestCase

collisdigital:
Make IntegrationTestCase smarter so that the integration tests themselves are simpler.

**For High Level Review**

- Integration tests can now interact with Flask more like a browser with get/post operations.
- get/post operations follow redirects and cache the results for asserting against
- Added a series of `assert` functions that can be used to verify the previous get/post e.g. `assertInPage` to check for content in the HTML.
- Major regex find/replace to update tests to use new methods then cleaned them up a bit.
- Changed tests not to depend on EQ_DEV_MODE being set (test sets it on setUp) if needed
- Removed settings.EQ_SERVER_SIDE_STORAGE_DATABASE_URL = "sqlite://" as it had no effect due to import order; however setting it outside of the integration tests improves performance by about 30% (memory db rather than file) and also avoids trashing your local SQLite database. Should consider setting this in run_test_unit.sh (I've enabled for Travis to test)?
- Moved MyMockSubmitter into IntegrationTestCase and removed the old derived class; any component test that submitted data was failing unless the RabbitMQ settings were disabled; so this allows all tests to submit and the data submitted to be investigated.
        

### How to review 
- Check IntegrationTestCase to start with and see the new methods
- Look at some of the integration test cases to see the difference
- Is the approach useful/better?
- Is the naming sensible?
- Pros/Cons?


collisdigital:
Travis didn't like this branch and wouldn't build it until I removed:

```
branches:
    only:
        - master
```

But I don't know why :(

collisdigital:
This is why this PR wasn't building: https://www.traviscistatus.com/incidents/j7f0sp50n9xw I've reverted the .travis.yml change

tests/integration/household_composition/test_household_question.py
```diff
@@ -1,198 +1,108 @@
 from werkzeug.datastructures import MultiDict
```
benjefferies, line 1: Nice tidy up

tests/integration/household_composition/test_repeating_household.py
```diff
@@ -1,141 +1,76 @@
...
-        form_data.add("action[save_continue]", '')
-        resp = self.get_and_post_with_csrf_token(person2_age_page, data=form_data, follow_redirects=False)
-        person2_shoe_size_page = resp.location
-
-        navigate_to_page(self.client, person2_shoe_size_page)
-        form_data = MultiDict()
-        form_data.add("what-is-your-shoe-size", '9993')
-        form_data.add("action[save_continue]", '')
-        self.get_and_post_with_csrf_token(person2_shoe_size_page, data=form_data, follow_redirects=False)
+        self.post({'what-is-your-age': '9990'})
```
benjefferies, line 26: Seeing it like this versus the old diff makes you question how did it get so bad

tests/integration/integration_test_case.py
```diff
@@ -2,18 +2,50 @@
...
+        self.last_response = None
+        self.last_csrf_token = None
 
+        # Perform setup steps
+        self._setUpSettings()
+        self._setUpMockSubmitter()
+        self._setUpApp()
+
+    @staticmethod
+    def _setUpSettings():
```
benjefferies, line 47: I question the naming convention here

tests/integration/integration_test_case.py
```diff
@@ -22,78 +54,163 @@ def setUp(self):
             password = os.getenv(password_name, dev_default)
             vars(settings)[password_name] = password  # assigns attribute to this module
 
-        self.application = create_app()
-        self.client = self.application.test_client()
+    def _setUpApp(self):
+        self._application = create_app()
+        self._client = self._application.test_client()
+        # Clear any existing state before the test starts
+        self.clearDatabase()
```
benjefferies, line 61: Do we need to execute on set up and teardown or at all now we are using an in memory database?

benjefferies (changes requested):

tests/integration/integration_test_case.py
collisdigital: Happy to change; what do you suggest?

tests/integration/household_composition/test_repeating_household.py
collisdigital: I guess Flask doesn't help; it should be simpler out the box to write clean integration tests; the docs are also not particularly helpful. Once you know enough to write tests it's already too late!

tests/integration/integration_test_case.py
collisdigital: Because we init the database on import rather than setup it will contain data from other tests in the same test case. I thought each test should be run with a clean db so wipe it here.

tests/integration/integration_test_case.py
benjefferies: I was just concerned with the camel casing version snake case

tests/integration/integration_test_case.py
collisdigital: yeah tricky as the underlying base class (`TestCase`) uses that format but PEP8 is `_` based; so I think the `Assert` calls should be consistent with TestCase but not sure about these ones; as `setUp` is from `TestCase` but the private functions could be `_` based?

tests/integration/integration_test_case.py
collisdigital: yeah tricky as the underlying base class (`TestCase`) uses that format but PEP8 is `_` based; so I think the `Assert` calls should be consistent with TestCase but not sure about these ones; as `setUp` is from `TestCase` but the private functions could be `_` based?

tests/integration/integration_test_case.py
benjefferies: I guess we only need a clear of the database on the set up or teardown not both

app/views/dev_mode.py
```diff
@@ -20,7 +20,7 @@
 def dev_mode():
     if request.method == "POST":
         form = request.form
-        user = form.get("user_id")
+        user_id = form.get("user_id")
```
LJBabbage, line 23: sneaky, but better

requirements_for_test.txt
```diff
@@ -7,4 +7,4 @@ jsonschema==2.5.1
 pylint==1.6.4
 flask-debugtoolbar==0.10.0
 pylint-mccabe==0.1.3
--e git+https://github.com/kuc2477/extraction.git@8595eb187c575cb7bcf13efa2bc0b8c4f86e908b#egg=extraction
+BeautifulSoup4==4.5.3
```
LJBabbage, line 10: Nice additions

LJBabbage (approved):

samiwel (approved):
Looks a lot better to me.

benjefferies (approved):

PR #1038, opened 2017-03-15
Allow functional test to be run on remote environments.

jonnyshaw89:
This PR adds the ability for the functional tests to be run against a defined environment

### How to review 
Run the following script but add your dev environment url
`REMOTE_ADDRESS=concourse-surveys.dev.eq.ons.digital EQ_FUNCTIONAL_TEST_ENV=remote TRAVIS=true sh scripts/run_tests_functional.sh`

### DIFF
--- a/gulpfile.babel.js
+++ b/gulpfile.babel.js
@@ -15,12 +15,16 @@ import a11ym from './gulp/a11ym'
 import {fonts} from './gulp/fonts'
 
 const getEnv = () => {
+  var env = yargs.argv.env
+  if (env.startsWith("http")) {
+    return env
+  }
   const envs = {
     local: 'http://localhost:5000',
     docker: 'http://localhost',
-    preprod: 'https://preprod-surveys.eq.ons.digital'
+    preprod: 'https://eq.onsdigital.uk',
   }
-  return envs[yargs.argv.env] || envs['local']
+  return envs[env] || envs['local']
 }
 
 gulp.task('test:a11ym', (done) => {
--- a/README.md
+++ b/README.md
@@ -247,6 +247,11 @@ NEW_RELIC_LICENSE_KEY - Enable new relic monitoring by supplying a New Relic lic
 NEW_RELIC_APP_NAME - The name to display for the application in New Relic
 ```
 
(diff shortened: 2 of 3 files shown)


gulpfile.babel.js
```diff
@@ -18,7 +18,8 @@ const getEnv = () => {
   const envs = {
     local: 'http://localhost:5000',
     docker: 'http://localhost',
-    preprod: 'https://preprod-surveys.eq.ons.digital'
+    preprod: 'https://preprod-surveys.eq.ons.digital',
+    remote: 'https://' + process.env.REMOTE_ADDRESS
```
collisdigital: should this be EQ_FUNCTIONAL_TEST_REMOTE_ADDRESS and added to README.md?

collisdigital (changes requested):
Tested against pre-prod; works !

gulpfile.babel.js
```diff
@@ -18,7 +18,8 @@ const getEnv = () => {
   const envs = {
     local: 'http://localhost:5000',
     docker: 'http://localhost',
-    preprod: 'https://preprod-surveys.eq.ons.digital'
+    preprod: 'https://preprod-surveys.eq.ons.digital',
```
collisdigital: Worth updating this to the new pre-prod address as well?

gulpfile.babel.js
ajmaddaford: I think it's confusing having a command line argument to choose the environment and an environment variable to set the URL. Maybe the `env` command line argument could take an environment name e.g. preprod, or a URL? If provided with a URL it would use that, else try and lookup an environment by name.

Something like:
```
const getEnv = () => {
  var env = yargs.argv.env
  if (env.startsWith("http")) {
    return env
  }
  const envs = {
    local: 'http://localhost:5000',
    docker: 'http://localhost',
    preprod: 'https://eq.onsdigital.uk',
  }
  return envs[env] || envs['local']
}
```

ajmaddaford (changes requested):

ajmaddaford (approved):

collisdigital (approved):

PR #1039, opened 2017-03-15
Add logging

ajmaddaford:
Adds additional logging to enable better cross service monitoring and tracing of requests

- Logs the payload and the result of queueing the response in RabbitMQ separately - we need the payload (encrypted) as a last ditch backup should the queues fail, but it can't be fed into Splunk as it's too large.
- Logs ru_ref, collection_exercise_sid, and tx_id in a single event on launch
- Logs out each request visited with a tx_id
- Logs a unique `request_id` for each request

### How to review 
Run and verify the log output matches the changes listed above

### DIFF
--- a/app/submitter/submitter.py
+++ b/app/submitter/submitter.py
@@ -52,7 +52,8 @@ class Submitter(object):
 class LogSubmitter(Submitter):
 
     def send_message(self, message, queue):
-        logger.info("message submitted", message=message)
+        logger.info("sending message")
+        logger.info("message payload", message=message)
         return True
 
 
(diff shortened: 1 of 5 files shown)


app/views/questionnaire.py
```diff
@@ -42,11 +43,13 @@
 @questionnaire_blueprint.before_request
 @login_required
 def check_survey_state():
```
jonnyshaw89: I think this name is due a change as its not just `check_survey_state` its more like a generic pre request setup and checks.

jonnyshaw89 (approved):

app/submitter/submitter.py
```diff
@@ -86,7 +86,8 @@ def send_message(self, message, queue):
         :return: a boolean value indicating if it was successful
         """
         message_as_string = str(message)
-        logger.info("sending message to rabbit mq", message=message_as_string)
+        logger.info("sending message to rabbit mq")
+        logger.info("message payload", message=message_as_string)
```
collisdigital: Should this and all log lines here have RabbitMQ in them or as a prefix/label etc. so they can easily be filtered/searched for.

app/views/session.py
```diff
@@ -19,6 +21,12 @@
 session_blueprint = Blueprint('session', __name__)
 
 
+@session_blueprint.before_request
+def add_request_id():
+    request_id = str(uuid4())
```
collisdigital: Why is this duplicated from `questionnaire.py` ?

app/submitter/submitter.py
collisdigital: Discussing with @eldeal it would be good to have a `category='RabbitMQ'` or similar

ajmaddaford: I've added removed 'rabbit mq' from the event and added `category=rabbitmq` to all the log lines

app/views/session.py
ajmaddaford: Good spot, I've moved this to the application `before_request`

samiwel (approved):

benjefferies (approved):

collisdigital (approved):

PR #1044, opened 2017-03-16
Fixes #1043 submit answers shown when not on routing path.

### LINKED ISSUE #1043: Submit answers link displayed even though confirmation page not on routing path
### Expected behaviour
When user should not be able to route to the summary/confirmation page the summary/confirmation navigation entry should not appear in the Navigation menu.

### Actual behaviour
Summary/Confirmation link is displayed in the navigation menu

Open census household
Answer No to first question
Summary/Confirmation link displayed in Navigation

### Screenshot
image

samiwel:
Fixes #1043
Fixes the issue by ensuring that the summary / confirmation blocks are on the routing path before displaying the navigation link.

### How to review 
Re-test the defect.

### DIFF
--- a/app/questionnaire/navigation.py
+++ b/app/questionnaire/navigation.py
@@ -38,26 +38,30 @@ class Navigation(object):
         last_group_id = SchemaHelper.get_last_group_id(self.survey_json)
 
         for group in filter(lambda x: 'hide_in_navigation' not in x, self.survey_json['groups']):
-            first_location = Location(group['id'], 0, group['blocks'][0]['id'])
-
             logger.debug("building frontend navigation", group_id=group['id'])
 
-            skip_group = self._should_skip_group(current_group_instance, group)
-            last_block_in_current_group = group['blocks'][-1]
-            is_summary_or_confirmation = last_block_in_current_group and \
-                last_block_in_current_group['type'] in ('Summary', 'Confirmation')
+            first_location = Location(group['id'], 0, group['blocks'][0]['id'])
+            last_block_in_group = SchemaHelper.get_last_block_in_group(group)
 
+            is_summary_or_confirm_group = last_block_in_group and \
+                SchemaHelper.is_summary_or_confirmation(last_block_in_group)
+
+            last_block_location = Location(group['id'], 0, last_block_in_group['id'])
+            can_get_to_summary = is_summary_or_confirm_group and self.routing_path and \
+                last_block_location in self.routing_path
+
+            skip_group = self._should_skip_group(current_group_instance, group)
             if not skip_group:
                 repeating_rule = SchemaHelper.get_repeat_rule(group)
                 if repeating_rule:
                     logger.debug("building repeating navigation", group_id=group['id'])
                     navigation.extend(self._build_repeating_navigation(repeating_rule, group, current_group_id,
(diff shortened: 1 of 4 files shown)


tests/app/questionnaire/test_navigation.py
```diff
@@ -1039,3 +1039,58 @@ def test_build_navigation_submit_answers_link_not_visible_when_no_completed_bloc
...
+        confirmation_link = {
+            'link_name': 'Summary',
+            'highlight': False,
+            'repeating': False,
+            'completed': False,
+            'link_url': Location('summary-group', 0, 'summary').url(metadata)
+        }
+
+        navigation_links = navigation.build_navigation('property-details', 0)
+        self.assertNotIn(confirmation_link, navigation_links)
```
benjefferies, line 1096: Worth adding a check for number of nav links

benjefferies (changes requested):

collisdigital (approved):
Tested against UKIS, MWSS and Census; working as expected.

benjefferies (approved):

app/questionnaire/navigation.py
```diff
@@ -44,20 +44,22 @@ def build_navigation(self, current_group_id, current_group_instance):
 
             skip_group = self._should_skip_group(current_group_instance, group)
             last_block_in_current_group = group['blocks'][-1]
```
LJBabbage: I know this would be from the original PR but think as its a for loop for 'group' we don't need to define _current_group, additionally we are putting things like group['blocks'][-1] in SchemaHelper (it might already be in there).

app/questionnaire/navigation.py
```diff
@@ -44,20 +44,22 @@ def build_navigation(self, current_group_id, current_group_instance):
...
-            is_summary_or_confirmation = last_block_in_current_group and \
+            is_summary_or_confirm_group = last_block_in_current_group and \
                 last_block_in_current_group['type'] in ('Summary', 'Confirmation')
+            can_get_to_summary = is_summary_or_confirm_group and self.routing_path and \
+                Location(group['id'], 0, last_block_in_current_group['id']) in self.routing_path
```
LJBabbage: Think we need to extract the Location(group['id'], 0, last_block_in_current_group['id']) out like we do above for first_location, it will makes it more consistent and easier to read, it will also make the if statement nicer

LJBabbage (changes requested):

LJBabbage:
Also think a reorder makes it easy to read

```
 for group in filter(lambda x: 'hide_in_navigation' not in x, self.survey_json['groups']):
            
            logger.debug("building frontend navigation", group_id=group['id'])
            
            first_block_location = Location(group['id'], 0, group['blocks'][0]['id'])
            last_block = SchemaHelper.get_last_block(self.survey_json)
            last_block_location = Location(group['id'], 0, last_block['id'])

            is_summary_or_confirm_group = last_block and last_block['type'] in ('Summary', 'Confirmation')
            can_get_to_summary = is_summary_or_confirm_group and self.routing_path and \
                                 last_block_location in self.routing_path
            
            skip_group = self._should_skip_group(current_group_instance, group)
            
            if not skip_group:
```

LJBabbage (approved):

PR #1048, opened 2017-03-20
CSRF attack should not save answers on page

benjefferies:
[trello](https://trello.com/c/VLUhPJEE/1006-csrf-behaviour)

If a csrf token is invalid log the user out and do not save the answers.

### How to review 
Tests in test_questionnaire_csrf.py demonstrate how it works.

To manually test
1. Open a questionnaire
1. Open chrome developer console
1. Search for csrf_token
1. Change csrf_token
1. Submit page

Expect to be signed out

### DIFF
--- a/app/views/questionnaire.py
+++ b/app/views/questionnaire.py
@@ -11,6 +11,7 @@ from flask_login import login_required
 from flask_login import logout_user
 from flask_themes2 import render_theme_template
 from structlog import get_logger
+from werkzeug.exceptions import Forbidden
 
 from app import settings
 from app.authentication.session_storage import session_storage
@@ -117,7 +118,7 @@ def post_block(eq_id, form_type, collection_id, group_id, group_instance, block_
 
     if 'action[save_sign_out]' in request.form:
         return _save_sign_out(collection_id, eq_id, form_type, current_location, form)
-    elif not form.validate():
+    elif _is_invalid_form(form):
         context = {'form': form, 'block': block}
         return _build_template(current_location, context, template=block['type'], routing_path=full_routing_path)
     else:
@@ -155,7 +156,7 @@ def post_household_composition(eq_id, form_type, collection_id, group_id):  # py
         remove_empty_household_members_from_answer_store(answer_store, group_id)
         return response
 
-    if not form.validate() or 'action[add_answer]' in request.form or 'action[remove_answer]' in request.form:
+    if _is_invalid_form(form) or 'action[add_answer]' in request.form or 'action[remove_answer]' in request.form:
         context = {'form': form, 'block': block}
         return _build_template(current_location, context, template='questionnaire')
     else:
@@ -260,7 +261,7 @@ def _save_sign_out(collection_id, eq_id, form_type, this_location, form):
     questionnaire_store = get_questionnaire_store(current_user.user_id, current_user.user_ik)
     block_json = _render_schema(this_location)
 
-    if not form.validate():
+    if _is_invalid_form(form):
         content = {'block': block_json, 'form': form}
         return _build_template(this_location, content, template=block_json['type'])
     else:
@@ -376,6 +377,15 @@ def update_questionnaire_store_with_answer_data(questionnaire_store, location, a
 
     if location not in questionnaire_store.completed_blocks:
         questionnaire_store.completed_blocks.append(location)
+
+
+def _is_invalid_form(form):
+    is_valid = form.validate()
+    if form.csrf_token.errors:
+        logger.warn('Invalid csrf token')
+        raise Forbidden
+
+    return not is_valid
 
 
 def _delete_user_data():
--- a/tests/integration/questionnaire/test_questionnaire_csrf.py
+++ b/tests/integration/questionnaire/test_questionnaire_csrf.py
@@ -1,27 +1,121 @@
+import json
+
 from tests.integration.integration_test_case import IntegrationTestCase
 
 
-class TestQuestionnaireInterstitial(IntegrationTestCase):
+class TestQuestionnaireCsrf(IntegrationTestCase):
 
-    def setUp(self):
-        super().setUp()
+    def test_given_on_interstitial_page_when_submit_with_no_csrf_token_then_forbidden(self):
+        # Given
         self.launchSurvey('test', 'interstitial_page')
-        self.assertInPage('Your response is legally required')
-        self.first_url = self.last_url
-
-    def test_form_not_processed_with_no_csrf_token(self):
         self.last_csrf_token = None
-        self.post(action='start_questionnaire')
-        self.assertStatusOK()
-        self.assertEqualUrl(self.first_url)
 
-    def test_form_not_processed_with_incorrect_csrf_token(self):
+        # When
+        self.post(action='start_questionnaire')
+
+        # Then
+        self.assertStatusForbidden()
+        self.assertEqualUrl(self.last_url)
+
+    def test_given_on_interstitial_page_when_submit_with_invalid_csrf_token_then_forbidden(self):
+        # Given
+        self.launchSurvey('test', 'interstitial_page')
         self.last_csrf_token = 'made-up-token'
-        self.post(action='start_questionnaire')
-        self.assertStatusOK()
-        self.assertEqualUrl(self.first_url)
 
-    def test_form_processed_with_correct_csrf_token(self):
+        # When
         self.post(action='start_questionnaire')
+
+        # Then
+        self.assertStatusForbidden()
+        self.assertEqualUrl(self.last_url)
+
+    def test_given_on_introduction_page_when_submit_valid_token_then_redirect_to_next_page(self):
+        # Given
+        self.launchSurvey('test', 'interstitial_page')
+
+        # When
+        self.post(action='start_questionnaire')
+
+        # Then
         self.assertStatusOK()
         self.assertInPage('What is your favourite breakfast food')
+
+    def test_given_answered_question_when_change_answer_with_invalid_csrf_token_then_answers_not_saved(self):
+        # Given
+        self.launchSurvey('test', 'interstitial_page', roles=['dumper'])
+        self.post()
+        self.post({'favourite-breakfast': 'Muesli'})
+
+        # When
+        self.last_csrf_token = 'made-up-token'
+        self.post({'favourite-breakfast': 'Pancakes'})
+
+        # Then
+        self.assertStatusForbidden()
+        self.get('/dump/answers')
+        answers = json.loads(self.getResponseData())
+        self.assertEqual('Muesli', answers['answers'][0]['value'])
+
+    def test_given_valid_answers_on_household_composition_when_answer_with_invalid_csrf_token_then_answers_not_saved(self):
+        # Given
+        self.launchSurvey('census', 'household', roles=['dumper'])
+        post_data = {'first_name': 'Joe'}
+
+        # When
+        self.last_csrf_token = 'made-up-token'
+        self.post(url='/questionnaire/census/household/789/who-lives-here/0/household-composition', post_data=post_data)
+
+        # Then
+        self.assertStatusForbidden()
+        self.get('/dump/answers')
+        answers = json.loads(self.getResponseData())
+        self.assertEqual(0, len(answers['answers']))
+
+    def test_given_valid_answers_when_save_and_sign_out_with_invalid_csrf_token_then_answers_not_saved(self):
+        # Given
+        self.launchSurvey('test', 'interstitial_page', roles=['dumper'])
+        self.post()
+        post_data = {'favourite-breakfast': 'Muesli'}
+
+        # When
+        self.last_csrf_token = 'made-up-token'
+        self.post(post_data=post_data, action='save_sign_out')
+
+        # Then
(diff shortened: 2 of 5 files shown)


jonnyshaw89:
@benjefferies Do we want to log the user out?

tests/integration/questionnaire/test_questionnaire_csrf.py
```diff
@@ -8,38 +10,97 @@ class TestQuestionnaireInterstitial(IntegrationTestCase):
...
         post_data = {
-            'action[start_questionnaire]': 'Start survey'
         }
 
         resp = self.get_and_post_with_csrf_token(self.BASE_URL + 'favourite-foods/0/introduction', post_data)
 
         self.assertEquals(302, resp.status_code)
 
+    def test_given_valid_answers_when_posting_invalid_csrf_token_then_answers_not_saved(self):
+        # Given
```
collisdigital: :-1: for tests in existing format; I've seen the light!

collisdigital: Not that there's anything to be done until the integration test PR merges; I'll re-base (the other PR) and update this when it merges to the new format.

collisdigital:
As a general service comment I think we need to prioritise the work to put the log me back in on the signed-out page as we keep adding to the number of ways a user can get logged out with now easy way for them to get back in

collisdigital:
I've commented on this https://trello.com/c/CDmNd8hd/901-respondent-journey

benjefferies:
I agree it will be good to review the user journey to reflect on how it currently is and talk through approaches of improving it. 

It's worth baring in mind that this scenario will be the minority as you'd not imagine there would be many users targeted with csrf but if one does it would be unclear at the moment of what went wrong and how to recover.

collisdigital:
I recommend we ignore the POST if invalid and re-direct to a 403 Forbidden on the current URL; if they refresh they'll recover back to the empty/last state of the page. This also means if they have an existing tab/window open that /is/ logged in they can continue there without having to sign-in again. 

As part of the Respondent Journey the error pages can be reviewed.

collisdigital:
I've added A/C to https://trello.com/c/VLUhPJEE/1006-csrf-behaviour

benjefferies:
It's worth discussing that point with the security team as Dan had suggested logging the user out. Not sure if he's identified a session hijacking scenario that the log out suggestion was to protect. His suggestion was to: `Options: Destroy the session; display an error age from which the valid user cannot progress.`

collisdigital:
I've also added some scenarios to the card; really this card shouldn't have come into development until these were agreed...

collisdigital:
I've discussed with Dan, he's happy with the scenarios I've added.

tests/integration/questionnaire/test_questionnaire_csrf.py
```diff
@@ -1,27 +1,124 @@
...
+
+        # Then
+        self.assertStatusForbidden()
+        self.get('/dump/answers')
+        answers = json.loads(self.last_response.get_data(True))
+        self.assertEqual('Muesli', answers['answers'][0]['value'])
+
+    def test_given_valid_answers_on_household_composition_when_answer_with_invalid_csrf_token_then_answers_not_saved(self):
+        # Given
+        self.launchSurvey('census', 'household', roles=['dumper'])
```
collisdigital, line 61: Any reason to use a different survey; in that this overrides what happens in `setUp` above?

tests/integration/questionnaire/test_questionnaire_csrf.py
```diff
@@ -1,27 +1,124 @@
...
-    def test_form_processed_with_correct_csrf_token(self):
+    def test_given_on_introduction_page_when_submit_valid_token_then_redirect_to_next_page(self):
         self.post(action='start_questionnaire')
         self.assertStatusOK()
         self.assertInPage('What is your favourite breakfast food')
+
+    def test_given_answered_question_when_change_answer_with_invalid_csrf_token_then_answers_not_saved(self):
+        # Given
+        self.post()
+        post_data = {
```
collisdigital: I had been in-lining these to keep the verbosity down e.g.

`self.post({'favourite-breakfast': 'Muesli'})`

tests/integration/questionnaire/test_questionnaire_csrf.py
```diff
@@ -1,27 +1,124 @@
...
+            'favourite-breakfast': 'Pancakes'
+        }
+        self.last_csrf_token = 'made-up-token'
+        self.post(post_data=post_data)
+
+        # Then
+        self.assertStatusForbidden()
+        self.get('/dump/answers')
+        answers = json.loads(self.last_response.get_data(True))
```
collisdigital: There is a helper for this `self.getResponseData()`

tests/integration/questionnaire/test_questionnaire_csrf.py
```diff
@@ -1,27 +1,124 @@
...
+    def test_given_valid_answers_on_household_composition_when_answer_with_invalid_csrf_token_then_answers_not_saved(self):
+        # Given
+        self.launchSurvey('census', 'household', roles=['dumper'])
+        post_data = {
+            'first_name': 'Joe'
+        }
+
+        # When
+        self.last_csrf_token = 'made-up-token'
+        self.post(url='/questionnaire/census/household/789/who-lives-here/0/household-composition', post_data=post_data)
```
collisdigital: If we can avoid putting URLs in it makes the tests more resilient to changes in future, if you use the survey launched in `setUp()` you can also use `self.first_url` or record a URL after a post/get from `self.last_url` to re-submit to the same page.

tests/integration/questionnaire/test_questionnaire_csrf.py
```diff
@@ -1,27 +1,124 @@
...
+
+        # Then
+        self.assertStatusForbidden()
+        self.get('/dump/answers')
+        answers = json.loads(self.last_response.get_data(True))
```
collisdigital: There is a helper for this `self.getResponseData()` wonder if making last_response private is better (not sure if any other tests access it directly)

tests/integration/questionnaire/test_questionnaire_csrf.py
```diff
@@ -1,27 +1,124 @@
...
+        self.assertEqual(0, len(answers['answers']))
+
+    def test_given_valid_answers_when_save_and_sign_out_with_invalid_csrf_token_then_answers_not_saved(self):
+        # Given
+        self.post()
+        post_data = {
+            'favourite-breakfast': 'Muesli',
+            'action[save_sign_out]': ''
```
collisdigital: you can pass the action to `post()` e.g.

`self.post({'favourite-breakfast': 'Muesli'}, action='save_sign_out')`

tests/integration/questionnaire/test_questionnaire_csrf.py
```diff
@@ -1,27 +1,124 @@
...
+        }
+
+        # When
+        self.last_csrf_token = 'made-up-token'
+        self.post(post_data=post_data)
+
+        # Then
+        self.assertStatusForbidden()
+        self.get('/dump/answers')
+        answers = json.loads(self.last_response.get_data(True))
```
collisdigital: Here too

tests/integration/questionnaire/test_questionnaire_csrf.py
```diff
@@ -1,27 +1,124 @@
...
+        self.assertEqual(0, len(answers['answers']))
+
+    def test_given_csrf_attack_when_submit_new_answers_then_answers_saved(self):
+        # Given
+        self.post()
+        post_data = {
+            'favourite-breakfast': 'Pancakes',
+            'csrf_token': 'made-up-token'
```
collisdigital: align with other tests; use `self.last_csrf_token`

tests/integration/questionnaire/test_questionnaire_csrf.py
```diff
@@ -1,27 +1,124 @@
...
+        # When
+        post_data = {
+            'favourite-breakfast': 'Pancakes'
+        }
+        self.post(post_data=post_data)
+
+        # Then
+        self.assertStatusOK()
+        self.get('/dump/answers')
+        answers = json.loads(self.last_response.get_data(True))
```
collisdigital: Here too

tests/integration/questionnaire/test_questionnaire_csrf.py
benjefferies: Done

tests/integration/questionnaire/test_questionnaire_csrf.py
benjefferies: Done

tests/integration/questionnaire/test_questionnaire_csrf.py
benjefferies: The household composition question has a bespoke end point so ensuring there isn't a csrf bug or regression there

tests/integration/questionnaire/test_questionnaire_csrf.py
benjefferies: This page is a few blocks into the questionnaire so don't think it's worth progressing through the questionnaire just to post to this question

tests/integration/questionnaire/test_questionnaire_csrf.py
benjefferies: Done

tests/integration/questionnaire/test_questionnaire_csrf.py
benjefferies: Done

tests/integration/questionnaire/test_questionnaire_csrf.py
benjefferies: Done

tests/integration/questionnaire/test_questionnaire_csrf.py
```diff
@@ -1,27 +1,105 @@
...
+
+    def test_given_csrf_attack_when_submit_new_answers_then_answers_saved(self):
+        # Given
+        self.post()
+        self.last_csrf_token = 'made-up-token'
+        self.post({'favourite-breakfast': 'Pancakes'})
+
+        # When
+        self.get(self.last_url)
+        self.post({'favourite-breakfast': 'Pancakes'})
```
collisdigital: Would be safer to use a value other than Pancakes here?

benjefferies: Yeah think it was muesli and pancake but maybe copy and paste errror

tests/integration/questionnaire/test_questionnaire_csrf.py
collisdigital: Maybe put this in a separate test class then as this test is doing a load of stuff in setUp and the overwriting it, I guess that's not a major issue just seems wrong

benjefferies: Not sure if the set up is really necessary as each test could do it's own set up. I think the introduction check is probably redundant as it would be annoying if all tests fail for an unrelated issue. Also I don't think repetitiveness in tests are that bad and sometimes can help with the readability of the scenario as each test is essentially the scenario

collisdigital: Agreed, ditch the setUp

collisdigital: Also only some tests need the dumper roles if you remove the common setup

tests/integration/questionnaire/test_questionnaire_csrf.py
```diff
@@ -1,27 +1,105 @@
+import json
+
 from tests.integration.integration_test_case import IntegrationTestCase
 
 
 class TestQuestionnaireInterstitial(IntegrationTestCase):
```
jonnyshaw89: This needs to be updated from the copy/paste?

app/views/questionnaire.py
```diff
@@ -378,6 +379,15 @@ def update_questionnaire_store_with_answer_data(questionnaire_store, location, a
         questionnaire_store.completed_blocks.append(location)
 
 
+def _is_invalid_form(form):
+    is_valid = form.validate()
+    if form.csrf_token.errors:
+        logger.warn('Invalid csrf token')
+        raise Forbidden
```
jonnyshaw89, line 386: This exception doesn't appear to be caught by our errors view to render our happy error pages.

jonnyshaw89 (changes requested):

tests/integration/questionnaire/test_questionnaire_csrf.py
benjefferies: Done

tests/integration/questionnaire/test_questionnaire_csrf.py
benjefferies: Done

app/views/questionnaire.py
benjefferies: Good spot, done

collisdigital (approved):

jonnyshaw89 (approved):

collisdigital:
Fix failing tests and squash, then it's ready.

PR #1049, opened 2017-03-22
Replace git ref

LJBabbage:
Remove all references to EQ_GIT_REF and inject a plaintext file into the root of the app at build time

N.B Jenkins will need to be updated after this card has been merged

### How to review 
Make sure every is working as normal


jonnyshaw89:
@LJBabbage if you look at the codecov report. There is some more code you can remove

LJBabbage:
Due to line number changes the patch is moaning about things that haven't changed

app/__init__.py
```diff
@@ -214,28 +191,24 @@ def setup_babel(application):
     application.jinja_env.add_extension('jinja2.ext.i18n')
 
 
-def add_health_check(application, headers):
-    from healthcheck import HealthCheck
```
collisdigital, line 218: Does this mean a package can be removed from requirements.txt now?

app/settings.py
```diff
@@ -27,6 +27,13 @@ def get_key(_key_name):
     else:
         return None
 
+
+def get_application_version_from_file():
```
collisdigital: I think there's already a function below that does this (although it might be key related) possible refactor to use a common one?

app/__init__.py
LJBabbage: Yep, has been removed

app/settings.py
```diff
@@ -18,11 +18,11 @@ def parse_mode(string):
...
 
 
-def get_key(_key_name):
-    if _key_name:
-        logger.debug("reading key from file", filename=_key_name)
-        key = open(_key_name, 'r')
-        contents = key.read()
+def read_file(file_name):
+    if os.path.isfile(file_name):
+        logger.error("reading from file", filename=file_name)
```
benjefferies: I guess the error level is an accident and should probably be debug

app/settings.py
```diff
@@ -18,11 +18,11 @@ def parse_mode(string):
...
+        f = open(file_name, "r")
+        contents = f.read().strip()
```
benjefferies: Do we need to strip the text read in?

benjefferies (changes requested):

app/settings.py
```diff
@@ -18,11 +18,11 @@ def parse_mode(string):
     return string.upper() != 'FALSE'
 
 
-def get_key(_key_name):
-    if _key_name:
-        logger.debug("reading key from file", filename=_key_name)
-        key = open(_key_name, 'r')
-        contents = key.read()
+def read_file(file_name):
```
collisdigital, line 21: :+1:

benjefferies (approved):
Good job on removing the redundant code

collisdigital (approved):
Squash and rebase then this is ready.

PR #1058, opened 2017-03-24
Non mandatory date of birth field retains date

LJBabbage:
Non mandatory date of birth field retains date even after respondent deletes the date. A check to see if the dict keys all contain empty values, if they do we can remove the answer from the store

### How to review 
Using test_dates.json, fill out all of the questionnaire, including the last question. Check the summary page is right. Using previous, change the answer to the last question, taking out all values. Submit and make sure the summary screen changes to 'No answer provided'. Hit previous again and make sure the last question is still empty.

### DIFF
--- a/app/views/questionnaire.py
+++ b/app/views/questionnaire.py
@@ -338,34 +338,46 @@ def update_questionnaire_store_with_form_data(questionnaire_store, location, ans
 
             # Dates are comprised of 3 string values
             if isinstance(answer_value, dict):
-                is_day_month_year = 'day' in answer_value and 'month' in answer_value and 'year' in answer_value
-                is_month_year = 'day' not in answer_value and 'year' in answer_value and 'month' in answer_value
-
-                if is_day_month_year and answer_value['day'] and answer_value['month']:
-                    date_str = "{:02d}/{:02d}/{}".format(
-                        int(answer_value['day']),
-                        int(answer_value['month']),
-                        answer_value['year'],
-                    )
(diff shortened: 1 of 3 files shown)


yoganandkunche:
Tested on branch and is working as expected.

app/views/questionnaire.py
```diff
@@ -337,28 +337,16 @@ def update_questionnaire_store_with_form_data(questionnaire_store, location, ans
...
-                    date_str = "{:02d}/{:02d}/{}".format(
-                        int(answer_value['day']),
-                        int(answer_value['month']),
-                        answer_value['year'],
-                    )
-                    answer = Answer(answer_id=answer_id, value=date_str, location=location)
-                elif is_month_year and answer_value['month']:
-                    date_str = "{:02d}/{}".format(int(answer_value['month']), answer_value['year'])
-                    answer = Answer(answer_id=answer_id, value=date_str, location=location)
+                if answer_value_empty(answer_value):
```
benjefferies, line 341: Maybe it would be better to do this in `update_questionnaire_store_with_form_data` before it's added

benjefferies (changes requested):

app/views/questionnaire.py
LJBabbage: It is in update_questionnaire_store_with_form_data and is before its stored ????

benjefferies (approved):

jonnyshaw89 (approved):

PR #1059, opened 2017-03-27
Escaping quotes and backslash in answers

### LINKED ISSUE #1060: Error 500 generated on names containing double quote
Observed in pre-prod 21/03/17 with the Census on the household composition page with the following fields: First name; Middle name; Last name. Suspected on other free-text fields.

This generates an HTTP 500 with the message "Sorry, something has gone wrong".

Questions: Should this be allowed? Is the double quote valid in someone's name?

If not, could this be handled gracefully with the standard validation message prompting for correction?

benjefferies:
Fixes #1060 

We use jinja to inject strings into the questionnaire schemas however if
quotes or backslashes are injected it will break.

Quotes nor backslashes can be in JSON without being escaped. Both could
get into the JSON via user answers if they are not escaped before hand.

This fix is a minimal fix for the problem described.

### How to review 
1. Load census
1. Try combinations of special characters on household composition screen

Suggest better alternatives

### DIFF
--- a/app/templating/schema_context.py
+++ b/app/templating/schema_context.py
@@ -25,7 +25,10 @@ def _map_alias_to_answers(aliases, answer_store):
         matching_answers = []
 
         answers = answer_store.filter(answer_id=answer_info['answer_id'], limit=settings.EQ_MAX_NUM_REPEATS)
-        matching_answers.extend([answer['value'] for answer in answers])
+
+        for answer in answers:
+            safe_answer = json_safe(answer['value'])
+            matching_answers.append(safe_answer)
 
         number_of_matches = len(matching_answers)
 
@@ -52,7 +55,14 @@ def _build_exercise(metadata):
 
 def _build_respondent(metadata):
     return {
-        "ru_name": metadata["ru_name"],
-        "trad_as": metadata["trad_as"],
-        "trad_as_or_ru_name": metadata["trad_as"] or metadata["ru_name"],
+        "ru_name": json_safe(metadata["ru_name"]),
+        "trad_as": json_safe(metadata["trad_as"]),
+        "trad_as_or_ru_name": json_safe(metadata["trad_as"] or metadata["ru_name"]),
     }
+
+
+def json_safe(data):
+    if isinstance(data, str):
+        return data.replace('\\', r'\\').replace('"', r'\"')
+    else:
+        return data
--- a/tests/app/templating/test_schema_context.py
+++ b/tests/app/templating/test_schema_context.py
@@ -5,7 +5,7 @@ from app.templating.schema_context import build_schema_context
 from tests.app.framework.survey_runner_test_case import SurveyRunnerTestCase
 
 
-class TestSchemaContext(SurveyRunnerTestCase):
+class TestSchemaContext(SurveyRunnerTestCase):  # pylint: disable=too-many-public-methods
 
     metadata = {
         'return_by': '2016-10-10',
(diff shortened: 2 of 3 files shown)


app/templating/schema_context.py
```diff
@@ -56,3 +59,7 @@ def _build_respondent(metadata):
         "trad_as": metadata["trad_as"],
         "trad_as_or_ru_name": metadata["trad_as"] or metadata["ru_name"],
     }
+
+
+def json_safe(data):
+    return data.replace('\\', r'\\').replace('"', r'\"')
```
collisdigital: Did we consider using Bleach again (https://pypi.python.org/pypi/bleach) ?

benjefferies: Yes I looked to see if bleach could resolve the issue but it wouldn't make the data "json safe"

collisdigital: Ok, what about `json.dumps()` (http://stackoverflow.com/a/5997098) ?

collisdigital: Just thinking it would be good to sanitise generally rather than specifically for this case to catch any other issues outside of quotes we might hit.

benjefferies: Yeah fully understand. I'm not satisfied with the solution but cannot think of any other way without rewriting how we do string injection. Regarding the `json.dumps` we would have to do `answer = json.dumps(answer)[1:-1] if answer else answer` as it adds in extra quotes

collisdigital: Ok 👌 that's nasty; stick with what you have

yoganandkunche:
Tested on branch works as expected for First name Last name and middle name.

But fails if the name of the company/trading in as have a " or \ (example: Apple & "Co" )

jonnyshaw89 (dismissed):

app/templating/schema_context.py
```diff
@@ -55,8 +55,8 @@ def _build_exercise(metadata):
 
 def _build_respondent(metadata):
     return {
-        "ru_name": metadata["ru_name"],
-        "trad_as": metadata["trad_as"],
+        "ru_name": json_safe(metadata["ru_name"]),
+        "trad_as": json_safe(metadata["trad_as"]),
         "trad_as_or_ru_name": metadata["trad_as"] or metadata["ru_name"],
```
jonnyshaw89: what about this field?

jonnyshaw89 (changes requested):

app/templating/schema_context.py
benjefferies: Done 👍

app/templating/schema_context.py
```diff
@@ -52,7 +55,14 @@ def _build_exercise(metadata):
 
 def _build_respondent(metadata):
     return {
-        "ru_name": metadata["ru_name"],
-        "trad_as": metadata["trad_as"],
-        "trad_as_or_ru_name": metadata["trad_as"] or metadata["ru_name"],
+        "ru_name": json_safe(metadata["ru_name"]),
+        "trad_as": json_safe(metadata["trad_as"]),
+        "trad_as_or_ru_name": json_safe(metadata["trad_as"] or metadata["ru_name"]),
```
jonnyshaw89: but not this one? `metadata["ru_name"]`?

jonnyshaw89 (approved):

collisdigital:
Can this be re-based

app/templating/schema_context.py
```diff
@@ -25,7 +25,10 @@ def _map_alias_to_answers(aliases, answer_store):
         matching_answers = []
 
         answers = answer_store.filter(answer_id=answer_info['answer_id'], limit=settings.EQ_MAX_NUM_REPEATS)
-        matching_answers.extend([answer['value'] for answer in answers])
+
+        for answer in answers:
+            answer = json_safe(answer['value'])
```
LJBabbage: Not a fan of overriding 'answer', can we use safe_answer or safe_value 
```
for answer in answers:
   safe_value = json_safe(answer['value'])
   matching_answers.append(safe_value)

File: app/templating/schema_context.py (hunk 3 above)
Comment (LJBabbage) [inline app/templating/schema_context.py]: json_safe(metadata["ru_name"]) ?

File: tests/app/templating/test_schema_context.py [hunk 5]
```diff
@@ -233,3 +233,112 @@ def test_respondent_display_name_is_ru_name_as_when_trading_as_empty(self):
         schema_context = build_schema_context(metadata, {}, self.answer_store)
 
         self.assertEqual(schema_context['respondent']['trad_as_or_ru_name'], metadata['ru_name'])
+
```
LJBabbage, line 236: Do we really need to test json_safe on every one(ru_name, trade_as etc) surely we just need to test json_safe with one example

tests/integration/questionnaire/test_questionnaire_piping.py
```diff
@@ -0,0 +1,30 @@
...
+        self.assertStatusOK()
+        self.assertInPage('Joe Bloggs "Junior"')
+
+    def test_given_backslash_in_answer_when_piped_into_page_then_backslash_on_page(self):
+        # Given
+        self.launchSurvey('census', 'household')
+        self.post({'permanent-or-family-home-answer': 'Yes'})
+
+        # When
+        self.post({'household-0-first-name': 'Joe Bloggs\\John Doe'})
```
LJBabbage, line 25: Double back slash?

LJBabbage (changes requested):

tests/app/templating/test_schema_context.py
benjefferies: What if someone removes the json_safe from one of the 'names'? I guess that issue would regress?

app/templating/schema_context.py
benjefferies: Done

tests/integration/questionnaire/test_questionnaire_piping.py
benjefferies: Escaped backslash = single backslash

LJBabbage (approved):

LJBabbage:
Rebase and squash please

PR #1067, opened 2017-03-29
added unit test for coverage

dcdarrell9:
Added a unit test to add code coverage in views/dev_mode

### How to review 
See whether the coverage is right, and advise on how to get 100% coverage

### DIFF
--- a/tests/app/views/test_dev_mode.py
+++ b/tests/app/views/test_dev_mode.py
@@ -1,6 +1,6 @@
 from app import settings
 from tests.integration.integration_test_case import IntegrationTestCase
-from app.views.dev_mode import *
+from app.views.dev_mode import extract_eq_id_and_form_type
 
 
 class TestDevMode(IntegrationTestCase):
--- a/tests/app/views/test_dev_mode.py
+++ b/tests/app/views/test_dev_mode.py
@@ -1,5 +1,6 @@
 from app import settings
 from tests.integration.integration_test_case import IntegrationTestCase
+from app.views.dev_mode import *
 
 
 class TestDevMode(IntegrationTestCase):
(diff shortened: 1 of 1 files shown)


tests/app/views/test_dev_mode.py
```diff
@@ -1,5 +1,6 @@
 from app import settings
 from tests.integration.integration_test_case import IntegrationTestCase
+from app.views.dev_mode import *
```
jonnyshaw89: Be more specific on your imports

tests/app/views/test_dev_mode.py
```diff
@@ -43,3 +44,10 @@ def test_dev_flush_mode_submission(self):
                       'collection_exercise_sid': '789',
                       'ru_ref': '12346789012A'})
         self.assertStatusNotFound()
+
+    def test_extract_eq_id_and_form_type_no_form_type(self):
+        #Given
```
collisdigital, line 49: Probably not worth having this comment here really

tests/app/views/test_dev_mode.py
```diff
@@ -43,3 +44,10 @@ def test_dev_flush_mode_submission(self):
...
+        data_to_extract = extract_eq_id_and_form_type('rogueone.json')
+
+        self.assertEqual(data_to_extract, ('rogueone', '-1'))
```
collisdigital, line 52: Do we also need a test for when there is an id value e.g. 'census_household.json' should extract to ('household', 'census') ?

dcdarrell9: There could be plenty of tests involved, it depends on how many you want written? The integration tests cover a lot of the code

collisdigital: It's worth reading up about the testing pyramid (e.g. http://www.agilenutshell.com/episodes/41-testing-pyramid and http://www.velocitypartners.net/blog/2014/01/28/agile-testing-the-agile-test-automation-pyramid/

In summary there should be more test coverage at the unit test level than component or UI; as these tests tend to be easier to write, faster to run and less likely to break. Worth chatting about this some more.

PR #1069, opened 2017-03-30
eq 1053 code coverage dev mode

dcdarrell9:
Changed how the function works (added regex) and added units tests in to make sure all input/output is right.

### How to review 
Check whether the function still works in the right way.
Check through to see if there is a more efficient or better way, and run the unit tests.


app/views/dev_mode.py
```diff
@@ -91,19 +92,14 @@ def dev_flush_mode_get():
 
 def extract_eq_id_and_form_type(schema_name):
     logger.debug("extracting eq_id and form type from schema name", schema_name=schema_name)
-    if "_" in schema_name:
-        split_schema_name = schema_name.split("_", 1)
-        if len(split_schema_name) != 2:
-            raise ValueError("Schema file name incorrect %", schema_name)
+    if re.search(r"^([a-z0-9]+)_(\w+)\.json$", schema_name):
+        split_schema_name = schema_name.split("_", maxsplit=1)
```
collisdigital: Looking good, would be better to use re.match here then you can extract the values that matched directly e.g.

```python
match = re.match(r"^([a-z0-9]+)_(\w+)\.json$", schema_name)
if match:
    eq_id = match.group(1)
    form_type = match.group(2)
```

collisdigital (changes requested):

collisdigital:
You need to run the linting locally and fix the errors:

`./scripts/run_lint.sh`

This is what Travis is failing on above (https://travis-ci.org/ONSdigital/eq-survey-runner/builds/216774124)

app/views/dev_mode.py
```diff
@@ -91,19 +92,12 @@ def dev_flush_mode_get():
...
-        form_type = split_rest_of_name[0]
+    match = re.match(r"^([a-z0-9]+)_(\w+)\.json$", schema_name)
+    if match:
+        eq_id = match.group(1)
+        form_type = match.group(2)
     else:
-        # No form type associated with
-        eq_id = schema_name.split(".", 1)[0]
-        form_type = "-1"
+        raise ValueError('Invalid schema format')
```
collisdigital: Last comment, previously this output the filename of the invalid file on error, it would be good to keep that:

`raise ValueError("Schema filename incorrect format: %", schema_name)`

jonnyshaw89: Agree with Chris. This error message changes the context of the error

jonnyshaw89 (changes requested):

collisdigital (approved):

PR #1070, opened 2017-03-30
Add save and continue to summary page

### LINKED ISSUE #1004: TImeout: 'Save and Sign out' button is not working on the summary screens where we don't have 'Save and complete' later link
### Expected behaviour
Must be able to click save and Signout on the timeout pop up.

### Actual behaviour
Save and Signout button does not do anything when the respondent is on a page where there is no 'Save and complete' later link

Launch a survey which has a summary screen
fill the survey and navigate to summary page
allow to time out 
and click on [Save and Sign out]

### Screenshot
screen shot 2017-03-01 at 11 17 12

benjefferies:
Fixes #1004 meaning users will be signed out when timing out on summary.

By adding save and complete later link to the summary this fixes the issue.

### How to review 
See #1004 on how steps to reproduce issue and confirm actual behaviour is expected

### DIFF
--- a/app/helpers/form_helper.py
+++ b/app/helpers/form_helper.py
@@ -84,7 +84,7 @@ def post_form_for_location(block_json, location, answer_store, request_form, err
 
 
 def disable_mandatory_answers(block_json):
-    for section_json in block_json['sections']:
+    for section_json in block_json.get('sections', []):
         for question_json in section_json['questions']:
             for answer_json in question_json['answers']:
                 if 'mandatory' in answer_json and answer_json['mandatory'] is True:
--- a/app/templates/summary.html
+++ b/app/templates/summary.html
@@ -1,3 +1,4 @@
+{% import 'macros/helpers.html' as helpers %}
 {% extends theme('layouts/_onecol.html') -%}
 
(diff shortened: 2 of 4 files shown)


dcdarrell9 (approved):

app/helpers/form_helper.py
```diff
@@ -84,7 +84,7 @@ def post_form_for_location(block_json, location, answer_store, request_form, err
 
 
 def disable_mandatory_answers(block_json):
-    for section_json in block_json['sections']:
+    for section_json in block_json.get('sections', []):
```
jonnyshaw89, line 87: if your changing this for one it makes sense to change for the subsequent for loops as well?

jonnyshaw89 (changes requested):

app/helpers/form_helper.py
benjefferies: Sections are optional but I believe questions/answers are not. It could be changed though if you think it should

jonnyshaw89 (approved):

jonnyshaw89:
@benjefferies squash?

benjefferies:
Rebased

PR #1071, opened 2017-03-30
Move delete from thank you to submit

LJBabbage:
Remove all state changes  on end points (thank you, signed out and session expired) and introduces survey_completed_metadata in cookie so submit data can clear the questionnaire store

### How to review 
1. Make sure you can't jump to the thank-you page without completing a survey first
2. Make sure once you are on the survey page and try to go back it stays on thank-you
3. Make sure session time works as expected
4. Make sure sign-out works as before

### DIFF
--- a/app/assets/js/app/modules/timeout.js
+++ b/app/assets/js/app/modules/timeout.js
@@ -64,6 +64,7 @@ domready(() => {
   const sessionTimeout = window.__EQ_SESSION_TIMEOUT__
   const promptTime = window.__EQ_SESSION_TIMEOUT_PROMPT__
   const sessionExpiredUrl = window.__EQ_SESSION_EXPIRED_URL__
+  const expireSessionUrl = window.__EQ_EXPIRE_SESSION_URL__
   const sessionContinueUrl = window.__EQ_SESSION_CONTINUE_URL__
 
   const continueBtn = document.querySelector('.js-timeout-continue')
@@ -124,7 +125,10 @@ domready(() => {
     let countDown = timeoutUI.onTick()
     if (countDown < 1) {
       window.clearInterval(this)
-      window.location = sessionExpiredUrl
+      fetch(expireSessionUrl, { method: 'POST' })
+        .then(() => {
+          window.location = sessionExpiredUrl
+        })
     }
 
     if (countDown < promptTime) {
--- a/app/authentication/session_storage.py
+++ b/app/authentication/session_storage.py
@@ -8,6 +8,7 @@ from app.data_model.database import db_session
 
 USER_ID = "user_id"
 USER_IK = "user_ik"
+SURVEY_COMPLETED_METADATA = "survey_completed_metadata"
 EQ_SESSION_ID = "eq-session-id"
 
 logger = get_logger()
@@ -30,9 +31,9 @@ class SessionStorage:
             # session has a add function but it is wrapped in a session_scope which confuses pylint
             db_session.add(eq_session)
 
-    def clear(self):
+    def delete_session_from_db(self):
         """
-        Removes a user id from the session
+        deletes user session from database
         """
         if EQ_SESSION_ID in session:
             eq_session_id = session[EQ_SESSION_ID]
@@ -88,6 +89,29 @@ class SessionStorage:
         session[USER_IK] = user_ik
 
     @staticmethod
+    def store_survey_completed_metadata(tx_id, period_str, ru_ref):
+        """
+        Store survey_completed_metadata in the cookie for retrieval later
+        """
+        session[SURVEY_COMPLETED_METADATA] = {
+            'tx_id': tx_id,
+            'period_str': period_str,
+            'ru_ref': ru_ref,
+        }
+
+    @staticmethod
+    def has_survey_completed_metadata():
+        """
+        Checks if a user has survey completed metadata
+        :return: boolean value
+        """
+
+        if SURVEY_COMPLETED_METADATA in session:
+            return session[SURVEY_COMPLETED_METADATA] is not None
+        else:
+            return False
+
+    @staticmethod
     def has_user_ik():
         """
         Checks if a user has a stored ik
@@ -97,6 +121,24 @@ class SessionStorage:
             return session[USER_IK] is not None
         else:
             return False
+
+    def remove_user_ik(self):
+        """
+        Removes a user's ik from the session
+        """
+
+        if self.has_user_ik():
+            session.pop(USER_IK)
+
+    def get_survey_completed_metadata(self):
+        """
+        Retrieves survey completed metadata
+        :return: the survey completed metadata
+        """
+        if self.has_survey_completed_metadata():
+            return session[SURVEY_COMPLETED_METADATA]
+        else:
+            return None
 
     def get_user_ik(self):
         """
--- a/app/views/questionnaire.py
+++ b/app/views/questionnaire.py
@@ -25,11 +25,11 @@ from app.questionnaire.path_finder import PathFinder
 from app.questionnaire.rules import evaluate_skip_condition
 from app.submitter.converter import convert_answers
 from app.submitter.submitter import SubmitterFactory
-from app.templating.metadata_context import build_metadata_context
+from app.templating.metadata_context import build_metadata_context, build_metadata_context_for_survey_completed
 from app.templating.schema_context import build_schema_context
 from app.templating.summary_context import build_summary_rendering_context
 from app.templating.template_renderer import renderer, TemplateRenderer
-from app.utilities.schema import load_schema_from_metadata
+from app.utilities.schema import load_schema_from_metadata, load_schema_from_params
 from app.views.errors import MultipleSurveyError
 
 logger = get_logger()
@@ -41,16 +41,16 @@ questionnaire_blueprint = Blueprint(name='questionnaire',
 
 
 @questionnaire_blueprint.before_request
-@login_required
 def before_request():
-    metadata = get_metadata(current_user)
-    logger.bind(tx_id=metadata['tx_id'])
-    g.schema_json = load_schema_from_metadata(metadata)
     values = request.view_args
     logger.info('questionnaire request', eq_id=values['eq_id'], form_type=values['form_type'],
                 ce_id=values['collection_id'], method=request.method, url_path=request.full_path)
 
-    _check_same_survey(values['eq_id'], values['form_type'], values['collection_id'])
+    metadata = get_metadata(current_user)
+    if metadata:
+        logger.bind(tx_id=metadata['tx_id'])
+        g.schema_json = load_schema_from_metadata(metadata)
+        _check_same_survey(values['eq_id'], values['form_type'], values['collection_id'])
 
 
 @questionnaire_blueprint.after_request
@@ -78,19 +78,18 @@ def get_block(eq_id, form_type, collection_id, group_id, group_instance, block_i
     metadata = get_metadata(current_user)
     answer_store = get_answer_store(current_user)
     path_finder = PathFinder(g.schema_json, answer_store, metadata)
-
     valid_group = group_id in SchemaHelper.get_group_ids(g.schema_json)
     full_routing_path = path_finder.get_routing_path()
     is_valid_location = valid_group and current_location in path_finder.get_routing_path(group_id, group_instance)
     latest_location = path_finder.get_latest_location(get_completed_blocks(current_user), routing_path=full_routing_path)
     if not is_valid_location:
-        return _redirect_to_latest_location(collection_id, eq_id, form_type, latest_location)
+        return _redirect_to_location(collection_id, eq_id, form_type, latest_location)
 
     block = _render_schema(current_location)
     block_type = block['type']
     is_skipping_to_end = block_type in ['Summary', 'Confirmation'] and current_location != latest_location
     if is_skipping_to_end:
-        return _redirect_to_latest_location(collection_id, eq_id, form_type, latest_location)
+        return _redirect_to_location(collection_id, eq_id, form_type, latest_location)
 
     context = _get_context(block, current_location, answer_store)
     return _build_template(current_location, context, template=block_type, routing_path=full_routing_path)
@@ -109,7 +108,7 @@ def post_block(eq_id, form_type, collection_id, group_id, group_instance, block_
     is_valid_location = valid_group and current_location in path_finder.get_routing_path(group_id, group_instance)
     if not is_valid_location:
         latest_location = path_finder.get_latest_location(get_completed_blocks(current_user), routing_path=full_routing_path)
-        return _redirect_to_latest_location(collection_id, eq_id, form_type, latest_location)
+        return _redirect_to_location(collection_id, eq_id, form_type, latest_location)
 
     error_messages = SchemaHelper.get_messages(g.schema_json)
     block = _render_schema(current_location)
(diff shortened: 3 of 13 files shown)


app/views/questionnaire.py
```diff
@@ -216,18 +211,20 @@ def get_timeout_continue(eq_id, form_type, collection_id):  # pylint: disable=un
...
 @login_required
-def get_session_expired(eq_id, form_type, collection_id):  # pylint: disable=unused-argument
+def post_session_expired(eq_id, form_type, collection_id):  # pylint: disable=unused-argument
     session_storage.clear()
     logout_user()
-    theme = g.schema_json.get('theme', None)
-    logger.debug("theme selected", theme=theme)
-    return render_theme_template(theme, template_name='session-expired.html',
-                                 legal_basis=g.schema_json['legal_basis'],
+    return redirect(url_for('.get_session_expired', eq_id=eq_id, form_type=form_type, collection_id=collection_id))
```
jonnyshaw89: instead of redirecting can't you just `return self.get_session_expired(eq_id, form_type, collection_id)`
Given you are essentially redirecting to the same page?

app/views/questionnaire.py
```diff
@@ -272,15 +267,16 @@ def is_survey_completed(answer_store, metadata):
...
     if _is_invalid_form(form):
         content = {'block': block_json, 'form': form}
         return _build_template(this_location, content, template=block_json['type'])
     else:
         _update_questionnaire_store(this_location, form)
-
         if this_location in questionnaire_store.completed_blocks:
             questionnaire_store.completed_blocks.remove(this_location)
+            questionnaire_store.add_or_update()
+        session_storage.clear()
```
jonnyshaw89: personal preference but a line break after the indentation makes it look clearer that the if has finished?

jonnyshaw89 (changes requested):

app/authentication/session_storage.py
```diff
@@ -88,6 +89,32 @@ def store_user_ik(user_ik):
         session[USER_IK] = user_ik
 
     @staticmethod
+    def store_survey_completed_metadata(tx_id, period_str, ru_ref):
+        """
+        Store survey_completed_metadata in the cookie for retrieval later
+        :param tx_id: the transaction id
```
collisdigital: Not sure these `:param` descriptions add anything over the variable names; if they really do then consider just naming the variables more verbose instead.

app/views/questionnaire.py
```diff
@@ -41,16 +42,16 @@
 
 
 @questionnaire_blueprint.before_request
-@login_required
```
collisdigital, line 44: What's the implication of removing `@login_required` here?

app/views/questionnaire.py
```diff
@@ -216,18 +211,20 @@ def get_timeout_continue(eq_id, form_type, collection_id):  # pylint: disable=un
     return 'true'
 
 
-@questionnaire_blueprint.route('session-expired', methods=["GET"])
+@questionnaire_blueprint.route('session-expired', methods=["POST"])
```
collisdigital: Not sure the name of this make sense now; should it be an action like `sign-out` as that's what it's doing?

collisdigital: The sign-out link on GitHub.com is a POST to `/logout`; which also raises the question of why ours is not on the root URL `/`, since our `sign-in` is i.e. `/session?token=....`

app/views/questionnaire.py
collisdigital: See my above comment about renaming this (and possibly moving the endpoint); if that happens it would make sense to re-direct to the expired page.

app/views/questionnaire.py
```diff
@@ -216,18 +211,20 @@ def get_timeout_continue(eq_id, form_type, collection_id):  # pylint: disable=un
...
     session_storage.clear()
     logout_user()
-    theme = g.schema_json.get('theme', None)
-    logger.debug("theme selected", theme=theme)
-    return render_theme_template(theme, template_name='session-expired.html',
-                                 legal_basis=g.schema_json['legal_basis'],
+    return redirect(url_for('.get_session_expired', eq_id=eq_id, form_type=form_type, collection_id=collection_id))
+
+
+@questionnaire_blueprint.route('session-expired', methods=["GET"])
```
collisdigital: As per above; should this endpoint be on the root URL e.g. `/session-expired` ?

app/views/questionnaire.py
```diff
@@ -172,41 +175,33 @@ def post_household_composition(eq_id, form_type, collection_id, group_id):  # py
...
-        full_routing_path = path_finder.get_routing_path()
+        path_finder = PathFinder(schema, answer_store, metadata)
+        routing_path = path_finder.get_routing_path()
         latest_location = path_finder.get_latest_location(get_completed_blocks(current_user),
-                                                          routing_path=full_routing_path)
+                                                          routing_path=routing_path)
         return _redirect_to_latest_location(collection_id, eq_id, form_type, latest_location)
 
 
 @questionnaire_blueprint.route('signed-out', methods=["GET"])
```
collisdigital: See my later comments; should this be on the root URL now e.g. `/signed-out` ?

app/views/questionnaire.py
```diff
@@ -238,10 +235,8 @@ def submit_answers(eq_id, form_type, collection_id, metadata, answer_store):
         submitter = SubmitterFactory.get_submitter()
         message = convert_answers(metadata, g.schema_json, answer_store, path_finder.get_routing_path())
         submitter.send_answers(message)
-        questionnaire_store = get_questionnaire_store(current_user.user_id, current_user.user_ik)
-        metadata['submitted'] = True
-        questionnaire_store.metadata = metadata
-
+        session_storage.store_survey_completed_metadata(metadata['tx_id'], metadata['period_str'], metadata['ru_ref'])
+        get_questionnaire_store(current_user.user_id, current_user.user_ik).delete()
```
collisdigital, line 238: should the user also be signed-out here (without clearing their cookie)?

collisdigital: Otherwise they only get signed-out by a timeout as the thank-you page has no way to sign-out manually.

app/views/questionnaire.py
LJBabbage: If its on the route, without the user having metadata how am I going to get the theme and the survey title? At the minute I am getting the eq_id and form_type from the URL then using the schema_loader to get me those details

app/views/questionnaire.py
LJBabbage: Problem is I have end points which you are no longer are signed in for, the before_request will naturally fire for everything requests, if you have to be logged in for the before_request then those end points can never be reached

app/views/questionnaire.py
LJBabbage: Tricky, we already have a sign-out page and it carries a different template meaning I can't merge them together

app/views/questionnaire.py
LJBabbage: I can sign them out, but I used the login_required decorator on the thank-you GET meaning someone can't just randomly jump to the thank you page. I guess it gives us a choice, sign them out and allow anyone to get to thank you or keep them signed in and only allow legitimate signed in users to get there

app/authentication/session_storage.py
```diff
@@ -30,9 +31,9 @@ def store_user_id(user_id):
             # session has a add function but it is wrapped in a session_scope which confuses pylint
             db_session.add(eq_session)
 
-    def clear(self):
+    def delete_session_from_db(self):
         """
-        Removes a user id from the session
+        deletes user session from database
```
collisdigital, line 36: This looks like it removes a cookie value not something from the database?

app/authentication/session_storage.py
```diff
@@ -98,6 +122,24 @@ def has_user_ik():
         else:
             return False
 
+    def remove_user_ik(self):
+        """
+        Removes a user's ik from the session
+        """
+
+        if self.has_user_ik():
+            session.pop('user_ik')
```
collisdigital: Use the constant: `USER_IK`

app/views/questionnaire.py
```diff
@@ -170,43 +173,39 @@ def post_household_composition(eq_id, form_type, collection_id, group_id):  # py
...
-        full_routing_path = path_finder.get_routing_path()
-        latest_location = path_finder.get_latest_location(get_completed_blocks(current_user),
-                                                          routing_path=full_routing_path)
-        return _redirect_to_latest_location(collection_id, eq_id, form_type, latest_location)
+        return get_latest_location(collection_id, eq_id, form_type, schema)
 
 
-@questionnaire_blueprint.route('signed-out', methods=["GET"])
 @login_required
+def get_latest_location(collection_id, eq_id, form_type, schema):
```
collisdigital: maybe this should be `_redirect_to_latest_location` and the original `_redirect_to_latest_location` should be renamed `_redirect_to_location` as it just take a `location` anyway?

app/views/questionnaire.py
```diff
@@ -170,43 +173,39 @@ def post_household_composition(eq_id, form_type, collection_id, group_id):  # py
...
+    answer_store = get_answer_store(current_user)
+    path_finder = PathFinder(schema, answer_store, metadata)
+    routing_path = path_finder.get_routing_path()
+    latest_location = path_finder.get_latest_location(get_completed_blocks(current_user),
+                                                      routing_path=routing_path)
+    return _redirect_to_latest_location(collection_id, eq_id, form_type, latest_location)
+
+
+@questionnaire_blueprint.route('signed-out', methods=["GET"])
 def get_sign_out(eq_id, form_type, collection_id):  # pylint: disable=unused-argument
```
collisdigital: Does this require a `login_required` now?

app/views/questionnaire.py
collisdigital: yeah leave it for now, something for the future

app/views/questionnaire.py
```diff
@@ -216,18 +215,19 @@ def get_timeout_continue(eq_id, form_type, collection_id):  # pylint: disable=un
...
 
-@questionnaire_blueprint.route('session-expired', methods=["GET"])
+@questionnaire_blueprint.route('session-expired', methods=["POST"])
 @login_required
+def post_session_expired(eq_id, form_type, collection_id):  # pylint: disable=unused-argument
+    _remove_survey_session_data()
+    return get_session_expired(eq_id, form_type, collection_id)
+
+
+@questionnaire_blueprint.route('session-expired', methods=["GET"])
```
collisdigital, line 221: this used to have `@login_required` should it still ?

app/views/questionnaire.py
collisdigital: Only someone with the values in their cookie should be able to get to the thank-you page and they should only have them if we put them there after they were logged in previously to submit

app/views/questionnaire.py
LJBabbage: No the user is signed out at this point

app/views/questionnaire.py
LJBabbage: The user is signed out by this point

app/views/questionnaire.py
collisdigital: ok, can you raise a debt card about knowing the theme from the URL rather than having to load the schema etc?

app/views/questionnaire.py
collisdigital: Ok see comment below

app/authentication/session_storage.py
LJBabbage: The main code this function is calling is db_session.delete(eq_session). So I have just aligned the function name, clear was a rubbish, it confused me as I thought it was clearing the whole contents of the session, wrong

app/views/questionnaire.py
LJBabbage: Ok

app/authentication/session_storage.py
collisdigital: ok, makes sense :+1:

.application-version
```diff
@@ -0,0 +1 @@
+e116268968fec6ef4114cf6a5e5eb9bf00d98823
```
collisdigital: should this file be here? Perhaps it is useful to have one that contains 'not set' or similar?

app/authentication/session_storage.py
```diff
@@ -88,6 +89,29 @@ def store_user_ik(user_ik):
         session[USER_IK] = user_ik
 
     @staticmethod
+    def store_survey_completed_metadata(tx_id, period_str, ru_ref):
+        """
+        Store survey_completed_metadata in the cookie for retrieval later
+        """
+        session[SURVEY_COMPLETED_METADATA] = {
+            'tx_id': tx_id,
```
collisdigital, line 97: Do we also need the survey title as it's currently shown on the thank-you page?

app/views/questionnaire.py
collisdigital: OK

app/views/questionnaire.py
```diff
@@ -449,7 +449,7 @@ def extract_answer_id_and_instance(answer_instance_id):
     return answer_id, int(index)
 
 
-def _redirect_to_latest_location(collection_id, eq_id, form_type, latest_location):
+def _redirect_to_location(collection_id, eq_id, form_type, latest_location):
```
collisdigital: `latest_location` should probably be just `location` now

tests/integration/star_wars/test_backwards_navigation_after_submission.py
```diff
@@ -21,9 +21,6 @@ def test_backwards_navigation_star_wars(self):
         self.backwards_navigation()
 
     def backwards_navigation(self):
-        # Introduction
```
collisdigital, line 24: Shouldn't this fail/work/pass still?

collisdigital:
Do we have integration tests that verify the acceptance criteria for this work?

app/authentication/session_storage.py
LJBabbage: Comes from the schema, so no

tests/integration/star_wars/test_backwards_navigation_after_submission.py
LJBabbage: It was a duplication, look at the last assertion its introduction as well

LJBabbage:
We are well covered on integration tests, although I have added one more for when a survey is completed and then you try to go back

app/views/questionnaire.py
ajmaddaford: We should at least rename the route something like `expire-session`

yarn.lock
```diff
@@ -1,25 +1,5 @@
 # THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
 # yarn lockfile v1
```
ajmaddaford: Should this have been checked in? Seems like a lot of dependency changes.

app/assets/js/app/modules/timeout.js
```diff
@@ -124,7 +124,10 @@ domready(() => {
     let countDown = timeoutUI.onTick()
     if (countDown < 1) {
       window.clearInterval(this)
-      window.location = sessionExpiredUrl
+      fetch(sessionExpiredUrl, { method: 'POST' })
+        .then(() => {
+          window.location = sessionExpiredUrl
```
ajmaddaford: We should use two URLs - one for expiring the session and another for displaying the session expired page.

app/views/questionnaire.py
jonnyshaw89: Does the JavaScript need updating to POST to this now?

app/views/questionnaire.py
jonnyshaw89: as `get_thank_you` no longer has the `login_required` decorator i think @collisdigital is correct

jonnyshaw89 (changes requested):

jonnyshaw89 (approved):

.application-version
```diff
@@ -0,0 +1 @@
+fe98719296c18e99b6cee7fee01e8bdf80ab08b1
```
ajmaddaford: This file shouldn't be checked in

ajmaddaford (changes requested):

app/views/questionnaire.py
```diff
@@ -216,18 +212,19 @@ def get_timeout_continue(eq_id, form_type, collection_id):  # pylint: disable=un
     return 'true'
 
 
-@questionnaire_blueprint.route('session-expired', methods=["GET"])
+@questionnaire_blueprint.route('expire-session', methods=["POST"])
 @login_required
+def post_session_expired(eq_id, form_type, collection_id):  # pylint: disable=unused-argument
```
ajmaddaford: This method should be renamed `post_expire_session` to match the url change

tests/integration/questionnaire/test_questionnaire_previous_link.py
```diff
@@ -62,3 +62,14 @@ def test_previous_link_on_relationship(self):
         self.get('questionnaire/census/household/789/who-lives-here-relationship/1/household-relationships')
         self.assertInPage('Previous')
 
+    def test_when_completed_questionnaire_previous_on_thank_you_returns_unauthorised(self):
```
ajmaddaford: This test isn't testing the previous link so probably shouldn't be in this test

tests/integration/questionnaire/test_questionnaire_session_expired.py
```diff
@@ -5,7 +5,7 @@ class TestSessionExpired(IntegrationTestCase):
 
     def test_session_expired_should_log_user_out(self):
         self.launchSurvey('1', '0205')
-        self.get('/questionnaire/1/0205/789/session-expired')
+        self.post(url='/questionnaire/1/0205/789/session-expired')
```
ajmaddaford: These URLs are now `/expire-session`

tests/integration/session/test_timeout.py
```diff
@@ -18,9 +18,9 @@ def test_timeout_continue_returns_200(self):
 
     def test_session_expired_clears_session(self):
         self.launchSurvey('test', 'timeout')
-        self.get('/questionnaire/test/timeout/789/session-expired')
+        self.post(url='/questionnaire/test/timeout/789/session-expired')
         self.assertStatusOK()
-        self.get('/questionnaire/test/timeout/789/session-expired')
+        self.post(url='/questionnaire/test/timeout/789/session-expired')
```
ajmaddaford: Isn't this test a duplicate of `test_session_expired_should_log_user_out` in `test_questionnaire_session_expired.py`?

ajmaddaford (changes requested):

jonnyshaw89:
@LJBabbage You have many build errors.

tests/integration/questionnaire/test_questionnaire_endpoint_redirects.py
```diff
@@ -72,3 +72,39 @@ def test_given_complete_questionnaire_when_submitted_then_data_is_deleted(self):
...
+        self.assertInUrl('questionnaire/test/percentage/789/thank-you')
+
+    def test_when_survey_submitted_re_submitting_returns_unauthorised(self):
+        # Given we have submitted the test_percentage survey
+        self.launchSurvey('test', 'percentage', roles=['dumper'])
+        self.post({'answer': '99'})
+        self.post(action=None)
+
+        # When we try to submit the survey again
+        self.post(url='questionnaire/2/0001/789/qbs/0/summary')
```
ajmaddaford: Any reason this is another survey and not the percentage survey?

ajmaddaford (changes requested):

ajmaddaford (approved):

yoganandkunche:
Passed QA in pre-prod

PR #1072, opened 2017-03-30
Remove schema_parser and move metadata_parser

collisdigital:
https://trello.com/c/umt7a3S8/1034-remove-schema-parser

Remove schemas_parser and associated files; we validate all schemas so parsing doesn't add anything and the code is left over from legacy processing.

Also moved metadata_parser.py so the whole `parser/` folder is gone.

Removed some duplication around setUp of Flask app context on some unit tests - they're actually component tests really. 

### How to review 
1. Check that the schema_parser wasn't adding enough value to keep it
2. Is the new location metadata_parser.py ok?
3. Are the related changes ok?


dcdarrell9:
All looks fine to me, maybe worth having someone else look at it though?

LJBabbage (changes requested):
app/parser/metadata_parser.py → app/authentication/metadata_parser.py

hat on a girl → school for boys

collisdigital:
@LJBabbage done

LJBabbage:
@collisdigital Have you included an extra 30 files by accident on that last push?

collisdigital:
Making these changes revealed other issues (related to caching the schemas) which I fixed which identified more duplicate, which I fixed, which revealed that some unit tests were actually integration tests, which I moved. So yeah, I think I can justify the changes being related. Although I have accidentally included the Star Wars `__init__.py` I wanted a blank one; will sort it.

LJBabbage:
No problem, it would help if you didn't squash commits until its finished, thank you

jonnyshaw89:
Agree with @LJBabbage we agreed a while back that PRs shouldn't be squashed until it has 2 ticks as it makes reviewing ongoing fixes a lot more difficult.

collisdigital:
Sorry, was merging some conflicts from master and squashed it at the same time; I've removed the star wars init and reverted a few changes that weren't needed. We actually have a load of other 'unit tests' that are 'integration tests' really, in that they require the full flask app to be created, they now all use the new `AppContextTestCase` as they all performed the same `setUp` steps to init the flask app. Outside this PR it might be worth deciding what to do with them.

app/utilities/schema.py
```diff
@@ -1,37 +1,49 @@
...
-        raise ValueError("No schema available")
+def load_schema_from_params(eq_id, form_type, language_code='en'):
+    return load_schema_file("{}_{}.json".format(eq_id, form_type), language_code)
+
+
+def load_schema_file(schema_file, language_code='en'):
+    logger.info("loading schema", filename=schema_file, language_code=language_code)
+
+    try:
+        if language_code is not None and language_code != 'en':
```
jonnyshaw89: language_code can't be None? as you are defaulting it in the method

tests/app/forms/test_household_composition.py
```diff
@@ -27,7 +22,7 @@ def _error_exists(answer_id, msg, mapped_errors):
         return any(a_id == answer_id and msg in ordered_errors for a_id, ordered_errors in mapped_errors)
 
     def test_get_name_form_generates_name_form(self):
-        with create_app().test_request_context():
```
jonnyshaw89, line 30: 👍 this probably removes a lot of overhead!

jonnyshaw89 (changes requested):

app/utilities/schema.py
collisdigital: Someone can pass it None though

collisdigital: In fact we do pass in None if the language isn't in their metadata

jonnyshaw89: Fair enough. didn't realise that None would override the default. However if we are defaulting in this function maybe we should not explicitly pass in None if the language isn't in their metadata?

collisdigital: Can change that but it will make the calling code more complex i.e.

```python
@cache.memoize()
 def load_schema_from_metadata(metadata):
     return load_schema_from_params(metadata['eq_id'], metadata['form_type'], metadata.get('language_code'))
```

Would become:

```python
@cache.memoize()
 def load_schema_from_metadata(metadata):
    language_code = metadata.get('language_code')
    if language_code:
      return load_schema_from_params(metadata['eq_id'], metadata['form_type'], language_code))
    else:
      return load_schema_from_params(metadata['eq_id'], metadata['form_type'])
```

jonnyshaw89: Fair enough.

jonnyshaw89 (approved):

app/utilities/schema.py
collisdigital: Tried a refactor, see what you think

app/utilities/schema.py
```diff
@@ -20,25 +20,35 @@ def load_schema_from_params(eq_id, form_type, language_code='en'):
...
+    """
+    language_code = language_code or 'en'
+    schema_path = _get_schema_file_path(schema_file, language_code)
 
-    try:
-        if language_code is not None and language_code != 'en':
-            schema_dir = os.path.join(settings.EQ_SCHEMA_DIRECTORY, language_code)
-        else:
-            schema_dir = settings.EQ_SCHEMA_DIRECTORY
+    if language_code != 'en' and not os.path.exists(schema_path):
```
jonnyshaw89: do we need `language_code != 'en'`?

collisdigital: only in that the call would be identical to one above if the language code was 'en'

app/utilities/schema.py
```diff
@@ -1,37 +1,59 @@
...
     """
-    Use the schema loader to get the schema from disk. Then use the parse to construct the object schema
-    :param eq_id: the id of the questionnaire
-    :param form_type: the form type
-    :return: (json, schema) # Tuple of json and schema object from schema file
+    Load a schema, optionally for a specified language.
+    :param language_code: ISO 2-character code for language e.g. 'en', 'cy'
     """
-    # load the schema
+    language_code = language_code or 'en'
```
LJBabbage: Why are we defaulting in the function call to None, to then do it here again to en, might as well just do language_code=en?

app/utilities/schema.py
```diff
@@ -1,37 +1,59 @@
...
+    schema_path = _get_schema_file_path(schema_file, language_code)
 
-    json_schema = load_schema(eq_id, form_type, language_code)
-    if json_schema:
-        parser = SchemaParser(json_schema)
-        parser.parse()
-        return json_schema
+    if language_code != 'en' and not os.path.exists(schema_path):
+        schema_path = _get_schema_file_path(schema_file, 'en')
```
LJBabbage: Not sure about this, I don't feel if we can't find a language schema we should be defaulting to english. Surely it should be game over, as this could hide a problem we don't realise is happening.

app/utilities/schema.py
```diff
@@ -1,37 +1,59 @@
...
+        with open(schema_path, encoding="utf8") as json_data:
+            return json.load(json_data)
+
+    except FileNotFoundError as e:
+        logger.error("no schema file exists", filename=schema_file)
+        raise e
+
+
+def _get_schema_file_path(schema_file, language_code):
+    if language_code != 'en':
```
LJBabbage, line 49: Not sure there should be logic like this here, shouldn't something else deal with creating the path. This function should just go get

app/utilities/schema.py
```diff
@@ -1,37 +1,59 @@
...
+        schema_dir = os.path.join(settings.EQ_SCHEMA_DIRECTORY, language_code)
     else:
-        raise ValueError("No schema available")
+        schema_dir = settings.EQ_SCHEMA_DIRECTORY
+
+    return os.path.join(schema_dir, schema_file)
+
+
+def available_schemas():
```
LJBabbage: Would prefer a name with json in, as the function assumes you know schemas means json files in a given directory.

LJBabbage (changes requested):

app/utilities/schema.py
collisdigital: Because calling code can pass in None; would still need this line or force callers to not pass None in. Not sure if it's better for the caller to deal with it or the callee.

app/utilities/schema.py
collisdigital: That does mean anyone launching EQ has to know if there's a specific language version available before choosing it; we don't really have anything in place to make that clear (like a schema registy). Having an 'en' fallback is also how the Babel and other translation processes normally work I.e translate or show English if not available.

collisdigital: I haven't changed this, it's always supported this fallback method

app/utilities/schema.py
collisdigital: I think if we had an 'en' path it would be cleaner, not sure what the best approach is?

app/utilities/schema.py
collisdigital: Ok

app/utilities/schema.py
LJBabbage: No problem, just wanted to ask the question, maybe a debug message might be good to highlight a missing language.

collisdigital: Good idea, will add something

app/utilities/schema.py
collisdigital: See https://github.com/ONSdigital/eq-survey-runner/pull/1072#discussion_r109137285

app/utilities/schema.py
collisdigital: Done

app/utilities/schema.py
collisdigital: Done

collisdigital:
All comments I think needed to be addressed have been @LJBabbage best to discuss the others ?

LJBabbage (approved):

LJBabbage:
Squash please

app/utilities/schema.py
ajmaddaford: I think it makes sense to move the schemas to an `en` directory. We should also consider moving them out of the app directory. We should create a card for it.

collisdigital: Card added for that: https://trello.com/c/5RCkEGjF

collisdigital:
@LJBabbage squashed and rebased, waiting for tests to pass

collisdigital:
@LJBabbage shouldn't have merged; was waiting for Codacy to complete ;)

PR #1077, opened 2017-03-31
Credstash as secret management solution

jonnyshaw89:
WIP example of how credstash (https://github.com/fugue/credstash) could be used as a secret management solution

### How to review 
With EQ_DEV_KEYS set to False the application should reach out to DynamoDB and AWS KMS for its secrets


collisdigital:
Nice, what needs to be setup in AWS to enable this to work?

jonnyshaw89:
@collisdigital a KMS alias and a DynamoDB table
And obviously the keys loaded in.

jonnyshaw89:
@collisdigital was also talking to @ajmaddaford (MadDog) and this could also be used for rolling our encryption keys as they secrets can have versions and are immutable

collisdigital:
👌 ; would be good to compare with Vault; but looks promising.

jonnyshaw89:
@collisdigital the issue i can see with Vault is that we would need to deploy our own instances with cluster and autoscaling.
Which is an option, but also its a single point of failure.

config.py
```diff
@@ -23,7 +23,7 @@ class Config(object):
 
 class Test(Config):
     DEBUG = True
-    SECRET_KEY = settings.EQ_SECRET_KEY
```
collisdigital: Not sure if this whole file is even used/needed?

jonnyshaw89: Apparently not. Ill delete it.

PR #1084, opened 2017-04-05
Code coverage on errors and schema_helper

dcdarrell9:
- Added new functions to check whether there are any duplicate aliases
- Removed unused code 

### How to review 
- Check to see tests work correct
- Check that code was definitely unused

- Still some left to do on errors.py, not in for next two days, didn't want to leave as is.

### DIFF
--- a/app/helpers/schema_helper.py
+++ b/app/helpers/schema_helper.py
@@ -2,10 +2,6 @@ from app.questionnaire.location import Location
 from app.validation.error_messages import error_messages
 
 
-class QuestionnaireException(Exception):
-    pass
-
-
 class SchemaHelper(object):  # pylint: disable=too-many-public-methods
 
     @staticmethod
@@ -86,13 +82,6 @@ class SchemaHelper(object):  # pylint: disable=too-many-public-methods
         return next(b for b in cls.get_blocks(survey_json) if b['id'] == block_id)
 
     @classmethod
-    def get_block_ids(cls, survey_json):
-        block_ids = []
-        for block_json in cls.get_blocks(survey_json):
-            block_ids.append(block_json['id'])
-        return block_ids
-
-    @classmethod
     def get_group_ids(cls, survey_json):
         group_ids = []
         for group_json in cls.get_groups(survey_json):
@@ -145,13 +134,12 @@ class SchemaHelper(object):  # pylint: disable=too-many-public-methods
         for block in cls.get_blocks(survey_json):
             for question in cls.get_questions_for_block(block):
                 for answer in question['answers']:
-                    if 'alias' in answer and answer['alias'] not in aliases:
+                    if 'alias' in answer:
+                        assert answer['alias'] not in aliases, 'Duplicate alias found: ' + answer['alias']
                         aliases[answer['alias']] = {
                             'answer_id': answer['id'],
                             'repeats': answer['type'] == 'Checkbox' or question['type'] == 'RepeatingAnswer',
                         }
-                    elif 'alias' in answer:
-                        raise QuestionnaireException('{} is not a unique alias'.format(answer['alias']))
         return aliases
 
     @classmethod
@@ -207,14 +195,6 @@ class SchemaHelper(object):  # pylint: disable=too-many-public-methods
             group_id=cls.get_first_group_id(survey_json),
             group_instance=0,
             block_id=cls.get_first_block_id(survey_json),
-        )
-
-    @classmethod
-    def get_last_location(cls, survey_json):
-        return Location(
-            group_id=cls.get_last_group_id(survey_json),
-            group_instance=0,
-            block_id=cls.get_last_block_id(survey_json),
         )
 
     @classmethod
--- a/tests/app/data/test_schema_validation.py
+++ b/tests/app/data/test_schema_validation.py
@@ -67,21 +67,27 @@ class TestSchemaValidation(unittest.TestCase):
 
         return errors
 
-    def test_no_duplicate_ids_in_schema(self):
-        schema_files = self.all_schema_files()
-
+    def test_no_duplicate_id_in_schema(self):
         # Certain keys need to be ignored to avoid false positives.
         ignored_keys = ['routing_rules', 'skip_condition']
+        self.find_duplicates(ignored_keys, 'id')
 
+    def test_no_duplicate_alias_in_schema(self):
+        # Certain keys need to be ignored to avoid false positives.
+        ignored_keys = ['routing_rules', 'skip_condition']
+        self.find_duplicates(ignored_keys, 'alias')
+
+    def find_duplicates(self, ignored_keys, special_key):
+        schema_files = self.all_schema_files()
         for schema_file in schema_files:
-            unique_id = []
             with open(schema_file, encoding="utf8") as file:
                 schema_json = load(file)
-                for id_value in self._parse_id_values(schema_json, ignored_keys):
-                    if id_value in unique_id:
-                        self.fail('Duplicate Id found. schema: %s, id: %s' % (schema_file, id_value))
+                unique_items = []
+                for value in self._parse_values(schema_json, ignored_keys, special_key):
+                    if value in unique_items:
+                        self.fail('Duplicate {} found. Schema: {}, value {}'.format(special_key, schema_file, value))
                     else:
-                        unique_id.append(id_value)
+                        unique_items.append(value)
 
     def test_all_schemas_contain_confirmation_page(self):
         schema_files = self.all_schema_files()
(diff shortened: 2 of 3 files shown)


tests/app/data/test_schema_validation.py
```diff
@@ -148,6 +164,18 @@ def _parse_id_values(self, schema_json, ignored_keys):
                     if isinstance(schema_item, dict):
                         yield from self._parse_id_values(schema_item, ignored_keys)
 
+    def _parse_alias_values(self, schema_json, ignored_keys):
+        for k, v in schema_json.items():
```
jonnyshaw89: can we give v and k more logical name?
it names reading the function a lot easier.

app/helpers/schema_helper.py
```diff
@@ -145,13 +134,12 @@ def get_aliases(cls, survey_json):
         for block in cls.get_blocks(survey_json):
             for question in cls.get_questions_for_block(block):
                 for answer in question['answers']:
-                    if 'alias' in answer and answer['alias'] not in aliases:
+                    if 'alias' in answer:
+                        assert answer['alias'] not in aliases, 'Duplicate alias found.'
```
jonnyshaw89: What now catches this exception?

jonnyshaw89 (changes requested):

app/helpers/schema_helper.py
collisdigital: maybe output the `answer['alias']` as part of the message?

tests/app/data/test_schema_validation.py
```diff
@@ -83,6 +83,22 @@ def test_no_duplicate_ids_in_schema(self):
                     else:
                         unique_id.append(id_value)
 
+    def test_no_duplicate_aliases_in_schema(self):
+        schema_files = self.all_schema_files()
+
+        # Certain keys need to be ignored to avoid false positives.
+        ignored_keys = ['routing_rules', 'skip_condition']
```
collisdigital: Do we need to ignore these for this checker?

tests/app/data/test_schema_validation.py
dcdarrell9: What would be the best way to name them? The _parse_id_values uses the same so can change that as well as.

tests/app/data/test_schema_validation.py
```diff
@@ -148,6 +164,18 @@ def _parse_id_values(self, schema_json, ignored_keys):
                     if isinstance(schema_item, dict):
                         yield from self._parse_id_values(schema_item, ignored_keys)
 
+    def _parse_alias_values(self, schema_json, ignored_keys):
```
collisdigital: This function can be merged with the one above by pulling the `'alias'` into a parameter to it...

collisdigital (changes requested):
Nice to see some more code removed :+1:

app/helpers/schema_helper.py
collisdigital: It's an assert, i.e. should never happen, so nothing will catch it

tests/app/data/test_schema_validation.py
dcdarrell9: How would you do this?
The way I just changed it to (_parse_values)... makes the test fail for both alias and id if there is only one duplicate

tests/app/data/test_schema_validation.py
LJBabbage: They are the 'key' and the 'value'

app/helpers/schema_helper.py
LJBabbage: No maybe, definitely, otherwise debug is a pain

tests/app/data/test_schema_validation.py
```diff
@@ -148,6 +164,18 @@ def _parse_id_values(self, schema_json, ignored_keys):
...
+        for k, v in schema_json.items():
+            if k == 'alias':
+                yield v
+            elif k in ignored_keys:
```
LJBabbage: The function needs a name change or the content needs reducing. If its purely dealing with alias' why are we worrying about 171 down.

tests/app/data/test_schema_validation.py
```diff
@@ -67,21 +67,31 @@ def validate_schema(self, schema, file, json_to_validate):
...
+                        if key == 'id':
+                            if value in unique_id:
+                                self.fail('Duplicate ID found. schema: %s, id: %s' % (schema_file, value))
+                            else:
+                                unique_id.append(value)
+                        elif key == 'alias':
+                            if value in unique_alias:
+                                self.fail('Duplicate alias found. schema: %s, alias: %s' % (schema_file, value))
+                            else:
+                                unique_alias.append(value)
```
dcdarrell9: Issue with this is it won't check alias if there is a failure in id.

samiwel: I think checking for duplicate Ids and duplicate aliases should be two separate tests. Just makes it clearer when one of them fails, you could re-use much of the logic here by extracting a method and passing 'id' or 'alias' as a parameter.

app/helpers/schema_helper.py
```diff
@@ -145,13 +134,12 @@ def get_aliases(cls, survey_json):
         for block in cls.get_blocks(survey_json):
             for question in cls.get_questions_for_block(block):
                 for answer in question['answers']:
-                    if 'alias' in answer and answer['alias'] not in aliases:
+                    if 'alias' in answer:
+                        assert answer['alias'] not in aliases, 'Duplicate alias found: ' + answer['alias']
```
samiwel, line 138: I'd question the need for an assert here at all. It's not a bad thing per se, and you may have a good reason for having it here, I don't know. But I think having a test in `test_schema_validation.py` and catching it at build time is sufficient. Since we're not dealing with the AssertionError at runtime anyway. Your call though.

samiwel (changes requested):
Looks good Darrel, I've added a few minor comments. But feel free to disagree with me if you want. 👍

app/helpers/schema_helper.py
collisdigital: in theory asserts are removed from production code (you have to turn on the Python optimise flags) so it's a developer aid in case anything ever goes wrong (some code changes break it for example or the schema validator stopped working)

tests/app/data/test_schema_validation.py
```diff
@@ -67,21 +67,31 @@ def validate_schema(self, schema, file, json_to_validate):
...
             with open(schema_file, encoding="utf8") as file:
                 schema_json = load(file)
-                for id_value in self._parse_id_values(schema_json, ignored_keys):
-                    if id_value in unique_id:
-                        self.fail('Duplicate Id found. schema: %s, id: %s' % (schema_file, id_value))
-                    else:
-                        unique_id.append(id_value)
+                for key in special_key:
+                    for value in self._parse_values(schema_json, ignored_keys, key):
+                        if key == 'id':
```
collisdigital: you could remove all this `if` logic if you reset `unique_id` at the start of the for loop `for key in special_key:` e.g.

```python
for key in special_key:
    unique_items = []
    ....
```

app/helpers/schema_helper.py
samiwel: Yeah, that's fair enough, I don't have strong views on this. Just feels a bit odd checking for duplicates in the same place where we're retrieving the values, but it's an additional safeguard I guess.

collisdigital (approved):
Looks good; @samiwel do you approve as well? If so @dcdarrell9 can get this rebased, squashed and we can merge.

collisdigital:
@jonnyshaw89 happy with this PR now?

app/helpers/schema_helper.py
jonnyshaw89: @collisdigital do we have the Python optimise flag currently set in production?

collisdigital: Not that I know of; I think when we checked we would need to change the EB instance internally to turn it on, there be dragons!

samiwel (approved):

app/helpers/schema_helper.py
samiwel: Codacy flags this up as an issue, if we're agreeing leaving the assert in is okay, is there some way of configuring codacy not to flag it up?

jonnyshaw89: I think we either need to use asserts and enable the Python optimise flag or not use asserts.

ajmaddaford: We have used asserts elsewhere in the code, so I've created a card for reviewing whether we should or not. For now I've disabled the Codacy rule.

jonnyshaw89 (approved):

Issue #833, opened 2017-01-11
CSS/JS assets are not minified


hamishtaplin:
### Expected behaviour

CSS and JS files should be minified in pre-prod/prod

### Actual behaviour

Assets are not minified

### Steps to reproduce the behaviour

- Load page in any browser
- Confirm CSS/JS assets do not use the minified versions via the filename (eg. should be .min.js)

#### Browser

All

### Screenshot

<img width="1020" alt="screenshot 2017-01-11 11 18 19" src="https://cloud.githubusercontent.com/assets/930398/21846639/f4a5aa1a-d7ef-11e6-9df8-634bd3f52c36.png">

adoublebarrel:
The good news is that in terms of code functionality we seem to have everything we need.

The bad news is that the minification build step isn't in place so if we turned on serving the minified versions right now things would go bang.

### Steps To Fix

1. Enable minification for pre-prod and prod builds (and dev?)
1. Switch the EQ_MINIMIZE_ASSETS env var to True in Jenkins for pre-prod and prod 

Source maps is all setup so we could use the minified assets in dev, @ajmaddaford is likely to check if that works for @hamishtaplin

PR #1085, opened 2017-04-06
Add tests for app/__init__.py + boy scouting

adoublebarrel:
[Trello Card](https://trello.com/c/Nm0Hy1Z3/1041-code-coverage-on-app-init-py)

Aim is to improve code coverage of app/__init__.py. New tests exercise existing code/behaviour that isn't already covered by existing integration tests. I've included comments identifying the behaviour that *isn't* tested and where they seem to be covered already.

Have also removed the Markdown plugin as it's no longer used. Fixed a
typo in a test filename.

### How to review 
- Test file can be found in tests/integration/test_app_create.py
- All tests should pass


jonnyshaw89 (commented):
Would be good to add a test for `get_minimized_asset` so that we can get 100% coverage on `__init__.py`

collisdigital:
I think we might have an issue there as well John! https://github.com/ONSdigital/eq-survey-runner/issues/833

adoublebarrel:
Good point, I've added a unit test to cover get_minimized_asset's behaviour. I've also added a couple of direct tests of versioned_url_for.

I added a comment addressing #883 as well.

collisdigital (approved):
Looks good

ajmaddaford (approved):

tests/integration/test_app_create.py
```diff
@@ -0,0 +1,165 @@
...
+        self.assertEqual('nosniff', headers['X-Content-Type-Options'])
+
+
+    # Indirectly covered by higher level integration
+    # tests, keeping to highlight that create_app is where
+    # it happens.
+    def test_adds_blueprints(self):
+        self.assertGreater(len(app.create_app().blueprints), 0)
+
+    # covered by session/test_login.py
```
jonnyshaw89: Should these commented tests be removed?

jonnyshaw89 (changes requested):

tests/integration/test_app_create.py
adoublebarrel: I left them in because I thought they *might* be useful to anyone who comes after who might wonder why functionality X of create_app() isn't covered in this file.

I'm happy to remove them though, or I could remove the defs and just keep one line comments highlighting where the behaviour is tested?

jonnyshaw89 (approved):

PR #1086, opened 2017-04-06
QBS Increase test coverage of the downstream data

yoganandkunche:
Dumps the answers that are going to be submitted and compares with the answers that were entered.

### How to review 
Run the integration test.

### DIFF
--- a/tests/integration/integration_test_case.py
+++ b/tests/integration/integration_test_case.py
@@ -1,6 +1,7 @@
 import os
 import re
 import unittest
+import json
 
 from bs4 import BeautifulSoup
 
@@ -91,6 +92,28 @@ class IntegrationTestCase(unittest.TestCase):  # pylint: disable=too-many-public
         """
         token = create_token(form_type_id=form_type_id, eq_id=eq_id, **payload_kwargs)
         self.get('/session?token=' + token.decode())
+
+    def dumpAnswers(self):
+
+        self.get('/dump/answers')
+
+        # Then I get a 200 OK response
+        self.assertStatusOK()
+
+        # And the JSON response contains the data I submitted
+        dump_answers = json.loads(self.getResponseData())
+        return dump_answers
+
+    def dumpSubmission(self):
+
+        self.get('/dump/submission')
+
+        # Then I get a 200 OK response
+        self.assertStatusOK()
+
+        # And the JSON response contains the data I submitted
+        dump_submission = json.loads(self.getResponseData())
+        return dump_submission
 
     def get(self, url):
         """
--- a/tests/integration/mci/test_change_answers.py
+++ b/tests/integration/mci/test_change_answers.py
@@ -1,3 +1,4 @@
+
 from tests.integration.integration_test_case import IntegrationTestCase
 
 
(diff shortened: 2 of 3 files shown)


samiwel (approved):
Looks good to me.

tests/integration/mci/test_change_answers.py
```diff
@@ -47,3 +49,54 @@ def happy_path(self, eq_id, form_type_id):
...
         self.assertInPage('Please check carefully before submission')
         self.assertInPage('>Submit answers<')
+
+        self.get('/dump/answers')
+
+        # Then I get a 200 OK response
+        self.assertStatusOK()
+
+        # And the JSON response contains the data I submitted
+        actual = json.loads(self.getResponseData())
```
collisdigital: I wonder if we should move this into the base class now so we can call `self.dumpAnswers()` which results the parsed json dict so move lines 53 through lines 59 into the IntegrationTestCase class and just call that here

`actual = self.dumpAnswers()`

tests/integration/mci/test_change_answers.py
```diff
@@ -47,3 +49,54 @@ def happy_path(self, eq_id, form_type_id):
         self.assertInPage('>Your responses<')
         self.assertInPage('Please check carefully before submission')
         self.assertInPage('>Submit answers<')
+
+        self.get('/dump/answers')
```
collisdigital: We should test against the submitted data: `self.get('/dump/submission')`

tests/integration/mci/test_change_answers.py
collisdigital: With above comment, it would be `actual = self.dumpSubmission()`

tests/integration/mci/test_change_answers.py
```diff
@@ -1,3 +1,5 @@
+import json
```
collisdigital: if `dumpSubmission` moves to the base class you'll need to remove this import.

collisdigital (changes requested):
Looking good, some refactoring to reduce duplication needed and should be checking the submission dump instead of answer store.

yoganandkunche:
@collisdigital addressed all the comments

tests/integration/integration_test_case.py
```diff
@@ -92,6 +93,28 @@ def launchSurvey(self, eq_id='test', form_type_id='radio', **payload_kwargs):
...
+        self.get('/dump/answers')
+
+        # Then I get a 200 OK response
+        self.assertStatusOK()
+
+        # And the JSON response contains the data I submitted
+        dump_answers = json.loads(self.getResponseData())
+        return dump_answers
+
+    def dumpSubmissions(self):
```
collisdigital: drop the 's' : Submission

tests/integration/qbs/test_qbs_change_answers.py
```diff
@@ -0,0 +1,72 @@
...
+        # We proceed to the questionnaire
+        self.post(action='start_questionnaire')
+
+        # We are in the Questionnaire
+        self.assertInPage('>Quarterly Business Survey</')
+        self.assertInPage('On 1 April 2016 what was the number of employees for Integration Tests?')
+        self.assertInPage('>Save and continue<')
+
+        # When I submit an answer
+        self.post(action='start_questionnaire')
```
collisdigital: Should this post be here? I'm interested why the test passes if this shouldn't be

tests/integration/integration_test_case.py
yoganandkunche: made the change

tests/integration/qbs/test_qbs_change_answers.py
yoganandkunche: removed the extra unnecessary post.. not sure yet why it still passes with an extra post.

collisdigital (approved):
Looks good, rebase, squash then ready to merge :+1:

PR #1087, opened 2017-04-06
eq-1044 Submission button

hamishtaplin:
As a user, I need the action of submitting my answers to be clear, reliable and resilient so that I am confident my answers are being submitted and I am prevented from breaking the process.

# Acceptance Criteria

- User is given an indication when submitting their answers that something is happening
- User is prevented (as much as possible) from accidentally double-submitting/causing a UX failure
- Solution can be a progressive enhancement (not available to all users).
- Solution extensively tested on all browsers and devices, this could break submission.


armstrongb:
Can we change this so that the spinner doesn't replace the copy entirely but is shown alongside copy: "[spinner] Submitting..."?
submission-spinner

yoganandkunche:
@hamishtaplin: as discussed please add the spinner 'submitting' to MWSS, UKIS and Census submission screens.

yoganandkunche:
@hamishtaplin timeout pop up displays '**Continuing**..' PFA -- Bug introduced on this branch.

screen shot 2017-04-12 at 16 32 37

yoganandkunche:
@collisdigital I had a discussion with @hamishtaplin and we want you to take a decision before we go ahead with this feature... as at the moment the "[spinner] Submitting..." thing comes up for 5 secs only and the scenario where it is misleading to the respondent is when the network is bad and it takes more than 5 secs before it get submitted the [Submit survey] button is displayed again which might confuse the respondent to click it again which results in 'Session timeout'.

collisdigital:
Show me on Tuesday

collisdigital:
Having seen it on a slow connection, can we increase the submitting timer to 10 seconds instead of 5?

bennpearson (changes requested):
Animation appears to stop after 5 seconds. Not sure if this is something that chrome dev tools throttling is introducing though

ajmaddaford:
I've just tried this and the animation runs for 10 seconds. The button resizes when the text changes to "Submitting..." which we may not want?

app/assets/js/app/modules/submitter.js
```diff
@@ -0,0 +1,25 @@
...
+
+      count++
+
+      trackEvent('send', {
+        hitType: 'event',
+        eventCategory: 'Error',
+        eventAction: 'Submission timeout',
+        eventLabel: `Attempts: ${count.toString()}`
+      })
+    }, 10000)
```
ajmaddaford: We should use a constant for this as it's a little unclear using the value directly.

ajmaddaford (changes requested):

app/assets/js/app/modules/submitter.js
hamishtaplin: Do we want to make this configurable or just more explicit?

ajmaddaford: Just more explicit

bennpearson (approved):

ajmaddaford (approved):

jonnyshaw89:
@hamishtaplin Can you update the description on this PR?

jonnyshaw89 (approved):

PR #1088, opened 2017-04-07
Secret management

jonnyshaw89:
Load all credentials and secrets from CredStash

### How to review 
Ensure that there are no credentials or secrets that need to be stored with the app.


jonnyshaw89:
Splitting to other PRs

PR #1090, opened 2017-04-10
Timeout Javascript to cancel interval

jonnyshaw89:
The timeout Javascript was causing requests to be made over 1 second if the browser did not load the expired page within 1 second

### How to review 
Test the timeout on a slow network connection (simulate in chrome) when the page expires then only 1 network request should be make to the expired page.


collisdigital (approved):
Tested before and after, works as described.

hamishtaplin (approved):

PR #1091, opened 2017-04-11
added unit tests in user.py and factory.py increase coverage

dcdarrell9:
- Added 2 simple unit tests in user.py and factory.py to increase coverage.

### How to review 
- Check all unit tests/lint work.
- Suggest improvements?


tests/app/authentication/test_user.py
```diff
@@ -22,3 +22,8 @@ def test_get_user_ik(self):
         with self.application.test_request_context():
             user = User("1", "2")
             self.assertEqual("2", user.user_ik)
+
+    def test_no_user(self):
+        with self.application.test_request_context():
+            with self.assertRaises(ValueError):
+                User("", "")
```
collisdigital, line 34: Worth a test passing None as well? And below...

dcdarrell9: Added...
Just a question, with this `if user_id and user_ik:` 
Is it possible to have one and not have the other, or do they both have to be there?
By this I mean in the app, not in this logic

collisdigital: Both should be mandatory

tests/app/utilities/test_factory.py
```diff
@@ -32,5 +32,10 @@ def test_register_all(self):
         self.assertIsInstance(factory.create("test-a"), TestClassA)
         self.assertIsInstance(factory.create("test-b"), TestClassB)
 
+    def test_create_value_error(self):
```
jonnyshaw89: I can't find anywhere that we actually use `Factory` within the code.
Maybe it should just be removed?

jonnyshaw89 (changes requested):

tests/app/utilities/test_factory.py
dcdarrell9: I think you're right, I can't seem to find it being used, 'SubmitterFactory' is used but that's a different thing.

jonnyshaw89 (approved):

Issue #965, opened 2017-02-20
Summary page bottom border cut off for Other answer


benjefferies:
### Expected behaviour
Summary bottom border should be visible.

### Actual behaviour
Summary bottom border is hidden.

### Steps to reproduce the behaviour
1. Go to https://preprod-surveys.eq.ons.digital/dev
1. Start test_checkbox.json
1. Complete questionnaire
1. Observe missing bottom border on summary

#### Browser
Google Chrome Version 56.0.2924.87 (64-bit)

#### Operating System
10.12.3 (16D32)

### Screenshot
![image](https://cloud.githubusercontent.com/assets/15050019/23121148/2d78cfee-f757-11e6-8fec-e6841c2ed8fa.png)

bennpearson:
`overflow: visible` on `.summary__items` possible fix

yoganandkunche:
We need the question numbers displayed on the summary page.

PR #1094, opened 2017-04-12
Update summary screen with latest design

hamishtaplin:
Updates the summary screen with the latest design (see https://trello.com/c/0efomJKb/913-summary-screen-mwss)

### How to review 
Review against acceptance criteria from Trello card:
- Currency values are presented with £, commas and pence e.g. **£123,456.00**
- All other values as per design (BenA will check and update A/C if divergence)
- All questions/answers asked on the respondent's routing path are displayed on the summary screen.
- Summary screen functionality behaves as before (e.g. Edit etc.)
- MWSS summary screen is enabled.
- All existing surveys with a summary adopt the new design changes.
- An Analytics event is added for the use of the Edit links (Navigation, Edit, URL)
- Summary screen shows the question numbers for questions.


samiwel:
@hamishtaplin Over to you 👍

app/templating/summary_context.py
```diff
@@ -14,16 +14,11 @@ def build_summary_rendering_context(schema_json, answer_store, metadata):
...
-                link = url_for('questionnaire.get_block',
-                               eq_id=metadata['eq_id'],
-                               form_type=metadata['form_type'],
-                               collection_id=metadata['collection_exercise_sid'],
-                               group_id=group['id'],
-                               group_instance=0,
-                               block_id=block['id'])
-                sections.extend([Section(section, answer_store.map(), link) for section in block['sections']])
-    return sections
+        if 'title' in group and group['title'] not in ('Introduction', 'Summary') \
```
ajmaddaford: We don't want to be using the title to determine behaviour here. Is this being done so that we don't add groups that don't contain any answers?

app/data/test_language.json
```diff
@@ -11,18 +11,19 @@
...
+                "title": ""
+            },
+            {
+                "type": "Summary",
+                "id": "summary"
+            }
+        ]
     }]
-}
+}
```
ajmaddaford, line 29: It looks like some of the schemas have been reformatted, making it hard to see what the required changes are for this pull request. Please can any that have just been reformatted be reverted.

ajmaddaford (changes requested):

app/templating/summary_context.py
samiwel: There is no type to latch onto for the group. So any change here would require either looking at a block within the group (which could be problematic if there are multiple blocks in the group) or introducing some other property to identify that it should not be displayed in the summary (which may make more sense). Feel free to change this if it makes sense, I am not in for 3 weeks and won't get an opportunity to work on this, sorry.

app/data/test_language.json
samiwel: I think they were all highlighted during a yarn lint, which was breaking my build locally. I think I made a separate commit for the JSON formatting, so feel free to revert that.

bennpearson (approved):
Frontend LGTM 👍  My changes need reviewing by @armstrongb

app/templating/summary_context.py
ajmaddaford: I've changed this so it checks if there are any `Questionnaire` blocks in a group.

app/data/test_language.json
ajmaddaford: Fair enough 👍

yoganandkunche:
@bennpearson #965 still fails of a checkbox other fields.

ajmaddaford (approved):

PR #1095, opened 2017-04-13
Added tests in submitter and changed metadata_context

dcdarrell9:
Added tests for coverage and changed some code.

### How to review 
- Check to see if the tests do what they should.
- Is the code in metadata_context worth keeping? (validated elsewhere)
- Suggest improvements.

### DIFF
--- a/tests/app/submitter/test_submitter.py
+++ b/tests/app/submitter/test_submitter.py
@@ -4,7 +4,7 @@ from mock import patch, Mock, call
 from pika.exceptions import AMQPError
 
 from app import settings
-from app.submitter.submitter import RabbitMQSubmitter
+from app.submitter.submitter import RabbitMQSubmitter, LogSubmitter, SubmitterFactory, Submitter
 
 
 class TestSubmitter(TestCase):
@@ -78,3 +78,30 @@ class TestSubmitter(TestCase):
 
             # Then
             self.assertFalse(published, 'send_message should fail to publish message')
+
+    def test_log_submitter_send_message(self):
+        # Given
+        with patch('app.submitter.submitter.BlockingConnection'):
+            # When
+            sent_message = LogSubmitter().send_message(message={}, queue='test_queue')
+
+            # Then
+            self.assertEqual(sent_message, True)
+
+    def test_get_submitter_settings_enabled(self):
+        settings.EQ_RABBITMQ_ENABLED = True
+        test = SubmitterFactory.get_submitter()
+        self.assertIsInstance(type(test), cls=object)
+
+    def test_get_submitter_settings_not_enabled(self):
+        settings.EQ_RABBITMQ_ENABLED = False
+        test = SubmitterFactory.get_submitter()
+        self.assertIsInstance(type(test), cls=object)
+
+    def test_submitter_send_message(self):
+        test = Submitter.send_message(self, message={}, queue='test_queue')
+        self.assertEqual(test, False)
+
+    def test_submitter_encrypt_message(self):
+        test = Submitter.encrypt_message(message={})
+        self.assertIsInstance(test, cls=str)
--- a/tests/app/submitter/test_submitter.py
+++ b/tests/app/submitter/test_submitter.py
@@ -90,18 +90,19 @@ class TestSubmitter(TestCase):
 
     def test_get_submitter_settings_enabled(self):
         settings.EQ_RABBITMQ_ENABLED = True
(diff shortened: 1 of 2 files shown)


tests/app/submitter/test_submitter.py
```diff
@@ -78,3 +78,30 @@ def test_when_fail_to_publish_message_then_returns_false(self):
...
+            # When
+            sent_message = LogSubmitter().send_message(message={}, queue='test_queue')
+
+            # Then
+            self.assertEqual(sent_message, True)
+
+    def test_get_submitter_settings_enabled(self):
+        settings.EQ_RABBITMQ_ENABLED = True
+        test = SubmitterFactory.get_submitter()
+        self.assertIsInstance(type(test), cls=object)
```
jonnyshaw89: Should we also test that the correct object is returned?

tests/app/submitter/test_submitter.py
```diff
@@ -78,3 +78,30 @@ def test_when_fail_to_publish_message_then_returns_false(self):
...
+
+    def test_submitter_send_message(self):
+        test = Submitter.send_message(self, message={}, queue='test_queue')
+        self.assertEqual(test, False)
+
+    def test_submitter_encrypt_message(self):
+        test = Submitter.encrypt_message(message={})
+        self.assertIsInstance(test, cls=str)
```
jonnyshaw89: You should test that the messages is actually being encrypted rather than just writing a test to touch lines.

tests/app/submitter/test_submitter.py
```diff
@@ -78,3 +78,30 @@ def test_when_fail_to_publish_message_then_returns_false(self):
...
+        self.assertIsInstance(type(test), cls=object)
+
+    def test_get_submitter_settings_not_enabled(self):
+        settings.EQ_RABBITMQ_ENABLED = False
+        test = SubmitterFactory.get_submitter()
+        self.assertIsInstance(type(test), cls=object)
+
+    def test_submitter_send_message(self):
+        test = Submitter.send_message(self, message={}, queue='test_queue')
+        self.assertEqual(test, False)
```
jonnyshaw89: The name of the test implies that the result should be True?

jonnyshaw89 (changes requested):

tests/app/submitter/test_submitter.py
dcdarrell9: Agree, I know the right object is returned because it prints it into the terminal but I don't know how to check it?

jonnyshaw89: Can you do something like?
`self.assertIsInstance(type(test), RabbitMQSubmitter)`

dcdarrell9: `IsInstance` doesn't work, but `assertEqual` works, is that right?

jonnyshaw89: If it works i guess so 👍

tests/app/submitter/test_submitter.py
dcdarrell9: I'm not sure about this one because it's an abstract method, do they get tested differently?

tests/app/submitter/test_submitter.py
```diff
@@ -78,3 +78,31 @@ def test_when_fail_to_publish_message_then_returns_false(self):
...
+        self.assertEqual(type(log_submitter), LogSubmitter)
+
+    def test_submitter_send_message(self):
+        sent_message_test = Submitter.send_message(self, message={}, queue='test_queue')
+        self.assertEqual(sent_message_test, False)
+
+    def test_submitter_encrypt_message(self):
+        encrypt_message_test = Submitter.encrypt_message(message={'test': 'test'})
+        self.assertIsInstance(encrypt_message_test, cls=str)
+        self.assertNotEqual(encrypt_message_test, 'test')
```
dcdarrell9: From what I think I know... 'test' would get printed if it wasn't encrypted, am I right to say that?

jonnyshaw89: yes that is correct. however the test isn't. this is currently testing that the entire message content is 'test' rather than the messages containing the word test

tests/app/submitter/test_submitter.py
jonnyshaw89: Maybe test send_message with both the LogSubmitter and the RabbitMQSubmitter?
each will have a different action.

PR #1096, opened 2017-04-18
Stage 2 of using Flask config for settings

adoublebarrel:
[Trello Card](https://trello.com/c/4kjDSdfM/1049-move-app-config-from-settings-py-to-flask-application-factory)

Apologies for some of the white space edits, the latest version of
flake8 is more fussy than its previous incarnations.

Reworked the submitter so that it only sends messages rather than
encrypting and sending them. The encryption step has been moved up
into the questionnaire blueprint.

While making the changes I started having trouble with the way the
submitter was being mocked out. I've introduced a helper which makes
the submitter available as a global. That global can then be reliably
overridden on a per-test basis, see test_downstream_data_typing.

### How to review 

All existing behaviour should be intact. Other than that special consideration should be given to:

- The changes to the design of the questionnaire view/blueprint + submitter 
- How the submitter is now being mocked out should also have a critical eye cast over it


app/views/flush.py
```diff
@@ -54,9 +56,22 @@ def _submit_data(user):
...
+        encrypter = Encrypter(
+            current_app.config['EQ_SUBMISSION_SR_PRIVATE_SIGNING_KEY'],
+            current_app.config['EQ_SUBMISSION_SR_PRIVATE_SIGNING_KEY_PASSWORD'],
+            current_app.config['EQ_SUBMISSION_SDX_PUBLIC_KEY'],
+        )
+
+        submitter = get_submitter()
         message = convert_answers(metadata, schema, answer_store, routing_path)
-        submitter.send_answers(message)
+        message = encrypter.encrypt(message)
```
jonnyshaw89: Will this always encrypt the message?

app/views/questionnaire.py
```diff
@@ -222,21 +219,33 @@ def post_expire_session(eq_id, form_type, collection_id):  # pylint: disable=unu
...
+
+        encrypter = Encrypter(
+            current_app.config['EQ_SUBMISSION_SR_PRIVATE_SIGNING_KEY'],
+            current_app.config['EQ_SUBMISSION_SR_PRIVATE_SIGNING_KEY_PASSWORD'],
+            current_app.config['EQ_SUBMISSION_SDX_PUBLIC_KEY'],
+        )
+
         message = convert_answers(metadata, g.schema_json, answer_store, path_finder.get_routing_path())
-        submitter.send_answers(message)
+        message = encrypter.encrypt(message)
```
jonnyshaw89: Will this always encrypt the message?

jonnyshaw89:
Have you not committed `app.helpers.submitter_helper`?

app/views/questionnaire.py
adoublebarrel: Yes, I've updated the tests to decrypt the message when they need to.

I take your point though, is this something we want to switch off in some circumstances?

adoublebarrel:
🤦‍♂️ No, I've forgotten to commit the helper. I'll update it as soon as I'm back.

app/views/questionnaire.py
jonnyshaw89: IIRC in dev mode it wouldn't encrypt the message to it was logged out so we could read it.

adoublebarrel:
*have forgotten to commit the helper

app/cryptography/jwt_encoder.py
```diff
@@ -4,26 +4,22 @@
...
+    def __init__(self, private_key, private_key_password, public_key):
+
+        if private_key is None:
             raise Exception("EQ_USER_AUTHENTICATION_RRM_PRIVATE_KEY is None (check EQ_DEV_MODE?)")
-        if settings.EQ_USER_AUTHENTICATION_RRM_PRIVATE_KEY_PASSWORD is None:
+        if private_key_password is None:
             raise Exception("EQ_USER_AUTHENTICATION_RRM_PRIVATE_KEY_PASSWORD is None (check EQ_DEV_MODE?)")
-        if settings.EQ_USER_AUTHENTICATION_SR_PUBLIC_KEY is None:
+        if public_key is None:
             raise Exception("EQ_USER_AUTHENTICATION_SR_PUBLIC_KEY is None (check EQ_DEV_MODE?)")
```
ajmaddaford: It looks like these three values no longer come from the settings so the exception text doesn't make sense now. It would be better to use something like `private_key missing`, `private_key_password missing` and `public_key missing`

app/helpers/submitter_helper.py
```diff
@@ -0,0 +1,21 @@
...
+    if rabbit_enabled:
+        # pylint: disable=protected-access
+        g._submitter = RabbitMQSubmitter(
+            rabbit_primary_url,
+            rabbit_secondary_url,
+        )
+    else:
+        g._submitter = LogSubmitter()  # pylint: disable=protected-access,redefined-variable-type
+
+    return g._submitter  # pylint: disable=protected-access
```
ajmaddaford: Making the submitter global so we can override it in the tests doesn't seem right. After a conversation with @adoublebarrel we're going to move the initialisation of the submitter to application startup. This suits how it is used (the type of submitter is never changed once the app is running) and also means each time we submit we don't need to supply the config.

app/submitter/encrypter.py
```diff
@@ -20,10 +19,7 @@ class Encrypter(JWEEncrypter):
 
     # pylint: disable=maybe-no-member
```
ajmaddaford: I don't think this is needed now we no longer default the values from settings

app/submitter/encrypter.py
```diff
@@ -20,10 +19,7 @@ class Encrypter(JWEEncrypter):
...
     # password and key variables are dynamically assigned
-    def __init__(self,
-                 private_key=settings.EQ_SUBMISSION_SR_PRIVATE_SIGNING_KEY,
-                 private_key_password=settings.EQ_SUBMISSION_SR_PRIVATE_SIGNING_KEY_PASSWORD,
-                 public_key=settings.EQ_SUBMISSION_SDX_PUBLIC_KEY):
+    def __init__(self, private_key, private_key_password, public_key):
```
ajmaddaford, line 21: We should check that these parameters are provided in a similar way to the `Encoder` initialisation.

ajmaddaford (changes requested):

app/cryptography/jwt_encoder.py
adoublebarrel: Is there value in adding some stronger validation here as well? We could push the validation into encrypter and then this init method could be removed altogether from here

ajmaddaford (approved):

jonnyshaw89 (approved):

PR #1098, opened 2017-04-18
Simplify Dockerfile

jonnyshaw89:
The Dockerfile is very complicated.
This Dockerfile sets out how we cause containerise the application moving forward

### How to review 
Play with it and see if it works.
This is meant less for Developers to use as a development environment (as I've never seen that work well) and more as a packaging of the application.


ajmaddaford:
The reason for using Apache was to match the Elastic Beanstalk setup as closely as possible. If we were to use this in production we would need to use something like [gunicorn](http://gunicorn.org/) as the Flask development web server is [not suitable for production](http://flask.pocoo.org/docs/0.12/deploying/).

jonnyshaw89:
This PR is a bit before its time.

PR #1100, opened 2017-04-20
Added print tracking to GA

dcdarrell9:
- Added the ability to track when a user opens up a print option (file > print || cmd + p) through Google Analytics

### How to review 
- Need access to Google Analytics with 'eq.onsdigital' and run it on localhost with the UA ID
- Try a print option in a survey on dev_mode and check that it comes through on Google Analytics
- Suggest better place for code? (maybe in analytics.js?)

### DIFF
--- a/app/assets/js/app/modules/analytics.js
+++ b/app/assets/js/app/modules/analytics.js
@@ -46,6 +46,25 @@ export default function initAnalytics() {
       trackElement(target)
     }
   })
+
+  const afterPrint = () => {
+    return trackEvent('send', {
+      hitType: 'event',
+      eventCategory: 'Print Intent',
+      eventAction: 'Print Intent',
+      eventLabel: window.location.pathname.split('/').slice(-3).join('/')
+    })
+  }
+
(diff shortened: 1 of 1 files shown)


bennpearson (commented):
This will be wrapped within {% if analytics_ua_id  %} {% endif %} tags. These have been implemented in #1087 so this will need merging. Maybe best option would be to put this inside app/assets/js/app/modules/analytics.js Lets check with @hamishtaplin

hamishtaplin (changes requested):
I agree with Benn, this should be moved into the existing analytics code. There's a `trackEvent` function in there that should be used instead of the `ga` one. Could probably lose the `try/catch` then too as I think that just guards against errors when `ga` isn't available.

dcdarrell9:
@hamishtaplin tried with dev_mode on Safari and Chrome and works, not sure about testing on other browsers.

collisdigital:
@hamishtaplin where are we at with this one?

hamishtaplin (approved):

PR #1105, opened 2017-04-24
Moved default schemas to 'en' directory

dcdarrell9:
- Moved schemas from app/data to data/en directory within survey-runner 
- Removed some logic from the code
- Changed tests to match the new directory

### How to review 
- Check I haven't missed any schemas/references to schemas
- Check to see if there are any other dependencies
- Run tests/lint


app/utilities/schema.py
```diff
@@ -46,17 +47,13 @@ def load_schema_file(schema_file, language_code=None):
...
-
+    schema_dir = os.path.join(settings.EQ_SCHEMA_DIRECTORY, language_code)
     return os.path.join(schema_dir, schema_file)
 
 
 def available_json_schemas():
     files = []
-    for file in os.listdir(settings.EQ_SCHEMA_DIRECTORY):
-        if os.path.isfile(os.path.join(settings.EQ_SCHEMA_DIRECTORY, file)) and file.endswith(".json"):
+    for file in os.listdir(settings.EQ_SCHEMA_DIRECTORY_DEFAULT):
```
collisdigital: I think there should be a `_get_schema_path(language)` function that returns this value rather than have a new environment var to do it; basically factor line 50 above out into it's own function to be used here and in `_get_schema_file_path`

collisdigital: If you do this, then you can remove `EQ_SCHEMA_DIRECTORY_DEFAULT`

tests/app/framework/survey_runner_test_case.py
```diff
@@ -25,6 +25,6 @@ def setUp(self):
             with c.session_transaction() as sess:
                 sess['user_id'] = '1'
 
-        schema_file = open(os.path.join(settings.EQ_SCHEMA_DIRECTORY, "0_star_wars.json"))
+        schema_file = open(os.path.join(settings.EQ_SCHEMA_DIRECTORY_DEFAULT, "0_star_wars.json"))
```
collisdigital: use the `_get_schema_file_path` instead of constructing manually.

collisdigital (changes requested):
See comments on code

app/utilities/schema.py
dcdarrell9: Changed it so the function is there and the env var is gone, but I don't know whether having `language_code='en'` seems right?

app/utilities/schema.py
```diff
@@ -45,18 +46,18 @@ def load_schema_file(schema_file, language_code=None):
...
+    schema_dir = _get_schema_path(language_code)
     return os.path.join(schema_dir, schema_file)
 
 
 def available_json_schemas():
     files = []
-    for file in os.listdir(settings.EQ_SCHEMA_DIRECTORY):
-        if os.path.isfile(os.path.join(settings.EQ_SCHEMA_DIRECTORY, file)) and file.endswith(".json"):
+    for file in os.listdir(_get_schema_path()):
+        if os.path.isfile(os.path.join(_get_schema_path(), file)) and file.endswith(".json"):
```
collisdigital: Use `_get_schema_file_path` here instead of os.join

tests/app/data/test_schema_id_regex.py
```diff
@@ -14,7 +15,7 @@ def create_schema_with_id(schema_id='answer'):
     :param schema_id: The Id to use for the answer.
     :return: The JSON file with the Id swapped for schema_id
     """
-    json_file = open(os.path.join(settings.EQ_SCHEMA_DIRECTORY, "test_percentage.json"))
+    json_file = open(os.path.join(_get_schema_path, "test_percentage.json"))
```
collisdigital: Use _get_schema_file_path here

tests/app/framework/survey_runner_test_case.py
```diff
@@ -25,6 +26,6 @@ def setUp(self):
             with c.session_transaction() as sess:
                 sess['user_id'] = '1'
 
-        schema_file = open(os.path.join(settings.EQ_SCHEMA_DIRECTORY, "0_star_wars.json"))
+        schema_file = open(os.path.join(_get_schema_path, "0_star_wars.json"))
```
collisdigital: Use _get_schema_file_path here, anywhere else this is happening you could do the same replacement?

dcdarrell9: do you mean replace it with..

`open(_get_schema_file_path, "0_star_wars.json")`

Because this breaks the unit tests

tests/app/framework/survey_runner_test_case.py
```diff
@@ -25,6 +24,6 @@ def setUp(self):
             with c.session_transaction() as sess:
                 sess['user_id'] = '1'
 
-        schema_file = open(os.path.join(settings.EQ_SCHEMA_DIRECTORY, "0_star_wars.json"))
+        schema_file = open(_get_schema_file_path("0_star_wars.json", language_code='en'))
```
ajmaddaford: I think we should remove this loading of a schema and setting of `self.schema_json`. It's only used in `tests/app/routing/test_conditional_display.py`, which should also be removed as the tests it contains are already covered by `tests/app/questionnaire/test_rules.py`.

app/utilities/schema.py
```diff
@@ -6,17 +6,18 @@
 from app import cache
 from app import settings
 
-
 logger = get_logger()
 
+default_language_code = 'en'
```
ajmaddaford: Constant variable names should be all uppercase, so `DEFAULT_LANGUAGE_CODE` here

app/utilities/schema.py
```diff
@@ -45,18 +46,18 @@ def load_schema_file(schema_file, language_code=None):
         raise e
 
 
-def _get_schema_file_path(schema_file, language_code):
-    if language_code != 'en':
-        schema_dir = os.path.join(settings.EQ_SCHEMA_DIRECTORY, language_code)
-    else:
-        schema_dir = settings.EQ_SCHEMA_DIRECTORY
+def _get_schema_path(language_code='en'):
```
ajmaddaford: `language_code` should default to `DEFAULT_LANGUAGE_CODE`

app/utilities/schema.py
```diff
@@ -45,18 +46,18 @@ def load_schema_file(schema_file, language_code=None):
...
+    return os.path.join(settings.EQ_SCHEMA_DIRECTORY, language_code)
 
+
+def _get_schema_file_path(schema_file, language_code):
```
ajmaddaford: `language_code` should default to `DEFAULT_LANGUAGE_CODE`

app/utilities/schema.py
```diff
@@ -45,18 +46,18 @@ def load_schema_file(schema_file, language_code=None):
...
+    schema_dir = _get_schema_path(language_code)
     return os.path.join(schema_dir, schema_file)
 
 
-def available_json_schemas():
+def available_json_schemas(language_code='en'):
```
ajmaddaford: We don't need to have a `language_code` parameter here as we always load the list of available schemas from the default language directory.

app/utilities/schema.py
```diff
@@ -45,18 +46,18 @@ def load_schema_file(schema_file, language_code=None):
...
     files = []
-    for file in os.listdir(settings.EQ_SCHEMA_DIRECTORY):
-        if os.path.isfile(os.path.join(settings.EQ_SCHEMA_DIRECTORY, file)) and file.endswith(".json"):
+    for file in os.listdir(_get_schema_path(language_code)):
+        if os.path.isfile(_get_schema_file_path(file, language_code)) and file.endswith(".json"):
```
ajmaddaford: You shouldn't need to pass `language_code` to either of these methods as they should default to `DEFAULT_LANGUAGE_CODE`

tests/app/data/test_schema_id_regex.py
```diff
@@ -14,7 +13,7 @@ def create_schema_with_id(schema_id='answer'):
     :param schema_id: The Id to use for the answer.
     :return: The JSON file with the Id swapped for schema_id
     """
-    json_file = open(os.path.join(settings.EQ_SCHEMA_DIRECTORY, "test_percentage.json"))
+    json_file = open(_get_schema_file_path("test_percentage.json", language_code='en'))
```
ajmaddaford: If we're using `_get_schema_file_path` outside of `app.utilities.schema` we should make them public (remove the _ from the start of the method name)

tests/app/data/test_schema_id_regex.py
```diff
@@ -32,7 +31,7 @@ def validate_json_against_schema(json_to_validate, schema):
 class TestSchemaIdRegEx(unittest.TestCase):
 
     def setUp(self):
-        schema_file = open(os.path.join(settings.EQ_SCHEMA_DIRECTORY, "schema/schema_v1.json"), encoding="utf8")
+        schema_file = open(_get_schema_file_path("schema/schema_v1.json", language_code='en'), encoding="utf8")
```
ajmaddaford: I'd go back to using the original `os.path.join` as this method won't work for `schema_v1.json` and it doesn't seem right to add a method to our code for loading it when it's only used in the tests.

tests/app/data/test_schema_validation.py
```diff
@@ -135,7 +135,7 @@ def test_child_answers_define_parent(self):
     @staticmethod
     def all_schema_files():
         schema_files = []
-        for folder, _, files in os.walk(settings.EQ_SCHEMA_DIRECTORY):
+        for folder, _, files in os.walk(_get_schema_path()):
             for filename in files:
                 if filename.endswith(".json") and filename != 'schema_v1.json':
```
ajmaddaford: If we move `schema_v1.json` to the `data` directory, then we shouldn't need this logic excluding it.
Also, if we do that, can this method be replaced with a call to `available_json_schemas`?

ajmaddaford (changes requested):

ajmaddaford:
I think we should move the entire data directory to the top level i.e. at the same level as app

collisdigital:
Looks like some of the cy schemas have been comitted by accident. Jenkins may also need updating if the paths have changed?

dcdarrell9:
Should I just remove them? Was it because the .gitignore changed?

app/utilities/schema.py
```diff
@@ -45,18 +46,18 @@ def load_schema_file(schema_file, language_code=None):
         raise e
 
 
-def _get_schema_file_path(schema_file, language_code):
-    if language_code != 'en':
-        schema_dir = os.path.join(settings.EQ_SCHEMA_DIRECTORY, language_code)
-    else:
-        schema_dir = settings.EQ_SCHEMA_DIRECTORY
+def get_schema_path(language_code='en'):
```
ajmaddaford: `'en'` should be replaced with `DEFAULT_LANGUAGE_CODE`

app/utilities/schema.py
```diff
@@ -45,18 +46,18 @@ def load_schema_file(schema_file, language_code=None):
...
+    return os.path.join(settings.EQ_SCHEMA_DIRECTORY, language_code)
+
 
+def get_schema_file_path(schema_file, language_code):
```
ajmaddaford: `language_code` should default to `DEFAULT_LANGUAGE_CODE`

app/utilities/schema.py
```diff
@@ -45,18 +46,18 @@ def load_schema_file(schema_file, language_code=None):
...
+    schema_dir = get_schema_path(language_code)
     return os.path.join(schema_dir, schema_file)
 
 
 def available_json_schemas():
     files = []
-    for file in os.listdir(settings.EQ_SCHEMA_DIRECTORY):
-        if os.path.isfile(os.path.join(settings.EQ_SCHEMA_DIRECTORY, file)) and file.endswith(".json"):
+    for file in os.listdir(get_schema_path(DEFAULT_LANGUAGE_CODE)):
+        if os.path.isfile(get_schema_file_path(file, DEFAULT_LANGUAGE_CODE)) and file.endswith(".json"):
```
ajmaddaford: `get_schema_path` and `get_schema_file_path` both default to `DEFAULT_LANGUAGE_CODE`, so there is no need to pass it here.

tests/app/data/test_schema_id_regex.py
```diff
@@ -14,7 +15,7 @@ def create_schema_with_id(schema_id='answer'):
     :param schema_id: The Id to use for the answer.
     :return: The JSON file with the Id swapped for schema_id
     """
-    json_file = open(os.path.join(settings.EQ_SCHEMA_DIRECTORY, "test_percentage.json"))
+    json_file = open(get_schema_file_path("test_percentage.json", language_code='en'))
```
ajmaddaford: We can drop the `language_code` if we default it in `get_schema_file_path` as recommended above

ajmaddaford (changes requested):

app/utilities/schema.py
dcdarrell9: Does it need to be passed in because it should be defaulted?

jonnyshaw89 (approved):
Looks good to me.

collisdigital (approved):
This looks good now, will need to test language variants in pre-prod before deploying

gulpfile.babel.js
```diff
@@ -198,11 +198,11 @@ gulp.task('lint:json', () => {
...
+  gulp.src(['./data/en/*.json'])
     .pipe(prettify())
-    .pipe(gulp.dest('./app/data/'))
+    .pipe(gulp.dest('./data/en/'))
 
-  gulp.src(['./app/data/schema/schema-v1.json'])
+  gulp.src(['./data/en/schema/schema-v1.json'])
     .pipe(prettify())
-    .pipe(gulp.dest('./app/data/schema'))
+    .pipe(gulp.dest('./data/en/schema'))
```
ajmaddaford: These two directories are wrong, the schema is in `./data/schema`

scripts/translate_schemas.sh
```diff
@@ -2,7 +2,7 @@
 current_dir=$(dirname "${BASH_SOURCE[0]}")
 parent_dir=$(dirname "${current_dir}")
 path_to_parent="$( cd "${parent_dir}" && pwd )"
-schema_dir="${path_to_parent}"/app/data
+schema_dir="${path_to_parent}"/data
 temp_dir="${path_to_parent}"/temp
```
ajmaddaford, line 6: This script also needs to be amended to check out 1.1.0 of eq-translations

ajmaddaford (changes requested):

ajmaddaford (approved):

PR #1111, opened 2017-05-02
Move development default settings from survey runner code to run app

jonnyshaw89:
Removed all settings that default to developer values

### How to review 
Check that all default values in settings.py are not developer values.
Check that developer experience has not changed

### DIFF
--- a/app/settings.py
+++ b/app/settings.py
@@ -2,6 +2,8 @@ import os
 
 from structlog import get_logger
 
+from app.missing_setting_exception import MissingSettingException
+
 logger = get_logger()
 
 
@@ -28,7 +30,15 @@ def read_file(file_name):
         return None
 
 
-EQ_MINIMIZE_ASSETS = parse_mode(os.getenv('EQ_MINIMIZE_ASSETS', 'False'))
+def get_env_or_fail(key):
+    value = os.getenv(key)
+    if value is None:
+        raise MissingSettingException("Setting '{}' Missing".format(key))
+
+    return value
+
+
+EQ_MINIMIZE_ASSETS = parse_mode(os.getenv('EQ_MINIMIZE_ASSETS', 'True'))
 # max request payload size in bytes
 MAX_CONTENT_LENGTH = os.getenv('EQ_MAX_HTTP_POST_CONTENT_LENGTH', 65536)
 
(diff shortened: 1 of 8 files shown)


app/settings.py
```diff
@@ -37,29 +47,28 @@ def read_file(file_name):
...
 EQ_UA_ID = os.getenv('EQ_UA_ID', '')
-EQ_NEW_RELIC_ENABLED = os.getenv("EQ_NEW_RELIC_ENABLED", "False") == "True"
+EQ_NEW_RELIC_ENABLED = parse_mode(os.getenv("EQ_NEW_RELIC_ENABLED", "False"))
 EQ_APPLICATION_VERSION_PATH = '.application-version'
 EQ_APPLICATION_VERSION = read_file(EQ_APPLICATION_VERSION_PATH)
 
 EQ_SERVER_SIDE_STORAGE_ENCRYPTION = parse_mode(os.getenv('EQ_SERVER_SIDE_STORAGE_ENCRYPTION', 'True'))
-EQ_SERVER_SIDE_STORAGE_DATABASE_URL = os.getenv('EQ_SERVER_SIDE_STORAGE_DATABASE_URL', 'sqlite:////tmp/questionnaire.db')
+EQ_SERVER_SIDE_STORAGE_DATABASE_URL = get_env_or_fail('EQ_SERVER_SIDE_STORAGE_DATABASE_URL')
 EQ_SERVER_SIDE_STORAGE_DATABASE_SETUP_RETRY_COUNT = os.getenv('EQ_SERVER_SIDE_STORAGE_DATABASE_CONNECT_RETRY_COUNT', 10)
```
collisdigital, line 67: any value in adding a `get_env_or_default()` to make these consistent now (i.e. that a decision has been made to either default or fail)?

scripts/run_tests_unit.sh
```diff
@@ -31,6 +31,38 @@ if [ -z "$EQ_ENABLE_CACHE" ]; then
...
+
+if [ -z "$EQ_SECRET_KEY" ]; then
+  export EQ_SECRET_KEY="SuperSecretDeveloperKey"
+fi
+
+if [ -z "$EQ_SERVER_SIDE_STORAGE_USER_ID_SALT" ]; then
+  export EQ_SERVER_SIDE_STORAGE_USER_ID_SALT="developer.user.id.salt"
+fi
+
+if [ -z "$EQ_SERVER_SIDE_STORAGE_USER_IK_SALT" ]; then
```
collisdigital: Can this duplication with run_app.sh be removed (e.g. do a `source dev_vars.sh` or something?

collisdigital (changes requested):

app/settings.py
jonnyshaw89: I don't think it would add a lot of value. its basically going to wrap the existing call and add nothing.

collisdigital: Fair enough

collisdigital (approved):

tests/app/test_settings.py
```diff
@@ -24,5 +25,11 @@ def test_missing_get_application_version_from_file(self):
     def test_none_filename_does_not_attempt_to_load_file(self):
         self.assertEqual(None, settings.read_file(None))
 
+    def test_invalid_key_raises_exception(self):
+        with self.assertRaises(MissingSettingException) as exception:
+            settings.get_env_or_fail("InvalidKey")
```
ajmaddaford: Minor point, but it would be clearer if this was something like `MISSING_ENVIRONMENT_VARIABLE` rather than `InvalidKey`.

ajmaddaford (changes requested):

tests/app/test_settings.py
jonnyshaw89: Done

ajmaddaford (approved):

PR #1112, opened 2017-05-02
Check that a file name is provided before attempting to load it.

jonnyshaw89:
Currently on a non EQ_DEV_MODE deployment the Dev keys have a filename of None which the app attempts to load. This causes the App not to start

### How to review 
Test that the code does not attempt to load the file if the name is None

### DIFF
--- a/app/settings.py
+++ b/app/settings.py
@@ -18,12 +18,13 @@ def parse_mode(string):
 
 
 def read_file(file_name):
-    if os.path.isfile(file_name):
+    if file_name and os.path.isfile(file_name):
         logger.debug("reading from file", filename=file_name)
         f = open(file_name, "r")
         contents = f.read()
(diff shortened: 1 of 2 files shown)


ajmaddaford (approved):

app/settings.py
```diff
@@ -18,7 +18,7 @@ def parse_mode(string):
 
 
 def read_file(file_name):
-    if os.path.isfile(file_name):
+    if file_name and os.path.isfile(file_name):
```
collisdigital, line 21: maybe log out on the else?

jonnyshaw89: @collisdigital Done.

collisdigital: Thanks

collisdigital (approved):

PR #1113, opened 2017-05-02
Added tx_id to RabbitMQ header

jonnyshaw89:
I have added the `tx_id` to the RabbitMQ header so that it is possible to identify the transaction id for a message without decrypting/processing it.

### How to review 
Validate that the `tx_id` is in the header of the RabbitMQ message

### DIFF
--- a/app/submitter/submitter.py
+++ b/app/submitter/submitter.py
@@ -9,7 +9,7 @@ logger = get_logger()
 
 class LogSubmitter():  # pylint: disable=no-self-use
 
-    def send_message(self, message, queue):
+    def send_message(self, message, queue, tx_id):  # pylint: disable=unused-argument
         logger.info("sending message")
         logger.info("message payload", message=message, queue=queue)
         return True
@@ -42,7 +42,7 @@ class RabbitMQSubmitter():
         except AMQPError as e:
             logger.error("unable to close connection", exc_info=e, category="rabbitmq")
 
-    def send_message(self, message, queue):
+    def send_message(self, message, queue, tx_id):
         """
         Sends a message to rabbit mq and returns a true or false depending on if it was successful
         :param message: The message to send to the rabbit mq queue
@@ -56,13 +56,17 @@ class RabbitMQSubmitter():
             self._connect()
             channel = self.connection.channel()
             channel.queue_declare(queue=queue, durable=True)
+            properties = BasicProperties(headers={},
+                                         delivery_mode=2)
+
+            if tx_id:
+                properties.headers['tx_id'] = tx_id
+
             published = channel.basic_publish(exchange='',
                                               routing_key=queue,
                                               body=message_as_string,
                                               mandatory=True,
-                                              properties=BasicProperties(
-                                                  delivery_mode=2,
-                                              ))
+                                              properties=properties)
             if published:
                 logger.info("sent message", category="rabbitmq")
             else:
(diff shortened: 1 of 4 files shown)


app/views/questionnaire.py
```diff
@@ -226,7 +227,9 @@ def submit_answers(eq_id, form_type, collection_id, metadata, answer_store):
 
         message = convert_answers(metadata, g.schema_json, answer_store, path_finder.get_routing_path())
         message = current_app.eq['encrypter'].encrypt(message)
-        sent = current_app.eq['submitter'].send_message(message, current_app.config['EQ_RABBITMQ_QUEUE_NAME'])
+        sent = current_app.eq['submitter'].send_message(message,
+                                                        current_app.config['EQ_RABBITMQ_QUEUE_NAME'],
+                                                        metadata["tx_id"])
```
collisdigital: single quotes for dicts, isn't there a pylint rule for this?

tests/app/submitter/test_submitter.py
```diff
@@ -20,15 +20,15 @@ def test_when_fail_to_connect_to_queue_then_published_false(self):
             connection.side_effect = AMQPError()
 
             # When
-            published = self.submitter.send_message(message={}, queue=self.queue_name)
+            published = self.submitter.send_message(message={}, queue=self.queue_name, tx_id=12345)
```
collisdigital: are tx_id values strings or ints?

app/submitter/submitter.py
```diff
@@ -9,7 +9,7 @@
 
 class LogSubmitter():  # pylint: disable=no-self-use
 
-    def send_message(self, message, queue):
+    def send_message(self, message, queue, tx_id):  # pylint: disable=unused-argument
```
collisdigital, line 12: rather than pylint disable, why not log it out?

collisdigital (approved):
Needs testing with a real RabbitMQ server (Docker?)

app/submitter/submitter.py
jonnyshaw89: It is already bound to the logger.

collisdigital: that's annoying then!

tests/app/submitter/test_submitter.py
```diff
@@ -77,7 +77,22 @@ def test_when_fail_to_publish_message_then_returns_false(self):
         connection.channel.side_effect = Mock(return_value=channel)
         with patch('app.submitter.submitter.BlockingConnection', return_value=connection):
             # When
-            published = self.submitter.send_message(message={}, queue=self.queue_name)
+            published = self.submitter.send_message(message={}, queue=self.queue_name, tx_id='12345')
 
             # Then
             self.assertFalse(published, 'send_message should fail to publish message')
+
+    def test_when_message_sent_then_tx_id_is_sent_in_header(self):
```
ajmaddaford, line 85: We should also have a test for when tx_id is not passed into `send_message`

ajmaddaford (changes requested):

tests/app/submitter/test_submitter.py
jonnyshaw89: It shouldn't be None in our normal flow as tx_id has a generator. But i have added a test and the functionality so that we can call `send_message` without one.

ajmaddaford (approved):

Issue #1115, opened 2017-05-02
Can't override the developer page default keys


ajmaddaford:
### Expected behaviour
We should be able to override the keys and password used to generate the JWT via the developer page. 

### Actual behaviour
Developer page keys and password can't be overridden. If they are and the keys are missing from the `jwt-test-keys` directory, the application fails to start.

### Steps to reproduce the behaviour
* Move `sdc-user-authentication-signing-rrm-private-key.pem` to a new location e.g. `mv ./jwt-test-keys/sdc-user-authentication-signing-rrm-private-key.pem ./jwt-test-keys/sdc-user-authentication-signing-rrm-private-key2.pem`
* Export the environment variable EQ_USER_AUTHENTICATION_RRM_PRIVATE_KEY to point to the new location e.g. `export EQ_USER_AUTHENTICATION_RRM_PRIVATE_KEY=./jwt-test-keys/sdc-user-authentication-signing-rrm-private-key2.pem`

* Try starting the application

The application fails to start and the error should look something like: 
`FileNotFoundError: [Errno 2] No such file or directory: './jwt-test-keys/sdc-user-authentication-signing-rrm-private-key.pem'`

PR #1116, opened 2017-05-03
Fix developer page key overriding

ajmaddaford:
Allows overriding of developer page keys and password. 
Fixes https://github.com/ONSdigital/eq-survey-runner/issues/1115

### How to review 
* Move `sdc-user-authentication-signing-rrm-private-key.pem` to a new location e.g. `mv ./jwt-test-keys/sdc-user-authentication-signing-rrm-private-key.pem ./jwt-test-keys/sdc-user-authentication-signing-rrm-private-key2.pem`
* Start the application
* Logs should contain `Did not load file because filename supplied was None or not a file` 
* Try to start a survey, you should get a 500 error
* Export the environment variable EQ_USER_AUTHENTICATION_RRM_PRIVATE_KEY to point to the new location e.g. `export EQ_USER_AUTHENTICATION_RRM_PRIVATE_KEY=./jwt-test-keys/sdc-user-authentication-signing-rrm-private-key2.pem`
* Logs should **not** contain `Did not load file because filename supplied was None or not a file` 
* You should be able to start a survey
* Try similar changes to `EQ_USER_AUTHENTICATION_RRM_PRIVATE_KEY_PASSWORD` and `EQ_USER_AUTHENTICATION_SR_PUBLIC_KEY`

### DIFF
--- a/app/views/dev_mode.py
+++ b/app/views/dev_mode.py
@@ -3,7 +3,7 @@ import re
 import time
 from uuid import uuid4
 
-from flask import Blueprint, redirect, render_template, request
+from flask import Blueprint, redirect, render_template, request, current_app
 from structlog import get_logger
 
 from app.cryptography.jwt_encoder import Encoder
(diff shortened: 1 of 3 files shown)


jonnyshaw89 (approved):

collisdigital:
Kind of feel like the app should die on startup if this is missing, with this implementation it gets all the way into the encryption and blows up with `TypeError: object of type 'NoneType' has no len()` which isn't very safe/clean?

collisdigital (changes requested):
Let's talk about the startup behaviour when the file is missing

jonnyshaw89:
I think i probably agree with @collisdigital on this one. if the app starts Beanstalk will treat it as a successful deployment. If it fails to start then it won't break all the servers.

dcdarrell9 (approved):

ajmaddaford:
I agree we should not start the application if required keys or passwords are missing from the environment (or empty). I made the changes locally to do this, but this would also mean that in production the app wouldn't start as the developer page keys wouldn't be set. I think we should get this pull request merged as is, so that we can resume continual deployments to pre-prod, and create another card for separating out the developer page specific settings. We should also remove the default paths and passwords from the app.

collisdigital (approved):

collisdigital:
Approved

PR #1125, opened 2017-05-08
Refactor modules to use flask config instead of settings.

samiwel:
This PR is a continuation of the work to refactor the code to use Flask's config property rather than accessing properties of the settings.py file directly throughout the code.
This PR is primarily concerned with refactoring:
- rules.py
- encrypted_questionnaire_storgage.py, and
- storage_factory.py

I have refactored surrounding methods where applicable to pass in values (from the flask config) down the call stack.

### How to review 
All tests should pass functionally should behave the same.


app/storage/encrypted_questionnaire_storage.py
```diff
@@ -1,14 +1,13 @@
 import hashlib
```
ajmaddaford, line 1: I think we should drop the support for non-encrypted storage. Developers rarely used it previously, and now don't need to since we added the ability to dump the data via `/dump/answers` and `/dump/submission` in https://github.com/ONSdigital/eq-survey-runner/pull/1025

app/storage/encrypted_questionnaire_storage.py
```diff
@@ -1,14 +1,13 @@
...
-from app import settings
 from app.cryptography.jwe_decryption import JWEDirDecrypter
 from app.cryptography.jwe_encryption import JWEDirEncrypter
 from app.storage.questionnaire_storage import QuestionnaireStorage
 from app.utilities.strings import to_bytes
 from app.utilities.strings import to_str
 
 
-def generate_key(user_id, user_ik, pepper=settings.EQ_SERVER_SIDE_STORAGE_ENCRYPTION_USER_PEPPER):
+def generate_key(user_id, user_ik, pepper=''):
```
samiwel: Probably shouldn't default the pepper values here. Better to change the tests and explicitly pass pepper values in.

app/storage/encrypted_questionnaire_storage.py
```diff
@@ -21,13 +20,14 @@ def generate_key(user_id, user_ik, pepper=settings.EQ_SERVER_SIDE_STORAGE_ENCRYP
 
 class EncryptedQuestionnaireStorage(QuestionnaireStorage):
 
-    def __init__(self, user_id, user_ik):
+    def __init__(self, user_id, user_ik, pepper=''):
```
samiwel: Same as above

app/questionnaire/rules.py
```diff
@@ -44,27 +43,28 @@ def evaluate_goto(goto_rule, metadata, answer_store, group_instance):
     return True
 
 
-def evaluate_repeat(repeat_rule, answer_store):
+def evaluate_repeat(repeat_rule, answer_store, max_repeats=25):
```
samiwel: Need to find best place to pass the default value of `EQ_MAX_NUM_REPEATS` in.

app/storage/storage_factory.py
```diff
@@ -1,10 +1,9 @@
-from app import settings
 from app.storage.encrypted_questionnaire_storage import EncryptedQuestionnaireStorage
 from app.storage.questionnaire_storage import QuestionnaireStorage
 
 
-def get_storage(user_id, user_ik):
-    if settings.EQ_SERVER_SIDE_STORAGE_ENCRYPTION:
-        return EncryptedQuestionnaireStorage(user_id, user_ik)
+def get_storage(user_id, user_ik, use_encrypted_storage=False, pepper=''):
```
samiwel: Same as above. Shouldn't default pepper values.

app/questionnaire/rules.py
ajmaddaford: `max_repeats` should be defined as a constant at the top of this file, rather than passed in, as we never override it.

app/storage/encrypted_questionnaire_storage.py
ajmaddaford: We shouldn't default the pepper to `''`

ajmaddaford (changes requested):

app/storage/encrypted_questionnaire_storage.py
samiwel: Have removed the storage factory and now instantiate the encrypted storage directly. Also removed the environment variable which toggles encrypted storage since it's no longer used.

app/storage/encrypted_questionnaire_storage.py
```diff
@@ -1,14 +1,13 @@
...
-from app import settings
 from app.cryptography.jwe_decryption import JWEDirDecrypter
 from app.cryptography.jwe_encryption import JWEDirEncrypter
 from app.storage.questionnaire_storage import QuestionnaireStorage
 from app.utilities.strings import to_bytes
 from app.utilities.strings import to_str
 
 
-def generate_key(user_id, user_ik, pepper=settings.EQ_SERVER_SIDE_STORAGE_ENCRYPTION_USER_PEPPER):
+def generate_key(user_id, user_ik, pepper):
```
jonnyshaw89: Is there any reason this isn't just a method of `EncryptedQuestionnaireStorage`?

samiwel: Not as far as I can see, I'll pull that method into that class and refactor the tests.

app/storage/encrypted_questionnaire_storage.py
```diff
@@ -1,49 +1,88 @@
...
 
-class EncryptedQuestionnaireStorage(QuestionnaireStorage):
+class EncryptedQuestionnaireStorage:
 
-    def __init__(self, user_id, user_ik):
-        super().__init__(user_id)
+    def __init__(self, user_id, user_ik, pepper):
+        if user_id is None:
+            raise ValueError('User id must be set')
         if user_ik is None:
```
collisdigital, line 21: also check for a valid `pepper`

app/storage/encrypted_questionnaire_storage.py
```diff
@@ -1,49 +1,88 @@
...
 
-    def decrypt_data(self, user_id, user_ik, encrypted_data):
-        sha_key = generate_key(user_id, user_ik)
+    def decrypt_data(self, encrypted_data):
+        sha_key = self.generate_key()
         return self.decryption.decrypt(encrypted_data['data'], sha_key)
+
+    def generate_key(self):
+        sha256 = hashlib.sha256()
+        sha256.update(to_str(self.user_id).encode('utf-8'))
```
collisdigital: maybe we should do this `to_str` and `encode` on `__init__` once rather than here, unless the values are used in their raw form anywhere?

app/storage/encrypted_questionnaire_storage.py
```diff
@@ -1,49 +1,88 @@
...
+        sha256.update(to_str(self.user_ik).encode('utf-8'))
+        sha256.update(to_str(self.pepper).encode('utf-8'))
+
+        # we only need the first 32 characters for the CEK
+        cek = sha256.hexdigest()[:32]
```
collisdigital: in fact, maybe just generate the `cek` on `__init__` and ditch the salt, pepper etc?

collisdigital (changes requested):

app/storage/encrypted_questionnaire_storage.py
jonnyshaw89: and store it as bytes just to save the next line as well?

app/storage/encrypted_questionnaire_storage.py
samiwel: Added the check for a valid pepper and a test to assert the ValueError is raised.

app/storage/encrypted_questionnaire_storage.py
```diff
@@ -1,49 +1,89 @@
...
         encrypted = self.encryption.encrypt(data, sha_key)
         return {'data': encrypted}
 
-    def decrypt_data(self, user_id, user_ik, encrypted_data):
-        sha_key = generate_key(user_id, user_ik)
+    def decrypt_data(self, encrypted_data):
+        sha_key = self.generate_key()
         return self.decryption.decrypt(encrypted_data['data'], sha_key)
+
+    def generate_key(self):
```
jonnyshaw89: this method can be removed altogether if `self.cek = sha256.hexdigest()[:32]` becomes `self.cek = to_bytes(sha256.hexdigest()[:32])`

collisdigital: Or put all this back in the `generate_key` function and call it from `__init__`

samiwel: I'm reluctant to get rid of the `generate_key` method altogether since we have test coverage of this method that will tell us if the key generation is broken, which I think would be good to keep. I'll move the logic back into generate_key and call from __init__.

tests/app/storage/test_questionnaire_storage.py
```diff
@@ -4,37 +4,37 @@
...
         settings.EQ_SERVER_SIDE_STORAGE_DATABASE_URL = "sqlite://"
-        self.storage = QuestionnaireStorage("1")
+        self.storage = EncryptedQuestionnaireStorage("user_id", "user_ik", "pepper")
 
     def test_store(self):
-        data = {'test': 'test'}
+        data = 'test'
         self.assertIsNone(self.storage.add_or_update(data))
-        self.assertIsNotNone(self.storage._get())  # pylint: disable=protected-access
+        self.assertIsNotNone(self.storage._find_questionnaire_state())  # pylint: disable=protected-access
```
ajmaddaford: Any reason for using`_find_questionnaire_state` rather than `get_user_data`?

tests/app/storage/test_questionnaire_storage.py
```diff
@@ -4,37 +4,37 @@
 from sqlalchemy.exc import IntegrityError
 
 from app import settings
-from app.storage.questionnaire_storage import QuestionnaireStorage
+from app.storage.encrypted_questionnaire_storage import EncryptedQuestionnaireStorage
```
ajmaddaford: This test file should be merged into `tests/app/storage/test_encrypted_questionnaire_storage.py` - although not followed consistently, there should be a one to one mapping from code to test files for unit tests.

ajmaddaford (changes requested):

tests/app/storage/test_questionnaire_storage.py
samiwel: Not really, the test was changed as a result of renaming the `_get` method. I'll change this so that it tests against the public interface.

ajmaddaford (approved):

jonnyshaw89 (approved):

jonnyshaw89:
@collisdigital this is waiting on your review.

collisdigital (approved):

PR #1128, opened 2017-05-12
DynamoDB as a database?

jonnyshaw89:
Just a POC to show how we could use DynamoDB to replace Postgres on RDS

### How to review 
Have a play. see what you think.


collisdigital:
The previous discussion about using AWS APIs directly seems to apply here too...

jonnyshaw89:
@collisdigital So it would be feasible if we had VPC Endpoints?

collisdigital:
What does the interface look like to a VPC Endpoint (form an app perspective?)

jonnyshaw89:
@collisdigital maybe I'm confused as to what you are referring to be "previous discussion"

PR #1129, opened 2017-05-12
Reference config from Flask application factory (Part 3)

samiwel:
This PR aims to refactor some of the code to remove a dependency on settings and opt for using Flask config instead.
This particular PR is concerned with refactoring the following modules:
- `converter.py`
- `schema_context.py`, and
- `schema.py`

### How to review 
All existing tests and functionality should continue to pass/behave as expected.

### DIFF
--- a/app/submitter/converter.py
+++ b/app/submitter/converter.py
@@ -3,7 +3,6 @@ from datetime import datetime, timezone
 
 from structlog import get_logger
 
-from app import settings
 from app.helpers.schema_helper import SchemaHelper
 
 logger = get_logger()
@@ -124,11 +123,13 @@ def _get_checkbox_answer_data(answer_store, checkboxes_with_qcode, value):
 
         if option:
             if 'child_answer_id' in option:
-                filtered = answer_store.filter(answer_id=option['child_answer_id'], limit=1)
+                filtered = answer_store.filter(answer_id=option['child_answer_id'])
+
+                assert len(filtered) <= 1, 'Multiple answers found for "{0}"'.format(option['child_answer_id'])
 
                 # if the user has selected 'other' we need to find the value it refers to.
                 # when non-mandatory, the other box value can be empty, in this case we just use its value
-                checkbox_answer_data[option['q_code']] = filtered[0]['value'] if len(filtered) == 1 else option['value']
+                checkbox_answer_data[option['q_code']] = filtered[0]['value'] if len(filtered) >= 1 else option['value']
             else:
                 checkbox_answer_data[option['q_code']] = user_answer
 
@@ -193,8 +194,6 @@ def _build_metadata(metadata):
 def _encode_value(value):
     if isinstance(value, int):
         return str(value)
-    elif isinstance(value, datetime):
-        return value.strftime(settings.SDX_DATE_FORMAT)
     elif isinstance(value, str) and value == '':
         return None
     else:
--- a/app/templating/schema_context.py
+++ b/app/templating/schema_context.py
@@ -1,4 +1,3 @@
-from app import settings
 from app.utilities.date_utils import to_date
 
 
@@ -24,7 +23,7 @@ def _map_alias_to_answers(aliases, answer_store):
     for alias, answer_info in aliases.items():
         matching_answers = []
 
-        answers = answer_store.filter(answer_id=answer_info['answer_id'], limit=settings.EQ_MAX_NUM_REPEATS)
+        answers = answer_store.filter(answer_id=answer_info['answer_id'], limit=True)
 
         for answer in answers:
             safe_answer = json_safe(answer['value'])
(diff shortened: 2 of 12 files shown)


app/submitter/converter.py
```diff
@@ -192,7 +193,7 @@ def _encode_value(value):
     if isinstance(value, int):
         return str(value)
     elif isinstance(value, datetime):
-        return value.strftime(settings.SDX_DATE_FORMAT)
+        return value.strftime(SDX_DATE_FORMAT)
```
jonnyshaw89: Im not sure that this line ever gets hit anymore? as we no longer store datetime objects in the answer store.

samiwel: @jonnyshaw89 Seems we don't have a test which covers it either! I can remove it, do we just store dates as string values now?

jonnyshaw89: We possibly should have a test. Do we know what format we are actually sending dates as?

jonnyshaw89:
I think this last comment got collapsed. 
With regards to the SDX_DATE_FORMAT
"We possibly should have a test. Do we know what format we are actually sending dates as?"

samiwel:
@jonnyshaw89 I looked into it and I think you were right, we don't store dates as `datetime` values in the answer store, at least as far as I can tell.
We currently have support for Day/Month/Year fields or Month/Year fields. They are submitted as `dict`s and [formatted as string values](https://github.com/ONSdigital/eq-survey-runner/blob/master/app/views/questionnaire.py#L382) which get stored against the answer Id in the answer store.

app/templating/schema_context.py
```diff
@@ -1,10 +1,10 @@
-from app import settings
 from app.utilities.date_utils import to_date
 
 
-def build_schema_context(metadata, aliases, answer_store, group_instance=0):
+def build_schema_context(metadata, aliases, answer_store, max_repeats, group_instance=0):
```
adoublebarrel: Maybe worth setting a default for max_repeats as it's currently a limit we have to enforce? Tiny reduction in boiler plate for callers but every little helps.

app/templating/schema_context.py
```diff
@@ -1,6 +1,7 @@
-from app import settings
 from app.utilities.date_utils import to_date
 
+EQ_MAX_NUM_REPEATS = 25
```
collisdigital: Not ideal having this same value defined in multiple places, what happens if they get out of sync...

tests/app/data/test_schema_id_regex.py
```diff
@@ -32,8 +31,10 @@ def validate_json_against_schema(json_to_validate, schema):
 
 class TestSchemaIdRegEx(unittest.TestCase):
 
+    EQ_SCHEMA_DIRECTORY = 'data'
```
collisdigital: Assume we're happy hard-coding this now, before it was runtime configurable?

tests/app/data/test_schema_validation.py
```diff
@@ -19,8 +18,10 @@
 
 class TestSchemaValidation(unittest.TestCase):
 
+    EQ_SCHEMA_DIRECTORY = 'data'
```
collisdigital: Duplicated from before, any way to only define these once?

app/templating/schema_context.py
samiwel: I could revert some of the changes from my latest commit and put the maximum number of repeats value back into`settings.py`? I made this change after discussing with @jonnyshaw89  and @adoublebarrel, since we never override this using environment variables. Whilst using config from flask is the ideal, for this particular refactor, might as well just reference the constant directly (otherwise we have to pass the value down multiple times to get it to where it needs to be used. So as long as everyone's happy with referencing settings file, then I'll take that approach?

jonnyshaw89 (approved):

ajmaddaford (changes requested):
I think we should consider making the limit a property of the answer store, either as a constant or passed on initialisation. It would avoid having to pass the limit 'un-used' through other methods. It also seems like the right place to control the maximum number of answers we support.

samiwel:
@ajmaddaford Andrew, made the changes we discussed. Please check to make sure your happy with the changes. Think we should add a separate card to work on not being able to add > MAX number of answers.

tests/app/forms/test_household_relationship_form.py
```diff
@@ -215,5 +215,5 @@ def test_build_relationship_choices_limited(self):
 
         choices = build_relationship_choices(answer_store, 0)
 
-        self.assertEqual(settings.EQ_MAX_NUM_REPEATS - 1, len(choices))
-
+        # Should be 1 fewer relationship choices than max number of repeats.
+        self.assertEqual(len(choices), 24)
```
ajmaddaford: We should use `answer_store.EQ_MAX_NUM_REPEATS - 1` rather than 24.

ajmaddaford (changes requested):

ajmaddaford (approved):

PR #1133, opened 2017-05-16
Eq 744 cleanse input into survey runner

adoublebarrel:

Reworked the way user input is sanitised so that is clearer and cleaner. Moved all HTML encoding to the point where schema context generation happens and turned off autoescaping for theme templates, see card for more details.

[Trello Card](https://trello.com/c/FsZ1ute6/744-cleanse-input-into-survey-runner)

### How to review 

- All Existing functionality should be intact
- There should be no vectors for XSS via user input

### DIFF
--- a/app/jinja_filters.py
+++ b/app/jinja_filters.py
@@ -1,3 +1,5 @@
+# coding: utf-8
+
 import re
 import string
 
--- a/app/questionnaire/navigation.py
+++ b/app/questionnaire/navigation.py
@@ -2,6 +2,7 @@ import copy
 from collections import defaultdict
 
 from structlog import get_logger
+from jinja2 import escape
 
 from app.helpers.schema_helper import SchemaHelper
 from app.questionnaire.location import Location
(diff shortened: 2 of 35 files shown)


collisdigital (commented):
Is there a way to disable the `safe` filter in jinja2 so we don't use it by accident (or do we actually use it elsewhere)?

adoublebarrel:
Not in a nice way, we could override it in the filters dict of Jinjas environment but that's not very explicit which could cause someone a nasty surprise in the future.

The safe filter wont have any affect in the theme templates now that autoescaping is turned off, as effectively that says we consider the entire Jinja context to be 'safe'. So at worst it's a little extra processing.

We do need to make sure that everyone who is onboard/comes onboard is aware that autoescaping is turned off in the theme templates (because they consume system generated HTML content).  As well as knowing the safe filter is unnecessary it means that introducing user input outside of schema_context generation is a no, no.

ajmaddaford:
There places in the templates we use the metadata context e.g. introduction.html. These values aren't escaped or made json safe. It might also be worth reviewing the duplication of values in schema_context and metadata_context.

ajmaddaford (dismissed):

jonnyshaw89:
screen shot 2017-05-22 at 15 44 45
I Broke it

jonnyshaw89:
I entered `<script>alert('gotcha')</script>` into `test_textfield.json`

jonnyshaw89 (approved):

ajmaddaford (approved):

PR #1136, opened 2017-05-17
Testing travis build stages

jonnyshaw89:
Testing to see if we can use build stages to not repeat unit tests


jonnyshaw89:
I give up

PR #1137, opened 2017-05-17
Load secrets from yml file

jonnyshaw89:
Load the application secrets from a .yml file on disk

### How to review 
Check that no secrets are defined as environment variables


app/__init__.py
```diff
@@ -47,10 +48,16 @@ def __call__(self, environ, start_response):
...
 
-def create_app():  # noqa: C901  pylint: disable=too-complex
+def create_app(setting_overrides=None):  # noqa: C901  pylint: disable=too-complex
     application = Flask(__name__, static_url_path='/s', static_folder='../static')
     application.config.from_object(settings)
 
+    secrets = yaml.safe_load(open(os.getenv('EQ_SECRETS')))
+    application.config.update(secrets)
+
+    if setting_overrides:
```
adoublebarrel: Default setting_overrides to a dict and you wont need the if...

app/__init__.py
```diff
@@ -47,10 +48,16 @@ def __call__(self, environ, start_response):
         return self.app(environ, start_response)
 
 
-def create_app():  # noqa: C901  pylint: disable=too-complex
+def create_app(setting_overrides=None):  # noqa: C901  pylint: disable=too-complex
     application = Flask(__name__, static_url_path='/s', static_folder='../static')
     application.config.from_object(settings)
 
+    secrets = yaml.safe_load(open(os.getenv('EQ_SECRETS')))
```
adoublebarrel: Would this be better off in application.py? Not sure create_app should have knowledge of where/how the secrets are kept.

tests/integration/create_token.py
```diff
@@ -23,24 +22,28 @@
...
-    payload_vars['jti'] = str(uuid4())
-    payload_vars['iat'] = time.time()
-    payload_vars['exp'] = payload_vars['iat'] + float(3600)  # one hour from now
-    for key, value in extra_payload.items():
-        payload_vars[key] = value
+class TokenGenerator:
 
-    return generate_token(payload_vars)
+    def __init__(self, application):
+        self.application = application
```
adoublebarrel: Worth tightening this so that we just pass the private key, private key password and public key to the constructor? (looks like that's all it needs)

collisdigital (approved):
Looks good

adoublebarrel (commented):
Broadly looks good to me, just some minor suggestions.

app/__init__.py
jonnyshaw89: I initially had this in application.py as i thought it would then make testing easier but @ajmaddaford pointed out that it would be better if all of the loading of application config was in the same place.

adoublebarrel: Yes that makes total sense. Longer term I was thinking that we'd move all config loading out of create_app to another place (e.g. application.py)? So create_app would only be concerned with putting the app together based off the config object|dict that's passed to it. That will help with [simplifying create_app](https://trello.com/c/90vVTmqt/425-look-at-reducing-the-complexity-of-create-app-app-init-py).

I'm probably jumping the gun a bit here though, happy with it as is.

adoublebarrel (approved):
LGTM

jonnyshaw89:
No merge until https://github.com/ONSdigital/eq-terraform/pull/81 is in

app/secrets.py
```diff
@@ -0,0 +1,24 @@
...
+    "EQ_RABBITMQ_URL",
+    "EQ_RABBITMQ_URL_SECONDARY",
+    "EQ_SERVER_SIDE_STORAGE_DATABASE_URL",
+]
+
+
+def validate_required_secrets(secrets):
+    for required_secret in EXPECTED_SECRETS:
+        if required_secret not in secrets:
+            raise MissingSecretException("Missing Secret [{}]".format(required_secret))
```
ajmaddaford: I don't think we need to have a `MissingSecretException`, we could just use `Exception`. Also, the required_secret bit doesn't get printed, you just get `app.missing_secret_exception.MissingSecretException`. 

Replacing `MissingSecretException` with `Exception` works and prints something like `Exception: Missing Secret [EQ_SERVER_SIDE_STORAGE_DATABASE_URL]`

app/__init__.py
```diff
@@ -50,10 +52,17 @@ def __call__(self, environ, start_response):
...
```
ajmaddaford: I think we should rename this `EQ_SECRETS_FILE`

ajmaddaford (changes requested):

collisdigital:
Have you tried running this in Docker as well?

jonnyshaw89:
@collisdigital yes I've tried it, but that doesn't even work on master.

jonnyshaw89:
@collisdigital works after docker pull on base image

app/data_model/database.py
```diff
@@ -68,15 +68,17 @@ def __repr__(self):
...
-        self.database_url = database_url
+        self.database_url = "{}://{}:{}@{}".format(database_driver, username, password, database_url) if username and password \
+            else "{}://{}".format(database_driver, database_url)
         self.database_setup_retry_count = database_setup_retry_count
         self.database_setup_retry_delay = database_setup_retry_delay
 
         self._db_session = self._create_session_and_engine_with_retry()
 
     def _create_session_and_engine(self):
+        logger.error("Database Address", database_url=self.database_url)
```
ajmaddaford: Debug logging?

docker-compose.yml
```diff
@@ -31,9 +31,10 @@ services:
       EQ_ENABLE_CACHE: "True"
       EQ_DEVELOPER_LOGGING: "True"
       EQ_RABBITMQ_ENABLED: "True"
-      EQ_SECRET_KEY: "NOT_SECRET_DOCKER_KEY"
-      EQ_SERVER_SIDE_STORAGE_DATABASE_URL: postgres://digitaleq:digitaleq@db:5432/runner
-      EQ_RABBITMQ_URL: amqp://digitaleq:digitaleq@rabbit:5672/%2F
+      EQ_SECRETS_FILE: ".dev_secrets.yml"
+      EQ_SERVER_SIDE_STORAGE_DATABASE_URL: digitaleq:digitaleq@db:5432/runner
```
ajmaddaford: We should use the username and password environment variables here rather than overloading the URL

app/data_model/database.py
```diff
@@ -68,15 +68,17 @@ def __repr__(self):
 
 class Database:
 
-    def __init__(self, database_url, database_setup_retry_count, database_setup_retry_delay):
+    def __init__(self, database_driver, database_url, username, password, database_setup_retry_count, database_setup_retry_delay):
```
ajmaddaford: We're calling two different things `database_url` here. We should rename `database_url` (AWS calls this an endpoint in RDS) or separate into host and port.

app/submitter/submitter.py
```diff
@@ -16,9 +16,11 @@ def send_message(self, message, queue, tx_id):  # pylint: disable=unused-argumen
 
 
 class RabbitMQSubmitter():
-    def __init__(self, rabbitmq_url, rabbitmq_secondary_url):
-        self.rabbitmq_url = rabbitmq_url
-        self.rabbitmq_secondary_url = rabbitmq_secondary_url
+    def __init__(self, rabbitmq_url, rabbitmq_secondary_url, username, password):
+        self.rabbitmq_url = "amqp://{}:{}@{}/%2F".format(username, password, rabbitmq_url) if username and password \
```
ajmaddaford: Same as for the database - we're calling two different things `rabbitmq_url`

tests/app/authentication/test_authenticator.py
```diff
@@ -17,7 +17,7 @@ def setUp(self):
         application.permanent_session_lifetime = timedelta(seconds=1)
         self.application = application
         # Use an in memory database
-        self.database = Database("sqlite://", 1, 0)
+        self.database = Database("sqlite", "", None, None, 1, 0)
```
ajmaddaford: We should use named args when instantiating Database as it's unclear what all the arguments are for. It's also worth considering moving the retry count and retry delay values into constants in database.py

ajmaddaford (changes requested):

docker-compose.yml
jonnyshaw89: They arn't settable via env var any more.

app/__init__.py
```diff
@@ -63,8 +72,11 @@ def create_app():  # noqa: C901  pylint: disable=too-complex
     application.eq = {}
     if application.config['EQ_RABBITMQ_ENABLED']:
         application.eq['submitter'] = RabbitMQSubmitter(
-            application.config['EQ_RABBITMQ_URL'],
-            application.config['EQ_RABBITMQ_URL_SECONDARY'],
+            rabbitmq_host=application.config['EQ_RABBITMQ_HOST'],
+            rabbitmq_secondary_host=application.config['EQ_RABBITMQ_HOST_SECONDARY'],
+            rabbitmq_port=application.config['EQ_RABBITMQ_PORT'],
```
ajmaddaford: I think we should drop the `rabbitmq_` from these params

app/__init__.py
```diff
@@ -77,9 +89,12 @@ def create_app():  # noqa: C901  pylint: disable=too-complex
     )
 
     application.eq['database'] = Database(
-        application.config['EQ_SERVER_SIDE_STORAGE_DATABASE_URL'],
-        application.config['EQ_SERVER_SIDE_STORAGE_DATABASE_SETUP_RETRY_COUNT'],
-        application.config['EQ_SERVER_SIDE_STORAGE_DATABASE_SETUP_RETRY_DELAY_SECONDS'],
+        database_driver=application.config['EQ_SERVER_SIDE_STORAGE_DATABASE_DRIVER'],
+        database_host=application.config['EQ_SERVER_SIDE_STORAGE_DATABASE_HOST'],
+        database_port=application.config['EQ_SERVER_SIDE_STORAGE_DATABASE_PORT'],
+        database_name=application.config['EQ_SERVER_SIDE_STORAGE_DATABASE_NAME'],
```
ajmaddaford: I think we should drop the `database_` from the driver, host and port params. `database_name` should stay though as it's less clear what that is if we drop the `database_` from it.

app/data_model/database.py
```diff
@@ -65,21 +65,36 @@ def __init__(self, jti_claim):
...
+DATABASE_SETUP_ATTEMPTS = 10
+DATABASE_SETUP_RETRY_DELAY_SECONDS = 6
+
 
 class Database:
 
-    def __init__(self, database_url, database_setup_retry_count, database_setup_retry_delay):
+    def __init__(self, database_driver, database_name, database_host=None, database_port=None,
+                 username=None, password=None, database_setup_attempts=DATABASE_SETUP_ATTEMPTS,
+                 database_setup_retry_delay=DATABASE_SETUP_RETRY_DELAY_SECONDS):
```
ajmaddaford: I think we should also drop `database_` from the attempts and delay

app/submitter/submitter.py
```diff
@@ -16,9 +16,19 @@ def send_message(self, message, queue, tx_id):  # pylint: disable=unused-argumen
...
+                                                                                    host=rabbitmq_host,
+                                                                                    port=rabbitmq_port) \
+            if username and password else "amqp://{host}:{port}/%2F".format(host=rabbitmq_host,
+                                                                            port=rabbitmq_port)
+        self.rabbitmq_secondary_url = "amqp://{username}:{password}@{host}:{port}/%2F".format(username=username,
+                                                                                              password=password,
+                                                                                              host=rabbitmq_secondary_host,
+                                                                                              port=rabbitmq_port) \
+            if username and password else "amqp://{host}:{port}/%2F".format(host=rabbitmq_secondary_host,
+                                                                            port=rabbitmq_port)
```
ajmaddaford: I think this would read more easily if the username and password conditional was moved to the top level, something like:
```
if username and password:
    ...
else:
    ...
```

docker-compose.yml
```diff
@@ -31,9 +31,10 @@ services:
...
+      EQ_RABBITMQ_URL: digitaleq:digitaleq@rabbit:5672
+      EQ_RABBITMQ_URL_SECONDARY: digitaleq:digitaleq@rabbit:5672
```
ajmaddaford: I think these need to be replaced with the new host and port environment variables

ajmaddaford (changes requested):

tests/app/data_model/test_database_setup.py
```diff
@@ -47,11 +52,32 @@ def test_database_retry_success_after_attempts(self):
...
+                                database_port=5432,
+                                username="testUsername",
+                                password="testPassword",
+                                database_name="digitaleqrds",
+                                database_setup_attempts=1,
+                                database_setup_retry_delay=0)
+
+            # pylint: disable=protected-access
+            # Then a value is successfully returned
+            self.assertEqual(database._database_url, 'postgresql://testUsername:testPassword@localhost:5432/digitaleqrds')
```
ajmaddaford: If the database url generation was factored out into it's own method, we wouldn't need to mock the database to test it.

tests/app/submitter/test_submitter.py
```diff
@@ -9,10 +9,11 @@
 class TestSubmitter(TestCase):
     def setUp(self):
         self.queue_name = 'test_queue'
-        self.url1 = 'amqp://localhost:5672/%2F'
-        self.url2 = 'amqp://localhost:5672/%2F'
+        self.url1 = 'localhost'
+        self.url2 = 'localhost'
```
ajmaddaford: These should be host1 and host2

ajmaddaford (changes requested):

ajmaddaford (approved):

scripts/generate_secrets.py
```diff
@@ -0,0 +1,48 @@
...
+        if trim:
+            data = data.rstrip('\r\n')
+    return data
+
+
+keys_folder = sys.argv[1]
+
+keys = {}
+
+keys['EQ_SUBMISSION_SDX_PUBLIC_KEY'] = get_file_contents(keys_folder + 'sdc-submission-encryption-sdx-public-key.pem')
```
ajmaddaford: Path concatenation should be done with `os.path.join`, this will provide more flexibility with what input will be allowed (currently the trailing slash is required). Also, it's probably worth doing the concatenation in `get_file_contents` as currently every method is passing keys_folder.

scripts/generate_secrets.py
```diff
@@ -0,0 +1,48 @@
...
+
+def get_file_contents(filename, trim=False):
+    with open(filename, 'r') as f:
+        data = f.read()
+        if trim:
+            data = data.rstrip('\r\n')
+    return data
+
+
+keys_folder = sys.argv[1]
```
ajmaddaford: It's probably worth printing an error and exiting if no arg is passed

ajmaddaford (changes requested):

ajmaddaford (approved):

PR #1141, opened 2017-05-22
move create_app() out of __init__.py

jonnyshaw89:
Moving create_app() out of __init__.py means that required settings don't need to be passed when running most unit tests.

This make testing individual test in PyCharm a lot easier

### How to review 
Test everything works as normal. Check that unit tests that do no start the app don't require the mandatory settings to be set.


app/setup.py
```diff
@@ -0,0 +1,266 @@
...
+
+
+def get_minimized_asset(filename):
+    """
+    If we're in production and it's a js or css file, return the minified version.
+    :param filename: the original filename
+    :return: the new file name will be .min.css or .min.js
+    """
+    if settings.EQ_MINIMIZE_ASSETS:
+        if 'css' in filename:
```
collisdigital, line 262: This looks dangerous, shouldn't it use `endswith`?

jonnyshaw89: Probably, but EQ_MINIMIZE_ASSETS doesn't actually work anyway. and isn't really the context of this change.

collisdigital: Add at tech debt card to either remove or fix this then?

jonnyshaw89: https://trello.com/c/jK5F7hgr/432-enable-asset-minification-by-default ?

collisdigital: :+1: I added some more text to that card

ajmaddaford (approved):

dcdarrell9 (approved):

PR #1145, opened 2017-05-25
Improves Other textfield answer display on summary screen

yoganandkunche:
Improves Other text field answer display on summary screen

### How to review 
Launch test_checkbox and test_radio schemas
Add values to the Other 'text field' and confirm 
That the display on the summary screen is as per the new designs attached to the trello card.

https://trello.com/c/LatUHa9o/1066-other-answers-on-summary-screen

### DIFF
--- a/app/templates/partials/summary/checkbox.html
+++ b/app/templates/partials/summary/checkbox.html
@@ -1,5 +1,18 @@
+{% set not_answered = 'No answer provided' %}
 <ul class="u-m-no">
-  {% for selected_answer in answer.value %}
-    <li>{{selected_answer}}</li>
-  {% endfor %}
+      {% for selected_answer in answer.value %}
+        <li>{{selected_answer.label}}
+            {%- if answer.child_answer_value is not none and selected_answer.should_display_other %}
+            <ul>
+                <li>
+                    {%- if answer.child_answer_value == ''  -%}
+                        {{not_answered}}
+                    {%- else -%}
+                        {{ answer.child_answer_value }}
+                    {%- endif -%}
+                </li>
+            </ul>
+            {%- endif -%}
+        </li>
+      {% endfor %}
 </ul>
--- a/app/templates/partials/summary/radio.html
+++ b/app/templates/partials/summary/radio.html
@@ -1,0 +1,18 @@
+{% set not_answered = 'No answer provided' %}
+<ul class="u-m-no">
+    <li>{{answer.value}}
+    {%- if answer.child_answer_value is not none -%}
+        <ul>
+            <li>
+                {%- if answer.child_answer_value == ''  -%}
+                    {{not_answered}}
+                {%- else -%}
+                    {{ answer.child_answer_value }}
+                {%- endif -%}
+            </li>
+        </ul>
+        {%- endif -%}
+    </li>
+</ul>
+
+
--- a/app/templates/summary.html
+++ b/app/templates/summary.html
@@ -47,26 +47,24 @@
                 </div>
 
                 {% for answer in question.answers %}
+                      {% if question.answers|length > 1 and answer.label %}
+                        <div class="summary__question summary__question--sub">
+                          {{answer.label|striptags}}
+                        </div>
+                      {% endif %}
 
-                  {% if question.answers|length > 1 %}
-                    <div class="summary__question summary__question--sub">
-                      {{answer.label|striptags}}
-                    </div>
-                  {% endif %}
-
-                  <div class="summary__answer">
-                    <div class="summary__answer-text" id="{{answer.id}}-answer" data-qa="{{answer.id}}-answer">
-                      {%- if answer.value is none or answer.value == '' -%}
-                        {{not_answered}}
-                      {%- else -%}
-                        {% include theme(['partials/summary/' ~ answer.type ~ '.html', 'partials/summary/default.html']) %}
-                      {%- endif -%}
-                    </div>
-                    <div class="summary__edit">
-                      <a href="{{ section.link }}#{{answer.id}}" class="summary__edit-link" aria-describedby="{{answer.id}} {{answer.id}}-answer" data-qa="{{answer.id}}-edit" {{helpers.track('click', 'Summary', 'Edit click')}}>Edit <span class="u-vh">your answer</span></a>
-                    </div>
-                  </div>
-
+                      <div class="summary__answer">
+                        <div class="summary__answer-text" id="{{answer.id}}-answer" data-qa="{{answer.id}}-answer">
+                          {%- if answer.value is none or answer.value == '' -%}
+                            {{not_answered}}
+                          {%- else -%}
+                            {% include theme(['partials/summary/' ~ answer.type ~ '.html', 'partials/summary/default.html']) %}
+                          {%- endif -%}
+                        </div>
+                        <div class="summary__edit">
+                          <a href="{{ section.link }}#{{answer.id}}" class="summary__edit-link" aria-describedby="{{answer.id}} {{answer.id}}-answer" data-qa="{{answer.id}}-edit" {{helpers.track('click', 'Summary', 'Edit click')}}>Edit <span class="u-vh">your answer</span></a>
+                        </div>
+                      </div>
                 {% endfor %}
 
             {% endfor %}
--- a/app/templating/summary/answer.py
+++ b/app/templating/summary/answer.py
@@ -1,7 +1,9 @@
 class Answer:
 
-    def __init__(self, answer_schema, answer):
+    def __init__(self, answer_schema, answer, child_answer_value=None):
         self.id = answer_schema['id']
-        self.label = answer_schema['label'] if 'label' in answer_schema else None
+        self.label = answer_schema.get('label')
         self.value = answer
         self.type = answer_schema['type'].lower()
+        self.parent_answer_id = answer_schema.get('parent_answer_id')
+        self.child_answer_value = child_answer_value
--- a/app/templating/summary/question.py
+++ b/app/templating/summary/question.py
@@ -1,3 +1,5 @@
+import collections
+
 from app.questionnaire.rules import evaluate_rule
 from app.data_model.answer_store import iterate_over_instance_ids
 from app.templating.summary.answer import Answer
@@ -33,18 +35,31 @@ class Question:
         summary_answers = []
         answers_iterator = iter(answer_schema)
         for answer_schema in answers_iterator:
+            if 'parent_answer_id' in answer_schema:
+                continue
             if question_schema['type'] == 'RepeatingAnswer':
                 for answer_id, answer_index in iterate_over_instance_ids(answers.keys()):
                     if answer_schema['id'] == answer_id:
                         key = answer_id if answer_index == 0 else '_'.join([answer_id, str(answer_index)])
                         answer = cls._build_answer(question_schema, answer_schema, answers, answers_iterator, key)
                         summary_answers.append(Answer(answer_schema, answer))
-
             else:
                 answer = cls._build_answer(question_schema, answer_schema, answers, answers_iterator)
-                summary_answers.append(Answer(answer_schema, answer))
+                child_answer_value = cls._find_other_value_in_answers(answer_schema, answer, answers)
+                summary_answers.append(Answer(answer_schema, answer, child_answer_value))
 
         return summary_answers
+
+    @classmethod
+    def _find_other_value_in_answers(cls, answer_schema, answer, answers):
+        if answer is not None and 'options' in answer_schema:
+            options = answer_schema['options']
+            for option in options:
+                if option['label'] in [a[0] for a in answer] or option['label'] == answer:
+                    for child_answer in answers:
+                        if 'child_answer_id' in option and child_answer == option['child_answer_id']:
+                            return answers[child_answer]
+        return None
 
     @classmethod
     def _build_answer(cls, question_schema, answer_schema, answers, answers_iterator, answer_id=None):
@@ -65,20 +80,16 @@ class Question:
     @classmethod
     def _build_checkbox_answers(cls, answer, answer_schema):
         multiple_answers = []
+        CheckboxSummaryAnswer = collections.namedtuple('CheckboxSummaryAnswer', 'label should_display_other')
         for option in answer_schema['options']:
-            if option['value'] in answer:
-                if option['value'] == 'other':
-                    summary_option_display_value = cls._get_checkbox_other_display_value(answer, answer_schema, option)
-                    multiple_answers.append(summary_option_display_value)
+            if option['label'] in answer:
+                if option['value'].lower() == 'other':
+                    summary_option_display_value = option['label']
+                    multiple_answers.append(CheckboxSummaryAnswer(label=summary_option_display_value,
+                                                                  should_display_other=True))
                 else:
-                    multiple_answers.append(option['label'])
+                    multiple_answers.append(CheckboxSummaryAnswer(label=option['label'], should_display_other=False))
         return multiple_answers
-
-    @staticmethod
-    def _get_checkbox_other_display_value(answer, answer_schema, option):
-        options = {option['value'] for option in answer_schema['options']}
-        other_option_input = [option for option in set(answer) - set(options) if option.strip()]
-        return option['label'] if not other_option_input else other_option_input.pop()
 
     @staticmethod
     def _build_data_range_answer(answer, answers, answers_iterator):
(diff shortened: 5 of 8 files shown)


armstrongb:
The `summary__question summary__question--sub` which produces the blank box above the list should be removed if the `summary__question--sub` is unspecified.

ons-checkbox-other-not-specified-01

armstrongb:
The output code is not quite correct:
```
<ul class="u-m-no">
    <li>Cheese</li>
    <li>Pepperoni</li>
    <li>Other</li>
<ul>
    <li>No answer provided</li>
</ul>
</ul>
```
It should be nested as follows:

```
<ul class="u-m-no">
    <li>Cheese</li>
    <li>Pepperoni</li>
    <li>Other
        <ul>
            <li>No answer provided</li>
        </ul>
    </li>
</ul>
```

armstrongb (changes requested):
Correction of list code structure as per comment.

app/templating/summary/answer.py
```diff
@@ -1,7 +1,9 @@
 class Answer:
 
-    def __init__(self, answer_schema, answer):
+    def __init__(self, answer_schema, answer, other_value=None):
```
samiwel: Just wondering, rather than `other_value`, perhaps this should be `child_answer_value` so that it fits in better with the new nomenclature.

app/templating/summary/answer.py
```diff
@@ -1,7 +1,9 @@
...
         self.id = answer_schema['id']
         self.label = answer_schema['label'] if 'label' in answer_schema else None
         self.value = answer
         self.type = answer_schema['type'].lower()
+        self.parent_answer_id = answer_schema['parent_answer_id'] if 'parent_answer_id' in answer_schema else None
+        self.other_value = other_value
```
samiwel: Same as above. Make sense to call it `child_answer_value` so it's clearer from the schema what's being talked about.

app/templating/summary/question.py
```diff
@@ -35,14 +35,27 @@ def _build_answers(cls, question_schema, answer_schema, answers):
                         key = answer_id if answer_index == 0 else '_'.join([answer_id, str(answer_index)])
                         answer = cls._build_answer(question_schema, answer_schema, answers, answers_iterator, key)
                         summary_answers.append(Answer(answer_schema, answer))
-
             else:
                 answer = cls._build_answer(question_schema, answer_schema, answers, answers_iterator)
-                summary_answers.append(Answer(answer_schema, answer))
+                other_value = cls._find_other_value_in_answers(answer_schema, answer, answers)
```
samiwel: `find_child_answer_value`?

yoganandkunche:
Addressed all the above comments.

yoganandkunche:
Addressed all the above comments.

dcdarrell9:
test_checkbox and test_radio schemas LGTM 👍

app/templates/partials/summary/checkbox.html
```diff
@@ -1,5 +1,21 @@
...
 <ul class="u-m-no">
   {% for selected_answer in answer.value %}
-    <li>{{selected_answer}}</li>
+    <li>{{selected_answer[0]}}
+    {%- if answer.child_answer_value is none -%}
+        {%- else -%}
+          {% if selected_answer[1] %}
+            {%- if answer.child_answer_value == ''  -%}
+              <ul>
+                <li>{{not_answered}}</li>
```
jonnyshaw89: You could move this `%- if answer.child_answer_value == ''  -%}` to inside the `<li>`?
That would mean that the `<ul><li>` is not duplicated in the if/else

app/templates/partials/summary/radio.html
```diff
@@ -0,0 +1,19 @@
+{% set not_answered = 'No answer provided' %}
+<ul class="u-m-no">
+    <li>{{answer.value}}
+    {%- if answer.child_answer_value is none -%}
+        {%- else -%}
+            {%- if answer.child_answer_value == ''  -%}
+              <ul>
+                <li>{{not_answered}}</li>
```
jonnyshaw89: Same comment as checkbox

app/templating/summary/question.py
```diff
@@ -39,14 +39,27 @@ def _build_answers(cls, question_schema, answer_schema, answers):
...
+                child_answer_value = cls._find_other_value_in_answers(answer_schema, answer, answers)
+                summary_answers.append(Answer(answer_schema, answer, child_answer_value))
 
         return summary_answers
 
     @classmethod
+    def _find_other_value_in_answers(cls, answer_schema, answer, answers):
+        if answer is None:
+            return None
+        elif answer_schema['type'] == 'Checkbox' or answer_schema['type'] == 'Radio':
```
jonnyshaw89: Would this work if it was `elif  'options' in answer_schema:`?

app/templating/summary/question.py
```diff
@@ -39,14 +39,27 @@ def _build_answers(cls, question_schema, answer_schema, answers):
...
+            options = answer_schema['options']
+            for option in options:
+                if option['label'] in [a[0] for a in answer] or option['label'] == answer and 'child_answer_id' in option:
+                    for child_answer in answers:
+                        if 'child_answer_id' in option and child_answer == option['child_answer_id']:
```
jonnyshaw89: `if 'child_answer_id' in option` you already know this to be true from the previous if

app/templating/summary/question.py
```diff
@@ -39,14 +39,27 @@ def _build_answers(cls, question_schema, answer_schema, answers):
...
+                            return answers[child_answer]
```
jonnyshaw89: just return `child_answer`?

app/templating/summary/question.py
```diff
@@ -39,14 +39,27 @@ def _build_answers(cls, question_schema, answer_schema, answers):
...
         return summary_answers
 
     @classmethod
+    def _find_other_value_in_answers(cls, answer_schema, answer, answers):
+        if answer is None:
+            return None
+        elif answer_schema['type'] == 'Checkbox' or answer_schema['type'] == 'Radio':
+            options = answer_schema['options']
+            for option in options:
+                if option['label'] in [a[0] for a in answer] or option['label'] == answer and 'child_answer_id' in option:
```
jonnyshaw89: This statement is quite long and hard to work out what its doing.

app/templating/summary/question.py
```diff
@@ -67,11 +80,11 @@ def _build_checkbox_answers(cls, answer, answer_schema):
         multiple_answers = []
         for option in answer_schema['options']:
             if option['value'] in answer:
-                if option['value'] == 'other':
+                if option['value'].lower() == 'other':
```
jonnyshaw89: you should probably add the `.lower()` to radio also. this will keep the behaviour consistent

app/templating/summary/question.py
```diff
@@ -67,11 +80,11 @@ def _build_checkbox_answers(cls, answer, answer_schema):
...
                     summary_option_display_value = cls._get_checkbox_other_display_value(answer, answer_schema, option)
-                    multiple_answers.append(summary_option_display_value)
+                    multiple_answers.append((summary_option_display_value, True))
```
jonnyshaw89: I can't easily see why the booleans have been added?

jonnyshaw89 (changes requested):

armstrongb:
The output code when using the checkbox schema to test looks good to me.

armstrongb (approved):

app/templating/summary/question.py
samiwel: The Boolean values indicate if this option is the "other" option.  If you look at the surrounding `if` statement then really those boolean values are equivalent to `option['value'].lower() == 'other'` but I agree, it's not immediately obvious. Perhaps this should be refactored by extracting a variable?

app/templating/summary/question.py
samiwel: I don't think so, `child_answer`refers to a dictionary _key_ in this case. The return value would be the _value_ associated with the `child_answer` key

app/templating/summary/question.py
yoganandkunche: Made the statement short by checking 'child_answer_id' in option only once

app/templating/summary/question.py
yoganandkunche: addressed it and now I am only checking it once.

app/templating/summary/question.py
yoganandkunche: Made this change to radio as well.

app/templating/summary/question.py
yoganandkunche: it worked and I made the change.

app/templates/partials/summary/radio.html
yoganandkunche: Made this change

app/templates/partials/summary/checkbox.html
yoganandkunche: Made this change.

app/templates/partials/summary/checkbox.html
```diff
@@ -1,5 +1,21 @@
+{% set not_answered = 'No answer provided' %}
 <ul class="u-m-no">
-  {% for selected_answer in answer.value %}
-    <li>{{selected_answer}}</li>
-  {% endfor %}
+      {% for selected_answer in answer.value %}
+        <li>{{selected_answer[0]}}
+        {%- if answer.child_answer_value is none -%}
+            {%- else -%}
+              {% if selected_answer[1] %}
```
jonnyshaw89: The 3 lines above can probably be 1?
something like
`{%- if answer.child_answer_value is not none and selected_answer[1] %}`?

app/templates/partials/summary/radio.html
```diff
@@ -0,0 +1,19 @@
+{% set not_answered = 'No answer provided' %}
+<ul class="u-m-no">
+    <li>{{answer.value}}
+    {%- if answer.child_answer_value is none -%}
+        {%- else -%}
```
jonnyshaw89: redundant else. just reverse the if statement

app/templating/summary/question.py
```diff
@@ -39,14 +39,27 @@ def _build_answers(cls, question_schema, answer_schema, answers):
...
+                child_answer_value = cls._find_other_value_in_answers(answer_schema, answer, answers)
+                summary_answers.append(Answer(answer_schema, answer, child_answer_value))
 
         return summary_answers
 
     @classmethod
+    def _find_other_value_in_answers(cls, answer_schema, answer, answers):
+        if answer is None:
+            return None
+        elif 'options' in answer_schema:
```
jonnyshaw89: this could be 
`if answer is not None and 'options' in answer_schema:`

jonnyshaw89 (changes requested):

app/templates/partials/summary/checkbox.html
```diff
@@ -1,5 +1,18 @@
...
+                  <ul>
+                    <li>
+                        {%- if answer.child_answer_value == ''  -%}
+                            {{not_answered}}
+                        {%- else -%}
+                            {{ answer.child_answer_value }}
+                    </li>
+                  </ul>
+                        {%- endif -%}
+            {%- endif -%}
```
jonnyshaw89: these don't look correct

app/templates/partials/summary/radio.html
```diff
@@ -0,0 +1,18 @@
...
+              <ul>
+                <li>
+                    {%- if answer.child_answer_value == ''  -%}
+                        {{not_answered}}
+                    {%- else -%}
+                        {{ answer.child_answer_value }}
+                </li>
+              </ul>
+                    {%- endif -%}
+        {%- endif -%}
```
jonnyshaw89: nor do these

jonnyshaw89 (changes requested):

jonnyshaw89 (approved):

app/templates/summary.html
```diff
@@ -47,26 +47,26 @@ <h2 class="summary__title saturn" id="{{group.id}}">{{group.title}}</h2>
...
-                      {%- else -%}
-                        {% include theme(['partials/summary/' ~ answer.type ~ '.html', 'partials/summary/default.html']) %}
-                      {%- endif -%}
-                    </div>
-                    <div class="summary__edit">
-                      <a href="{{ section.link }}#{{answer.id}}" class="summary__edit-link" aria-describedby="{{answer.id}} {{answer.id}}-answer" data-qa="{{answer.id}}-edit" {{helpers.track('click', 'Summary', 'Edit click')}}>Edit <span class="u-vh">your answer</span></a>
-                    </div>
-                  </div>
-
+                    {% if not answer.parent_answer_id %}
```
ajmaddaford: We should not append answers with a `parent_answer_id` in the `_build_answers` method in the summary question.py to avoid this if statement in the template

app/templating/summary/answer.py
```diff
@@ -1,7 +1,9 @@
 class Answer:
 
-    def __init__(self, answer_schema, answer):
+    def __init__(self, answer_schema, answer, child_answer_value=None):
         self.id = answer_schema['id']
         self.label = answer_schema['label'] if 'label' in answer_schema else None
         self.value = answer
         self.type = answer_schema['type'].lower()
+        self.parent_answer_id = answer_schema['parent_answer_id'] if 'parent_answer_id' in answer_schema else None
```
ajmaddaford: We should use the dict get method here - `self.parent_answer_id = answer_schema.get('parent_answer_id')` - it will default to None if the key doesn't exist. Same for the existing `answer_schema['label']` above.

app/templating/summary/question.py
ajmaddaford: We should use a dict here rather than a tuple so we can label the values and replace the references in the templates of `0` and `1` with keys e.g. `multiple_answers.append({'label': option['label'], 'is_other_answer': False})`.  This would make this code and the templates much clearer.

app/templating/summary/question.py
```diff
@@ -67,11 +78,11 @@ def _build_checkbox_answers(cls, answer, answer_schema):
         multiple_answers = []
         for option in answer_schema['options']:
             if option['value'] in answer:
-                if option['value'] == 'other':
+                if option['value'].lower() == 'other':
                     summary_option_display_value = cls._get_checkbox_other_display_value(answer, answer_schema, option)
```
ajmaddaford: Do we ever use anything other than the label for this? i.e. is a call to `_get_checkbox_other_display_value` needed, or can we just use `option['label']`?

ajmaddaford (changes requested):

app/templating/summary/question.py
samiwel: Agree that dict would provide better clarity, but Yogi and I tried this last week and a simple dict did not guarantee ordering of answers on Summary screen. So we tried using an ordered dict, but ran into some other problem higher up in the call stack when using a dict (I can't remember what the problem was now, sorry - perhaps Yogi remembers?)

yoganandkunche: I think the ordered dict was displaying the selected values in random order on the summary page.

app/templating/summary/question.py
samiwel: Until last week it was never even getting into  `_get_checkbox_other_display_value` due to not performing case-insensitive match on `option['value']` so I'd suggest it can be replaced with option['label'] as you mention. Particularly now that the requirement seems to be display "Other - The other value"

app/templating/summary/answer.py
yoganandkunche: Made the changes.

app/templating/summary/question.py
yoganandkunche: Made the changes.

yoganandkunche:
@ajmaddaford replaced the references in the templates using a namedtuple, please let me know if this is ok?

app/templating/summary/question.py
ajmaddaford: If we don't need the `_get_checkbox_other_display_value` method we should remove it and just use the label

app/templating/summary/question.py
```diff
@@ -65,13 +80,15 @@ def _build_answer(cls, question_schema, answer_schema, answers, answers_iterator
     @classmethod
     def _build_checkbox_answers(cls, answer, answer_schema):
         multiple_answers = []
+        CheckboxSummaryAnswer = collections.namedtuple('CheckboxSummaryAnswer', 'label should_display_other')
         for option in answer_schema['options']:
-            if option['value'] in answer:
-                if option['value'] == 'other':
+            if option['label'] in answer:
+                if option['value'].lower() == 'other':
                     summary_option_display_value = cls._get_checkbox_other_display_value(answer, answer_schema, option)
```
ajmaddaford: I think this has got lost in the previous comments. If we don't need the `_get_checkbox_other_display_value` method we should remove it and just use the label.

ajmaddaford (changes requested):

ajmaddaford (approved):

PR #1151, opened 2017-06-07
Check if pre-prod secrets exist

jonnyshaw89:
Check if the pre-production secrets exist before loading them into the app.
In Prod we don't want to attempt to load these secrets as they won't be there.

### How to review 
Check that Secrets are only loaded if they files are present.


ajmaddaford (approved):

adoublebarrel (approved):
LGTM

PR #1152, opened 2017-06-12
BrowserStack integration

jonnyshaw89:
Added capability to run functional tests via BrowserStack from the command line. Using the BrowserStack local tunnel.

### How to review 

Trigger the tests locally by running the following command.

`BROWSERSTACK_USER=#### BROWSERSTACK_ACCESS_KEY=#### yarn test_functional_browserstack`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ language: node_js
 node_js: 6
 env:
   global:
-    - EQ_ENABLE_CACHE=True
     - EQ_RUN_LOCAL_LINT=True
     - EQ_RUN_LOCAL_TESTS=True
     - EQ_RUN_LOCAL=True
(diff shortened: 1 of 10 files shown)


ajmaddaford:
This is proving to be un-reliable - sometimes the tests don't run, and sometimes they run multiple times on BrowserStack. Also the new functional tests don't seem to work - they launch locally in Chrome. I think we should park this and pick it back up once all the functional tests are in one place. We should also consider the effort it would take to move to ChimpJS as this may prove more reliable than the code we've written to do this.

ajmaddaford:
I've thought some more about this, given that it doesn't break any existing functionality and it gives us a way to run functional tests against BrowserStack I think we should merge this.

ajmaddaford (approved):

adoublebarrel (approved):

PR #1157, opened 2017-06-16
Eq 1029 freetext max characters

adoublebarrel:
[Trello Card](https://trello.com/c/LVWhB7Td/1029-freetext-max-characters-s)

Three main areas have been changed:

- Additions to schema_v1.json; I'm not sure if I'm allowed to just add stuff without bumping a version?
- Frontend, template changes; a new JS module and associated SCSS
- Backend; new validator and update to textarea field so that it always does a max length check (defaults to 2000 characters)

### How to review 

The test_textarea schema has been updated with a custom limit of 20 characters, you can select it from the dev page. There are 3 main scenarios:

- IE8 + no-js; No change to the display of the text area, too many chars fails validation on submission
- IE9; "Characters remaining" displayed and can't go negative. This is a change in behaviour resulting from enforcing double counting of newlines in the browser.

- Chrome + Others: "Characters remaining" displayed, can't add characters past the limit

This is my first PR with frontend on this project so you may have to cast a more critical eye over those parts than normal.


yoganandkunche:
@adoublebarrel Looks like the application is throwing an error even when the max char limit is not reached. (PFA) 

It throws an error when Characters remaining limit start to go under 25

screen shot 2017-06-16 at 14 58 41

adoublebarrel:
Oof, thanks, I'll have a look at that.

yoganandkunche:
@adoublebarrel Char count does not work on Firefox

screen shot 2017-06-16 at 15 19 58

adoublebarrel:
@yoganandkunche awesome catches, I've made fixes which should resolve both those issues

ajmaddaford:
I think we need to do something so that the default max length is only defined in one place (currently it's defined in the validator and the template). Some options I've discussed with Alex are:
- Enforcing the max length property in the schema. This doesn't fulfil the acceptance criteria that states it should be configurable, but default to a defined limit.
- Adding a post-processing step to the schema parsing to add a max_length property to any text areas that don't have one already. It feels a bit premature to introduce this without the context of more examples.
- Creating a custom TextArea field that has a max length property. This property could be accessed from both the template and the validation initialisation. This seems like the lowest impact approach that is in line with our usage of WTForms.

app/assets/js/app/modules/analytics.js
```diff
@@ -1,16 +1,13 @@
 import domready from './domready'
 
 export let trackEvent = (event, data) => {
-  console.log(`[Analytics disabled] Event: ${event}`)
-  console.log(data)
+  // No-op
```
hamishtaplin: This is really useful for debugging analytics events. I think just adding some `eslint-ignore` comments is better than removing it altogether.

app/assets/styles/partials/components/_character-limit.scss
```diff
@@ -0,0 +1,20 @@
+textarea.js-charlimit-exceeded {
```
hamishtaplin: This styling should be part of the `input` block as`js-` classes are hooks only. This would be more consistent with the rest of the codebase and also allow styling to be applied without JS. Something like `input__limit`.

app/templates/partials/answers/textarea.html
```diff
@@ -16,9 +16,16 @@
...
+      "maxlength": answer.max_length|default(2000, true)
     } %}
 
     {% set answer_helper = answer.name ~ "-helper" %}
 
     <textarea {{attr|xmlattr}}>{{ form.get_data(answer['id']) }}</textarea>
+
+    {% set remainingAttr = {
+      "id": answer.id ~ "-remaining",
+      "class": "js-charlimit-remaining u-mt-xs mercury"
```
hamishtaplin: Perhaps move margin into an element of `input` (eg. `input__limit`)

hamishtaplin (changes requested):
Couple of code style comments.

However, I believe the AC states that users can exceed the limit, it functions as a warning?

yoganandkunche:
@adoublebarrel *Issue with Safari*: the moment respondent hits the max char limit while pasting text in the comments field, Safari does not allow the respondent to remove and add text via the keyboard.

adoublebarrel:
@yoganandkunche This looks like it could be a bug in Safaris textarea + maxlength implementation. I've created a plain html textarea and get the same behaviour. Can you try it independently and see if you get the same results?

```
<html>
    <body>
        <textarea maxlength=2000></textarea>
    </body>
</html>
```

yoganandkunche:
@adoublebarrel I tried it myself on various browsers and I agree with you that it is a bug with Safari browser.

app/forms/custom_fields.py
```diff
@@ -0,0 +1,7 @@
+from wtforms import TextAreaField
+
+
+class MaxTextAreaField(TextAreaField):
+    def __init__(self, label='', validators=None, maxlength=10000, **kwargs):
+        super(MaxTextAreaField, self).__init__(label, validators, **kwargs)
+        self.maxlength = maxlength
```
ajmaddaford, line 7: We should also move `CustomIntegerField` here

app/validation/error_messages.py
```diff
@@ -10,4 +10,5 @@
     'INTEGER_TOO_LARGE': _('This number is too big.'),
     'INVALID_DATE_RANGE_TO_BEFORE_FROM': _("The 'period to' date cannot be before the 'period from' date."),
     'INVALID_DATE_RANGE_TO_FROM_SAME': _("The 'period to' date must be different to the 'period from' date."),
+    'MAX_LENGTH_EXCEEDED': _("Your answer is too long, it has to be less than %%(max)d characters"),
```
ajmaddaford: I think we can avoid this double escaping `%%` by dropping the `_` gettext call for all of these strings as WTForms passes all error messages through gettext.

ajmaddaford (changes requested):
I think we also need an integration test that tests the WTForms validation error contains the correct length value.

adoublebarrel:
Done, done and done.

hamishtaplin (approved):

ajmaddaford (approved):

PR #1158, opened 2017-06-16
Eq 1114 make question skip conditions consistent with groups

dwyeradam:
Allow for Questions to be skipped with an array of possible (or) conditions in line with those used for Groups and Blocks.
Allow for the summary page to check if a question has been skipped by metadata values.\

### How to review

The test_skip_condition_question schema has been created with the new style skip array. On the first page you can select Yes to either/both of the skip questions. If yes is selected on either then the second question will not be present on the next page. You can also set your user_id to "Skip" and that will skip the second question based on metadata.


app/templating/summary/question.py
```diff
@@ -17,17 +17,15 @@ def __init__(self, question_schema, answers):
...
-                    answer = answer.unescape()
-
-                if not evaluate_rule(when_rule, answer):
-                    return False
-            return True
-
+            for when_rules in self.skip_condition:
+                for when_rule in when_rules['when']:
+                    if 'id' in when_rule:
+                        answer = answer_store.map().get(when_rule['id'])
```
ajmaddaford: We shouldn't use the map method to get an answer, we should be using the `filter` method on `answer_store` to return all matching answers.

app/templating/summary/question.py
```diff
@@ -17,17 +17,15 @@ def __init__(self, question_schema, answers):
...
+                        if isinstance(answer, Markup):
+                            answer = answer.unescape()
```
ajmaddaford: `answer` isn't used

app/templating/summary_context.py
```diff
@@ -27,6 +27,6 @@ def build_summary_rendering_context(schema_json, answer_store, metadata):
                 if isinstance(value, str):
                     answers[answer_id] = escape(value)
 
-            groups.extend([Group(group, answers, path, metadata, url_for)])
+            groups.extend([Group(group, answer_store, path, metadata, url_for)])
```
ajmaddaford: Changing this to use the answer_store rather than answers means answers aren't being escaped.

ajmaddaford (changes requested):

ajmaddaford:
I think it would be good to get together with @adoublebarrel tomorrow to review the changes to ensure we don't lose any of the escaping fixes he recently added.

ajmaddaford (approved):

tests/functional/spec/skip-condition-question.spec.js
```diff
@@ -0,0 +1,126 @@
+import {getRandomString, openQuestionnaire, startQuestionnaire} from '../helpers'
+
+import DoYouWantToSkip from '../pages/surveys/skip_condition_question/do-you-want-to-skip.page.js'
+import ShouldSkip from '../pages/surveys/skip_condition_question/should-skip.page.js'
+import SummaryPage from '../pages/summary.page'
+
+describe('Skip condition question', function() {
```
jonnyshaw89: shouldn't we be making new tests under `functional_new`

data/schema/schema_v1.json
```diff
@@ -364,7 +364,7 @@
                                                         },
                                                         "skip_condition": {
```
jonnyshaw89: Maybe this should be renamed to `skip_conditions`?

jonnyshaw89 (dismissed):

jonnyshaw89 (changes requested):
See previous comments

dcdarrell9 (commented):
Not sure if the tests should be under functional_new like Jon said, but everything else looks good.

jonnyshaw89 (approved):

jonnyshaw89:
@dwyeradam squash?

PR #1159, opened 2017-06-16
Changed NI logo for UKIS theme + base template change

adoublebarrel:
[Trello Card](https://trello.com/c/ym9XnMFm/1149-update-ni-logo-on-surveys)

I've also experimentally added a conditional around page__subheader so that it wont be output if it doesn't have any content. It has padding applied to it in the CSS so when there's nothing to display the page has extra white space, Matthew noticed this and did not like.

#### For the logo

Go to the dev page and select 1_0001.json from the list of schemas. The first page displays the logo, let me know if there should be changes to logo alignment, margins etc.

#### For the base template change

You can remove the conditional from the base template to see how it currently looks on master. If you click past the first page you will see that the previous link is still displayed, as well as the mobile nav for smaller screens.


yoganandkunche:
The new NIR logo looks good to me, However GB-ENG logo seems to be a bit off the place when compared to how it is in PROD. (PFA)

Currently in PROD: 
screen shot 2017-06-16 at 15 44 58

On current Branch:
screen shot 2017-06-16 at 15 45 09

adoublebarrel:
@yoganandkunche which browser is the screen grab from?

yoganandkunche:
@adoublebarrel 'chrome'

adoublebarrel:
@yoganandkunche Thanks for feedback, I've made a tweak which should make the placement consistent

bennpearson (changes requested):
Space at top of page needs to be consistent. Bit tricky to do, so probably best leave .page__suheader on the page

bennpearson (approved):

PR #1162, opened 2017-06-22
Allow multiple encryption and signing keys

jonnyshaw89:
Survey Runner is able to accept multiple keys from upstream systems. This allows for multiple systems to integrate and for multiple keys to be supported for each system. Thus making the rotation of keys a zero downtime process.

### How to review 
Add Multiple keys to the secrets and test authenticating with different kid values matching the keys you have added.


app/authentication/jwt_decoder.py
```diff
@@ -22,27 +22,24 @@ class JWTDecryptor(JWERSAOAEPDecryptor):
...
             if signed_token:
                 logger.debug("decoding signed jwt", jwt_token=strings.to_str(signed_token))
                 self._check_token(signed_token)
-                token = jwt.decode(signed_token, self.rrm_public_key, algorithms=['RS256'],
+
+                kid = self._extract_kid(signed_token)
+                public_key = self._secret_store.get_public_key_by_kid(kid)
+
+                # oddly the python cryptography library needs these as bytes string
+                rrm_public_key_as_bytes = strings.to_bytes(public_key)
```
collisdigital: avoid calling this rrm anymore?

app/authentication/jwt_decoder.py
```diff
@@ -169,6 +172,8 @@ def __check_jwe_protected_header(self, header):
             raise InvalidTokenException("Missing Encoding")
         if header_data.get("enc") != "A256GCM":
             raise InvalidTokenException("Invalid Encoding")
+        if not header_data.get("kid"):
+            raise InvalidTokenException("Missing kid")
```
collisdigital: Need to check with RRM, we may have to support kid as optional initially

app/cryptography/jwe_decryption.py
```diff
@@ -52,19 +48,31 @@ def decrypt(self, token):
...
 
         signed_token = self._decrypt_cipher_text(cipher_text, iv, decrypted_key, tag, jwe_protected_header)
         return signed_token
 
-    def _decrypt_key(self, encrypted_key):
+    def _decrypt_key(self, encrypted_key, kid):
+
+        private_key = serialization.load_pem_private_key(
+            self._secret_store.get_private_key_by_kid(kid).encode(),
+            password="digitaleq".encode(),
```
collisdigital: hmmm, hard coded...

app/secrets.py
```diff
@@ -15,8 +13,52 @@
...
+]
+
 
 def validate_required_secrets(secrets):
     for required_secret in EXPECTED_SECRETS:
-        if required_secret not in secrets:
+        if required_secret not in secrets['secrets']:
             raise Exception("Missing Secret [{}]".format(required_secret))
+
+    # for required_private_key in EXPECTED_PRIVATE_KEYS:
```
collisdigital: Remove

app/secrets.py
```diff
@@ -15,8 +13,52 @@
...
+    def __init__(self, secrets):
+        if 'secrets' in secrets:
+            self.secrets = secrets['secrets']
+        if 'private_keys' in secrets:
+            self.private_keys = secrets['private_keys']
+        if 'public_keys' in secrets:
+            self.public_keys = secrets['public_keys']
+
+    def get_secret_by_name(self, secret_name):
+        if secret_name in self.secrets:
```
collisdigital: should these raise exceptions if not found?

app/secrets.py
```diff
@@ -15,8 +13,52 @@
...
+            return self.secrets[secret_name]
+
+    def get_private_key_by_kid(self, kid):
```
collisdigital: we discussed the concept of key purposes, so you restrict even further, e.g. `get_private_key_by_kid(self, kid, purpose)`

collisdigital (changes requested):

app/secrets.py
jonnyshaw89: Yeh, that was next on the list to think about.

app/cryptography/jwe_decryption.py
jonnyshaw89: Open question as to wether or not we still want passwords on the private keys?

app/secrets.py
jonnyshaw89: The validate_required_secrets validates that on startup so the application won't deploy if its missing a secret. If its not required then None seems like a good value to return?

app/cryptography/jwe_decryption.py
collisdigital: Seeing as the password is public it looks to offer little value, go-launch-a-survey also doesn't support password protected private keys.

app/secrets.py
collisdigital: in that case you can just use `self.secrets.get(secret_name)`

app/secrets.py
collisdigital: 👍

app/secrets.py
```diff
@@ -18,5 +16,31 @@
...
+        if kid in self.private_keys:
+            return self.private_keys[kid]
+
+        raise InvalidTokenException("Invalid Private Key Identifier [{}]".format(kid))
+
+    def get_public_key_by_kid(self, kid):
+        if kid in self.public_keys:
```
collisdigital: same about using `.get()`

collisdigital: do we need to do any validation that `kid` is valid (e.g. is 160 bits etc.)?

scripts/generate_secrets.py
```diff
@@ -11,23 +16,73 @@ def get_file_contents(folder, filename, trim=False):
...
+def to_bytes(bytes_or_str):
+    if isinstance(bytes_or_str, str):
+        value = bytes_or_str.encode()
+    else:
+        value = bytes_or_str
+    return value
+
+
+def generate_kid_from_key(pem_data):
+    key_as_bytes = to_bytes(pem_data)
```
collisdigital: Not sure this is correct, I'd expect to load the PEM data using python cryptography and convert the public key to bytes, otherwise you're hashing the PEM text not the key bytes itself (see https://www.ietf.org/rfc/rfc3280.txt section 4.2.1.2)

scripts/generate_secrets.py
```diff
@@ -11,23 +16,73 @@ def get_file_contents(folder, filename, trim=False):
...
+    dict[kid] = key
+
+
+def add_private_key_to_dict(dict, key):
+
+    private_key = load_pem_private_key(key.encode(), password="digitaleq".encode(), backend=backend)
+
+    pub_key = private_key.public_key()
+
+    pub_bytes = pub_key.public_bytes(Encoding.PEM, PublicFormat.SubjectPublicKeyInfo)
```
collisdigital: This looks better, do similar on line 27?

scripts/generate_secrets.py
```diff
@@ -11,23 +16,73 @@ def get_file_contents(folder, filename, trim=False):
...
+
 if len(sys.argv) < 2:
     raise Exception("Please specify the directory containing the secrets")
 
 keys_folder = sys.argv[1]
 
-keys = {}
+public_keys = {}
+
+# add_key_to_dict(public_keys, get_file_contents(keys_folder, 'sdc-submission-encryption-sdx-public-key.pem'))
```
collisdigital: Remove?

scripts/generate_secrets.py
```diff
@@ -11,23 +16,73 @@ def get_file_contents(folder, filename, trim=False):
...
+private_keys['EQ_USER_AUTHENTICATION_SR_PRIVATE_KEY'] = get_file_contents(keys_folder, 'sdc-user-authentication-encryption-sr-private-key.pem')
+
+private_keys['EQ_SUBMISSION_SR_PRIVATE_SIGNING_KEY'] = get_file_contents(keys_folder, 'sdc-submission-signing-sr-private-key.pem')
 
 if os.path.exists(os.path.join(keys_folder, 'sdc-user-authentication-signing-rrm-private-key.pem')):
-    keys['EQ_USER_AUTHENTICATION_RRM_PRIVATE_KEY'] = get_file_contents(keys_folder, 'sdc-user-authentication-signing-rrm-private-key.pem')
+    private_keys['EQ_USER_AUTHENTICATION_RRM_PRIVATE_KEY'] = get_file_contents(keys_folder, 'sdc-user-authentication-signing-rrm-private-key.pem')
+
+keys = {
+    "public_keys": public_keys,
```
collisdigital: May have to expand this to include key purpose

collisdigital (changes requested):

app/secrets.py
jonnyshaw89: I agree. But, Would need to work out how we identify the keys where we only have one version. e.g. how would we know what public key we would use for submission?

scripts/generate_secrets.py
jonnyshaw89: Both methods create identical hex values

app/secrets.py
collisdigital: if we had the concept of key types/purposes you could request the key by purpose (and use the first?)

jonnyshaw89: Or require just 1 to be present when starting the app?

app/authentication/authenticator.py
```diff
@@ -76,19 +78,13 @@ def store_session(metadata):
...
-        current_app.config['EQ_USER_AUTHENTICATION_SR_PRIVATE_KEY'],
-        current_app.config['EQ_USER_AUTHENTICATION_SR_PRIVATE_KEY_PASSWORD'],
-        current_app.config['EQ_USER_AUTHENTICATION_RRM_PUBLIC_KEY'],
-    )
+    jwt_token = decrypt_jwe(encrypted_token, current_app.eq['secret_store'], purpose=PURPOSE_AUTHENTICATION,
+                            default_kid='EQ_USER_AUTHENTICATION_SR_PRIVATE_KEY')
 
-    decrypted_token = decoder.decrypt_jwt_token(
-        encrypted_token,
-        current_app.config['EQ_JWT_LEEWAY_IN_SECONDS'],
```
collisdigital, line 90: Is the leeway still handled somewhere?

app/authentication/jwt_decoder.py
```diff
@@ -1,179 +0,0 @@
...
-
-        self._check_headers(header_data)
-        self._check_header_values(token_as_str)
-        self._check_payload(payload_data)
-
-    @staticmethod
-    def _check_header_values(token):
-        header = jwt.get_unverified_header(token)
-
-        if not header:
```
collisdigital, line 74: Being lazy, are all these checks still present somewhere else?

app/authentication/jwt_decoder.py
```diff
@@ -1,179 +0,0 @@
...
-                store[key] = value
-        return store
-
-    def _check_payload(self, payload_data):
-        try:
-            payload = self._base64_decode(payload_data)
-            if not payload:
-                raise InvalidTokenException("Missing Payload")
-            payload_decoded = payload.decode()
-            if payload_decoded == "{}":
```
collisdigital, line 120: And these, are they checked elsewhere now?

app/authentication/jwt_decoder.py
```diff
@@ -1,179 +0,0 @@
...
-                raise NoTokenException("JWT Missing")
-        except (jwt.DecodeError, InvalidTag, InternalError, ValueError, AssertionError) as e:
-            raise InvalidTokenException(repr(e))
-
-    def __check_jwe_protected_header(self, header):
-        header = self._base64_decode(header).decode()
-        header_data = json.loads(header)
-        if not header_data.get("alg"):
-            raise InvalidTokenException("Missing Algorithm")
-        if header_data.get("alg") != "RSA-OAEP":
```
collisdigital, line 166: And here

app/cryptography/token_helper.py
```diff
@@ -0,0 +1,97 @@
...
+
+        jwe_kid = extract_kid_from_header(encrypted_token, default_kid)
+
+        sr_private_jwk = secret_store.get_private_key_by_kid(purpose, jwe_kid)
+
+        jwe_token.decrypt(sr_private_jwk)
+
+        return jwe_token.payload.decode()
+    except InvalidJWEData as e:
+        raise InvalidTokenException(repr(e))
```
collisdigital: you can do `raise InvalidTokenException from e` which chains the exceptions together

app/cryptography/token_helper.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+def encrypt_jwe(payload, kid, secret_store, purpose, alg="RSA-OAEP", enc="A256GCM", protected_header=None):
+
+    try:
+        downstream_public_jwk = secret_store.get_public_key_by_kid(purpose, kid)
+
+        _protected_header = {"alg": alg, "enc": enc, "kid": kid} if protected_header is None else protected_header
```
collisdigital: seems like this function is overloaded to do two things, can't this creation of a protected header be moved out so this function just takes the header?

app/secrets.py
```diff
@@ -15,8 +13,44 @@
...
     for required_secret in EXPECTED_SECRETS:
-        if required_secret not in secrets:
+        if required_secret not in secrets['secrets']:
             raise Exception("Missing Secret [{}]".format(required_secret))
+
+
+class SecretStore:
+    def __init__(self, secrets):
+        if 'secrets' in secrets:
+            self.secrets = secrets['secrets']
```
collisdigital: maybe the `if` statements should be removed and replace with `self.secrets = secrets['secrets'] or None` so at least the attribute is always set at `__init__` even if to `None` ?

app/secrets.py
```diff
@@ -15,8 +13,44 @@
...
+            self.private_keys = secrets['private_keys']
+        if 'public_keys' in secrets:
+            self.public_keys = secrets['public_keys']
+
+    def get_secret_by_name(self, secret_name):
+        return self.secrets.get(secret_name)
+
+    def get_private_key_by_kid(self, purpose, kid):
+        if purpose in self.private_keys and kid in self.private_keys[purpose]:
+            return jwk.JWK.from_pem(self.private_keys[purpose][kid].encode('utf-8'), 'digitaleq'.encode('utf-8'))
```
collisdigital: Do we have a tech debt card about the password?

app/submitter/encrypter.py
```diff
@@ -1,83 +1,15 @@
...
-        cipher = Cipher(algorithms.AES(self.cek), modes.GCM(self.iv), backend=backend)
-        encryptor = cipher.encryptor()
-
-        encryptor.authenticate_additional_data(jwe_protected_header)
-
-        ciphertext = encryptor.update(payload) + encryptor.finalize()
+def encrypt(json, secret_store):
 
-        tag = encryptor.tag
+    jwt_kid = secret_store.get_private_kids_by_purpose(PURPOSE_SUBMISSION)
```
collisdigital: Maybe an assert or similar to ensure we only get 1 key back?

app/views/dev_mode.py
```diff
@@ -6,13 +6,17 @@
...
-from app.cryptography.jwt_encoder import Encoder
+from app.cryptography.token_helper import encrypt_jwe, encode_jwt
+from app.secrets import PURPOSE_AUTHENTICATION
 from app.utilities.schema import available_json_schemas
 
 # pylint: disable=too-many-locals
 logger = get_logger()
 dev_mode_blueprint = Blueprint('dev_mode', __name__, template_folder='templates')
 
+EQ_USER_AUTHENTICATION_RRM_PRIVATE_KEY_KID = "709eb42cfee5570058ce0711f730bfbb7d4c8ade"
```
collisdigital: Do we have examples in the ons-schema-docs of the extract format of the kid like this, i.e. a lowercase alpha-numeric byte (0-9a-f) string exactly (?) characters long?

collisdigital: Makes me think we should put a regex validator in for checking KIDs conform.

scripts/generate_secrets.py
collisdigital: ok

scripts/generate_secrets.py
```diff
@@ -11,23 +19,75 @@ def get_file_contents(folder, filename, trim=False):
...
+# Support default value until all upstream systems are passing kid
+private_keys[PURPOSE_AUTHENTICATION]['EQ_USER_AUTHENTICATION_SR_PRIVATE_KEY'] = \
+    get_file_contents(keys_folder, 'sdc-user-authentication-encryption-sr-private-key.pem')
+
+keys = {
+    "public_keys": public_keys,
+    "private_keys": private_keys,
+}
 
 with open('secrets.yml', 'w') as f:
```
collisdigital: This file is getting more complex, should we document the structure of it somewhere, perhaps in the new design/architecture decisions log or README?

tests/app/authentication/__init__.py
```diff
@@ -40,7 +40,7 @@
 -----END RSA PRIVATE KEY-----"""
 
 
-TEST_DO_NOT_USE_RRM_PUBLIC_PEM = """-----BEGIN PUBLIC KEY-----
+TEST_DO_NOT_USE_UPSTREAM_PUBLIC_PEM = """-----BEGIN PUBLIC KEY-----
```
collisdigital: Maybe the name should include the purpose so it's obvious what it's for?

tests/app/authentication/__init__.py
```diff
@@ -50,7 +50,7 @@
 kwIDAQAB
 -----END PUBLIC KEY-----"""
 
-TEST_DO_NOT_USE_RRM_PRIVATE_PEM = """-----BEGIN RSA PRIVATE KEY-----
+TEST_DO_NOT_USE_UPSTREAM_PRIVATE_PEM = """-----BEGIN RSA PRIVATE KEY-----
```
collisdigital: Maybe the name should include the purpose so it's obvious what it's for?

tests/app/authentication/__init__.py
```diff
@@ -83,29 +83,32 @@
...
-            "BslOcCLtAlGdRsptS6-L4YHxaGEAvwXD1heQADoAabjGCA-BZGPdc6h8A6y1QR7m2P33KEz83h141X4Ftyd9kTGxgjQ8-UrIGvu9b1wbm" \
-            "cjTWK3hiiyC3o0guBbzBq0SflX-RG9_zyK4sAIksiXj9mNRah1IUjUT5zr9cs1c5i4WwJdpJ28IzKvczgQ63B_t1xEpGsb6A35EhxzRkG" \
-            "7BQ.I9WExvUvZVXdwjxz8EPtCg"
 
 TEST_DO_NOT_USE_SR_PRIVATE_PEM_PASSWORD = "digitaleq"
 
 TEST_DO_NOT_USE_PASSWORD = "digitaleq"
 
-TEST_DO_NOT_USE_RRM_PRIVATE_KEY = """-----BEGIN RSA PRIVATE KEY-----
+TEST_DO_NOT_USE_UPSTREAM_PRIVATE_KEY = """-----BEGIN RSA PRIVATE KEY-----
```
collisdigital: Maybe the name should include the purpose so it's obvious what it's for?

collisdigital (approved):
Looks good!

app/authentication/jwt_decoder.py
jonnyshaw89: Yes the library check all of this for us. the tests are still present that used to test this.

app/authentication/jwt_decoder.py
jonnyshaw89: accepted alg is passed into the library when decoding

app/submitter/encrypter.py
jonnyshaw89: I might add that logic to validate_secrets so the app won't start without one.

app/authentication/jwt_decoder.py
collisdigital: Great

app/authentication/jwt_decoder.py
collisdigital: :+1:

scripts/generate_secrets.py
jonnyshaw89: Not the readme, already too much info in that.

app/submitter/encrypter.py
collisdigital: I was also thinking that there isn't more than 1 (there shouldn't be)

jonnyshaw89: agreed.

app/secrets.py
```diff
@@ -15,8 +13,44 @@
...
+
+        raise InvalidTokenException("Invalid Public Key Identifier [{}] for Purpose [{}]".format(kid, purpose))
+
+    def get_private_kids_by_purpose(self, purpose):
+        if purpose in self.private_keys:
+            return list(self.private_keys[purpose].keys())
+
+    def get_public_kids_by_purpose(self, purpose):
+        if purpose in self.public_keys:
+            return list(self.public_keys[purpose].keys())
```
ajmaddaford: I think using a nested structure with the kid and purpose as keys rather than values makes this code hard to understand. Flattening the keys down to something like below would make accessing keys clearer and more flexible.
```
"keys":
  - "kid": "e19091072f920cbf3ca9f436ceba309e7d814a62"
    "type": "public"
    "purpose": "authentication"
    "value": |
      -----BEGIN ENCRYPTED PRIVATE KEY-----
      ...
```

ajmaddaford: I also think it might be worth considering the SecretStore returning instances of a Key class, that has properties for kid, purpose, value etc and a method `as_jwk` to return a JWK version of a key.

app/cryptography/token_helper.py
```diff
@@ -0,0 +1,97 @@
...
+
+    try:
+        protected_header = base64url_decode(header).decode()
+
+        protected_header_data = json.loads(protected_header)
+
+        if 'kid' in protected_header:
+            return protected_header_data['kid']
+        elif default_kid:
+            return default_kid
```
ajmaddaford: I don't think this method should know anything about a default_kid, the defaulting should be done where this method is called in `token_helper.py`

app/cryptography/token_helper.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+def decrypt_jwe(encrypted_token, secret_store, purpose, default_kid=None):
+    try:
+        jwe_token = jwe.JWE(algs=['RSA-OAEP', 'A256GCM'])
+        jwe_token.deserialize(encrypted_token)
+
+        jwe_kid = extract_kid_from_header(encrypted_token, default_kid)
+
+        sr_private_jwk = secret_store.get_private_key_by_kid(purpose, jwe_kid)
```
ajmaddaford: I think we should drop the `sr_` from this variable name

app/cryptography/token_helper.py
```diff
@@ -0,0 +1,97 @@
...
+            InvalidJWSSignature,
+            JWTInvalidClaimFormat,
+            JWTMissingClaim,
+            JWTExpired,
+            ValueError) as e:
+        raise InvalidTokenException(repr(e))
+
+
+def encode_jwt(claims, kid, secret_store, purpose):
+    sr_private_jwk = secret_store.get_private_key_by_kid(purpose, kid)
```
ajmaddaford: I think we should drop the `sr_` from this variable name

app/cryptography/token_helper.py
```diff
@@ -0,0 +1,97 @@
...
+        return token.serialize(compact=True)
+    except InvalidJWEData as e:
+        raise InvalidTokenException(repr(e))
+
+
+def decode_jwt(jwt_token, secret_store, purpose, check_claims=None):
+    try:
+        jwt_kid = extract_kid_from_header(jwt_token)
+
+        upstream_public_jwk = secret_store.get_public_key_by_kid(purpose, jwt_kid)
```
ajmaddaford: I think we should drop the `upstream_` from this variable name

app/cryptography/token_helper.py
ajmaddaford: Is the protected_header arg used anywhere?

app/cryptography/token_helper.py
```diff
@@ -0,0 +1,97 @@
...
+
+        _check_claims = {"exp": None, "iat": None} if check_claims is None else check_claims
```
ajmaddaford: It looks like check_claims is only overridable for the tests, I think it shouldn't be and we should re-work the tests so it's not needed.

app/storage/encrypted_questionnaire_storage.py
```diff
@@ -68,11 +68,39 @@ def _generate_key(user_id, user_ik, pepper):
...
 
     def _encrypt_data(self, data):
-        encrypted = JWEDirEncrypter().encrypt(data, self._cek)
-        return {'data': encrypted}
+        protected_header = {
+            "alg": "dir",
+            "enc": "A256GCM",
+            "kid": "1,1",
+        }
+        token = jwe.JWE(plaintext=base64url_encode(data), protected=protected_header)
```
ajmaddaford: It's probably worth calling this `jwe_token` for clarity

app/authentication/authenticator.py
jonnyshaw89: It is now.

app/authentication/jwt_decoder.py
jonnyshaw89: All the tests have been ported across.

app/storage/encrypted_questionnaire_storage.py
```diff
@@ -68,11 +68,39 @@ def _generate_key(user_id, user_ik, pepper):
...
-        return JWEDirDecrypter().decrypt(encrypted_data['data'], self._cek)
+
+        encrypted_token = encrypted_data['data']
+
+        password = {
+            "kty": "oct",
+            "k": base64url_encode(self._cek),
+        }
+
+        key = jwk.JWK(**password)
```
ajmaddaford: Also probably better calling this `jwk` than `key`

app/storage/encrypted_questionnaire_storage.py
```diff
@@ -68,11 +68,39 @@ def _generate_key(user_id, user_ik, pepper):
...
+            "kid": "1,1",
+        }
+        token = jwe.JWE(plaintext=base64url_encode(data), protected=protected_header)
+
+        password = {
+            "kty": "oct",
+            "k": base64url_encode(self._cek),
+        }
+
+        key = jwk.JWK(**password)
```
ajmaddaford: Also probably better calling this `jwk` than `key`

app/cryptography/token_helper.py
jonnyshaw89: Thats a lot neater. The code i had was pulled from the old code.

app/storage/encrypted_questionnaire_storage.py
ajmaddaford: Might be worth factoring out creating the jwk into a method that takes the cek as a value. This method could be used by both the encrypt and decrypt methods.

app/secrets.py
```diff
@@ -15,8 +13,44 @@
     "EQ_SERVER_SIDE_STORAGE_DATABASE_PASSWORD",
 ]
 
+PURPOSE_AUTHENTICATION = 'authentication'
+PURPOSE_SUBMISSION = 'submission'
```
ajmaddaford: I think it would be clearer in the places these are used if they were `KEY_PURPOSE_AUTHENTICATION` and `KEY_ PURPOSE_SUBMISSION`

ajmaddaford (changes requested):

app/cryptography/token_helper.py
jonnyshaw89: However your code doesn't work

app/cryptography/token_helper.py
jonnyshaw89: Or remove it because i don't use arg any more

app/storage/encrypted_questionnaire_storage.py
jonnyshaw89: i can't as thats what the library is called

app/authentication/jwt_decoder.py
collisdigital: :+1:

app/secrets.py
collisdigital: `self.secrets = secrets.get('secrets')` even

scripts/generate_secrets.py
collisdigital: somewhere else then (top of this file??)

scripts/generate_secrets.py
```diff
@@ -2,6 +2,52 @@
...
+        scalar = representer(dumper, data)
+        scalar.style = style
+        return scalar
+    return new_representer
+
+represent_literal_unicode = change_style('|', SafeRepresenter.represent_str)
+yaml.add_representer(LiteralUnicode, represent_literal_unicode)
+
+
+if len(sys.argv) < 2:
```
adoublebarrel: Would be nice to introduce use of argparse under a if __name__ == "__main__" block. Argparse will handle your argument validation for you and it gives a distinct visual split between class/func defs and the code that is actually doing stuff.

Also means you can import the script for unit testing, if you want to.

scripts/generate_secrets.py
```diff
@@ -2,6 +2,52 @@
 import yaml
```
adoublebarrel, line 5: Worth sticking a #!/usr/bin/env python line at the top so the script can be executed from a shell without having to prefix it with "python"?

collisdigital:
Latest changes look good to me 👍

ajmaddaford (commented):
Looks good 👍  
I'll leave the blocking review until we're ready to deploy this breaking change to pre-prod.

ajmaddaford (approved):

PR #1169, opened 2017-06-29
Eq 1113 refactor navigation

dwyeradam:
Currently the Navigation has been overloaded into the groups in the schema. This was done based on the assumption there was a one-to-one mapping between a group and a navigation section. As the groups have been used to control flow and repeating, this is not the case. 

We should re-implement navigation to be defined outside of the group structure in the schema

### How to review 
- It is possible to identify which groups need to be completed for a navigation item to be marked as complete (get a tick) - multiple groups are allowed.
- What constitutes a group being complete is defined by the blocks within it and the route taken (all routed blocks must be completed).
- The group to jump to when using a navigation link can be specified. Only one group can be given.
- The title of the navigation link (section) can be customised.

### DIFF
--- a/app/questionnaire/navigation.py
+++ b/app/questionnaire/navigation.py
@@ -31,110 +31,144 @@ class Navigation(object):
         :param current_group_instance:
         :return:
         """
-        if self.survey_json.get('navigation', False) is False:
+
+        navigation_block = self.survey_json.get('navigation')
+        if navigation_block is None or navigation_block.get("visible", True) is False:
             return None
 
         navigation = []
 
-        last_group_id = SchemaHelper.get_last_group_id(self.survey_json)
+        for section in navigation_block['sections']:
+            non_skipped_groups = self._get_non_skipped_groups(section)
+            if len(non_skipped_groups) == 0:
+                continue
 
-        for group in filter(lambda x: 'hide_in_navigation' not in x, self.survey_json['groups']):
-            logger.debug("building frontend navigation", group_id=group['id'])
+            first_group_id = non_skipped_groups[0]
+            first_block_id = SchemaHelper.get_first_block_id_for_group(self.survey_json, first_group_id)
 
-            first_location = Location(group['id'], 0, group['blocks'][0]['id'])
-            last_block_in_group = SchemaHelper.get_last_block_in_group(group)
+            completion_group_block_id = self._get_completion_group_block_id(non_skipped_groups)
+            if completion_group_block_id is None:
+                continue
 
-            is_summary_or_confirm_group = last_block_in_group and \
-                SchemaHelper.is_summary_or_confirmation(last_block_in_group)
+            completion_group_id = completion_group_block_id['group_id']
+            completion_block_id = completion_group_block_id['block_id']
 
-            last_block_location = Location(group['id'], 0, last_block_in_group['id'])
-            can_get_to_summary = is_summary_or_confirm_group and self.routing_path and \
-                last_block_location in self.routing_path
+            if SchemaHelper.is_summary_or_confirmation(SchemaHelper.get_block(self.survey_json, first_block_id)) \
+                    and self._can_reach_summary_confirmation(completion_group_id, completion_block_id) is False:
+                continue
 
-            skip_group = self._should_skip_group(current_group_instance, group)
-            if not skip_group:
-                repeating_rule = SchemaHelper.get_repeat_rule(group)
-                if repeating_rule:
-                    logger.debug("building repeating navigation", group_id=group['id'])
-                    navigation.extend(self._build_repeating_navigation(repeating_rule, group, current_group_id,
-                                                                       current_group_instance))
-                elif last_group_id == group['id'] and can_get_to_summary:
-                    logger.debug("building navigation", group_id=group['id'])
-                    if len(self.completed_blocks) > 0 and set(self.routing_path[:-1]).issubset(self.completed_blocks):
-                        navigation.append(self._build_single_navigation(group, current_group_id, first_location))
-                elif last_group_id != group['id'] and not is_summary_or_confirm_group:
-                    logger.debug("building navigation", group_id=group['id'])
-                    navigation.append(self._build_single_navigation(group, current_group_id, first_location))
+            repeating_rule = SchemaHelper.get_repeat_rule(
+                SchemaHelper.get_group(self.survey_json, first_group_id))
+            if repeating_rule:
+                navigation.extend(self._build_repeating_navigation(repeating_rule, section, current_group_id,
+                                                                   current_group_instance, completion_block_id))
+            else:
+                navigation.append(self._build_single_navigation(section, current_group_id,
+                                                                Location(first_group_id, 0, first_block_id),
+                                                                completion_block_id))
 
         return navigation
 
-    def _should_skip_group(self, current_group_instance, group):
+    def _get_non_skipped_groups(self, section):
+        non_skipped_groups = []
+        for group_id in section['group_order']:
+            if not self._should_skip_group(group_id):
+                non_skipped_groups.append(group_id)
 
+        return non_skipped_groups
+
+    def _should_skip_group(self, group_id):
         skip_group = False
+        group = SchemaHelper.get_group(self.survey_json, group_id)
         skip_conditions = SchemaHelper.get_skip_condition(group)
 
         if skip_conditions:
-            skip_group = evaluate_skip_conditions(skip_conditions, self.metadata, self.answer_store,
-                                                  current_group_instance)
+            skip_group = evaluate_skip_conditions(skip_conditions, self.metadata, self.answer_store)
+
         return skip_group
 
-    def _build_repeating_navigation(self, repeating_rule, group, current_group_id, current_group_instance):
+    def _get_completion_group_block_id(self, group_ids):
+        """
+         Navigate backwards through each non skipped group in a section
+          checking each block in reverse until a non skipped
+          block is obtained.
+
+          Group instance is not used to evaluate skip conditions:
+          Assumption for repeating groups is that the
+          final block is constant across group instances
+
+        :param group_ids: List of groups to check
+        :return:
+        """
+        for group_id in reversed(group_ids):
+            group = SchemaHelper.get_group(self.survey_json, group_id)
+            for block in reversed(group['blocks']):
+                skip_conditions = block.get('skip_conditions')
+                if evaluate_skip_conditions(skip_conditions, self.metadata, self.answer_store) is False:
+                    return {'group_id': group_id, 'block_id': block['id']}
+
+        return None
+
+    def _can_reach_summary_confirmation(self, group_id, block_id):
+        if self.routing_path and Location(group_id, 0, block_id) in self.routing_path and \
+                len(self.completed_blocks) > 0 and set(self.routing_path[:-1]).issubset(self.completed_blocks):
+            return True
+
+        return False
+
+    def _build_single_navigation(self, section, current_group_id, first_location, completion_block):
+        is_highlighted = current_group_id in section['group_order']
+        is_completed = self._is_completed(completion_block)
+
+        return self._generate_item(section['title'], is_completed, first_location, is_highlighted)
+
+    def _build_repeating_navigation(self, repeating_rule, section, current_group_id,
+                                    current_group_instance, completed_id):
+        group = SchemaHelper.get_group(self.survey_json, section['group_order'][0])
         first_location = Location(group['id'], 0, group['blocks'][0]['id'])
         no_of_repeats = evaluate_repeat(repeating_rule, self.answer_store)
 
         repeating_nav = []
 
         if repeating_rule['type'] == 'answer_count':
-            completed_id = group['completed_id'] if 'completed_id' in group else group['blocks'][-1]['id']
             is_current_group = group['id'] == current_group_id
-            link_names = self._generate_link_names(repeating_rule)
+            link_names = self._generate_link_names(section['title_from_answers'])
 
             repeating_nav = self._generate_repeated_items(link_names, first_location, completed_id, no_of_repeats,
                                                           is_current_group, current_group_instance)
 
         elif no_of_repeats > 0:
-            item = self._build_single_navigation(group, current_group_id, first_location)
+            item = self._build_single_repeating_navigation(section, current_group_id, first_location,
+                                                           completed_id, no_of_repeats)
             repeating_nav.append(item)
 
         return repeating_nav
 
-    def _build_single_navigation(self, group, current_group_id, first_location):
-        completed_id = group['completed_id'] if 'completed_id' in group else group['blocks'][-1]['id']
-        default_highlight = self._should_highlight(group, current_group_id)
-        is_completed = self._is_complete_non_repeating(completed_id)
+    def _build_single_repeating_navigation(self, section, current_group_id, first_location,
+                                           completed_id, no_of_repeats):
+        is_highlighted = current_group_id in section['group_order']
+        is_completed = self._is_completed_single_repeating(no_of_repeats, completed_id)
+        return self._generate_item(section['title'], is_completed, first_location, is_highlighted)
 
-        return self._generate_item(group['title'], is_completed, first_location, default_highlight)
+    def _is_completed_single_repeating(self, no_of_repeats, completed_id):
+        for i in range(no_of_repeats):
+            if self._is_completed(completed_id, i) is False:
+                return False
 
-    @staticmethod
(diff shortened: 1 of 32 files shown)


app/questionnaire/navigation.py
```diff
@@ -31,86 +31,96 @@ def build_navigation(self, current_group_id, current_group_instance):
...
+                elif last_group_id != first_sect_group and not is_summary_or_confirm_group:
+                    navigation.append(self._build_single_navigation(section, current_group_id,
+                                                                    first_block_location, last_block))
 
         return navigation
 
-    def _should_skip_group(self, current_group_instance, group):
-
-        skip_group = False
+    def _get_last_sect_group(self, section, current_group_instance):
```
jonnyshaw89: Is there any need for the abbreviation in this method?

jonnyshaw89 (changes requested):

dcdarrell9 (commented):
Looks good 👍

app/questionnaire/navigation.py
dwyeradam: Nope so now fully formed as _get_last_section_group.

jonnyshaw89: Is it?

dwyeradam: Yes he says sheepishly. Seems I lost a couple of changes during the conflict resolution. Done now.

jonnyshaw89 (dismissed):

dwyeradam:
Adding functional tests to check for completed status.

app/questionnaire/navigation.py
```diff
@@ -31,110 +31,133 @@ def build_navigation(self, current_group_id, current_group_instance):
...
 
-            is_summary_or_confirm_group = last_block_in_group and \
-                SchemaHelper.is_summary_or_confirmation(last_block_in_group)
+            last_section_group_id = self._get_last_non_skipped_group_in_section(section)
+            last_block_id = self._get_last_block_id(last_section_group_id, current_group_instance)
 
-            last_block_location = Location(group['id'], 0, last_block_in_group['id'])
-            can_get_to_summary = is_summary_or_confirm_group and self.routing_path and \
-                last_block_location in self.routing_path
+            if first_block_id in ('summary', 'confirmation') \
```
ajmaddaford: This needs to use the `is_summary_or_confirmation` method in `SchemaHelper` so that we're comparing against block types rather than ids

app/questionnaire/navigation.py
```diff
@@ -31,110 +31,133 @@ def build_navigation(self, current_group_id, current_group_instance):
...
             return None
 
         navigation = []
 
-        last_group_id = SchemaHelper.get_last_group_id(self.survey_json)
-
-        for group in filter(lambda x: 'hide_in_navigation' not in x, self.survey_json['groups']):
-            logger.debug("building frontend navigation", group_id=group['id'])
+        for section in navigation_block['sections']:
+            first_section_group_id = self._get_first_non_skipped_group_in_section(section)
```
ajmaddaford: I think it would be clearer if this was named `first_group_id_in_section` or possibly dropping the section and making it `first_group_id`

app/questionnaire/navigation.py
```diff
@@ -31,110 +31,133 @@ def build_navigation(self, current_group_id, current_group_instance):
...
+            if first_section_group_id is None:
+                continue
 
-            first_location = Location(group['id'], 0, group['blocks'][0]['id'])
-            last_block_in_group = SchemaHelper.get_last_block_in_group(group)
+            first_block_id = SchemaHelper.get_first_block_id_for_group(self.survey_json, first_section_group_id)
 
-            is_summary_or_confirm_group = last_block_in_group and \
-                SchemaHelper.is_summary_or_confirmation(last_block_in_group)
+            last_section_group_id = self._get_last_non_skipped_group_in_section(section)
```
ajmaddaford: Same as comment above, this could be `last_group_id`
Also, as this variable is only used if the block is a summary or confirmation, we should move it under that conditional check, something like:
```
if first_block_id in ('summary', 'confirmation'):
    last_section_group_id = self._get_last_non_skipped_group_in_section(section)
    if self._can_reach_summary_confirmation(last_section_group_id, last_block_id) is False:
        continue
```

app/questionnaire/navigation.py
```diff
@@ -31,110 +31,133 @@ def build_navigation(self, current_group_id, current_group_instance):
...
+
+        return None
 
-        skip_group = False
+    def _get_last_non_skipped_group_in_section(self, section):
+        for group_id in reversed(section['group_order']):
+            if not self._should_skip_group(group_id):
+                return group_id
+
+        return section['group_order'][0]
```
ajmaddaford: Why does this return the first group id even if it's skipped?

app/questionnaire/navigation.py
```diff
@@ -31,110 +31,133 @@ def build_navigation(self, current_group_id, current_group_instance):
...
 
         if skip_conditions:
-            skip_group = evaluate_skip_conditions(skip_conditions, self.metadata, self.answer_store,
-                                                  current_group_instance)
-        return skip_group
+            skip_nav = evaluate_skip_conditions(skip_conditions, self.metadata, self.answer_store)
+
+        return skip_nav
+
+    def _get_last_block_id(self, last_section_group_id, current_group_instance):
```
ajmaddaford: `last_section_group_id` should be `group_id` - the method isn't concerned with where the group id came from

app/questionnaire/navigation.py
```diff
@@ -31,110 +31,133 @@ def build_navigation(self, current_group_id, current_group_instance):
...
 
-    def _build_single_navigation(self, group, current_group_id, first_location):
-        completed_id = group['completed_id'] if 'completed_id' in group else group['blocks'][-1]['id']
-        default_highlight = self._should_highlight(group, current_group_id)
-        is_completed = self._is_complete_non_repeating(completed_id)
-
-        return self._generate_item(group['title'], is_completed, first_location, default_highlight)
+    def _build_single_repeating_navigation(self, section, current_group_id, first_location,
+                                           completed_id, no_of_repeats):
+        default_highlight = current_group_id in section['group_order']
```
ajmaddaford: Not sure why this was called `default_highlight` previously, it should be renamed something like `should_highlight`

ajmaddaford (changes requested):

jonnyshaw89:
@dwyeradam can you rebase this branch so it had the latest changes from master?

app/questionnaire/navigation.py
```diff
@@ -31,110 +31,145 @@ def build_navigation(self, current_group_id, current_group_instance):
...
+        :param group_instance: Used for repeating groups
+        :return:
+        """
+        last_group_block_id = None
+        for group_id in reversed(group_list):
+            group = SchemaHelper.get_group(self.survey_json, group_id)
+            for block in reversed(group['blocks']):
+                skip_cons = block.get('skip_conditions')
+                if evaluate_skip_conditions(skip_cons, self.metadata, self.answer_store, group_instance) is False:
+                    last_group_block_id = {'group_id': group_id, 'block_id': block['id']}
```
jonnyshaw89: could you not just return here?

jonnyshaw89:
One issue i found was that if i complete a block but then go back and change an answer that routes. The section is still marked as complete when the newly exposed block has not been completed.
The result is that when reaching the end of the survey i am redirected back to the un-completed block.
Discussed with @dwyeradam and a separate card will be raised to address this.

dwyeradam:
Just to expand and clarify on  @jonnyshaw89 comment above. This is where the new routing path shares the same end point as a previously completed path. This is because the code is only looking at the completed id of the last answer which in this instance would remain completed. 
Changes to answers where the end point is changed as a result will remove the completed mark.

tests/new_functional/spec/navigation.spec.js
```diff
@@ -0,0 +1,156 @@
...
+const InsuranceTypePage = require('../pages/surveys/navigation/insurance-type.page.js');
+const RepeatingBlock1Page = require('../pages/surveys/navigation/repeating-block-1.page.js');
+const RepeatingBlock2Page = require('../pages/surveys/navigation/repeating-block-2.page.js');
+const SecurityCodePage = require('../pages/surveys/navigation/security-code.page.js');
+const SecurityCodeInterstitialPage = require('../pages/surveys/navigation/security-code-interstitial.page.js');
+const SkipInterstitialPage = require('../pages/surveys/navigation/skip-interstitial.page.js');
+const SkipPaymentPage = require('../pages/surveys/navigation/skip-payment.page.js');
+const SummaryPage = require('../pages/surveys/navigation/summary.page.js');
+
+describe('Navigation', function() {
```
jonnyshaw89: Were we going to add tests for the section 'ticks'?

dwyeradam: Yes if I can figure out how. Had some difficulty getting it to recognise them.

yoganandkunche:
@dwyeradam: As discussed earlier if you can add at least one more question to that group which has repeating questions will allow me to cover more scenarios where I can confirm that the last question after the repeating questions is considered as the trigger to set the group as completed (Green Tick).

app/questionnaire/navigation.py
```diff
@@ -31,110 +31,141 @@ def build_navigation(self, current_group_id, current_group_instance):
...
+
+        :param group_list: List of groups to check
+        :param group_instance: Used for repeating groups
+        :return:
+        """
+        for group_id in reversed(group_list):
+            group = SchemaHelper.get_group(self.survey_json, group_id)
+            for block in reversed(group['blocks']):
+                skip_cons = block.get('skip_conditions')
+                if evaluate_skip_conditions(skip_cons, self.metadata, self.answer_store, group_instance) is False:
```
ajmaddaford: `group_instance` is the current group instance the user is on when `build_navigation` is called, and doesn't change each time `_get_last_group_and_block_id` is called. I think for now we should remove this argument and let `evaluate_skip_conditions` use the default of `0`.  This will mean that for every repeat of a group the completed block will be the same, but we should tackle that problem when we look at using the routing path to determine if a section is complete.

ajmaddaford (changes requested):

app/questionnaire/navigation.py
```diff
@@ -31,110 +31,143 @@ def build_navigation(self, current_group_id, current_group_instance):
...
+        return skip_nav
+
+    def _get_last_group_and_block_id(self, group_list):
+        """
+        Navigate backwards through each non skipped group in a section
+         checking each block in reverse until a non skipped
+         block is obtained.
+         Assumption for repeating groups is that the
+         final block is
+         constant across group instances
```
ajmaddaford: Minor point, but not sure what's going on with the new lines in this documentation.

app/questionnaire/navigation.py
```diff
@@ -31,110 +31,143 @@ def build_navigation(self, current_group_id, current_group_instance):
...
+        group = SchemaHelper.get_group(self.survey_json, group_id)
         skip_conditions = SchemaHelper.get_skip_condition(group)
 
         if skip_conditions:
-            skip_group = evaluate_skip_conditions(skip_conditions, self.metadata, self.answer_store,
-                                                  current_group_instance)
-        return skip_group
+            skip_nav = evaluate_skip_conditions(skip_conditions, self.metadata, self.answer_store)
+
+        return skip_nav
```
ajmaddaford: `skip_nav` should be renamed something like `skip_group` as this method doesn't know anything about navigation

app/questionnaire/navigation.py
```diff
@@ -31,110 +31,143 @@ def build_navigation(self, current_group_id, current_group_instance):
...
+                continue
 
-        for group in filter(lambda x: 'hide_in_navigation' not in x, self.survey_json['groups']):
-            logger.debug("building frontend navigation", group_id=group['id'])
+            first_group_id = non_skipped_groups[0]
+            first_block_id = SchemaHelper.get_first_block_id_for_group(self.survey_json, first_group_id)
 
-            first_location = Location(group['id'], 0, group['blocks'][0]['id'])
-            last_block_in_group = SchemaHelper.get_last_block_in_group(group)
+            last_group_block_id = self._get_last_group_and_block_id(non_skipped_groups)
```
ajmaddaford: I think it would be clearer if we replaced references to `last` with `completion`, so we'd have `completion_group_and_block_id`, `_get_completion_group_and_block_id`, and further on `completion_group_id` and `completion_block_id`

app/questionnaire/navigation.py
```diff
@@ -31,110 +31,143 @@ def build_navigation(self, current_group_id, current_group_instance):
...
+         final block is
+         constant across group instances
+
+        :param group_list: List of groups to check
+        :return:
+        """
+        for group_id in reversed(group_list):
+            group = SchemaHelper.get_group(self.survey_json, group_id)
+            for block in reversed(group['blocks']):
+                skip_cons = block.get('skip_conditions')
```
ajmaddaford: `skip_cons` shouldn't be abbreviated, it should be `skip_conditions`

app/questionnaire/navigation.py
```diff
@@ -31,110 +31,143 @@ def build_navigation(self, current_group_id, current_group_instance):
...
+    def _get_last_group_and_block_id(self, group_list):
+        """
+        Navigate backwards through each non skipped group in a section
+         checking each block in reverse until a non skipped
+         block is obtained.
+         Assumption for repeating groups is that the
+         final block is
+         constant across group instances
+
+        :param group_list: List of groups to check
```
ajmaddaford: It might be clearer to call this `group_ids` rather than `group_list`

ajmaddaford (changes requested):

yoganandkunche:
@dwyeradam Error 500 on final submission screen needs fixing

tests/new_functional/config.js
```diff
@@ -38,6 +39,9 @@ let config = {
   before: function(capabilities, specs) {
     // Allow chaining of browser promises with chai assertions
     chaiAsPromised.transferPromiseness = browser.transferPromiseness;
+
+    browser.addCommand("checkCompletedTick", helpers.checkCompletedTick);
```
ajmaddaford: As discussed, I think we should look at how we can implement this as a helper method in the test rather than as a browser command.

tests/new_functional/helpers.js
```diff
@@ -81,6 +81,21 @@ function isViewSectionsVisible() {
   return browser.isExisting(viewSectionsLink) && browser.isVisibleWithinViewport(viewSectionsLink);
 }
 
+function checkCompletedTick(link_name) {
+  return browser
+    .element(link_name)
+    .element('..')
+    .getAttribute('class')
+    .then(function (class_string) {
+      if (class_string.includes('nav__item--completed')) {
```
ajmaddaford: We should use a `data-qa` attribute rather than rely on a css attribute

tests/new_functional/spec/navigation.spec.js
```diff
@@ -0,0 +1,204 @@
...
+        .isVisible('a=Test User').should.eventually.be.true
+        .isVisible('a=Another User').should.eventually.be.true
+        .click('a=Household Composition')
+        .click(HouseholdCompositionPage.removePerson())
+        // Wait until page refresh
+        .waitUntil(function () {
+          return browser.elements('.question__answers > div').then(function (e) {
+            return e.value.length === 1;
+          });
+        }, 2000, 'Person not removed in time')
```
ajmaddaford: Not sure why we need to wait for the person to be removed before submitting. Is this an issue we need to look into as it could affect users completing the survey?

ajmaddaford (changes requested):

tests/new_functional/spec/navigation.spec.js
dwyeradam: Potentially yes but in reality I doubt it will really happen but we should look into it. In practice you have to be very quick with the mouse but if you are quick enough then the submit can preempt the remove.

tests/new_functional/helpers.js
```diff
@@ -81,6 +81,39 @@ function isViewSectionsVisible() {
   return browser.isExisting(viewSectionsLink) && browser.isVisibleWithinViewport(viewSectionsLink);
 }
 
+function checkCompletedTick(link_name) {
```
ajmaddaford: I think this method can be removed now it has been replaced with `isSectionComplete`?

ajmaddaford (changes requested):

ajmaddaford (approved):

adoublebarrel (approved):

yoganandkunche (approved):
LGTM 👍🏼

shahi645 (dismissed):
testing fourth wall

PR #1171, opened 2017-07-04
eq-1190 allow conditional date piping

berianjones:
this change allows a choice between submitted dates or default to metadata dates

New def format_cond_date added to jinja_filters.py and change reflected in template_renderer.py
This allows the use of default dates from metadata

### How to review 
I modified an existing json (not checked in), made the date entry fields non-mandatory, created an alias for the period-from and period-to answers (cf with census-household.json). I then added the following to a title held in the subsequent questions in the json 
```
{{[answers.period_to, exercise.start_date]|format_cond_date}}
{{[answers.period_from, exercise.end_date]|format_cond_date}}
```
where period_from and period_to are the aliases and exercise.start_date/end_date are the metadata values. Tested for blank dates, invalid dates

### DIFF
--- a/app/jinja_filters.py
+++ b/app/jinja_filters.py
@@ -43,6 +43,29 @@ def format_date(value):
     return value.strftime('%-d %B %Y')
 
 
+@blueprint.app_template_filter()
+def format_conditional_date(date1=None, date2=None):
+    """
+    This function format_conditional_date accepts two dates, a user submitted date and a metadata date
+
+    :param date1 user entered date
+    :param date2 is metadata date
+    :return: the value of the date to be piped
+    """
+    if date1:
+        date = date1
+    else:
+        date = date2
+
+    if date is None:
+        raise Exception("No valid dates passed to format_conditional_dates filter")
+
+    if isinstance(date, datetime):
+        return format_date(date)
+    else:
+        return format_date(datetime.strptime(date, "%d/%m/%Y"))
+
+
 def format_str_as_short_date(value):
     return datetime.strptime(value, "%d/%m/%Y").strftime('%d %B %Y')
 
@@ -96,5 +119,10 @@ def format_number_to_alphabetic_letter(number):
 
 
 @blueprint.app_context_processor
-def utility_processor():
+def start_end_date_check():
     return dict(format_start_end_date=format_start_end_date)
+
+
+@blueprint.app_context_processor
+def conditional_dates_check():
+    return dict(format_conditional_date=format_conditional_date)
--- a/data/en/test_conditional_dates.json
+++ b/data/en/test_conditional_dates.json
@@ -1,0 +1,97 @@
+{
+    "mime_type": "application/json/ons/eq",
+    "schema_version": "0.0.1",
+    "data_version": "0.0.1",
+    "survey_id": "023",
+    "title": "Date formats",
+    "description": "A test schema for different date formats",
+    "theme": "default",
+    "legal_basis": "StatisticsOfTradeAct",
+    "groups": [{
+        "blocks": [{
+                "type": "Questionnaire",
+                "id": "date-block",
+                "sections": [{
+                    "description": "",
+                    "id": "date-section",
+                    "questions": [{
+                        "answers": [{
+                                "alias": "date_start_from",
+                                "id": "date-start-from",
+                                "label": "Period from",
+                                "mandatory": false,
+                                "options": [],
+                                "q_code": "1111",
+                                "type": "Date",
+                                "validation": {
+                                    "messages": {
+                                        "INVALID_DATE": "The date entered is not valid.  Please correct your answer.",
+                                        "MANDATORY": "Please provide an answer to continue."
+                                    }
+                                }
+                            },
+                            {
+                                "alias": "date_end_to",
+                                "id": "date-end-to",
+                                "label": "Period to",
+                                "mandatory": false,
+                                "options": [],
+                                "q_code": "1112",
+                                "type": "Date",
+                                "validation": {
+                                    "messages": {
+                                        "INVALID_DATE": "The date entered is not valid.  Please correct your answer.",
+                                        "MANDATORY": "Please provide an answer to continue."
+                                    }
+                                }
+                            }
+                        ],
+                        "description": "",
+                        "id": "date-pipe-question",
+                        "title": "Pipe range",
+                        "type": "DateRange"
+                    }],
+                    "title": "Date Examples"
+                }],
+                "title": "Date Examples"
+            },
+            {
+                "type": "Questionnaire",
+                "id": "date-value-test",
+                "sections": [{
+                    "description": "A test to see if dates are correctly passed",
+                    "id": "total-date-test",
+                    "questions": [{
+                        "answers": [{
+                            "id": "date-test-answer",
+                            "label": "Piped dates",
+                            "mandatory": false,
+                            "options": [],
+                            "q_code": "20",
+                            "type": "Integer",
+                            "validation": {
+                                "messages": {
+                                    "INTEGER_TOO_LARGE": "The maximum value allowed is 9999999999. Please correct your answer.",
+                                    "MANDATORY": "Please provide a value, even if your value is 0.",
+                                    "NEGATIVE_INTEGER": "The value cannot be negative. Please correct your answer.",
+                                    "NOT_INTEGER": "Please only enter whole numbers into the field."
+                                }
+                            }
+                        }],
+                        "id": "total-retail-turnover-question",
+                        "title": "For the period {{ format_conditional_date (answers.date_start_from, exercise.start_date)}} to {{ format_conditional_date (answers.date_end_to, exercise.end_date)}}, what was the value of the business’s total retail turnover?",
+                        "type": "General"
+                    }],
+                    "title": "Conditional Date Validation"
+                }],
+                "title": "Conditional Date Validation"
+            },
+            {
+                "type": "Summary",
+                "id": "summary"
+            }
+        ],
+        "id": "dates",
+        "title": ""
+    }]
+}
--- a/tests/app/test_jinja_filters.py
+++ b/tests/app/test_jinja_filters.py
@@ -3,7 +3,8 @@ from unittest import TestCase
 
 from mock import Mock
 
-from app.jinja_filters import format_date, format_currency, format_multilined_string, format_percentage, format_start_end_date
+from app.jinja_filters import format_date, format_conditional_date, format_currency, format_multilined_string, \
+     format_percentage, format_start_end_date
 from app.jinja_filters import format_household_member_name
 from app.jinja_filters import format_str_as_date
 from app.jinja_filters import format_str_as_date_range
@@ -94,6 +95,53 @@ class TestJinjaFilters(TestCase):  # pylint: disable=too-many-public-methods
         format_value = format_date(date)
 
         self.assertEqual(format_value, '1 January 2017')
+
+    def test_format_conditional_date_not_date(self):
+        # Given       no test for integers this check was removed from jinja_filters
+
+        invalid_input = [('1', None),
+                         ('1/1/1', None)]
+
+        # When
+        for nonsense in invalid_input:
+            date1 = nonsense[0]
+            date2 = nonsense[1]
+            with self.assertRaises(Exception) as exception:
+                format_conditional_date(date1, date2)
+        # Then
+            self.assertIn("does not match format '%d/%m/%Y'", str(exception.exception))
+
+    def test_format_conditional_date_not_set(self):
+        # Given
+
+        # When
+        with self.assertRaises(Exception) as exception:
+            format_conditional_date(None, None)
+
+        # Then
+        self.assertIn("No valid dates passed to format_conditional_dates filter", str(exception.exception))
+
+    def test_format_conditional_date(self):
+        # Given
+
+        datelist = [('12/01/2016', '12/02/2016', '12 January 2016'),
+                    ('23/12/2017', None, '23 December 2017'),
+                    (datetime(2019, 5, 12), None, '12 May 2019'),
+                    (None, datetime(2017, 6, 22), '22 June 2017'),
+                    ('12/08/2017', datetime(2017, 9, 10), '12 August 2017'),
+                    (datetime(2018, 4, 7), "12/3/2018", '7 April 2018'),
+                    (None, datetime(2017, 10, 12), '12 October 2017'),
+                    (datetime(2019, 10, 12), datetime(2017, 9, 12), '12 October 2019')]
+
+        # When
+        for triple in datelist:
+            date1 = triple[0]
+            date2 = triple[1]
+            #dates = (date1, date2)
+            format_value = format_conditional_date(date1, date2)
+
+            # Then
+            self.assertEqual(format_value, triple[2])
 
     def test_format_start_end_date(self):
         # Given
--- a/tests/new_functional/pages/surveys/dates_conditional/summary.page.js
+++ b/tests/new_functional/pages/surveys/dates_conditional/summary.page.js
@@ -1,0 +1,19 @@
+// >>> WARNING THIS PAGE WAS AUTO-GENERATED - DO NOT EDIT!!! <<<
+const QuestionPage = require('../question.page');
+
+class SummaryPage extends QuestionPage {
+
+  constructor() {
+    super('summary');
+  }
+
+  answer() {
+    return '#date-start-from-answer';
+  }
+
+  otheranswer() {
+    return '#date-test-answer-answer';
+  }
+
+}
+module.exports = new SummaryPage();
--- a/tests/new_functional/spec/dates_conditional.spec.js
+++ b/tests/new_functional/spec/dates_conditional.spec.js
@@ -1,0 +1,57 @@
+const helpers = require('../helpers');
+
+const DatesPage = require('../pages/surveys/dates_conditional/date-block.page');
+const DatesConfirmationPage = require('../pages/surveys/dates_conditional/date-value-test.page');
+const SummaryPage = require('../pages/surveys/dates_conditional/summary.page');
+
+describe('Piped Dates', function () {
+
+  it('Given the test_conditional_dates survey is selected when dates are entered then the summary screen shows the conditional piped dates entered formatted', function() {
+
+            // Given the test_dates survey is selected
+            return helpers.openQuestionnaire('test_conditional_dates.json').then(() => {
+            return browser
+
+                    // When dates are entered
+                .setValue(DatesPage.dateStartFromday(), 11)
+                .selectByValue(DatesPage.dateStartFrommonth(), 10)
+                .setValue(DatesPage.dateStartFromyear(), 2017)
+                .setValue(DatesPage.dateEndToday(), 3)
+                .selectByValue(DatesPage.dateEndTomonth(), 12)
+                .setValue(DatesPage.dateEndToyear(), 2017)
+                .click(DatesPage.submit())
+                .setValue(DatesConfirmationPage.answer(), 1)
+                .click(DatesConfirmationPage.submit())
+
+                // Then the summary screen shows the dates entered formatted
+
+                .getText(SummaryPage.answer()).should.eventually.contain('11 October 2017 to 03 December 2017');
+
+})
+        })
+
+  it('Given the test_conditional_dates survey is selected when no dates are entered then the summary screen shows the conditional piped dates entered formatted', function() {
+
+            // Given the test_dates survey is selected
+           return helpers.openQuestionnaire('test_conditional_dates.json').then(() => {
+           return browser
+
+                    // When dates are entered
+                .setValue(DatesPage.dateStartFromday(), '')
+                .selectByValue(DatesPage.dateStartFrommonth(), '')
+                .setValue(DatesPage.dateStartFromyear(), '')
+                .setValue(DatesPage.dateEndToday(), '')
+                .selectByValue(DatesPage.dateEndTomonth(), '')
+                .setValue(DatesPage.dateEndToyear(), '')
+                .click(DatesPage.submit())
+                .setValue(DatesConfirmationPage.answer(), 2)
+                .click(DatesConfirmationPage.submit())
+
+
+            // Then the summary screen shows the dates entered formatted
+
+           .getText(SummaryPage.answer()).should.eventually.contain('No answer provided');
+
+         })
+       })
+})
--- a/app/templating/template_renderer.py
+++ b/app/templating/template_renderer.py
@@ -5,17 +5,21 @@ import json
 import re
 from jinja2 import Environment
 
-from app.jinja_filters import format_date, format_currency, format_household_member_name, format_household_summary, format_str_as_date
+from app.jinja_filters import format_date, format_currency, format_household_member_name, \
+    format_household_summary, format_str_as_date, format_conditional_date
 
 
 class TemplateRenderer:
     def __init__(self):
         self.environment = Environment()
+
         self.environment.filters['format_date'] = format_date
         self.environment.filters['format_str_as_date'] = format_str_as_date
         self.environment.filters['format_currency'] = format_currency
         self.environment.filters['format_household_name'] = format_household_member_name
         self.environment.filters['format_household_summary'] = format_household_summary
+
+        self.environment.globals['format_conditional_date'] = format_conditional_date
 
     def render(self, renderable, **context):
         """
--- a/tests/app/templating/test_template_renderer.py
+++ b/tests/app/templating/test_template_renderer.py
@@ -146,6 +146,13 @@ class TestTemplateRenderer(unittest.TestCase):
 
         self.assertEqual(safe_content, 'Is … correct?')
 
+    def test_should_replace_jinja_template_condtional_dates(self):
+        content = 'Is {{ format_condtional_date(meta.survey.start_date, meta.survey.end_date)}} correct?'
+
+        safe_content = TemplateRenderer.safe_content(content)
+
+        self.assertEqual(safe_content, 'Is … correct?')
+
     def test_should_replace_simple_html_tags(self):
         content = 'This <em>string contains</em> <p><strong>some</strong></p><i>HTML</i>tags?'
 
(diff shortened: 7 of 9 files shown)


app/jinja_filters.py
```diff
@@ -43,6 +43,18 @@ def format_date(value):
     return value.strftime('%-d %B %Y')
 
 
+""" This function format_cond_date accepts two dates, if a date is submitted that is used and is available, if not the metadata date is made available"""
+""" The format (example) in the json is {{[answers.date_from, exercise.start_date]|format_cond_date}}"""
+
+
+@blueprint.app_template_filter()
+def format_cond_date(dates):
```
jonnyshaw89: I can't see any reason to abbreviate this name?

jonnyshaw89 (changes requested):
I think you also need to add some tests for this new filter.

jonnyshaw89:
Did you accidentally commit `yarn.lock`?

app/jinja_filters.py
```diff
@@ -43,6 +43,18 @@ def format_date(value):
     return value.strftime('%-d %B %Y')
 
 
+@blueprint.app_template_filter()
+def format_conditional_date(dates):
+    """ This function format_cond_date accepts two dates, if a date is submitted that is used and is available,
```
jonnyshaw89: the function name in the comment is now incorrect

dwyeradam:
I agree it would be worth having a test for this. There is already a test_dates.json that could be expanded to include an extra page with piping.

collisdigital:
Please use commit messages to describe the change being made, it's making it hard to see what's going on 
image

app/jinja_filters.py
```diff
@@ -43,6 +43,18 @@ def format_date(value):
...
+    if not the metadata date is made available The format (example) in the json is
+    {{[answers.date_from, exercise.start_date]|format_cond_date}}"""
+
+    if dates[0]:
+        return datetime.strptime(dates[0], "%d/%m/%Y").strftime('%d %B %Y')
+    else:
+        return datetime.strptime(dates[1], "%d/%m/%Y").strftime('%d %B %Y')
```
ajmaddaford: You should be able to replace these lines with calls to `format_str_as_short_date` e.g. `format_str_as_short_date(dates[0])`

ajmaddaford (changes requested):

app/jinja_filters.py
```diff
@@ -43,6 +43,18 @@ def format_date(value):
...
+@blueprint.app_template_filter()
+def format_conditional_date(dates):
+    """ This function format_cond_date accepts two dates, if a date is submitted that is used and is available,
+    if not the metadata date is made available The format (example) in the json is
+    {{[answers.date_from, exercise.start_date]|format_cond_date}} """
+
+    if dates[0]:
+        return format_str_as_short_date(dates[0])
+    else:
+        return format_date(dates[1])
```
jonnyshaw89: why do the 2 return types use different methods to format the date?

package.json
```diff
@@ -142,6 +142,7 @@
     "eslint-plugin-import": "^2.3.0",
     "eslint-plugin-node": "^5.0.0",
     "events": "^1.1.1",
+    "fibers": "^2.0.0",
```
jonnyshaw89: What is this used for?

jonnyshaw89 (changes requested):

app/jinja_filters.py
berianjones: I think this is due to the date[1] being derived from the metadata. It errors if I use the same function for both

package.json
berianjones: Nothing, it must have been a consequence f the yarn.lock accidental commit

berianjones:
Added tests and addressed comments, reverted changes to yarn.lock and packages.json

app/jinja_filters.py
jonnyshaw89: I think these should both use the same function otherwise the returned date will be formatted differently.
If the 2 values are in different formats then that just needs to be handled. 
This function shouldn't be implemented to just handle some data coming from metadata.

jonnyshaw89:
@berianjones the `yarn.lock` and `package.json` changes are still in this PR?

app/jinja_filters.py
ajmaddaford: Dates in answers are stored as a string in a `dd/mm/yyyy` format, whereas dates in the metadata and survey contexts are stored as Python date objects. 

As this method should be able to compare any dates, it should deal with both date formats in either argument. I think for each date it would be fine to check if it's a Python date object and if not assume it's a date in a `dd/mm/yyyy` format.

This has highlighted some inconsistencies in the way that we handle dates that we should review.

tests/app/test_jinja_filters.py
```diff
@@ -95,6 +96,28 @@ def test_format_date(self):
...
+        # Given
+        date = None
+        date2 = datetime.strptime('01/01/17', '%d/%m/%y')
+        dates = ([date, date2])
+
+        # When
+        format_value = format_conditional_date(dates)
+
+        self.assertEqual(format_value, '1 January 2017')
+
```
ajmaddaford: If we adapt the method to work with dates in either format for both arguments to `format_conditional_date`, then we'll need additional tests to deal with all of the combinations.

ajmaddaford (changes requested):

collisdigital:
I'm still seeing a commits on this branch with the same commit message, please be descriptive when committing changes as it makes reviewing the history hard otherwise

berianjones:
Ready for review

app/jinja_filters.py
ajmaddaford: When documenting the arguments and return values in the docstring we should follow the standard used elsewhere e.g. https://github.com/ONSdigital/eq-survey-runner/blob/master/app/data_model/answer_store.py#L153

app/jinja_filters.py
```diff
@@ -43,6 +43,32 @@ def format_date(value):
...
+    if not the metadata date is made available The format (example) in the json is
+    {{[answers.date_from, exercise.start_date]|format_conditional_date}} """
+
+    if dates[0]:
+        datex = dates[0]
+    else:
+        datex = dates[1]
+
+    fmts = ('%Y', '%b %d, %Y', '%b %d, %Y', '%B %d, %Y', '%B %d %Y', '%m/%d/%Y', '%m/%d/%y', '%b %Y', '%B%Y',
+            '%b %d,%Y', '%d/%m/%Y', '%Y-%m-%d %H:%M:%S')
```
ajmaddaford: As discussed, we should assume this method is called with only two date 'formats' - a `dd/mm/yyyy` string date from the answer store, or a Python Date object from the metadata. Then, for each date argument we can just test if it's already a Date object and if not convert it to one (assuming a string in the format of `dd/mm/yyyy`). 
The Date object can then be output as a date string in the appropriate format using the `format_date` method.

tests/app/test_jinja_filters.py
```diff
@@ -95,6 +96,28 @@ def test_format_date(self):
...
+        # Given
+        date = None
+        date2 = '10/12/17'
+        dates = ([date, date2])
+
+        # When
+        format_value = format_conditional_date(dates)
+
+        self.assertEqual(format_value, '12 October 2017')
+
```
ajmaddaford: We should have more unit tests. Currently there are no tests for when dates passed to this method are Python Date objects, and there are no tests with just the first date argument set.

tests/app/test_jinja_filters.py
```diff
@@ -95,6 +96,28 @@ def test_format_date(self):
...
+        date = '03/11/2017'
+        date2 = '04/08/2018'
+        dates = ([date, date2])
+
+        # When
+        format_value = format_conditional_date(dates)
+
+        self.assertEqual(format_value, '11 March 2017')
+
+    def test_format_conditional_date2(self):
```
ajmaddaford: The method name should reflect what the method is testing, so something like `test_format_conditional_date_first_arg_none` would be better. 

It might also be worth considering creating a data structure to represent the test variations and loop over it e.g. 
```
test_dates = [
  ['03/11/2017', '04/08/2018', '11 March 2017'],
  [None, '10/12/17', '12 October 2017']
]
```
or possibly clearer using dicts:
```
test_dates = [
  {
    'date1': '03/11/2017', 
    'date2': '04/08/2018', 
    'result': '11 March 2017'
  },
  {
    'date1': None, 
    'date2': '10/12/17', 
    'result': '12 October 2017'
  }
]
```

ajmaddaford (changes requested):

berianjones:
Ready for review. New tests added and code changes made. All tests passed

app/jinja_filters.py
jonnyshaw89: @berianjones this comment is still incorrect

app/jinja_filters.py
```diff
@@ -43,6 +43,23 @@ def format_date(value):
...
+    """ This function format_cond_date accepts two dates, if a date is submitted that is used and is available,
+    if not the metadata date is made available The format (example) in the json is
+    {{[answers.date_from, exercise.start_date]|format_conditional_date}} """
+
+    if dates[0]:
+        date = dates[0]
+    else:
+        date = dates[1]
+
+    if isinstance(date, datetime):
```
jonnyshaw89: I wonder if this logic could go inside the `format_date` function?

package.json
jonnyshaw89: @berianjones this new dependency is still here. if we don't use it, it should be removed.

yarn.lock
```diff
@@ -1090,7 +1090,7 @@ block-stream@*:
   dependencies:
     inherits "~2.0.0"
 
-bluebird@^3.0.5, bluebird@^3.3.0:
+bluebird@^3.0.5, bluebird@^3.3.0, bluebird@^3.5.0:
```
jonnyshaw89: @berianjones given we shouldn't have any changes to the node dependencies then there is no reason that the lock file should be updated.

jonnyshaw89 (changes requested):

app/jinja_filters.py
```diff
@@ -43,6 +43,23 @@ def format_date(value):
...
+def format_conditional_date(dates):
+    """ This function format_cond_date accepts two dates, if a date is submitted that is used and is available,
+    if not the metadata date is made available The format (example) in the json is
+    {{[answers.date_from, exercise.start_date]|format_conditional_date}} """
+
+    if dates[0]:
+        date = dates[0]
+    else:
+        date = dates[1]
+
```
ajmaddaford: I think we should raise an Exception here if date is None, something like:
```
if date is None:
    raise Exception('No dates passed to format_conditional_date filter')
```

tests/app/test_jinja_filters.py
```diff
@@ -95,6 +96,39 @@ def test_format_date(self):
...
+    def test_format_conditional_date(self):
+        # Given
+
+        datelist = [('12/01/2016', '12/02/2016', '12 January 2016'), ('23/12/2017', None, '23 December 2017'),
+                    (None, datetime.strptime("12102017", "%d%m%Y"), '12 October 2017'),
+                    (datetime.strptime("12102019", "%d%m%Y"),
+                     datetime.strptime("12092017", "%d%m%Y"), '12 October 2019'),
+                    (datetime.strptime("12/10/2018", "%d/%m/%Y"),
+                     datetime.strptime("12/09/2017", "%d/%m/%Y"), '12 October 2018'),
+                    (None, datetime(1000, 1, 1), '1 January 1000')]
```
ajmaddaford: This is a little hard to read, I think it would be improved if:
- Each item was on it's own line
- The `datetime.strptime` method calls were replaced with `datetime.date` calls

Something like:
```
datelist = [
    ('12/01/2016', '12/02/2016', '12 January 2016'), 
    ('23/12/2017', None, '23 December 2017'),
    (None, datetime.date(2017, 10, 12), '12 October 2017'),
    (datetime.date(2019, 10, 12), datetime.date(2017, 9, 12), '12 October 2019'),
    (datetime.date(2018, 10, 12), datetime.date(2017, 9, 12), '12 October 2018'),
    (None, datetime.date(1000, 1, 1), '1 January 1000')
]
```
This also more clearly highlights some missing tests:
```
(None, '23/12/2017', '23 December 2017')
(datetime.date(2017, 10, 12), None, '12 October 2017'),
('23/12/2017', datetime.date(2017, 10, 12), '23 December 2017')
(datetime.date(2017, 10, 12), '23/12/2017', '12 October 2017')
```
And some duplicate tests:
```
(None, datetime.date(1000, 1, 1), '1 January 1000')
(datetime.date(2018, 10, 12), datetime.date(2017, 9, 12), '12 October 2018')
```

tests/app/test_jinja_filters.py
```diff
@@ -95,6 +96,39 @@ def test_format_date(self):
...
+
+        # When
+        for triple in datelist:
+            date = triple[0]
+            date2 = triple[1]
+            dates = ([date, date2])
+
+            format_value = format_conditional_date(dates)
```
ajmaddaford: There's no need to create the dates variable, can't we just call `format_conditional_date([date, date2])`?

tests/app/test_jinja_filters.py
```diff
@@ -95,6 +96,39 @@ def test_format_date(self):
...
 
+    def test_format_conditional_date_blank(self):
+        # Given
+
+        date = None
+        date2 = None
+        dates = ([date, date2])
+
+        with self.assertRaises(TypeError) as exception:
+            format_conditional_date(dates)
```
ajmaddaford: I think it would be just as clear to test `format_conditional_date([None, None])`

data/en/test_conditional_dates.json
```diff
@@ -0,0 +1,214 @@
+{
```
ajmaddaford, line 1: I think if we're adding a new survey to test the conditional dates it should be the minimal content required, this looks like it's a copy of the original test_dates survey with an additional set of dates and an extra page to test the conditional date filter. We should remove all date questions that don't relate to testing the conditional date filter.

data/en/test_dates.json
```diff
@@ -89,26 +89,28 @@
...
+                                "MANDATORY": "Please provide an answer to continue."
+                              }
+                            }
+                          }
+                        ],
+                        "description": "",
+                        "id": "non-mandatory-date-question",
+                        "title": "Non Mandatory",
+                        "type": "General"
+                      }
```
ajmaddaford: This just looks like formatting changes, running `yarn format` should remove these unnecessary changes.

tests/functional/spec/dates.spec.js
```diff
@@ -239,4 +240,73 @@ describe('Date checks', function() {
...
+              .setDateOfBirthYear(1999)
+              .setStartReportingPeriodDay(11)
+              .setStartReportingPeriodMonth(10)
+              .setStartReportingPeriodYear(2017)
+              .setEndReportingPeriodDay(3)
+              .setEndReportingPeriodMonth(12)
+              .setEndReportingPeriodYear(2017)
+              .submit()
+              .setDatesAnswer(1)
+              .submit()
```
ajmaddaford: If a page has been added to the survey, then a page object should be generated and this submit should be called on that.

ajmaddaford (changes requested):

app/jinja_filters.py
jonnyshaw89: We should probably also validate for the existence of `dates[1]`

berianjones:
Code and tests changed as requested in reviews

berianjones:
All revisions made and new new_functional test added - ready for review

tests/new_functional/spec/dates_conditional.spec.js
```diff
@@ -0,0 +1,57 @@
...
+                .click(DatesConfirmationPage.submit())
+
+                // Then the summary screen shows the dates entered formatted
+
+                .getText(SummaryPage.answer()).should.eventually.contain('11 October 2017 to 03 December 2017')
+
+})
+        })
+
+  it('Given the test_dates survey is selected when no dates are entered then the summary screen shows the conditional piped dates entered formatted', function() {
```
ajmaddaford: test_dates should be test_conditional_dates

tests/new_functional/pages/surveys/dates_conditional/summary.page.js
```diff
@@ -0,0 +1,19 @@
...
+    super('summary');
+  }
+
+ answer() {
+    return '#date-start-from-answer ';
+  }
+
+ otheranswer() {
+    return '#date-test-answer-answer ';
+  }
```
ajmaddaford: Spacing is a bit off here

tests/functional/spec/dates.spec.js
```diff
@@ -215,6 +214,7 @@ describe('Date checks', function() {
               .setNonMandatoryDateAnswerYear(2016)
               .submit()
 
+
```
ajmaddaford: These extra new lines should be removed

tests/functional/pages/surveys/dates/dates-answers.page.js
```diff
@@ -79,6 +79,42 @@ class DatesPage extends QuestionPage {
...
+  setEndReportingPeriodYear(year) {
+    browser.setValue('[name="date-end-to-year"]', year)
+    return this
+  }
+
+  setDatesAnswer(year) {
+    browser.setValue('[name="date-test-answer"]', year)
+    return this
+  }
+
```
ajmaddaford: I don't think there should be any changes to anything in tests/functional in this pull request? (as the new conditional tests are done in tests/new_functional)

ajmaddaford (changes requested):

app/jinja_filters.py
berianjones: Existence of date[0] checked, if no date[0], date is set to value of date[1]. Date is then checked and if None the exception is thrown. Thus the existence check on date[1] is carried out implicitly. A second check could be added, but this would add little value.

package.json
berianjones: OK

tests/new_functional/spec/dates_conditional.spec.js
berianjones: done

tests/new_functional/pages/surveys/dates_conditional/summary.page.js
berianjones: done

tests/functional/spec/dates.spec.js
berianjones: contents are now identical with master branch version

app/jinja_filters.py
jonnyshaw89: @berianjones line 58 would fail with `IndexError: list index out of range` so would not get to line 60 to be checked.

jonnyshaw89 (changes requested):

data/en/test_conditional_dates.json
berianjones: done

app/jinja_filters.py
```diff
@@ -43,6 +43,29 @@ def format_date(value):
     return value.strftime('%-d %B %Y')
 
 
+@blueprint.app_template_filter()
+def format_conditional_date(dates):
```
jonnyshaw89: i think that if we make this method take 2 parameters then that would handle the None default for parameter 2.
you would just need to declare an additional `@blueprint.app_context_processor` for this

berianjones:
Updated ninja_filters and test_jinja_filters to cater for error conditions of empty values passed, numbers rather than dates and date strings which are too short

berianjones:
ready for review

app/jinja_filters.py
berianjones: changes made to cater for this

app/jinja_filters.py
berianjones: List index out of range error case addressed. Looking at 2 param option

tests/functional/pages/surveys/dates/dates-answers.page.js
berianjones: all removed

app/jinja_filters.py
```diff
@@ -43,6 +43,40 @@ def format_date(value):
...
+    """
+    This function format_conditional_date accepts two dates, a user submitted date and a metadata date
+
+    :param dates: a list of dates first one is user entered, second is metadata date
+    :return: the value of the date to be piped
+    """
+    dates = ([date1, date2])
+
+    if not len(dates):
+        raise Exception("Empty dates passed to format_conditional_dates filter")
```
ajmaddaford: I don't think this Exception can ever get raised. Both `date1` and `date2` default to `None` if not passed, so dates will always be of length 2?

app/jinja_filters.py
```diff
@@ -43,6 +43,40 @@ def format_date(value):
...
+
+    if any(isinstance(item, int) for item in dates):
```
ajmaddaford: Why are we checking if either of the dates is an `int`?

app/jinja_filters.py
```diff
@@ -43,6 +43,40 @@ def format_date(value):
...
+        date = date1
+    else:
+        date = date2
+
+    if date is None:
+        raise Exception("No valid dates passed to format_conditional_dates filter")
+
+    d = str(date)
+    if len(d) < 8:
+        raise Exception("Date too short passed to format_conditional_dates filter")
```
ajmaddaford: This check is only relevant for strings, but placing it here will also check Date objects. Moving it to the else below would be better.

ajmaddaford (changes requested):

app/jinja_filters.py
```diff
@@ -43,6 +43,40 @@ def format_date(value):
...
 
+@blueprint.app_template_filter()
+def format_conditional_date(date1=None, date2=None):
+    """
+    This function format_conditional_date accepts two dates, a user submitted date and a metadata date
+
+    :param dates: a list of dates first one is user entered, second is metadata date
+    :return: the value of the date to be piped
+    """
+    dates = ([date1, date2])
```
jonnyshaw89: this code is pointless

app/jinja_filters.py
```diff
@@ -98,3 +132,8 @@ def format_number_to_alphabetic_letter(number):
 @blueprint.app_context_processor
 def utility_processor():
```
jonnyshaw89: Might be worth updating this name to make it more obvious while we are making changes here.

tests/app/test_jinja_filters.py
```diff
@@ -95,6 +96,71 @@ def test_format_date(self):
...
+                      (None, 1),
+                      (1, None),
+                      (1, ''),
+                      ('', 9),
+                      (1, 3)]
+
+        for double in error_list:
+            date = double[0]
+            date2 = double[1]
+        with self.assertRaises(Exception) as exception:
```
jonnyshaw89: I think this indentation is incorrect so its only currently testing the last value in the list?

tests/app/test_jinja_filters.py
```diff
@@ -95,6 +96,71 @@ def test_format_date(self):
...
+                      ('1', None),
+                      (None, '2'),
+                      ('1', ''),
+                      ('7/7/77', '2'),
+                      ('', '4/4/444')]
+
+        for double in error_list:
+            date1 = double[0]
+            date2 = double[1]
+        with self.assertRaises(Exception) as exception:
```
jonnyshaw89: same indentation issue?

jonnyshaw89 (changes requested):

app/jinja_filters.py
berianjones: Name updated, so that it reflects what its doing. I've changed the method format_condtional_date to accept two dates. Tests revised in consultation with Andrew.

tests/app/test_jinja_filters.py
berianjones: dealt with as part of test revision

tests/app/test_jinja_filters.py
berianjones: dealt with as part of test revision

berianjones:
All tests passed, ready for review

app/jinja_filters.py
berianjones: done

tests/app/test_jinja_filters.py
berianjones: changed

app/jinja_filters.py
berianjones: removed as agreed

app/jinja_filters.py
berianjones: removed as agreed

tests/app/test_jinja_filters.py
berianjones: changed

app/jinja_filters.py
berianjones: removed now as part of param change

tests/functional/spec/dates.spec.js
berianjones: added

data/en/test_dates.json
berianjones: done

tests/app/test_jinja_filters.py
berianjones: changed now as calling using 2 date parameters not list

tests/app/test_jinja_filters.py
berianjones: added

tests/app/test_jinja_filters.py
berianjones: was added, but params now different

yarn.lock
berianjones: reverted

ajmaddaford (approved):

berianjones:
coverage issue addressed

jonnyshaw89:
@berianjones you just need to revert your changes to `package.json` and `yarn.lock` then this is ready to go.

jonnyshaw89:
@berianjones i don't think its the same as master. your master version may have been out of date.
Its still showing changes in this PR. possibly reverting some of the changes @adoublebarrel has recently made.

berianjones:
@jonnyshaw89 can you review please. I've reverted package.json to @adoublebarrel 's changes

jonnyshaw89 (approved):

jonnyshaw89:
@berianjones Can you squash your commits down to 1 before we merge?

PR #1173, opened 2017-07-07
Added Integration testing of naughty strings

adoublebarrel:

[Trello Card](https://trello.com/c/y3bbPczC/1102-add-the-big-list-of-naughty-strings-to-our-user-input-tests-s)

Includes the addition of a schema with 507 text areas =:o This allows
all the naughty strings to be tested at once which is *much* faster
than doing 507 POSTs.

The strings are downloaded via curl into a temp directory as part of
the tests script and removed once the script finishes.

### How to review 
Run the tests and they should all pass.


tests/integration/test_textarea.py
```diff
@@ -27,3 +31,15 @@ def test_acceptable_submission(self):
...
         self.post(action=None)
         self.assertInUrl('thank-you')
+
+    def test_big_list_of_naughty_strings(self):
+        self.launchSurvey('test', 'big_list_naughty_strings')
+
+        answers = {}
+        for i in range(0, len(NAUGHTY_STRINGS)):
+            k = 'answer{}'.format(i)
+            answers[k] = NAUGHTY_STRINGS[i]
```
jonnyshaw89, line 41: What happens if new strings get added? it won't be able to find the answer index e.g. 508?

scripts/run_tests_unit.sh
```diff
@@ -4,9 +4,15 @@
...
 # NOTE: This script expects to be run from the project root with
 # ./scripts/run_tests_unit.sh
+current_dir=$(dirname "${BASH_SOURCE[0]}")
+parent_dir=$(dirname "${current_dir}")
+path_to_parent="$( cd "${parent_dir}" && pwd )"
+temp_dir="${path_to_parent}"/temp
 
 set -o pipefail
 
+curl 'https://raw.githubusercontent.com/minimaxir/big-list-of-naughty-strings/master/blns.json' --compressed --create-dirs -o "${temp_dir}"/blns.json
```
ajmaddaford: I don't think we should be curling the file down in the run tests script as this means you can't run the test independently of the script. It might be better to just include the file in the tests folder (this would also mean the issue Jon raised about additional strings wouldn't be a problem).

ajmaddaford (changes requested):

tests/integration/test_textarea.py
adoublebarrel: Have added the list of naughty strings as a fixture, see Andrews comment below.

scripts/run_tests_unit.sh
adoublebarrel: Ok, I've added a fixtures folder under the tests directory and added committed the json file. Seem sensible?

tests/integration/test_textarea.py
```diff
@@ -27,3 +31,15 @@ def test_acceptable_submission(self):
...
+
+        self.post(answers)
+        self.assertInUrl('summary')
+        self.assertEqual(200, self.last_response.status_code)
```
jonnyshaw89, line 45: We are only testing that the summary page is rendered? not that any of these strings haven't broken our safe strings and are displaying "dodginess" on the summary page?

jonnyshaw89 (changes requested):

tests/integration/test_textarea.py
adoublebarrel: Yes, this is a large grain test to make sure that the application doesn't fall over when handling "naughty" input. Finer grained tests, e.g. for XSS vectors, already exist. We should continue to review fine grained security test coverage as the platform evolves. For example, if in the future we consider SQL injection to be a potential vector we should cover that in specific tests.

jonnyshaw89: Would it make the test really slow if we asserted that each of those strings is displayed as entered on the summary page?

jonnyshaw89: agreed with @adoublebarrel that this isn't the right test to verify the output

jonnyshaw89 (approved):

ajmaddaford (approved):

PR #1179, opened 2017-07-13
Eq 1221 Numeric Range and Decimal Validation

dwyeradam:
As a respondent, 
I want to know if I have entered an unacceptable numeric value, 
so that I submit acceptable values

As a respondent,
I want to be able to enter figures to n decimal places,
so that I can give accurate figures

### How to review 

- Author must be able to set a maximum and/or minimum numeric value
- The system must alert the user when a number outside the range has been entered
- The error messages are as follows (to be confirmed - could just be option 3 and use default max and minimum:

-- The maximum value allowed is <number>. Please correct your answer.
-- The minimum value allowed is <number>. Please correct your answer.
-- The value must be between <min number> and <max number>. Please correct your answer.

The system default values will always apply if author does not set any limits
-- The default minimum will be 0
-- The default maximum will be 9999999999
error is displayed when user hits save and continue


dwyeradam:
Not ready for review yet.

dwyeradam:
See also https://github.com/ONSdigital/go-launch-a-survey/pull/14/files to be able to select the new json.

app/validation/validators.py
```diff
@@ -71,11 +72,35 @@ def __init__(self, minimum=None, maximum=None, messages=None):
...
     def __call__(self, form, field):
         data = field.data
+        error_message = None
         if data is not None:
             if self.minimum is not None and data < self.minimum:
-                raise validators.ValidationError(self.messages['NEGATIVE_INTEGER'])
+                error_message = self.messages.get('OUT_OF_RANGE', self.messages['NUMBER_TOO_SMALL'])
             elif self.maximum is not None and data > self.maximum:
-                raise validators.ValidationError(self.messages['INTEGER_TOO_LARGE'])
+                error_message = self.messages.get('OUT_OF_RANGE', self.messages['NUMBER_TOO_LARGE'])
```
ajmaddaford, line 80: Shouldn't 'OUT_OF_RANGE' be the second arg here?
As in `self.messages.get('NUMBER_TOO_LARGE', self.messages['OUT_OF_RANGE'])`

ajmaddaford (changes requested):
Nice one, this is looking good 👍 

I think we need to look at how we handle user provided min and max values:
- There are no inputs in test_numbers.json where the min or max value is defined explicitly in the schema (as in not derived from a previous answer)
- Leading on from this the min and max values should really be numbers in the schema rather than strings. We'd need to discuss how we'd work using answer values into this.
- You can input a maximum that is less than a minimum, which stops you from entering any value on the next screen. We should probably throw an exception in `get_number_field` when this happens

app/validation/validators.py
dwyeradam: It all depends how we want to handle the error messages for these. This way allows us to use the new style message only where needed. Since OUT_OF_RANGE is the new message not currently used I have made the assumption that if we add this to a schema then it is because we want to use it. Otherwise we will use the currently used too large/small messages. Similarly that is why I have not added OUT_OF_RANGE to error_messages.py.

app/data_model/questionnaire_store.py
```diff
@@ -1,4 +1,4 @@
-import json
+import simplejson
```
adoublebarrel, line 1: What caused the switch to simplejson?

app/forms/fields.py
```diff
@@ -163,24 +164,61 @@ def get_integer_field(answer, label, guidance, error_messages):
...
-        description=guidance,
-        validators=validate_with,
-    )
+    min_value = get_schema_defined_max_min_value(answer.get('min_value'), answer_store) or 0
+
+    max_value = get_schema_defined_max_min_value(answer.get('max_value'), answer_store) \
+        or get_max_min_system_limits('max', max_decimals)
+
+    if min_value > max_value:
+        raise Exception('min_value: ' + str(min_value) + ' > max_value: ' + str(max_value) +
```
adoublebarrel: Using a format string here will make it easier to read what the error message will be, and you wont need to explicitly cast the values to str

app/forms/fields.py
```diff
@@ -1,30 +1,31 @@
...
 from structlog import get_logger
 
 MAX_LENGTH = 2000
+MAX_NUMBER_LENGTH = 10
+MAX_DECIMAL_PLACES = MAX_NUMBER_LENGTH - 2
 logger = get_logger()
 
 
-def get_field(answer, label, error_messages):
+def get_field(answer, label, error_messages, answer_store):
```
adoublebarrel, line 19: We're increasing the use of the answer store, which contains tainted input, in the codebase. While the context here *might* mean there isn't a vulnerability it's a pattern of use that needs consideration.

We should have a discussion about whether we default to escaping values returned by the answer store, enforce access via a proxy or some other solution.

app/forms/questionnaire_form.py
```diff
@@ -92,7 +92,7 @@ def get_other_answer(self, answer_id, option_index):
         return None
 
 
-def get_answer_fields(question, data, error_messages):
+def get_answer_fields(question, data, error_messages, answer_store):
```
adoublebarrel, line 95: See direct access to answer store comment above.

app/forms/questionnaire_form.py
```diff
@@ -101,7 +101,7 @@ def get_answer_fields(question, data, error_messages):
                 next(a['mandatory'] for a in question['answers'] if a['id'] == answer['parent_answer_id'])
 
         name = answer.get('label') or question.get('title')
-        answer_fields[answer['id']] = get_field(answer, name, error_messages)
+        answer_fields[answer['id']] = get_field(answer, name, error_messages, answer_store)
```
adoublebarrel, line 104: and again, I'll stop calling them out now!

tests/app/forms/test_custom_fields.py
```diff
@@ -1,18 +1,37 @@
...
+            self.fail("Exceptions not thrown by CustomIntegerField")
+
+    def test_decimal_field(self):
+        decimal_field = CustomDecimalField(_form=self.mock_form, _name='aName')
+        self.assertIsInstance(decimal_field, Field)
+
+        try:
+            decimal_field.process_formdata(['NonDecimal'])
+        except:
+            self.fail("Exception not be thrown by CustomDecimalField")
```
adoublebarrel: `<pedant>`missing a "should" here?`</pedant>`

app/forms/fields.py
```diff
@@ -163,24 +164,61 @@ def get_integer_field(answer, label, guidance, error_messages):
...
+        )
+
+
+def get_schema_defined_max_min_value(definition, answer_store):
+    if definition:
+        if 'value' in definition:
+            return definition['value']
+        else:
+            answer_id = definition['answer_id']
+            for answer in answer_store.answers:
```
ajmaddaford: `answer_store.filter(answer_id=answer_id)` will return a list of answers that match an answer_id and should be used.

app/forms/fields.py
```diff
@@ -163,24 +164,61 @@ def get_integer_field(answer, label, guidance, error_messages):
...
+    for x in range(0, MAX_NUMBER_LENGTH):
+        if x == 0 and min_or_max == 'min':
+            default = '-' + default
+        else:
+            if decimals > 0 and x == MAX_NUMBER_LENGTH - (decimals+1):
+                default = default + '.'
+            else:
+                default = default + '9'
+
+    return int(default) if decimals == 0 else float(default)
```
ajmaddaford: I know we discussed doing this, but seeing it I think we're doing the wrong thing. Using limits based on the 'string length' of a number are downstream limitations. I think any downstream limits should be explicit in the schema. 

For now I think eQ system limits of -9999999999 to 9999999999 and a maximum of 6 decimal places makes sense, and defaults of 0 to 9999999999 with no decimal places. I also think we should throw an exception if we get any values that exceed these (rather than setting to the system maximum)

ajmaddaford (changes requested):

app/forms/fields.py
ajmaddaford: You're right that there's not a vulnerability here as in the context it's being used only numeric values are allowed (the `validate_range_type` test in `test_schema_validation` ensures this). 

I agree we should have a discussion about use of tainted values from the answer store - I've added a note to Tuesdays tech discussion agenda.

app/data_model/questionnaire_store.py
dwyeradam: simplejson has a higher native support for Decimal values.

tests/app/forms/test_fields.py
```diff
@@ -322,15 +330,17 @@ def test_percentage_field(self):
             "label": "New to the market in 2014-2016",
             "mandatory": False,
             "q_code": "0810",
-            "type": "Percentage",
+            "type": "Percentage", "max_value": {"value": 100},
```
ajmaddaford: `max_value` should be on a new line (also applies to all other occurrences of this)

tests/app/validation/test_number_range_validator.py
```diff
@@ -70,3 +107,275 @@ def test_within_range_at_max(self):
...
+                   'options': [], 'mandatory': False,
+                   'validation':
+                      {'messages':
+                           {
+                            'INVALID_NUMBER': 'Please only enter whole numbers into the field.',
+                            'NUMBER_TOO_SMALL': 'The minimum value allowed is 10. Please correct your answer.'
+                           }
+                      },
+                   'description': '', 'id': 'test-range', 'type': 'Currency'
+                  }
```
ajmaddaford: The formatting needs looking at, at a minimum each key should be on a new line

ajmaddaford (changes requested):

data/schema/schema_v1.json
```diff
@@ -476,12 +481,37 @@
...
+                                                                    "max_value": {
+                                                                        "type": "object",
+                                                                        "properties": {
+                                                                            "value": {
+                                                                                "type": "integer"
+                                                                            },
+                                                                            "answer_id": {
+                                                                                "type": "string"
+                                                                            }
+                                                                        }
```
ajmaddaford: We should also be able to enforce the one of `value` or `answer_id` being present using the `anyOf` pattern used elsewhere in the schema.

ajmaddaford (changes requested):

ajmaddaford (approved):

adoublebarrel (approved):
LGTM

PR #1183, opened 2017-07-25
Removed default KID and added logging for KID use

jonnyshaw89:
I have removed the default KID of `EQ_USER_AUTHENTICATION_SR_PRIVATE_KEY` when no KID is passed for the JWE.
This is because all upstream systems are passing a KID for the JWE

### How to review 
Check that you are still able to authenticate.

### DIFF
--- a/app/cryptography/token_helper.py
+++ b/app/cryptography/token_helper.py
@@ -1,4 +1,5 @@
 import json
+from structlog import get_logger
 
 from jwcrypto import jwe, jwt
 from jwcrypto.common import base64url_decode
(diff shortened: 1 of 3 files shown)


app/cryptography/token_helper.py
```diff
@@ -8,19 +9,17 @@
...
-        try:
-            jwe_kid = extract_kid_from_header(encrypted_token)
-        except InvalidTokenException as e:
-            if default_kid:
-                jwe_kid = default_kid
-            else:
-                raise e
+        jwe_kid = extract_kid_from_header(encrypted_token)
+
+        logger.info("Decrypting JWE", KID=jwe_kid)
```
ajmaddaford: From the logging standards at https://digitaleq.atlassian.net/wiki/display/SDC/Logging, the keys should be lower case - KID -> kid

ajmaddaford (changes requested):

app/cryptography/token_helper.py
jonnyshaw89: I did have that originally, ill change it back :)

tests/app/cryptography/test_token_helper.py
```diff
@@ -87,16 +87,6 @@ def test_enc_missing(self):
 
         self.assert_in_decrypt_exception(jwe.decode(), "Algorithm not allowed")
 
-    def test_missing_kid_uses_default(self):
```
ajmaddaford, line 90: Should this be converted to a test for an exception when kid is missing from the JWE header rather than removed?

ajmaddaford (changes requested):

tests/app/cryptography/test_token_helper.py
jonnyshaw89: yeh probably.

ajmaddaford (approved):

adoublebarrel (approved):

PR #1185, opened 2017-07-27
Failed requests for URLs containing <no source> (Issue)

### LINKED ISSUE #1053: Failed requests for URLs containing <no source>
### Expected behaviour
No requests for URLs containing `<no source>`

### Actual behaviour
Logs show failed requests to `/s/css/census/themes/census/<no%20source>`

Unknown, these errors are from the server logs

adoublebarrel:
[Trello Card](https://trello.com/c/SX3FWZrm/1025-eq-survey-runner-1053-failed-requests-for-urls-containing-no-source-issue-xs)

Fixes https://github.com/ONSdigital/eq-survey-runner/issues/1053

After some poking around (and upgrading the gulp sass plugin) it looks
like a problem with the globs + postCSS + cssnano was resulting in a `themes/census/<no
source>` path being added to the gulp styles task. Seems the strongest candidate for the root cause so I've fixed it.

Minified assets with source maps are now generated for all environments.

### How to review
You can run the app locally as usual and everything is fine.

- You'll get minified versions of the CSS and JS
- You should see a comment at the end of the minified files detailing the source map url.
- In chrome you can search for the mapped files using CMD-P (CTRL-P for non-OSX)

Run `EQ_MINIMIZE_ASSETS=False yarn compile`
Run `EQ_MINIMIZE_ASSETS=False ./scripts/run_app.sh` from runners root dir

- All surveys should look the same
- Viewing the network traffic in dev tools/reviewing the server log should show non minified versions of the CSS and JS are being used.


gulp/styles.js
```diff
@@ -35,21 +35,31 @@ export function lint() {
 
 export function styles() {
   const minifyAssets = process.env.EQ_MINIMIZE_ASSETS === 'True'
+  const doSourceMaps = process.env.EQ_DEV_MODE === 'True'
```
jonnyshaw89: We are trying to stop the use of the generic "Dev Mode" so maybe if we want this feature it should have its own switch?

app/settings.py
```diff
@@ -36,7 +36,7 @@ def get_env_or_fail(key):
     return value
 
 
-EQ_MINIMIZE_ASSETS = parse_mode(os.getenv('EQ_MINIMIZE_ASSETS', 'False'))
+EQ_MINIMIZE_ASSETS = parse_mode(os.getenv('EQ_MINIMIZE_ASSETS', 'TRUE'))
```
ajmaddaford: This should be 'True' to be consistent with the other `parse_mode` calls

hamishtaplin (approved):

ajmaddaford (changes requested):

docker-compose.yml
```diff
@@ -31,6 +31,7 @@ services:
       EQ_ENABLE_CACHE: "True"
       EQ_DEVELOPER_LOGGING: "True"
       EQ_RABBITMQ_ENABLED: "True"
+      EQ_MINIMIZE_ASSETS: "True"
```
jonnyshaw89: No need to set this to True if the app defaults to True

jonnyshaw89 (changes requested):

app/settings.py
adoublebarrel: 🤦‍♂️

scripts/dev_settings.sh
```diff
@@ -9,7 +9,7 @@ if [ -z "$EQ_RABBITMQ_ENABLED" ]; then
 fi
 
 if [ -z "$EQ_MINIMIZE_ASSETS" ]; then
-  export EQ_MINIMIZE_ASSETS=False
+    export EQ_MINIMIZE_ASSETS=True
```
jonnyshaw89: If the `settings.py` defaults this to on then there is no need to also set it here.

jonnyshaw89 (changes requested):

jonnyshaw89 (approved):

ajmaddaford (approved):

jonnyshaw89:
@adoublebarrel squash?

PR #1186, opened 2017-07-28
Constructed Leaner Dockerfile

jonnyshaw89:
Created a much leaner Dockerfile that runs directly onto of the python base image.
Also created a container for static assets that can be used in AWS.

### How to review 
try `docker-compose build` then `docker-compose up -d`


Dockerfile
```diff
@@ -1,4 +1,7 @@
-FROM onsdigital/docker-aws-apache-wsgi:python-3.4.2-onbuild
+FROM python:3.4-onbuild
```
collisdigital, line 1: can we consider moving to Python 3.6 now?

jonnyshaw89: Maybe we do that as a separate card, as i think some of our code isn't compatible and would need a slight re-work

adoublebarrel (approved):
LGTM

collisdigital (approved):

PR #1187, opened 2017-07-28
Make jti claim mandatory

jonnyshaw89:
jti claim is now being passed by all upstream systems so we can now make it mandatory

### How to review 
Check that not passing a jti claim causes an authentication failure

### DIFF
--- a/app/storage/metadata_parser.py
+++ b/app/storage/metadata_parser.py
@@ -42,7 +42,6 @@ class MetadataField(object):
 
 
 metadata_fields = {
-    "jti": MetadataField(mandatory=False),
     "user_id": MetadataField(),
     "ru_ref": MetadataField(),
     "ru_name": MetadataField(),
--- a/tests/integration/session/test_login.py
+++ b/tests/integration/session/test_login.py
@@ -48,17 +48,13 @@ class TestLogin(IntegrationTestCase):
         # Then
         self.assertStatusUnauthorised()
 
-    def test_login_with_token_twice_is_authorised_when_no_jti(self):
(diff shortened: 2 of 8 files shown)


collisdigital (approved):

app/storage/metadata_parser.py
```diff
@@ -42,7 +42,7 @@ def generate(self):
 
 
 metadata_fields = {
-    "jti": MetadataField(mandatory=False),
+    "jti": MetadataField(),
```
ajmaddaford: I think we should consider moving the validation of the `jti` claim into `token_helper.py` - it's a JWT standard claim rather than an ONS metadata claim (longer term this would then be included in the [sdc-cryptography](https://github.com/ONSdigital/sdc-cryptography) library)

ajmaddaford (changes requested):

tests/integration/session/test_login.py
```diff
@@ -48,18 +48,6 @@ def test_login_with_token_twice_is_unauthorised_when_same_jti_provided(self):
...
-        token = self.token_generator.create_token_without_jti('0205', '1')
-        self.get('/session?token=' + token)
-
-        # When
-        self.get('/session?token=' + token)
-
-        # Then
-        self.assertStatusOK()
-        self.assertInUrl('/questionnaire/1/0205')
-
```
ajmaddaford: We need to have a test that tests what happens when a token doesn't have a jti

ajmaddaford (changes requested):

ajmaddaford (approved):

PR #1188, opened 2017-07-31
Investigate Travis Functional Test Failures

dwyeradam:
Travis functional tests are failing intermittently.

When capturing which tests fail it is normally one of the following:
'Given a currency option, a user should be able to click the label of the currency field to focus' or
'Given I am answering details correct question in the individual detail section, When I do not select any response, Then I am routed to Are you over 16'

These are the first tests in their specific test suites. 

Could be being caused by JWT generation taking too long. Short term fix could be to increase the sleep time. Other cards to migrate functional tests to the new format should be prioritised.

### DIFF
--- a/tests/functional/helpers.js
+++ b/tests/functional/helpers.js
@@ -12,7 +12,7 @@ export const startCensusQuestionnaire = (schema, sexualIdentity = false, region 
       return browser.url('/session?token=' + token)
     })
 
-  browser.pause(1000) // Shudder!!!
+  browser.pause(2000) // Double Shudder!!!
 }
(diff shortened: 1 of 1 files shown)


jonnyshaw89 (approved):

dwyeradam:
I have run Travis 6 times with no failures. Additional time on tests should be just under 2 minutes which was confirmed by my 6 tests (longest running 15 minutes 14 seconds).

adoublebarrel (approved):
We should move everything over to the new style but as a sticking plaster this will do...

PR #1190, opened 2017-08-03
Removed and replaced existing assert tests with exceptions with addit…

MIslam92:
…ion of new methods in answer_store

Context of this PR:
Refactoring of tests to remove all assertions in code and to be replaced with exceptions & also to add tests for these.
Also, new methods added to deal with "Location" in the answer_store

Trello Card: https://trello.com/c/93EgMm8q

How to review: 
Run tests

### DIFF
--- a/app/data_model/answer_store.py
+++ b/app/data_model/answer_store.py
@@ -159,23 +159,36 @@ class AnswerStore(object):
         """
         self._validate(answer)
 
-        return len(self.filter(answer.group_id, answer.block_id, answer.answer_id, answer.group_instance, answer.answer_instance))
+        return len(self.filter(answer.group_id, answer.block_id, answer.answer_id, answer.group_instance,
+                               answer.answer_instance))
 
-    def filter(self, group_id=None, block_id=None, answer_id=None, group_instance=None, answer_instance=None, location=None, limit=None):
+    def filter_by_location(self, location):
+        """
+        Find all answers in he answer store for a given location
+
+        :param location: The location to filter results by
+        return: Return a list of answers which match the location
+        """
+        return self.filter(group_id=location.group_id, block_id=location.block_id, group_instance=location.
+                           group_instance)
+
+    def filter(self, group_id=None, block_id=None, answer_id=None, group_instance=None, answer_instance=None,
+               limit=None):
         """
         Find all answers in the answer store for a given set of filter parameter matches.
         If no filter parameters are passed it returns a copy of the list of all answers.
+
 
         :param answer_id: The answer id to filter results by
         :param block_id: The block id to filter results by
         :param group_id: The group id to filter results by
         :param answer_instance: The answer instance to filter results by
         :param group_instance: The group instance to filter results by
-        :param location: The location to filter results by (takes precedence over group_id, group_instance and block_id)
         :param limit: True | False Limit the number of answers returned
         :return: Return a list of answers which satisfy the filter criteria
         """
         filtered = []
+
         filter_vars = {
             "answer_id": answer_id,
             "block_id": block_id,
@@ -183,13 +196,6 @@ class AnswerStore(object):
             "answer_instance": answer_instance,
             "group_instance": group_instance,
         }
-        if location:
-            assert not (group_id or group_instance or block_id), \
-                "Expected either a location object or one or more of group_id, group_instance, block_id params"
-
-            filter_vars['group_id'] = location.group_id
-            filter_vars['group_instance'] = location.group_instance
-            filter_vars['block_id'] = location.block_id
 
         for answer in self.answers:
             matches = True
@@ -229,14 +235,13 @@ class AnswerStore(object):
     def remove_answer(self, answer):
         """
         Removes an answer from the answer store.
-
         :param answer: A dict of flattened answer details.
         """
         index = self.find(answer)
         if index is not None:
             del self.answers[index]
 
-    def remove(self, group_id=None, block_id=None, answer_id=None, group_instance=None, answer_instance=None, location=None):
+    def remove(self, group_id=None, block_id=None, answer_id=None, group_instance=None, answer_instance=None):
         """
         Removes answer(s) from the answer store.
 
@@ -245,11 +250,19 @@ class AnswerStore(object):
         :param group_id: The group id to filter results to remove
         :param answer_instance: The answer instance to filter results to remove
         :param group_instance: The group instance to filter results to remove
-        :param location: The location to filter results to remove (takes precedence over group_id, group_instance and block_id)
         """
 
-        for answer in self.filter(group_id, block_id, answer_id, group_instance, answer_instance, location):
+        for answer in self.filter(group_id, block_id, answer_id, group_instance, answer_instance):
             self.answers.remove(answer)
+
+    def remove_by_location(self, location):
+        """
+        Removes answer(s) from the answer store based on Location
+
+        :param location: The location to filter results
+        """
+
+        self.remove(group_id=location.group_id, block_id=location.block_id, group_instance=location.group_instance)
 
 
 def number_else_string(text):
@@ -259,6 +272,7 @@ def number_else_string(text):
 def natural_order(key):
     """
     Orders a set of items according to
+
     :param key:
     :return:
     """
--- a/app/forms/household_composition_form.py
+++ b/app/forms/household_composition_form.py
@@ -1,3 +1,5 @@
+import logging
+from structlog import wrap_logger
 from flask_wtf import FlaskForm
 from wtforms import FieldList, Form, FormField
 
@@ -6,6 +8,8 @@ from app.forms.fields import get_string_field
 from app.helpers.schema_helper import SchemaHelper
 
 from werkzeug.datastructures import MultiDict
+
+logger = wrap_logger(logging.getLogger(__name__))
 
 
 def get_name_form(block_json, error_messages):
@@ -29,6 +33,7 @@ def serialise_composition_answers(location, data):
     for index, person_data in enumerate(data):
         for answer_id in person_data.keys():
             answer = Answer(
+
                 location=location,
                 answer_id=answer_id,
                 answer_instance=index,
@@ -79,11 +84,8 @@ def generate_household_composition_form(block_json, data, error_messages):
         def remove_person(self, index_to_remove):
             popped = []
 
-            assert len(self.household.data) >= index_to_remove, "Expected household data length >= index to remove"
-
             while index_to_remove != len(self.household.data):
                 popped.append(self.household.pop_entry())
-
             popped.reverse()
 
             for field in popped[1:]:
--- a/app/helpers/schema_helper.py
+++ b/app/helpers/schema_helper.py
@@ -127,7 +127,8 @@ class SchemaHelper(object):  # pylint: disable=too-many-public-methods
             for question in cls.get_questions_for_block(block):
                 for answer in question['answers']:
                     if 'alias' in answer:
-                        assert answer['alias'] not in aliases, 'Duplicate alias found: ' + answer['alias']
+                        if answer['alias'] in aliases:
+                            raise AssertionError('Duplicate alias found: ' + answer['alias'])
                         aliases[answer['alias']] = {
                             'answer_id': answer['id'],
                             'repeats': answer['type'] == 'Checkbox' or question['type'] == 'RepeatingAnswer',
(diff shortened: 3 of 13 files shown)


app/data_model/answer_store.py
```diff
@@ -159,9 +160,24 @@ def count(self, answer):
...
+    def filter_by_location(self, location):
+        """
+        Returns the value of the location.
+
+        :param location: The location to filter results by (takes precedence over group_id, group_instance and block_id)
+        :return: The value of the location found
+        """
+        group_id = location.group_id
+        group_instance = location.group_instance
+        block_id = location.block_id
```
jonnyshaw89: is there any point assigning these variable if they are only used once?

app/forms/household_composition_form.py
```diff
@@ -6,7 +8,9 @@
 from app.helpers.schema_helper import SchemaHelper
 
 from werkzeug.datastructures import MultiDict
+from werkzeug.exceptions import InternalServerError, NotFound, BadRequest
 
+logger = wrap_logger(logging.getLogger(__name__))
```
jonnyshaw89: we should stick to the standard of
```
from structlog import get_logger
logger = get_logger()
```

app/questionnaire/rules.py
```diff
@@ -114,7 +114,8 @@ def evaluate_when_rules(when_rules, metadata, answer_store, group_instance):
             answer_index = when_rule['id']
             filtered = answer_store.filter(answer_id=answer_index, group_instance=group_instance)
 
-            assert len(filtered) <= 1, "Condition will not met: Multiple ({:d}) answers found".format(len(filtered))
+            if len(filtered) <= 1:
+                raise Exception("Condition will not met: Multiple answers found".format(len(filtered)))
```
jonnyshaw89: Condition will not **be** met

app/submitter/converter.py
```diff
@@ -122,16 +122,17 @@ def _get_checkbox_answer_data(answer_store, checkboxes_with_qcode, value):
...
 
         if option:
-            if 'child_answer_id' in option:
-                filtered = answer_store.filter(answer_id=option['child_answer_id'])
+                if 'child_answer_id' in option:
+                    filtered = answer_store.filter(answer_id=option['child_answer_id'])
 
-                assert len(filtered) <= 1, 'Multiple answers found for "{0}"'.format(option['child_answer_id'])
+                    if len(filtered) >= 1:
+                        raise AssertionError("Multiple answers found {}".format(option['child_answer_id']))
```
jonnyshaw89: you have lost the 'for' from the message. is that intentional?

tests/app/data_model/test_answer_store.py
```diff
@@ -631,7 +631,7 @@ def test_remove_multiple_answers_by_location(self):
 
         location = Location("group1", 0, "block1")
 
-        self.store.remove(location=location)
+        self.store.remove_by_location(location)
```
jonnyshaw89: you should try and call these consistently. `(location=location)` vs `(location)`

tests/app/forms/test_household_composition.py
```diff
@@ -1,3 +1,4 @@
+# from werkzeug.exceptions import InternalServerError
```
jonnyshaw89: if we don't need it remove it.

jonnyshaw89 (dismissed):

app/data_model/answer_store.py
```diff
@@ -4,7 +4,8 @@
 
 
 class Answer(object):
-    def __init__(self, group_id=None, block_id=None, answer_id=None, value=None, group_instance=0, answer_instance=0, location=None):
+    def __init__(self, group_id=None, block_id=None, answer_id=None, value=None, group_instance=0, answer_instance=0,
+                 location=None):
         valid = (group_id or answer_id or block_id or value) is not None
 
         if location:
```
ajmaddaford: There is still an assert after this line. Looking at the code that uses the Answer class, we should be able to drop group_id, block_id and group_instance from the init params of this method and only accept location.

app/data_model/answer_store.py
```diff
@@ -248,9 +255,17 @@ def remove(self, group_id=None, block_id=None, answer_id=None, group_instance=No
         :param location: The location to filter results to remove (takes precedence over group_id, group_instance and block_id)
         """
 
-        for answer in self.filter(group_id, block_id, answer_id, group_instance, answer_instance, location):
+        for answer in self.filter(group_id, block_id, answer_id, group_instance, answer_instance):
             self.answers.remove(answer)
 
+    def remove_by_location(self, location):
+
```
ajmaddaford: This method should have documentation consistent with the other similar methods

app/data_model/answer_store.py
```diff
@@ -248,9 +255,17 @@ def remove(self, group_id=None, block_id=None, answer_id=None, group_instance=No
...
+        group_id = location.group_id
+        group_instance = location.group_instance
+        block_id = location.block_id
+
+        self.remove(group_id=group_id, block_id=block_id, group_instance=group_instance)
```
ajmaddaford: This method should be consistent with `filter_by_location` and not create the `group_id`, `group_instance` and `block_id` variables.

app/data_model/answer_store.py
```diff
@@ -159,9 +160,20 @@ def count(self, answer):
...
-        return len(self.filter(answer.group_id, answer.block_id, answer.answer_id, answer.group_instance, answer.answer_instance))
+        return len(self.filter(answer.group_id, answer.block_id, answer.answer_id, answer.group_instance,
+                               answer.answer_instance))
 
-    def filter(self, group_id=None, block_id=None, answer_id=None, group_instance=None, answer_instance=None, location=None, limit=None):
+    def filter_by_location(self, location):
+        """
+        Returns the value of the location.
+
+        :param location: The location to filter results by (takes precedence over group_id, group_instance and block_id)
```
ajmaddaford: The documentation for this should be something like:
```
Find all answers in the answer store for a given location

:param location: The location to filter results by
return: Return a list of answers which match the location
```

app/data_model/answer_store.py
```diff
@@ -159,9 +160,20 @@ def count(self, answer):
...
+        :return: The value of the location found
+        """
+        return self.filter(group_id=location.group_id, block_id=location.block_id, group_instance=location.group_instance)
+
+    def filter(self, group_id=None, block_id=None, answer_id=None, group_instance=None, answer_instance=None,
+               location=None, limit=None):
```
ajmaddaford: `location` should be removed from the parameter list (and the docstring)

app/data_model/answer_store.py
```diff
@@ -236,7 +242,8 @@ def remove_answer(self, answer):
         if index is not None:
             del self.answers[index]
 
-    def remove(self, group_id=None, block_id=None, answer_id=None, group_instance=None, answer_instance=None, location=None):
+    def remove(self, group_id=None, block_id=None, answer_id=None, group_instance=None, answer_instance=None,
+               ):
```
ajmaddaford: The `):` should be on the end of the line above (and the trailing command should be removed)

app/forms/household_composition_form.py
```diff
@@ -79,11 +83,12 @@ def answer_errors(self, input_id):
...
-
-            while index_to_remove != len(self.household.data):
-                popped.append(self.household.pop_entry())
-
+            try:
+                while index_to_remove != len(self.household.data):
+                    popped.append(self.household.pop_entry())
+            except AssertionError as e:
+                logger.error("Expected household data length greater then index to remove. {}".format(e))
+                raise AssertionError
```
ajmaddaford: Looking at the test for this below, we're checking for an `IndexError` rather than an `AssertionError`. Having seen how this works I think we can drop this try and except and if it happens it will be raised as an `IndexError` exception.

app/helpers/schema_helper.py
```diff
@@ -127,7 +127,8 @@ def get_aliases(cls, survey_json):
             for question in cls.get_questions_for_block(block):
                 for answer in question['answers']:
                     if 'alias' in answer:
-                        assert answer['alias'] not in aliases, 'Duplicate alias found: ' + answer['alias']
+                        if answer['alias'] not in aliases:
```
ajmaddaford: Shouldn't this be checking if `answer['alias']` is in `aliases`?

app/questionnaire/rules.py
```diff
@@ -114,7 +114,8 @@ def evaluate_when_rules(when_rules, metadata, answer_store, group_instance):
             answer_index = when_rule['id']
             filtered = answer_store.filter(answer_id=answer_index, group_instance=group_instance)
 
-            assert len(filtered) <= 1, "Condition will not met: Multiple ({:d}) answers found".format(len(filtered))
+            if len(filtered) <= 1:
+                raise Exception("Condition will not be met: Multiple answers found".format(len(filtered)))
```
ajmaddaford: - This is missing the `{:d}` from the original message
- The exception text could be improved, something like `Multiple answers ({:d}) found evaluating when rule for answer ({}).format(len(filtered), answer_index)`

app/questionnaire/rules.py
```diff
@@ -125,7 +126,7 @@ def evaluate_when_rules(when_rules, metadata, answer_store, group_instance):
             value = get_metadata_value(metadata, key)
             if not evaluate_rule(when_rule, value):
                 return False
-    return True
+        return True
```
ajmaddaford: I don't think this `return True` should have been moved?

app/submitter/converter.py
```diff
@@ -122,16 +122,17 @@ def _get_checkbox_answer_data(answer_store, checkboxes_with_qcode, value):
         option = next((option for option in checkboxes_with_qcode['options'] if option['value'] == user_answer), None)
 
         if option:
-            if 'child_answer_id' in option:
-                filtered = answer_store.filter(answer_id=option['child_answer_id'])
+                if 'child_answer_id' in option:
```
ajmaddaford: The indentation is off on these lines

app/submitter/converter.py
```diff
@@ -122,16 +122,17 @@ def _get_checkbox_answer_data(answer_store, checkboxes_with_qcode, value):
...
+                    filtered = answer_store.filter(answer_id=option['child_answer_id'])
 
-                assert len(filtered) <= 1, 'Multiple answers found for "{0}"'.format(option['child_answer_id'])
+                    if len(filtered) >= 1:
+                        raise AssertionError("Multiple answers found for {}".format(option['child_answer_id']))
```
ajmaddaford: We should be consistent in out use of `AssertionError`. `Exception` is used elsewhere in this pull request

tests/app/helpers/test_schema_helper.py
```diff
@@ -97,6 +97,14 @@ def test_get_parent_options_for_block(self):
 
         self.assertEqual(parent_options, expected)
 
+    def test_get_duplicate_aliases(self):
+        survey_json = load_schema_file("test_repeating_household.json")
+
+        with self.assertRaises(AssertionError)as err:
+            aliases = SchemaHelper.get_aliases(survey_json)
+            for alias in aliases:
+                self.assertEquals(('Duplicate alias found: ' + alias)), err(err.exception)
```
ajmaddaford: I'm not really sure what this is testing, you should need a survey with duplicate aliases to test the exception

tests/app/questionnaire/test_rules.py
```diff
@@ -426,3 +426,30 @@ def test_should_minus_one_from_maximum_repeats(self):
...
+                }
+            ]
+        }
+
+        answer_store = AnswerStore()
+        answer_store.add(Answer(answer_id='my_answers', value='1'))
+
+        # When
+        with self.assertRaises(Exception) as err:
+            evaluate_when_rules(when['when'], None, answer_store, 0)
```
ajmaddaford: To test this exception, you should have multiple answers with the same answer_id in the answer store

tests/app/submitter/test_converter.py
```diff
@@ -198,6 +198,84 @@ def test_answer_with_multiple_instances(self):
...
+                            }
+                        ]
+                    }
+                ]
+            }
+
+            with self.assertRaises(AssertionError) as err:
+                options = convert_answers(metadata, questionnaire, AnswerStore(answers), routing_path)
+                for option in options:
+                    self.assertEquals("Multiple answers found {}".format(option['child_answer_id'])),str(err.exception)
```
ajmaddaford: Looping over the options doesn't make any sense as `str(err.exception)` will always be the same value

ajmaddaford (dismissed):

app/data_model/answer_store.py
```diff
@@ -46,6 +46,8 @@ def matches_dict(self, answer_dict):
         :param answer_dict: A dict providing
         :return:
         """
+        # location = Location(group_id=answer_dict['group_id'], group_instance=answer_dict['group_instance'],
+        #                     block_id=answer_dict['block_id'])
```
shahi645: Delete these 2 if not being used?

shahi645 (commented):
LGTM👍🏾

PR #1191, opened 2017-08-07
displaying decimal places for currency on summary screen

shahi645:
All currency values piped in a survey or on the summary screen are shown with two decimal places

### How to review 
clone and run survey runner...summary screen should show currency with 2 DP.

### DIFF
--- a/app/jinja_filters.py
+++ b/app/jinja_filters.py
@@ -15,7 +15,7 @@ blueprint = flask.Blueprint('filters', __name__)
 @blueprint.app_template_filter()
 def format_currency(value):
     if value is not None and len(str(value)) > 0:
-        return "£{:,}".format(value)
+        return "£{:,.2f}".format(value)
     else:
         return ""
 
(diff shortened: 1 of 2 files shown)


app/jinja_filters.py
```diff
@@ -1,23 +1,18 @@
...
 blueprint = flask.Blueprint('filters', __name__)
 
 
 @blueprint.app_template_filter()
 def format_currency(value):
-    if value is not None and len(str(value)) > 0:
-        return "£{:,}".format(value)
-    else:
-        return ""
+    return "£{:,.2f}".format(value)
```
dwyeradam: Why has the if statement been removed? If a None or empty string value is passed to format_currency an exception will  now be thrown. You could have just added the .2f to the existing return.

shahi645: oh my fault...was meant to be there but i removed by accident...reverting

shahi645: done

yoganandkunche (approved):
LGTM :+1: work as expected. Failing functional test though

shahi645:
errors from pylint as i changed the imports around..changes reverted and functional tests pass

dwyeradam (approved):

ajmaddaford (approved):

PR #1192, opened 2017-08-07
Run surveys from urls

jonnyshaw89:
Survey Runner is able to load schemas from a URL as well as working as it currently does by loading local schemas

### How to review 
run docker-compose and test using `test_register' schema from https://github.com/ONSdigital/go-launch-a-survey/pull/19
This schema will load the survey json from the url


app/utilities/schema.py
```diff
@@ -37,23 +42,28 @@ def load_schema_file(schema_file, language_code=None):
...
+def load_schema(survey_url):
+    logger.info("loading schema from URL", survey_url=survey_url)
 
+    req = requests.get(survey_url)
+    schema_response = req.content.decode()
 
-def get_schema_path(schema_dir=DEFAULT_SCHEMA_DIR, language_code=DEFAULT_LANGUAGE_CODE):
-    return os.path.join(schema_dir, language_code)
+    if req.status_code != 200 or not schema_response:
+        logger.error("no schema exists", survey_url=survey_url)
```
adoublebarrel: Better to log this error on a 404 and log a dependency error of some kind on other non-200 codes/no response body?

tests/app/schema_loader/test_schema_loader.py
```diff
@@ -1,29 +1,54 @@
...
 from app.utilities.schema import load_schema_from_params
 
 from tests.app.app_context_test_case import AppContextTestCase
 
 
 class SchemaLoaderTest(AppContextTestCase):
 
+    @staticmethod
+    @urlmatch(netloc=r'(.*\.)?')
+    def register_mock(url, request): # pylint: disable=unused-argument
```
adoublebarrel: I *think* pylint will ignore vars beginning with `_` so you should be able to replace `request` with `_` and drop the disable comment.

tests/app/schema_loader/test_schema_loader.py
```diff
@@ -1,29 +1,54 @@
...
+        schema_path = os.path.join("tests/fixtures/data/", url.path[1:] + ".json")
+
+        with open(schema_path, encoding="utf8") as json_data:
+            return json_data.read()
+
+    @staticmethod
+    @urlmatch(netloc=r'(.*\.)?')
+    def register_mock_not_found(url, request): # pylint: disable=unused-argument
```
adoublebarrel: Same here (see above)

tests/integration/session/test_login.py
```diff
@@ -91,3 +112,27 @@ def test_login_with_missing_mandatory_claims_should_be_forbidden(self):
 
         # Then
         self.assertStatusForbidden()
+
+    def test_login_token_with_survey_url_should_redirect_to_survey(self):
+
+        survey_url = "http://eq-survey-register/my-test-schema"
+
+        # Given
+        token = self.token_generator.create_token_with_survey_url('0205', '1', survey_url)
```
adoublebarrel: After the conversation we had this morning, should we start moving away from using "real" surveys for testing? i.e. any new tests/modified tests are switched to use a suitable test_* fixture as we do other work.

adoublebarrel (commented):
It looks like you've moved generated schemas (mci_*, rsi_*) into the fixtures folder. I think they got accidentally committed, Yogi has added a git ignore rule for them.

Also looks like there's an accidental yarn.lock commit in here.

README.md
```diff
@@ -94,6 +94,11 @@ Note, you will also need to run an upstream tool (eg, https://github.com/ONSDigi
 docker run -it -p 8000:8000 onsdigital/go-launch-a-survey:latest
 ```
 
+You will also need to run a Survey Register for Survey Runner to load Schemas from 
+```
+docker run -it -p 8080:8080 onsdigital/eq-survey-register:simple-rest-api
+```
```
ajmaddaford: The survey register isn't required yet, as the existing production form types will still be in survey runner.

app/utilities/schema.py
```diff
@@ -13,7 +15,10 @@
 
 @cache.memoize()
 def load_schema_from_metadata(metadata):
-    return load_schema_from_params(metadata['eq_id'], metadata['form_type'], metadata.get('language_code'))
+    if 'survey_url' in metadata and metadata['survey_url'] is not None:
```
ajmaddaford: This could be simplified to `if metadata.get('survey_url'):`

tests/integration/create_token.py
```diff
@@ -53,6 +53,19 @@ def create_token_without_jti(self, form_type_id, eq_id, **extra_payload):
 
         return self.generate_token(payload_vars)
 
+    def create_token_with_survey_url(self, form_type_id, eq_id, survey_url, **extra_payload):
```
ajmaddaford: It might be worth creating a method that returns a payload to share between the other methods in this file, then adding the survey_url to that before calling generate_token.

ajmaddaford (changes requested):

jonnyshaw89:
@ajmaddaford @adoublebarrel ready for re-review

tests/integration/session/test_login.py
```diff
@@ -91,3 +110,41 @@ def test_login_with_missing_mandatory_claims_should_be_forbidden(self):
...
+    def test_login_token_with_survey_url_should_redirect_to_survey(self):
+
+        survey_url = "http://eq-survey-register/my-test-schema"
+
+        # Given
+        token = self.token_generator.create_token_with_survey_url('textarea', 'test', survey_url)
+
+        # When
+        with HTTMock(self.register_mock):
+            self.get('/session?token=' + token)
```
ajmaddaford: Is there an assert missing here?

tests/integration/session/test_login.py
```diff
@@ -91,3 +110,41 @@ def test_login_with_missing_mandatory_claims_should_be_forbidden(self):
...
+        # When
+        with HTTMock(self.register_mock_404):
+            self.get('/session?token=' + token)
+
+        # Then
+        self.assertStatusNotFound()
+
+    @staticmethod
+    @urlmatch(netloc=r'(.*\.)?')
+    def register_mock(_url, _request):
```
ajmaddaford: I think this method should be called `survey_url_mock` and the pattern match should match the `http://eq-survey-register/my-test-schema` url

tests/integration/session/test_login.py
```diff
@@ -91,3 +110,41 @@ def test_login_with_missing_mandatory_claims_should_be_forbidden(self):
...
+        schema_path = get_schema_file_path("1_0205.json")
+
+        with open(schema_path, encoding="utf8") as json_data:
+            return json_data.read()
+
+    @staticmethod
+    @urlmatch(netloc=r'(.*\.)?')
+    def register_mock_404(_url, _request):
```
ajmaddaford: Similar to above, and I think this should be named `survey_url_404_mock`

tests/integration/session/test_login.py
```diff
@@ -91,3 +110,41 @@ def test_login_with_missing_mandatory_claims_should_be_forbidden(self):
...
+        # Given
+        token = self.token_generator.create_token_with_survey_url('textarea', 'test', survey_url)
+
+        # When
+        with HTTMock(self.register_mock):
+            self.get('/session?token=' + token)
+
+    def test_login_token_with_incorrect_survey_url_results_in_404(self):
+
+        survey_url = "http://eq-survey-register/my-test-schema"
```
ajmaddaford: Although not required, it would be clearer if this url was different to the other one, something like `/my-test-schema-not-found`

tests/integration/views/test_jwt_authentication.py
```diff
@@ -49,8 +53,15 @@ def test_fully_encrypted(self):
...
-        self.assertEqual(302, response.status_code)
+        schema_path = get_schema_file_path("{}_{}.json".format("test", "0203"))
+
+        with open(schema_path, encoding="utf8") as json_data:
+            schema = json.load(json_data)
+
+            with patch('app.utilities.schema.load_schema_file', return_value=schema):
+
+                response = self.client.get('/session?token=' + encrypted_token)
+                self.assertEqual(302, response.status_code)
```
ajmaddaford: Why do we now need to patch load_schema_file?

ajmaddaford (changes requested):

tests/integration/views/test_jwt_authentication.py
jonnyshaw89: We don't any more

app/utilities/schema.py
```diff
@@ -11,9 +13,11 @@
 DEFAULT_LANGUAGE_CODE = 'en'
 
 
-@cache.memoize()
 def load_schema_from_metadata(metadata):
-    return load_schema_from_params(metadata['eq_id'], metadata['form_type'], metadata.get('language_code'))
+    if metadata.get('survey_url') and metadata.get('survey_url') is not "":
```
adoublebarrel: Pretty sure you can just do `if metadata.get('survey_url')` here as `""` is considered falsy

adoublebarrel (approved):
One small comment but nothing that should block this moving forward. LGTM.

tests/app/app_context_test_case.py
```diff
@@ -11,7 +11,7 @@ class AppContextTestCase(unittest.TestCase):
     def setUp(self):
         setting_overrides = {
             "EQ_SERVER_SIDE_STORAGE_DATABASE_DRIVER": "sqlite",
-            "EQ_SERVER_SIDE_STORAGE_DATABASE_NAME": ""
+            "EQ_SERVER_SIDE_STORAGE_DATABASE_NAME": "",
```
ajmaddaford: Extra `,` in an otherwise untouched file

ajmaddaford (changes requested):

ajmaddaford (approved):

PR #1196, opened 2017-08-09
Unit Patterns: Length, Area and Volume

dwyeradam:
- Allow for the display of Length, Area and Volume units
- New Unit schema type created with unit param
- Units defined by Unicode Common Locale Data Repository
- Allowable units can be found at http://unicode.org/repos/cldr/tags/latest/common/validity/unit.xml
- test_unit_patterns.json shows range of Length, Area and Volume fields initially requested.


requirements.txt
```diff
@@ -1,6 +1,5 @@
-Babel==2.2.0 \
-    --hash=sha256:fed07cbcdcb3de79b53a8220eebed21c93f8dbb3dbce1d9c6b1c4b09e8aecf2b \
-    --hash=sha256:d8cb4c0e78148aee89560f9fe21587aa57739c975bb89ff66b1e842cc697428f
+Babel==2.4.0 \
+    --hash=sha256:e86ca5a3a6bb64b9bbb62b9dac37225ec0ab5dfaae3c2492ebd648266468042f
```
jonnyshaw89: i think your missing a 2nd hash. if you run `hashin Babel` it will add the additional one.

jonnyshaw89 (changes requested):
Other than the small change all looks good!

app/templates/partials/summary/unit.html
```diff
@@ -0,0 +1 @@
+{{answer.unit|format_unit (answer.value)}}
```
ajmaddaford: The syntax for this looks odd, if two arguments are used they should be passed like `format_unit(answer.unit, answer.value)`. This is done for the `format_start_end_date` filter (using a context processor).

app/templates/partials/answers/unit.html
```diff
@@ -0,0 +1,26 @@
...
+        {% set input = {
+        'type': 'number',
+        'classes': input_classes,
+        'hasType': true,
+        'data': form.get_data(answer['id'])|e,
+        'name': input.name,
+        'id': answer.id
+        } %}
+        {% include 'partials/forms/input.html' %}
+        <span class="input-type__type" id="{{answer.id}}-type">{{answer.unit| format_unit ('')}}</span>
```
ajmaddaford: The second argument to `format_unit` should default to `''` so you wouldn't need to pass an empty string here.

data/schema/schema_units.json
```diff
@@ -0,0 +1,145 @@
+{
```
ajmaddaford: - I think we should just name this file `units.json` as it's already in a `schema` folder
- I think the first line should be `"$schema": "http://json-schema.org/draft-04/schema"`
- `unit` should have a type of `string`

ajmaddaford (changes requested):

app/jinja_filters.py
```diff
@@ -126,3 +133,8 @@ def start_end_date_check():
 @blueprint.app_context_processor
 def conditional_dates_check():
     return dict(format_conditional_date=format_conditional_date)
+
+
+@blueprint.app_context_processor
+def unit_check():
```
ajmaddaford: I think it would be clearer for this to be called `format_unit_processor`, I'm not sure why the others are called `_check`

app/jinja_filters.py
```diff
@@ -25,6 +27,11 @@ def format_percentage(value):
     return "{}%".format(value)
 
 
+@blueprint.app_template_filter()
```
ajmaddaford: I don't think this annotation is relevant now it's not a filter

ajmaddaford (changes requested):

dwyeradam:
@hamishtaplin  or @armstrongb  Could you give the couple of front end changes a review please. Changed unit block to allow for sq mi (square miles). May need to be bigger.

ajmaddaford (approved):

jonnyshaw89 (approved):

app/assets/styles/partials/components/_input-type.scss
```diff
@@ -73,3 +73,28 @@
     }
   }
 }
+
+.input-type--unit {
```
hamishtaplin: No need to duplicate the CSS from `.input-type--percentage`, just add the selector for `.input-type--unit` to that block instead:

```scss
  .input-type--percentage,
  .input-type--unit {
    // css
  }
```

hamishtaplin (dismissed):

app/assets/styles/partials/components/_input-type.scss
shahi645: done

jonnyshaw89:
@hamishtaplin this one is ready if you are happy with the fix?

hamishtaplin (approved):

jonnyshaw89:
@dwyeradam Can you squash this on?

PR #1198, opened 2017-08-10
Numeric decimal issue

### LINKED ISSUE #1197: A decimal after a numeric value does not throw error and displays as no answer provided on summary
### Expected behaviour
Must display an validation error

### Actual behaviour
No validation error displayed and on the summary page, it is displayed as no answer provided.
 
Launch any business survey which has numeric field entry question (ex. £ or employees field)
Enter a numeric value followed by a '.'
and click save and continue 
navigate to the summary page

yoganandkunche:
fixes #1197  

### How to review 
retest #1197

### DIFF
--- a/app/validation/error_messages.py
+++ b/app/validation/error_messages.py
@@ -8,6 +8,7 @@ error_messages = {
     'NUMBER_TOO_LARGE': 'This number is too big.',
     'INVALID_DATE_RANGE_TO_BEFORE_FROM': "The 'period to' date cannot be before the 'period from' date.",
     'INVALID_DATE_RANGE_TO_FROM_SAME': "The 'period to' date must be different to the 'period from' date.",
-    'MAX_LENGTH_EXCEEDED': "Your answer is too long, it has to be less than %(max)d characters",
-    'INVALID_DECIMAL': "Please only enter numbers to %(max)d decimal places into the field.",
+    'MAX_LENGTH_EXCEEDED': "Your answer is too long, it has to be less than %(max)d characters.",
+    'INVALID_DECIMAL': "Enter a number rounded to %(max)d decimal places.",
+    'INVALID_INTEGER': "Please only enter whole numbers into the field.",
 }
--- a/app/validation/validators.py
+++ b/app/validation/validators.py
@@ -99,8 +99,11 @@ class DecimalPlaces(object):
 
     def __call__(self, form, field):
         data = field.raw_data[0]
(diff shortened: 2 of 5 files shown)


jonnyshaw89 (approved):

app/validation/error_messages.py
```diff
@@ -9,5 +9,5 @@
     'INVALID_DATE_RANGE_TO_BEFORE_FROM': "The 'period to' date cannot be before the 'period from' date.",
     'INVALID_DATE_RANGE_TO_FROM_SAME': "The 'period to' date must be different to the 'period from' date.",
     'MAX_LENGTH_EXCEEDED': "Your answer is too long, it has to be less than %(max)d characters",
-    'INVALID_DECIMAL': "Please only enter numbers to %(max)d decimal places into the field.",
+    'INVALID_DECIMAL': "Please only enter whole numbers into the field.",
```
adoublebarrel: This copy change doesn't make sense for situations where there are decimal places required. The only reason it passes the functional tests is due to the error message being overridden for the test field, which I doubt we want to force authors to do for every decimal field that appears in a survey.

I suggest that we add an `INVALID_INTEGER` error message and adjust the code accordingly. We should also remove the overridden `INVALID_DECIMAL` error message from `test_numbers.json`

app/validation/validators.py
```diff
@@ -99,7 +99,7 @@ def __init__(self, max_decimals=0, messages=None):
 
     def __call__(self, form, field):
         data = field.raw_data[0]
-        if data is not None and '.' in data and len(data.split('.')[1]) > self.max_decimals:
+        if data and '.' in data and (self.max_decimals == 0 or len(data.split('.')[1]) > self.max_decimals):
```
adoublebarrel: This line's pretty hard to read;

1. `if data and '.' in data` could be rewritten as `if '.' in data`
1. `... and (self.max_decimals == 0 or len(data.split('.')[1]) > self.max_decimals`; looks like you could end up comparing two different types with the `>` opeartor, i.e. `True > 0`. This will work but it's pretty confusing.

This code should get simpler if we bring in an `INVALID_INTEGER` message but for future reference when you find yourself trying to pack this much on one line it's usually best to split it out across multiple lines and/or wrap part of it in a function.

adoublebarrel (changes requested):

adoublebarrel (approved):

PR #1201, opened 2017-08-14
Eq allow decimal places mwss pay increase

yoganandkunche:

This change will now allow respondents to enter 2 decimal places for MWSS 'pay increases' percentage field.

### How to review 

Launch MWSS survey 
Scenario 1: navigate to 2.10, 3.10, 4.10, 5.10 & 6.10 questions and enter percentage with 2 decimal places.
make sure the entered values are pushed down streams (dump)
Scenario 2: enter more than 2 decimal place must throw an error "Enter a number rounded to 2 decimal places."

### DIFF
--- a/data/en/1_0005.json
+++ b/data/en/1_0005.json
@@ -733,6 +733,7 @@
                                     "max_value": {
                                         "value": 100
                                     },
+                                    "decimal_places": 2,
                                     "validation": {
                                         "messages": {
                                             "NUMBER_TOO_LARGE": "The maximum value allowed is 100. Please correct your answer."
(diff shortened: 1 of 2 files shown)


shahi645:
In section 2.10 of MWSS there is a possible second percentage answer without the validation on 2 DP

data/en/1_0005.json
```diff
@@ -733,6 +733,7 @@
                                     "max_value": {
                                         "value": 100
                                     },
+                                    "decimal_places": 2,
                                     "validation": {
                                         "messages": {
                                             "NUMBER_TOO_LARGE": "The maximum value allowed is 100. Please correct your answer."
```
shahi645, line 739: does line 760: "weekly-pay-significant-changes-pay-rates-increase-percent-employees-answer" need:
validation and 2 decimal places?

yoganandkunche: No. I don't think so. as it is a free text field.

dwyeradam (approved):

shahi645 (approved):

PR #1202, opened 2017-08-16
Eq 1181 in service feedback feedback links

adoublebarrel:
[Trello Card](https://trello.com/c/Niqhzv70/1181-in-service-feedback-feedback-links-l)

This only has the No JS version of the feedback form available.

There's some heavy changes to the questionnaire view as I tried to understand and leverage existing questionnaire code in the feedback view. Form validation and template rendering is now re-used for both views. The new template rendering solution makes heavy use of decorators as this seemed a nice balance of declaring what your template dependencies are while also not drawing focus away from custom view rendering steps. There may be too many/few decorators, maybe decorators aren't the way to go.

### How to review 

For manual testing there's a feedback link below the header for all questionnaires which will lead you to the feedback form. Note that validation of the form is explicitly ruled out of the acceptance criteria for this card.

For automated testing there is an integration test which makes direct GETS/POSTS and a functional test which simulates the user.

There's also a strong code quality element to the changes; the questionnaire views actions for block GETS/POSTS should be more legible. The newly minted functions they're using need more work, tips and pattern suggestions for those are more than welcome.


app/submitter/converter.py
```diff
@@ -198,3 +198,48 @@ def _encode_value(value):
...
+        },
+        "submitted_at": "2016-03-07T15:28:05Z",
+        "metadata": {
+          "user_id": "789473423",
+          "ru_ref": "432423423423"
+        },
+        "data": {}
+      }
+    """
+    survey_id = "021"
```
jonnyshaw89: Is this a fixed value?

app/templates/feedback_sent.html
```diff
@@ -0,0 +1,16 @@
...
+
+{% block main %}
+
+<div class="u-mb-s">
+    <h1 class="saturn">Feedback Sent</h1>
+    <p>Thank you for your feedback.</p>
+</div>
+
+<div class="u-mt-s">
+    <p class="mars">It is now safe to close this window.</p>
```
jonnyshaw89, line 13: I think we need to look at the wording here. If it opens in a new tab will people close the entire browser window?

app/templates/layouts/_base.html
```diff
@@ -119,7 +117,12 @@
...
       <script src="{{ url_for('static', filename='js/bundle.js') }}"></script>
     {% endblock scripts %}
-  	<!--<![endif]-->
+  	  <!--<![endif]-->
 
+      <!-- START: ABS should of removed this before comitting -->
+     <script id="__bs_script__">//<![CDATA[
+         document.write("<script async src='http://HOST:3000/browser-sync/browser-sync-client.js?v=2.18.12'><\/script>".replace("HOST", location.hostname));
+         //]]></script>
+     <!-- END: ABS should of removed this before comitting -->
```
jonnyshaw89, line 123: Just commenting on this to give it more visibility

app/views/feedback.py
```diff
@@ -0,0 +1,108 @@
...
+    return redirect(url_for('feedback.thankyou'))
+
+
+@feedback_blueprint.route('/thankyou', methods=['GET'])
+@login_required
+def thankyou():
+    return _render_template('feedback_sent', page_title='Thankyou for your feedback')
+
+
+def _render_template(template, **kwargs):
```
jonnyshaw89: Is this duplicated from questionnaire view?

data/schema/feedback_schema_001.json
```diff
@@ -0,0 +1,70 @@
+{
```
jonnyshaw89: 👍

jonnyshaw89 (changes requested):
Just a few comments so far.

app/submitter/converter.py
adoublebarrel: Yes but it shouldn't be!

data/schema/feedback_schema_001.json
```diff
@@ -0,0 +1,70 @@
...
+  "$schema": "http://json-schema.org/draft-06/schema",
+  "$id": "https://edc.ons.gov.uk/schemas/feeback_schema_001.json",
```
ajmaddaford: - I think we can drop `schema` from the filename, seems a bit redundant as it's in a `schemas` folder. 
- The mapping from version number to filename won't work very well - version 1.0.0 would be `feeback_schema_100.json`. It might be worth sticking with a simple version number e.g. v1.

app/templates/partials/feedback_form.html
```diff
@@ -0,0 +1,28 @@
+<form method="POST" class="feeback__form {{ 'u-hidden' if feedback_form_hidden }}">
+    {{ form.csrf_token }}
+    <div class="group">
+        <div class="feedback__field">
+            <label id="feedback-message-label" class="label venus" for="feedback-message">How can we improve this survey?</label>
+            {% if feedback_form_hidden %}
+            <a class="feeback__close" href="#close">Close</a>
+            {% endif %}
+            <textarea class="input input--textarea js-charlimit-input" name="message" id="feedback-message" rows="8" cols="60" maxlength="2000"></textarea>
+            <div id="feedback-message-remaining" class="input__limit u-mt-xs mercury">Characters remaining: <span>-</span></div>
```
charlotte-hirst, line 11: The "Characters remaining" either needs to be the same font as the "Please include your name.." or bit of padding between that and the "Name" label

app/templates/partials/feedback_form.html
```diff
@@ -0,0 +1,28 @@
...
+                <div class="label__inner">
+                    Name (optional)
+                    <br />
+                    <span class="label__description pluto">Please include your name and email address if you're interested in taking part in further research to improve this survey</span>
+                </div>
+            </label>
+            <input type="text" class="input" name="name" id="feedback-name" value=""/>
+        </div>
+        <div class="feedback__field">
+            <label id="feedback-email-label" class="label venus" for="feedback-email">Email address (optional)</label>
```
charlotte-hirst, line 24: Again I think needs a little padding looks too close to above field

app/templates/partials/feedback_form.html
```diff
@@ -0,0 +1,28 @@
+<form method="POST" class="feeback__form {{ 'u-hidden' if feedback_form_hidden }}">
```
charlotte-hirst, line 2: When it opens new tab should title be "ONS feedback" or "Survey feedback" as its currently "ONS survey" which is misleading

app/templates/partials/feedback_form.html
```diff
@@ -0,0 +1,28 @@
...
+        <div class="feedback__field">
+            <label id="feedback-message-label" class="label venus" for="feedback-message">How can we improve this survey?</label>
+            {% if feedback_form_hidden %}
+            <a class="feeback__close" href="#close">Close</a>
+            {% endif %}
+            <textarea class="input input--textarea js-charlimit-input" name="message" id="feedback-message" rows="8" cols="60" maxlength="2000"></textarea>
+            <div id="feedback-message-remaining" class="input__limit u-mt-xs mercury">Characters remaining: <span>-</span></div>
+        </div>
+        <div class="feedback__field">
+            <label id="feedback-name-label" class="label venus" for="feedback-name">
```
charlotte-hirst, line 14: as there are no limits on name and email address, i take it there is no problem with sending long strings to SDX, i put over 2000characters in name field?

charlotte-hirst:
Does it store URL at the point they gave feedback?

app/templates/partials/feedback_form.html
adoublebarrel: The latest copy revision makes the form exclusively reference the survey so I guess we go with Survey feedback?

app/templates/partials/feedback_form.html
adoublebarrel: We don't have length validation on any text input fields. You wouldn't be able do that on pre-prod or live due to the WAF

app/templates/partials/feedback_form.html
adoublebarrel: ~~That's the same component as we use in the questionnaires so changing the styling will change it across the whole product. Does that give you a preference over padding vs font change?~~ Problem was due to the most recent CSS files not coming through see next comment.

app/templates/partials/feedback_form.html
adoublebarrel: ~~Shall I just go back to the padding we use in questionnaires?~~ This was because the styles weren't loading properly; I hadn't manually update the static server instance with the latest changes.

app/submitter/converter.py
```diff
@@ -200,3 +200,47 @@ def _encode_value(value):
...
         return value
+
+
+def convert_feedback(metadata, feedback_json, survey_id):
+    """
+    Create the JSON answer format for down stream processing
+    :param metadata: metadata for the questionnaire
+    :param questionnaire_json: the questionnaire json
+    :param answer_store: the users answers
+    :param routing_path: the path followed by the user when answering the questionnaire
```
jonnyshaw89: the `:param` values need updating in this comment

adoublebarrel:
No. I got it in my head the referrer was just for sending the user back to the page they came from to give feedback so thought we didn't need it as a new tab/window is opened.

I'm adding it to the message payload we send to SDX.

app/templating/template_renderer.py
```diff
@@ -20,6 +20,8 @@ def __init__(self):
         self.environment.filters['format_household_summary'] = format_household_summary
         self.environment.globals['format_conditional_date'] = format_conditional_date
 
+        self.environment.globals['format_conditional_date'] = format_conditional_date
```
jonnyshaw89: Duplicated line?

app/views/questionnaire.py
```diff
@@ -507,43 +590,28 @@ def get_page_title_for_location(schema_json, current_location):
 
 def _build_template(current_location, context, template, routing_path=None):
     metadata = get_metadata(current_user)
-    metadata_context = build_metadata_context(metadata)
-
     answer_store = get_answer_store(current_user)
     front_end_navigation = _get_front_end_navigation(answer_store, current_location, metadata, routing_path)
-
-    path_finder = PathFinder(g.schema_json, answer_store, metadata)
+    logger.info("Building")
```
jonnyshaw89: prob don't need this any more?

app/views/questionnaire.py
```diff
@@ -219,40 +210,125 @@ def get_session_expired(eq_id, form_type, collection_id):  # pylint: disable=unu
...
+
+
+def _valid_location(location):
+    return location in path_finder.get_routing_path(
+        location.group_id,
+        location.group_instance,
+    )
+
+
+def _valid_group(group_id):
```
jonnyshaw89: I think we should make the naming of these private methods consistent?
e.g. the ones that return a bool start with is. e.g. `_is_skipping_to_the_end`
it just makes it clearer in my opinion

app/views/questionnaire.py
```diff
@@ -507,43 +590,28 @@ def get_page_title_for_location(schema_json, current_location):
...
-
-    survey_data = metadata_context['survey']
-    url_prefix = '/questionnaire/{}/{}/{}'.format(survey_data['eq_id'], survey_data['form_type'], survey_data['collection_id'])
 
+@template_helper.with_session_timeout
+@template_helper.with_questionnaire_url_prefix
+@template_helper.with_metadata
+@template_helper.with_analytics
+@template_helper.with_legal_basis
+def _render_template(context, current_location, template, front_end_navigation, previous_url, **kwargs):
```
jonnyshaw89, line 560: 👍 I really like this

jonnyshaw89 (changes requested):

app/templating/template_renderer.py
adoublebarrel: 🤦‍♂️

app/templates/feedback_sent.html
adoublebarrel: Had this reviewed for content and flagged with BenA, looks like it's going to stay as-is.

app/helpers/form_helper.py
```diff
@@ -137,3 +142,12 @@ def clear_other_text_field(data, questions_for_block):
                 form_data[answer['id']] = ''
 
     return form_data
+
+
+def is_valid_form(form):
+    is_valid = form.validate()
+    if form.csrf_token.errors:
+        logger.warn('Invalid csrf token')
+        raise Forbidden
```
ajmaddaford: We shouldn't raise `Forbidden` here, this method should just return if the form is valid or not. It might be worth looking at adding CSRF protection more globally for FlaskForm objects - http://flask-wtf.readthedocs.io/en/stable/csrf.html

app/views/feedback.py
```diff
@@ -0,0 +1,93 @@
...
+        g.schema_json = load_schema_from_metadata(metadata)
+
+
+@feedback_blueprint.route('', methods=['GET'])
+@login_required
+def get_form():
+    form = FeedbackForm()
+    return _render_template(
+        'feedback',
+        page_title='Survey feedback',
```
ajmaddaford: The title text should be set in the template, should be settable via `{% block page_title %}Survey feedback{% endblock %}`

app/views/feedback.py
```diff
@@ -0,0 +1,93 @@
...
+
+        metadata = get_metadata(current_user)
+
+        message = convert_feedback(
+            metadata,
+            {
+                'message': escape(request.form.get('message')),
+                'name': escape(request.form.get('name')),
+                'email': escape(request.form.get('email')),
+                'referer': request.referrer or '',
```
ajmaddaford: - I don't think the property should be called `referer`, that's just how we're getting the URL that they are providing feedback for. Something more generic like `url` would be better.
- I think the four message properties should be passed explicitly to the convert_feedback method - `convert_feedback(message, name, email, url, metadata, survey_id)`, as it stands it implies any payload can be provided

app/views/feedback.py
```diff
@@ -0,0 +1,93 @@
...
+        if not sent:
+            raise SubmissionFailedException()
+
+    return redirect(url_for('feedback.thankyou'))
+
+
+@feedback_blueprint.route('/thankyou', methods=['GET'])
+@login_required
+def thankyou():
+    return _render_template('feedback_sent', page_title='Thankyou for your feedback')
```
ajmaddaford: Same as previous comment, the page title should be set in the template.

ajmaddaford (changes requested):
Looking good 👍  
I've done an initial review, but need to take a deeper look into the changes to questionnaire.py

app/views/questionnaire.py
```diff
@@ -219,40 +210,125 @@ def get_session_expired(eq_id, form_type, collection_id):  # pylint: disable=unu
...
+        disable_mandatory)
+
+    return wtf_form
+
+
+def _submit_answers(eq_id, form_type, collection_id):
+    metadata = get_metadata(current_user)
+    answer_store = get_answer_store(current_user)
+
+    return submit_answers(eq_id, form_type, collection_id, metadata, answer_store)
```
ajmaddaford: I'm not sure it's worth having this `_submit_answers` method given that's in only called in one place

app/views/questionnaire.py
```diff
@@ -219,40 +210,125 @@ def get_session_expired(eq_id, form_type, collection_id):  # pylint: disable=unu
...
+
+def _next_location_url(location):
+    metadata = get_metadata(current_user)
+    return location.url(metadata)
+
+
+def _is_end_of_questionnaire(block, location):
+    end_blocks = ["Summary", "Confirmation"]
+
+    return location is None and \
```
ajmaddaford: It's a little confusing as it reads like `location` would be the location of the block, it would be clearer named `next_location`

app/views/questionnaire.py
```diff
@@ -219,40 +210,125 @@ def get_session_expired(eq_id, form_type, collection_id):  # pylint: disable=unu
...
+        block['type'] in end_blocks
+
+
+def _is_skipping_to_the_end(block, current_location):
+    latest_location = path_finder.get_latest_location(
+        get_completed_blocks(current_user),
+        routing_path=path_finder.get_routing_path(),
+    )
+    end_blocks = ['Summary', 'Confirmation']
```
ajmaddaford: `end_blocks` is defined twice, it might be worth moving these into a constant

app/views/questionnaire.py
```diff
@@ -219,40 +210,125 @@ def get_session_expired(eq_id, form_type, collection_id):  # pylint: disable=unu
...
+        _remove_repeating_on_household_answers(get_answer_store(current_user), group_id)
+    return post_block(eq_id, form_type, collection_id, group_id, group_instance, 'permanent-or-family-home')
+
+
+def _render_block(block, current_location, post_form=None):
+
+    if post_form is None:
+        context = _get_context(block, current_location, get_answer_store(current_user))
+    else:
+        context = {'form': post_form, 'block': block}
```
ajmaddaford: This context is created in two other places - `_save_sign_out`, `post_household_composition` - it would be worth exploring getting them to call `_render_block` instead of `_build_template`

app/views/questionnaire.py
```diff
@@ -457,6 +538,8 @@ def _get_context(block, current_location, answer_store):
 
     error_messages = SchemaHelper.get_messages(g.schema_json)
     form, template_params = get_form_for_location(block, current_location, answer_store, error_messages)
+    form.validate_on_submit()
```
ajmaddaford: What is this for?

ajmaddaford (changes requested):

app/helpers/form_helper.py
adoublebarrel: Yeah, I lazily transplanted it from [the questionnaire view](https://github.com/ONSdigital/eq-survey-runner/blob/master/app/views/questionnaire.py#L409). I'll give a custom error handler [as the docs detail](http://flask-wtf.readthedocs.io/en/stable/csrf.html#customize-the-error-response) a go. It should hopefully just slot straight in.

app/views/questionnaire.py
adoublebarrel: :man_facepalming:

app/views/questionnaire.py
adoublebarrel: Um... Yeah, that's random. I had a lot of fun with rebase earlier in the week so I'm going to blame it on that.

app/helpers/form_helper.py
adoublebarrel: While I'm here, technically this should probably be a Bad Args rather than Forbidden. Is there a specific reason it's a Forbidden error code currently?

ajmaddaford: I don't think so. Looking at FlaskWTF there is a [CSRFError](http://flask-wtf.readthedocs.io/en/stable/api.html#flask_wtf.csrf.CSRFError) that generates a 400 error.

app/views/questionnaire.py
```diff
@@ -341,7 +312,7 @@ def is_survey_completed():
     return True, None
 
 
-def _save_sign_out(collection_id, eq_id, form_type, this_location, form):
+def _save_sign_out(_cid, _eqid, _formtype, this_location, form):
```
ajmaddaford: These parameters should be removed if they're not used

data/schema/feedback_v1.json
```diff
@@ -0,0 +1,73 @@
+{
+  "$schema": "http://json-schema.org/draft-06/schema",
+  "$id": "https://edc.ons.gov.uk/schemas/feeback_schema_001.json",
```
ajmaddaford: Should be `feedback_v1.json`

app/helpers/form_helper.py
```diff
@@ -5,6 +5,10 @@
 from app.forms.household_relationship_form import build_relationship_choices, deserialise_relationship_answers, generate_relationship_form
 from app.forms.questionnaire_form import generate_form
 
+from structlog import get_logger
+
+logger = get_logger()
+
```
ajmaddaford: It doesn't look like logger is used

app/views/errors.py
```diff
@@ -73,6 +81,7 @@ def log_exception(error, status_code):
 def _render_error_page(status_code):
     tx_id = get_tx_id()
     user_agent = user_agent_parser.Parse(request.headers.get('User-Agent', ''))
+    logger.info('rendering error.html')
```
ajmaddaford: I'm not sure this log line is needed?

app/templates/csrf_exception.html
```diff
@@ -0,0 +1,14 @@
...
+
+{% block page_title %}Information{% endblock %}
+
+{% block main %}
+  <div>
+    <h1 class="saturn">Information</h1>
+  </div>
+  <div class="panel panel--simple panel--error panel--spacious" data-qa="stale-survey-error">
+    <p><strong>Your session has gone stale</strong></p>
+    <div>To continue with this survey please log in again.</div>
```
ajmaddaford: We'll need to check the content for this

tests/functional/pages/multiple-survey.page.js
```diff
@@ -1,9 +0,0 @@
-class MultipleSurveys {
-
-  hasMultipleSurveyError() {
-    return browser.isExisting('[data-qa="multiple-survey-error"]')
-  }
-
-}
-
-export default new MultipleSurveys()
```
ajmaddaford, line 9: Do we still get the multiple surveys error on GET requests to 'old' surveys?

ajmaddaford (changes requested):

data/schema/feedback_v1.json
adoublebarrel: 🤦‍♂️

app/assets/js/app/modules/timeout.js
```diff
@@ -64,9 +64,9 @@ class TimeoutUI {
 domready(() => {
   const sessionTimeout = window.__EQ_SESSION_TIMEOUT__
   const promptTime = window.__EQ_SESSION_TIMEOUT_PROMPT__
-  const sessionExpiredUrl = window.__EQ_SESSION_EXPIRED_URL__
-  const expireSessionUrl = window.__EQ_EXPIRE_SESSION_URL__
-  const sessionContinueUrl = window.__EQ_SESSION_CONTINUE_URL__
+  const sessionExpiredUrl = '/session-expired'
```
jonnyshaw89, line 67: 👍

tests/integration/questionnaire/test_questionnaire_session_expired.py
```diff
@@ -5,7 +5,8 @@ class TestSessionExpired(IntegrationTestCase):
 
     def test_session_expired_should_log_user_out(self):
         self.launchSurvey('1', '0205')
-        self.post(url='/questionnaire/1/0205/789/expire-session')
+        startPage = self.last_url
+        self.post(url='/expire-session')
         self.assertStatusOK()
-        self.post(url='/questionnaire/1/0205/789/expire-session')
+        self.get(url=startPage)
```
jonnyshaw89, line 11: 👍

tests/functional/pages/multiple-survey.page.js
adoublebarrel: Yes and I've changed the behaviour for the CSRF failing to display the same message rather the "stale session" message now. Means that as far as a user is concerned nothing has changed.

app/views/feedback.py
```diff
@@ -0,0 +1,83 @@
...
+                                                        current_app.config['EQ_RABBITMQ_QUEUE_NAME'],
+                                                        metadata['tx_id'])
+
+        if not sent:
+            raise SubmissionFailedException()
+
+    return redirect(url_for('feedback.thankyou'))
+
+
+@feedback_blueprint.route('/thankyou', methods=['GET'])
```
ajmaddaford: The url should be hyphen separated - `thank-you` - to be consistent with the other urls

ajmaddaford (changes requested):

ajmaddaford (approved):

jonnyshaw89 (approved):

PR #1203, opened 2017-08-17
Remove Sections from Schema

dwyeradam:
- Remove sections from within Blocks in schemas
- New json format: Group -> Block -> Questions -> Answers


jonnyshaw89 (approved):
Not really that much code changed here. Just a lot of schemas to update.

adoublebarrel (approved):
I think you should get some kind of award for having to do all of that by hand.

PR #1205, opened 2017-08-17
Client Side Packages Updated and Chrome Headless

adoublebarrel:
This PR has been lifted out of the feedback work I've been doing as:

a) it will reduce the review burden in the feedback PR
b) Switching to Chrome headless for Karma tests ~~and Travis functional tests is a one liner but isn't relevant to the feedback work~~

All client dependencies have been updated to their latest versions; this made using hot-reloading via browsersync much more workable. There are knock on changes to some config (e.g. eslint) where packages had deprecated some features.

PhantomJS has been switched out in favour of Chrome headless for Karma tests only. This removes a dev dependency from the client and means the Karma tests on Travis now use the latest stable version of Chrome rather than a version of webkit that is slowly atrophying.

Sadly we can't currently make use of Chrome headless for the functional tests. Chrome in headless mode doesn't support the API that ChromeDriver currently uses to resize browser windows. The functional tests resize the browser window for its mobile testing (e.g. for testing the navigation). There are potential workarounds but sticking with Phantom until a new release of ChromeDriver is the easiest.

### How to review

All yarn commands/gulp scripts should work as before. The Travis build should pass and if you look in the logs you should see that the Travis chrome add on is being used when running the Karma tests.


jonnyshaw89 (changes requested):
I get errors when i run `./scripts/run_tests_functional.sh` also there are still phantomjs dependencies in the package?

jonnyshaw89:
```
18 08 2017 09:22:09.378:ERROR [plugin]: Error during loading "karma-phantomjs-launcher" plugin:
  Path must be a string. Received null
```

adoublebarrel:
Ah, overlap! Realised that I'd only switched out Chrome for PhantomJS in Karma. Was all working for me though so not sure why your functional tests bombed. Will catch up with you about that on Slack.

adoublebarrel:
Ok, Chrome headless is a no go for the functional tests :disappointed:  I've updated the PR accordingly.

jonnyshaw89 (approved):

dwyeradam (approved):

jonnyshaw89:
@adoublebarrel Squish Squash?

PR #1206, opened 2017-08-21
Eq 1299 currency unit patterns

shahi645:
Implemented currency unit pattern, and currency fields allow for commas to be entered 

### How to review 
Run survey-runner and load any survey with currency. Test_currency.json allows for decimal places, but not implemented in live surveys. 

Playback also formats entered currency correctly


jonnyshaw89:
@shahi645 Can you rebase this so it remove the unit pattern work from the diff?

app/forms/fields.py
```diff
@@ -168,7 +168,11 @@ def get_number_field(answer, label, guidance, error_messages, answer_store):
 
     mandatory_or_optional = get_mandatory_validator(answer, answer_errors)
 
-    max_decimals = answer.get('decimal_places', 0)
+    if answer['type'] == "Currency":
+        max_decimals = 2
```
dwyeradam: Defaulting currency to two decimal places should be removed until agreed by all business areas as this is not the behaviour of production. When rolling out the use of pence within the currency we can add the "decimal_places": 2 to the schema.

data/en/1_0001.json
```diff
@@ -4714,7 +4724,8 @@
     "messages": {
         "NUMBER_TOO_LARGE": "The maximum value allowed is 9999999999. Please correct your answer.",
         "NUMBER_TOO_SMALL": "The value cannot be negative. Please correct your answer.",
-        "INVALID_NUMBER": "The value you have entered is invalid. Please enter a numeric value."
+        "INVALID_NUMBER": "The value you have entered is invalid. Please enter a numeric value.",
+        "INVALID_DECIMAL": "Please enter a number to 2 decimal places."
```
dwyeradam: If we remove the default pence allowed for currency values then we also need to remove the INVALID_DECIMAL error message.

tests/app/test_jinja_filters.py
```diff
@@ -3,37 +3,39 @@
...
+        format_value = get_currency_symbol(currency)
 
-        self.assertEqual(format_value, '£1.00')
+        self.assertEqual(format_value, '£')
 
     def test_format_currency(self):
         # Given
-        currency = 1.12
+        currency = "GBP"
+        value = '11'
```
jonnyshaw89: Do you have a test with a decimal value?

jonnyshaw89: You said you couldn't accept decimals? I've just run this with a decimal and it worked?
US$2.45

dwyeradam: The format_currency will accept any number of decimal places and will return it rounded to 2.
It would therefore be worth testing values with 2 and more decimal places.

shahi645: ok will make the changes accordingly

tests/new_functional/spec/dates_conditional.spec.js
```diff
@@ -26,7 +26,7 @@ describe('Piped Dates', function () {
       // Then the summary screen shows the dates entered formatted
       .getText(SummaryPage.answer()).should.eventually.contain('11 October 2017 to 03 December 2017');
 
-    });
```
jonnyshaw89: Why have you removed this semi-colon?

shahi645: this was failing the functional tests in travis

app/jinja_filters.py
```diff
@@ -2,24 +2,44 @@
...
+    if value:
+        try:
+            formatted_number = numbers.format_number(value, locale='en_GB')
+            if decimal_places and decimal_places > 0:
+                if '.' not in formatted_number:
+                    formatted_number += '.'
+                while len(formatted_number.split('.')[1]) < decimal_places:
+                    formatted_number += '0'
+        except InvalidOperation:
+            return value
```
ajmaddaford: When is an InvalidOperation raised?

app/jinja_filters.py
```diff
@@ -2,24 +2,44 @@
...
+def get_formatted_number(value=None, decimal_places=0):
+    formatted_number = ''
+    if value:
+        try:
+            formatted_number = numbers.format_number(value, locale='en_GB')
+            if decimal_places and decimal_places > 0:
+                if '.' not in formatted_number:
+                    formatted_number += '.'
+                while len(formatted_number.split('.')[1]) < decimal_places:
+                    formatted_number += '0'
```
ajmaddaford: I don't think we want to be padding all decimal numbers up to their decimal place limit with 0's. For example, if someone enters 100.1 for a field that can take 3 decimal places, we'd output 100.100 which doesn't look right.

I think the zero padding is only a currency issue, and this method should be made specific for that, something like:
```
def format_currency_for_input(value, decimal_places=None):
    if decimal_places == 0:
        format = return numbers.format_number(value, locale='en_GB')
    else:
        return numbers.format_decimal(value, format='#,###.00', locale='en_GB')
```
Also, as shown above, the zero padding should be possible using a babel format string.

app/jinja_filters.py
```diff
@@ -2,24 +2,44 @@
...
+
+
+@blueprint.app_template_filter()
+def get_currency_symbol(currency="GBP"):
+    return numbers.get_currency_symbol(currency, locale='en_GB')
+
+
+@blueprint.app_template_filter()
+def format_currency(value=None, currency="GBP"):
+    return numbers.format_currency(number=value, currency=currency, locale='en_GB')
```
ajmaddaford: Given we're using the `en_GB` locale in a few places it would we better to define it as a constant, something like `DEFAULT_LOCALE='en_GB'`

data/en/test_currency.json
```diff
@@ -10,26 +10,61 @@
     "messages": {
         "NUMBER_TOO_LARGE": "Number is too large",
         "NUMBER_TOO_SMALL": "Number cannot be less than zero",
-        "INVALID_NUMBER": "Please enter an integer"
+        "INVALID_DECIMAL": "Please enter a number to 2 decimal places"
```
ajmaddaford: You should use the placeholder value `%(max)d` instead of `2` so that if the answers change the message will be updated automatically. This would also mean the answers can have varied decimal places.

data/schema/currencies.json
```diff
@@ -0,0 +1,23 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema",
```
ajmaddaford, line 2: Where did the list of currencies come from? We should add a description line to document what standard they follow e.g. ISO 4217

tests/app/test_jinja_filters.py
```diff
@@ -3,39 +3,33 @@
...
-        # When
-        format_value = format_currency(currency)
-
-        self.assertEqual(format_value, '£1.00')
+    def test_get_formatted_number(self):
+        self.assertEqual(get_formatted_number('100.00', 2), '100.00')
+        self.assertEqual(get_formatted_number('1000'), '1,000')
+        self.assertEqual(get_formatted_number('10000'), '10,000')
+        self.assertEqual(get_formatted_number('100000000'), '100,000,000')
+        self.assertEqual(get_formatted_number('100.90', 2), '100.90')
```
ajmaddaford: Needs tests added for:
- One decimal place in input, output defined as more - `self.assertEqual(get_formatted_number('100.1', 3), '100.100')`
- Output containing `,` and `.` - `self.assertEqual(get_formatted_number('10000', 2), '10,000.00')`

app/helpers/form_helper.py
```diff
@@ -128,7 +128,9 @@ def clear_other_text_field(data, questions_for_block):
     """
     form_data = MultiDict(data)
     for question in questions_for_block:
-        for answer in question['answers']:
+        for answer in question.get('answers', []):
+            if answer.get('type') == 'Currency':
+                form_data[answer['id']] = form_data.get(answer['id'], '').replace(',', '')
```
ajmaddaford: This shouldn't be done here, as it has nothing to do with the `other text field` functionality. It should be moved to the `post_form_for_location` or `generate_form` method.

ajmaddaford (changes requested):

app/jinja_filters.py
dwyeradam: This function is called before the validators errors are handled. Therefore if someone entered a text rather than numeric value the babel formatter would throw the InvalidOperation exception. By passing the value as it stands allows our number validator to raise the exception instead.

ajmaddaford:
I think we have an issue with amending the value when there is a validation error. For both of these examples the error does not make sense as the input has been fixed.

In the test_currency survey, after entering 1.256 in the first input and clicking submit, I get:
image

In the RSI 0102 survey, after entering 1.00 for Total retail turnover and clicking submit, I get:
image

app/jinja_filters.py
```diff
@@ -9,26 +9,51 @@
...
+@blueprint.app_template_filter()
+def format_currency_for_input(value=None, decimal_places=0, locale=DEFAULT_LOCALE):
+    formatted_number = ''
+    if value:
+        if decimal_places is None or decimal_places == 0:
+            formatted_number = numbers.format_number(value, locale=locale)
+        else:
+            formatted_number = format_currency(value, locale=locale).replace(get_currency_symbol(locale=locale), '')
+
+    return formatted_number
```
ajmaddaford: It might be cleaner and more consistent with the `format_number` method to remove the `formatted_number` variable and return as soon as we have a formatted number
```
return numbers.format_number(value, locale=locale)
return format_currency(value, locale=locale).replace(get_currency_symbol(locale=locale), '')
return ''
```

ajmaddaford (approved):

yoganandkunche (changes requested):
Summary screen does not render the value when respondent enter a '0'

app/templates/partials/answers/currency.html
```diff
@@ -9,18 +9,25 @@
 
     <div class="input-type input-type--currency">
 
-      <span class="input-type__type" id="{{answer.id}}-type">£</span>
+      <span class="input-type__type" id="{{answer.id}}-type">{{ get_currency_symbol(answer.currency) }}</span>
+
+        {% if form.errors or form.question_errors %}
+            {% set data_input = form.get_data(answer['id'])|e %}
```
adoublebarrel: Why are we having to escape here? Ideally you shouldn't have to be worrying about it at this point.

adoublebarrel (changes requested):

app/templating/template_renderer.py
```diff
@@ -15,10 +15,12 @@ def __init__(self):
 
         self.environment.filters['format_date'] = format_date
         self.environment.filters['format_str_as_date'] = format_str_as_date
-        self.environment.filters['format_currency'] = format_currency
         self.environment.filters['format_household_name'] = format_household_member_name
         self.environment.filters['format_household_summary'] = format_household_summary
         self.environment.globals['format_conditional_date'] = format_conditional_date
+        self.environment.filters['format_currency'] = format_currency
```
adoublebarrel: How about we switch to using [Jinja filter decorators](http://flask.pocoo.org/docs/0.12/templating/#registering-filters) in app.jinja_filters?

adoublebarrel (changes requested):

tests/new_functional/spec/numbers.spec.js
```diff
@@ -32,12 +32,12 @@ describe('NumericRange', function() {
...
           .setValue(TestMinMax.testRange(), '9')
           .setValue(TestMinMax.testMin(), '0')
-          .setValue(TestMinMax.testMax(), '789')
+          .setValue(TestMinMax.testMax(), '12345')
           .setValue(TestMinMax.testPercent(), '101')
           .click(TestMinMax.submit())
           .getText('ul > li:nth-child(1) > a').should.eventually.contain('The value must be between 10 and 20. Please correct your answer.')
           .getText('ul > li:nth-child(2) > a').should.eventually.contain('The minimum value allowed is 123. Please correct your answer.')
-          .getText('ul > li:nth-child(3) > a').should.eventually.contain('The maximum value allowed is 456. Please correct your answer.')
+          .getText('ul > li:nth-child(3) > a').should.eventually.contain('The maximum value allowed is 1234. Please correct your answer.')
```
yoganandkunche: Now that we are saving the data with en_GB format with ',' I think it is worth adding having the same format in validation error messages (example: The maximum value allowed is **1,234**. Please correct your answer.)

yoganandkunche (dismissed):

app/forms/custom_fields.py
```diff
@@ -25,7 +25,7 @@ def process_formdata(self, valuelist):
 
         if valuelist:
             try:
-                self.data = int(valuelist[0])
+                self.data = int(valuelist[0].replace(',', ''))
```
dwyeradam: We should use Babel to provide us with the thousand separator for the locale (get_group_symbol()). This is also used for CustomDecimal and in validators.py

app/validation/validators.py
```diff
@@ -19,8 +19,8 @@ def __init__(self, message=None):
 
     def __call__(self, form, field):
         try:
-            Decimal(field.raw_data[0])
-        except (ValueError, TypeError, InvalidOperation):
+            Decimal(field.raw_data[0].replace(',', ''))
```
dwyeradam: We should use Babel to return the thousands separators for the locale (get_group_symbol()). In the DecimalPlaces validator we should also use babel's get_exponential_symbol to get the decimal place symbol for the locale. We should make DEFAULT_LOCALE a global variable that can be used by jinja_filters.py, custom_fields.py and validators.py.

app/templating/template_renderer.py
```diff
@@ -15,11 +15,13 @@ def __init__(self):
 
         self.environment.filters['format_date'] = format_date
```
adoublebarrel, line 16: Following this up from my last comment; I see we already have decorators for these in the jinja_filters module. Do we know why we're effectively duplicating the assignment to the environment here?

adoublebarrel (changes requested):
My questions haven't been addressed yet (see above). 

In the case of my question around the Jinja filters I recognise that we're talking about code that was present before this PR but it is [part of our workflow](https://digitaleq.atlassian.net/wiki/spaces/EQ/pages/3735569/Developer+Workflow#DeveloperWorkflow-Refactoring). (Specifically Refactoring)

app/templating/template_renderer.py
shahi645: they are in the template renderer, so that we can use them in the schema for piped currencies, or date in this instance

adoublebarrel: But they are also decorated. See [Flasks docs](http://flask.pocoo.org/docs/0.12/templating/#registering-filters). My reading of that is that we shouldn't be doing both.

adoublebarrel: Yes but they are also decorated in the jinja_filters module. See [Flasks docs](http://flask.pocoo.org/docs/0.12/templating/#registering-filters). My reading of that is they're equivalent and so we shouldn't be doing both?

adoublebarrel: Direct assignment is used as TemplateRenderer has its own Jinja environment while the decorators are used for the rest of the application.

TemplateRenderer is a bit of a misnomer as it doesn't render templates, it does variable substitution on dicts (usually dicts populated by json.load) and strings, returning the resulting dict or string.

There is a card on the Tech board to re-evaluate this approach and tackling it now is beyond the scope of this PR.

app/templates/partials/answers/currency.html
adoublebarrel: This could be a legacy code and/or there for a very good reason. To be investigated tomorrow.

shahi645: removed manual escaping from html templates

adoublebarrel (approved):
👍

yoganandkunche (approved):
looks good to me :+1:

adoublebarrel:
Time to squash the commits I think...

jonnyshaw89:
@shahi645 can you update the commit messages based on what we went through in the tech session?

ajmaddaford (commented):
The commit message title should be more descriptive and not contain hyphens, for example `Update currency unit pattern to be more flexible` would cover that we are enhancing the existing currency inputs.

Also, the body of the message doesn't mention number input types (we updated these to also allow commas)

PR #1210, opened 2017-08-29
Apply Conditional dates to RSI

jonnyshaw89:
Conditional date logic has been applied to the RSI questionnaires as per the Google Doc (https://drive.google.com/drive/folders/0BwTdPM-J8JFBa2F4RFQyTlN2bms)

### How to review 
Check that each of the RSI questionnaires matches the Google Doc


MIslam92 (approved):
Seems to match up with all surveys listed 👍

data/en/1_0102.json
```diff
@@ -82,47 +82,78 @@
...
+                            }
+                        ],
+                        "mandatory": true
+                    }],
+                    "id": "reporting-period-question-1",
+                    "title": "Are you able to report for the period from {{exercise.start_date|format_date}} to {{exercise.end_date|format_date}}?"
+                }],
+                "type": "Questionnaire",
+                "title": "Reporting period",
+                "id": "reporting-period-block-1",
```
ajmaddaford: For the new id values I think we should have a clearer prefix of `reporting-period-choice` and drop the word `block`, so the id values would be:
```
reporting-period-choice
reporting-period-choice-question
reporting-period-choice-answer
```

ajmaddaford (dismissed):

data/en/1_0102.json
```diff
@@ -82,18 +82,62 @@
                 }]
             },
             {
+                "questions": [{
+                    "description": "",
+                    "type": "General",
+                    "answers": [{
+                        "label": "",
+                        "description": "",
```
ajmaddaford: Can we remove the empty values?

ajmaddaford (changes requested):

data/en/1_0102.json
```diff
@@ -82,18 +82,62 @@
...
+                    "title": "Are you able to report for the period from {{exercise.start_date|format_date}} to {{exercise.end_date|format_date}}?"
+                }],
+                "type": "Questionnaire",
+                "title": "Reporting period",
+                "id": "reporting-period-choice",
+                "routing_rules": [{
+                        "goto": {
+                            "when": [{
+                                "value": "Yes",
+                                "id": "reporting-period-answer-1",
```
ajmaddaford: The routing is broken as this id has been changed

ajmaddaford (changes requested):

jonnyshaw89:
@ajmaddaford can you look at this again?

ajmaddaford (commented):
If you:
- Start a survey using the code from master
- Change the period dates
- Resume the survey wit the conditional dates
- Choose to keep the default dates

Then all subsequent screens still use the previously provided dates.

shahi645:
works for the rsi form types listed on the google doc 👍🏿

tests/app/templating/test_schema_context.py
```diff
@@ -1,11 +1,12 @@
...
 
 from app.data_model.answer_store import AnswerStore
+from app.questionnaire.location import Location
 from app.templating.schema_context import build_schema_context
-from tests.app.framework.survey_runner_test_case import SurveyRunnerTestCase
+from tests.app.app_context_test_case import AppContextTestCase
 
 
-class TestSchemaContext(SurveyRunnerTestCase):  # pylint: disable=too-many-public-methods
+class TestSchemaContext(AppContextTestCase):  # pylint: disable=too-many-public-methods
```
ajmaddaford, line 7: I think we need another test for the new evaluation of whether an answer is on the routing path, something like:
- Add 2 answers to the answer store with different locations
- Create a routing path that includes one answer location but not the other
- Verify the first answer in the schema context, and the second isn't

ajmaddaford (changes requested):

ajmaddaford (approved):

PR #1212, opened 2017-08-30
Update default error messages

jonnyshaw89:
Updated error messages on listed MCI/RSI form types so they match approved error message wording

### How to review 
Check that the error messages meet the Specs defined in https://trello.com/c/1sA9qHlv/1279-update-survey-default-error-messages-m


app/validation/validators.py
```diff
@@ -75,12 +75,12 @@ def __call__(self, form, field):
         error_message = None
         if data is not None:
             if self.minimum is not None and data < self.minimum:
-                error_message = self.messages.get('OUT_OF_RANGE', self.messages['NUMBER_TOO_SMALL'])
+                error_message = self.messages.get('OUT_OF_RANGE', self.messages['NUMBER_TOO_SMALL']) % dict(min=self.minimum, max=self.maximum)
```
dwyeradam: Do we really need to cover both format methods? We could remove the % dict from this line and line 80 and put it back in on line 83 where the error is raised.

dwyeradam (changes requested):

yoganandkunche:
@charlottehirst1 @sarahbrooks6 @donnapalmer3 I have done a quick review of this card on the branch and I have come across conflicting validation error messages which I think has to be finalised before signing off this card (Example: When a negative value '-2' is entered in a £'s field the spec's and confluence pages give different msgs: 1) Enter a positive number. & 2) Enter a number more than 0 (where '0' is also a valid answer).

tests/new_functional/spec/numbers.spec.js
```diff
@@ -35,10 +35,10 @@ describe('NumericRange', function() {
...
           .setValue(TestMinMax.testPercent(), '101')
           .click(TestMinMax.submit())
-          .getText('ul > li:nth-child(1) > a').should.eventually.contain('The value must be between 10 and 20. Please correct your answer.')
-          .getText('ul > li:nth-child(2) > a').should.eventually.contain('The minimum value allowed is 123. Please correct your answer.')
-          .getText('ul > li:nth-child(3) > a').should.eventually.contain('The maximum value allowed is 456. Please correct your answer.')
-          .getText('ul > li:nth-child(4) > a').should.eventually.contain('Percentages must be between 0 and 100.');
+          .getText('ul > li:nth-child(1) > a').should.eventually.contain("Enter a number more than %(min)s.".replace("%(min)s", "10"))
+          .getText('ul > li:nth-child(2) > a').should.eventually.contain("Enter a number more than %(min)s.".replace("%(min)s", "123"))
+          .getText('ul > li:nth-child(3) > a').should.eventually.contain("Enter a number less than %(max)s.".replace("%(max)s", "456"))
+          .getText('ul > li:nth-child(4) > a').should.eventually.contain("Enter a number less than %(max)s.".replace("%(max)s", "100"));
```
ajmaddaford: I don't think we need the string replacement here

tests/new_functional/spec/numbers.spec.js
```diff
@@ -52,8 +52,8 @@ describe('NumericRange', function() {
           .setValue(TestMinMax.testRange(), '12.3')
           .setValue(TestMinMax.testDecimal(), '1.234')
           .click(TestMinMax.submit())
-          .getText('ul > li:nth-child(1) > a').should.eventually.contain('Please only enter whole numbers into the field.')
-          .getText('ul > li:nth-child(2) > a').should.eventually.contain('Enter a number rounded to 2 decimal places.');
+          .getText('ul > li:nth-child(1) > a').should.eventually.contain("Enter a whole number.")
+          .getText('ul > li:nth-child(2) > a').should.eventually.contain("Enter a number rounded to %(max)d decimal places.".replace("%(max)d", "2"));
```
ajmaddaford: I don't think we need the string replacement here

ajmaddaford (changes requested):

ajmaddaford (approved):

dwyeradam (changes requested):
Two tests in test_number_range_validator.py should now fail. However they do not as they are currently not working as they should. Lines 239 and 290 are tabbed and therefore are not being evaluated but are using the no longer existing 'OUT_OF_RANGE' error message. Line 167 in test_number_validator.py is also tabbed when it shouldn't be but other than the tab removed needs no further changes.

dwyeradam (approved):

yoganandkunche:
@jonnyshaw89 error message needs updating 
Expected: 
Enter a number more than or equal to {min}.

Actual: Enter a number more than 0.

yoganandkunche (changes requested):
Percentage validation error message needs updating ass discussed.

yoganandkunche (approved):

PR #1213, opened 2017-08-31
MWSS Pay Pattern summary

jonnyshaw89:
I have added the Summary of selected Pay Patterns.
Also this PR includes the ability to route between groups

### How to review 
Check that the behaviour of the MWSS Pay Pattern summary matches the slides


adoublebarrel (approved):
I'm still not a fan of using Jinja filters like this but it works.

yoganandkunche (approved):

app/submitter/converter.py
```diff
@@ -94,6 +94,9 @@ def convert_answers_to_data(answer_store, questionnaire_json, routing_path):
 
             if answer_schema is not None and value is not None and 'parent_answer_id' not in answer_schema:
                 if answer_schema['type'] != 'Checkbox' or any('q_code' not in option for option in answer_schema['options']):
+                    if 'q_code' not in answer_schema:
```
dwyeradam: Are we happy that we could lose data if q_codes in schemas have not been defined?

app/submitter/converter.py
```diff
@@ -94,6 +94,9 @@ def convert_answers_to_data(answer_store, questionnaire_json, routing_path):
...
+                        continue
+
```
dwyeradam, line 99: Are we happy that we could lose data if q_codes in schemas have not been defined?

jonnyshaw89: Yes, given that this isn't enforced in the schema then we shouldn't enforce it in the code.

dwyeradam (approved):
Based on the fact we no longer enforce full q_code use across the schema.

ajmaddaford (changes requested):
Leaving a blocking review as we don't want this merged yet.

app/jinja_filters.py
```diff
@@ -102,14 +102,26 @@ def format_household_member_name(names):
 
 
 @blueprint.app_template_filter()
+def format_summary(list_items):
```
ajmaddaford: I think this should be `format_list` rather than `format_summary`

data/schema/schema_v1.json
```diff
@@ -162,10 +162,25 @@
...
+                "type": "string"
+              },
               "when": {
                 "$ref": "#/definitions/when"
               }
             },
+            "oneOf": [
+              {
+                "required": [
+                  "id"
```
ajmaddaford, line 175: I think we should consider renaming `id` to `block` now that we have `group`

ajmaddaford (changes requested):

data/schema/schema_v1.json
jonnyshaw89: Added Card https://trello.com/c/zAJUVc8q/335-rename-id-in-routingrules-to-block

ajmaddaford (approved):

PR #1214, opened 2017-08-31
Database.py enhancement

jonnyshaw89:
I have centralised the use if json loads and dumps to `EncryptedQuestionnaireStorage` so it is all in one place.
This makes `database.py` a cleaner representation of stored state

### How to review 
Does it still work?
Do you like the change?
Does it have any impact?


app/data_model/database.py
```diff
@@ -20,19 +20,24 @@ class QuestionnaireState(base):
     __tablename__ = 'questionnaire_state'
     user_id = Column('userid', String, primary_key=True)
     state = Column('questionnaire_data', String)
+    data = None
 
     def __init__(self, user_id, data):
         self.user_id = user_id
+        self.data = data
         self.state = json.dumps(data)
```
adoublebarrel: Now that we hold onto data I think we could populate state on access which will save a little bit of time as well. Probably justifies using an @property decorator.

adoublebarrel (changes requested):
It's hard to tell if it makes much difference as the integration tests are so slow but it is an improvement code wise.

If we change it so state is populated on demand I think it will be even better! 😛

app/data_model/database.py
jonnyshaw89: Im not sure its that easy is it when using sqlalchemy?

app/data_model/database.py
```diff
@@ -20,19 +20,24 @@ class QuestionnaireState(base):
...
 
     def set_data(self, data):
         logger.debug("setting data for questionnaire state")
-        self.state = json.dumps(data)
+        self._data = data
+        self.state = json.dumps(self._data)
 
     def get_data(self):
-        data = json.loads(self.state)
+        if not self._data:
```
collisdigital: What happens if `self.state` has changed since the last call to `get_data`?

collisdigital: i.e. can `self.state` be made private?

jonnyshaw89: state can't be private otherwise sqlalchemy doesn't persist it.

app/data_model/database.py
```diff
@@ -19,23 +19,28 @@
...
     user_id = Column('userid', String, primary_key=True)
-    state = Column('questionnaire_data', String)
+    _state = Column('questionnaire_data', String)
+    _data = None
 
     def __init__(self, user_id, data):
         self.user_id = user_id
-        self.state = json.dumps(data)
+        self._data = data
+        self._state = json.dumps(self._data)
```
adoublebarrel: This isn't Private with a capital "P", it's just a Python convention that indicates it's not a good idea to access/assign to it directly.

adoublebarrel (commented):
I don't know the internals of Sqlalchemy well enough to be sure; but if it accesses `state` via the object than having a property accessor should work.

Something like:

```
@property
    def state(self):
        """Get the current state."""
        return json.dumps(self._data)
```

and switch the table definition back to using `state` instead of `_state`.

jonnyshaw89:
@adoublebarrel yeh i tried that but the test failed with it not persisting the data.

adoublebarrel:
Oh, Boo. Never mind then.

adoublebarrel (dismissed):

ajmaddaford (changes requested):
I don't think we should be adding a `data` property to the `QuestionnaireState` class, I think this should be a clean mapping of the database table to a Python class. 

Looking at the code in `EncryptedQuestionnaireStorage` I think we can move the JSON serialisation bits there, so all interactions with the database object work by getting/setting the state property, and dropping the set_data and get_data methods from `QuestionnaireState`.

app/storage/encrypted_questionnaire_storage.py
```diff
@@ -26,11 +27,12 @@ def __init__(self, database, user_id, user_ik, pepper):
         self._cek = self._generate_key(user_id, user_ik, pepper)
 
     def add_or_update(self, data):
-        encrypted_data = self._encrypt_data(data)
+        serialised_data = simplejson.dumps(data, default=lambda o: o.__dict__)
```
adoublebarrel: It's generally bad form to access an objects dunder (`__xx__`) methods and in this specific case I'm not sure it's a good idea. `__dict__` doesn't always exist for an object and it doesn't take account of an objects inheritance path so attributes that you'd expect to be serialised might not be.

Am I right that this is to cope with serialising Location objects? If it is then it might be preferable to define a public interface for getting a dict version of it *or* making it dict-like by inheriting from dict directly or [one of the collections abstract base classes](https://docs.python.org/3/library/collections.abc.html#collections.abc.Mapping).

app/storage/encrypted_questionnaire_storage.py
```diff
@@ -1,4 +1,5 @@
 import hashlib
+import simplejson
```
adoublebarrel: Not directly related to this review as we started doing this elsewhere *but*; I read [the package documentation](https://pypi.python.org/pypi/simplejson/) and I *think* `simplejson` should be the same as the `json` package that's part of the standard library.

Begs the question as to why we were seeing different behaviour with decimals (I think) but probably worth checking before we go and embed simplejson as a dependency everywhere.

adoublebarrel (changes requested):

adoublebarrel (approved):
👍

app/data_model/questionnaire_store.py
```diff
@@ -1,56 +1,50 @@
...
-        self.answer_store.answers = data.get('ANSWERS') or []
-        self.completed_blocks = data.get('COMPLETED_BLOCKS') or []
-
-    def _get_data(self):
-        return {
+            self._deserialise(raw_data)
+
+    def _deserialise(self, data):
+        json_data = json.loads(data)
+        if isinstance(json_data, dict):
```
collisdigital: why would this be false and if it is does it indicate an error?

ajmaddaford: Without this pylint throws a `maybe-no-member` error (see https://stackoverflow.com/questions/21545409/pylint-reports-maybe-no-member-error-for-data-loaded-via-simplejson) as there are a few edge cases where the deserialised data may not be a dict e.g. the string "true" would be loaded as `True`. In might be clearer to just disable the rule here.

collisdigital: if the `if` stays should there be an `else` to log the error (if it is one?) or throw?

app/data_model/database.py
```diff
@@ -21,21 +20,12 @@ class QuestionnaireState(base):
...
-
-    def get_data(self):
-        data = json.loads(self.state)
-        logger.debug("loading questionnaire state")
-        return data
+        self.state = state
 
     def __repr__(self):
-        return "<QuestionnaireState('%s','%s')>" % (self.user_id, self.state)
+        return "<QuestionnaireState('%s','%s')>" % (self.user_id, self._state)
```
collisdigital, line 28: Do we have unit tests that check this function works when calling `repr()` ?

app/data_model/questionnaire_store.py
```diff
@@ -1,56 +1,50 @@
...
 
     def add_or_update(self):
-        data = self._get_data()
-        serialised_data = self._serialise(data)
-        self._storage.add_or_update(data=serialised_data)
+        data = self._serialise()
+        self._storage.add_or_update(data=data)
+
+    def _encode_questionnaire_store(self, o):
+        if hasattr(o, 'to_dict'):
```
collisdigital, line 46: should there be an `else` to this that throws or logs?

collisdigital: oh I see, please ignore

app/data_model/database.py
jonnyshaw89: no, https://codecov.io/gh/ONSdigital/eq-survey-runner/src/master/app/data_model/database.py

collisdigital: Should add some then...

collisdigital:
Please test this by running from master, partially completing and then switching to this branch to ensure no data loss/corruption.

collisdigital (approved):

jonnyshaw89:
@ajmaddaford Just needs a tick from you now.

ajmaddaford (approved):

app/data_model/database.py
collisdigital: Looking at it, this code will actually fail as the variable is named incorrectly _state

PR #1215, opened 2017-09-01
Convert all the data types into a string for data version 0.0.1

yoganandkunche:
Converts all data types entered into a string and pushes it to SDX

### How to review 
Launch any survey with dumper
enter different data type values into fields
and verify the dump

Expected: All the values entered must be displayed as strings on the dump

### DIFF
--- a/app/submitter/converter.py
+++ b/app/submitter/converter.py
@@ -194,12 +194,13 @@ def _build_metadata(metadata):
 
 
 def _encode_value(value):
-    if isinstance(value, int):
-        return str(value)
-    elif isinstance(value, str) and value == '':
-        return None
+    if isinstance(value, str):
+        if value == '':
+            return None
+        else:
+            return value
     else:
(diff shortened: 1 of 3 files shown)


ajmaddaford (approved):

dwyeradam (changes requested):
This only works for schemas with "data_version": "0.0.1". I'm not sure if this was the intention but if we start to use data version 2 at some point then the conversion will not take place.

ajmaddaford:
@dwyeradam That was the intention. The stringifcation of values is only required for surveys that use Q codes.

dwyeradam (approved):

ajmaddaford (changes requested):
The commit message needs changing, it's currently `resolved merge conflicts`.

The full commit message should also include why this is being changed.

ajmaddaford (approved):

PR #1221, opened 2017-09-06
Added population of path finder in g

adoublebarrel:
Before this the helper always returned a new path finder object which
made most of the code in the helper pointless.

### How to review 
All functional behaviour should be the same as before. If you set a breakpoint within the `if` you should find that you only hit it once per request to the questionnaire blueprint.

### DIFF
--- a/tests/app/helpers/test_path_finder_helper.py
+++ b/tests/app/helpers/test_path_finder_helper.py
@@ -1,0 +1,21 @@
+from unittest.mock import patch
+from flask import g
+from tests.app.app_context_test_case import AppContextTestCase
+
+from app.helpers.path_finder_helper import path_finder
+
+
+@patch('app.helpers.path_finder_helper.PathFinder')
+@patch('app.helpers.path_finder_helper.get_metadata')
+@patch('app.helpers.path_finder_helper.get_answer_store')
+class TestPathFinderHelper(AppContextTestCase):
+    def test_path_finder_instantiated_once(self, mock_path_finder, _, __):
+        g.schema_json = {}
+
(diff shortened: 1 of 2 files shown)


jonnyshaw89:
@adoublebarrel should this have a test?

tests/app/helpers/test_path_finder_helper.py
```diff
@@ -0,0 +1,11 @@
...
+from tests.app.app_context_test_case import AppContextTestCase
+
+from app.helpers.path_finder_helper import path_finder
+
+@patch('app.helpers.path_finder_helper.get_metadata')
+@patch('app.helpers.path_finder_helper.get_answer_store')
+@patch('app.helpers.path_finder_helper.PathFinder')
+class TestPathFinderHelper(AppContextTestCase):
+    def test_path_finder_is_a_PathFinder(self, mock_metadata, mock_answer_store, mock_path_finder):
+        self.assertEqual('', mock_path_finder)
```
jonnyshaw89: This isn't very obvious?
looks like its asserting a mock is an empty string?

jonnyshaw89: We should also test that calling it a 2nd time returns the same value and doesn't generate a new one?

jonnyshaw89 (changes requested):

adoublebarrel:
Sorry, I should of marked it as "in progress". This was a broken push so that I had everything tee'd up for being back in the office.

jonnyshaw89 (approved):

ajmaddaford (approved):

PR #1222, opened 2017-09-06
Only re-generate routing_path on changes to the answer_store

jonnyshaw89:
The routing path was being generated about 7 times per request.
This PR means that on a GET the routing_path is generated once and on a POST it is generated twice.

### How to review 
Do you like the implementation?
Does it work?


app/questionnaire/path_finder.py
```diff
@@ -91,19 +94,52 @@ def _remove_rule_answers(self, goto_rule):
...
 
         return self.build_path(self.get_blocks(), location)
 
+    def get_full_routing_path(self):
+        """
+        Returns a list of the block ids visited based on answers provided
+        :return: List of block location dicts
+        """
+
+        latest_answer_store_hash = hash(simplejson.dumps(self.answer_store, default=lambda o: o.__dict__))
```
collisdigital: Bit of duplication here with line 20, consider factoring out?

app/questionnaire/path_finder.py
```diff
@@ -91,19 +94,50 @@ def _remove_rule_answers(self, goto_rule):
...
 
-    def get_routing_path(self, group_id=None, group_instance=0):
+    def get_routing_path(self, group_id, group_instance=0):
         """
         Returns a list of the block ids visited based on answers provided
         :return: List of block location dicts
         """
-        if group_id is None:
-            group_id = SchemaHelper.get_first_group_id(self.survey_json)
+        self.get_full_routing_path()
```
collisdigital, line 105: feels a bit strange using it this way, would be clearer if the return from this call was assigned to a variable used later on?

app/questionnaire/path_finder.py
```diff
@@ -91,19 +94,50 @@ def _remove_rule_answers(self, goto_rule):
...
+        routing_path = None
+
+        for i, location in enumerate(self.full_routing_path):
+            if location.group_id == group_id and location.group_instance == group_instance:
+                routing_path = self.full_routing_path[i:]
+                break
+
+        if routing_path:
+            return routing_path
+
```
collisdigital: use `else:` for the rest?

app/questionnaire/path_finder.py
jonnyshaw89: it essentially is `self.full_routing_path`

app/questionnaire/path_finder.py
jonnyshaw89: Im not a fan of an `else` it generally indicates that the method does too much

collisdigital: 😜 may I suggest this method might be doing too much then

jonnyshaw89: not any more.

collisdigital: 👍

shahi645 (approved):
Routing works as before for me locally. All tests pass fine also 👍🏿

app/questionnaire/path_finder.py
```diff
@@ -208,3 +234,7 @@ def _relationship_previous_location(current_group_instance):
         else:
             previous_location = Location('who-lives-here-relationship', current_group_instance - 1, 'household-relationships')
         return previous_location
+
+    @staticmethod
+    def _generate_hash_of_answer_store(answer_store):
+        return hash(simplejson.dumps(answer_store, default=lambda o: o.__dict__))
```
ajmaddaford: I think we should consider adding a `get_hash` method on the answer_store, and return the hash of answers rather than a hash of the AnswerStore object, something like:
```
def get_hash(self):
     return hash(self.answers)
```
We could then remove this method from here and call `answer_store.get_hash()` instead

app/questionnaire/path_finder.py
```diff
@@ -1,5 +1,6 @@
 import copy
 
+import simplejson
```
jcox-dev: To stay consistent (with, for example, #1214), should this be changed to `import simplejson as json`?

app/data_model/answer_store.py
```diff
@@ -264,6 +265,9 @@ def remove_by_location(self, location):
 
         self.remove(group_id=location.group_id, block_id=location.block_id, group_instance=location.group_instance)
 
+    def get_hash(self):
+        return hash(json.dumps(self.answers))
```
collisdigital: Not sure if this is really something to be concerned with here, but I don't believe that json.dumps is guaranteed to return the same JSON string for the same Python dict each time (e.g. key orders etc.) even if logically the dict is identical. Relying on this to check if something has changed could lead to false-positives where the same dict will look like it has changed because the hash of the raw string will be different even if the actual data isn't. 

Again, we may not care?

app/questionnaire/path_finder.py
collisdigital: maybe it should be called `generate_full_routing_path` or something since `get` implies this function returns a value rather than changing the state of the object?

app/views/questionnaire.py
```diff
@@ -171,12 +174,12 @@ def get_thank_you(eq_id, form_type, collection_id):  # pylint: disable=unused-ar
                                                    survey_title=TemplateRenderer.safe_content(schema['title']))
         return thank_you_template
     else:
-        return _redirect_to_latest_location(collection_id, eq_id, form_type)
+        routing_path = path_finder.get_full_routing_path()
```
collisdigital, line 181: why not use the function decorator here?

app/helpers/path_finder_helper.py
```diff
@@ -19,3 +21,11 @@ def get_path_finder():
 
 
 path_finder = LocalProxy(get_path_finder)
+
+
+def has_path_finder_route(function):
```
collisdigital: I wonder if this name is best; looking at the usage it feels more like `requires_full_routing_path`

app/views/questionnaire.py
jcox-dev: @collisdigital could do but the fully calculated routing path is only required if the survey is not complete (for the redirect), not for rendering the thank you template.

app/questionnaire/path_finder.py
jcox-dev: `generate_full_routing_path` would be clearer that it's performing a more complex task but the [`build_path`](https://github.com/ONSdigital/eq-survey-runner/blob/261a2de597db5ec44af8ec6b2426c22dd1cabdbc/app/questionnaire/path_finder.py#L28) method ([called](https://github.com/ONSdigital/eq-survey-runner/blob/261a2de597db5ec44af8ec6b2426c22dd1cabdbc/app/questionnaire/path_finder.py#L132) from `get_full_routing_path`) is doing the real heavy lifting and `get_full_routing_path` does then fetch it from the result of that.

app/helpers/path_finder_helper.py
```diff
@@ -19,3 +21,11 @@ def get_path_finder():
 
 
 path_finder = LocalProxy(get_path_finder)
+
+
+def requires_full_routing_path(function):
```
collisdigital: Sorry to be a pain but I now realise that the convention seems to be to put `required` at the end e.g. `login_required` and `role_required` so this should probably be `full_routing_path_required`?

jcox-dev: @collisdigital good spot.

app/views/questionnaire.py
collisdigital: ok

app/data_model/answer_store.py
collisdigital: @ajmaddaford @jonnyshaw89 @jcox-dev  thoughts?

jonnyshaw89: Does it work?

jcox-dev: It does but we could ensure it's accurate each time with the `sort_keys` flag if we think that's necessary:

```python
    def get_hash(self):
        return hash(json.dumps(self.answers, sort_keys=True))
```

collisdigital: I'd query the performance hit of this kind of thing, how often will this code run?

jcox-dev: Shouldn't take more than about 0.1ms (0.15ms with key sorting) to calculate the hash (on a large, e.g: census, answer list).

app/questionnaire/path_finder.py
```diff
@@ -16,6 +15,8 @@ def __init__(self, survey_json, answer_store=None, metadata=None):
         self.answer_store = answer_store or AnswerStore()
         self.metadata = metadata or {}
         self.survey_json = survey_json
+        self.answer_store_hash = self.answer_store.get_hash()
```
collisdigital: should these be `_` members?

collisdigital (approved):

app/questionnaire/path_finder.py
jcox-dev: @collisdigital I think you're right, `.answer_store_hash` and `.full_routing_path` should be as they are only to be used and calculated by the instance's methods.

PR #1231, opened 2017-09-07
Maximum value for number field not accepting zero

### LINKED ISSUE #1228: min and max value run time validation not as expected
### Expected behaviour
Must fail as the min value is greater than max value

### Actual behaviour
Accepts even if the min value is greater than max value

Launch test_numbers
enter a min value as 10 and max value as '0'

dwyeradam:
Defaulting to MAX_NUMBER as OR recognises int 0 as false.
Fixes #1228

### DIFF
--- a/tests/app/validation/test_number_range_validator.py
+++ b/tests/app/validation/test_number_range_validator.py
@@ -197,6 +197,79 @@ class TestNumberRangeValidator(unittest.TestCase):
         except ValidationError:
             self.fail("Valid integer raised ValidationError")
 
+
+    def test_zero_max(self):
+        max_value = 0
+        answer = {
+            'max_value': {
+                "value": max_value
+            },
+            'label': 'Max Test',
+            'mandatory': False,
+            'id': 'test-range',
+            'type': 'Currency'
+        }
+        label = answer['label']
+        error_message = error_messages['NUMBER_TOO_LARGE'] % dict(max=max_value)
+
+        integer_field = get_number_field(answer, label, '', error_messages, self.store)
+
+        self.assertTrue(integer_field.field_class == CustomIntegerField)
+
+        for validator in integer_field.kwargs['validators']:
+            if isinstance(validator, NumberRange):
+                test_validator = validator
+
+        mock_form = Mock()
+        integer_field.data = 1
+
+        with self.assertRaises(ValidationError) as ite:
+            test_validator(mock_form, integer_field)
+
+        self.assertEqual(str(ite.exception), error_message)
+
+        try:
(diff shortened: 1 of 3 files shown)


tests/app/validation/test_number_range_validator.py
```diff
@@ -195,6 +195,51 @@ def test_manual_max(self):
...
+
+    def test_zero_max(self):
+        answer = { 'max_value': {
+                        "value": 0
+                    },
+                   'label': 'Max Test',
+                   'guidance': '',
+                   'options': [],
+                   'mandatory': False,
+                   'validation':
```
jonnyshaw89: You don't really need to define this here, it just bloats the tests when you could just include `error_messages.py`

tests/new_functional/spec/numbers.spec.js
```diff
@@ -57,4 +57,15 @@ describe('NumericRange', function() {
...
   });
 
+  it('Given a range is set when the maximum is less than the minimum then a run time error should occur', function() {
+    return helpers.openQuestionnaire(number_schema)
+      .then(() => {
+        return browser
+          .setValue(SetMinMax.setMinimum(), '10')
+          .setValue(SetMinMax.setMaximum(), '0')
+          .click(SetMinMax.submit())
+          .getText('#main > h1').should.eventually.contain('Error 500');
```
jonnyshaw89: Do we want to be testing this sort of things as a functional test?
Seems to me this is more the sort of thing that can be tested with a unit or integration test.

jonnyshaw89 (dismissed):

tests/new_functional/spec/numbers.spec.js
yoganandkunche: @jonnyshaw89 Removed the functional test as we already have a unit test test_number_range_validator

tests/app/validation/test_number_range_validator.py
yoganandkunche: Done

shahi645 (approved):

tests/app/validation/test_number_range_validator.py
```diff
@@ -197,6 +197,44 @@ def test_manual_max(self):
         except ValidationError:
             self.fail("Valid integer raised ValidationError")
 
+
+    def test_zero_max(self):
```
ajmaddaford, line 201: I think we should also have a test for a min_value of `0` defined in the schema

ajmaddaford (dismissed):

jcox-dev (approved):
Tests for both changes are in. Looks good to me.

PR #1232, opened 2017-09-08
Migrate Functional Tests

dwyeradam:
It's months since the re-work of the functional tests was done and we still have a split between the "new" and "old" tests. Having both is a source of confusion/friction. This PR moves tests from the "old" functional tests to the "new", deleting and refactoring as appropriate.


data/en/test_error_messages.json
```diff
@@ -0,0 +1,72 @@
+{
+    "mime_type": "application/json/ons/eq",
+    "schema_version": "0.0.1",
+    "data_version": "0.0.2",
+    "survey_id": "023",
+    "title": "Test Error Messages",
+    "description": "Test Description",
+    "theme": "default",
+    "legal_basis": "StatisticsOfTradeAct",
+    "messages": {
```
jonnyshaw89: We should lean on the defaults unless we really need to override

data/en/test_error_messages.json
```diff
@@ -0,0 +1,72 @@
...
+                            "id": "test-percent",
+                            "description": "",
+                            "label": "Percent Test (0 to 100)",
+                            "mandatory": false,
+                            "options": [],
+                            "type": "Percentage",
+                            "max_value": {
+                                "value": 100
+                            },
+                            "validation": {
```
jonnyshaw89: same comment about using the defaults

jonnyshaw89 (dismissed):

ajmaddaford (approved):

shahi645 (approved):
All functional tests work locally for, was getting an error with 'in-eduction spec' running as part of the test suite, but passes individually.

PR #1233, opened 2017-09-08
Updated selenium-standalone package to work with chrome 61

adoublebarrel:
Travis builds started failing due to the release of Chrome 61 which we use in headless mode for the new_functional tests. This is a trial of the latest version of selenium standalone which should resolve this issue.

### How to review 
Builds should pass.

### DIFF
--- a/tests/new_functional/config.js
+++ b/tests/new_functional/config.js
@@ -44,14 +44,6 @@ let config = {
   }
 };
 
-const sauceLabsConfig = {
-  services: ['sauce'],
-  sauceConnect: true,
-  user: process.env.SAUCE_USERNAME,
-  key: process.env.SAUCE_ACCESS_KEY,
-  capabilities: [capabilities.firefox]
-};
-
 const browserStackConfig = {
   services: ['browserstack'],
   user: process.env.BROWSERSTACK_USER,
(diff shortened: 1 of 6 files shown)


jonnyshaw89 (approved):

tests/new_functional/config.js
```diff
@@ -44,48 +44,47 @@ let config = {
...
+      chaiAsPromised.transferPromiseness = browser.transferPromiseness;
+      browser.setViewportSize({
+        width: 1280,
+        height: 1024
+      });
+    }
   }
-
-};
+);
```
samiwel, line 73: What was the reason for having to extend the Phantom config. Was this required to get the chrome headless working or is it an unrelated change? Sorry just being nosey on a Runner PR since it would be nice to get the headless chrome running on Author as well :-)

adoublebarrel: Unrelated. I almost changed back to running Phantom to unblock us and realised that I'd broken the configuration in the move to Chrome headless. So this is really re-instating some of that config so we can easily switch back to Phantom if we need to.

Phantom doesn't cope with our new multiple surveys test though so we'd have to disable them if we did switch back.

yoganandkunche (approved):

PR #1234, opened 2017-09-12
Inline feedback form

adoublebarrel:
[Trello Card](https://trello.com/c/loROoVWw/1325-in-service-feedback-inline-js-s)

### How to review 
The feedback link now opens the inline form when JS is enabled, the non-JS experience still opens the standalone form in a new tab.

A new JS module has been added and updates made to the feedback scss.


yoganandkunche:
@adoublebarrel looks like there is an issue. error 500 on every scheme. 
TemplateNotFound(template)

jonnyshaw89:
@adoublebarrel did you forget to commit a file ?
`jinja2.exceptions.TemplateNotFound: partials/feedback/expandable_inline.html`

app/templates/layouts/_base.html
```diff
@@ -115,5 +115,8 @@
       <script src="{{ url_for('static', filename='js/bundle.js') }}"></script>
     {% endblock scripts %}
   	  <!--<![endif]-->
+      <script id="__bs_script__">//<![CDATA[
```
jonnyshaw89: Don't forget to take this out again.

jonnyshaw89:
I think there should be a way to close the feedback form if i change my mind?

jonnyshaw89:
After posting the feedback form i get a 302 and the browser performs a GET request in the background to `http://localhost:5000/feedback/thank-you` for this form i think it should just return a 200 status code

jonnyshaw89 (changes requested):
Comments above

yoganandkunche (dismissed):
The top previous link stops working after submitting a feedback.

adoublebarrel:
@jonnyshaw89 
The XHR request no longer gets redirected.

I agree with your comment re. being able to collapse the form but having implemented this now I think the general UX should be re-thought ("modal" overlay rather than expanding content FWIW).

We need to get better with engaging with these designs earlier before they get to this point in the process. As it stands the current behaviour is to the agreed spec when this card was accepted for development.

jonnyshaw89:
@adoublebarrel I think we should log the POST request out as we do with `questionnaire request`

jonnyshaw89 (approved):

jcox-dev:
Echoing what @jonnyshaw89 was saying about being able to close the inline feedback form, perhaps we can keep the "beta" line, instead of hiding it, and clicking the "Help us improve it" with the form visible hides it again.

jcox-dev:
Missed the import for the helpers in the inline template.

shahi645 (approved):
New changes look good to me an work fine

charlotte-hirst:
works fine in chrome but cant submit in Safari. Havent got IE to test it with

app/assets/js/app/modules/inline-feedback.js
```diff
@@ -0,0 +1,124 @@
...
+      e.preventDefault()
+
+      this.collapse(this.form)
+    })
+  }
+
+  sendFeedback() {
+    const form = this.form.getElementsByTagName('form')[0]
+    const formData = new FormData(form)
+    formData.set('redirect', 'false')
```
jcox-dev: The [set interface](https://developer.mozilla.org/en-US/docs/Web/API/FormData/set) for FormData is not implemented across all browsers. Need to polyfill or handle redirect in another way other than manipulating the form.

jcox-dev (changes requested):

armstrongb:
### The textarea focus state is displaying incorrectly in Chrome.

Both edges are cropped by 1px each when focussed. Both edges should display the same thickness as the top and bottom edges. If the text area is dragged inwards from the right-hand edge the `box-shadow` shows fully on the rh side.

### When at default width:
ons-screenshot-feedback-textarea-focus

### When dragged inwards:
ons-screenshot-feedback-textarea-focus-infromright

armstrongb:
This also happens to the input fields (lefthand `box-shadow`):

ons-screenshot-feedback-input-focus

armstrongb:
I think that the feedback 'Cancel' link should have `title="Cancel sending your feedback"` added to the link so that its purpose is stated. Alternatively the 'sending your feedback' could be: `<span class="u-vh"> sending your feedback</span>`

@hamishtaplin which is our preferred option for this (or both)?

armstrongb (changes requested):
Can the visual defects and 'Cancel' link attributes be resolved before proceeding please?

jcox-dev (approved):

jcox-dev:
Spotted a visible gap (between the feedback CTA and header) introduced with these changes but currently unable to determine how. 

ons-screenshot-feedback-gap

@armstrongb mentioned an old "previous" link might be the culprit.

ons-screenshot-previous-link

bennpearson:
`overflow: hidden` just needs removing from `.feedback__inline` to fix the border issue.

app/assets/js/app/modules/inline-feedback.js
```diff
@@ -0,0 +1,127 @@
+import domready from './domready'
+
+export const classCallToAction = 'feedback__action'
```
hamishtaplin: Don't use CSS classes as selectors (as changing them will break the JS)

Instead add `js-` prefixed hooks.

app/assets/js/app/modules/inline-feedback.js
```diff
@@ -0,0 +1,127 @@
...
+    element.classList.add(classHidden)
+  }
+
+  show(element) {
+    element.classList.remove(classHidden)
+  }
+}
+
+export default function feedback() {
+  const elFeedback = document.getElementsByClassName(classInlineForm)[0]
```
hamishtaplin: use `document.querySelector` which will return the first matched element.

app/assets/styles/partials/components/_feedback.scss
```diff
@@ -1,3 +1,23 @@
+.feedback__inline {
+  margin-bottom: 0 !important;
```
hamishtaplin: Don't use `!important`. We should fix the [greedy selector](https://github.com/ONSdigital/eq-survey-runner/blame/eq-1325-inline-feedback-js/app/assets/styles/partials/objects/_page.scss#L27) that's making this necessary.

app/assets/styles/partials/components/_feedback.scss
```diff
@@ -1,3 +1,23 @@
...
+  overflow: hidden;
+  width: 100%;
+  opacity: 0;
+  transition: max-height 500ms ease-out, opacity 500ms ease-out, margin-top 500ms ease-out;
+  max-height: 0;
+  &.is-expanded {
+    margin-top: 1rem !important;
```
hamishtaplin: Again with `!important` 😲

This transition looks weird because of the `margin-top`—better to handle the whitespace with an internal `div` with padding than override margins and then attempt to transition them.

app/templates/partials/feedback/expandable_inline.html
```diff
@@ -0,0 +1,8 @@
+{% import 'macros/helpers.html' as helpers %}
+{% include theme('partials/feedback_call_to_action.html') %}
+<div class="feedback__inline">
+    <div class="feedback__border">
+        {% include theme('partials/feedback_form.html') %}
+        <p class="mars feedback__cancel"><a href="" data-qa="a-feedback-close" {{helpers.track('click', 'Feedback', 'Cancel feedback link click')}}>Cancel</a></p>
```
hamishtaplin: Why is this a `p`? Don't use semantic tags for layout—it's not a paragraph of text :)

The `a` has no `href`—use a `button`. I guess it is a link because it's designed to look like one, but it isn't actually a link because it doesn't take you to a new page (or a page fragment). It should look like and _be_ a `button`.

app/templates/partials/feedback_form.html
```diff
@@ -1,12 +1,12 @@
-<form method="POST" class="feeback__form {{ 'u-hidden' if feedback_form_hidden }}">
+<form id="feedback-form" method="POST" class="feedback__form {{ 'u-hidden' if feedback_form_hidden }}">
     {{ form.csrf_token }}
     <div class="group">
         <div class="feedback__field">
             <label id="feedback-message-label" class="label venus" for="feedback-message">How can we improve this survey?</label>
             {% if feedback_form_hidden %}
-            <a class="feeback__close" href="#close">Close</a>
+            <a class="feedback__close" href="#close">Close</a>
```
hamishtaplin: Not sure how to test this but it's a link to a page fragment (`#close`) that doesn't seem to exist anywhere?

tests/new_functional/spec/textarea.spec.js
```diff
@@ -5,6 +5,7 @@ const TextareaSummary = require('../pages/surveys/textarea/textarea-summary.page
 describe('Textarea', function() {
 
   const textarea_schema = 'test_textarea.json';
+  const textarea_limit = '[data-qa="textarea-with-limit"]';
```
hamishtaplin, line 8: Much better than using CSS classes 👍

app/assets/js/app/modules/inline-feedback.js
```diff
@@ -0,0 +1,127 @@
...
+
+      this.cancel = cancel
+      this.registerCancel(this.cancel)
+    }
+  }
+
+  registerForm(element) {
+    element.setAttribute(attrHidden, 'true')
+
+    let elSubmit = element.getElementsByClassName(classSubmit)[0]
```
hamishtaplin: This never gets reassigned so should be `const`.

app/assets/js/app/modules/inline-feedback.js
```diff
@@ -0,0 +1,127 @@
...
+  registerCancel(element) {
+    element.addEventListener('click', e => {
+      e.preventDefault()
+
+      this.collapse(this.form)
+    })
+  }
+
+  sendFeedback() {
+    const form = this.form.getElementsByTagName('form')[0]
```
hamishtaplin: `this.form.querySelector("form")`

hamishtaplin (changes requested):

hamishtaplin:
screenshot 2017-09-22 10 00 50

Looks like this now 😲

hamishtaplin:
@armstrongb I don't think it should be a link in the first place? It doesn't go anywhere—it should be a button. Using a `title` just gets you a tooltip, [screenreaders don't read them](https://silktide.com/i-thought-title-text-improved-accessibility-i-was-wrong/).

hamishtaplin:
I don't know if this is intentional, but putting the form on the top of the page but hiding it without `display: none` means it's still in the tab index—meaning that tabbing through the page takes you through the "hidden" form. Confusing for anyone who uses a keyboard but also bad for accessibility as users of things like Dragon will receive announcements for elements they can't see.

Personally, I would have tested this with DAC _before_ implementing it.

hamishtaplin:
@jcox-dev click the "Help us improve it" link again feels a bit weird for closing the form? Maybe toggle it to "Don't help us improve it"?  

We already have a "cancel" link/button—isn't this a bit redundant anyway?

I agree with what @adoublebarrel said—this might have been better as a modal as it's pretty large and it being in the DOM permanently, on every page, especially at the top could be problematic.

app/assets/styles/partials/components/_feedback.scss
armstrongb: @hamishtaplin Is there not a `u-mb-0` utility class too?

hamishtaplin: @armstrongb yeah that would also work 👍

jcox-dev:
@armstrongb @hamishtaplin changes pushed.

armstrongb:
The 'Previous' link is too close to the border of the feedback form.
`.page__subheader .container>*` bottom margin in `_page.scss` is preventing the addition of a bottom margin (e.g. `.u-mb-l`)
image

armstrongb:
Can we remove the `.btn--lg` from the 'Send feedback' button as the type is untidily stepped at present. The resolved layout is shown below:

image

armstrongb:
I don't think that `.feedback__border` needs the padding-top applied to it.

jcox-dev:
@armstrongb amendments made.

hamishtaplin (approved):

jcox-dev:
There was a discussion today which covered:

- moving the link to the bottom of page (above footer), similar to other .gov sites
- whether "Cancel" should be a link or a button (split on agreement)
- display: none should be default for the form (even without JS) to avoid tabbing through hidden form fields and issues with screenreader
- moving implementation to modal rather than expandable inline form

@hamishtaplin @armstrongb @bennpearson

armstrongb:
@jcox-dev apologies - the button should have been `.btn--link` so still a button but one styled as a link.

If we are to proceed with this version I think that we should change 
> Cancel

to

> I don't want to provide feedback

which would also not require any visually hidden content.

armstrongb:
Can we amend the 
> I don't want to provide feedback

to 

> I don’t want to provide feedback
(change = use of apostrophe not straight single quote - use `alt+shift+]` on Mac)

,also:
> Please include your name and email address if you're interested in taking part in further research to improve this survey

change
> ... you're ... 
to
> ... you’re ...

?

Thanks.

jcox-dev:
@armstrongb amendments pushed. 👍

armstrongb (approved):

PR #1218, opened 2017-09-05
Eq 1319 cookie and contact us pages

MIslam92:
### What is the context of this PR?
Addition of new static pages to survey runner i.e contact us page & cookies and privacy page

### How to review 
run ./scripts/run_app.sh and docker then navigate to url and append /contact-us or /cookies-privacy e.g. localhost:5000/contact-us to check that pages match design.
Yarn compile/start may be needed to run for styles.

PR #1235, opened 2017-09-13
New Table component to be added to eq-survey-runner

MIslam92:
New component being added from the global design patterns

### How to review 
Just check in context to PR https://github.com/ONSdigital/eq-survey-runner/pull/1218
If the tables are displayed correctly in the cookie & privacy pages

### DIFF
--- a/app/assets/styles/partials/components/_table.scss
+++ b/app/assets/styles/partials/components/_table.scss
@@ -1,0 +1,32 @@
+.table {
+
+}
+caption {
+  text-align: left;
+}
+.table__dense {
+  font-size: 81.25%;
+
+.table--header--cell {
+    padding: 0.5rem 0 0.25rem 1rem;
+    @include nth-element(1, 0);
+  }
+}
+
+.table--row {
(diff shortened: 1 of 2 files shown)


shahi645 (approved):
Works on my machine locally, code follows standards of survey runner... LGTM 👍🏿

app/assets/styles/partials/components/_table.scss
```diff
@@ -0,0 +1,32 @@
...
+.table--header--cell {
+  padding: 0.5rem 0 0.5rem 1rem;
+  border-bottom: 2px solid $color-very-dark-grey;
+  @include nth-element(1, 0);
+}
+.table--cell__numeric {
+  text-align: right;
+}
+
+.table--cell {
```
adoublebarrel, line 29: The designs have a smaller font size for the table cells. You can see it applied in the prototype where it sets it to 12pt. That's probably not how we want to implement it in runners CSS but we certainly want the same behaviour.

adoublebarrel (changes requested):

PR #1208, opened 2017-08-24
Adds preview questions and new content on rsi/mci

yoganandkunche:
### What is the context of this PR?
his PR is to added preview content to MCI and RSI Survey's

### How to review 
Launch all the Business surveys 0102,0112,0203,0205,0213 & 2015 and also 0005, 1_0001 & 2_0002 and verify that the landing page match the content and preview structure defined in the below doc:

https://docs.google.com/document/d/10v30YjBp9wX5wYJURSnPAZZ4BkbVLfCBcj414ri5OPQ/edit#heading=h.fcltikn7ehz4

PR #1240, opened 2017-09-14
Removing conditional dates and revert content from RSI and MCI surveys

MIslam92:
- All RSI and MCI types are reverted to their previous versions

- All survey launch pages for RSI and MCI types are reverted to their previous versions

See also: #1208, #1210

### How to review 
Launch all RSI & MCI surveys and compare that they all match previous version
Surveys:
0102
0112
0203
0205
0213
0215

### DIFF
--- a/data/en/1_0102.json
+++ b/data/en/1_0102.json
@@ -12,115 +12,27 @@
                 "type": "Introduction",
                 "id": "introduction",
                 "primary_content": [{
-                    "id": "get-started",
-                    "title": "Get started",
+                    "id": "use-of-information",
                     "content": [{
-                            "description": "On average it takes 10 minutes to complete this survey once you’ve collected the information."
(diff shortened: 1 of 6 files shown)


charlotte-hirst (approved):
Checked all surveys are working but does need someone like @yoganandkunche to carefully check them side by side as we cant take any chances with RSI

yoganandkunche (approved):
@charlotte-hirst Looks good to me. I will be review this again in pre-prod and verify the .pck files which go downstreams.

yoganandkunche:
@jonnyshaw89 could you please review it code wise.

jonnyshaw89 (changes requested):
It looks like some of the blocks have `id` defined twice. if you can remove the duplicate then I'm happy

jonnyshaw89 (approved):

PR #1243, opened 2017-09-15
Enable no-sandbox on chrome headless

jonnyshaw89:
This allows the Functional Tests to run correctly in Concourse

`--no-sandbox` is required for chrome headless to run properly inside a docker container
Headless chrome does not run properly in docker when running in a sandbox https://github.com/SeleniumHQ/docker-selenium/issues/520

### How to review 
Test that the functional tests can run inside a docker container.

### DIFF
--- a/tests/karma/karma.conf.js
+++ b/tests/karma/karma.conf.js
@@ -40,7 +40,14 @@ module.exports = function(config) {
 
     reporters: ['mocha', 'progress', 'coverage'],
 
-    browsers: ['ChromeHeadless'],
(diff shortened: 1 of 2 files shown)


shahi645:
functional tests are still very flakey, getting server side errors for numbers.spec.js, but is occurring both with and without this change

ajmaddaford:
It would be helpful to explain, or link to something that explains, the reasons for this.

ajmaddaford (approved):
It would be useful if the context of the PR bit was also in the commit message.

MIslam92 (approved):
👍

Issue #1230, opened 2017-09-07
500 Error from /session-expired


jonnyshaw89:
### Expected behaviour
When i hit the `/session-expired` url i should get a valid 200 response regardless of wether i still have a valid cookie

### Actual behaviour
When `/session-expired` is hit and no cookie exists a 500 error is thrown

### Steps to reproduce the behaviour
hit `/session-expired` in an incognito tab

### Technical information
```
{
    "url": "http://eq.onsdigital.uk/session-expired",
    "event": "an error has occurred",
    "created": "2017-09-07T11:16:36.555259Z",
    "service": "eq-survey-runner",
    "exception": [
        "Traceback (most recent call last):",
        "  File '/opt/python/run/venv/lib/python3.4/site-packages/flask/app.py', line 1612, in full_dispatch_request",
        "    rv = self.dispatch_request()",
        "  File '/opt/python/run/venv/lib/python3.4/site-packages/flask/app.py', line 1598, in dispatch_request",
        "    return self.view_functions[rule.endpoint](**req.view_args)",
        "  File '/opt/python/current/app/app/views/session.py', line 95, in get_session_expired",
        "    return render_theme_template(session['theme'], template_name='session-expired.html',",
        "  File '/opt/python/run/venv/lib/python3.4/site-packages/werkzeug/local.py', line 377, in <lambda>",
        "    __getitem__ = lambda x, i: x._get_current_object()[i]",
        "KeyError: 'theme'"
    ],
    "request_id": "2e8cd714-dae4-4053-a920-a4e97b2b40be",
    "level": "error",
    "status_code": 500
}
```

collisdigital:
I get this all the time, not cookie related?

PR #1244, opened 2017-09-18
Fix 500 Error from /session-expired and /signed-out

shahi645:
feature includes:

- when signed in user gets session-expired/signed-out with theme and title of survey
launched

- when not signed in (and no cookie) theme is set as default and no title

Resolves: #1230

### How to review 
Load survey runner launch a survey and navigate to 'localhost:5000/session-expired' or 'localhost:5000/signed-out'. page should match survey and survey title. 

Load survey, delet cookie, then navigate to above pages....user should see default page and no title

### DIFF
--- a/app/helpers/session_helper.py
+++ b/app/helpers/session_helper.py
@@ -1,4 +1,4 @@
-from flask import current_app, session
+from flask import current_app
 from flask_login import current_user, logout_user
 
 from app.globals import get_questionnaire_store
@@ -18,6 +18,3 @@ def end_session_with_schema_context(schema=None):
         schema = load_schema_from_metadata(metadata)
 
     _remove_survey_session_data()
-
-    session['theme'] = schema['theme']
-    session['survey_title'] = schema['title']
--- a/app/views/errors.py
+++ b/app/views/errors.py
@@ -1,4 +1,4 @@
-from flask import Blueprint, request, current_app
+from flask import Blueprint, request, current_app, session
 from flask_themes2 import render_theme_template
 from flask_login import current_user
 from flask_wtf.csrf import CSRFError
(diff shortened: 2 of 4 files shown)


app/helpers/session_helper.py
```diff
@@ -19,5 +19,4 @@ def end_session_with_schema_context(schema=None):
 
     _remove_survey_session_data()
 
-    session['theme'] = schema['theme']
-    session['survey_title'] = schema['title']
+
```
ajmaddaford: Extra line should be removed

ajmaddaford (approved):
The commit message should be `Fix 500 error from /session-expired`

ajmaddaford (changes requested):
I think we should have a test for the `/session-expired` and `/signed-out` urls returning a 200 rather than a 500.

collisdigital (approved):
Tested locally; works as described.

collisdigital (changes requested):
Changes to be made as discussed so that theme is used even on 401 if possible

collisdigital (approved):
Checked with @charlotte-hirst and working as needed; please squash.

app/views/errors.py
```diff
@@ -94,3 +95,13 @@ def get_tx_id():
...
+
+
+def render_template(template_name):
+    tx_id = get_tx_id()
+    user_agent = user_agent_parser.Parse(request.headers.get('User-Agent', ''))
+
+    return render_theme_template(session.get('theme', 'default'), template_name=template_name,
+                                 analytics_ua_id=current_app.config['EQ_UA_ID'],
+                                 ua=user_agent, tx_id=tx_id,
+                                 survey_title=TemplateRenderer.safe_content(session.get('survey_title', '')))
```
ajmaddaford: Each argument should be on a new line as it's far easier to read

ajmaddaford (changes requested):

yoganandkunche (approved):
@shahi645 Tested locally.. happy with the solution and it works as expected.

ajmaddaford (approved):

PR #1245, opened 2017-09-18
eq-json-descriptions Add descriptions

berianjones:
Add descriptions to json schemas as appropriate to aid documentation, re-use and development

Added descriptions to the schemas under data/schema to aid explanation. Also added descriptions to test_unit_patterns.json and test_big_list_naughty_strings.json to align them with other test schemas

New pull request on new branch to remove extraneous commit.
### How to review 
No tests as such, as no functionality has changed, just check that the descriptions are sensible. I have checked that the descriptions don't leak when the schemas are rendered

### DIFF
--- a/data/schema/feedback_v1.json
+++ b/data/schema/feedback_v1.json
@@ -5,6 +5,7 @@
   "properties": {
     "tx_id": {
       "type": "string",
+      "description": "Survey transaction identifier. This will be the same for all feedback during a survey session.",
       "pattern": "^[a-z0-9]{8}-[a-z0-9]{4}-[a-z0-9]{4}-[a-z0-9]{4}-[a-z0-9]{12}$"
     },
     "type": {
@@ -21,6 +22,7 @@
     },
     "survey_id": {
       "type": "string",
+      "description": "The ONS id of the survey derived from the inquiry code already in use for that survey.",
       "pattern": "^[0-9]{3}$"
     },
     "collection": {
@@ -39,6 +41,7 @@
     },
     "submitted_at": {
       "type": "string",
+      "description": "Time at which survey feedback was submitted.",
       "format": "date-time"
     },
     "metadata": {
--- a/data/schema/schema_v1.json
+++ b/data/schema/schema_v1.json
@@ -14,16 +14,16 @@
   "definitions": {
     "id": {
       "type": "string",
-      "description": "A lowercase alphanumeric string separated by hyphens.",
+      "description": "Used to identify groups, blocks, questions and answers.",
       "pattern": "^[a-z0-9][a-z0-9\\-]*[a-z0-9]$"
     },
     "q_code": {
       "type": "string",
-      "description": "A question code used by downstream systems to identify answers",
+      "description": "A question code used by downstream systems to identify answers.",
       "pattern": "^[0-9a-z]+$"
     },
     "guidance_content": {
-      "description": "Allows customisation of guidance text",
+      "description": "Allows customisation of guidance text.",
       "type": "array",
       "minItems": 1,
       "items": {
@@ -86,7 +86,8 @@
             "type": "string"
           },
           "meta": {
-            "type": "string"
+            "type": "string",
+            "description": "Metadata provided by the calling service. This will vary between surveys."
           },
           "condition": {
             "enum": [
@@ -124,6 +125,7 @@
     },
     "messages": {
       "type": "object",
+      "description": "These messages override the standard error messages.",
       "properties": {
         "MANDATORY_TEXTFIELD": {
           "type": "string"
@@ -172,6 +174,7 @@
     },
     "routing_rules": {
       "type": "array",
+      "description": "Used to direct the journey through a survey (in conjunction with navigation).",
       "items": {
         "type": "object",
         "properties": {
@@ -179,10 +182,12 @@
             "type": "object",
             "properties": {
               "id": {
-                "type": "string"
+                "type": "string",
+                "description": "The id of another block in the same group."
               },
               "group": {
-                "type": "string"
+                "type": "string",
+                "description": "The id of another group. The first block in that group will be routed to."
               },
               "when": {
                 "$ref": "#/definitions/when"
@@ -266,7 +271,8 @@
       ]
     },
     "survey_id": {
-      "type": "string"
+      "type": "string",
+      "description": "The ONS id of the survey derived from the inquiry code already in use for that survey."
     },
     "session_timeout_in_seconds": {
       "description": "The amount of time in seconds before timing out a users session.",
@@ -293,6 +299,7 @@
     },
     "navigation": {
       "type": "object",
+      "description": "Used in conjunction with routing to take user through a survey. ",
       "properties": {
         "visible": {
           "type": "boolean"
@@ -300,24 +307,36 @@
         "sections": {
           "type": "array",
           "items": {
-            "title": "string",
-            "title_from_answers": "array",
-            "group_order": "array",
-            "required": [
-              "group_order"
-            ],
-            "oneOf": [
-              {
-                "required": [
-                  "title"
-                ]
+            "type": "object",
+            "properties": {
+              "title": {
+                "type": "string",
+                "description": "Title for a given id."
               },
-              {
-                "required": [
-                  "title_from_answers"
-                ]
+              "title_from_answers": {
+                "type": "array",
+                "description": "Takes a list of answer ids. Title will be generated from answer values concatenated together with spaces."
+              },
+              "group_order": {
+                "type": "array",
+                "description": "Order in which groups will be navigated within this section."
               }
-            ]
+            },
+              "required": [
+                "group_order"
+              ],
+              "oneOf": [
+                {
+                  "required": [
+                    "title"
+                  ]
+                },
+                {
+                  "required": [
+                    "title_from_answers"
+                  ]
+                }
+              ]
           }
         }
       }
@@ -335,17 +354,6 @@
           },
           "title": {
             "type": "string"
-          },
-          "completed_id": {
-            "type": "string"
-          },
-          "highlight_when": {
-            "type": "array",
-            "items": {
-              "type": "string"
-            },
-            "minItems": 1,
-            "uniqueItems": true
           },
           "hide_in_navigation": {
             "type": "boolean"
@@ -469,7 +477,8 @@
                               "$ref": "#/definitions/id"
                             },
                             "parent_answer_id": {
-                              "type": "string"
+                              "type": "string",
+                              "description": "The id of an answer which is the parent of this child answer, for example a radio option leading to a text field answer."
                             },
                             "q_code": {
                               "$ref": "#/definitions/q_code"
@@ -481,10 +490,12 @@
                               "type": "object",
                               "properties": {
                                 "show_guidance": {
-                                  "type": "string"
+                                  "type": "string",
+                                  "description": "The text that is used for the 'Show guidance' link."
                                 },
                                 "hide_guidance": {
-                                  "type": "string"
+                                  "type": "string",
+                                  "description": "The text that is used for the 'Hide guidance' link."
                                 },
                                 "content": {
                                   "$ref": "#/definitions/guidance_content"
@@ -524,7 +535,8 @@
                                     "$ref": "#/definitions/q_code"
                                   },
                                   "child_answer_id": {
-                                    "type": "string"
+                                    "type": "string",
+                                    "description": "The id of an answer that should be treated as a child of this one, for example a text field answer can be the child of a radio option. The id must be of an answer in this list of answers."
                                   }
                                 },
                                 "required": [
@@ -540,22 +552,23 @@
                               "type": "integer"
                             },
                             "decimal_places": {
-                              "type": "integer"
+                              "type": "integer",
+                              "description": "Number of decimal places allowed"
                             },
                             "alias": {
-                              "type": "string"
-                            },
-                            "repeats": {
-                              "type": "boolean"
+                              "type": "string",
+                              "description": "An alias that can be used when piping this answer."
                             },
                             "max_value": {
                               "type": "object",
                               "properties": {
                                 "value": {
-                                  "type": "integer"
+                                  "type": "integer",
+                                  "description": "Maximum numeric value allowed."
                                 },
                                 "answer_id": {
-                                  "type": "string"
+                                  "type": "string",
+                                  "description": "The id of an answer from which to obtain the max_value"
                                 }
                               },
                               "oneOf": [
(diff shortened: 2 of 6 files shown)


shahi645 (approved):

data/schema/feedback_v1.json
```diff
@@ -5,6 +5,7 @@
   "properties": {
     "tx_id": {
       "type": "string",
+      "description": "A lowercase alphanumeric string separated by hyphens.",
```
ajmaddaford: The description should describe what is it is rather than it's structure (which is covered by the regex), something like "Survey transaction identifier. Will be the same for all feedback during a survey session."

data/schema/feedback_v1.json
```diff
@@ -39,6 +41,7 @@
     },
     "submitted_at": {
       "type": "string",
+      "description": "Survey submission date-time-stamp.",
```
ajmaddaford: This should refer to feedback rather than survey, something like "Time at which the feedback was submitted"

data/schema/schema_v1.json
```diff
@@ -114,6 +114,7 @@
     },
     "messages": {
       "type": "object",
+      "description": "Standardised error and warning messages for all schemas.",
```
ajmaddaford: This isn't quite right, messages contains content that overrides the standard error messages. Something like "These messages override the standard error messages" would be correct.

data/schema/schema_v1.json
```diff
@@ -162,6 +163,7 @@
     },
     "routing_rules": {
       "type": "array",
+      "description": "Used to direct the journey through a schema (in conjunction with navigation).",
```
ajmaddaford: I think "schema" should be "survey"

data/schema/schema_v1.json
```diff
@@ -283,6 +286,7 @@
     },
     "navigation": {
       "type": "object",
+      "description": "Used in conjunction with routing to take user through a schema.",
```
ajmaddaford: I think "schema" should be "survey"

ajmaddaford (changes requested):
There are a few elements that don't have descriptions that probably should:
- when `meta`
- routing_rules `goto.id`, `goto.group`
- navigation `title`, `title_from_answers`, `group_order`
- answer `options.child_answer_id`, `guidance.show_guidance`, `guidance.hide_guidance`, `alias`, `repeats`, `max_value.answer_id`, `min_value.answer_id`, `calculated`

I think we should also remove elements that are no longer used:
- groups `completed_id`, `highlight_when`

data/schema/feedback_v1.json
berianjones: Done. Is the description for schema.json correct? I took the description for feedback.json from this.

data/schema/feedback_v1.json
berianjones: Done

data/schema/schema_v1.json
berianjones: Done

data/schema/schema_v1.json
berianjones: Done

data/schema/schema_v1.json
berianjones: Done

data/schema/feedback_v1.json
ajmaddaford: Good spot, I think that should be something like "Used to identify groups, blocks, questions and answers"

berianjones: "There are a few elements that don't have descriptions that probably should:
when meta
routing_rules goto.id, goto.group
navigation title, title_from_answers, group_order
answer options.child_answer_id, guidance.show_guidance, guidance.hide_guidance, alias, repeats, max_value.answer_id, min_value.answer_id, calculated"

Descriptions added for these

berianjones: "I think we should also remove elements that are no longer used:

groups completed_id, highlight_when"

These have been removed

data/schema/schema_v1.json
```diff
@@ -162,17 +164,20 @@
...
       "items": {
         "type": "object",
         "properties": {
           "goto": {
             "type": "object",
             "properties": {
               "id": {
-                "type": "string"
+                "type": "string",
+                "description": "Moves the user to an id of a particular part of a survey, either within the current block or elsewhere."
```
ajmaddaford: This documents how this property is used, rather than what the property e.g "The id of another block in the same group."

data/schema/schema_v1.json
```diff
@@ -162,17 +164,20 @@
...
               },
               "group": {
-                "type": "string"
+                "type": "string",
+                "description": "Moves the user to the start of a particular group within a survey."
```
ajmaddaford: Same as above, this should be something like "The id of another group. The first block in that group will be routed to".

data/schema/feedback_v1.json
berianjones: Amended the id description on schema.json to contain the description above, will commit and push after any other feedback.

data/schema/schema_v1.json
```diff
@@ -283,16 +289,19 @@
...
           "type": "boolean"
         },
         "sections": {
           "type": "array",
+          "description": "Title - initial title, title_from_answers - title with information derived from answers",
           "items": {
             "title": "string",
             "title_from_answers": "array",
             "group_order": "array",
+            "description": "Order in which groups will appear during survey",
```
ajmaddaford: This adds a description property to `sections`, rather than add a description to `group_order`. To document each property in sections, you should be able to do something like:
```
"items": {
    "type": "object",
    "properties": {
        "title": {
            "type": "string"
            "description": "..."
        },
        "title_from_answers": {
            "type": "array"
            "description": "..."
        },
        "group_order": {
            "type": "array"
            "description": "..."
        }
    }
}
```

data/schema/schema_v1.json
```diff
@@ -471,10 +469,12 @@
                               "type": "object",
                               "properties": {
                                 "show_guidance": {
-                                  "type": "string"
+                                  "type": "string",
+                                  "description": "Enables guidance on how to answer the question to be displayed."
```
ajmaddaford: This property is the text that is used for the 'Show guidance' link

data/schema/schema_v1.json
```diff
@@ -471,10 +469,12 @@
...
                                 },
                                 "hide_guidance": {
-                                  "type": "string"
+                                  "type": "string",
+                                  "description": "Enables guidance on how to answer the question to be hidden."
```
ajmaddaford: This property is the text that is used for the 'Hide guidance' link

data/schema/schema_v1.json
```diff
@@ -514,7 +514,8 @@
                                     "$ref": "#/definitions/q_code"
                                   },
                                   "child_answer_id": {
-                                    "type": "string"
+                                    "type": "string",
+                                    "description": "Response to follow up question resulting from an answer."
```
ajmaddaford: This isn't very clear, and I think needs more than a short description, something like 'The id of an answer that should be treated as a child of this one, for example a text field answer can be the child of a radio option. The id must be of an answer in this list of answers.'

data/schema/schema_v1.json
```diff
@@ -533,16 +534,19 @@
...
                             },
                             "alias": {
-                              "type": "string"
+                              "type": "string",
+                              "description": "Allows another name for a variable to be used."
                             },
                             "repeats": {
-                              "type": "boolean"
+                              "type": "boolean",
+                              "description": "Value showing whether there are multiple answers for a question or not."
```
ajmaddaford: I don't think repeats is actually used. It only appears in `test_navigation_confirmation.json`. If it's not used it should be removed from `test_navigation_confirmation.json` and this schema.

data/schema/schema_v1.json
```diff
@@ -533,16 +534,19 @@
                               "type": "integer"
                             },
                             "alias": {
-                              "type": "string"
+                              "type": "string",
+                              "description": "Allows another name for a variable to be used."
```
ajmaddaford: Doesn't quite capture what this is for, something like "An alias that can be used when piping this answer" would be better.

data/schema/schema_v1.json
```diff
@@ -533,16 +534,19 @@
...
+                              "type": "boolean",
+                              "description": "Value showing whether there are multiple answers for a question or not."
                             },
                             "max_value": {
                               "type": "object",
                               "properties": {
                                 "value": {
-                                  "type": "integer"
+                                  "type": "integer",
+                                  "description": "Maximum numeric value allowed in answer."
```
ajmaddaford: - I think we can drop the 'in answer' bit. Same for `min_value`. 
- The `answer_id` also needs a description, something like "The id of an answer from which to obtain the max_value", and something similar for the `min_value.answer_id`.

data/schema/schema_v1.json
```diff
@@ -76,7 +76,8 @@
             "type": "string"
           },
           "meta": {
-            "type": "string"
+            "type": "string",
+            "description": "Metadata item which will take a particular action in a survey"
```
ajmaddaford: This should be "Metadata provided by the calling service. This will vary between surveys."

ajmaddaford (changes requested):

data/schema/schema_v1.json
berianjones: Done

data/schema/schema_v1.json
berianjones: Done

data/schema/schema_v1.json
berianjones: Done

data/schema/schema_v1.json
berianjones: Done

data/schema/schema_v1.json
berianjones: Done

data/schema/schema_v1.json
berianjones: Done, should we add a description for Parent_Id, perhaps an inverse of the child_answer_id,

'The id of an answer which is the parent of one or more child answers, for example a radio option leading to a text field answer.'

data/schema/schema_v1.json
berianjones: done

data/schema/schema_v1.json
berianjones: done

data/schema/schema_v1.json
berianjones: Removed these after testing test_navigation_confirmation without the repeats element

data/schema/schema_v1.json
```diff
@@ -296,12 +296,19 @@
...
           "type": "array",
-          "description": "Title - initial title, title_from_answers - title with information derived from answers",
           "items": {
-            "title": "string",
-            "title_from_answers": "array",
-            "group_order": "array",
-            "description": "Order in which groups will appear during survey",
+            "title": {
+            "type": "string",
+            "description": "Initial title."
```
ajmaddaford: - The formatting is slightly off here, `type` and `description` should be indented
- The description is a bit vague, if we can't think of anything of value it might be worth just dropping the description.

data/schema/schema_v1.json
```diff
@@ -296,12 +296,19 @@
...
+            },
+            "title_from_answers": {
+            "type": "array",
+            "description": "Modified title with information derived from answers."
+            },
+            "group_order": {
+            "type": "array",
+            "description": "Order in which groups will appear during survey"
```
ajmaddaford: I think we should change "will appear during survey" to "will be navigated within this section"

data/schema/schema_v1.json
```diff
@@ -296,12 +296,19 @@
...
-            "title_from_answers": "array",
-            "group_order": "array",
-            "description": "Order in which groups will appear during survey",
+            "title": {
+            "type": "string",
+            "description": "Initial title."
+            },
+            "title_from_answers": {
+            "type": "array",
+            "description": "Modified title with information derived from answers."
```
ajmaddaford: I think we should extend this description with "Takes a list of answer ids. Title will be generated from answer values concatenated together with spaces".

ajmaddaford (changes requested):

data/schema/schema_v1.json
ajmaddaford: I can't see `parent_id` in the schema

data/schema/schema_v1.json
berianjones: Indentation done, title description amended to 

'Title for given id'

data/schema/schema_v1.json
berianjones: done

data/schema/schema_v1.json
berianjones: done

data/schema/schema_v1.json
berianjones: sorry should have written parent_answer_id, it was getting late and schema blind

data/schema/schema_v1.json
```diff
@@ -459,7 +467,8 @@
                               "$ref": "#/definitions/id"
                             },
                             "parent_answer_id": {
-                              "type": "string"
+                              "type": "string",
+                              "description": "The id of an answer which is the parent of one or more child answers, for example a radio option leading to a text field answer."
```
ajmaddaford: At the moment a parent answer can only have one child.

data/schema/schema_v1.json
```diff
@@ -599,7 +614,8 @@
                               }
                             },
                             "calculated": {
-                              "type": "boolean"
+                              "type": "boolean",
+                              "description": "Value calculated from answers."
```
ajmaddaford: This isn't a calculated value, it's a boolean that indicates if this answer should be treated as the result of a calculation.

ajmaddaford (changes requested):

data/schema/schema_v1.json
berianjones: updated

data/schema/schema_v1.json
berianjones: Indeed. New text

"Shows that answer is to be treated as result of calculation or not."

data/schema/schema_v1.json
```diff
@@ -468,7 +468,7 @@
                             },
                             "parent_answer_id": {
                               "type": "string",
-                              "description": "The id of an answer which is the parent of one or more child answers, for example a radio option leading to a text field answer."
+                              "description": "The id of an answer which is the parent of a child answer, for example a radio option leading to a text field answer."
```
ajmaddaford: This should be '... of *this* child answer'

data/schema/schema_v1.json
```diff
@@ -615,7 +615,7 @@
                             },
                             "calculated": {
                               "type": "boolean",
-                              "description": "Value calculated from answers."
+                              "description": "Shows that answer is to be treated as result of calculation or not."
```
ajmaddaford: This would read better as 'Indicates that the answer should be treated as the result of a calculation.'

ajmaddaford (changes requested):

data/schema/schema_v1.json
berianjones: changed

data/schema/schema_v1.json
berianjones: agreed, changed

ajmaddaford (approved):

berianjones:
Squashed and ready

Issue #1249, opened 2017-09-19
Error on MWSS when calendar monthly is first pay pattern


ajmaddaford:
### Expected behaviour
When completing an MWSS survey and confirming choice of just "Calendar monthly" the calendar monthly introduction screen should be shown.

### Actual behaviour
A HTTP 500 error

### Steps to reproduce the behaviour
- Start an MWSS survey
- Choose just "Calendar monthly" as a pay period
- Confirm choice

PR #1250, opened 2017-09-19
Fix MWSS Calendar monthly routing error

ajmaddaford:
Fixes an issue with MWSS that had a routing rule to a group that didn't
exist. Also adds a schema validation test to validate all schemas with
routing rules to groups only route to groups that exist.

Fixes: #1249

See #1249

### How to review 
- Start an MWSS survey
- Choose just "Calendar monthly" as a pay period
- Confirm choice
- Confirm 500 error doesn't occur

### DIFF
--- a/data/en/1_0005.json
+++ b/data/en/1_0005.json
@@ -294,7 +294,7 @@
                         },
                         {
                             "goto": {
-                                "group": "calendar-monthly-pay",
+                                "group": "calendar-monthly",
                                 "when": [{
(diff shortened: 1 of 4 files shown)


collisdigital (approved):
Tested locally; MWSS working and tried corrupting schema and validation script picks it up.

yoganandkunche (approved):
Looks good to me. Raised another issue #1251  which I will discussed with @charlotte-hirst if we want it for this release.

PR #1265, opened 2017-09-26
Fixes exponential numeric values issue

yoganandkunche:
This will now restrict the respondent to enter exponential values with in numeric fields (Example: 2e2)

### How to review 
Launch any business survey where you can enter a numeric value
Enter a numeric value as: 2e2 and submit
Must now throw an error: Enter a number.


berianjones (approved):
Tested with upper and lower case e, and tried different combinations of letters/numbers

jcox-dev:
Tested and LGTM 👍

jcox-dev (approved):
Tested and LGTM 👍

PR #1269, opened 2017-09-28
Fix for #1247, able to submit with invalid answer id

### LINKED ISSUE #1247: Unable to submit with invalid answer id in answer store
### Expected behaviour
- Submission succeeds even if an answer id in the answer stores doesn't exist in the schema

### Actual behaviour
- Submission fails with a 500 error.

- Start a survey and enter data
- Change an answer id in the schema 
- Resume the survey
- Try and dump the submission or submit
- See a 500 error with a key error on the now missing answer id.

### Info
This will occur if a schema changes (removes an answer id) and a respondent has pre-existing data stored.

### Screenshot
Launch pre-prod with:
image

Try submitting or dumping submission:
screen shot 2017-09-19 at 09 05 59

shahi645:
scenario:
- when SR launched and data entered and then 'saved for later'
- if the answer id was changed in survey json schema
- then SR launched and going back to saved data

previously:
- resulted in a 500 error

currently:
- allows submission

### How to review 
run scenario described in context, on submission should now see successful submission not a 500

Fixes #1247

### DIFF
--- a/app/submitter/converter.py
+++ b/app/submitter/converter.py
@@ -89,11 +89,17 @@ def convert_answers_to_data(answer_store, questionnaire_json, routing_path):
         answer_schema_list = SchemaHelper.get_answers_by_id_for_block(block_json)
 
         for answer in answers_in_block:
-            answer_schema = answer_schema_list[answer['answer_id']]
+            try:
+                answer_schema = answer_schema_list[answer['answer_id']]
+            except KeyError:
+                logger.error("No matching answer schema found in the store")
+                continue
+
             value = answer['value']
 
             if answer_schema is not None and value is not None and 'parent_answer_id' not in answer_schema:
-                if answer_schema['type'] != 'Checkbox' or any('q_code' not in option for option in answer_schema['options']):
+                if answer_schema['type'] != 'Checkbox' or any('q_code' not in option for option in
+                                                              answer_schema['options']):
                     if 'q_code' not in answer_schema:
                         continue
 
--- a/tests/app/submitter/test_converter.py
+++ b/tests/app/submitter/test_converter.py
@@ -2,10 +2,12 @@ from datetime import datetime, timedelta, timezone
 
 import dateutil.parser
 
+from mock import patch
+
 from app.data_model.answer_store import AnswerStore
 from app.questionnaire.location import Location
 from app.storage.metadata_parser import parse_metadata
-from app.submitter.converter import convert_answers, DataVersionError
+from app.submitter.converter import convert_answers, DataVersionError, convert_answers_to_data
 from tests.app.framework.survey_runner_test_case import SurveyRunnerTestCase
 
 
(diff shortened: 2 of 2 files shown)


app/submitter/converter.py
```diff
@@ -89,7 +89,7 @@ def convert_answers_to_data(answer_store, questionnaire_json, routing_path):
         answer_schema_list = SchemaHelper.get_answers_by_id_for_block(block_json)
 
         for answer in answers_in_block:
-            answer_schema = answer_schema_list[answer['answer_id']]
+            answer_schema = answer_schema_list.get(answer.get('answer_id', ''), None)
```
jcox-dev: A little unreadable. Perhaps change to the following to keep it simple?

```python
try:
    answer_schema = answer_schema_list[answer['answer_id']]
except KeyError:
    answer_schema = None
```

jcox-dev (changes requested):

app/submitter/converter.py
jonnyshaw89: Doesn't `.get()` default to returning `None` so no need to specify it?

jonnyshaw89 (changes requested):

app/submitter/converter.py
shahi645: i thought the '.get' might be better but this does look more readable

jcox-dev (approved):

app/submitter/converter.py
```diff
@@ -89,7 +89,10 @@ def convert_answers_to_data(answer_store, questionnaire_json, routing_path):
         answer_schema_list = SchemaHelper.get_answers_by_id_for_block(block_json)
 
         for answer in answers_in_block:
-            answer_schema = answer_schema_list[answer['answer_id']]
+            try:
+                answer_schema = answer_schema_list[answer['answer_id']]
+            except KeyError:
+                answer_schema = None
```
collisdigital: Can we get some error logging here

collisdigital: ideally with some info about the actual id/answer causing the issue

jonnyshaw89: I think we should add logging as @collisdigital mentions. Also given that the next statement is `if answer_schema is not None` we should just `return data` and not bother evaluating the rest of the code?

jonnyshaw89 (changes requested):

app/submitter/converter.py
```diff
@@ -92,11 +92,14 @@ def convert_answers_to_data(answer_store, questionnaire_json, routing_path):
             try:
                 answer_schema = answer_schema_list[answer['answer_id']]
             except KeyError:
-                answer_schema = None
+                logger.info("There is no answer id in the store", answer_id=answer['answer_id'])
```
jcox-dev: Change to `logger.error("No matching answer schema found in the store")` to avoid a second `KeyError`.

tests/app/submitter/test_converter.py
```diff
@@ -131,6 +134,40 @@ def test_convert_answers(self):
...
+                                }]
+                            }
+                        ]
+                    }
+                ]
+            }
+
+            routing_path = [Location(group_id='group-1', group_instance=0, block_id='block-1')]
+            with patch('app.submitter.converter.logger') as patched_logger:
+                convert_answers(AnswerStore(user_answer), questionnaire, routing_path)
```
jcox-dev: Change to `convert_answers_to_data(...)` (fails with TypeError - missing metadata - otherwise).

tests/app/submitter/test_converter.py
```diff
@@ -1,11 +1,14 @@
 from datetime import datetime, timedelta, timezone
 
 import dateutil.parser
+import logging
```
jcox-dev: `logging` is unused here and should be removed.

tests/app/submitter/test_converter.py
```diff
@@ -131,6 +134,40 @@ def test_convert_answers(self):
...
+                            }
+                        ]
+                    }
+                ]
+            }
+
+            routing_path = [Location(group_id='group-1', group_instance=0, block_id='block-1')]
+            with patch('app.submitter.converter.logger') as patched_logger:
+                convert_answers(AnswerStore(user_answer), questionnaire, routing_path)
+                self.assertEqual(patched_logger.info.call_count, 1)
```
jcox-dev: Change to `...patched_logger.error...` once changes to logging in [converter](https://github.com/ONSdigital/eq-survey-runner/blob/6c2d5cbaed01fd3fbc7407f806b4c370b1fd1289/app/submitter/converter.py#L95) have been applied.

jcox-dev (changes requested):

jcox-dev (approved):

app/submitter/converter.py
```diff
@@ -89,11 +89,17 @@ def convert_answers_to_data(answer_store, questionnaire_json, routing_path):
         answer_schema_list = SchemaHelper.get_answers_by_id_for_block(block_json)
 
         for answer in answers_in_block:
-            answer_schema = answer_schema_list[answer['answer_id']]
+            try:
+                answer_schema = answer_schema_list[answer['answer_id']]
+            except KeyError:
+                logger.error("No matching answer schema found in the store")
+                return data
```
jonnyshaw89: Thinking about it, this should be continue not return.
As if you have multiple answers it will return after the 1st one.
You should add a test with a mixture of valid and invalid answers and test that all of the value answers are returned by this method

jonnyshaw89 (changes requested):

tests/app/submitter/test_converter.py
```diff
@@ -131,6 +133,53 @@ def test_convert_answers(self):
...
+                        ]
+                    }
+                ]
+            }
+
+            routing_path = [Location(group_id='group-1', group_instance=0, block_id='block-1')]
+            with patch('app.submitter.converter.logger') as patched_logger:
+                answer_object = convert_answers_to_data(AnswerStore(user_answer), questionnaire, routing_path)
+                self.assertEqual(patched_logger.error.call_count, 2)
+                self.assertEqual(answer_object['002'], '2016-03-30')
```
jonnyshaw89, line 181: You should also check that this is the only answer?

jonnyshaw89 (changes requested):

jonnyshaw89 (approved):

Issue #1266, opened 2017-09-26
Redundant code generated a HTTP 500 error


ajmaddaford:
### Expected behaviour
Visiting the `/session-expired` endpoint with a cookie and no valid metadata should not give a HTTP 500 error.

### Actual behaviour
Seen in production:
```
"exception": [
        "Traceback (most recent call last):",
        "  File '/opt/python/run/venv/lib/python3.4/site-packages/flask/app.py', line 1612, in full_dispatch_request",
        "    rv = self.dispatch_request()",
        "  File '/opt/python/run/venv/lib/python3.4/site-packages/flask/app.py', line 1598, in dispatch_request",
        "    return self.view_functions[rule.endpoint](**req.view_args)",
        "  File '/opt/python/current/app/app/views/session.py', line 95, in get_session_expired",
        "    end_session_with_schema_context()",
        "  File '/opt/python/current/app/app/helpers/session_helper.py', line 18, in end_session_with_schema_context",
        "    schema = load_schema_from_metadata(metadata)",
        "  File '/opt/python/current/app/app/utilities/schema.py', line 20, in load_schema_from_metadata",
        "    return load_schema_from_params(metadata['eq_id'], metadata['form_type'], metadata.get('language_code'))",
        "KeyError: 'eq_id'"
    ]
```

### Steps to reproduce the behaviour
- Start the same survey in two different browsers
- Complete the survey in the first browser
- Visit the `/session-expired` page in the other browser

ajmaddaford:
There is some unnecessary code in `session_helper.py` that attempts to load the schema from metadata leading to this error. As part of fixing this we should investigate if there is a lint rule that can fail builds on unused variables.

PR #1270, opened 2017-09-28
Fix redundant code generating a HTTP 500 error

jcox-dev:
Session helper [`end_session_with_schema_context`](https://github.com/ONSdigital/eq-survey-runner/commit/339f0baadda65717227968e421973fbe080f8d81#diff-94ead32a421bdfa2e563a3fe15b53589R14) was necessary before template function refactor to add theme context to a session but is now redundant and has been removed.

Following the amendments to [how templates are rendered](https://github.com/ONSdigital/eq-survey-runner/commit/ad4f69a0b564141e90305da46721a9cf0ce2beac), the helper method `end_session_with_schema_context` is no longer required and in fact can cause an error when trying to use metadata after a submission #1266.

### How to review 
Ensure all references to old function (`end_session_with_schema_context`) has been removed; all tests pass; and themes are still rendered after signing out / submitting.

### DIFF
--- a/app/helpers/session_helper.py
+++ b/app/helpers/session_helper.py
@@ -1,20 +1,8 @@
 from flask import current_app
-from flask_login import current_user, logout_user
-
-from app.globals import get_questionnaire_store
-from app.utilities.schema import load_schema_from_metadata
+from flask_login import logout_user
 
 
-def _remove_survey_session_data():
+def remove_survey_session_data():
     current_app.eq['session_storage'].delete_session_from_db()
     current_app.eq['session_storage'].remove_user_ik()
     logout_user()
-
-
-def end_session_with_schema_context(schema=None):
-    if schema is None:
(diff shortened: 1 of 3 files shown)


app/helpers/session_helper.py
```diff
@@ -1,20 +1,8 @@
 from flask import current_app
-from flask_login import current_user, logout_user
-
-from app.globals import get_questionnaire_store
-from app.utilities.schema import load_schema_from_metadata
+from flask_login import logout_user
 
 
 def _remove_survey_session_data():
```
ajmaddaford: We should remove the leading `_`

ajmaddaford (changes requested):

yoganandkunche (approved):
LGTM :+1:

ajmaddaford (approved):

PR #1272, opened 2017-09-28
Bind tx_id to logger for all requests

jonnyshaw89:
tx_id was not being logged for all questionnaire requests so this has not been added.
tx_id was not being logged in any error logs so this has now been added to log on error messages where we have a metadata.

### How to review 
Test that all log messages contain a `tx_id`

### DIFF
--- a/app/views/errors.py
+++ b/app/views/errors.py
@@ -22,12 +22,6 @@ class MultipleSurveyError(Exception):
     pass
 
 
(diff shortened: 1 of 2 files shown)


jcox-dev (approved):
Tested error pages and questionnaire requests and tx_id is present and bound to logger. 👍

dwyeradam (approved):

PR #1282, opened 2017-10-03
eq-1374 Revisit stylelint rules

hamishtaplin:

Tightens up stylelint rules:

- made stylelint actually work again 😯
- switched to `gulp-stylelint`
- disallows use of `!important`
- no `id`s as selectors
- 4 levels of specificity on classes (this is too much but would require too much refactoring to fix)
- 1 level of specificity using tag names as selectors
- double-quotes everywhere
- removed unused prism library css
- moved vendor folder out of components

NOTE: some changes are from my editor formatting stuff

### How to review 

This shouldn't introduce and styling changes but it's worth running through a survey to be sure.


app/assets/styles/partials/base/_print.scss
```diff
@@ -1,3 +1,5 @@
+/* stylelint-disable */
```
shahi645, line 1: Is this disable because of the '!important' as below we use *stylelint-disable declaration-no-important*

hamishtaplin: @shahi645 the print stylesheet would require quite a refactor to pass lint so I just disabled it completely

shahi645 (approved):
Tested by changing quotes inside and outside where *stylelint* is used, as runs as expected. LGTM 👍🏿

bennpearson (dismissed):
Seems to be formatting to single quotes instead of double. Might be my editor though

shahi645:
@bennpearson worked for double quotes with me, and then when i changed to single quotes threw up lint errors

bennpearson (approved):

PR #1286, opened 2017-10-04
Company details display changes

dwyeradam:
As a respondent
I want to have the company details displayed
So that I am clear which company I am completing for and I can use as a reference if I choose to keep a copy of the answers

### How to review 
Period details are shown in all banners and changes with conditional dates.
Introduction page shows company details in following format:
"You are completing this for _Company Name_. (trading as _Company Trading Name_.)"


dwyeradam:
topbar.html has far too many if statements in. I would appreciate some comments regarding how to better handle this.

yoganandkunche (changes requested):
@dwyeradam 
1. When there is no RU_Name in the meta data - period is not displayed in the header
2. UKIS text alignment is not looking right - as discussed
3. We will treat your data securely and confidentially. — this is not consistent on all surveys not sure it was intentional.?

app/templates/partials/topbar.html
```diff
@@ -1,13 +1,38 @@
...
+        {% if content and content.variables.period %}
+            {% set period = content.variables.period %}
+        {% endif %}
+
+        {% if meta and meta.respondent.address %}
+            {% set respondent = meta.respondent.address.name %}
+        {% endif %}
+    {% endif %}
+
+    {% if is_intro %}
```
ajmaddaford: I think we should remove `is_intro` and create a new `topbar-introduction.html` that extends `topbar.html`, overriding the bar content, something like:
```
{% extends theme('partials/topbar.html') %}

{% block topbar_content %}
      <h1 class="bar__title jupiter">{{ survey_title }}
        {% if period %}
            <p class="u-mb-xs pluto">
                <span class="config2">{{ period }}</span>
            </p>
        {% endif %}
      </h1>
{% endblock %}
```

app/templates/partials/topbar.html
```diff
@@ -1,13 +1,38 @@
...
     {% else %}
-      <div class="bar__title venus">{{ survey_title }}</div>
+      <div class="bar__title venus">{{ survey_title }}
+        <span class="mars">
+        {% if respondent %}
+            for&nbsp;<span class="config1">{{ respondent }}</span>
+        {% endif %}
+        {% if period %}
+            <p class="u-mb-s pluto" style="margin-bottom:0;">Period:
+                <span class="config2">{{ period }}</span>
```
ajmaddaford: Where do the classes `config1` and `config2` come from?

app/templates/partials/topbar.html
```diff
@@ -1,13 +1,38 @@
...
+      </h1>
     {% else %}
-      <div class="bar__title venus">{{ survey_title }}</div>
+      <div class="bar__title venus">{{ survey_title }}
+        <span class="mars">
+        {% if respondent %}
+            for&nbsp;<span class="config1">{{ respondent }}</span>
+        {% endif %}
+        {% if period %}
+            <p class="u-mb-s pluto" style="margin-bottom:0;">Period:
```
ajmaddaford: I don't think we should be doing any inline styles

app/views/questionnaire.py
```diff
@@ -84,12 +84,14 @@ def get_block(routing_path, eq_id, form_type, collection_id, group_id, group_ins
     if not _is_valid_group(group_id) or not _is_valid_location(current_location):
         return _redirect_to_latest_location(routing_path, collection_id, eq_id, form_type)
 
-    block = _render_schema(routing_path, current_location)
+    rendered_schema = _render_schema(routing_path, current_location)
+    block = rendered_schema['block']
+    schema_variables = rendered_schema['variables']
```
ajmaddaford: Can we move the rendering of the block and variables to `_render_block`?

ajmaddaford (changes requested):

app/templates/partials/topbar.html
shahi645: i couldn't find them either

dwyeradam: Copied from prototype without thinking.

app/templates/partials/topbar.html
dwyeradam: Copied from the prototype but removed and uses class="u-mb-xs pluto" for the same effect.

yoganandkunche:
@dwyeradam as discussed 1) text 'period:' to be added on the landing page for all business surveys
2) QBS period must display to 'from' and 'to' dates if exits in metadata

app/templates/introduction.html
```diff
@@ -2,92 +2,76 @@
...
-
-{%- if legal_basis == 'StatisticsOfTradeAct' -%}
-  {% set legal_text = "Your response is legally required" %}
-{%- endif -%}
+    {% set form = content.form %}
 
-{% block page_title %}{{survey_title}}{% endblock %}
+    {% block page_title %}
+        {{survey_title}}
+    {% endblock %}
```
ajmaddaford: I think this should be left in one line as it's just replacing the title of the page. Also it's consistently on one line in other templates.

app/templates/introduction.html
```diff
@@ -2,92 +2,76 @@
...
   {% include theme('partials/feedback/expandable_inline.html') %}
 
   {% block business_details %}
 
-    <div class="u-mt-s">
+            <div class="grid">
+            <div class="grid__col col-7@m">
+
+            <h2 class="saturn">You are completing this for
+                <span class="config1">{{ meta.respondent.address.name}}</span>
```
ajmaddaford: - This needs a proper class, or the config1 removed?
- Can we take this as an opportunity to remove the unnecessary address element in the metadata?

app/templates/introduction.html
```diff
@@ -2,92 +2,76 @@
...
+                {% if meta.respondent.address.trading_as %}
+                    <span class="trading_as">(trading as&nbsp;<span class="trad_as">{{meta.respondent.address.trading_as }}</span>)</span>
+                {% endif %}
+            </h2>
+
+            <p class="mars">If the company details or structure have changed contact us
+                </br>on {{ helpers.telephone_number() }} or email {{ helpers.email_address('details-changed-title', subject="Change of details reference " + meta.respondent.respondent_id) }}.
```
ajmaddaford: Having a break in the middle of a sentence causes some issues when the screen with is made smaller:
image

app/templates/partials/topbar-intro.html
```diff
@@ -0,0 +1,11 @@
+<div class="bar" role="banner">
```
ajmaddaford: We don't tend to shorten the filenames of templates, this should use be `topbar-introduction.html`

tests/integration/questionnaire/test_questionnaire_page_titles.py
```diff
@@ -8,7 +8,7 @@ def test_should_have_question_in_page_title_when_loading_introduction(self):
         # Given, When
         self.launchSurvey('test', 'final_confirmation')
         # Then
-        self.assertEqualPageTitle('Final confirmation to submit')
+        self.assertEqualPageTitle('\n        Final confirmation to submit\n    ')
```
ajmaddaford: This looks a little odd, why does the page title now have newlines and spaces in it?

app/views/questionnaire.py
```diff
@@ -497,40 +499,52 @@ def _redirect_to_location(collection_id, eq_id, form_type, location):
...
-        metadata = get_metadata(current_user)
-        aliases = SchemaHelper.get_aliases(g.schema_json)
-        schema_context = build_schema_context(metadata=metadata,
-                                              aliases=aliases,
-                                              answer_store=answer_store,
-                                              routing_path=full_routing_path)
         rendered_schema_json = renderer.render(g.schema_json, **schema_context)
-        content.update({'summary': build_summary_rendering_context(rendered_schema_json, answer_store, metadata)})
+        context.update({'summary': build_summary_rendering_context(rendered_schema_json, answer_store, metadata)})
+
```
ajmaddaford: I think the summary block needs to be dealt with first, as it doesn't need `block` and `form` but they're both generated anyway.

app/views/questionnaire.py
```diff
@@ -497,40 +499,52 @@ def _redirect_to_location(collection_id, eq_id, form_type, location):
...
+    schema_context = _get_schema_context(full_routing_path, current_location, metadata, answer_store)
+
+    block = renderer.render(block, **schema_context)
+
+    variables = g.schema_json.get('variables')
+    if variables:
+        variables = renderer.render(variables, **schema_context)
+
+    if post_form:
+        return {'form': post_form, 'block': block, 'variables': variables}
```
ajmaddaford: I think the we should return from the same place for `post_form` or `form` as the only difference would be the form object generated. As it stands it's not clear they have the same format return data.

ajmaddaford (changes requested):

app/views/questionnaire.py
dwyeradam: It appears summary does require "form". Have pushed the rest of the changes and will investigate this further.

app/templates/introduction.html
```diff
@@ -2,92 +2,74 @@
...
-{%- if legal_basis == 'StatisticsOfTradeAct' -%}
-  {% set legal_text = "Your response is legally required" %}
-{%- endif -%}
+    {% set form = content.form %}
 
-{% block page_title %}{{survey_title}}{% endblock %}
+    {% block page_title %}{{survey_title}}{% endblock %}
 
-{% block main %}
+    {% block main %}
```
ajmaddaford: These shouldn't be indented as they're all top level blocks

app/templates/introduction.html
```diff
@@ -2,92 +2,74 @@
...
+                {% if meta.respondent.trading_as %}
+                    <span class="trading_as">(trading as&nbsp;<span class="trad_as">{{meta.respondent.trading_as }}</span>)</span>
+                {% endif %}
+            </h2>
+
+            <p class="mars">If the company details or structure have changed contact us on {{ helpers.telephone_number() }}
+                or email {{ helpers.email_address('details-changed-title', subject="Change of details reference " + meta.respondent.respondent_id) }}.
+            </p>
+
+        {% endblock business_details %}
```
ajmaddaford: This should line up with the start of the `business_details` block

ajmaddaford (changes requested):

ajmaddaford:
The company name is missing on the submission successful screen:
image

ajmaddaford:
Maybe one for @armstrongb, is the duplication of the period below the section title expected, or should it be removed?
image

armstrongb:
@dwyeradam , @ajmaddaford the duplication of date can be removed from the group description. thanks.

ajmaddaford (approved):

yoganandkunche (approved):
LGTM 👍

yoganandkunche:
@dwyeradam rebase and squash?

PR #1290, opened 2017-10-09
eq-1404 add pylint rule for enforcing single quotes for variables 

jcox-dev:
Adds new pylint plugin, [pylint-quotes](https://github.com/edaniszewski/pylint-quotes), for enforcing single quotes for Python variables.


dwyeradam (approved):

app/authentication/jti_claim_storage.py
```diff
@@ -15,7 +15,7 @@ def __init__(self, jti_claim):
         self.jti_claim = jti_claim
 
     def __str__(self, *args, **kwargs):
-        return "jti claim '{jti_claim}' has already been used".format(jti_claim=self.jti_claim)
+        return 'jti claim \'{jti_claim}\' has already been used'.format(jti_claim=self.jti_claim)
```
ajmaddaford: Google's Python style guide suggests permitting using the alternate quote style if the string contains a quote i.e. use `"` if `'` is used in the string to avoid the backslashes - https://google.github.io/styleguide/pyguide.html?showone=Strings#Strings

ajmaddaford (changes requested):

app/authentication/jti_claim_storage.py
jcox-dev: Looks like the package we use to enforce these checks can be configured for "smart" checking with `single-avoid-escape`. I'll update the config to use that.

shahi645 (approved):
Tested locally, and changes behave as expected. 👍🏿

ajmaddaford (approved):

shahi645:
@jcox-dev squishy squashy

PR #1301, opened 2017-10-13
Switch to Pipenv for dependency management

jcox-dev:
Change package management process to using [Pipfile](https://github.com/pypa/pipfile) with the preferred tool [Pipenv](https://docs.pipenv.org/index.html)

- New workflow for package management documented in README.
- Truly isolated, deterministic, and temporary virtual environments for development created without having to use pyenv directly.
- Dockerfile updated to avoid using [deprecated](https://github.com/docker-library/official-images/issues/2076) "-onbuild" image and to use Pipenv to install pinned requirements.
- Travis build now uses Pipenv-created environment for running each script.
- Security check provided by Pipenv (and [safety](https://github.com/pyupio/safety)) added to Travis build (should fail on security issues found).
- Upgraded dependencies to latest versions (where possible).
- Evaluated new pylint and flake8 rules and applied changes across codebase.

See card [#1409](https://trello.com/c/9hQEiVPp) for more detail of some of the research done for this.

### How to review 
- Try out new installation steps listed in [README](https://github.com/ONSdigital/eq-survey-runner/blob/eq-1409-pip-dependencies/README.md)
- Review changes to [travis](https://github.com/ONSdigital/eq-survey-runner/blob/eq-1409-pip-dependencies/.travis.yml) and [Dockerfile](https://github.com/ONSdigital/eq-survey-runner/blob/eq-1409-pip-dependencies/Dockerfile)
- Review linting changes
- Discuss alternatives and/or new process


shahi645:
~some requirements missing from `requirements.in` @jcox-dev looking at what is used and not used~

Dockerfile
```diff
@@ -1,4 +1,16 @@
-FROM python:3.4-onbuild
+# https://github.com/docker-library/official-images/issues/2076
```
jonnyshaw89: What is this comment for?

Dockerfile
```diff
@@ -1,4 +1,16 @@
...
+RUN mkdir -p /usr/src/app
+WORKDIR /usr/src/app
+
+COPY Pipfile Pipfile
+COPY Pipfile.lock Pipfile.lock
+COPY . /usr/src/app
+
+RUN pipenv install --deploy --system
 
 ENV AWS_DEFAULT_REGION eu-west-1
```
jonnyshaw89: You should put this above the COPYs to improve layer caching

Dockerfile
```diff
@@ -1,4 +1,16 @@
...
+FROM python:3.4
+
+RUN pip install pipenv --upgrade
+
+RUN mkdir -p /usr/src/app
+WORKDIR /usr/src/app
+
+COPY Pipfile Pipfile
+COPY Pipfile.lock Pipfile.lock
+COPY . /usr/src/app
```
jonnyshaw89: If we put this at the bottom of the file it will be most efficient for layer caching

README.md
```diff
@@ -164,7 +145,7 @@ pyenv deactivate                 # to exit the virtual environment
 sudo pip install awsebcli        # install the eb cli tools
 ```
 
-Using the Elastic Beanstalk CLI is quite simple but mus tbe done *outside* the virtual environment of the project itself.  It will use the requirements.txt to ensure any requirements are at the right version for the deployed application.
```
jonnyshaw89: Do we know if ElasticBeanstalk works without a `requirements.txt`

app/authentication/roles.py
```diff
@@ -26,7 +26,7 @@ def role_required_decorator(func):
         @wraps(func)
         def role_required_wrapper(*args, **kwargs):
             metadata = get_metadata(current_user)
-            roles = metadata and metadata.get('roles', []) or []
+            roles = metadata and metadata.get('roles', []) or []  # pylint: disable=consider-using-ternary
```
jonnyshaw89: We should probably just fix this line?

Dockerfile
```diff
@@ -1,4 +1,16 @@
-FROM python:3.4-onbuild
+# https://github.com/docker-library/official-images/issues/2076
+FROM python:3.4
+
+RUN pip install pipenv --upgrade
+
+RUN mkdir -p /usr/src/app
+WORKDIR /usr/src/app
+
+COPY Pipfile Pipfile
```
ajmaddaford: Do we need to copy the `Pipfile` in for the pipenv install, or will it work with just the `Pipfile.lock`?

jonnyshaw89 (changes requested):
I think we should split out the linting changes to a separate PR?
It will make this PR a lot cleaner so we can see the specific changes

README.md
```diff
@@ -164,7 +145,7 @@ pyenv deactivate                 # to exit the virtual environment
...
+Using the Elastic Beanstalk CLI is quite simple but mus tbe done *outside* the virtual environment of the project itself.
```
ajmaddaford: Minor typo (existing) - `mus tbe`

ajmaddaford (changes requested):

README.md
ajmaddaford: I just checked and it doesn't 😞 
I think we should update the docs here with a line about generating one - https://docs.pipenv.org/advanced.html#generating-a-requirements-txt. We can update Jenkins to do the same.

Dockerfile
jcox-dev: @jonnyshaw89 mostly to draw attention to the deprecation issue. Can be removed now that it's been discussed.

README.md
jcox-dev: Could we generate a `requirements.txt` file as part of the travis build / provisioning?

app/authentication/roles.py
jcox-dev: Would be replaced with something like this to ensure list over None:

```
roles = (metadata.get('roles', []) or []) if metadata else []
```

Dockerfile
jcox-dev: I was using the [Pipenv Dockerfile](https://hub.docker.com/r/kennethreitz/pipenv/~/dockerfile/) as an example but I think that was only so the developer is able to play around with the full functionality. It's probably unnecessary for us.

jcox-dev: Turns out you do need at the very least a Pipfile otherwise Pipenv fails hard on Docker when it's unable to create one.

shahi645 (approved):
Happy that the pipenv changes work 👍🏿

README.md
```diff
@@ -161,10 +142,13 @@ You will need to install the EB CLI tools using PIP.
 
 ```
 pyenv deactivate                 # to exit the virtual environment
-sudo pip install awsebcli        # install the eb cli tools
+pip install --user awsebcli        # install the eb cli tools
```
ajmaddaford: Is the deactivate line needed as the virtualenv will only be used if the command is prefixed with `pipenv run` now?

README.md
```diff
@@ -306,10 +290,8 @@ will load the file into the interactive browser where it can be sorted and queri
 
 ## Updating / Installing dependencies
 
-We make use of python pip's support for only installing packages if their sha-256 hash matches a known good value.
-To add a new dependency, try installing the `hashin` python package that calculates a hash and adds it to the
-local `requirements.txt`. When installing the dependencies, always use `--require-hashes` to force this check
-on downloaded packages.
+We make use of Python pip's support for only installing packages if their sha-256 hash matches a known good value.
+To add a new dependency, use `pipenv install [package-name]`, which not only installs the package but Pipenv will also go to the trouble of updating the Pipfile as well.
```
ajmaddaford: It might be worth mentioning here that both the Pipfile and Pipfile.lock will need to be committed to source control.

jcox-dev:
@jonnyshaw89 @ajmaddaford I also think we should separate the linting changes from this PR. As discussed, we need to make sure this one has the old version of pylint pinned to the Pipfile and the latest in another PR.

jcox-dev:
Commits amended and dropped to keep PR relevant to Pipenv.

README.md
```diff
@@ -157,14 +142,16 @@ These options can be combined with `test_functional` or `test_functional_sauce`,
 
 You will need to install the EB CLI tools using PIP.
 
-*NOTE:* The EB tools do not currently work with Python 3.5.  I installed the EB CLI tools *outside* my virtual environment and installed installed them globally using the following commands
+*NOTE:* The EB tools do not currently work with Python 3.5.
```
ajmaddaford: We should remove this line as it looks like this is no longer the case - I can't find any reference to it in the AWS documentation. Looking at https://stackoverflow.com/questions/33801834/error-when-installing-awsebcli it looks like it used to have something in the documentation about it.

README.md
```diff
@@ -53,38 +53,23 @@ It is preferable to use the version of Python locally that matches that
...
-pyenv virtualenv <your env name>
-pyenv activate <your env name>
-```
 
 Upgrade pip and install dependencies:
 
 ```
-pip install --upgrade pip setuptools
-pip install -r requirements.txt
+pip install --upgrade pip setuptools pipenv pyenv
```
ajmaddaford: I don't think you can install pyenv via pip, the docs at https://github.com/pyenv/pyenv#installation recommend using brew to install it on OSX

README.md
```diff
@@ -53,38 +53,23 @@ It is preferable to use the version of Python locally that matches that
...
+pipenv install --dev
 ```
 
 If you need to run the tests:
 ```
 pip install -r requirements_for_test.txt
```
ajmaddaford: This isn't needed any more

tests/new_functional/spec/simultaneous-surveys.spec.js
```diff
@@ -17,7 +17,6 @@ describe('Given the user launches two surveys', function() {
 
     let launchSecondTab = browser
         .newWindow('/status', this.secondTab)
-        .setViewportSize({ width: 1024, height: 1158 })
```
ajmaddaford: I think this is from another merged pull request (and shouldn't be in this one)

docker-compose.yml
```diff
@@ -54,7 +54,7 @@ services:
       - "5000:5000"
 
   go-launch-a-survey:
-    image: onsdigital/go-launch-a-survey:load-schemas-from-register
+    image: onsdigital/go-launch-a-survey:latest
```
ajmaddaford: I think this is from another merged pull request (and shouldn't be in this one)

Pipfile
```diff
@@ -0,0 +1,56 @@
...
+Flask-Script = "*"
+Flask-SQLAlchemy = "*"
+"Flask-Themes2" = "*"
+Flask-WTF = "*"
+gevent = "*"
+gunicorn = "*"
+newrelic = "*"
+pika = "*"
+"psycopg2" = "*"
+PyYAML = "*"
```
ajmaddaford: pipenv seems to lowercase all packages and quote package names that contain numbers. It's probably worth starting with that as pipenv seems to update them all to match that when another package is added.

ajmaddaford (changes requested):

jcox-dev:
@ajmaddaford comments addressed in README and commits amended to drop the crept in changes from other branches.

ajmaddaford (approved):
Nice one! 👍

jonnyshaw89 (approved):

jonnyshaw89:
@jcox-dev can you squash this PR?

jcox-dev:
@jonnyshaw89 done!

PR #1308, opened 2017-10-19
Remove map method from answer store

jcox-dev:
From @ajmaddaford: "The Answer Store map method was originally created to output a WTForms POST data compatible format. This is then used to serialise the form data back into the WTForms form object. 

The map method has since been used in the summary context, accessing the answer values from this flattened list. Accessing the answers in this way is unclear and from https://github.com/ONSdigital/eq-survey-runner/pull/1158 will mean both the mapped answer values (used for answer values) and the answer store are being used."

Latest changes include adding method chaining to the answer store objects as well as reducing the overall complexity of building summary context.

### How to review 
* Complete survey and check summary page displays expected results
* Ensure changes are clear and obvious
* Check tests all pass across environments

### DIFF
--- a/app/data_model/answer_store.py
+++ b/app/data_model/answer_store.py
@@ -1,5 +1,5 @@
-from collections import OrderedDict
 import re
+from jinja2 import escape
 import simplejson as json
 
 
@@ -68,6 +68,12 @@ class AnswerStore(object):
     def __init__(self, existing_answers=None):
         self.answers = existing_answers or []
 
+    def __iter__(self):
+        return iter(self.answers)
+
+    def __getitem__(self, key):
+        return self.answers[key]
+
     @staticmethod
     def _validate(answer):
         if not isinstance(answer, Answer):
@@ -83,7 +89,7 @@ class AnswerStore(object):
 
         answer_to_add = answer.__dict__.copy()
 
-        if self.exists(answer):
+        if self.find(answer) is not None:
             raise ValueError('Answer instance already exists in store')
         else:
             self.answers.append(answer_to_add)
@@ -107,24 +113,10 @@ class AnswerStore(object):
 
         :param answer: An answer object.
         """
-        if self.exists(answer):
+        try:
             self.update(answer)
-        else:
+        except ValueError:
             self.add(answer)
-
-    def get(self, answer):
-        """
-        Returns the value of an answer.
-
-        :param answer: The ids for the answer to return
-        :return: The value of the answer found
-        """
-        position = self.find(answer)
-
-        if position is None:
-            raise ValueError('Answer instance does not exist in store')
-        else:
-            return self.answers[position]['value']
 
     def find(self, answer):
         """
@@ -141,30 +133,42 @@ class AnswerStore(object):
 
         return None
 
-    def exists(self, answer):
+    def count(self):
         """
-        Checks to see if an answer exists in the answer store.
+        Count of the number of answers in the answer store.
+        NB: can be combined with `filter` method to find count of an answer, e.g.:
 
-        :param answer: A dict of flattened answer details.
-        :return: True if the answer is in the store, False if not.
+            `answer_store.filter(answer_id=example_id).count()`
+
+        :return: Number of answers in this store.
         """
-        return self.find(answer) is not None
+        return len(self.answers)
 
-    def count(self, answer):
+    def values(self):
         """
-        Count of the number of instances of an answer in the answer store.
+        Return a flat list of all values in the answer store.
 
-        :param answer: A dict of flattened answer details.
-        :return: 0 if the answer doesn't exist, otherwise the number of instances.
+        :return: Return a list of answer values
         """
-        self._validate(answer)
+        return [answer['value'] for answer in self.answers]
 
-        return len(self.filter(answer.group_id, answer.block_id, answer.answer_id, answer.group_instance,
-                               answer.answer_instance))
+    def escaped(self):
+        """
+        Escape all answer values and return a new AnswerStore instance.
+
+        :return: Return a new AnswerStore object with escaped answers for chaining
+        """
+        escaped = []
+        for answer in self.answers:
+            answer = answer.copy()
+            if isinstance(answer['value'], str):
+                answer['value'] = escape(answer['value'])
+            escaped.append(answer)
+        return self.__class__(escaped)
 
     def filter_by_location(self, location):
         """
-        Find all answers in he answer store for a given location
+        Find all answers in the answer store for a given location
 
         :param location: The location to filter results by
         return: Return a list of answers which match the location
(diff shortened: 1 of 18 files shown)


app/templating/summary/question.py
```diff
@@ -1,70 +1,73 @@
...
+        for answer_schema in self.answers_iterator:
             if 'parent_answer_id' in answer_schema:
                 continue
             if question_schema['type'] == 'RepeatingAnswer':
-                for answer_id, answer_index in iterate_over_instance_ids(answers.keys()):
-                    if answer_schema['id'] == answer_id:
-                        key = answer_id if answer_index == 0 else '_'.join([answer_id, str(answer_index)])
-                        answer = cls._build_answer(question_schema, answer_schema, answers, answers_iterator, key)
-                        summary_answers.append(Answer(answer_schema, answer))
+                for answer_value in self._get_answers(answer_schema['id']):
```
ajmaddaford, line 36: The answer store already provides getting answers by id via the filter method - is there any reason to not use `answer_store.filter(answer_id=answer_schema['id'])`?

ajmaddaford (changes requested):

app/templating/summary/question.py
jcox-dev: @ajmaddaford keeping the helper methods in there for convenience but I agree that it was overkill rewriting a filter when one exists already.

app/data_model/answer_store.py
```diff
@@ -203,35 +204,18 @@ def filter(self, group_id=None, block_id=None, answer_id=None, group_instance=No
                 if v is not None:
                     matches = matches and answer[k] == v
             if matches:
-                filtered.append(answer)
+                if escape_values:
+                    answer = answer.copy()
+                    if isinstance(answer['value'], str):
+                        answer['value'] = escape(answer['value'])
+                filtered.append(answer['value'] if flat_values else answer)
```
ajmaddaford: I'm not sure `flat_values` makes sense in this method - if the filter is returning values for more than one answer id, won't it just be an array of values with no way of identifying which answer they are for?

ajmaddaford (changes requested):

app/data_model/answer_store.py
jcox-dev: @ajmaddaford true. It was more of a feature creep for convenience. It probably makes more sense to keep it specific to the helpers in Question.py while it's not required elsewhere.

dwyeradam (dismissed):

shahi645:
All changes work for me, as they should

jcox-dev:
@shahi645 @ajmaddaford @dwyeradam a lot of big changes to the answer store interface just pushed up.

shahi645:
Summary pages display as expected, and all tests pass. LGTM 👍🏿

app/data_model/answer_store.py
```diff
@@ -150,21 +156,45 @@ def exists(self, answer):
         """
         return self.find(answer) is not None
 
-    def count(self, answer):
+    def count(self, answer=None):
```
ajmaddaford: Do we ever call count with an answer? It's a bit odd as filtering by the answer instance would return you a count of 1, as there can only be one answer that will match both answer id and answer instance.
Also, now that count can be chained off filter it's probably clearer for people to do `answer_store.filter(..).count()` anyway.

app/data_model/answer_store.py
```diff
@@ -176,16 +206,15 @@ def filter(self, group_id=None, block_id=None, answer_id=None, group_instance=No
...
+        If no filter parameters are passed it returns a copy of the instance.
 
         :param answer_id: The answer id to filter results by
         :param block_id: The block id to filter results by
         :param group_id: The group id to filter results by
         :param answer_instance: The answer instance to filter results by
         :param group_instance: The group instance to filter results by
         :param limit: True | False Limit the number of answers returned
-        :return: Return a list of answers which satisfy the filter criteria
+        :return: Return a new AnswerStore object with filtered answers for chaining
```
ajmaddaford, line 191: Nice! 👍

app/helpers/form_helper.py
```diff
@@ -143,3 +146,25 @@ def clear_other_text_field(data, questions_for_block):
                 form_data[answer['id']] = ''
 
     return form_data
+
+
+def get_mapped_answers(answer_store, group_id=None, block_id=None, answer_id=None, group_instance=None, answer_instance=None):
+    """
+    Maps the answers in this store to a dictionary of key, value answers. Keys include instance
```
ajmaddaford: Now this method has moved, "in this store" no longer makes sense.

app/submitter/converter.py
```diff
@@ -135,12 +135,12 @@ def _get_checkbox_answer_data(answer_store, checkboxes_with_qcode, value):
...
                 filtered = answer_store.filter(answer_id=option['child_answer_id'])
 
-                if len(filtered) > 1:
+                if filtered.count() > 1:
                     raise Exception('Multiple answers found for {}'.format(option['child_answer_id']))
 
                 # if the user has selected 'other' we need to find the value it refers to.
                 # when non-mandatory, the other box value can be empty, in this case we just use its value
-                checkbox_answer_data[option['q_code']] = filtered[0]['value'] if len(filtered) > 1 else option['value']
+                checkbox_answer_data[option['q_code']] = filtered[0]['value'] if filtered.count() > 1 else option['value']
```
ajmaddaford: I can see this is existing code, but given we raise an exception on line 138 if `filtered.count() > 1`, should this just be `checkbox_answer_data[option['q_code']] = option['value']`?

app/templating/summary/question.py
```diff
@@ -1,70 +1,71 @@
...
-    def __init__(self, question_schema, answers_map, answer_store, metadata):
+    def __init__(self, question_schema, answer_store, metadata):
         self.id = question_schema['id']
         self.type = question_schema['type']
         self.skip_conditions = question_schema.get('skip_conditions')
         answer_schema = question_schema['answers']
         self.title = question_schema['title'] or answer_schema[0]['label']
         self.number = question_schema.get('number', None)
-        self.answers = self._build_answers(question_schema, answer_schema, answers_map)
+        self._answers = answer_store
```
ajmaddaford: I think this would be clearer if it was `_answer_store`

app/templating/summary/question.py
```diff
@@ -1,70 +1,71 @@
...
+        self.answers_iterator = iter(answer_schema)
```
ajmaddaford: - I think this should be `answer_schemas` rather than `answers_iterator`. This would capture what it is more clearly (schemas rather than answer store answers) and read better when used in `for answer_schema in answer_schemas`
- This should be prefixed with `_` as it's not exposed outside of this class

ajmaddaford (changes requested):

app/submitter/converter.py
jcox-dev: Nice spot! I was worried about some magical case when it would get past that exception.

ajmaddaford (approved):

dwyeradam (approved):
Much cleaner.

dwyeradam:
@jcox-dev  Squishy squashy

PR #1310, opened 2017-10-20
Add AWS CLI to docker container to get secrets on startup

ajmaddaford:
Adds the AWS CLI to the docker container so it can get the secrets from S3 on startup.

### How to review 
- Build and run the docker container
- Verify the `aws` command is working in the container e.g. `docker exec -ti eqsurveyrunner_eq-survey-runner_1 aws`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,7 @@
 FROM python:3.4
 
-RUN pip install -U pipenv==8.2.7
+RUN pip install -U pipenv==8.2.7 \
+  && pip install -U awscli==1.11.174
 
 RUN mkdir -p /usr/src/app
 WORKDIR /usr/src/app


Dockerfile
```diff
@@ -1,6 +1,7 @@
 FROM python:3.4
 
-RUN pip install -U pipenv==8.2.7
+RUN pip install -U pipenv==8.2.7 \
+  && pip install -U awscli==1.11.174
```
samiwel, line 4: Why are the versions specified in-line here rather than in a requirements.txt file?

jonnyshaw89: The aren't dependencies of the application. they are dependencies of the docker environment

jonnyshaw89 (approved):

Dockerfile
ajmaddaford: Because these are dependencies required to bootstrap survey runner - they aren't required to run survey runner.

samiwel: gotchya

MIslam92 (approved):

PR #1312, opened 2017-10-23
Remove numbering from Census

MIslam92:
Removed section numbering & removed question numbering
As this does not provide user value and is incorrect when routed

### How to review 
- Run survey runner and run though all Census surveys to check that no numbers in questions/sections are displayed.

### DIFF
--- a/data/en/census_communal.json
+++ b/data/en/census_communal.json
@@ -17,7 +17,6 @@
                 "questions": [{
                     "id": "establishment-type-question",
                     "title": "How would you describe your establishment?",
-                    "number": "1",
                     "type": "General",
                     "answers": [{
                             "id": "establishment-type-answer",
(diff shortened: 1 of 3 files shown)


jcox-dev (changes requested):
Couldn't see any numbering but is there a possibility for a test for this sort of thing? (As none were broken when these changes were introduced)

shahi645 (changes requested):
Need to remove numbering from welsh versions.

shahi645:
Also english version LGTM 👍🏿

MIslam92:
@jcox-dev this should not be a problem as there was no tests previously testing the numbers on a page and if there was it would specifically check for the content on the page. Which would be to specific to a survey which should not be the case.

shahi645 (approved):
Welsh changes implemented on translate_schemas script

jcox-dev (approved):

PR #1314, opened 2017-10-23
New_functional test folder to replace functional test folder

dwyeradam:
Now that all functional tests have been migrated the new_functional folder tests needs to become the only source for functional test scripts. Therefore the tests/new_functional folder is to replace the tests/functional folder. Any config file with reference to both files needs to be updated to only use the single recently replaced tests/functional folder.

### How to review 
All functional tests should pass.
Functional tests should still be able to be called on a single, suite or all level.

I have applied the config changes into the initial commit for easier reviews. The second commit is the folder move.


jcox-dev (approved):
Tests all pass. Spec arg still works. Suite arg still works. LGTM 👍

MIslam92 (approved):

MIslam92:
@dwyeradam "Squishy Squashy"

PR #1315, opened 2017-10-24
Remove virtual env warning from app script

jcox-dev:
virtualenv does not currently set the VIRTUAL_ENV environment variable when activate_this.py is used (https://github.com/pypa/virtualenv/pull/1057). pipenv run uses activate_this.py rather than one of the other activation scripts. We should also not care about reporting a lack of virtual environment now that we use Pipenv to manage it.

### DIFF
--- a/scripts/run_app.sh
+++ b/scripts/run_app.sh
@@ -1,11 +1,5 @@
 #!/bin/bash
 
-if [ -n "$VIRTUAL_ENV" ]; then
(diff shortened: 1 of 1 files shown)


MIslam92 (approved):
👍

dwyeradam (approved):

dwyeradam:
@jcox-dev  rebase please.

PR #1325, opened 2017-10-30
removed section headings from census household

MIslam92:
Removing section headings from census household 
Except for summary pages and printing answers after submission

### How to review 
Run SR and look though Census Household to check that all section headings are removed and run tests etc.


jonnyshaw89:
@MIslam92 might it be easier to just remove these form the census theme?

MIslam92:
@jonnyshaw89 good point but I'm only removing it from census household as off now (assume gonna remove from rest of them however not mentioned on card)

PR #1326, opened 2017-10-31
Removed census household section titles

MIslam92:
## **Context of this PR**

Removing section headings from census household
Except for summary pages and printing answers after submission

## **How to review**

Run SR and look though Census Household to check that all section headings are removed and run tests


jonnyshaw89:
@MIslam92 Did you look at just removing it from the theme?

MIslam92:
@jonnyshaw89 I did however if I removed it from the census theme it would remove it from household, communal & individual? While the card only requests it to be removed from household?

data/en/census_household.json
```diff
@@ -37,7 +37,7 @@
             "blocks": [{
                     "type": "Questionnaire",
                     "id": "permanent-or-family-home",
-                    "title": "Who lives here?",
+                    "title": "",
```
shahi645: Get rid of these titles, to match rest of schema changes?

shahi645 (changes requested):
Section headings still remain for individual sections but not on all questions. So need to remove of keep all. 
Also ` def complete_individual_section_person_1` failing in `test_census_household_submission_data`

tests/functional/spec/census-household.spec.js
```diff
@@ -157,7 +157,7 @@ describe('Census Household', function () {
 
   });
 
-  it('Given a census household survey, When one person in household, Then persons name should appear on member details pages.', function () {
+ /* it('Given a census household survey, When one person in household, Then persons name should appear on member details pages.', function () {
```
jonnyshaw89: If we remove this then this needs to be deleted now commented

data/schema/schema_v1.json
```diff
@@ -646,7 +646,6 @@
                     "additionalProperties": false,
                     "required": [
                       "id",
-                      "title",
```
jonnyshaw89: You also need to add validation to say that if it is specified then it can't be empty.

jonnyshaw89 (changes requested):

shahi645 (approved):
LGTM, all section titles gone and tests pass 👍🏿

jonnyshaw89 (approved):

jonnyshaw89:
@MIslam92 Squashy Squishy

PR #1329, opened 2017-11-01
Added a check/validation for figures being '<=' previous figure for MWSS

shahi645:
To add a check and validation to figure linked to a previous figure, which will make sure that said figure is less than or equal to the previous figure

### How to review 
- Launch MWSS

- make any figure linked to a previous figure  more than said figure

- check for an enhanced error message for validation errors

- run all tests

### DIFF
--- a/app/forms/fields.py
+++ b/app/forms/fields.py
@@ -193,7 +193,7 @@ def get_number_field(answer, label, guidance, error_messages, answer_store):
 
     validate_with = mandatory_or_optional + [
         NumberCheck(answer_errors['INVALID_NUMBER']),
-        NumberRange(minimum=min_value, maximum=max_value, messages=answer_errors),
+        NumberRange(minimum=min_value, maximum=max_value, messages=answer_errors, currency=answer.get('currency')),
         DecimalPlaces(max_decimals=max_decimals, messages=answer_errors),
     ]
 
--- a/app/utilities/schema.py
+++ b/app/utilities/schema.py
@@ -1,6 +1,6 @@
-import json
 import os
 import requests
+import simplejson as json
 
 from structlog import get_logger
 from werkzeug.exceptions import NotFound
@@ -42,7 +42,7 @@ def load_schema_file(schema_file, language_code=None):
 
     try:
         with open(schema_path, encoding='utf8') as json_data:
-            return json.load(json_data)
+            return json.load(json_data, use_decimal=True)
 
     except FileNotFoundError as e:
         logger.error('no schema file exists', filename=schema_path)
--- a/app/validation/validators.py
+++ b/app/validation/validators.py
@@ -3,7 +3,7 @@ from datetime import datetime
 
 from babel import numbers
 
-from app.jinja_filters import format_number
+from app.jinja_filters import format_number, format_currency
 from app.settings import DEFAULT_LOCALE
 from app.validation.error_messages import error_messages
 
@@ -68,24 +68,30 @@ class NumberRange(object):
         The maximum value of the number. If not provided, maximum value
         will not be checked.
     """
-    def __init__(self, minimum=None, maximum=None, messages=None):
+    def __init__(self, minimum=None, maximum=None, messages=None, currency=None):
         self.minimum = minimum
         self.maximum = maximum
         if not messages:
             messages = error_messages
         self.messages = messages
+        self.currency = currency
 
     def __call__(self, form, field):
         data = field.data
         error_message = None
         if data is not None:
             if self.minimum is not None and data < self.minimum:
-                error_message = self.messages['NUMBER_TOO_SMALL'] % dict(min=format_number(self.minimum))
+                error_message = self.messages['NUMBER_TOO_SMALL'] % dict(min=self.format_min_max(self.minimum))
             elif self.maximum is not None and data > self.maximum:
-                error_message = self.messages['NUMBER_TOO_LARGE'] % dict(max=format_number(self.maximum))
+                error_message = self.messages['NUMBER_TOO_LARGE'] % dict(max=self.format_min_max(self.maximum))
 
             if error_message:
                 raise validators.ValidationError(error_message)
+
+    def format_min_max(self, value):
+        if self.currency:
+            return format_currency(value, self.currency)
+        return format_number(value)
 
 
 class DecimalPlaces(object):
--- a/app/views/questionnaire.py
+++ b/app/views/questionnaire.py
@@ -1,5 +1,6 @@
 import re
 from collections import defaultdict
+import simplejson as json
 
 from flask import Blueprint, g, redirect, request, url_for, current_app
 from flask_login import current_user, login_required
@@ -258,12 +259,12 @@ def submit_answers(routing_path, eq_id, form_type, collection_id):
     answer_store = get_answer_store(current_user)
 
     if is_completed:
-        message = convert_answers(
+        message = json.dumps(convert_answers(
             metadata,
             g.schema_json,
             answer_store,
             routing_path,
-        )
+        ))
 
         encrypted_message = encrypt(message, current_app.eq['key_store'], KEY_PURPOSE_SUBMISSION)
         sent = current_app.eq['submitter'].send_message(
--- a/data/en/1_0005.json
+++ b/data/en/1_0005.json
@@ -433,7 +433,15 @@
                                 "q_code": "60",
                                 "type": "Currency",
                                 "currency": "GBP",
-                                "decimal_places": 2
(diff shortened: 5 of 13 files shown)


app/validation/validators.py
```diff
@@ -68,21 +68,28 @@ class NumberRange(object):
...
         self.messages = messages
+        self.format_type = format_type
 
     def __call__(self, form, field):
         data = field.data
         error_message = None
         if data is not None:
             if self.minimum is not None and data < self.minimum:
-                error_message = self.messages['NUMBER_TOO_SMALL'] % dict(min=format_number(self.minimum))
+                if self.format_type == 'Currency':
```
jonnyshaw89: it might be nicer to modify `format_number` to take in `format_type` and then do the formatting stitch to currency in their.
It would simplify this if conditions here a lot.

data/en/1_0005.json
```diff
@@ -433,7 +433,16 @@
...
                                 "type": "Currency",
                                 "currency": "GBP",
-                                "decimal_places": 2
+                                "decimal_places": 2,
+                                "max_value": {
+                                    "answer_id": "weekly-pay-gross-pay-answer"
+                                },
+                                "validation": {
+                                    "messages": {
+                                        "NUMBER_TOO_LARGE": "Enter a number less than or equal to £%(max)s."
```
jonnyshaw89: If we are formatting as currency, can we revert to using the default error messages?

jonnyshaw89 (dismissed):

data/en/1_0005.json
shahi645: @jonnyshaw89 but the default error messages don't contain the previous figure in them

jonnyshaw89: don't they?
https://github.com/ONSdigital/eq-survey-runner/blob/master/app/validation/error_messages.py#L10

MIslam92 (dismissed):
@shahi645 Address build failures

MIslam92 (approved):

app/forms/fields.py
```diff
@@ -219,7 +219,7 @@ def get_schema_defined_limit(answer_id, definition, answer_store):
             source_answer_id = definition.get('answer_id')
             answer_list = answer_store.filter(answer_id=source_answer_id)
             value = answer_list[0].get('value')
-            if not isinstance(value, int) and not isinstance(value, Decimal):
+            if not isinstance(value, int) and not isinstance(value, Decimal) and not isinstance(value, float):
```
dwyeradam: How are these float values now getting into the answer store? We should only have numbers of type integer or decimal.

dwyeradam (changes requested):
Do we also need to put in the bespoke error messages? For example "Enter a number less than or the same as the total number of employees (pipe in previous answer)"
https://trello.com/c/dGBfHvXj

app/validation/validators.py
```diff
@@ -68,25 +68,31 @@ class NumberRange(object):
...
             elif self.maximum is not None and data > self.maximum:
-                error_message = self.messages['NUMBER_TOO_LARGE'] % dict(max=format_number(self.maximum))
+                error_message = self.messages['NUMBER_TOO_LARGE'] % dict(max=self.format_min_max(self.maximum))
 
             if error_message:
                 raise validators.ValidationError(error_message)
 
+    def format_min_max(self, value):
+        if self.format_type == 'Currency':
+            return format_currency(value)
```
dwyeradam: Should we not also pass in the "currency" field to ensure the error message matches the answer asked? At the moment this will always default to GBP.

app/views/questionnaire.py
```diff
@@ -273,6 +274,7 @@ def submit_answers(routing_path, eq_id, form_type, collection_id):
             routing_path,
         )
 
+        message = json.dumps(message)
```
jonnyshaw89: I think that this should be done within `convert_answers` and `convert_answers` should now return a string.
Its not best practice to re-set a variable.

jonnyshaw89 (changes requested):

app/views/questionnaire.py
```diff
@@ -266,12 +267,12 @@ def submit_answers(routing_path, eq_id, form_type, collection_id):
...
 
     if is_completed:
-        message = convert_answers(
+        message = json.dumps(convert_answers(
             metadata,
             g.schema_json,
             answer_store,
             routing_path,
-        )
+        ))
```
ajmaddaford, line 267: We need to be happy that passing a JSON string rather than a Python object to encrypt does the same thing. The encrypt method calls the `JWTHelper.encode` method in sdc-cryptography JWTHelper, but that method doesn't have any tests. We should add a test there that tests a dict and a `json.dumps` of that dict produces the same payload.

JSON claims as a string is supported by JWCrypto, but there aren't any tests for it. I've raised an issue for this - https://github.com/latchset/jwcrypto/issues/96

ajmaddaford (changes requested):

app/views/questionnaire.py
shahi645: 👍 will add a test this morning

yoganandkunche (dismissed):
@shahi645 On the total gross <> pay question the validation message when respondent enters a negative value must be `Enter a number more than or equal to 0`

app/validation/validators.py
```diff
@@ -68,25 +68,31 @@ class NumberRange(object):
...
             elif self.maximum is not None and data > self.maximum:
-                error_message = self.messages['NUMBER_TOO_LARGE'] % dict(max=format_number(self.maximum))
+                error_message = self.messages['NUMBER_TOO_LARGE'] % dict(max=self.format_min_max(self.maximum))
 
             if error_message:
                 raise validators.ValidationError(error_message)
 
+    def format_min_max(self, value):
+        if self.currency:
+            return format_currency(value, self.currency)
```
ajmaddaford, line 93: What about percentages and units?

app/utilities/schema.py
```diff
@@ -42,7 +42,7 @@ def load_schema_file(schema_file, language_code=None):
 
     try:
         with open(schema_path, encoding='utf8') as json_data:
-            return json.load(json_data)
+            return json.load(json_data, use_decimal=True)
```
ajmaddaford, line 45: Have you done any testing with decimals saved using a master version of runner, and then resuming with this change?

tests/app/forms/test_questionnaire_form.py
```diff
@@ -178,7 +178,7 @@ def test_answer_errors_are_mapped(self):
 
             form.validate()
             answer_errors = form.answer_errors('total-retail-turnover-answer')
-            self.assertIn(error_messages["NUMBER_TOO_SMALL"] % dict(min=0), answer_errors)
+            self.assertIn(error_messages["NUMBER_TOO_SMALL"] % dict(min='£0.00'), answer_errors)
```
ajmaddaford: I think it would be clearer if we changed the answer to one that generates an error message that doesn't require interpolation of a value e.g. using `a` instead of `-1` would generate a `INVALID_NUMBER` error.

ajmaddaford (changes requested):

app/utilities/schema.py
shahi645: Yes this morning, behaves as expected 👍

app/validation/validators.py
shahi645: You're right ill add them in

tests/app/forms/test_questionnaire_form.py
shahi645: @ajmaddaford I think as the test is to map a previous answer to an error, i think using a default error message goes against the test itself. Maybe we should add a new test for this case?

app/validation/validators.py
shahi645: These are now going to be done in the next PR for the error messages

tests/app/forms/test_questionnaire_form.py
ajmaddaford: My bad, I mis-understood the test. 

Having said that I think it would be clearer if we changed the answer to be a number e.g. `total-number-employees` rather than a currency answer.

I also think we should rename the test, as it's not mapping any errors (some previous tests are testing the `map_errors` method). Something like `test_answer_errors_are_interpolated` would be clearer.

ajmaddaford (changes requested):

ajmaddaford (approved):

jonnyshaw89 (approved):

dwyeradam (approved):

dwyeradam:
@shahi645  Squishy Squashy.

PR #1330, opened 2017-11-01
Navigation completeness from routing path

jonnyshaw89:
This PR changes how Navigation sections are marked as completed by using the routing path to ensure that all blocks within the routing path for a navigation section are completed rather then just relying on the final block being complete.

This fixes issues with a scenario when their are multiple paths through a section and the route is modified which opens up new questions but the navigation does not then mark the section as incomplete.

This also removes the need to have a fixed end block to a section.

### How to review 
Test navigating around surveys opening and closing conditional routes and check that the section ticks on the navigation behave as expected.

### DIFF
--- a/app/questionnaire/navigation.py
+++ b/app/questionnaire/navigation.py
@@ -15,11 +15,11 @@ class Navigation(object):
     Reads navigation config from the schema and returns collections of dicts that describe
     completion status of each group
     """
-    def __init__(self, survey_json, answer_store, metadata=None, completed_blocks=None, routing_path=None):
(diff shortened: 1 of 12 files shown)


MIslam92 (approved):
Behaves as expected and tests pass however please make a conscious effort to follow eq practices and add the card number prior to your PR's i.e. eq-1419-navigation-completeness-from-routing-path as this has been mentioned in our tech sessions many times Thank you for your cooperation 🙃

shahi645 (approved):
@jonnyshaw89 need to update the go-launcher with the new navigation schema but otherwise LGTM 👍🏿
Squishy-squashy time

PR #1338, opened 2017-11-06
Eq 1458 enter address page census

shahi645:
Added an 'What is your address page' for respondents to enter address and have 'address line 1' piped into future survey questions. Address line 1 also mandatory

### How to review 
Launch census household leave address line 1 blank and verify that you get an error

### DIFF
--- a/data/en/census_household.json
+++ b/data/en/census_household.json
@@ -32,6 +32,74 @@
         ]
     },
     "groups": [{
+            "id": "what-is-your-address-group",
+            "title": "What is your address?",
+            "blocks": [{
+                "type": "Questionnaire",
+                "id": "what-is-your-address",
+                "questions": [{
+                    "id": "what-is-your-address-question",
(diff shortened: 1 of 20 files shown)


yoganandkunche:
@charlotte-hirst couple of questions for this PR. At the moment when census is launched app takes the respondent straight to the address question where address line 1 is mandatory but because the respondent have access to the navigation they can actually bypass the address question and still submit the survey and also there is no previous/section navigation to that question if the respondent chooses to change their address.

charlotte-hirst:
@yoganandkunche this is a temporary and will be used in the user testing and the user  researchers will be with them doing it so i think its ok

yoganandkunche (approved):
LGTM :+1:

MIslam92 (approved):

MIslam92:
@shahi645 "Squishy squashy"

PR #1345, opened 2017-11-07
Allow viewing of submitted answers

dwyeradam:
**As a** respondent, 
**I need** to access my responses to questions 
**so that** I can refer to them if my data is queried by ONS at a later date or used as a reference for next time I complete the same survey

### How to review 
Easiest to test using docker (`docker-compose up -d`) to ensure the dynamoDb container is created.

Use `test_submitted_responses.json` schema which allows answers to be viewed for 45 seconds.

### DIFF
--- a/app/data_model/session_store.py
+++ b/app/data_model/session_store.py
@@ -1,6 +1,8 @@
+import re
 from structlog import get_logger
 from app.data_model.models import EQSession, db
 from app.data_model.session_data import SessionData
+from app.storage.storage_encryption import StorageEncryption
 import simplejson as json
 
 logger = get_logger()
@@ -8,11 +10,13 @@ logger = get_logger()
 
 class SessionStore:
 
-    def __init__(self, eq_session_id=None):
+    def __init__(self, user_ik, pepper, eq_session_id=None):
         self.eq_session_id = eq_session_id
         self.user_id = None
+        self.user_ik = user_ik
         self.session_data = None
         self._eq_session = None
+        self.pepper = pepper
 
         if eq_session_id:
             self._load()
@@ -24,7 +28,7 @@ class SessionStore:
         self.eq_session_id = eq_session_id
         self.user_id = user_id
         self.session_data = session_data
-        self._eq_session = EQSession(self.eq_session_id, self.user_id, json.dumps(vars(self.session_data)))
+        self._eq_session = EQSession(self.eq_session_id, self.user_id)
 
         return self
 
@@ -33,7 +37,8 @@ class SessionStore:
         save session
         """
         if self._eq_session:
-            self._eq_session.session_data = json.dumps(vars(self.session_data))
+            self._eq_session.session_data = \
+                StorageEncryption(self.user_id, self.user_ik, self.pepper).encrypt_data(vars(self.session_data))
 
             # pylint: disable=maybe-no-member
             db.session.add(self._eq_session)
@@ -69,8 +74,13 @@ class SessionStore:
 
         if self._eq_session:
             self.user_id = self._eq_session.user_id
+
             if self._eq_session.session_data:
-                self.session_data = json.loads(self._eq_session.session_data, object_hook=lambda d: SessionData(**d))
+                session_data = self._eq_session.session_data
+                if self._is_session_data_encrypted(session_data):
+                    session_data = StorageEncryption(self.user_id, self.user_ik, self.pepper).decrypt_data(session_data)
+
+                self.session_data = json.loads(session_data, object_hook=lambda d: SessionData(**d))
 
             logger.debug('found matching eq_session for eq_session_id in database',
                          session_id=self._eq_session.eq_session_id,
@@ -79,3 +89,9 @@ class SessionStore:
             logger.debug('eq_session_id not found in database', eq_session_id=self.eq_session_id)
 
         return self._eq_session
+
+    @staticmethod
+    def _is_session_data_encrypted(session_data):
+        # pylint: disable=anomalous-backslash-in-string
+        encrypted_token_regex = '^[\w\-]+?\.([\w\-]+)?\.[\w\-]+\.[\w\-]+\.[\w\-]+?$'
+        return bool(re.match(encrypted_token_regex, session_data))
--- a/app/data_model/submitted_responses.py
+++ b/app/data_model/submitted_responses.py
@@ -1,0 +1,37 @@
+from datetime import datetime
+from botocore.exceptions import BotoCoreError, ConnectionError as BotoConnectionError
+from structlog import get_logger
+
+logger = get_logger()
+
+
+class SubmittedResponses(object):
+
+    def __init__(self, table):
+        self.table = table
+
+    def put_item(self, item):
+        """Insert an item to table"""
+        if item and item['valid_until']:
+            item['valid_until'] = int(item['valid_until'].strftime('%s'))  # convert to epoch
+
+        try:
+            response = self.table.put_item(Item=item)['ResponseMetadata']['HTTPStatusCode']
+            return response == 200
+
+        except (BotoCoreError, BotoConnectionError, ConnectionError):
+            logger.warn('Could not load data into DynamoDB')
+            return False
+
+    def get_item(self, tx_id):
+        """ Get an item given its key """
+        try:
+            response = self.table.get_item(Key={'tx_id': tx_id})
+            item = response.get('Item', None)
+            if item:
+                item['valid_until'] = datetime.utcfromtimestamp(item['valid_until'])
+            return item
+
+        except (BotoCoreError, BotoConnectionError, ConnectionError):
+            logger.warn('Could not get data from DynamoDB')
+            return None
--- a/app/jinja_filters.py
+++ b/app/jinja_filters.py
@@ -77,6 +77,11 @@ def format_month_year_date(value, date_format='%B %Y'):
 
 
 @blueprint.app_template_filter()
+def format_datetime(value, date_format='%d %B %Y at %H:%M'):
+    return "<span class='date'>{date}</span>".format(date=datetime.strptime(value, '%Y-%m-%dT%H:%M:%S.%f').strftime(date_format))
+
+
+@blueprint.app_template_filter()
 def format_conditional_date(date1=None, date2=None):
     """
     This function format_conditional_date accepts two dates, a user submitted date and a metadata date
--- a/app/libs/utils.py
+++ b/app/libs/utils.py
@@ -2,6 +2,11 @@ import uuid
 
 
 def convert_tx_id(tx_id):
+    # converts the guid tx_id to string of 16 characters with a dash between every 4 characters
+    return (tx_id[:4] + '-' + tx_id[4:])[:19]
+
+
+def convert_tx_id_for_boxes(tx_id):
     # converts the guid tx_id to string of 16 characters with a space between every 4 characters
     tx_id = uuid.UUID(tx_id)
     tx_id = tx_id.hex
--- a/app/settings.py
+++ b/app/settings.py
@@ -65,6 +65,11 @@ EQ_SERVER_SIDE_STORAGE_DATABASE_NAME = os.getenv('EQ_SERVER_SIDE_STORAGE_DATABAS
 EQ_SERVER_SIDE_STORAGE_USER_ID_ITERATIONS = ensure_min(os.getenv('EQ_SERVER_SIDE_STORAGE_USER_ID_ITERATIONS', 10000),
                                                        1000)
 
+EQ_DYNAMODB_ENABLED = parse_mode(os.getenv('EQ_DYNAMODB_ENABLED', 'True'))
+if EQ_DYNAMODB_ENABLED:
+    EQ_DYNAMODB_ENDPOINT = os.getenv('EQ_DYNAMODB_ENDPOINT')
+    EQ_SUBMITTED_RESPONSES_TABLE_NAME = get_env_or_fail('EQ_SUBMITTED_RESPONSES_TABLE_NAME')
+
 RESPONDENT_ACCOUNT_URL = os.getenv('RESPONDENT_ACCOUNT_URL', 'https://survey.ons.gov.uk/')
 
 EQ_DEV_MODE = parse_mode(os.getenv('EQ_DEV_MODE', 'False'))
--- a/app/storage/encrypted_questionnaire_storage.py
+++ b/app/storage/encrypted_questionnaire_storage.py
@@ -1,14 +1,9 @@
-import hashlib
 import simplejson as json
 
-from jwcrypto import jwe, jwk
-from jwcrypto.common import base64url_encode, base64url_decode
 from structlog import get_logger
 
+from app.storage.storage_encryption import StorageEncryption
 from app.data_model.models import QuestionnaireState, db
-from app.utilities.strings import to_bytes
-from app.utilities.strings import to_str
-
 logger = get_logger()
 
 
@@ -17,15 +12,12 @@ class EncryptedQuestionnaireStorage:
     def __init__(self, user_id, user_ik, pepper):
         if user_id is None:
             raise ValueError('User id must be set')
-        if user_ik is None:
-            raise ValueError('User ik must be set')
-        if pepper is None:
-            raise ValueError('Pepper must be set')
+
         self._user_id = user_id
-        self._cek = self._generate_key(user_id, user_ik, pepper)
+        self.encrypter = StorageEncryption(user_id, user_ik, pepper)
 
     def add_or_update(self, data, version):
-        encrypted_data = self._encrypt_data(data)
+        encrypted_data = self.encrypter.encrypt_data(data)
         encrypted_data_json = json.dumps({'data': encrypted_data})
         questionnaire_state = self._find_questionnaire_state()
         if questionnaire_state:
@@ -48,7 +40,7 @@ class EncryptedQuestionnaireStorage:
             version = questionnaire_state.version or 0
 
             if 'data' in data:
-                decrypted_data = self._decrypt_data(data['data'])
+                decrypted_data = self.encrypter.decrypt_data(data['data'])
                 return decrypted_data, version
 
         return None, None
@@ -62,50 +54,8 @@ class EncryptedQuestionnaireStorage:
             db.session.delete(questionnaire_state)
             db.session.commit()
 
-    @staticmethod
-    def _generate_key(user_id, user_ik, pepper):
-        sha256 = hashlib.sha256()
-        sha256.update(to_str(user_id).encode('utf-8'))
-        sha256.update(to_str(user_ik).encode('utf-8'))
-        sha256.update(to_str(pepper).encode('utf-8'))
-
-        # we only need the first 32 characters for the CEK
-        return to_bytes(sha256.hexdigest()[:32])
-
-    def _encrypt_data(self, data):
-        protected_header = {
-            'alg': 'dir',
-            'enc': 'A256GCM',
-            'kid': '1,1',
-        }
-        jwe_token = jwe.JWE(plaintext=base64url_encode(data), protected=protected_header)
-
-        key = self.generate_jwk_from_cek(self._cek)
-
-        jwe_token.add_recipient(key)
-
-        return jwe_token.serialize(compact=True)
-
-    def _decrypt_data(self, encrypted_token):
-
-        key = self.generate_jwk_from_cek(self._cek)
-
-        jwe_token = jwe.JWE(algs=['dir', 'A256GCM'])
-        jwe_token.deserialize(encrypted_token, key)
-
-        return base64url_decode(jwe_token.payload.decode()).decode()
-
     def _find_questionnaire_state(self):
         logger.debug('getting questionnaire data', user_id=self._user_id)
         # pylint: disable=maybe-no-member
         # SQLAlchemy doing declarative magic which makes session scope query property available
         return QuestionnaireState.query.filter(QuestionnaireState.user_id == self._user_id).first()
-
-    @staticmethod
-    def generate_jwk_from_cek(cek):
-        password = {
-            'kty': 'oct',
-            'k': base64url_encode(cek),
-        }
-
-        return jwk.JWK(**password)
--- a/app/storage/storage_encryption.py
+++ b/app/storage/storage_encryption.py
@@ -1,0 +1,64 @@
+import hashlib
+
+from jwcrypto import jwe, jwk
+from jwcrypto.common import base64url_encode, base64url_decode
+from structlog import get_logger
+
+from app.utilities.strings import to_bytes
+from app.utilities.strings import to_str
+import simplejson as json
+
+logger = get_logger()
+
+
+class StorageEncryption:
+
+    def __init__(self, user_id, user_ik, pepper):
+        if user_id is None:
+            raise ValueError('User id must be set')
+        if user_ik is None:
+            raise ValueError('User ik must be set')
+        if pepper is None:
+            raise ValueError('Pepper must be set')
+
+        self.key = self._generate_key(user_id, user_ik, pepper)
+
+    @staticmethod
+    def _generate_key(user_id, user_ik, pepper):
+        sha256 = hashlib.sha256()
+        sha256.update(to_str(user_id).encode('utf-8'))
+        sha256.update(to_str(user_ik).encode('utf-8'))
+        sha256.update(to_str(pepper).encode('utf-8'))
+
+        # we only need the first 32 characters for the CEK
+        cek = to_bytes(sha256.hexdigest()[:32])
+
+        password = {
+            'kty': 'oct',
+            'k': base64url_encode(cek),
+        }
+
+        return jwk.JWK(**password)
+
+    def encrypt_data(self, data):
+        if not isinstance(data, str):
+            data = json.dumps(data)
+
+        protected_header = {
+            'alg': 'dir',
+            'enc': 'A256GCM',
+            'kid': '1,1',
+        }
+
+        jwe_token = jwe.JWE(plaintext=base64url_encode(data), protected=protected_header)
+
+        jwe_token.add_recipient(self.key)
+
+        return jwe_token.serialize(compact=True)
+
+    def decrypt_data(self, encrypted_token):
+
+        jwe_token = jwe.JWE(algs=['dir', 'A256GCM'])
+        jwe_token.deserialize(encrypted_token, self.key)
+
+        return base64url_decode(jwe_token.payload.decode()).decode()
--- a/app/templates/summary.html
+++ b/app/templates/summary.html
@@ -4,8 +4,6 @@
 
 {% block page_title -%}{{_("Summary")}} - {{survey_title}}{% endblock -%}
 
-{% set not_answered = 'No answer provided' %}
-
 {% block sidebar %}
   {% if navigation %}
     {% include theme('partials/navigation.html') %}
@@ -20,65 +18,22 @@
 {% block inline_feedback %}{{ super() }}{% endblock %}
 
 <div>
-  <h1 class="saturn">Your responses</h1>
+  <h1 class="saturn">Your answers</h1>
   <div class="print__message panel panel--simple panel--error">
-    <h2 class="saturn">Please remember to submit these responses.</h2>
+    <h2 class="saturn">Please remember to submit these answers.</h2>
   </div>
   <div class="print__hidden panel panel--simple panel--warn">
-    <h2 class="neptune">Please check your responses carefully before submitting.</h2>
+    <h2 class="neptune">Please check your answers carefully before submitting.</h2>
   </div>
 </div>
 
-<div>
-  <div class="summary">
+{% include theme('partials/summary/summary.html') %}
 
-  {% for group in content.summary %}
-
-      <h2 class="summary__title saturn" id="{{group.id}}">{{group.title}}</h2>
-
-      <div class="summary__block">
-
-     {% for block in group.blocks %}
-
-        <div class="summary__items">
-
-            {% for question in block.questions %}
-
-                <div class="summary__question" id="{{question.id}}">
-                  {{question.number or ''}} {{question.title|striptags}}
-                </div>
-
-                {% for answer in question.answers %}
-                      {% if question.answers|length > 1 and answer.label %}
-                        <div class="summary__question summary__question--sub">
-                          {{answer.label|striptags}}
-                        </div>
-                      {% endif %}
-
-                      <div class="summary__answer">
-                        <div class="summary__answer-text" id="{{answer.id}}-answer" data-qa="{{answer.id}}-answer">
-                          {%- if answer.value is none or answer.value == '' -%}
-                            {{not_answered}}
-                          {%- else -%}
-                            {% include theme(['partials/summary/' ~ answer.type ~ '.html', 'partials/summary/default.html']) %}
-                          {%- endif -%}
-                        </div>
-                        <div class="summary__edit">
-                          <a href="{{ block.link }}#{{answer.id}}" class="summary__edit-link" aria-describedby="{{answer.id}} {{answer.id}}-answer" data-qa="{{answer.id}}-edit" {{helpers.track('click', 'Summary', 'Edit click')}}>Edit <span class="u-vh">your answer</span></a>
-                        </div>
-                      </div>
-                {% endfor %}
-
-            {% endfor %}
-
-        </div>
-
-     {% endfor %}
-
-     </div>
-  {% endfor %}
+{% if content.is_view_submission_response_enabled %}
+  <div class="u-mb-s u-mb-m@s">
+    <div class="venus">You will have the opportunity to view and print a copy of your answers after submitting this questionnaire</div>
   </div>
-</div>
+{% endif %}
 
 {{super()}}
 
--- a/app/templates/thank-you.html
+++ b/app/templates/thank-you.html
@@ -6,25 +6,41 @@
 
 <div class="u-mb-s">
     <h1 class="saturn">Submission Successful</h1>
-    <p>Thank you for submitting the {{ survey_title }}, your responses have been successfully received.</p>
 </div>
 
-<div class="panel panel--simple panel--success panel--spacious">
-    <p class="mars">Transaction ID (Please quote this reference for any support queries)</p>
-    {% if meta.respondent %}
-    <ul class="list list--boxes">
-        {%for id_part in meta.respondent.tx_id%}<li class="list__item">{{id_part}}</li>{% endfor %}
-    </ul>
+<div class="u-mb-s">
+    <div class="panel panel--simple panel--success panel--spacious">
+        <p class="mars">Your answers were submitted for <b>{{meta.respondent.ru_name}}</b> on {{meta.survey.submitted_time|format_datetime}}</p>
+        <p class="mars">Transaction ID: <b>{{meta.respondent.tx_id}}</b></p>
+    </div>
+</div>
+
+<div class="u-mb-s">
+    <p>Your answers will be processed in the next few weeks.
+        We may contact you to query your answers via phone or secure message.</p>
+</div>
+
+{% block view_submission %}
+{% if is_view_submitted_response_enabled %}
+    <div class="u-mb-s">
+    {% if view_submission_url %}
+        <p class="u-mb-m@s mars" data-qa="view-submission"><a href={{view_submission_url}}>View and print a copy of your answers</a></p>
+        <p class="u-mb-s mars">Your answers will only be available for <b>{{view_submission_duration}}</b> to keep them secure.</p>
+    {% else %}
+        <div class="panel panel--simple panel--info" data-qa="view-submission-expired">
+            <div class="panel__body">
+                <p>You are no longer able to view and print your answers.</p>
+                <p>The link expires to maintain the security of your information.</p>
+            </div>
+        </div>
     {% endif %}
-    <p class="u-mb-no mars">You may wish to save or print this page for your records.</p>
-</div>
+    </div>
+{% endif %}
+{% endblock %}
 
-{{answer_guidance}}
-
-<div class="u-mt-s">
+<div class="u-mb-s">
     <div class="mars">For more information on how we use this data.</div>
     <p class="u-mb-m@s mars"><a href="https://www.ons.gov.uk/surveys">https://www.ons.gov.uk/surveys</a></p>
-    <p class="mars">It is now safe to close this window.</p>
 </div>
 
 {% endblock %}
--- a/app/templates/view-submission.html
+++ b/app/templates/view-submission.html
@@ -1,0 +1,23 @@
+{% extends theme('layouts/_twocol.html') %}
+{% import 'macros/helpers.html' as helpers %}
+
+{% block page_title -%}{{_("Submission")}} - {{survey_title}}{% endblock -%}
+
+{% block main -%}
+
+<div>
+  <h1 class="saturn">Submitted answers</h1>
+</div>
+
+<div class="u-mb-s">
+    <div class="panel panel--simple panel--success panel--spacious">
+        <p class="mars">Your answers were submitted for <b>{{meta.respondent.ru_name}}</b> on {{meta.survey.submitted_time|format_datetime}}</p>
+        <p class="mars">Transaction ID: <b>{{meta.respondent.tx_id}}</b></p>
+    </div>
+</div>
+
+<button class="btn btn--tertiary print__hidden u-no-js-hide" onclick="window.print()">Print this page <span class="btn__icon btn__icon--print-link"></span></button>
+
+{% include theme('partials/summary/summary.html') %}
+
+{% endblock -%}
--- a/app/templating/metadata_context.py
+++ b/app/templating/metadata_context.py
@@ -47,12 +47,15 @@ def _build_survey_meta(metadata):
     }
 
 
-def build_metadata_context_for_survey_completed(survey_completed_metadata):
+def build_metadata_context_for_survey_completed(session_data):
     return {
         'survey': {
-            'period_str': survey_completed_metadata['period_str'],
+            'period_str': session_data.period_str,
+            'submitted_time': session_data.submitted_time
         },
         'respondent': {
-            'tx_id': convert_tx_id(survey_completed_metadata['tx_id']),
+            'tx_id': convert_tx_id(session_data.tx_id),
+            'ru_name': session_data.ru_name,
+            'ru_ref': session_data.ru_ref,
         },
     }
--- a/app/views/questionnaire.py
+++ b/app/views/questionnaire.py
@@ -1,7 +1,8 @@
 import re
 from collections import defaultdict
-from datetime import datetime
+from datetime import datetime, timedelta
 import simplejson as json
+import humanize as humanize
 
 from flask import Blueprint, g, redirect, request, url_for, current_app
 from flask_login import current_user, login_required, logout_user
@@ -11,16 +12,19 @@ from flask_wtf import FlaskForm
 
 from structlog import get_logger
 
-from app.data_model.answer_store import Answer
-from app.globals import get_answer_store, get_completed_blocks, get_metadata, get_questionnaire_store, get_session_store
+from app.globals import get_session_store
+from app.data_model.answer_store import Answer, AnswerStore
+from app.globals import get_answer_store, get_completed_blocks, get_metadata, get_questionnaire_store
 from app.helpers.form_helper import get_form_for_location, post_form_for_location
 from app.helpers.path_finder_helper import path_finder, full_routing_path_required
 from app.helpers import template_helper
 from app.questionnaire.location import Location
 from app.questionnaire.navigation import Navigation
+from app.questionnaire.path_finder import PathFinder
 
 from app.questionnaire.rules import evaluate_skip_conditions
 from app.keys import KEY_PURPOSE_SUBMISSION
+from app.storage.storage_encryption import StorageEncryption
 from app.submitter.converter import convert_answers
 from app.submitter.submission_failed import SubmissionFailedException
 from app.templating.metadata_context import build_metadata_context_for_survey_completed
@@ -208,20 +212,70 @@ def get_thank_you(eq_id, form_type):  # pylint: disable=unused-argument
 
     if session_data.submitted_time:
         schema = load_schema_from_params(session_data.eq_id, session_data.form_type)
-        metadata_context = build_metadata_context_for_survey_completed(vars(session_data))
-        thank_you_template = render_theme_template(schema.json['theme'],
-                                                   template_name='thank-you.html',
-                                                   meta=metadata_context,
-                                                   analytics_ua_id=current_app.config['EQ_UA_ID'],
-                                                   survey_id=schema.json['survey_id'],
-                                                   survey_title=TemplateRenderer.safe_content(schema.json['title']))
-        return thank_you_template
+        metadata_context = build_metadata_context_for_survey_completed(session_data)
+
+        view_submission_url = None
+        view_submission_duration = 0
+        if _is_submission_viewable(schema.json, session_data.submitted_time):
+            view_submission_url = url_for('.get_view_submission', eq_id=eq_id, form_type=form_type)
+            view_submission_duration = humanize.naturaldelta(timedelta(seconds=schema.json['view_submitted_response']['duration']))
+
+        return render_theme_template(schema.json['theme'],
+                                     template_name='thank-you.html',
+                                     meta=metadata_context,
+                                     analytics_ua_id=current_app.config['EQ_UA_ID'],
+                                     survey_id=schema.json['survey_id'],
+                                     survey_title=TemplateRenderer.safe_content(schema.json['title']),
+                                     is_view_submitted_response_enabled=is_view_submitted_response_enabled(schema.json),
+                                     view_submission_url=view_submission_url,
+                                     view_submission_duration=view_submission_duration)
 
     metadata = get_metadata(current_user)
     g.schema = load_schema_from_metadata(metadata)
     routing_path = path_finder.get_full_routing_path()
     collection_id = metadata['collection_exercise_sid']
     return _redirect_to_latest_location(routing_path, collection_id, metadata.get('eq_id'), metadata.get('form_type'))
+
+
+@post_submission_blueprint.route('view-submission', methods=['GET'])
+@login_required
+def get_view_submission(eq_id, form_type):  # pylint: disable=unused-argument
+
+    session_data = get_session_store().session_data
+    g.schema = load_schema_from_params(session_data.eq_id, session_data.form_type)
+
+    if _is_submission_viewable(g.schema.json, session_data.submitted_time):
+        submitted_data = current_app.eq['submitted_responses'].get_item(session_data.tx_id)
+
+        if submitted_data:
+
+            metadata_context = build_metadata_context_for_survey_completed(session_data)
+
+            pepper = current_app.eq['secret_store'].get_secret_by_name('EQ_SERVER_SIDE_STORAGE_ENCRYPTION_USER_PEPPER')
+            encrypter = StorageEncryption(current_user.user_id, current_user.user_ik, pepper)
+
+            submitted_data = json.loads(encrypter.decrypt_data(submitted_data['data']))
+            answer_store = AnswerStore(existing_answers=submitted_data.get('answers'))
+
+            metadata = submitted_data.get('metadata')
+
+            routing_path = PathFinder(g.schema, answer_store, metadata).get_full_routing_path()
+
+            schema_context = _get_schema_context(routing_path, 0, metadata, answer_store)
+            rendered_schema = renderer.render(g.schema.json, **schema_context)
+            summary_rendered_context = {'summary': build_summary_rendering_context(g.schema, rendered_schema, answer_store, metadata),
+                                        'variables': None,
+                                        'answers_are_editable': False}
+
+            return render_theme_template(g.schema.json['theme'],
+                                         template_name='view-submission.html',
+                                         meta=metadata_context,
+                                         analytics_ua_id=current_app.config['EQ_UA_ID'],
+                                         survey_id=g.schema.json['survey_id'],
+                                         survey_title=TemplateRenderer.safe_content(g.schema.json['title']),
+                                         content=summary_rendered_context)
+
+    return redirect(url_for('post_submission.get_thank_you', eq_id=eq_id, form_type=form_type))
 
 
 def _redirect_to_latest_location(routing_path, collection_id, eq_id, form_type):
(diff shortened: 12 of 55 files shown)


app/data_model/submitted_responses.py
```diff
@@ -0,0 +1,83 @@
...
+from structlog import get_logger
+
+logger = get_logger()
+
+
+class SubmittedResponses(object):
+
+    def __init__(self, table_name='submitted-responses'):
+        if EQ_ENVIRONMENT_PREFIX == 'dev':
+            self.dynamoDB = boto3.resource('dynamodb', endpoint_url='http://localhost:6060')
```
jonnyshaw89: I think we should explicitly pass in the endpoint and call this line only if one is passed in rather then switching on a hard coded value of 'dev'

app/data_model/submitted_responses.py
```diff
@@ -0,0 +1,83 @@
...
+        else:
+            self.dynamoDB = boto3.resource('dynamodb')
+
+        self.tableName = EQ_ENVIRONMENT_PREFIX + '-' + table_name
+
+    def insert_item(self, tx_id, answers, metadata, submitted, time_to_live=EQ_SUBMISSION_VALIDITY):
+        """Insert an item to table"""
+
+        if not time_to_live or time_to_live > EQ_SUBMISSION_VALIDITY:
```
jonnyshaw89: Can `time_to_live` be none of we default it in the method?

app/data_model/submitted_responses.py
```diff
@@ -0,0 +1,83 @@
...
+        try:
+            table = self.dynamoDB.Table(self.tableName)
+            response = table.put_item(Item=item)['ResponseMetadata']['HTTPStatusCode']
+            if response == 200:
+                return True
+            else:
+                raise ClientError('Could not load data into DynamoDB table', 'put_item')
+        except ClientError:
+            logger.bind(tx_id=tx_id)
+            logger.info('submitted_responses dynamoDB', 'Could not load data into DynamoDB table:')
```
jonnyshaw89: This log line contains a trailing `:`

app/data_model/submitted_responses.py
```diff
@@ -0,0 +1,83 @@
...
+
+        item = {
+            'tx_id': tx_id,
+            'answers': answers,
+            'metadata': metadata,
+            'valid_until': valid_until
+        }
+
+        try:
+            table = self.dynamoDB.Table(self.tableName)
```
jonnyshaw89: you could set `self.table` in the constructor instead of `self.tableName`

app/data_model/submitted_responses.py
```diff
@@ -0,0 +1,83 @@
...
+    def get_item(self, tx_id):
+        """
+        Get an item given its key
+        """
+        try:
+            table = self.dynamoDB.Table(self.tableName)
+            response = table.get_item(Key={'tx_id': tx_id})
+            return response['Item']
+        except ClientError:
+            return None
```
jonnyshaw89: If the response is empty we should return None. If we get an exception we should log it.

app/data_model/submitted_responses.py
```diff
@@ -0,0 +1,83 @@
...
+                Key={
+                    'tx_id': tx_id
+                },
+                AttributesToGet=[
+                    'valid_until'
+                ]
+            )
+            return response['Item']['valid_until']
+        except (KeyError, ClientError):
+            return 0
```
jonnyshaw89: same as above

app/libs/utils.py
```diff
@@ -2,6 +2,11 @@
 
 
 def convert_tx_id(tx_id):
+    # converts the guid tx_id to string of 16 characters with a dash between every 4 characters
+    return (tx_id[:4] + '-' + tx_id[4:])[:19]
+
+
+def convert_tx_id_for_boxes(tx_id):
```
jonnyshaw89, line 9: Do we use the boxes anymore? if not we should remove this.

app/views/questionnaire.py
```diff
@@ -260,7 +312,7 @@ def _is_skipping_to_the_end(routing_path, block, current_location):
         block['type'] in END_BLOCKS
 
 
-def submit_answers(routing_path, eq_id, form_type, collection_id):
+def submit_answers(routing_path, eq_id, form_type, collection_id, block_id):
```
jonnyshaw89: rather than pass the `block_id` into this method, we have the `g.schema_json` so just lookup if it has a summary

app/views/questionnaire.py
```diff
@@ -538,8 +603,11 @@ def _get_block_json(current_location):
     return _evaluate_skip_conditions(block_json, current_location, answer_store, metadata)
 
 
-def _get_schema_context(full_routing_path, current_location, metadata, answer_store):
-    aliases = SchemaHelper.get_aliases(g.schema_json)
+def _get_schema_context(full_routing_path, current_location, metadata, answer_store, schema=None):
+    if schema:
```
jonnyshaw89: Im not sure why we have done this?

jonnyshaw89 (changes requested):

app/libs/utils.py
dwyeradam: Used in an error page but i'll need to check what context that error page would get used.

app/views/questionnaire.py
dwyeradam: Going to refactor the way we define if we show the submitted responses or not. This will come out as part of that. In future we'll define it within an object in the schema so it won't be dependent on having a summary page.

app/views/questionnaire.py
dwyeradam: You're right, it's no longer required. It was left in from a time when the schema wasn't available after submission. That however was fixed.

dwyeradam:
Still need to add tests around this but ready for a high level review. I'd also like to discuss issues around security of answers.

app/libs/utils.py
jonnyshaw89: It might been good to make the error page show the tx_id in a consistent way to the thank you?

app/data_model/submitted_responses.py
```diff
@@ -0,0 +1,81 @@
...
+from app.settings import EQ_ENVIRONMENT_PREFIX, EQ_SUBMISSION_VALIDITY, EQ_DYNAMODB_ENDPOINT
+from structlog import get_logger
+
+logger = get_logger()
+
+
+class SubmittedResponses(object):
+
+    def __init__(self):
+        if EQ_DYNAMODB_ENDPOINT:
```
jonnyshaw89: if `EQ_DYNAMODB_ENDPOINT` defaults to None, what happens if you pass None into the `endpoint_url`?
ie, do you need the if/else?

app/data_model/submitted_responses.py
```diff
@@ -0,0 +1,81 @@
...
+        }
+
+        try:
+            response = self.table.put_item(Item=item)['ResponseMetadata']['HTTPStatusCode']
+            if response == 200:
+                return True
+            else:
+                raise BotoConnectionError()
+        except (BotoCoreError, BotoConnectionError):
+            logger.bind(tx_id=tx_id)
```
jonnyshaw89: Is `tx_id` not already bound? if not we should do it towards the beginning of the request

app/data_model/submitted_responses.py
```diff
@@ -0,0 +1,81 @@
...
+        return {k: v for k, v in ((k, self.clean_empty(v)) for k, v in d.items()) if v}
+
+    def get_item(self, tx_id):
+        """
+        Get an item given its key
+        """
+        try:
+            response = self.table.get_item(Key={'tx_id': tx_id})
+            return response['Item']
+        except (BotoCoreError, BotoConnectionError, KeyError):
```
jonnyshaw89: We should also log these exceptions?

app/data_model/submitted_responses.py
```diff
@@ -0,0 +1,81 @@
...
+            return [v for v in (self.clean_empty(v) for v in d) if v]
+        return {k: v for k, v in ((k, self.clean_empty(v)) for k, v in d.items()) if v}
+
+    def get_item(self, tx_id):
+        """
+        Get an item given its key
+        """
+        try:
+            response = self.table.get_item(Key={'tx_id': tx_id})
+            return response['Item']
```
jonnyshaw89: If there are no results does the 'Item' key exist?

app/views/questionnaire.py
```diff
@@ -172,22 +181,59 @@ def post_household_composition(routing_path, **kwargs):
 @questionnaire_blueprint.route('thank-you', methods=['GET'])
 def get_thank_you(eq_id, form_type, collection_id):  # pylint: disable=unused-argument
     survey_completed_metadata = current_app.eq['session_storage'].get_survey_completed_metadata()
-    schema = load_schema_from_params(eq_id, form_type)
 
     if survey_completed_metadata:
+        view_submission = None
+        tx_id = survey_completed_metadata['tx_id']
```
jonnyshaw89: Put this inside the If, as we dont need it outside

app/views/questionnaire.py
```diff
@@ -273,6 +319,7 @@ def submit_answers(routing_path, eq_id, form_type, collection_id):
             routing_path,
         )
 
+        message = json.dumps(message)
```
jonnyshaw89: why are we doing this? does it change the structure of what we encrypt?

app/views/questionnaire.py
```diff
@@ -300,11 +360,19 @@ def submit_answers(routing_path, eq_id, form_type, collection_id):
         return redirect(invalid_location.url(metadata))
 
 
+def store_submitted_responses():
+    if g.schema_json.get('view_submitted_responses'):
+        if g.schema_json['view_submitted_responses'].get('enabled'):
```
jonnyshaw89: what if enabled is set to `False`?

jonnyshaw89: Assuming it returns a boolean then you can just return this line? rather than it being an if statement.

docker-compose.yml
```diff
@@ -44,6 +44,8 @@ services:
       EQ_RABBITMQ_HOST_SECONDARY: rabbit
       EQ_SECRETS_FILE: docker-secrets.yml
       EQ_KEYS_FILE: docker-keys.yml
+      EQ_ENVIRONMENT_PREFIX: dev
+      EQ_DYNAMODB_ENDPOINT: http://localhost:6060
     restart: always
     depends_on:
```
jonnyshaw89: Add a depends on for `eq-docker-dynamodb`

docker-compose.yml
```diff
@@ -44,6 +44,8 @@ services:
       EQ_RABBITMQ_HOST_SECONDARY: rabbit
       EQ_SECRETS_FILE: docker-secrets.yml
       EQ_KEYS_FILE: docker-keys.yml
+      EQ_ENVIRONMENT_PREFIX: dev
+      EQ_DYNAMODB_ENDPOINT: http://localhost:6060
```
jonnyshaw89: you should route the communication within the docker network by changing this to `http://eq-docker-dynamodb:6060`

jonnyshaw89 (changes requested):

app/data_model/submitted_responses.py
dwyeradam: I've badly used Exception handling to deal with expected behaviour which I'll refactor.

app/views/questionnaire.py
dwyeradam: See #1329

ajmaddaford:
- The answers need to be stored encrypted exactly as they are when being stored in Postgres. For this we will need to keep the user signed in (so we can encrypt and decrypt the answers with the user provided keys). 
- We should look at doing this by adding configurable backends for questionnaire store. This could potentially allow us to switch out Postgres for DynamoDB for survey state during surveys.
- The extra information required on the thank you and answers after submission screen e.g. ru name, submission time, should be stored in the server side session. 
- In doing this we should also be able to store the eq id and form type in the session and remove them from the urls for post submission screens.

app/views/questionnaire.py
shahi645: @dwyeradam this has been changed to be done earlier : `message = json.dumps(convert_answers(`

app/data_model/submitted_responses.py
dwyeradam: Removed default.

shahi645:
@dwyeradam still getting 5xx's on submission, to do with Boto3's `PutItem` class

jonnyshaw89:
@dwyeradam @shahi645 this is because there is an extra `$` on line 36 of `dev_settings.sh`

app/views/questionnaire.py
```diff
@@ -276,13 +333,32 @@ def submit_answers(routing_path, eq_id, form_type, collection_id):
...
+            submission_validity = g.schema_json['view_submitted_responses'].get('timeout', EQ_SUBMISSION_VALIDITY)
+            encrypter = StorageEncryption(current_user.user_id, current_user.user_ik)
+            encrypted_data = encrypter.encrypt_data(
+                {
+                    'answers': answer_store.answers,
+                    'metadata': metadata
+                }
+            )
+
+            SubmittedResponses().insert_item(metadata['tx_id'], encrypted_data, submitted_time, submission_validity)
```
jonnyshaw89: I think we should look at if we can extend `EncryptedQuestionnaireStorage` and provide multiple implementations then we can just update the storage mechanism on `QuestionnaireStore` and save a copy?
This would also set us in a good place for if we want to use DynamoDB for storing answers eventually

app/views/questionnaire.py
```diff
@@ -293,11 +369,18 @@ def submit_answers(routing_path, eq_id, form_type, collection_id):
         return redirect(invalid_location.url(metadata))
 
 
+def store_submitted_responses():
```
jonnyshaw89: I think this should be renamed as it reads as if the function is doing the storing rather then returning a boolean

app/views/questionnaire.py
```diff
@@ -293,11 +369,18 @@ def submit_answers(routing_path, eq_id, form_type, collection_id):
...
+
+    return False
+
+
 def is_survey_completed(full_routing_path):
     completed_blocks = get_completed_blocks(current_user)
 
     for location in full_routing_path:
-        if location.block_id in ['thank-you', 'summary', 'confirmation']:
+        if location.block_id in ['thank-you', 'summary', 'confirmation', 'view-submission']:
```
jonnyshaw89: I think these need to be changed to look at the Block type. But that may be another PR?

jonnyshaw89 (changes requested):

app/templates/thank-you.html
```diff
@@ -9,19 +9,37 @@ <h1 class="saturn">Submission Successful</h1>
...
-    <p class="mars">Transaction ID (Please quote this reference for any support queries)</p>
-    {% if meta.respondent %}
-    <ul class="list list--boxes">
-        {%for id_part in meta.respondent.tx_id%}<li class="list__item">{{id_part}}</li>{% endfor %}
-    </ul>
-    {% endif %}
-    <p class="u-mb-no mars">You may wish to save or print this page for your records.</p>
+<div class="u-mb-s">
+    <div class="panel panel--simple panel--success panel--spacious">
+        <p class=""mars>Your answers were submitted for <b>{{meta.respondent.ru_name}}</b> on {{meta.survey.submitted_time}}</p>
```
hamishtaplin: The class name is outside the quotes.

app/templates/thank-you.html
```diff
@@ -9,19 +9,37 @@ <h1 class="saturn">Submission Successful</h1>
...
+    <p>Your answers will be processed in the next few weeks.
+        We may contact you to query your answers via phone or secure message</p>
 </div>
 
-{{answer_guidance}}
+{% block view_submission %}
+{% if meta.survey.submission_validity > 0 %}
+    <div class="u-mb-s">
+    {% if view_submission_url %}
+        <a class="btn btn--primary btn--lg" data-qa="btn-view-submission" href={{view_submission_url}}>View and print a copy of your answers</a>
```
hamishtaplin: Links shouldn't look like buttons.

app/templates/view-submission.html
```diff
@@ -0,0 +1,73 @@
...
+
+{% block main -%}
+
+<div>
+  <h1 class="saturn">Submitted answers</h1>
+</div>
+
+<div class="u-mb-s">
+    <div class="panel panel--simple panel--success panel--spacious">
+        <p class=""mars>Your answers were submitted for <b>{{meta.respondent.ru_name}}</b> on {{meta.survey.submitted_time}}</p>
```
hamishtaplin: Class outside the quotes.

hamishtaplin (changes requested):

app/views/questionnaire.py
dwyeradam: Changing to is_submission_stored

app/data_model/submitted_responses.py
```diff
@@ -0,0 +1,64 @@
...
+    def __init__(self):
+        config = Config(retries={'max_attempts': 1})   # Number of additional connection attempts
+        dynamoDB = boto3.resource('dynamodb', endpoint_url=EQ_DYNAMODB_ENDPOINT, config=config)
+        tableName = EQ_ENVIRONMENT_PREFIX + '-submitted-responses'
+        self.table = dynamoDB.Table(tableName)
+
+    def insert_item(self, tx_id, data, submitted, time_to_live):
+        """Insert an item to table"""
+
+        if not time_to_live or time_to_live > EQ_SUBMISSION_VALIDITY:
```
jonnyshaw89: i don't think we need `if not time_to_live` as `time_to_live` is always set

app/settings.py
```diff
@@ -73,6 +76,7 @@ def get_env_or_fail(key):
 
 EQ_JWT_LEEWAY_IN_SECONDS = 120
 DEFAULT_LOCALE = 'en_GB'
+EQ_SUBMISSION_VALIDITY = 900
```
jonnyshaw89: Was this now mandatory in the schema?

app/storage/storage_encryption.py
```diff
@@ -0,0 +1,72 @@
...
+
+    @staticmethod
+    def _generate_key(user_id, user_ik, pepper):
+        sha256 = hashlib.sha256()
+        sha256.update(to_str(user_id).encode('utf-8'))
+        sha256.update(to_str(user_ik).encode('utf-8'))
+        sha256.update(to_str(pepper).encode('utf-8'))
+
+        # we only need the first 32 characters for the CEK
+        return to_bytes(sha256.hexdigest()[:32])
```
jonnyshaw89: if we wrap this in `generate_jwk_from_cek()` then it would save having to call `key = self.generate_jwk_from_cek(self._cek)` whenever we use it.

app/templates/thank-you.html
dwyeradam: @hamishtaplin, @armstrongb If we don't use a button I need some advice on how to make this link a primary feature of the page please.

app/views/questionnaire.py
```diff
@@ -199,6 +215,50 @@ def get_thank_you():
...
+
+    if submission_valid_until > int(datetime.utcnow().strftime('%s')):
+
+        submitted_data = SubmittedResponses().get_item(session_data.tx_id)
+
+        if submitted_data and submitted_data['valid_until'] > int(datetime.utcnow().strftime('%s')):
+
+            metadata_context = build_metadata_context_for_survey_completed(vars(session_data))
+
+            current_location = Location('group', 0, 'view-submission')
```
jonnyshaw89: This code smells a bit. Why do we have a hard-coded Location?

jonnyshaw89: The only thing that this location is used for is to go into `_get_schema_context()` and that only uses the `group_instance` so it might be worth refactoring `_get_schema_context()` to just take in the `group_instance`

app/views/questionnaire.py
```diff
@@ -199,6 +215,50 @@ def get_thank_you():
...
+            submission_view_template = render_theme_template(g.schema_json['theme'],
+                                                             template_name='view-submission.html',
+                                                             meta=metadata_context,
+                                                             analytics_ua_id=current_app.config['EQ_UA_ID'],
+                                                             survey_id=g.schema_json['survey_id'],
+                                                             survey_title=TemplateRenderer.safe_content(g.schema_json['title']),
+                                                             summary=summary_rendered_context)
+            return submission_view_template
+
+    return get_thank_you()
```
jonnyshaw89: I think we should re-direct to the thank you page otherwise you will have the thank you page with the url of `/view-submission`

app/views/questionnaire.py
```diff
@@ -283,17 +343,47 @@ def submit_answers(routing_path):
...
+            if SubmittedResponses().insert_item(metadata['tx_id'], encrypted_data, submitted_time.strftime('%s'),
+                                                submission_validity) is False:
+                submission_validity = 0  # failed load so viewing submission is not possible
 
         session_store = get_session_store()
-        session_store.session_data.submitted_time = datetime.utcnow().isoformat()
+        session_data = session_store.session_data
+        session_data.submitted_time = submitted_time.isoformat()
+        session_data.submission_validity = submission_validity
+        session_store.session_data = session_data
```
jonnyshaw89: Is this line needed? we are editing the object so no need to set it back?

app/views/questionnaire.py
```diff
@@ -283,17 +343,47 @@ def submit_answers(routing_path):
...
         return redirect(invalid_location.url(metadata))
 
 
+def submitted_responses_validity():
+    enabled = False
+    validity = 0
+    if g.schema_json.get('view_submitted_responses'):
+        enabled = g.schema_json['view_submitted_responses'].get('enabled', False)
+        if enabled:
+            validity = g.schema_json['view_submitted_responses'].get('timeout', EQ_SUBMISSION_VALIDITY)
```
jonnyshaw89: Didn't we stop this being optional in the schema?

jonnyshaw89 (changes requested):

app/data_model/answer_store.py
```diff
@@ -122,6 +122,20 @@ def add_or_update(self, answer):
         except ValueError:
             self.add(answer)
 
+    def add_multiple_answers(self, answers):
+        """
+        Add a new answers into the answer store, or update if it exists.
```
shahi645: typo get rid of s

app/data_model/submitted_responses.py
```diff
@@ -0,0 +1,61 @@
+import boto3
+from botocore.exceptions import BotoCoreError, ConnectionError as BotoConnectionError
+from botocore.config import Config
+from app.settings import EQ_ENVIRONMENT_PREFIX, EQ_DYNAMODB_ENDPOINT
+from structlog import get_logger
+
+logger = get_logger()
+
+
+class SubmittedResponses(object):
```
jonnyshaw89, line 8: I think we should create this once for the application maybe in `setup.py` and then reuse it?

jonnyshaw89 (changes requested):

app/data_model/submitted_responses.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def __init__(self):
+        config = Config(retries={'max_attempts': 1})   # Number of additional connection attempts
```
shahi645: Should `max_attempts` be ore than 1 ?

dwyeradam: We'll check the database is available twice as this is additional checks. It would be very unlikely it wouldn't be able to access it first time as it's AWS to AWS. Any further attempts would cause a time delay felt by the user.

shahi645:
Looks good so far, just tests and print button to go! 👍

app/data_model/answer_store.py
jonnyshaw89: you mean the `a` yeh?

shahi645: yes srry method is multiple answers, my bad

tests/functional/spec/features/submitted_responses.spec.js
```diff
@@ -0,0 +1,64 @@
...
+  });
+
+  describe('Try to view submission after timeout', function() {
+    it('Given I complete to Thank-You page, When I click the submitted answers link after the timeout period, Then I should not be able to view my submitted answers', function() {
+        return browser
+          .isVisible(ThankYouPage.viewSubmitted())
+          .pause(5000)
+          .click(ThankYouPage.viewSubmitted())
+          .getUrl().should.eventually.contain(ThankYouPage.pageName)
+          .getSource().should.eventually.contain('You are no longer able to view and print your answers.');
```
jonnyshaw89: Can you access this text with a selector? it will make the error message better if so

tests/functional/spec/features/submitted_responses.spec.js
```diff
@@ -0,0 +1,64 @@
...
+    });
+  });
+
+    describe('Try to view submission after timeout', function() {
+    it('Given I complete to Thank-You page, When I refresh after the timeout period, Then I should not be able to view my submitted answers', function() {
+        return browser
+          .isVisible(ThankYouPage.viewSubmitted())
+          .pause(5000)
+          .refresh()
+          .getUrl().should.eventually.contain(ThankYouPage.pageName)
```
jonnyshaw89: I don't think we need this check as the url shouldn't change on a refresh?

tests/functional/spec/features/submitted_responses.spec.js
```diff
@@ -0,0 +1,64 @@
...
+          .isVisible(ThankYouPage.viewSubmitted())
+          .pause(5000)
+          .click(ThankYouPage.viewSubmitted())
+          .getUrl().should.eventually.contain(ThankYouPage.pageName)
+          .getSource().should.eventually.contain('You are no longer able to view and print your answers.');
+    });
+  });
+
+    describe('Try to view submission after timeout', function() {
+    it('Given I complete to Thank-You page, When I refresh after the timeout period, Then I should not be able to view my submitted answers', function() {
```
jonnyshaw89: I think the indentation is incorrect here.

jonnyshaw89 (changes requested):

app/storage/encrypted_questionnaire_storage.py
```diff
@@ -42,14 +34,15 @@ def add_or_update(self, data, version):
...
+        data = None
         questionnaire_state = self._find_questionnaire_state()
         if questionnaire_state:
             data = json.loads(questionnaire_state.state)
             version = questionnaire_state.version or 0
 
-            if 'data' in data:
-                decrypted_data = self._decrypt_data(data['data'])
-                return decrypted_data, version
+        if data and 'data' in data:
```
jonnyshaw89: If we keep this nested inside the `if` above then we don't need the extra code here?

jonnyshaw89 (changes requested):

app/data_model/submitted_responses.py
dwyeradam: Done.

app/templates/view-submission.html
```diff
@@ -0,0 +1,73 @@
...
+
+<div class="u-mb-s">
+    <div class="panel panel--simple panel--success panel--spacious">
+        <p class="mars">Your answers were submitted for <b>{{meta.respondent.ru_name}}</b> on {{meta.survey.submitted_time}}</p>
+        <p class="mars">Transaction ID: <b>{{meta.respondent.tx_id}}</b></p>
+    </div>
+</div>
+
+<div>
+  <div class="summary">
```
ajmaddaford: We should try and share this summary html with the summary screen - is the only difference no 'Edit' links?

app/templating/metadata_context.py
```diff
@@ -51,8 +54,12 @@ def build_metadata_context_for_survey_completed(survey_completed_metadata):
     return {
         'survey': {
             'period_str': survey_completed_metadata['period_str'],
+            'submission_validity': round(Decimal(survey_completed_metadata['submission_validity'] / 60), 2),
```
ajmaddaford: `survey_completed_metadata` has been refactored into SessionData, can we pass the SessionData class into this method instead?

app/templating/metadata_context.py
```diff
@@ -51,8 +54,12 @@ def build_metadata_context_for_survey_completed(survey_completed_metadata):
...
+            'submitted_time': dateutil.parser.parse(survey_completed_metadata['submitted_time']).strftime('%d %B %Y at %H:%M')
```
ajmaddaford: An ISO format date should be set here and the template should convert it (probably via a jinja filter) into the `%d %B %Y at %H:%M` format

app/views/questionnaire.py
```diff
@@ -224,6 +235,48 @@ def get_thank_you(eq_id, form_type):  # pylint: disable=unused-argument
...
+
+        if submitted_data and submitted_data['valid_until'] > int(datetime.utcnow().strftime('%s')):
+
+            metadata_context = build_metadata_context_for_survey_completed(vars(session_data))
+
+            encrypter = StorageEncryption(current_user.user_id, current_user.user_ik)
+
+            answer_store = AnswerStore()
+            submitted_data = json.loads(encrypter.decrypt_data(submitted_data['data']))
+            answer_store.add_multiple_answers(submitted_data.get('answers'))
```
ajmaddaford: You can pass a dict of answers to AnswerStore on initialisation - `answer_store = AnswerStore(existing_answers= submitted_data.get('answers'))` - which should remove the need for the new `add_multiple_answers` method.

app/views/questionnaire.py
```diff
@@ -224,6 +235,48 @@ def get_thank_you(eq_id, form_type):  # pylint: disable=unused-argument
     return _redirect_to_latest_location(routing_path, collection_id, metadata.get('eq_id'), metadata.get('form_type'))
 
 
+@post_submission_blueprint.route('view-submission', methods=['GET'])
+@login_required
+def get_view_submission(eq_id, form_type):  # pylint: disable=unused-argument
+
+    session_data = get_session_store().session_data
+    submission_valid_until = int(dateutil.parser.parse(session_data.submitted_time).strftime('%s')) + session_data.submission_validity
```
ajmaddaford: I think it would be clearer if we used `datetime.timedelta` and compared date objects rather than converting dates to ints. Something like:
```
submission_time = dateutil.parser.parse(session_data.submitted_time)
submission_valid_until = submission_time + datetime.timedelta(seconds=submission_validity)
```
The comparison would then be:
```
if submission_valid_until > datetime.utcnow():
```

This applies to the other places where dates are converted to ints

app/views/questionnaire.py
```diff
@@ -303,20 +356,52 @@ def submit_answers(routing_path, eq_id, form_type):
         if not sent:
             raise SubmissionFailedException()
 
-        get_questionnaire_store(current_user.user_id, current_user.user_ik).delete()
+        store_submitted_responses_and_session_data(answer_store.answers, metadata)
```
ajmaddaford: I don't think we should be combining this into one method. We always store the session data, but only store the submitted responses if it's enabled for the questionnaire.

app/views/questionnaire.py
```diff
@@ -303,20 +356,52 @@ def submit_answers(routing_path, eq_id, form_type):
...
+def submitted_responses_validity():
+    enabled = False
+    validity = 0
+    view_submitted_responses = g.schema.json.get('view_submitted_responses')
+    if view_submitted_responses:
+        enabled = view_submitted_responses['enabled']
+        if enabled:
+            validity = view_submitted_responses['timeout']
+
+    return enabled, validity
```
ajmaddaford: I think this method should just return whether the schema has view_submitted_responses enabled or not. We can enforce in the schema that timeout should be greater than 0.

app/views/questionnaire.py
```diff
@@ -303,20 +356,52 @@ def submit_answers(routing_path, eq_id, form_type):
...
+
+        if current_app.eq['submitted_responses'].insert_item(metadata['tx_id'],
+                                                             encrypted_data, submitted_time.strftime('%s'),
+                                                             submission_validity) is False:
+            submission_validity = 0  # failed load so viewing submission is not possible
+
+    session_store = get_session_store()
+    session_data = session_store.session_data
+    session_data.submitted_time = submitted_time.isoformat()
+    session_data.submission_validity = submission_validity
```
ajmaddaford: We should keep this simple and just store a boolean to say whether a submission has been persisted to DyanmoDB or not. When we need to later on we can get the timeout from the schema.

ajmaddaford (changes requested):

app/views/questionnaire.py
```diff
@@ -224,6 +237,45 @@ def get_thank_you(eq_id, form_type):  # pylint: disable=unused-argument
...
+
+            metadata = submitted_data.get('metadata')
+
+            routing_path = PathFinder(g.schema, answer_store, metadata).get_full_routing_path()
+
+            schema_context = _get_schema_context(routing_path, 0, metadata, answer_store)
+            rendered_schema = renderer.render(g.schema.json, **schema_context)
+            summary_rendered_context = build_summary_rendering_context(g.schema, rendered_schema, answer_store, metadata)
+
+            submission_view_template = render_theme_template(g.schema.json['theme'],
```
jonnyshaw89: You should just return here rather than assign a variable

jonnyshaw89 (changes requested):

app/views/questionnaire.py
```diff
@@ -208,13 +212,22 @@ def get_thank_you(eq_id, form_type):  # pylint: disable=unused-argument
 
     if session_data.submitted_time:
         schema = load_schema_from_params(session_data.eq_id, session_data.form_type)
-        metadata_context = build_metadata_context_for_survey_completed(vars(session_data))
+        metadata_context = build_metadata_context_for_survey_completed(session_data)
+
+        view_submission_url, view_submission_duration = None, 0
+        if session_data.is_viewable_submission_stored:
```
ajmaddaford: I think we should drop `is_viewable_submission_stored` and rely on checking the submitted time against the view submission duration. If nothing has gone wrong then it would work in the same way. If it hasn't been stored then we can error or redirect back to the thank-you page.

data/en/test_submitted_responses.json
```diff
@@ -0,0 +1,113 @@
...
+    "data_version": "0.0.2",
+    "survey_id": "0",
+    "title": "Other input fields",
+    "theme": "default",
+    "legal_basis": "StatisticsOfTradeAct",
+    "description": "A questionnaire to demo radio field Other input.",
+    "view_submitted_responses": {
+        "enabled": true,
+        "timeout": 5
+    },
```
ajmaddaford: I think we can improve the naming here. It would be clearer if we renamed:
- `view_submitted_responses` to `view_submitted_response` as it's for a single questionnaire response, and
- `timeout` to `duration`

ajmaddaford (changes requested):

ajmaddaford (approved):
I think we need to have an environment variable to allow not using DynamoDB, similar to EQ_RABBITMQ_ENABLED. We'd have to make the EQ_SUBMITTED_RESPONSES_TABLE_NAME environment variable mandatory only when DynamoDB is enabled. This will also allow it to be disabled for the majority of the create_app tests that mock boto3.

ajmaddaford (changes requested):
I think we need to have an environment variable to allow not using DynamoDB, similar to EQ_RABBITMQ_ENABLED. We'd have to make the EQ_SUBMITTED_RESPONSES_TABLE_NAME environment variable mandatory only when DynamoDB is enabled. This will also allow it to be disabled for the majority of the create_app tests that mock boto3.

ajmaddaford (changes requested):
I think we need to have an environment variable to allow not using DynamoDB, similar to EQ_RABBITMQ_ENABLED. We'd have to make the EQ_SUBMITTED_RESPONSES_TABLE_NAME environment variable mandatory only when DynamoDB is enabled. This will also allow it to be disabled for the majority of the create_app tests that mock boto3.

tests/integration/integration_test_case.py
```diff
@@ -40,12 +40,25 @@ def setUp(self):
...
+    def SetUpWithDynamoDB(self):
+        # Cache for requests
+        self.last_url = None
+        self.last_response = None
+        self.last_csrf_token = None
+
+        # Perform setup steps
+        self._setUpApp(True)
+
+    def _setUpApp(self, enableDynamoDB=False):
```
jonnyshaw89: variable names should have `_` rather then capitals.

jonnyshaw89 (changes requested):

app/views/questionnaire.py
```diff
@@ -224,6 +236,44 @@ def get_thank_you(eq_id, form_type):  # pylint: disable=unused-argument
...
 
+@post_submission_blueprint.route('view-submission', methods=['GET'])
+@login_required
+def get_view_submission(eq_id, form_type):  # pylint: disable=unused-argument
+
+    session_data = get_session_store().session_data
+    g.schema = load_schema_from_params(session_data.eq_id, session_data.form_type)
+
+    if _is_submission_viewable(g.schema.json, session_data.submitted_time):
+        submitted_data = current_app.eq['submitted_responses'].get_item(session_data.tx_id)
```
jonnyshaw89, line 248: we need to check here is `submitted_responses` in present
`if 'submitted_responses' in current_app.eq:`

dwyeradam: Added check to `def is_view_submitted_response_enabled(schema):` so that we don't attempt anything around submitted answers unless avaliable.

app/data_model/session_store.py
```diff
@@ -24,7 +26,7 @@ def create(self, eq_session_id, user_id, session_data):
         self.eq_session_id = eq_session_id
         self.user_id = user_id
         self.session_data = session_data
-        self._eq_session = EQSession(self.eq_session_id, self.user_id, json.dumps(vars(self.session_data)))
+        self._eq_session = EQSession(self.eq_session_id, self.user_id, vars(self.session_data))
```
ajmaddaford: I think it would be clearer to not set `session_data` on `EQSession` until `save` is called, as this looks like we're storing session data unencrypted, which isn't the case.

app/data_model/submitted_responses.py
```diff
@@ -0,0 +1,44 @@
...
+
+logger = get_logger()
+
+
+class SubmittedResponses(object):
+
+    def __init__(self, table):
+        self.table = table
+
+    def insert_item(self, tx_id, data, submitted_time, time_to_live):
```
ajmaddaford: I think it would be clearer to call this method `put_item` to be consistent with the DynamoDB naming. I also highlights that `put` will replace an item if it already exists (not that this should happen).

app/data_model/submitted_responses.py
```diff
@@ -0,0 +1,44 @@
...
+            return False
+
+    def get_item(self, tx_id):
+        """ Get an item given its key """
+        try:
+            response = self.table.get_item(Key={'tx_id': tx_id})
+            Item = response.get('Item', None)
+            if Item:
+                Item['valid_until'] = datetime.utcfromtimestamp(Item['valid_until'])
+            return Item
```
ajmaddaford: The variable `item` should be lowercase

app/data_model/submitted_responses.py
```diff
@@ -0,0 +1,44 @@
...
+class SubmittedResponses(object):
+
+    def __init__(self, table):
+        self.table = table
+
+    def insert_item(self, tx_id, data, submitted_time, time_to_live):
+        """Insert an item to table"""
+
+        submitted_time = int(submitted_time.strftime('%s'))  # convert to epoch
+        valid_until = submitted_time + time_to_live
```
ajmaddaford: I think we should create `item` outside of the put method, so the get and put are consistent (the get returns an item dict). It could take a dict with valid until as a time and convert it to epoch in this method, which would also mirror what's done in the get method.

app/jinja_filters.py
```diff
@@ -76,6 +76,12 @@ def format_month_year_date(value, date_format='%B %Y'):
     return "<span class='date'>{date}</span>".format(date=datetime.strptime(value, '%Y-%m').strftime(date_format))
 
 
+@blueprint.app_template_filter()
+def format_iso_date_string(value, date_format='%d %B %Y at %H:%M'):
```
ajmaddaford: - We should call this method `format_datetime` to be consistent with `format_date` and `format_month_year_date`
- We should drop the test variable and just return the string as done in the `format_date` and `format_month_year_date` methods

app/settings.py
```diff
@@ -44,6 +44,7 @@ def get_env_or_fail(key):
 EQ_SECRETS_FILE = os.getenv('EQ_SECRETS_FILE', 'secrets.yml')
 EQ_KEYS_FILE = os.getenv('EQ_KEYS_FILE', 'keys.yml')
 
+EQ_SUBMITTED_RESPONSES_TABLE_NAME = get_env_or_fail('EQ_SUBMITTED_RESPONSES_TABLE_NAME')
```
ajmaddaford: This should only be mandatory if `EQ_DYNAMODB_ENABLED` is `True`

app/templates/thank-you.html
```diff
@@ -9,19 +9,37 @@ <h1 class="saturn">Submission Successful</h1>
...
 
-<div class="panel panel--simple panel--success panel--spacious">
-    <p class="mars">Transaction ID (Please quote this reference for any support queries)</p>
-    {% if meta.respondent %}
-    <ul class="list list--boxes">
-        {%for id_part in meta.respondent.tx_id%}<li class="list__item">{{id_part}}</li>{% endfor %}
-    </ul>
-    {% endif %}
-    <p class="u-mb-no mars">You may wish to save or print this page for your records.</p>
+<div class="u-mb-s">
```
ajmaddaford, line 11: There is extra content on this screen that isn't in the designs. It looks like the "Thank you for submitting the <survey name>, your responses have been successfully received." and "It is now safe to close this window." has been dropped. We should check with @armstrongb.

app/templates/thank-you.html
```diff
@@ -9,19 +9,37 @@ <h1 class="saturn">Submission Successful</h1>
...
+{% if view_submission_duration > 0 %}
+    <div class="u-mb-s">
+    {% if view_submission_url %}
+        <a class="btn btn--primary btn--lg" data-qa="btn-view-submission" href={{view_submission_url}}>View and print a copy of your answers</a>
+        <p class="u-mb-s mars">Your answers will only be available for <b>{{view_submission_duration}} minutes</b>.</p>
+    {% else %}
+        <div class="panel panel--simple panel--info" data-qa="view-submission-expired">
+            <div class="panel__body">
+                <p>You are no longer able to view and print your answers.</p>
+                <p>The link expires to maintain the security of your information.</p>
```
ajmaddaford, line 33: Where does this line come from? I can't see it in the designs.

app/views/questionnaire.py
```diff
@@ -208,14 +212,22 @@ def get_thank_you(eq_id, form_type):  # pylint: disable=unused-argument
...
-                                                   analytics_ua_id=current_app.config['EQ_UA_ID'],
-                                                   survey_id=schema.json['survey_id'],
-                                                   survey_title=TemplateRenderer.safe_content(schema.json['title']))
-        return thank_you_template
+        metadata_context = build_metadata_context_for_survey_completed(session_data)
+
+        view_submission_url, view_submission_duration = None, 0
+        if is_view_submitted_response_enabled(schema.json):
+            view_submission_duration = schema.json['view_submitted_response']['duration']
+            if _is_submission_viewable(schema.json, session_data.submitted_time):
```
ajmaddaford: Duration is only needed if the submission is viewable, so it would be better to move it to the point we work out the url. The logic for converting it to minutes should also move here.

app/views/questionnaire.py
```diff
@@ -208,14 +212,22 @@ def get_thank_you(eq_id, form_type):  # pylint: disable=unused-argument
...
-        thank_you_template = render_theme_template(schema.json['theme'],
-                                                   template_name='thank-you.html',
-                                                   meta=metadata_context,
-                                                   analytics_ua_id=current_app.config['EQ_UA_ID'],
-                                                   survey_id=schema.json['survey_id'],
-                                                   survey_title=TemplateRenderer.safe_content(schema.json['title']))
-        return thank_you_template
+        metadata_context = build_metadata_context_for_survey_completed(session_data)
+
+        view_submission_url, view_submission_duration = None, 0
```
ajmaddaford: I don't think we gain anything by these being on one line and I think it would be clearer on two.

app/views/questionnaire.py
```diff
@@ -224,6 +236,44 @@ def get_thank_you(eq_id, form_type):  # pylint: disable=unused-argument
...
+@login_required
+def get_view_submission(eq_id, form_type):  # pylint: disable=unused-argument
+
+    session_data = get_session_store().session_data
+    g.schema = load_schema_from_params(session_data.eq_id, session_data.form_type)
+
+    if _is_submission_viewable(g.schema.json, session_data.submitted_time):
+        submitted_data = current_app.eq['submitted_responses'].get_item(session_data.tx_id)
+
+        if submitted_data and submitted_data['valid_until'] > datetime.utcnow():
```
ajmaddaford: Do we need to check `submitted_data['valid_until'] > datetime.utcnow()` as it's already checked in the call to `_is_submission_viewable`?

app/views/questionnaire.py
```diff
@@ -303,20 +353,59 @@ def submit_answers(routing_path, eq_id, form_type):
...
-            eq_id=eq_id,
-            form_type=form_type))
+        get_questionnaire_store(current_user.user_id, current_user.user_ik).delete()
+
+        return redirect(url_for('post_submission.get_thank_you', eq_id=eq_id, form_type=form_type))
     else:
         return redirect(invalid_location.url(metadata))
 
 
+def _store_session_data(submitted_time):
```
ajmaddaford: It would be clearer if we named this method `_store_submitted_time_in_session`

app/views/questionnaire.py
```diff
@@ -303,20 +353,59 @@ def submit_answers(routing_path, eq_id, form_type):
...
+    session_store = get_session_store()
+    session_data = session_store.session_data
+    session_data.submitted_time = submitted_time.isoformat()
+    session_store.save()
+
+
+def _store_viewable_submission(answers, metadata, submitted_time):
+    if is_view_submitted_response_enabled(g.schema.json):
```
ajmaddaford: I think it would be clearer if the if statement was move outside of this method, to make it clear we only store the viewable submission if the schema enables it.

app/data_model/session_store.py
```diff
@@ -69,8 +72,10 @@ def _load(self):
 
         if self._eq_session:
             self.user_id = self._eq_session.user_id
+
             if self._eq_session.session_data:
-                self.session_data = json.loads(self._eq_session.session_data, object_hook=lambda d: SessionData(**d))
+                session_data = StorageEncryption(self.user_id, self.user_ik).decrypt_data(self._eq_session.session_data)
```
ajmaddaford: This assumes that the session is encrypted, which isn't the case if someone already has an active session when we deploy this. We should be able to make it work with un-encrypted sessions on load, which will then be encrypted on save.

ajmaddaford (changes requested):

app/views/questionnaire.py
dwyeradam: No, probably a bit of overkill checking that both times match. I'll remove.

app/views/questionnaire.py
dwyeradam: I was using duration as the check for showing either the link or the timeout panel on the thank you page. If it was greater than 0 then it showed the view submission content. Instead i'm now passing a boolean of `is_view_submission_enabled`

app/templates/thank-you.html
dwyeradam: This was agreed with @charlotte-hirst and the line is taken from the error page *ONS-screenshot-expired-print-link-error.png* that we decided not to use but felt this line gave additional useful information here.

app/templates/thank-you.html
dwyeradam: This matches https://docs.google.com/presentation/d/1fU8J2iV35wSEnWuZ1-MNvuCCE4vTxSmZdH5TyXNICSw/edit#slide=id.g27fc58c6a0_0_1

dwyeradam: Talked with @armstrongb and have removed both of these.

app/data_model/session_store.py
```diff
@@ -74,7 +75,10 @@ def _load(self):
             self.user_id = self._eq_session.user_id
 
             if self._eq_session.session_data:
-                session_data = StorageEncryption(self.user_id, self.user_ik).decrypt_data(self._eq_session.session_data)
+                session_data = self._eq_session.session_data
+                if self._is_session_data_encrypted(session_data):
```
jonnyshaw89: We should add a blocked card in the backlog to remove this.

app/settings.py
```diff
@@ -67,7 +66,10 @@ def get_env_or_fail(key):
                                                        1000)
 
 EQ_DYNAMODB_ENABLED = parse_mode(os.getenv('EQ_DYNAMODB_ENABLED', 'True'))
-EQ_DYNAMODB_ENDPOINT = os.getenv('EQ_DYNAMODB_ENDPOINT', None)
+EQ_DYNAMODB_ENDPOINT = os.getenv('EQ_DYNAMODB_ENDPOINT')
```
jonnyshaw89: Can the endpoint go inside the `if` too?

yoganandkunche (changes requested):
@dwyeradam I don't think we need the help us improve on the view answer page as the session is already expired it will only give an error 403.

jonnyshaw89 (approved):

app/templates/summary.html
```diff
@@ -20,65 +18,22 @@
...
-
-        </div>
-
-     {% endfor %}
-
-     </div>
-  {% endfor %}
+{% if content.is_view_submission_enabled %}
+  <div class="u-mb-s u-mb-m@s">
+    <div class="venus download__text">You will have the opportunity to view and print a copy of your answers after submitting this questionnaire</div>
```
hamishtaplin: `download__text` class shouldn't be there

app/templates/view-submission.html
```diff
@@ -0,0 +1,23 @@
...
+</div>
+
+<div class="u-mb-s">
+    <div class="panel panel--simple panel--success panel--spacious">
+        <p class="mars">Your answers were submitted for <b>{{meta.respondent.ru_name}}</b> on {{meta.survey.submitted_time|format_datetime}}</p>
+        <p class="mars">Transaction ID: <b>{{meta.respondent.tx_id}}</b></p>
+    </div>
+</div>
+
+<button class="btn btn--tertiary print__hidden u-no-js-hide" onclick="window.print()">Print this page <span class="btn__icon btn__icon--print-link"></span></button>
```
hamishtaplin, line 19: `.print__hidden` class not needed

hamishtaplin (changes requested):
Couple of minor things

hamishtaplin (changes requested):
Couple of minor things

ajmaddaford:
I get a `TypeError: __init__() missing 2 required positional arguments: 'ru_name' and 'ru_ref'` error when trying to continue a survey with an active session after switching to this version.

data/en/test_submitted_responses.json
```diff
@@ -0,0 +1,113 @@
+{
```
ajmaddaford: It would be clearer if this file was named `test_view_submitted_response.json`

tests/functional/spec/features/submitted_responses.spec.js
```diff
@@ -0,0 +1,66 @@
+const helpers = require('../../helpers');
+
+const RadioPage = require('../../pages/features/submitted_responses/radio.page.js');
+const NumberBlockPage = require('../../pages/features/submitted_responses/test-number-block.page.js');
+const SummaryPage = require('../../pages/summary.page.js');
+const ThankYouPage = require('../../pages/thank-you.page.js');
+const SubmissionPage = require('../../pages/features/submitted_responses/submission.page.js');
+
+
+describe('@watch Submitted Responses', function() {
```
ajmaddaford: - The `@watch` tag needs to be removed
- To be consistent with the other feature test it should be prefixed with `Feature: `

app/templates/view-submission.html
dwyeradam: Without this *Print this page* text appears on the print preview and i'm assuming the print itself. I feel this spoils the look of a printed page.

app/storage/storage_encryption.py
```diff
@@ -0,0 +1,66 @@
...
+
+class StorageEncryption:
+
+    def __init__(self, user_id, user_ik):
+        if user_id is None:
+            raise ValueError('User id must be set')
+        if user_ik is None:
+            raise ValueError('User ik must be set')
+
+        pepper = current_app.eq['secret_store'].get_secret_by_name('EQ_SERVER_SIDE_STORAGE_ENCRYPTION_USER_PEPPER')
```
ajmaddaford: I don't think this class should access current_app, as this introduces the dependency of being in a flask app. Testing this also requires the app context whereas if pepper was passed in it could be tested in isolation.

app/templates/summary.html
ajmaddaford: "a copy of" seems redundant

app/templates/view-submission.html
ajmaddaford, line 19: Is this a button that looks like a link?

app/views/questionnaire.py
```diff
@@ -533,7 +631,11 @@ def _get_context(full_routing_path, block, current_location, form):
         rendered_schema_json = renderer.render(g.schema.json, **schema_context)
         form = form or FlaskForm()
         summary_rendered_context = build_summary_rendering_context(g.schema, rendered_schema_json, answer_store, metadata)
-        context = {'form': form, 'summary': summary_rendered_context, 'variables': variables}
+        context = {'form': form,
+                   'summary': summary_rendered_context,
+                   'variables': variables,
+                   'is_view_submission_enabled': is_view_submitted_response_enabled(g.schema.json),
```
ajmaddaford: I think we need to keep the naming consistent - is there any reason not to call the boolean passed to the template `is_view_submitted_response_enabled`?

app/views/questionnaire.py
```diff
@@ -533,7 +631,11 @@ def _get_context(full_routing_path, block, current_location, form):
...
+                   'is_editable': True}
```
ajmaddaford: I didn't get what this was until I viewed the template. I think something like `answers_are_editable` would be clearer.

docker-compose.yml
```diff
@@ -44,10 +44,15 @@ services:
       EQ_RABBITMQ_HOST_SECONDARY: rabbit
       EQ_SECRETS_FILE: docker-secrets.yml
       EQ_KEYS_FILE: docker-keys.yml
+      EQ_SUBMITTED_RESPONSES_TABLE_NAME: dev-submitted-responses
+      EQ_DYNAMODB_ENDPOINT: http://eq-docker-dynamodb:8000
+      AWS_ACCESS_KEY_ID: dummy
+      AWS_SECRET_ACCESS_KEY: dummy
```
ajmaddaford: We should discourage people from accidentally committing their AWS credentials by having a default and allowing overriding via environment variables (this is done in https://github.com/ONSdigital/eq-compose/blob/master/eq.yml#L91) e.g. `AWS_ACCESS_KEY_ID: "${AWS_ACCESS_KEY_ID:-dummy}"`

tests/app/data_model/test_submitted_responses.py
```diff
@@ -0,0 +1,55 @@
...
+        valid_until = datetime.utcnow() + timedelta(seconds=10)
+
+        self.item = {
+            'tx_id': 'test_tx_id',
+            'data': 'test_data',
+            'valid_until': valid_until
+        }
+
+
+    def test_insert_item(self):
```
ajmaddaford: This needs to be renamed `test_put_item` to match the renamed method

ajmaddaford (changes requested):

app/templates/view-submission.html
dwyeradam: This is taken from the Design Pattern Library so we need to review this along with the opposite link looking like a button to get to the view-submission page.

yoganandkunche (approved):
LGTM 👍🏼

ajmaddaford (approved):

hamishtaplin (approved):

PR #1347, opened 2017-11-07
Handle dates consistently throughout survey runner

jonnyshaw89:
The way that we handle dates within the application has been changed to always hold dates in the full-date format documented in [RFC3339](https://xml2rfc.tools.ietf.org/public/rfc/html/rfc3339.html#anchor14) - YYYY-MM-DD e.g. 2017-01-22

### How to review 
Run `cat database-scripts/pr-1347-apply.sql | sqlite3 /tmp/questionnaire.db`
Test that `test_dates.json` works as before.

Also.
Try launching `test_dates.json` on master and entering data.
Switch to this branch
Re-load the questionnaire with the existing data.
The data should still be available.

### DIFF
--- a/data/en/1_0005.json
+++ b/data/en/1_0005.json
@@ -47,6 +47,8 @@
                 "primary_content": [{
                     "id": "get-started",
                     "content": [{
+                        "description": "<div class=\"panel panel--simple panel--info\"><div class=\"panel__body\"><h2 class=\"venus\">Important notice for December period</h2><p>We understand that many UK businesses have ‘holiday shutdowns’ in December. If this applies to your business, please share your weekly wage figures from a more representative week.</p></div></div>"
+                    }, {
                         "list": [
                             "On average it takes 10 minutes to complete this survey.",
                             "You can provide informed estimates if actual figures aren't available.",
@@ -326,7 +328,7 @@
             "blocks": [{
                     "id": "weekly-pay-introduction",
                     "title": "Weekly Pay",
-                    "description": "<p>The next section covers Weekly pay.</p><p>Please exclude the following from your figures for this section:</p><p><ul><li>trainees on government schemes</li><li>director\u2019s fees</li><li>employer\u2019s NI and contribution to pension schemes</li><li>employees working abroad unless paid directly from this business\u2019s GB payroll</li><li>signing on fees (sporting professionals)</li><li>payment in lieu of notice</li><li>redundancy pay (taxable and non-taxable)</li><li>accrued holiday pay</li><li>employees in Northern Ireland</li><li>benefits employees receive through pay, for example, family working tax credit</li><li>expenses payments for attending meetings, for example, councillors</li></ul></p>",
+                    "description": "<div class=\"panel panel--simple panel--info\"><div class=\"panel__body\"><h2 class=\"venus\">Important notice for December period</h2><p>We understand that many UK businesses have ‘holiday shutdowns’ in December. If this applies to your business, please share your weekly wage figures from a more representative week.</p></div></div><br/><p>The next section covers Weekly pay.</p><p>Please exclude the following from your figures for this section:</p><p><ul><li>trainees on government schemes</li><li>director\u2019s fees</li><li>employer\u2019s NI and contribution to pension schemes</li><li>employees working abroad unless paid directly from this business\u2019s GB payroll</li><li>signing on fees (sporting professionals)</li><li>payment in lieu of notice</li><li>redundancy pay (taxable and non-taxable)</li><li>accrued holiday pay</li><li>employees in Northern Ireland</li><li>benefits employees receive through pay, for example, family working tax credit</li><li>expenses payments for attending meetings, for example, councillors</li></ul></p>",
                     "type": "Interstitial"
                 },
                 {
@@ -895,7 +897,7 @@
             "blocks": [{
                     "id": "fortnightly-pay-introduction",
                     "title": "Fortnightly Pay",
-                    "description": "<p>The next section covers Fortnightly Pay.</p><p>Please exclude the following from your figures for this section:</p><p><ul><li>trainees on government schemes</li><li>director\u2019s fees</li><li>employer\u2019s NI and contribution to pension schemes</li><li>employees working abroad unless paid directly from this business\u2019s GB payroll</li><li>signing on fees (sporting professionals)</li><li>payment in lieu of notice</li><li>redundancy pay (taxable and non-taxable</li><li>accrued holiday pay</li><li>employees in Northern Ireland</li><li>benefits employees receive through pay, for example, family working tax credit</li><li>expenses payments for attending meetings, for example, councillors</li></ul></p>",
+                    "description": "<div class=\"panel panel--simple panel--info\"><div class=\"panel__body\"><h2 class=\"venus\">Important notice for December period</h2><p>We understand that many UK businesses have ‘holiday shutdowns’ in December. If this applies to your business, please share your weekly wage figures from a more representative week.</p></div></div><br/><p>The next section covers Fortnightly Pay.</p><p>Please exclude the following from your figures for this section:</p><p><ul><li>trainees on government schemes</li><li>director\u2019s fees</li><li>employer\u2019s NI and contribution to pension schemes</li><li>employees working abroad unless paid directly from this business\u2019s GB payroll</li><li>signing on fees (sporting professionals)</li><li>payment in lieu of notice</li><li>redundancy pay (taxable and non-taxable</li><li>accrued holiday pay</li><li>employees in Northern Ireland</li><li>benefits employees receive through pay, for example, family working tax credit</li><li>expenses payments for attending meetings, for example, councillors</li></ul></p>",
                     "type": "Interstitial"
                 },
                 {


MIslam92 (approved):
Works as expected

app/data_model/answer_store.py
```diff
@@ -247,6 +254,26 @@ def get_hash(self):
         """
         return hash(json.dumps(self.answers, sort_keys=True))
 
+    def upgrade(self):
+        if self.version < self._LATEST_VERSION:
+            if self.version < 1:
```
dwyeradam: Can we split this out to make it clearer that upgrade is a generic change to the answer store structure and the second "if" is for the specific date upgrade.

dwyeradam (changes requested):

tests/app/validation/test_month_year_validator.py
```diff
@@ -83,7 +83,7 @@ def test_month_year_date_validator_invalid_year(self):
 
         mock_form = Mock()
         mock_form.month.data = "12"
-        mock_form.year.data = "17"
```
dwyeradam: Does the fact 17 no longer fails this test mean we need to consider validation around years with only 2 digits.

tests/app/validation/test_date_check_validator.py
```diff
@@ -118,7 +118,7 @@ def test_date_type_validator_invalid_year(self):
         mock_form = Mock()
         mock_form.day.data = '01'
         mock_form.month.data = '1'
-        mock_form.year.data = '16'
```
dwyeradam: As above. Is 2 digit year validation now required?

tests/app/templating/test_metadata_context.py
```diff
@@ -38,9 +38,9 @@ def test_build_metadata_context(self):
         survey_data = render_data['survey']
         self.assertIsNotNone(survey_data)
 
-        self.assertEqual('7 July 2016', survey_data['return_by'].strftime('%-d %B %Y'))
-        self.assertEqual('2 February 2016', survey_data['start_date'].strftime('%-d %B %Y'))
-        self.assertEqual('3 March 2016', survey_data['end_date'].strftime('%-d %B %Y'))
+        self.assertEqual('2016-07-07', survey_data['return_by'])
```
dwyeradam: I know you're using the existing test dates but can we avoid testing with Days and Months being the same. I also always prefer to set days to be greater than 12 so it's easier to differentiate. There are a few instances of this.

tests/app/templating/test_schema_context.py
```diff
@@ -51,11 +51,11 @@ def test_build_exercise(self):
...
         exercise = schema_context['exercise']
-        self.assertEqual('2016-10-11', exercise['start_date'].date().isoformat())
-        self.assertEqual('2016-10-12', exercise['end_date'].date().isoformat())
+        self.assertEqual('2016-10-11', exercise['start_date'])
+        self.assertEqual('2016-10-12', exercise['end_date'])
         self.assertEqual('October 2016', exercise['period_str'])
-        self.assertEqual('2016-10-09', exercise['employment_date'].date().isoformat())
-        self.assertEqual('2016-10-10', exercise['return_by'].date().isoformat())
+        self.assertEqual('2016-10-09', exercise['employment_date'])
+        self.assertEqual('2016-10-10', exercise['return_by'])
```
dwyeradam: As above. Testing where day = month.

dwyeradam (dismissed):

app/data_model/answer_store.py
```diff
@@ -1,7 +1,12 @@
 import re
+
+from datetime import datetime
+from flask import g
```
ajmaddaford: We should avoid flask globals being used outside of the views and globals.py - this only appears to be used to get the schema json, can this be passed to the answer store on initialisation?

app/data_model/answer_store.py
```diff
@@ -247,6 +254,26 @@ def get_hash(self):
...
         return hash(json.dumps(self.answers, sort_keys=True))
 
+    def upgrade(self):
+        if self.version < self._LATEST_VERSION:
+            if self.version < 1:
+                # Update Date formats
+                for answer in self.answers:
+                    block_json = SchemaHelper.get_block(g.schema_json, answer['block_id'])
+                    answer_schema_list = SchemaHelper.get_answers_by_id_for_block(block_json)
+                    answer_schema = answer_schema_list[answer['answer_id']]
```
ajmaddaford: I think this logic should live in SchemaHelper, so we could call something like `SchemaHelper.get_answer_schema_for_answer_id(answer['answer_id'])`, or maybe even going as far as a `get_answer_type` method in SchemaHelper.

app/data_model/answer_store.py
```diff
@@ -247,6 +254,26 @@ def get_hash(self):
...
+
+                    if answer_schema['type'] == 'Date':
+                        answer['value'] = datetime.strptime(answer['value'], '%d/%m/%Y').strftime('%Y-%m-%d')
+                        continue
+
+                    if answer_schema['type'] == 'MonthYearDate':
+                        answer['value'] = datetime.strptime(answer['value'], '%m/%Y').strftime('%Y-%m-%d')
```
ajmaddaford: I think we should handle month-year dates as YYYY-MM as mentioned in https://www.w3.org/TR/NOTE-datetime. Currently:
- It's not clear when storing it with the day what the day would be
- Given just the data you don't know whether its a full date for a month-year date

app/data_model/answer_store.py
```diff
@@ -247,6 +254,26 @@ def get_hash(self):
         """
         return hash(json.dumps(self.answers, sort_keys=True))
 
+    def upgrade(self):
```
ajmaddaford: I think we need to consider how we manage multiple versions and upgrades of answers across multiple version changes e.g. loading data v1 into data v3. We also need to consider how we test these migrations.

ajmaddaford (changes requested):

app/jinja_filters.py
```diff
@@ -102,19 +100,14 @@ def format_start_end_date(start_date, end_date=None):
 
 @blueprint.app_template_filter()
 def format_str_as_date_range(value):
-    from_date = format_str_as_date(value['from'])
-    to_date = format_str_as_date(value['to'])
+    from_date = format_date(value['from'])
+    to_date = format_date(value['to'])
```
ajmaddaford: If we replace the only call to this method in `templates/partials/summary/date.html` with `format_start_end_date(answer.value.from, answer.value.to)` we can remove this method. We should also rename `format_start_end_date` to `format_date_range`.

app/jinja_filters.py
```diff
@@ -102,19 +100,14 @@ def format_start_end_date(start_date, end_date=None):
...
 
 @blueprint.app_template_filter()
 def format_str_as_month_year_date(value):
-    return datetime.strptime(value, '%m/%Y').strftime('%B %Y')
-
-
-@blueprint.app_template_filter()
-def format_str_as_date(value):
-    return format_date(datetime.strptime(value, '%d/%m/%Y'))
+    return format_date(value, '%B %Y')
```
ajmaddaford: - Now we have a consistent date format I think `str` should be removed from method names, so this should be `format_month_year_date`. This would also then be consistent with `format_date` and `format_date_range`.
- Does it matter that the date will now be output in a `<span>` with a `date` class? (this applies to any methods that have been changed to use `format_date`)

app/submitter/converter.py
```diff
@@ -261,3 +288,7 @@ def convert_feedback(message, name, email, url, metadata, survey_id):
     }
 
     return payload
+
+
+def format_date_for_submission(value, date_format='%d/%m/%Y'):
+    return datetime.strptime(value, '%Y-%m-%d').strftime(date_format)
```
ajmaddaford: Does this work for month year dates in the format `YYYY-mm`?

ajmaddaford (changes requested):

app/jinja_filters.py
jonnyshaw89: the `<span>` stops the date line breaking in the middle, which i think it something we want to keep across the board

jonnyshaw89:
I am currently making changes to move the `version` onto the `QuestionnaireStore` and for that version to be stored un-encrypted in the database

jonnyshaw89:
@ajmaddaford We need to decide how we manage database changes.
As previously discussed. i have tested the app on `master` with extra columns in the database and there is no impact.

app/data_model/database.py
```diff
@@ -19,10 +19,16 @@ class QuestionnaireState(base):
     __tablename__ = 'questionnaire_state'
     user_id = Column('userid', String, primary_key=True)
     state = Column('questionnaire_data', String)
+    version = Column('version', Integer)
+    last_updated = Column('last_updated', Integer)
+    created = Column('created', Integer)
```
ajmaddaford: - I think the updated and created columns should use the sqlalchemy DateTime type. This will serialise into Python datetime objects which should make dealing with these columns in code more straightforward. 
- We should have a convention for naming date and date-time columns, one I've used previously that works well is ending the column name with `_at` for times and `_on` for dates. So we'd have `updated_at` and `created_at`
- Can we use sqlalchemy defaults for the updated and created times? (http://sqlalchemy-html.readthedocs.io/en/rel_1_0_7/core/defaults.html#python-executed-functions)

app/data_model/answer_store.py
```diff
@@ -247,6 +252,28 @@ def get_hash(self):
         """
         return hash(json.dumps(self.answers, sort_keys=True))
 
+    def upgrade(self, latest_verison, current_version):
```
ajmaddaford: Not sure about 'upgrade', is 'migrate' clearer?

app/globals.py
```diff
@@ -30,6 +30,8 @@ def get_metadata(user):
 def get_answer_store(user):
     questionnaire_store = get_questionnaire_store(user.user_id, user.user_ik)
 
+    questionnaire_store.version = questionnaire_store.answer_store.upgrade(questionnaire_store.LATEST_VERSION, questionnaire_store.version)
```
ajmaddaford: - I think we should check if `questionnaire_store.version` is less than `questionnaire_store.LATEST_VERSION` before calling upgrade
- Shouldn't questionnaire_store.version always be `LATEST_VERSION` after calling this, so no need to return it?
- Should this logic be in `get_questionnaire_store`?

app/submitter/converter.py
```diff
@@ -261,3 +288,13 @@ def convert_feedback(message, name, email, url, metadata, survey_id):
...
+
+
+def format_date_for_submission(value, date_format='%d/%m/%Y'):
+    separators = value.count('-')
+
+    if separators == 2:
+        return datetime.strptime(value, '%Y-%m-%d').strftime(date_format)
+
+    if separators == 1:
+        return datetime.strptime(value, '%Y-%m').strftime(date_format)
```
ajmaddaford: I'm not sure it's worth having this method, if this formatting was done in `format_dates` we wouldn't need to work out the date formatting using the separator.

app/submitter/converter.py
```diff
@@ -183,10 +190,30 @@ def convert_answers_to_census_data(answer_store, routing_path):
...
-        data.extend(answers_in_block)
+        block_json = SchemaHelper.get_block(questionnaire_json, location.block_id)
+        answer_schema_list = SchemaHelper.get_answers_by_id_for_block(block_json)
+        for answer in answers_in_block:
+            answer_schema = answer_schema_list[answer['answer_id']]
+
+            if answer_schema['type'] in ('Date', 'MonthYearDate'):
+                answer['value'] = format_dates(answer_schema['type'], answer['value'])
+                data.append(answer)
+                continue
```
ajmaddaford: I think we should review whether we need to keep the old date format for the v2 data format. I'll ask @charlotte-hirst

database_scripts/PR1347/apply.sql
```diff
@@ -0,0 +1,8 @@
+ALTER TABLE questionnaire_state
+  ADD COLUMN created INT;
+
+ALTER TABLE questionnaire_state
+  ADD COLUMN last_updated INT;
+
+ALTER TABLE questionnaire_state
+  ADD COLUMN version INT;
```
ajmaddaford: - Directory names should be hyphen separated 
- I think we can drop a directory level, so we'd have `database-scripts/pr-1347-apply.sql`

ajmaddaford (changes requested):

app/data_model/database.py
jonnyshaw89: @ajmaddaford do we want to be holding dates as Python datetime objects?

app/submitter/converter.py
jonnyshaw89: @ajmaddaford any answer on this?

ajmaddaford: @charlotte-hirst has confirmed we're safe to change the date format

app/views/questionnaire.py
```diff
@@ -432,14 +432,18 @@ def _format_answer_value(answer_value):
...
+            int(answer_value['year']),
             int(answer_value['month']),
-            answer_value['year'],
+            int(answer_value['day'])
+        )
+    elif is_month_year:
+        formatted_answer_value = '{:04d}-{:02d}-{:02d}'.format(
+            int(answer_value['year']),
+            int(answer_value['month']),
+            int(1)
```
ajmaddaford: Shouldn't this return answers formatted as `YYYY-mm`?

ajmaddaford (changes requested):

app/data_model/database.py
```diff
@@ -1,7 +1,7 @@
 import datetime
 import time
 
-from sqlalchemy import Column
+from sqlalchemy import Column, Integer
```
ajmaddaford: Minor, we should be consistent with imports - either group all on one line or use a line for each

app/submitter/converter.py
```diff
@@ -186,6 +193,16 @@ def convert_answers_to_census_data(answer_store, routing_path):
     return data
 
 
+def format_dates(answer_type, value):
+    if answer_type == 'Date':
+        return datetime.strptime(value, '%Y-%m-%d').strftime('%d/%m/%Y')
+
+    if answer_type == 'MonthYearDate':
+        return datetime.strptime(value, '%Y-%m').strftime('%m/%Y')
```
ajmaddaford: Is there value in having this method? We could just do this formatting in `_format_downstream_answer`

app/data_model/answer_store.py
```diff
@@ -247,6 +252,22 @@ def get_hash(self):
         """
         return hash(json.dumps(self.answers, sort_keys=True))
 
+    def upgrade(self, current_version):
```
ajmaddaford: I can't see any unit tests for this method

ajmaddaford (changes requested):

app/data_model/database.py
jonnyshaw89: Ill tidy these up but they all go in my other PR

yoganandkunche (changes requested):
Fails to launch a survey with existing data
ValueError: time data Markup('17/05/1985') does not match format '%Y-%m-%d'
127.0.0.1 - - [01/Dec/2017 09:02:02] "GET /questionnaire/1/0102/LsFRCVb5/rsi/0/total-retail-turnover-block HTTP/1.1" 500 -
127.0.0.1 - - [01/Dec/2017 09:02:02] "GET /s/css/default/responsive.min.css?q=e1a797a4c8e72c633389f4d4780f6a330085574d HTTP/1.1" 200 -
127.0.0.1 - - [01/Dec/2017 09:02:02] "GET /s/img/logo.svg?q=e1a797a4c8e72c633389f4d4780f6a330085574d HTTP/1.1" 200 -
127.0.0.1 - - [01/Dec/2017 09:02:02] "GET /s/js/bundle.min.js?q=e1a797a4c8e72c633389f4d4780f6a330085574d HTTP/1.1" 200 -
127.0.0.1 - - [01/Dec/2017 09:02:02] "GET /s/img/ogl.svg?q=e1a797a4c8e72c633389f4d4780f6a330085574d HTTP/1.1" 200 -

jonnyshaw89:
@yoganandkunche The issue identified has been fixed. It was failing to upgrade the AnswerStore

ajmaddaford:
I think the app should fail to start if the required database migration hasn't been run so that:
- We can warn developers that the database needs to be patched or removed
- The application won't start in pre-production and production in a broken state

yoganandkunche (approved):
Issue retest worked as expected and happy from a functional point of view

jonnyshaw89:
@ajmaddaford @dwyeradam Can you re-review?

app/data_model/models.py
```diff
@@ -12,13 +12,17 @@ class QuestionnaireState(db.Model):
     __tablename__ = 'questionnaire_state'
     user_id = db.Column('userid', db.String, primary_key=True)
     state = db.Column('questionnaire_data', db.String)
+    version = db.Column('version', db.Integer)
+    last_updated = db.Column('last_updated', db.Integer)
+    created = db.Column('created', db.Integer)
```
ajmaddaford: I think these column names have regressed in the merge of the database refactor, they should be `updated_at` and `created_at`

ajmaddaford (changes requested):

ajmaddaford (approved):

PR #1352, opened 2017-11-09
Remove period from RSI QBS Banners

dwyeradam:
Due to a missed requirement the period which has been added to the banner doesn't apply to employment questions within surveys so the employment area would like us to take it out of the banner for those surveys affected until we have done user research.

### How to review 
Check surveys 1_0112, 1_0213, 1_0215 and 2_0001 no longer have period information in their banners.


jonnyshaw89 (approved):

yoganandkunche (approved):
LGTM :+1:

PR #1358, opened 2017-11-10
Update how many vistors census household question

MIslam92:
Modified to pipe andrew line one now into question

Card: https://trello.com/c/W93tQly9/1463-update-how-many-visitors

### How to review 
Check on Census Household


charlotte-hirst (approved):
Think we need to look separately at the colour we use for highlighting as purple isnt great

dwyeradam (approved):

PR #1363, opened 2017-11-10
Repeat rules only apply from answers within the routing path

jonnyshaw89:
Repeating rules did not honour the routing path.
This PR addresses that so we no longer need to delete the answer driving a repeat when changing path through a survey

### How to review 
Test `census_household` by adding household members.
Go back and answer nobody lives at address, observe the household members are removed from the routing path. (navigation)
Re-answer "does anyone live here" question to yes, observe that previous answers for household members persisted.


yoganandkunche (approved):
LGTM :+1:

shahi645 (approved):
Works for me. @jonnyshaw89 Squash ?

PR #1373, opened 2017-11-16
Eq 1483 piping multiple into question and answer

shahi645:
Piping multiple fields into a question and an answer. Test schema(test_multiple_piping) created to show this, with new functional tests also added.

Trello Card:
https://trello.com/c/SDpyjNdX/1483-piping-multiple-answers-into-a-question-and-an-answer-option-s

### How to review

Run test_multiple_piping does it behave as card demonstrates?

Run unit tests and functional tests

### DIFF
--- a/app/jinja_filters.py
+++ b/app/jinja_filters.py
@@ -119,7 +119,7 @@ def format_str_as_date(value):
 
 @blueprint.app_template_filter()
 def format_household_member_name(names):
-    return ' '.join(map(lambda name: name.strip(), filter(None, names)))
+    return concatenated_list(list_items=names, delimiter=' ')
 
 
 @blueprint.app_template_filter()
@@ -133,7 +133,7 @@ def format_household_member_name_possessive(names):
 
 
 @blueprint.app_template_filter()
-def format_list(list_items):
+def format_unordered_list(list_items):
     summary_list = ''
 
     if list_items and list_items[0]:
@@ -146,13 +146,18 @@ def format_list(list_items):
 
 
 @blueprint.app_template_filter()
+def concatenated_list(list_items, delimiter=', '):
+    return delimiter.join(map(lambda item: item.strip(), filter(None, list_items)))
+
+
+@blueprint.app_template_filter()
 def format_household_summary(names):
     if names:
         person_list = []
         for first_name, middle_name, last_name in zip(names[0], names[1], names[2]):
             person_list.append(format_household_member_name([first_name, middle_name, last_name]))
 
-        return format_list([person_list])
+        return format_unordered_list([person_list])
     return ''
 
 
--- a/data/en/test_multiple_piping.json
+++ b/data/en/test_multiple_piping.json
@@ -1,0 +1,143 @@
+{
+    "mime_type": "application/json/ons/eq",
+    "schema_version": "0.0.1",
+    "data_version": "0.0.2",
+    "survey_id": "0",
+    "title": "Multiple piping test survey",
+    "theme": "default",
+    "legal_basis": "StatisticsOfTradeAct",
+    "description": "A questionnaire to test multiple piping into a question and answer",
+    "groups": [{
+        "id": "what-is-your-address-group",
+        "title": "What is your address?",
+        "blocks": [{
+                "type": "Questionnaire",
+                "id": "what-is-your-address",
+                "questions": [{
+                    "id": "what-is-your-address-question",
+                    "title": "What is your address?",
+                    "type": "General",
+                    "answers": [{
+                            "id": "building",
+                            "label": "Building",
+                            "mandatory": false,
+                            "type": "TextField",
+                            "alias": "address_building"
+                        },
+                        {
+                            "id": "address-line-1",
+                            "label": "Address Line 1",
+                            "mandatory": true,
+                            "type": "TextField",
+                            "alias": "address_line_1",
+                            "validation": {
+                                "messages": {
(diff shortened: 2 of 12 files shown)


app/jinja_filters.py
```diff
@@ -133,7 +133,7 @@ def format_household_member_name_possessive(names):
 
 
 @blueprint.app_template_filter()
-def format_list(list_items):
+def format_list_for_html(list_items):
```
MIslam92: Feel like a better name can be used for this to match the rest of the file style?

MIslam92 (changes requested):
Refer to below

app/jinja_filters.py
shahi645: @MIslam92 what would you suggest

MIslam92: format_unordered_list? similar to whats already being used i.e. format_multilined_string as its more explicit to what the function is actually doing

tests/functional/spec/piping.spec.js
```diff
@@ -0,0 +1,36 @@
...
+const PipingAddressPage = require('../pages/surveys/piping/what-is-your-address.page');
+const PipingPersonPage = require('../pages/surveys/piping/household-composition.page');
+const MultiplePipingPage = require('../pages/surveys/piping/term-time-location.page');
+
+describe('Multiple piping into question and answer', function() {
+
+    var piping_schema = 'test_multiple_piping.json';
+
+  it('Given into multiple address fields, when i navigate to term-time-location question, I should see all values separated  by commas and spaces', function() {
+    return helpers.openQuestionnaire(piping_schema).then(() => {
```
MIslam92: Instead of doing this for each scenario you can do a beforeEach to load the survey for each scenario instead e.g.` beforeEach('load the survey', function() {
return helpers.openQuestionnaire(piping_schema);}
);`

MIslam92 (commented):
See comments below

app/jinja_filters.py
```diff
@@ -145,14 +145,19 @@ def format_list(list_items):
     return summary_list
 
 
+@blueprint.app_template_filter()
+def format_list(list_items, delimiter=', '):
```
jonnyshaw89: Does the changing of the name of the function above imply that this list is ordered?
Is this not more like `join_list`?

data/en/test_multiple_piping.json
```diff
@@ -0,0 +1,150 @@
+{
```
jonnyshaw89, line 1: Have you added this to the launcher?

tests/functional/pages/surveys/question.page.js
```diff
@@ -6,6 +6,7 @@ class QuestionPage extends BasePage {
     super(pageName);
     this.questions = [];
   }
+  questionText() { return '.question__title'; }
```
jonnyshaw89: you should add a `data-qa` tag to this

jonnyshaw89 (changes requested):

app/jinja_filters.py
shahi645: @jonnyshaw89 it more implies this is the main list formatting method, but we can rename it `format_join_list`?

data/en/test_multiple_piping.json
shahi645: @jonnyshaw89 yes and merged

tests/functional/pages/surveys/question.page.js
shahi645: Change made

app/jinja_filters.py
jonnyshaw89: i don't see its more of a list than the other, just different ways of formatting a list.
If you look at it in the context for address, is it a list or is it a formatting style?

shahi645: Well it's listing the address 🙃

MIslam92 (approved):

jonnyshaw89:
Happy that this works. Think we should keep the name of the existing filter and find a new one for the new filter.

shahi645:
@jonnyshaw89 food for thought, i will check with Andrew when he comes in this morning

tests/app/test_jinja_filters.py
```diff
@@ -310,6 +310,24 @@ def test_format_household_member_name_possessive_trailing_s(self):
 
         self.assertEqual(format_value, 'John Does\u2019')
 
+    def test_concatenated_list(self):
```
jonnyshaw89, line 313: can you add a test with 1 item in the list?

jonnyshaw89 (changes requested):

tests/app/test_jinja_filters.py
shahi645: Yes, sorry my bad was supposed to do it as discussed

tests/functional/spec/multiple-piping.spec.js
```diff
@@ -0,0 +1,39 @@
...
+
+describe('Multiple piping into question and answer', function() {
+
+    var piping_schema = 'test_multiple_piping.json';
+
+    describe('Scenarios', function() {
+      beforeEach('load the survey', function() {
+        return helpers.openQuestionnaire(piping_schema);
+            });
+
```
jonnyshaw89: Can you correct the formatting of this file?

jonnyshaw89 (changes requested):

jonnyshaw89 (approved):

PR #1383, opened 2017-11-22
Fix error navigating sections

### LINKED ISSUE #1379: Navigating through sections using navigation results in 5xx's
### Expected behaviour
navigate forward to sections not visited yet should be fine

### Actual behaviour
when navigating to a different section, user is presented with a 5xx

load any survey with multiple sections and navigation, eg Census or MWSS

### Technical information
`TypeError` thrown

#### Browser
all 

### Screenshot
screen shot 2017-11-22 at 08 29 58

jonnyshaw89:
The first block of each section was not being added to the routing path if a group before got to a dead end in the routing.

Fixes https://github.com/ONSdigital/eq-survey-runner/issues/1379

### How to review 
Does it fix the issue?

### DIFF
--- a/tests/functional/spec/navigation.spec.js
+++ b/tests/functional/spec/navigation.spec.js
@@ -15,6 +15,7 @@ const SecurityCodeInterstitialPage = require('../pages/surveys/navigation/securi
 const SkipInterstitialPage = require('../pages/surveys/navigation/skip-interstitial.page.js');
 const SkipPaymentPage = require('../pages/surveys/navigation/skip-payment.page.js');
 const SummaryPage = require('../pages/surveys/navigation/summary.page.js');
+const GenericPage = require('../pages/surveys/generic.page.js');
 
 const CoffeePage = require('../pages/surveys/navigation/completeness/coffee.page.js');
 const ResponseYesPage = require('../pages/surveys/navigation/completeness/response-yes.page.js');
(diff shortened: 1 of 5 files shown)


dwyeradam (changes requested):
The group routing in test_routing_group.json no longer works.

shahi645 (changes requested):
Still getting 5xx's, i think we should test generated sections as thats what sees to cause the issue

jonnyshaw89:
@shahi645 i'm not sure what you mean. Can you give a detailed list of steps to reproduce?

tests/functional/spec/navigation.spec.js
```diff
@@ -151,6 +151,15 @@ describe('Navigation', function() {
       });
   });
 
+  it('Given I have a section that can not be reached from the start of the survey, When I click on the navigation link,' +
+    ' Then the first block of that section should be displayed', function() {
+    return helpers.openQuestionnaire('test_navigation_routing.json').then(() => {
```
MIslam92, line 157: Would it not be better to just do a beforeEach to open the questionnaire rather then having to do it for every scenario for the Navigation?

MIslam92 (dismissed):

tests/functional/spec/navigation.spec.js
jonnyshaw89: @MIslam92 this questionnaire is only used once.

MIslam92: @jonnyshaw89 Sorry, I was referring to when the test_navigation.json was used my bad

shahi645:
@jonnyshaw89 
- Load MWSS 
- Select several pay patterns
- Try navigating to a later section via navigation, and you should see a 500

shahi645:
@jonnyshaw89 Tested out just now, the changes are working. Cant reproduce error 👍

shahi645 (approved):
After a thorough test this is working as it should now, affected surveys work fine

dwyeradam (approved):

dwyeradam:
Have added an issue #1386 we found during this PR. Using routing to skip Sections should not be allowed at schema validation.

dwyeradam (changes requested):
Do we have a test that would currently fail on master?

dwyeradam (approved):

PR #1388, opened 2017-11-27
Add Exclusive Number Validation to MWSS

dwyeradam:
Missing requirements around the pay patterns we have applied to MWSS, we were asked for each of the 'of which' on every pay pattern to add "less than or equal to", this isn't quite right. In some instances it is less than only.

### How to review 
Run MWSS and check following changes:

- Holiday pay  -  weekly/fortnightly - 'less than'
- Arrears  - weekly/fortnightly/monthly/4 weekly & 5 weekly -  'less than' 
- Bonus/commission - calendar monthly, 4 weekly, 5 weekly - 'less than'

### DIFF
--- a/data/en/1_0005.json
+++ b/data/en/1_0005.json
@@ -435,11 +435,12 @@
                                 "currency": "GBP",
                                 "decimal_places": 2,
                                 "max_value": {
-                                    "answer_id": "weekly-pay-gross-pay-answer"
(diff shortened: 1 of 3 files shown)


shahi645 (approved):
tested very deeply 🙃

jonnyshaw89 (approved):

Issue #1376, opened 2017-11-20
Application doesn't reconnect to database after losing connection


ajmaddaford:
### Expected behaviour
When the application loses connectivity to the database, it should attempt to re-connect

### Actual behaviour
The application did not reconnect to the database.

### Steps to reproduce the behaviour
Unknown locally. This happened when down-sizing the RDS instance in the pre-production environment.

### Log
```
{
    "status_code": 500,
    "created": "2017-11-20T08:27:23.486477Z",
    "event": "an error has occurred",
    "url": "http://eq.onsdigital.uk/session?token=<token>",
    "service": "eq-survey-runner",
    "request_id": "2a19b14d-71ab-461b-a552-176011bdbeb7",
    "exception": [
        "Traceback (most recent call last):",
        "  File '/opt/python/run/venv/lib64/python3.4/site-packages/sqlalchemy/engine/base.py', line 1182, in _execute_context",
        "    context)",
        "  File '/opt/python/run/venv/lib64/python3.4/site-packages/sqlalchemy/engine/default.py', line 470, in do_execute",
        "    cursor.execute(statement, parameters)",
        "psycopg2.OperationalError: SSL connection has been closed unexpectedly",
        "",
        "",
        "The above exception was the direct cause of the following exception:",
        "",
        "Traceback (most recent call last):",
        "  File '/opt/python/run/venv/lib64/python3.4/site-packages/flask/app.py', line 1612, in full_dispatch_request",
        "    rv = self.dispatch_request()",
        "  File '/opt/python/run/venv/lib64/python3.4/site-packages/flask/app.py', line 1598, in dispatch_request",
        "    return self.view_functions[rule.endpoint](**req.view_args)",
        "  File '/opt/python/current/app/app/views/session.py', line 51, in login",
        "    jti_claim_storage.use_jti_claim(jti_claim)",
        "  File '/opt/python/current/app/app/authentication/jti_claim_storage.py', line 50, in use_jti_claim",
        "    self._database.add(jti)",
        "  File '/opt/python/current/app/app/data_model/database.py', line 119, in add",
        "    self._db_session.commit()",
        "  File '/opt/python/run/venv/lib64/python3.4/site-packages/sqlalchemy/orm/scoping.py', line 157, in do",
        "    return getattr(self.registry(), name)(*args, **kwargs)",
        "  File '/opt/python/run/venv/lib64/python3.4/site-packages/sqlalchemy/orm/session.py', line 906, in commit",
[...]

PR #1389, opened 2017-11-29
Removing custom DB logic in favour of flask-sqlalchemy

jonnyshaw89:
I have removed most of our Database connection and session handling logic as that is all done by `flask-sqlalchemy` which we are already importing as a dependency.
This also ties the database better into the request context.

### How to review 
Deploy to ElasticBeanstalk
Resize Database
Validate connections come back after resize has finished

Fixes https://github.com/ONSdigital/eq-survey-runner/issues/1376


app/data_model/database.py
```diff
@@ -62,64 +57,34 @@ def __repr__(self):
 
 class Database:
 
-    def __init__(self, driver, database_name, host=None, port=None,
-                 username=None, password=None, setup_attempts=SETUP_ATTEMPTS,
-                 setup_retry_delay=SETUP_RETRY_DELAY_SECONDS):
```
MIslam92: Any reason the all the setup_retry_delay stuff was removed? When your initialising the db etc. How does SQLAlchemy with the retry during setup if a error occurs itself?

tests/app/authentication/test_database_session_manager.py
```diff
@@ -16,10 +16,9 @@ class TestSessionManager(AppContextTestCase):
 
     def setUp(self):
         super().setUp()
-        self.database = Mock(Database(driver='sqlite',
-                                      database_name='',
-                                      setup_attempts=1,
-                                      setup_retry_delay=0))
+        self.database = Mock(Database(application=self._app,
+                                      driver='sqlite',
+                                      database_name=''))
```
MIslam92: Should this not have some name something like test_db?

app/data_model/database.py
jonnyshaw89: It kept retrying when i tested but you should test too.

tests/app/authentication/test_database_session_manager.py
jonnyshaw89: Its mocked so it makes no difference

MIslam92 (approved):

ajmaddaford:
I think we should look to move the database setup code out of database.py and into setup.py as:
- It doesn't make sense for the database connection code to sit with the models (we might want to consider renaming database.py to models.py)
- It would simplify the code to check the database state on startup in https://github.com/ONSdigital/eq-survey-runner/pull/1347

jonnyshaw89:
@ajmaddaford Ready for you to review again

ajmaddaford (approved):

PR #1390, opened 2017-11-30
fixes plurality of error message

### LINKED ISSUE #1377: Plurality of error messages
### Expected behaviour
When there is only one error message, it should be worded appropriately, something like:
- This page has an error
- This must be corrected to continue

### Actual behaviour
When only one error is raised the wording is:
- This page has 1 errors 
- These must be corrected to continue

On any screen that can generate an error, input invalid data and submit.

### Screenshot
image

yoganandkunche:
Fixes #1377 

### How to review 
Retest #1377 

And all tests should pass


app/templates/questionnaire.html
```diff
@@ -7,7 +7,11 @@
 
     <div class="panel panel--error u-mb-s" data-qa="error">
       <div class="panel__header">
-        <h1 class="panel__title venus">This page has {{mapped_errors|length}} errors</h1>
+          {% if mapped_errors|length > 1 %}
+            <h1 class="panel__title venus">This page has {{mapped_errors|length}} errors</h1>
+          {%- else -%}
+            <h1 class="panel__title venus">This page has an error</h1>
+          {%- endif -%}
```
dwyeradam: We should discuss the use of logic in HTML.

app/templates/questionnaire.html
```diff
@@ -7,7 +7,11 @@
...
       </div>
       <div class="panel__body" data-qa="error-body">
         <p class="mars">These <strong>must be corrected</strong> to continue.</p>
```
dwyeradam: Shouldn't "These" be "This" for a single error?

dwyeradam (changes requested):

app/templates/questionnaire.html
dwyeradam: Try this from flask babel:
{{ ngettext('This', 'These', mapped_errors|length) }} <strong>must be corrected</strong> to continue.

dwyeradam (changes requested):

tests/integration/star_wars/test_errors.py
```diff
@@ -41,4 +41,18 @@ def test_skip_question_errors(self):
 
         # Then errors exists on page
         self.assertStatusOK()
-        self.assertInPage('This page has 1 errors')
+        self.assertInPage('This page has an error')
+        self.assertInPage('This <strong>must be corrected</strong> to continue.')
```
dwyeradam: It may be worth adding a test for multiple errors showing "These" as well.

shahi645 (approved):
@yoganandkunche Looks good!

tests/integration/star_wars/test_errors.py
shahi645: @dwyeradam This is covered below see line 58

dwyeradam (approved):

dwyeradam:
@yoganandkunche Squishy Squashy

PR #1391, opened 2017-11-30
Re-Work Session Storage

dwyeradam:
Store session data in the server side session. The only data that should be in the client side cookie is enough information to identify the user.


app/authentication/authenticator.py
```diff
@@ -76,6 +77,17 @@ def store_session(metadata):
...
 
+def get_session_data(metadata):
+    session_data = {
+        'tx_id': metadata.get('tx_id'),
+        'schema_identifier': metadata.get('eq_id') + '_' + metadata.get('form_type'),
+        'ru_ref': metadata.get('ru_ref'),
+        'ru_name': metadata.get('ru_name'),
+        'period_str': metadata.get('period_str')
+    }
+    return json.dumps(session_data)
```
jonnyshaw89: I think we should be consistent in the `dumps` and `loads` we pass in json to the `SessionStore` but return a dict. I think we should pass in the Dict here and do all of the serialisation inside `SessionStore`

app/authentication/session_storage.py
```diff
@@ -18,47 +19,56 @@ class SessionStorage:
...
-        if EQ_SESSION_ID in session:
-            eq_session_id = session[EQ_SESSION_ID]
-            eq_session = self._get_user_session(eq_session_id)
-            if eq_session is not None:
-                self._database.delete(eq_session)
-            else:
-                logger.debug("eq_session_id from user's cookie not found in database")
-        else:
-            logger.debug("eq_session_id does not exist in user's cookie")
+        if eq_session is None:
```
jonnyshaw89: Why have we changed this behaviour? when would we try to delete a session we don't know exists?

app/authentication/session_storage.py
```diff
@@ -18,47 +19,56 @@ class SessionStorage:
...
+        """
+        eq_session = self._get_user_session()
+        if eq_session is not None:
+            return json.loads(eq_session.session_data)
+
+        return None
 
     @staticmethod
-    def _get_user_session(eq_session_id):
+    def _get_user_session():
```
jonnyshaw89: Should we be caching this result?
How many times in a request will this be hit?

app/views/questionnaire.py
```diff
@@ -40,21 +41,26 @@
...
+                                import_name=__name__,
+                                url_prefix='/thank-you')
+
 
 @questionnaire_blueprint.before_request
 def before_request():
     metadata = get_metadata(current_user)
-    if metadata:
-        logger.bind(tx_id=metadata['tx_id'])
+    if not bool(metadata):
```
jonnyshaw89: Why do we need to cast this to a bool? can't we check not None?

jonnyshaw89: Or even, why would it ever be None?

app/views/questionnaire.py
```diff
@@ -170,13 +176,15 @@ def post_household_composition(routing_path, **kwargs):
...
-    schema = load_schema_from_params(eq_id, form_type)
+@thank_you_blueprint.route('/', methods=['GET'])
+def get_thank_you():  # pylint: disable=unused-argument
+    session_data = current_app.eq['session_storage'].get_session_data()
+    eq_id, form_type = session_data.get('schema_identifier').split('_', maxsplit=1)
+    metadata = get_metadata(current_user)
 
-    if survey_completed_metadata:
-        metadata_context = build_metadata_context_for_survey_completed(survey_completed_metadata)
+    if not bool(metadata):
```
jonnyshaw89: Can't you just check for `None`?

app/views/questionnaire.py
```diff
@@ -273,19 +283,10 @@ def submit_answers(routing_path, eq_id, form_type, collection_id):
...
             raise SubmissionFailedException()
 
-        current_app.eq['session_storage'].store_survey_completed_metadata(
-            metadata['tx_id'],
-            metadata['period_str'],
-            metadata['ru_ref'])
-
         get_questionnaire_store(current_user.user_id, current_user.user_ik).delete()
-        remove_survey_session_data()
+        # remove_survey_session_data()
```
jonnyshaw89: Remove comment

jonnyshaw89 (changes requested):

app/authentication/session_storage.py
dwyeradam: I need to change this. It's actually to fix a test so I need to rewrite the test in test_database_session_manager.py.

app/authentication/session_storage.py
dwyeradam: I did originally cache it however it's only called once per request so not really required. By not caching it will make it easier to use session storage for dynamic values in the future.

app/views/questionnaire.py
dwyeradam: This is after submission when trying to get back into the questionnaire. get_metadata() returns an empty dict as the user is now still logged in but the metadata has been cleared. I'll look at the logic around returning none instead of an empty dict.

app/authentication/session_storage.py
```diff
@@ -69,7 +78,7 @@ def _get_user_session(eq_session_id):
             logger.debug('found matching eq_session for eq_session_id in database',
                          session_id=eq_session.eq_session_id,
                          user_id=eq_session.user_id,
-                         timestamp=eq_session.timestamp.isoformat())
+                         created_at=eq_session.created_at.isoformat())
```
jonnyshaw89: Do we need to log the created time?

app/data_model/database.py
```diff
@@ -32,15 +32,17 @@ class EQSession(base):
...
-    def __init__(self, eq_session_id, user_id):
+    def __init__(self, eq_session_id, user_id, session_data):
         self.eq_session_id = eq_session_id
         self.user_id = user_id
-        self.timestamp = datetime.datetime.now()
+        self.session_data = session_data
 
     def __repr__(self):
-        return "<EQSession('%s', '%s', '%s')>" % (self.eq_session_id, self.user_id, self.timestamp)
+        return "<EQSession('%s', '%s', '%s')>" % (self.eq_session_id, self.user_id, self.session_data)
```
jonnyshaw89: Is it valuable to have the data in this?

app/globals.py
```diff
@@ -24,7 +24,10 @@ def get_metadata(user):
         return None
 
     questionnaire_store = get_questionnaire_store(user.user_id, user.user_ik)
-    return questionnaire_store.metadata
+    if questionnaire_store.metadata:
```
jonnyshaw89: If this is false then `questionnaire_store.metadata` would be None? therefore can you not just return `questionnaire_store.metadata` as it was before?

app/views/questionnaire.py
```diff
@@ -40,21 +41,26 @@
                                     import_name=__name__,
                                     url_prefix='/questionnaire/<eq_id>/<form_type>/<collection_id>/')
 
+after_submission_blueprint = Blueprint(name='after_submission',
```
jonnyshaw89: after submission or post submission? just my opinion, ignore it if you like

app/views/questionnaire.py
```diff
@@ -40,21 +41,26 @@
...
+                                       url_prefix='/questionnaire')
+
 
 @questionnaire_blueprint.before_request
 def before_request():
     metadata = get_metadata(current_user)
-    if metadata:
-        logger.bind(tx_id=metadata['tx_id'])
+    if metadata is None:
+        raise NoTokenException(401)
```
jonnyshaw89: I do wonder if we are trying to re-implement authentication logic.

app/views/questionnaire.py
```diff
@@ -170,13 +176,15 @@ def post_household_composition(routing_path, **kwargs):
     return _render_page(routing_path, block, current_location, post_form=form)
 
 
-@questionnaire_blueprint.route('thank-you', methods=['GET'])
-def get_thank_you(eq_id, form_type, collection_id):  # pylint: disable=unused-argument
-    survey_completed_metadata = current_app.eq['session_storage'].get_survey_completed_metadata()
-    schema = load_schema_from_params(eq_id, form_type)
+@after_submission_blueprint.route('/thank-you', methods=['GET'])
+def get_thank_you():  # pylint: disable=unused-argument
```
jonnyshaw89: you can remove the pylint exclusion as there are no unused arguments on this funciton

jonnyshaw89 (changes requested):

app/authentication/session_storage.py
dwyeradam: We had it before (timestamp) so i've left it in. We probably don't though so worth discussing with @ajmaddaford.

app/data_model/database.py
dwyeradam: Guess not. Will remove.

app/views/questionnaire.py
dwyeradam: I'll toss a coin but I might weight it in your favour as I probably agree.

app/views/questionnaire.py
dwyeradam: We are really as at this point the user is still authenticated but we don't want to allow them back into the questionnaire. It may be worth creating a new error for this.

jonnyshaw89: Do we have the submitted flag/time in the session? if so can we check this and redirect them to the thankyou?

app/views/questionnaire.py
```diff
@@ -170,13 +176,15 @@ def post_household_composition(routing_path, **kwargs):
...
+    session_data = current_app.eq['session_storage'].get_session_data()
+    eq_id, form_type = session_data.get('schema_identifier').split('_', maxsplit=1)
+    metadata = get_metadata(current_user)
```
jonnyshaw89: Here we call `current_app.eq['session_storage'].get_session_data()` which queries `EQSession` and then 2 lines later we call `get_metadata(current_user)` which queries `EQSession` again.
Is there a way we can limit this to 1 call per request?

jonnyshaw89 (changes requested):

app/globals.py
dwyeradam: It will always exist but will be empty after submission. This logic stops it returning an empty dictionary. Will change the later logic to cope with the empty dictionary where expected.

jonnyshaw89: Yeh if this is the only place we set this to `None` then the other code should just handle the empty dict.

app/views/questionnaire.py
dwyeradam: This matches current behaviour if a user attempts to get back into the questionnaire after submission. Another one to discuss with @ajmaddaford. Argument for both but confusion could arise if the user is pressing back but just keeps getting routed to where they were in the first place without any kind of notification.

app/authentication/session_storage.py
```diff
@@ -18,13 +19,13 @@ class SessionStorage:
     def __init__(self, database):
         self._database = database
 
-    def store_user_id(self, user_id):
+    def store_user_id(self, user_id, session_data):
```
shahi645: should we rename this method? 🤷‍♂️

app/authentication/authenticator.py
```diff
@@ -76,6 +76,17 @@ def store_session(metadata):
     logger.info('user authenticated')
 
 
+def get_session_data(metadata):
+    session_data = {
+        'tx_id': metadata.get('tx_id'),
+        'schema_identifier': metadata.get('eq_id') + '_' + metadata.get('form_type'),
```
ajmaddaford: Why not just store `eq_id` and `form_type` in the session? The only place I can see `schema_identifier` used these values are split on the `_`.

app/authentication/authenticator.py
```diff
@@ -76,6 +76,17 @@ def store_session(metadata):
...
+        'ru_ref': metadata.get('ru_ref'),
+        'ru_name': metadata.get('ru_name'),
+        'period_str': metadata.get('period_str')
+    }
+    return session_data
```
ajmaddaford: - I think the definition of the session data should be with the session storage rather than authenticator. 
- We should consider making the session data a class so that we have a clearly defined thing that will be returned. It would also mean we could access session properties with `session.property`. Authenticator could create a SessionData class and call a `populate_from_metadata` method.

ajmaddaford (changes requested):
There is a card in our technical backlog to separate the cookie session out of session_storage, as it's confusing accessing the cookie session while also interacting with the database session. I think we should take this opportunity to do this work. 

We should be able to limit our use of the cookie session to authenticator.py and just pass the session id to session storage. One way we could make it clearer which session is which in authenticator.py would be to do something like `import session as cookie_session`.

app/authentication/authenticator.py
jonnyshaw89: If we only store `eq_id` and `form_type` then loading from url would break?

ajmaddaford: We would store that under a survey_url property, the same as it is in the metadata. Even if a survey is loaded from a url it still has an eq_id and form_type.

app/authentication/session_storage.py
```diff
@@ -15,52 +18,71 @@
...
+        session_data = self._create_session_data_from_dict(metadata)
+        eq_session = EQSession(eq_session_id, user_id, json.dumps(session_data.get_session_data_as_dict()))
 
         logger.debug('Adding eq_session to database', eq_session_id=eq_session_id, user_id=user_id)
         # pylint: disable=maybe-no-member
         db.session.add(eq_session)
         db.session.commit()
 
+    @staticmethod
+    def _create_session_data_from_dict(data_dict):
```
jonnyshaw89: should this be `from_metadata`? as its not just a dict its specifically the metadata?

jonnyshaw89 (changes requested):

app/authentication/authenticator.py
```diff
@@ -91,3 +93,40 @@ def decrypt_token(encrypted_token):
...
+    if USER_IK in cookie_session:
+        return cookie_session[USER_IK] is not None
+    return False
+
+
+def remove_user_ik():
+    """
+    Removes a user's ik from the session
+    """
+
```
jonnyshaw89: remove this line?

app/authentication/session_storage.py
dwyeradam: Removed this method. Instead of a second call to this I'm using the object_hook in json.loads to directly get the SessionData object.

app/authentication/session_storage.py
jonnyshaw89: With this in it errors when loading a session if that session was started on the previous version of code

app/views/questionnaire.py
```diff
@@ -170,13 +176,14 @@ def post_household_composition(routing_path, **kwargs):
...
-    schema = load_schema_from_params(eq_id, form_type)
+@post_submission_blueprint.route('/thank-you', methods=['GET'])
+def get_thank_you():
+    session_data = current_app.eq['session_storage'].get_session_data()
+    metadata = get_metadata(current_user)
 
-    if survey_completed_metadata:
-        metadata_context = build_metadata_context_for_survey_completed(survey_completed_metadata)
+    if not metadata:
+        schema = load_schema_from_params(session_data.eq_id, session_data.form_type)
```
jonnyshaw89: If the version was updated while the user was completing the questionnaire, when they reach the thank-you page it will error.
We should handle this somehow. maybe populate the session_data on submit if it isn't present

jonnyshaw89 (changes requested):

app/views/questionnaire.py
dwyeradam: I'll remove the created_at from the debug message as not really needed. I'll need to create an update for session_data that can be checked at submit.

app/authentication/session_storage.py
```diff
@@ -1,144 +1,98 @@
...
-    def store_user_id(user_id):
+    def store_session_variables(self, user_id, metadata, eq_session_id):
         """
         Create a new eq_session_id and associate it with the user_id specified
         """
-        eq_session_id = str(uuid4())
-        session[EQ_SESSION_ID] = eq_session_id
-        eq_session = EQSession(eq_session_id, user_id)
+        session_data = self._create_session_data_from_meta(metadata)
+        eq_session = EQSession(eq_session_id, user_id, json.dumps(session_data.__dict__))
```
jonnyshaw89: i think the best practice is not to use `__dict__` but to use `vars(session_data)` instead

app/authentication/session_storage.py
```diff
@@ -1,144 +1,98 @@
...
         db.session.commit()
 
+    def update_session_variables(self, metadata):
+        """
+        Update a user's session data
+        """
+        eq_session = self._get_user_session()
+        if eq_session:
+            session_data = self._create_session_data_from_meta(metadata)
+            eq_session.session_data = json.dumps(session_data.__dict__)
```
jonnyshaw89: same as above

jonnyshaw89 (changes requested):

jonnyshaw89 (dismissed):

ajmaddaford:
From discussions with  @dwyeradam and @jonnyshaw89 we're going to:
- Remove cookie helper
  It's not very useful having a helper class that only has methods for manipulating the cookie dict, it's also not clear where these methods are called that they're dealing with the cookie session.
- Use flask logout hook so we can remove session helper
  Session helper only has one method, it looks like we should be able to hook into the flask-login logout handling to do the same thing.
- Make session a global and migrate SessionStorage to SessionStore (a QuestionnaireStore like way of interacting with the current user session)
  Currently every time the session is accessed it is loaded from the database. It would be simpler and more efficient to load this once with each request in a similar way to questionnaire store. 
- Write session data on login, update with submitted time on submission
  Session data is known when someone signs in and will be used for the upcoming contact us page changes, so it makes more sense to write the session data on sign in. Writing the submitted time on submission will allow us to know if the user should be able to request the thank-you page.

app/authentication/authenticator.py
```diff
@@ -54,21 +65,24 @@ def store_session(metadata):
     :param metadata: metadata parsed from jwt token
     """
     # clear the session entry in the database
-    current_app.eq['session_storage'].delete_session_from_db()
+    session_store = get_session_store()
+    session_store.delete_session_from_db()
 
     # also clear the secure cookie data
-    session.clear()
+    cookie_session.clear()
```
jonnyshaw89: could you call `flask_logout()` here rather than deleting from db and clearing cookie?

app/authentication/session_storage.py
```diff
@@ -1,144 +1,112 @@
...
+            ru_name=metadata.get('ru_name'),
+            period_str=metadata.get('period_str'),
+        )
+        return json.dumps(vars(session_data))
+
+    def add_submission_time(self):
+        """
+        add submission time to session data
+        """
+        self.eq_session = self._get_user_session()
```
jonnyshaw89: we shouldn't need to re-assign `self.eq_session` as it is already assigned by `self._get_user_session()`

app/authentication/session_storage.py
```diff
@@ -1,144 +1,112 @@
...
-                # pylint: disable=maybe-no-member
-                db.session.delete(eq_session)
-                db.session.commit()
-            else:
-                logger.debug("eq_session_id from user's cookie not found in database")
-        else:
-            logger.debug("eq_session_id does not exist in user's cookie")
+        eq_session = self._get_user_session()
+        if eq_session:
+            self.eq_session = None
```
jonnyshaw89: I would set the None after the delete

app/globals.py
```diff
@@ -18,6 +19,15 @@ def get_questionnaire_store(user_id, user_ik):
     return store
 
 
+def get_session_store():
+    # Sets up a single SessionStorage instance throughout app.
```
jonnyshaw89: its not throughout the app its per request.

app/setup.py
```diff
@@ -231,10 +228,13 @@ def add_blueprints(application):
     csrf = CSRFProtect(application)
 
     # import and register the main application blueprint
-    from app.views.questionnaire import questionnaire_blueprint
+    from app.views.questionnaire import questionnaire_blueprint, post_submission_blueprint
     application.register_blueprint(questionnaire_blueprint)
     questionnaire_blueprint.config = application.config.copy()
 
+    application.register_blueprint(post_submission_blueprint)
```
jonnyshaw89: to maintain consistency i think the import should be above each blueprint?

app/views/questionnaire.py
```diff
@@ -273,19 +281,11 @@ def submit_answers(routing_path, eq_id, form_type, collection_id):
...
-
         get_questionnaire_store(current_user.user_id, current_user.user_ik).delete()
-        remove_survey_session_data()
 
-        return redirect(url_for(
-            '.get_thank_you',
-            eq_id=eq_id,
-            form_type=form_type,
-            collection_id=collection_id))
+        get_session_store().add_submission_time()
```
jonnyshaw89: do we want to be more explicit here and pass the time in?

app/views/session.py
```diff
@@ -87,21 +86,21 @@ def get_timeout_continue():
...
 @login_required
 def post_expire_session():
-    remove_survey_session_data()
+    logout_user()
     return get_session_expired()
 
 
 @session_blueprint.route('/session-expired', methods=['GET'])
 def get_session_expired():
     if current_user.is_active:
```
jonnyshaw89: Do we need this `if`? i think `logout_user()` checks this

jonnyshaw89 (changes requested):

app/globals.py
dwyeradam: Lazy cut and paste from questionnaire_store above. Corrected both.

app/authentication/authenticator.py
dwyeradam: This causes issues when trying to get the current_user for use in the get_answer_store in login() in session.py. We need a separate card now to ensure we're using the Flask login and logout effectively.

app/views/questionnaire.py
dwyeradam: I didn't but go on then. Passing in datetime.utcnow().isoformat() to follow runner standards.

app/authentication/session_storage.py
```diff
@@ -1,144 +1,111 @@
...
 
 logger = get_logger()
 
 
 class SessionStorage:
 
-    @staticmethod
-    def store_user_id(user_id):
+    def __init__(self):
+        self.eq_session = None
```
ajmaddaford: I think we need to simplify this:
- The class should expose the properties` user_id`, `eq_session_id` and `session_data` - this will remove the need for the `get_user_id` and `get_session_data` properties. The `session_data` property should be a `SessionData` class.
- We should just need `load`, `save` and `delete` methods. To update a session we would update the `session_data` property and call save.
- The eq_session_id from the cookie should be passed in, rather than having a dependency on the flask cookie session in this class

Also, we should rename this file session_store.py and move it alongside questionnaire_store.

app/authentication/session_storage.py
```diff
@@ -1,144 +1,111 @@
...
+        return json.dumps(vars(session_data))
+
+    def add_submission_time(self, submission_time):
+        """
+        add submission time to session data
+        """
+        eq_session = self._get_user_session()
+        if eq_session and eq_session.session_data:
+            session_data = json.loads(eq_session.session_data, object_hook=lambda d: SessionData(**d))
+            session_data.add_submitted_time(submission_time)
```
ajmaddaford: Can this just be `session_data.submitted_time = submission_time`?

ajmaddaford (changes requested):

app/authentication/session_storage.py
jonnyshaw89: i don't think we need `load`. We should do that on the init as we do with `QuestionnaireStore`

app/authentication/authenticator.py
```diff
@@ -28,14 +33,25 @@ def request_load_user(request):  # pylint: disable=unused-argument
     return load_user()
 
 
+@user_logged_out.connect_via(ANY)
+def when_user_logged_out(sender_app, user):  # pylint: disable=unused-argument
```
MIslam92, line 37: Could you not silence these by just adding a _prefix before each variable passed instead wherever we pass unused arguments?

MIslam92 (dismissed):

app/authentication/authenticator.py
ajmaddaford: I tried this and got an error when this method is called with a keyword argument that doesn't match the `_` prefixed parameters:
`when_user_logged_out() got an unexpected keyword argument 'user'`

ajmaddaford (approved):
@dwyeradam I made some changes to the pull request last week as I'm off this week. Please can you review the changes I've made.

dwyeradam (commented):
I've checked the changes Andrew made and i'm happy with the messing of my code.

jonnyshaw89 (approved):

MIslam92 (approved):
All above addressed

PR #1401, opened 2017-12-11
Corrected column types of date columns

jonnyshaw89:
questionnaire_state had incorrect column types for date columns

### How to review 
Launch runner using a new empty database.
it should startup and launch a survey without error.


MIslam92 (approved):
👍

ajmaddaford (approved):

Issue #1403, opened 2017-12-15
Error when resuming a survey when an answered answer has been removed from the schema


ajmaddaford:
### Expected behaviour
When I part complete a survey 
And one of the answers that I have completed is removed from the survey
Then when I resume the survey it should work

### Actual behaviour
A 500 error is generated.

This has been spotted in pre-production after the `religion-welsh-answer-other` was removed from the schema in https://github.com/ONSdigital/eq-survey-runner/pull/1337.

```
"exception": [
        "Traceback (most recent call last):",
        "  File '/opt/python/run/venv/local/lib64/python3.4/site-packages/flask/app.py', line 1612, in full_dispatch_request",
        "    rv = self.dispatch_request()",
        "  File '/opt/python/run/venv/local/lib64/python3.4/site-packages/flask/app.py', line 1598, in dispatch_request",
        "    return self.view_functions[rule.endpoint](**req.view_args)",
        "  File '/opt/python/current/app/app/views/session.py', line 75, in login",
        "    navigator = PathFinder(g.schema_json, get_answer_store(current_user), metadata)",
        "  File '/opt/python/current/app/app/globals.py', line 34, in get_answer_store",
        "    questionnaire_store.answer_store.upgrade(questionnaire_store.version, g.schema_json)",
        "  File '/opt/python/current/app/app/data_model/answer_store.py', line 260, in upgrade",
        "    answer_schema = SchemaHelper.get_answer_schema_for_answer_id(schema_json, answer['block_id'], answer['answer_id'])",
        "  File '/opt/python/current/app/app/helpers/schema_helper.py', line 211, in get_answer_schema_for_answer_id",
        "    return answer_schema_list[answer_id]",
        "KeyError: 'religion-welsh-answer-other'"
    ]
```

### Steps to reproduce the behaviour
- Part complete a survey
- Remove an answer from the schema that has been answered
- Resume the survey

PR #1404, opened 2017-12-18
Fix error when resuming a survey

jonnyshaw89:
When an answered answer has been removed from the schema, the upgrade now skips that answer as it cannot find it.

Fixes https://github.com/ONSdigital/eq-survey-runner/issues/1403

### DIFF
--- a/app/data_model/answer_store.py
+++ b/app/data_model/answer_store.py
@@ -259,13 +259,14 @@ class AnswerStore(object):
             for answer in self.answers:
                 answer_schema = SchemaHelper.get_answer_schema_for_answer_id(schema_json, answer['block_id'], answer['answer_id'])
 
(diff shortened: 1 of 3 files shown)


shahi645 (approved):
Tested and verified this works, test also passes

yoganandkunche (changes requested):
@jonnyshaw89 as discussed it fails with error 500 when the whole block is removed. Work when only the answer is removed.

shahi645:
@yoganandkunche I think thats what the issue is related to. Only when an answer is removed

yoganandkunche:
@shahi645 Yes the issue is only related to answer but it should also cater for the block removal, when we get a request from business comes to remove the whole question.

yoganandkunche (approved):
@jonnyshaw89 squishy-squashy?

PR #1405, opened 2017-12-19
Add a minimum exclusive value of 0 to total pay for each pay pattern

shahi645:
For each pay patterns total gross pay, the user may now no longer enter 0 as a value.

### How to review 
Run MWSS try total pay with `0` in each pay patter. Should give the correct validation error


yoganandkunche (dismissed):
@shahi645 the validation error messages need tweaking. When no value entered on mandatory field the app gives a validation error message : Enter an answer, even if it is 0. when entering '0' is not a valid entry

shahi645:
@yoganandkunche will look at this

jonnyshaw89:
@shahi645 This can't really be reviewed until we have the validator change

shahi645:
@jonnyshaw89 the fix can be reviewed, to see if it works.

MIslam92 (approved):

ajmaddaford (approved):

PR #1408, opened 2017-12-21
Enabling lint on Unit Tests

MIslam92:
I have added linting to run consistently on the unit tests.

### How to review 
Is all the linting fixed and passing in CI
run `find ./tests -type f -name "*.py" | xargs pipenv run pylint` to check all linting passes test_mwss_submission_data will be moved to a separate card so ignore this for now if it appears in lint errors.


tests/app/forms/test_household_relationship_form.py
```diff
@@ -131,14 +130,14 @@ def test_serialise_relationship_answers(self):
 
         expected_answers = [
             {
-                'group_id':"household-relationships",
+                'group_id': "household-relationships",
```
jonnyshaw89: do we not check the consistency of `'` and `"`?

tests/app/validation/test_number_range_validator.py
```diff
@@ -110,24 +110,25 @@ def test_within_range_at_max(self):
...
-                   'id': 'test-range',
-                   'type': 'Currency'
-                  }
+        answer = {
+            'min_value': {
+                "value": 10
+            },
+            'label': 'Min Test',
+            'guidance': '',
+            'options': [],
```
jonnyshaw89: only because you touched it, we shouldn't need to specify the values that we don't need.

jonnyshaw89 (changes requested):

tests/app/forms/test_household_relationship_form.py
shahi645: This should be picked up by linting as part of the PR done by Hamish

tests/app/validation/test_number_range_validator.py
```diff
@@ -154,24 +155,25 @@ def test_manual_min(self):
...
-                      },
-                   'description': '',
-                   'id': 'test-range',
-                   'type': 'Currency'
-                  }
+        answer = {
+            'max_value': {
+                "value": 20
+            },
+            'label': 'Max Test',
```
shahi645: @MIslam92 these use  `"` in our schemas, might be worth doing that here

tests/app/forms/test_household_relationship_form.py
ajmaddaford: Running pylint against the `tests` directory appears to not be linting all python files. If you run `find ./tests -type f -name "*.py" | xargs pipenv run pylint` a lot of lint errors (incluing ones for quotes) are generated.

ajmaddaford: From https://www.hurricanelabs.com/blog/pylint-the-secret-to-detecting-even-your-sneakiest-python-code-errors, without using xargs (and suppressing all output that isn't an error):
`find ./tests -name "*.py" -exec pipenv run pylint -E '{}' +`

.pylintrc
```diff
@@ -281,7 +281,7 @@ ignore-mixin-members=yes
 # (useful for modules/projects where namespaces are manipulated during runtime
 # and thus existing member attributes cannot be deduced by static analysis. It
 # supports qualified module names, as well as Unix pattern matching.
-ignored-modules=
+ignored-modules=test_mwss_submission_data
```
jonnyshaw89: why is this ignored?

jonnyshaw89 (changes requested):
The command in your description does not match the command run by `run_lint.sh` do we need to update this?

.pylintrc
MIslam92: How we deal with test_mwss_submission_data was going to be made into part of a separate card so for this PR thought it would be best to just ignore it. For the command in the description it seems to pick up more lint compared to what our `run_lint.sh` script does so might have to update it.

shahi645: @MIslam92 I think if this is what we are going to do, we need to update the run_lint script.

tests/app/forms/test_fields.py
```diff
@@ -9,7 +9,7 @@
 from app.data_model.answer_store import AnswerStore
 
 
-class TestFields(unittest.TestCase):
+class TestFields(unittest.TestCase):  # pylint: skip-file
```
shahi645: same as `test_mwss_submission`?

tests/app/validation/test_number_range_validator.py
```diff
@@ -1,15 +1,16 @@
+# pylint: disable=W,C,R
```
shahi645: is this needed if the pylint is skipping the file

shahi645 (changes requested):
Think we should definitely update the lint script to have the same linting as here

tests/app/forms/test_fields.py
MIslam92: These files are meant to be ignored as the changes suggested by lint to get the tests working cause more complexity and thus far less readable then whats already in place and working.

.pylintrc
jonnyshaw89: As @shahi645 we need to put this command into `run_lint.sh` otherwise the value will be lost as soon as this PR is closed

shahi645: agreed 👍

jonnyshaw89:
You need to add the following files to make sure that the standard lint command hits all of the files.
Without these files Python doesn't recognise these folders as modules
```
tests/app/confirmation_page/__init__.py
tests/app/forms/__init__.py
tests/app/templating/summary/__init__.py
tests/app/validation/__init__.py
```

jonnyshaw89:
@MIslam92 have you raised the card for test_mwss_submission_data?

MIslam92:
@jonnyshaw89 there might already be a card in a the backlog to look into work around current downstream tests etc. so Andrew will have a look into it whenever his free

shahi645 (dismissed):
Works as expected 👍

.pylintrc
ajmaddaford: The only lint rule that should be ignored in test_mwss_submission_data is for the number of asserts. Can we just use a `# pylint: disable=` in the test file?

tests/app/forms/test_fields.py
ajmaddaford: What lint rules fail? I don't think we should be skipping files, could we just disable the lint rules that cause issues?

tests/app/validation/test_number_range_validator.py
```diff
@@ -4,12 +4,12 @@
...
 from app.validation.error_messages import error_messages
 from app.validation.validators import NumberRange
-from wtforms.validators import ValidationError
 from app.forms.fields import get_number_field, CustomIntegerField
 from app.data_model.answer_store import Answer, AnswerStore
+from wtforms.validators import ValidationError
 
 
-class TestNumberRangeValidator(unittest.TestCase):
+class TestNumberRangeValidator(unittest.TestCase):  # pylint: skip-file
```
ajmaddaford: Same as comment above, we shouldn't be skipping files.

tests/functional/generate_pages.py
```diff
@@ -212,18 +217,19 @@ def long_names_required(question, num_questions):
...
-        ANSWER_GETTER.substitute({'answerName': prefix + 'day', 'answerId': answerId + '-day'}) + \
-        ANSWER_GETTER.substitute({'answerName': prefix + 'month', 'answerId': answerId + '-month'}) + \
-        ANSWER_GETTER.substitute({'answerName': prefix + 'year', 'answerId': answerId + '-year'})
+        ANSWER_GETTER.substitute({'answerName': prefix + 'day', 'answerId': answer_id + '-day'}) + \
+        ANSWER_GETTER.substitute({'answerName': prefix + 'month', 'answerId': answer_id + '-month'}) + \
+        ANSWER_GETTER.substitute({'answerName': prefix + 'year', 'answerId': answer_id + '-year'})
 
 
-def _write_month_year_date_answer(answer_name, answerId, prefix):
+def _write_month_year_date_answer(_answer_name, answer_id, prefix):
```
ajmaddaford: Can we just remove `_answer_name` as it's not used?

tests/functional/generate_pages.py
```diff
@@ -235,13 +241,13 @@ def find_kv(block, key, values):
     return False
 
 
-def process_block(block, dir_out, schema_data, spec_file, relative_require = '..'):
-    logger.debug("Processing Block: %s", block['id'])
+def process_block(block, dir_out, schema_data, required_spec_file, relative_require='..'):
```
ajmaddaford: Why has `spec_file` been renamed `required_spec_file`? It's confusing as it's not required (it only does something with it if it's passed in).

tests/integration/integration_test_case.py
```diff
@@ -50,16 +50,20 @@ def _setUpApp(self):
             'keys': {
                 EQ_USER_AUTHENTICATION_RRM_PRIVATE_KEY_KID: {'purpose': KEY_PURPOSE_AUTHENTICATION,
                                                              'type': 'private',
-                                                             'value': get_file_contents('third-party/sdc-rrm-authentication-signing-private-v1.pem')},
+                                                             'value': get_file_contents(
```
ajmaddaford: I think all of these would be clearer with a new line after the `{`, something like:
```
EQ_USER_AUTHENTICATION_RRM_PRIVATE_KEY_KID: {
    'purpose': KEY_PURPOSE_AUTHENTICATION,
    'type': 'private',		     
    ....
}
```

ajmaddaford (changes requested):

ajmaddaford (dismissed):

tests/functional/generate_pages.py
MIslam92: As the linting picked up on this and threw the below error: 
`"Shadows name 'spec file' from outer scope"`
So the name change got around it

tests/app/validation/test_number_validator.py
```diff
@@ -3,15 +3,16 @@
...
 from app.validation.error_messages import error_messages
 from app.validation.validators import NumberCheck, DecimalPlaces
-from wtforms.validators import StopValidation, ValidationError
 from app.forms.fields import get_number_field, CustomDecimalField, MAX_DECIMAL_PLACES
 from app.data_model.answer_store import AnswerStore
+from wtforms.validators import StopValidation, ValidationError
 
 
-class TestNumberValidator(unittest.TestCase):
+class TestNumberValidator(unittest.TestCase):  # pylint: skip-file
```
ajmaddaford: `skip-file` needs to be removed and replaced with the specific lint rule we're disabling

ajmaddaford (approved):

jonnyshaw89 (approved):

PR #1410, opened 2017-12-21
Close secrets and keys files after the content has been read

jonnyshaw89:
The Keys and Secrets file resources were not being closed after the file was read.
This PR ensures that they are closed

### DIFF
--- a/app/setup.py
+++ b/app/setup.py
@@ -64,8 +64,12 @@ def create_app(setting_overrides=None):  # noqa: C901  pylint: disable=too-compl
 
     application.eq = {}
 
-    secrets = yaml.safe_load(open(application.config['EQ_SECRETS_FILE']))
-    keys = yaml.safe_load(open(application.config['EQ_KEYS_FILE']))
+    with open(application.config['EQ_SECRETS_FILE']) as secrets_file:
+        secrets = yaml.safe_load(secrets_file)
+
+    with open(application.config['EQ_KEYS_FILE']) as keys_file:
+        keys = yaml.safe_load(keys_file)
+
     validate_required_secrets(secrets)
     validate_required_keys(keys, KEY_PURPOSE_SUBMISSION)
     application.eq['secret_store'] = SecretStore(secrets)


app/setup.py
```diff
@@ -64,8 +64,12 @@ def create_app(setting_overrides=None):  # noqa: C901  pylint: disable=too-compl
 
     application.eq = {}
 
-    secrets = yaml.safe_load(open(application.config['EQ_SECRETS_FILE']))
-    keys = yaml.safe_load(open(application.config['EQ_KEYS_FILE']))
+    with open(application.config['EQ_SECRETS_FILE']) as secrets_file:
+        secrets = yaml.safe_load(secrets_file)
+
+    with open(application.config['EQ_KEYS_FILE']) as keys_file:
+        keys = yaml.safe_load(keys_file)
```
shahi645, line 71: @jonnyshaw89 This may be a silly question but how does this ensure they get closed ?

jonnyshaw89: Have a google

shahi645: 🥇

shahi645 (approved):

app/setup.py
```diff
@@ -64,8 +64,12 @@ def create_app(setting_overrides=None):  # noqa: C901  pylint: disable=too-compl
 
     application.eq = {}
 
-    secrets = yaml.safe_load(open(application.config['EQ_SECRETS_FILE']))
-    keys = yaml.safe_load(open(application.config['EQ_KEYS_FILE']))
+    with open(application.config['EQ_SECRETS_FILE']) as secrets_file:
+        secrets = yaml.safe_load(secrets_file)
+
```
MIslam92, line 69: Should you not be calling the close method the file like this i.e. secrets.close()? as best practice or does it auto close?

MIslam92 (changes requested):

app/setup.py
jonnyshaw89: http://effbot.org/zone/python-with-statement.htm

MIslam92: ahhh i was looking at: http://www.tutorialspoint.com/python/python_files_io.htm but forgot bout the with part

MIslam92 (approved):

PR #1413, opened 2017-12-27
Use ChimpJS to run functional tests

jonnyshaw89:
ChimpJS simplifies how we run the functional tests and makes the running of them a lot more stable.
I have also removed the build matrix from travis as the tests run a lot faster.

### How to review 
Do the tests all still run.
Do they work reliably locally


shahi645 (approved):
So much better with chimp!

dwyeradam (changes requested):
@jonnyshaw89 and @ajmaddaford. We need to discuss if the loss of running suites and individual tests is worth the improvement in run time.

Personally running all the tests quickly and reliably is worth losing the suites for. We can also still run individual folders so that's much the same.
With regard to running an individual test then that is a loss. We could use the @watch tag but that's not as easy as just calling a test but would be great during the writing of the individual test.

.travis.yml
```diff
@@ -29,7 +23,7 @@ cache:
     - yarn
     - pip
     - directories:
-        - /home/travis/.pyenv/versions/
```
MIslam92: Could we not add something like a before_cache phase to delete unwanted files/dependencies/logs?

MIslam92 (changes requested):

.travis.yml
jonnyshaw89: I don't think this is relevant to this PR

ajmaddaford:
@jonnyshaw89 @dwyeradam happy for this to merge given the improvements to both speed and reliability. Adding @watch to tests to run individual ones seems like an acceptable compromise.

README.md
```diff
@@ -114,25 +112,25 @@ Command                                    | Task
...
 
-`--env=preprod` Will run the test against preprod.
-`--spec=mci` Will run a single test spec (`mci.spec.js`) instead of the entire suite.
-`--suite=core` Will run a suite of tests (core) instead of the entire suite.
+`--path=tests/functional/spec/components/` Will run all of the tests under this directory.
+
+These options can be combined with `test_functional` eg:
 
-These options can be combined with `test_functional` or `test_functional_sauce`, eg:
+`yarn test_functional -- --path tests/functional/spec/components/` Will run all of the component tests test.
```
ajmaddaford: I don’t think you need the extra `--` in the yarn command - `From Yarn 1.0 onwards, scripts don't require "--" for options to be forwarded. In a future version, any explicit "--" will be forwarded as-is to the scripts.`

ajmaddaford (changes requested):
Nice improvement 😃 
I think it would be useful to have something in the readme about how to use the @watch tag.

ajmaddaford (approved):

dwyeradam (approved):
The benefits outweigh the negatives.

MIslam92 (changes requested):
LGTM

MIslam92 (approved):

shahi645:
SQUISHY-SQUASHY

PR #1415, opened 2017-12-28
Added blinker as a dependency

jonnyshaw89:
The docker container for eq-survey-runner was failing to load blinker as a module

### How to review 
Run the container that is built by this PR and confirm that it starts correctly


ajmaddaford:
Closing this as blinker is still required as a dependency with this approach. New pull request raised to keep existing approach and add blinker as a dependency https://github.com/ONSdigital/eq-survey-runner/pull/1418

PR #1416, opened 2017-12-28
Fix Travis pipenv performance issue

jonnyshaw89:
By caching the Python versions that we use the we remove a 3(ish) minute build step that sets up the build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,8 @@ cache:
     - yarn
     - pip
     - directories:
-        - /home/travis/.pyenv/versions/
+        - /opt/python/
+        - /home/travis/.local/share/
 script:
     - set -e
     - yarn compile


shahi645 (dismissed):

MIslam92 (approved):

ajmaddaford (changes requested):
Before we merge this I think we need to understand how this works when we remove dependencies from the app, or downgrade the version of a dependency.

ajmaddaford:
When I remove a dependency locally from the Pipfile, it remains in the virtualenv. Only when I run the `pipenv uninstall` command is it removed from the virtualenv.

jonnyshaw89:
@ajmaddaford that isn't any different to how the pip cache used to work

ajmaddaford (approved):
Happy for this to merge on the basis it's the same thing that was happening previously when travis was caching the pip dependencies. This has highlighted the possible issue of the Python environment on Travis having dependencies that have been removed. To counter this we've raised a card to run the functional tests against the Docker container that gets built. This container is built cleanly on every build.

PR #1419, opened 2018-01-09
Eq 1749 remove unnecessary use of escape sequences in survey json

MIslam92:
Throughout the survey JSON escape sequences are used for non-standard characters like quotes a currency symbols. This makes them hard to read and author. All occurrences of these escape sequences should be replaced with the real characters.

### How to review 
Check all of the unnecessary escape sequences are escaped as well as the ones in the tests.

### DIFF
--- a/data/en/1_0112.json
+++ b/data/en/1_0112.json
@@ -225,7 +225,7 @@
                         "decimal_places": 2
                     }],
                     "id": "total-turnover-question",
-                    "title": "For the period {{ format_conditional_date (answers.period_from, exercise.start_date)}} to {{ format_conditional_date (answers.period_to, exercise.end_date)}}, what was the value of {{respondent.trad_as_or_ru_name}}\u2019s <em>total retail turnover</em>?",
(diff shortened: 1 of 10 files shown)


data/en/1_0112.json
```diff
@@ -289,7 +289,7 @@
                         "q_code": "146a",
                         "type": "Radio"
                     }],
-                    "description": "Please note: what constitutes a \u2018significant change\u2019 is dependent on your own interpretation in relation to {{respondent.trad_as_or_ru_name}}\u2019s figures from the previous reporting period and the same reporting period last year.<br><br>This information will help us to validate your data and should reduce the need to query any figures with you.",
+                    "description": "Please note: what constitutes a ‘significant change' is dependent on your own interpretation in relation to {{respondent.trad_as_or_ru_name}}'s figures from the previous reporting period and the same reporting period last year.<br><br>This information will help us to validate your data and should reduce the need to query any figures with you.",
```
jonnyshaw89, line 292: `‘significant change'` i don't think these closing quotes are correct.
There are many across this PR

jonnyshaw89 (changes requested):

PR #1420, opened 2018-01-09
Remove unnecessary use of escape sequences in survey JSON

MIslam92:
## **What is the context of this PR?**
Throughout the survey JSON escape sequences are used for non-standard characters like quotes a currency symbols. This makes them hard to read and author. All occurrences of these escape sequences should be replaced with the real characters.

## **How to review**
Check all of the unnecessary escape sequences are escaped as well as the ones in the tests.


dwyeradam (approved):
Also tested using BrowserStack in IE on Windows 7.

jonnyshaw89 (approved):

dwyeradam (changes requested):
I was a tiny bit premature with my approval. Only a little thing but you also need to update the .yaml files in data/sources/blocks that are using the escape sequences. Auto generated schemas will otherwise still have these escape sequences in them.

dwyeradam:
generate_schemas.py is still converting smart quotes into unicode escape sequences so we'll need to look at either changing that or possibly stop using it as it's only used to create two test schemas.

jonnyshaw89:
@MIslam92 how about the `yaml` files?

dwyeradam (changes requested):
We should remove the call in build.sh to generate_schemas as well.

dwyeradam (approved):

PR #1424, opened 2018-01-12
Move thank-you page back to original URL

jonnyshaw89:
The thank-you page was moved recently to `/questionnaire/thank-you` (https://github.com/ONSdigital/eq-survey-runner/pull/1391), but this breaks the analytics for survey completion. It needs to be moved to `/questionnaire/<eq_id>/<form_type>/thank-you`.

### How to review 
Has the `thank-you` URL changed back?

### DIFF
--- a/app/views/questionnaire.py
+++ b/app/views/questionnaire.py
@@ -43,11 +43,11 @@ questionnaire_blueprint = Blueprint(name='questionnaire',
 
 post_submission_blueprint = Blueprint(name='post_submission',
                                       import_name=__name__,
-                                      url_prefix='/questionnaire')
+                                      url_prefix='/questionnaire/<eq_id>/<form_type>/')
 
 
 @questionnaire_blueprint.before_request
-def before_request():
+def before_questionnaire_request():
     metadata = get_metadata(current_user)
     if not metadata:
         raise NoTokenException(401)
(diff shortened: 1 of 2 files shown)


app/views/questionnaire.py
```diff
@@ -43,11 +43,11 @@
 
 post_submission_blueprint = Blueprint(name='post_submission',
                                       import_name=__name__,
-                                      url_prefix='/questionnaire')
+                                      url_prefix='/questionnaire/<eq_id>/<form_type>/<collection_id>/')
```
dwyeradam: We don't need the Collection Exercise ID for analytics so should be removed. And therefore doesn't need to be added to session_data.

dwyeradam (changes requested):

app/views/questionnaire.py
```diff
@@ -59,8 +59,29 @@ def before_request():
...
+    session_data = session.session_data
+
+    logger.bind(tx_id=session_data.tx_id)
+
+    values = request.view_args
+    logger.bind(eq_id=values['eq_id'], form_type=values['form_type'],
+                ce_id=values['collection_id'])
+    logger.info('questionnaire request', method=request.method, url_path=request.full_path)
+
+    _check_same_survey(values['eq_id'], values['form_type'], values['collection_id'],
```
shahi645: 👍

app/views/questionnaire.py
```diff
@@ -59,8 +59,28 @@ def before_request():
                 ce_id=values['collection_id'])
     logger.info('questionnaire request', method=request.method, url_path=request.full_path)
 
+    _check_same_survey(values['eq_id'], values['form_type'], values['collection_id'],
+                       metadata['eq_id'], metadata['form_type'], metadata['collection_exercise_sid'])
```
ajmaddaford: We should use named parameters here

app/views/questionnaire.py
```diff
@@ -460,10 +483,9 @@ def update_questionnaire_store_with_answer_data(questionnaire_store, location, a
         questionnaire_store.completed_blocks.append(location)
 
 
-def _check_same_survey(eq_id, form_type, collection_id):
-    metadata = get_metadata(current_user)
-    current_survey = eq_id + form_type + collection_id
-    metadata_survey = metadata['eq_id'] + metadata['form_type'] + metadata['collection_exercise_sid']
+def _check_same_survey(current_eq_id, current_form_type, current_collection_id, stored_eq_id, stored_form_type, stored_collection_id):
```
ajmaddaford: I think it would be clearer if we named the values with url instead of current e.g.`url_eq_id` and session instead of stored e.g. `session_eq_id`

ajmaddaford (changes requested):

dwyeradam (approved):

ajmaddaford (approved):

PR #1428, opened 2018-01-19
Store the minimum required context for answers in the answer store

jonnyshaw89:
Only answer_id, group_instance and answer_instance should be needed.

Allows question to be moved between sections without breaking the existing answer store.

### How to review 
Would need to load previous version somehow (backwards compatible)

### DIFF
--- a/app/data_model/answer_store.py
+++ b/app/data_model/answer_store.py
@@ -6,22 +6,10 @@ import simplejson as json
 
 
 class Answer(object):
-    def __init__(self, group_id=None, block_id=None, answer_id=None, value=None, group_instance=0, answer_instance=0, location=None):
-        valid = (group_id or answer_id or block_id or value) is not None
+    def __init__(self, answer_id, value, group_instance=0, answer_instance=0):
+        if answer_id is None or value is None:
+            raise ValueError("Both 'answer_id' and 'value' must be set for Answer")
 
-        if location:
-            assert not (group_id or group_instance or block_id), \
-                'Expected either a location object or one or more of group_id, group_instance, block_id params'
-
-            group_id = location.group_id
-            group_instance = location.group_instance
-            block_id = location.block_id
-
-        if not valid:
-            raise ValueError("At least one of 'answer_id', 'group_id', 'block_id' or 'value' must be set for Answer")
-
-        self.group_id = group_id
-        self.block_id = block_id
         self.answer_id = answer_id
         self.group_instance = group_instance
         self.answer_instance = answer_instance
@@ -35,9 +23,7 @@ class Answer(object):
         :param answer: An answer to compare
         :return: True if both answers match, otherwise False.
         """
-        return self.group_id == answer.group_id and \
-            self.block_id == answer.block_id and \
-            self.answer_id == answer.answer_id and \
+        return self.answer_id == answer.answer_id and \
             self.group_instance == answer.group_instance and \
             self.answer_instance == answer.answer_instance
 
@@ -50,10 +36,8 @@ class Answer(object):
         """
 
         return self.matches(Answer(
-            answer_dict['group_id'],
-            answer_dict['block_id'],
             answer_dict['answer_id'],
-            '',
+            answer_dict['value'],
             answer_dict['group_instance'],
             answer_dict['answer_instance'],
         ))
@@ -168,25 +152,13 @@ class AnswerStore(object):
             escaped.append(answer)
         return self.__class__(existing_answers=escaped)
 
-    def filter_by_location(self, location):
-        """
-        Find all answers in the answer store for a given location
-
-        :param location: The location to filter results by
-        return: Return a list of answers which match the location
-        """
-        return self.filter(group_id=location.group_id, block_id=location.block_id, group_instance=location.
-                           group_instance)
-
-    def filter(self, group_id=None, block_id=None, answer_id=None, group_instance=None, answer_instance=None,
+    def filter(self, answer_ids, group_instance=None, answer_instance=None,
                limit=None):
         """
         Find all answers in the answer store for a given set of filter parameter matches.
         If no filter parameters are passed it returns a copy of the instance.
 
-        :param answer_id: The answer id to filter results by
-        :param block_id: The block id to filter results by
-        :param group_id: The group id to filter results by
+        :param answer_ids: The answer ids to filter results by
         :param answer_instance: The answer instance to filter results by
         :param group_instance: The group instance to filter results by
         :param limit: True | False Limit the number of answers returned
@@ -195,16 +167,14 @@ class AnswerStore(object):
         filtered = []
 
         filter_vars = {
-            'answer_id': answer_id,
-            'block_id': block_id,
-            'group_id': group_id,
+            'answer_id': answer_ids,
             'answer_instance': answer_instance,
             'group_instance': group_instance,
         }
 
         for answer in self.answers:
             matches = all(
-                answer[key] == value
+                answer[key] in value if isinstance(value, list) else answer[key] == value
                 for key, value in filter_vars.items()
                 if value is not None
             )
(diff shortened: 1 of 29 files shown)


app/data_model/answer_store.py
```diff
@@ -220,27 +210,17 @@ def clear(self):
         """
         self.answers.clear()
 
-    def remove(self, group_id=None, block_id=None, answer_id=None, group_instance=None, answer_instance=None):
+    def remove(self, answer_ids=None, group_instance=None, answer_instance=None):
         """
         Removes answer(s) *in place* from the answer store.
 
         :param answer_id: The answer id to filter results to remove
```
shahi645: remove/rename parameter

app/data_model/answer_store.py
```diff
@@ -6,22 +6,18 @@
...
 
 class Answer(object):
-    def __init__(self, group_id=None, block_id=None, answer_id=None, value=None, group_instance=0, answer_instance=0, location=None):
-        valid = (group_id or answer_id or block_id or value) is not None
+    def __init__(self, answer_id=None, value=None, group_instance=0, answer_instance=0, location=None):
+        valid = (answer_id or value) is not None
 
         if location:
-            assert not (group_id or group_instance or block_id), \
+            assert not group_instance, \
```
ajmaddaford: I think we should drop `location` now as we'd only be using it for the `group_instance` and that can be passed as a separate parameter.

app/data_model/answer_store.py
```diff
@@ -6,22 +6,18 @@
 
 
 class Answer(object):
-    def __init__(self, group_id=None, block_id=None, answer_id=None, value=None, group_instance=0, answer_instance=0, location=None):
-        valid = (group_id or answer_id or block_id or value) is not None
+    def __init__(self, answer_id=None, value=None, group_instance=0, answer_instance=0, location=None):
+        valid = (answer_id or value) is not None
```
ajmaddaford: I don't think this makes sense, shouldn't we always expect an answer_id and value?

shahi645 (approved):
Tested and verified. Can't see any problems with code, and removal of lines of code massively outweigh added lines...big 👍

app/templating/schema_context.py
```diff
@@ -45,10 +43,12 @@ def _map_alias_to_answers(aliases, answer_store, routing_path):
     return values
 
 
-def _is_answer_in_routing_path(routing_path, answer):
-    location = Location(answer['group_id'], answer['group_instance'], answer['block_id'])
+def _get_answers_in_path(schema, path):
```
ajmaddaford: This method is used in two places, I think it would make sense to add it to path_finder. Something like `get_answer_ids_on_routing_path`.

app/data_model/answer_store.py
```diff
@@ -168,25 +160,25 @@ def escaped(self):
             escaped.append(answer)
         return self.__class__(existing_answers=escaped)
 
-    def filter_by_location(self, location):
+    def filter_by_location(self, schema, location):
```
ajmaddaford: I think we should consider dropping this method and just using `filter`

app/data_model/answer_store.py
```diff
@@ -168,25 +160,25 @@ def escaped(self):
...
         return: Return a list of answers which match the location
         """
-        return self.filter(group_id=location.group_id, block_id=location.block_id, group_instance=location.
-                           group_instance)
+        answer_ids = schema.get_answer_ids_for_block(location.block_id)
 
-    def filter(self, group_id=None, block_id=None, answer_id=None, group_instance=None, answer_instance=None,
+        return self.filter(answer_ids=answer_ids, group_instance=location.group_instance)
+
+    def filter(self, answer_ids=None, group_instance=None, answer_instance=None,
```
ajmaddaford: Should answer_ids now be required? It doesn't make much sense to delete all answers for a given group or answer instance.

ajmaddaford (changes requested):

app/data_model/answer_store.py
```diff
@@ -6,22 +6,10 @@
 
 
 class Answer(object):
-    def __init__(self, group_id=None, block_id=None, answer_id=None, value=None, group_instance=0, answer_instance=0, location=None):
-        valid = (group_id or answer_id or block_id or value) is not None
+    def __init__(self, answer_id=None, value=None, group_instance=0, answer_instance=0):
```
ajmaddaford: We should drop the `None` default for `answer_id` and `value` to make it clear from the method signature that they aren't optional.

app/data_model/answer_store.py
```diff
@@ -168,25 +152,13 @@ def escaped(self):
...
-                           group_instance)
-
-    def filter(self, group_id=None, block_id=None, answer_id=None, group_instance=None, answer_instance=None,
+    def filter(self, answer_ids, group_instance=None, answer_instance=None,
                limit=None):
         """
         Find all answers in the answer store for a given set of filter parameter matches.
         If no filter parameters are passed it returns a copy of the instance.
 
         :param answer_id: The answer id to filter results by
```
ajmaddaford: answer_id is now answer_ids

app/questionnaire/navigation.py
```diff
@@ -98,7 +99,8 @@ def _build_single_navigation(self, section, current_group_id, first_location):
     def _build_repeating_navigation(self, repeating_rule, section, current_group_id, current_group_instance):
         group = self.schema.get_group(section['group_order'][0])
         first_location = Location(group['id'], 0, group['blocks'][0]['id'])
-        no_of_repeats = evaluate_repeat(repeating_rule, self.answer_store, self.routing_path)
+        answer_on_path = PathFinder.get_answer_ids_on_routing_path(self.schema, self.routing_path)
```
ajmaddaford: This variable name isn't very clear, it sounds singular and possibly boolean. It would be much clearer named something like `answer_ids_on_path`

app/questionnaire/path_finder.py
```diff
@@ -117,7 +117,8 @@ def _calculate_no_of_repeats(self, group, path):
         repeating_rule = self.schema.get_repeat_rule(group)
 
         if repeating_rule:
-            return evaluate_repeat(repeating_rule, self.answer_store, path)
+            answer_on_path = self.get_answer_ids_on_routing_path(self.schema, path)
```
ajmaddaford: Same as previous comment.

app/questionnaire/path_finder.py
```diff
@@ -241,6 +242,14 @@ def get_latest_location(self, completed_blocks, routing_path=None):
 
         return latest_location
 
+    @staticmethod
+    def get_answer_ids_on_routing_path(schema, path):
+        answer_ids_in_path = []
+        for location in path:
+            answer_ids_in_path.extend(schema.get_answers_by_id_for_block(location.block_id))
+
+        return answer_ids_in_path
```
ajmaddaford: For consistency we should use either `on` or `in`, not both.

app/questionnaire/rules.py
```diff
@@ -52,7 +50,7 @@ def evaluate_goto(goto_rule, metadata, answer_store, group_instance):
     return True
 
 
-def evaluate_repeat(repeat_rule, answer_store, current_path):
+def evaluate_repeat(repeat_rule, answer_store, answer_on_path):
```
ajmaddaford: Related to comment above, it would be clearer if `answer_on_path` was renamed `answer_ids_on_path`

app/templating/schema_context.py
```diff
@@ -16,20 +16,20 @@ def build_schema_context(metadata, aliases, answer_store, routing_path, group_in
...
     values = {}
     for alias, answer_info in aliases.items():
         matching_answers = []
 
-        answers = answer_store.filter(answer_id=answer_info['answer_id'], limit=True)
+        answers = answer_store.filter(answer_ids=[answer_info['answer_id']], limit=True)
 
         for answer in answers:
-            if _is_answer_in_routing_path(routing_path, answer):
+            if answer['answer_id'] in PathFinder.get_answer_ids_on_routing_path(schema, routing_path):
```
ajmaddaford: `get_answer_ids_on_routing_path` could be called outside of both loops? Or the answer_ids on the routing path could be passed in to build_schema_context removing the dependency on both schema and PathFinder.

ajmaddaford (changes requested):

ajmaddaford:
I couldn't test starting a survey on master and continuing on this branch as this branch doesn't have the session encryption changes from master.

jonnyshaw89:
@ajmaddaford ill rebase it so it has those changes

ajmaddaford (approved):

PR #1432, opened 2018-01-23
Make "Select all that apply" label on checkbox fieldsets consistent

hamishtaplin:
screen shot 2018-01-23 at 11 03 11

Adds a label to fieldsets of checkboxes that states "Select all that apply:".

- Introduces `.field__label` element of the `field` block.
- Added this to template for multiple-choice answers if the type is "checkbox".
- Removes existing instances of this text from schemas to prevent deuplication

### How to review 

- Navigate to "test_checkbox" schema.
- Confirm text/styling
- Confirm no existing schemas contain this text anymore in labels/descriptions.

### DIFF
--- a/data/en/1_0102.json
+++ b/data/en/1_0102.json
@@ -1,375 +1,435 @@
 {
-    "mime_type": "application/json/ons/eq",
-    "schema_version": "0.0.1",
-    "data_version": "0.0.1",
-    "survey_id": "023",
-    "title": "Monthly Business Survey - Retail Sales Index",
-    "description": "RSI Description",
-    "theme": "default",
-    "legal_basis": "StatisticsOfTradeAct",
-    "variables": {
-        "period": "{{ format_conditional_date (answers.period_from, exercise.start_date)}} to {{ format_conditional_date (answers.period_to, exercise.end_date)}}"
-    },
-    "groups": [{
-        "blocks": [{
-                "type": "Introduction",
-                "id": "introduction",
-                "primary_content": [{
-                    "id": "get-started",
-                    "content": [{
-                        "list": [
-                            "On average it takes 10 minutes to complete this survey once you’ve collected the information.",
-                            "Data should relate to all sites in England, Scotland and Wales.",
-                            "You can provide informed estimates if actual figures aren't available.",
-                            "We will treat your data securely and confidentially."
-                        ]
-                    }]
-                }],
-                "preview_content": {
-                    "id": "preview",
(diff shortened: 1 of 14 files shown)


data/en/1_0102.json
```diff
@@ -319,7 +319,7 @@
                         "type": "Checkbox"
                     }],
 
-                    "description": "Select all that apply",
+                    "description": "",
```
jonnyshaw89: These should fail validation but surprisingly don't.
if we don't have a value we should remove the description tag

shahi645 (changes requested):
There is a couple left out from `census_household`, including identity questions and religion, the central heating question also shows bold `select all that apply` text

data/en/1_0102.json
hamishtaplin: yeah I missed that one

hamishtaplin: done but there are 277 other instances of this across the schemas.

hamishtaplin:
@shahi645 There's one I hadn't spotted in the census schemas but the rest are in slightly different contexts I think. Mainly inside guidance.

jonnyshaw89:
@hamishtaplin you need to run `yarn format` as you have re-formatted one of the json files

hamishtaplin:
@jonnyshaw89 should probably add that process in a Git hook so nobody has to worry about it

data/en/1_0102.json
```diff
@@ -33,42 +33,43 @@
...
-                            "list": [
-                                "in-store / online promotions",
-                                "special events (e.g.  sporting events)",
-                                "calendar events (e.g.  Christmas, Easter, Bank Holiday)",
-                                "weather",
-                                "store closures/openings"
+                            "question": "Total retail turnover",
+                            "content": [{
+                                    "description": "Include:",
+                                    "list": ["VAT", "internet sales", "retail sales from outlets in Great Britain to customers abroad"]
```
jonnyshaw89, line 39: The formatting of the json seems to have changed how the lists are structured.
We should try to avoid formatting changes to the schemas as it makes it harder to see the actual changes for this PR

shahi645 (approved):
LGTM

data/en/1_0102.json
hamishtaplin: Agreed. If we set up a pre-commit hook to run the json formatter then it will eradicate accidental formatting changes so nobody has to worry about it. We do this with Prettier on author and it works great.

PR #1423, opened 2018-01-12
Rename block type from questionnaire to question

MIslam92:
### What is the context of this PR?
Changing of Questionnaire to Question now in blocks

### How to review 
Review this first: https://github.com/ONSdigital/eq-schema-validator/pull/14
Check all instances of Questionnaire have been changed and everything still functions i.e. tests etc.

PR #1434, opened 2018-01-24
Test Travis schema validation with broken schema

ajmaddaford:
A broken schema was recently merged in #1423 (e_commerce.json) that should have failed schema validation. This pull request re-adds the broken schema to see if it fails schema validation in Travis (it does fail when run locally).


ajmaddaford:
I've raised a card on the backlog to investigate this further

PR #1437, opened 2018-01-29
README updates for local setup

nealedj:
Update README instructions for local setup to include a step to install python3.4. Whilst setting up on a fresh machine I found that `pipenv install` was failing because it couldn't find python3.4.

Also when running `yarn` to install npm dependencies the fibers package was failing due to node-gyp not working on python3.

### How to review 
Remove python3.4 from your machine and follow the local setup steps. Verify that the app's pipenv gets set up correct.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,7 @@ Upgrade pip and install dependencies:
 
 ```
 brew install pyenv
+pyenv install
 pip install --upgrade pip setuptools pipenv
 pipenv install --dev
 ```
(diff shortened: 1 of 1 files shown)


shahi645 (approved):
I had this trouble also but had to use pipenv to set the path to python

MIslam92 (approved):

MIslam92:
@nealedj Squash and should be good to push

PR #1447, opened 2018-02-07
Add group_id and block_id back onto Answer

jonnyshaw89:
Make the Answer object forward compatible

### How to review 
When I start a survey on the old version of eQ, continue on the new version, and return to the old version, it should continue to work


MIslam92 (approved):

dwyeradam (approved):
Tested with flicking between master and current production tag and had the expected failures. These failures stopped when flicking between this branch and the current production tag.

PR #1449, opened 2018-02-07
removed redundant code and removed test testing redundant code

MIslam92:
Removing redundant code due to recent changes that accommodates for this.

Card: https://trello.com/c/4aJcpZAF/1812-remove-redundant-code

### How to review 
The usual

### DIFF
--- a/app/data_model/session_store.py
+++ b/app/data_model/session_store.py
@@ -1,4 +1,3 @@
-import re
 from structlog import get_logger
 from app.data_model.models import EQSession, db
 from app.data_model.session_data import SessionData
@@ -76,9 +75,9 @@ class SessionStore:
             self.user_id = self._eq_session.user_id
 
             if self._eq_session.session_data:
-                session_data = self._eq_session.session_data
-                if self._is_session_data_encrypted(session_data):
-                    session_data = StorageEncryption(self.user_id, self.user_ik, self.pepper).decrypt_data(session_data)
+                encrypted_session_data = self._eq_session.session_data
+                session_data = StorageEncryption(self.user_id, self.user_ik, self.pepper)\
+                    .decrypt_data(encrypted_session_data)
 
                 self.session_data = json.loads(session_data, object_hook=lambda d: SessionData(**d))
 
@@ -89,9 +88,3 @@ class SessionStore:
             logger.debug('eq_session_id not found in database', eq_session_id=self.eq_session_id)
 
         return self._eq_session
-
-    @staticmethod
-    def _is_session_data_encrypted(session_data):
-        # pylint: disable=anomalous-backslash-in-string
-        encrypted_token_regex = '^[\w\-]+?\.([\w\-]+)?\.[\w\-]+\.[\w\-]+\.[\w\-]+?$'
-        return bool(re.match(encrypted_token_regex, session_data))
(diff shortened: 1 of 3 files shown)


app/data_model/session_store.py
```diff
@@ -77,8 +77,7 @@ def _load(self):
 
             if self._eq_session.session_data:
                 session_data = self._eq_session.session_data
```
jonnyshaw89: Can we rename this to `encrypted_session_data`?

jonnyshaw89 (changes requested):

app/data_model/session_store.py
```diff
@@ -76,11 +76,10 @@ def _load(self):
...
 
             if self._eq_session.session_data:
-                session_data = self._eq_session.session_data
-                if self._is_session_data_encrypted(session_data):
-                    session_data = StorageEncryption(self.user_id, self.user_ik, self.pepper).decrypt_data(session_data)
+                encrypted_session_data = self._eq_session.session_data
+                encrypted_session_data = StorageEncryption(self.user_id, self.user_ik, self.pepper).decrypt_data(encrypted_session_data)
 
-                self.session_data = json.loads(session_data, object_hook=lambda d: SessionData(**d))
+                self.session_data = json.loads(encrypted_session_data, object_hook=lambda d: SessionData(**d))
```
dwyeradam: After getting **encrypted_session_data** the next variable should just be **session_data**. This shows that this step is around decrypting the stored encrypted data. So line 79 you get encrypted_session_data and then on line 80 you use that to create session_data which is then used to create the **self.session_data** on line 82.

app/data_model/session_store.py
```diff
@@ -76,11 +76,10 @@ def _load(self):
             self.user_id = self._eq_session.user_id
 
             if self._eq_session.session_data:
-                session_data = self._eq_session.session_data
-                if self._is_session_data_encrypted(session_data):
-                    session_data = StorageEncryption(self.user_id, self.user_ik, self.pepper).decrypt_data(session_data)
+                encrypted_session_data = self._eq_session.session_data
+                encrypted_session_data = StorageEncryption(self.user_id, self.user_ik, self.pepper).decrypt_data(encrypted_session_data)
```
dwyeradam: After getting encrypted_session_data the next variable should just be session_data. This shows that this step is around decrypting the stored encrypted data. So line 79 you get encrypted_session_data and then on line 80 you use that to create session_data which is then used to create the self.session_data on line 82.

dwyeradam (changes requested):

app/data_model/session_store.py
```diff
@@ -76,9 +76,9 @@ def _load(self):
             self.user_id = self._eq_session.user_id
 
             if self._eq_session.session_data:
-                session_data = self._eq_session.session_data
-                if self._is_session_data_encrypted(session_data):
```
dwyeradam, line 80: `def _is_session_data_encrypted(session_data)` is now no longer needed.

dwyeradam (changes requested):

app/data_model/session_store.py
MIslam92: @dwyeradam looking at my branch on Github its not there?
https://github.com/ONSdigital/eq-survey-runner/blob/eq-1812-remove-redundant-code/app/data_model/session_store.py#L78

dwyeradam: It's not that line that's the issue. It's the fact you still have the method in that the line was previously calling but is now no longer needed as nothing else calls it. 
https://github.com/ONSdigital/eq-survey-runner/blob/eq-1812-remove-redundant-code/app/data_model/session_store.py#L94

dwyeradam (approved):

jonnyshaw89 (approved):

PR #1462, opened 2018-02-14
Implement CSP headers

nealedj:
In order to mitigate against malicious attacks this PR adds [Flask-Talisman](https://github.com/GoogleCloudPlatform/flask-talisman) to handle the addition of security headers as recommended in the [Flask docs](http://flask-.readthedocs.io/en/latest/security/#security-headers). 

The main change here is that a [strict-dynamic](https://csp.withgoogle.com/docs/strict-csp.html) CSP header has been added.

Additionally Talisman has been extended to generate nonce value for each request that can be added to each request.

### How to review 
Run app and check that the following response headers are present:
- X-Frame-Options: DENY
- X-XSS-Protection: 1; mode=block
- X-Content-Type-Options: nosniff
- Content-Security-Policy: default-src 'self'; script-src 'self' strict-dynamic 'nonce-SOME-PER-REQUEST-NONCE'
- Referrer-Policy: strict-origin-when-cross-origin


jonnyshaw89:
This looks like a really positive change.
Are there going to be any issues with older browsers?

nealedj:
Older browsers that don't support CSP will simply not get the protection.

However, strict-dynamic isn't yet supported by Edge so we may want to consider explicitly specify trusted external hosts instead. That's been supported in modern browsers for quite some time. https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP#Browser_compatibility (see script-src and strict-dynamic).

app/setup.py
```diff
@@ -160,6 +164,18 @@ def override_url_for():  # pylint: disable=unused-variable
     return application
 
 
+def setup_secure_headers(application):
+    Talisman(
+        application,
+        content_security_policy=CSP_POLICY,
+        content_security_policy_nonce_in=['script-src'],
+        session_cookie_secure=not application.config['EQ_DEV_MODE'],
```
jonnyshaw89: We are trying to move away from the `EQ_DEV_MODE` master switch in favour of individual bespoke switches for specific features
So for this we would have something like `EQ_SECURE_SESSION_COOKIE` which would default to True (As we default to production setup) and then override to False in the dev settings

jonnyshaw89: Also does this mean we no longer need this? 
https://github.com/ONSdigital/eq-survey-runner/pull/1462/files#diff-03a224d9f3b49bcb5f5cd912c76ca683R321

app/talisman.py
```diff
@@ -0,0 +1,50 @@
...
+import random
+import string
+
+from flask import request
+from flask_talisman import Talisman as BaseTalisman
+
+NONCE_LENGTH = 16
+
+
+class Talisman(BaseTalisman):
```
jonnyshaw89, line 11: Is this something that we could potentially contribute back to flask-talisman?

jonnyshaw89 (changes requested):

app/setup.py
nealedj: Ah yeah - good call on the specific switch.

It looks like talisman just sets `SESSION_COOKIE_SECURE` internally anyway: https://github.com/GoogleCloudPlatform/flask-talisman/blob/master/flask_talisman/talisman.py#L189

It think it sounds sensible to allow the talisman object to keep control of this so that the secure headers stay handled in a single place. So yeah - let's remove our setting of `SESSION_COOKIE_SECURE`.

app/talisman.py
nealedj: Definitely - I didn't want to delay the implementation if we had to wait on the upstream contributions to get merged. Hopefully get this contributed back and we can remove our custom `Talisman` class.

app/settings.py
```diff
@@ -75,6 +75,7 @@ def get_env_or_fail(key):
 EQ_DEV_MODE = parse_mode(os.getenv('EQ_DEV_MODE', 'False'))
 EQ_ENABLE_CACHE = parse_mode(os.getenv('EQ_ENABLE_CACHE', 'True'))
 EQ_ENABLE_FLASK_DEBUG_TOOLBAR = parse_mode(os.getenv('EQ_ENABLE_FLASK_DEBUG_TOOLBAR', 'False'))
+EQ_SECURE_SESSION_COOKIE = parse_mode(os.getenv('EQ_SECURE_SESSION_COOKIE', 'False'))
```
jonnyshaw89: Just looking at this again we should probably follow the convention of having `ENABLE` in the name as its a boolean value.

jonnyshaw89: We should also use a production default and just override in Dev

config.py
```diff
@@ -14,7 +14,6 @@ class Config(object):
     DEBUG = True
```
jonnyshaw89: Is this file even still used?

jonnyshaw89 (changes requested):

app/setup.py
```diff
@@ -160,6 +164,18 @@ def override_url_for():  # pylint: disable=unused-variable
...
 
+def setup_secure_headers(application):
+    Talisman(
+        application,
+        content_security_policy=CSP_POLICY,
+        content_security_policy_nonce_in=['script-src'],
+        session_cookie_secure=not application.config['EQ_SECURE_SESSION_COOKIE'],
+        force_https=False,  # this is handled at the firewall
+        strict_transport_security=True,
+        strict_transport_security_max_age=31536000,
```
MIslam92: Do you not want to includeSubDomains for this?

config.py
nealedj: Good shout - doesn't look like it!

app/settings.py
nealedj: Ah, that was daft of me. Good catch!

MIslam92 (changes requested):

app/setup.py
nealedj: Yep that's included as the default HSTS value.

app/setup.py
```diff
@@ -29,20 +29,22 @@
...
     'Pragma': 'no-cache',
-    'Strict-Transport-Security': 'max-age=31536000; includeSubdomains',
-    'X-Frame-Options': 'DENY',
-    'X-Xss-Protection': '1; mode=block',
-    'X-Content-Type-Options': 'nosniff',
+}
+
+CSP_POLICY = {
+    'default-src': ["'self'", ],
+    'script-src': ["'self'", 'strict-dynamic', ]
```
MIslam92: I think you can add a `report-uri` property to this  as well so if the policy does get broken/violated we could debug though and have a log? Might actually be a separate card

MIslam92 (changes requested):

app/setup.py
```diff
@@ -29,20 +29,22 @@
...
     'Cache-Control': 'no-cache, no-store, must-revalidate',
     'Pragma': 'no-cache',
-    'Strict-Transport-Security': 'max-age=31536000; includeSubdomains',
-    'X-Frame-Options': 'DENY',
-    'X-Xss-Protection': '1; mode=block',
-    'X-Content-Type-Options': 'nosniff',
+}
+
+CSP_POLICY = {
+    'default-src': ["'self'", ],
```
MIslam92, line 46: You will also need to allow Google Analytics as we run those scripts and it seems like its violates the policy if not added

MIslam92 (changes requested):

app/setup.py
```diff
@@ -66,6 +68,8 @@ def create_app(setting_overrides=None):  # noqa: C901  pylint: disable=too-compl
     application = Flask(__name__, static_url_path='/s', static_folder='../static')
     application.config.from_object(settings)
 
+    setup_secure_headers(application)
```
jonnyshaw89: We should move this further down the setup. Especially after line 87 as that could change the behaviour.
I would probably put it after line 90 when we log out the version.

jonnyshaw89 (changes requested):

app/setup.py
nealedj: Ah yeah. strict-dynamic should allow GA to work but not all browsers support it.

I'm thinking that we should move to a whitelist TBH.

jonnyshaw89: And i guess we would want to add the CDN to this too, but we will do that when we integrate it.

jonnyshaw89 (changes requested):
There is an issue where images that are SVG in the CSS are not loading

jonnyshaw89:
screen shot 2018-02-15 at 12 04 12

MIslam92:
We could allow for data URIs to be used as a content src but by looks of things thats not very secure at all..

app/setup.py
```diff
@@ -44,8 +44,8 @@
 
 CSP_POLICY = {
     'default-src': ["'self'", ],
-    'script-src': ["'self'", 'strict-dynamic', ],
-    'img-src': ["'self'", "data:"]
+    'script-src': ["'self'", 'www.google-analytics.com', ],
```
jonnyshaw89: I think we can force this to only allow https too

nealedj: Fair enough, I can't think of a good reason why not.

app/settings.py
```diff
@@ -75,6 +75,7 @@ def get_env_or_fail(key):
 EQ_DEV_MODE = parse_mode(os.getenv('EQ_DEV_MODE', 'False'))
 EQ_ENABLE_CACHE = parse_mode(os.getenv('EQ_ENABLE_CACHE', 'True'))
 EQ_ENABLE_FLASK_DEBUG_TOOLBAR = parse_mode(os.getenv('EQ_ENABLE_FLASK_DEBUG_TOOLBAR', 'False'))
+EQ_ENABLE_SECURE_SESSION_COOKIE = parse_mode(os.getenv('EQ_SECURE_SESSION_COOKIE', 'True'))
```
jonnyshaw89: The env var name here is incorrect so i'm not sure setting it with have any effect.

jonnyshaw89 (changes requested):

app/settings.py
nealedj: Well spotted!

nealedj:
@MIslam92 Good suggestion. Let's consider that as a separate job and aim to hook it up to New Relic.

app/talisman.py
jonnyshaw89: We should raise this sooner rather than later because after this is done we will forget.

jonnyshaw89 (approved):

app/talisman.py
nealedj: Yep - PR is in https://github.com/GoogleCloudPlatform/flask-talisman/pull/18

jonnyshaw89: Fab 👍

MIslam92 (approved):

MIslam92:
@nealedj 'Squishy Squashy'

PR #1466, opened 2018-02-15
Deploy MWSS Content Changes

MIslam92:
Change to using the updated MWSS content in `1_0005.json`

Card: https://trello.com/c/AytnMjGi/1838-deploy-mwss-content-changes

Related: https://github.com/ONSdigital/go-launch-a-survey/pull/49

### How to review 
Open `1_0005.json` and check that it is the updated version & it displays all the correct content in Runner.


jonnyshaw89:
Have you ensured there are other tests that check the downstream values of different field types?

MIslam92:
@jonnyshaw89 By the looks of things in `test_converter.py` it checks the values for text areas and checkboxes.
Guess I should write some tests for other fields like radios etc.

nealedj (changes requested):
Yeah it makes sense to remove schema tests but let's make sure we don't lose any test coverage.

It looks like test_converter is only checking its processing of field types that it actually needs to do some conversion on.

It should also check field types that it does nothing to in order to assert that their values get passed without any change:

- Radio
- Number
- Percentage
- TextArea
- Currency

nealedj (approved):

nealedj:
Switching this to approved because there's some urgency to this change. Will raise the test coverage job separately.

jonnyshaw89 (approved):

PR #1467, opened 2018-02-15
Changes to support RAS Integration

jonnyshaw89:
Implemented the changed required to support the RAS integration as documented in this PR https://github.com/ONSdigital/ons-schema-definitions/pull/30

### How to review 
Are the new values present in the downstream data when launched with the version of launcher that supports these values https://github.com/ONSdigital/go-launch-a-survey/pull/50

### DIFF
--- a/app/data_model/session_data.py
+++ b/app/data_model/session_data.py
@@ -1,6 +1,19 @@
 class SessionData(object):
 
-    def __init__(self, tx_id, eq_id, form_type, period_str, language_code, survey_url, ru_name, ru_ref, submitted_time=None, **_):
+    def __init__(self,
+                 tx_id,
+                 eq_id,
+                 form_type,
+                 period_str,
+                 language_code,
+                 survey_url,
+                 ru_name,
+                 ru_ref,
+                 case_id=None,
+                 case_ref=None,
+                 account_service_url=None,
+                 submitted_time=None,
+                 **_):
         self.tx_id = tx_id
         self.eq_id = eq_id
         self.form_type = form_type
@@ -10,3 +23,6 @@ class SessionData(object):
         self.ru_name = ru_name
         self.ru_ref = ru_ref
         self.submitted_time = submitted_time
+        self.case_id = case_id
+        self.case_ref = case_ref
+        self.account_service_url = account_service_url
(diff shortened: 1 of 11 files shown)


app/data_model/session_data.py
```diff
@@ -1,6 +1,18 @@
...
+                 form_type,
+                 period_str,
+                 language_code=None,
+                 survey_url=None,
+                 ru_name=None,
+                 ru_ref=None,
+                 submitted_time=None,
+                 case_id=None,
+                 case_ref=None,
+                 account_service_url=None):
```
dwyeradam: We'll need to remember to consider forward compatibility for these new fields when used by the [session_store](https://github.com/ONSdigital/eq-survey-runner/blob/master/app/data_model/session_store.py#L82)

nealedj: Just chatted about this. Good point! Let's consider that a separate job and get the session data incompatibility fixed before the next deployment.

dwyeradam: Card raised: https://trello.com/c/mQrMxzDw

dwyeradam (approved):

app/views/errors.py
```diff
@@ -95,14 +96,21 @@ def get_tx_id():
...
+        account_url = cookie_session[ACCOUNT_URL]
+    return account_url
+
+
 def render_template(template_name):
     tx_id = get_tx_id()
     user_agent = user_agent_parser.Parse(request.headers.get('User-Agent', ''))
 
-    return render_theme_template(session.get('theme', 'default'),
+    return render_theme_template(cookie_session.get('theme', 'default'),
```
nealedj, line 110: It looks like there's no test coverage for the `respondent_account_url` getting set by the JWT payload.

nealedj (changes requested):

app/views/errors.py
jonnyshaw89: @nealedj i have added test coverage

app/views/errors.py
```diff
@@ -95,14 +96,21 @@ def get_tx_id():
     return tx_id
 
 
+def get_account_url():
+    account_url = current_app.config['RESPONDENT_ACCOUNT_URL']
+    if ACCOUNT_URL in cookie_session:
```
nealedj, line 101: If `cookie_session[ACCOUNT_URL` is present but null then `None` will be rendered in the `href`. We should check that there is a non-empty value for the URL.

nealedj (changes requested):

nealedj (approved):

PR #1468, opened 2018-02-19
SessionData to ignore new attributes

jonnyshaw89:
- SessionData should now be forward compatible

When new variables are added to session data during deployment if a session is created with the new code but then later accessed by the old code then an error is thrown.

### How to review 
Create a session on a version of the code with new session data (https://github.com/ONSdigital/eq-survey-runner/pull/1467) then return to this branch and you should be able to resume.

### DIFF
--- a/app/data_model/session_data.py
+++ b/app/data_model/session_data.py
@@ -1,6 +1,6 @@
 class SessionData(object):
 
-    def __init__(self, tx_id, eq_id, form_type, period_str, language_code=None, survey_url=None, ru_name=None, ru_ref=None, submitted_time=None):
+    def __init__(self, tx_id, eq_id, form_type, period_str, language_code, survey_url, ru_name, ru_ref, submitted_time=None, **_):
         self.tx_id = tx_id
         self.eq_id = eq_id
         self.form_type = form_type
--- a/tests/app/data_model/test_session_store.py
+++ b/tests/app/data_model/test_session_store.py
@@ -4,6 +4,7 @@ from mock import patch
 from app.data_model.session_store import SessionStore
 from app.data_model.session_data import SessionData
 from tests.app.app_context_test_case import AppContextTestCase
+from tests.app.data_model.TestSessionData import TestSessionData
 
 
 class SessionStoreTest(AppContextTestCase):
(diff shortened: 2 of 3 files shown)


app/data_model/session_data.py
```diff
@@ -1,6 +1,6 @@
 class SessionData(object):
 
-    def __init__(self, tx_id, eq_id, form_type, period_str, language_code=None, survey_url=None, ru_name=None, ru_ref=None, submitted_time=None):
+    def __init__(self, tx_id, eq_id, form_type, period_str, language_code=None, survey_url=None, ru_name=None, ru_ref=None, submitted_time=None, **kwargs):
```
dwyeradam: For everything other than **submitted_time** we should really remove the None defaults now. These were only needed for backward compatibility.

tests/app/data_model/test_session_store.py
```diff
@@ -66,3 +67,26 @@ def test_should_not_delete_when_no_session(self):
...
+        new_session_data = TestSessionData(
+            tx_id='tx_id',
+            eq_id='eq_id',
+            form_type='form_type',
+            period_str='period_str',
+            language_code=None,
+            survey_url=None,
+            ru_name='ru_name',
+            ru_ref='ru_ref',
+            additional_value='some cool new value you dont know about yet'
```
dwyeradam: I'll be the pedant `do not` rather than `dont`.

dwyeradam (changes requested):

tests/app/data_model/test_session_store.py
```diff
@@ -66,3 +67,26 @@ def test_should_not_delete_when_no_session(self):
...
+        new_session_data = TestSessionData(
+            tx_id='tx_id',
+            eq_id='eq_id',
+            form_type='form_type',
+            period_str='period_str',
+            language_code=None,
+            survey_url=None,
+            ru_name='ru_name',
+            ru_ref='ru_ref',
+            additional_value='some cool new value you do not know about yet'
```
dwyeradam, line 81: We should test with multiple additional variables.

dwyeradam (changes requested):

app/data_model/session_data.py
```diff
@@ -1,6 +1,6 @@
 class SessionData(object):
 
-    def __init__(self, tx_id, eq_id, form_type, period_str, language_code=None, survey_url=None, ru_name=None, ru_ref=None, submitted_time=None):
+    def __init__(self, tx_id, eq_id, form_type, period_str, language_code, survey_url, ru_name, ru_ref, submitted_time=None, **kwargs):
```
dwyeradam: `**kwargs` is failing linting. I believe `**_` is a valid alternative.

dwyeradam (approved):
Checked switching between this and **ras-integration-changes** branch and worked as expected.

nealedj (approved):

PR #1471, opened 2018-02-21
Add CDN to CSP header

nealedj:
Adds the ONS CDN domain to the CSP header.

### How to review 
- Check that the header contains the correct value
- Navigate to a page which shows an icon from the pattern library, check that the icons show and check that there are no errors in the console


jonnyshaw89 (approved):
Tested and works fine

dwyeradam (approved):

PR #1479, opened 2018-03-02
Enhance logic for section links

nealedj:
Delegates the logic for working how whether a section is not started, started, complete or skipped to a separate module. This is used by the navigation builder to work out which block to link to when a user clicks on a section link. It's also used to work out if a questionnaire has been completed and can be submitted.

- If a section has not been started the user should link to the first block.
- If it has been started then the user will link to the first incomplete block.
- If it has been completed and the final block is a summary or confirmation block then the user will link to the final block
- If it has been completed and the final block is not a summary or confirmation block then the user will link to the first block

### How to review 
- Test section navigation using surveys that make use of routing, skip conditions and repeated groups

### DIFF
--- a/app/globals.py
+++ b/app/globals.py
@@ -8,6 +8,7 @@ from app.settings import EQ_SESSION_ID, USER_IK
 from app.data_model.questionnaire_store import QuestionnaireStore
 from app.data_model.session_store import SessionStore
 from app.storage.encrypted_questionnaire_storage import EncryptedQuestionnaireStorage
+from app.questionnaire.completeness import Completeness
 
 logger = get_logger()
 
@@ -65,6 +66,23 @@ def get_answer_store(user):
 def get_completed_blocks(user):
     questionnaire_store = get_questionnaire_store(user.user_id, user.user_ik)
     return questionnaire_store.completed_blocks
+
+
+def get_completeness(user):
+    from app.helpers.path_finder_helper import path_finder
+
+    completeness_object = g.get('_completeness')
+
+    if completeness_object is None:
+        metadata = get_metadata(user)
+        answer_store = get_answer_store(user)
+        completed_blocks = get_completed_blocks(user)
+        routing_path = path_finder.get_full_routing_path()
+
+        completeness_object = g._completeness = Completeness(
+            g.schema, answer_store, completed_blocks, routing_path, metadata)
+
+    return completeness_object
 
 
 def get_dynamodb():
--- a/app/questionnaire/completeness.py
+++ b/app/questionnaire/completeness.py
@@ -1,0 +1,193 @@
+from app.questionnaire.location import Location
+from app.questionnaire.questionnaire_schema import QuestionnaireSchema
+from app.questionnaire.rules import evaluate_skip_conditions, get_number_of_repeats
+
+
+class Completeness:
+    NOT_STARTED = 'NOT_STARTED'
+    STARTED = 'STARTED'
+    SKIPPED = 'SKIPPED'
+    COMPLETED = 'COMPLETED'
+    INVALID = 'INVALID'
+
+    # for the purposes of deriving completeness all of the following states
+    # allow a block to be marked as complete so long as at least
+    # one item is complete
+    COMPLETED_STATES = (COMPLETED, INVALID, SKIPPED,)
+
+    def __init__(self, schema, answer_store, completed_blocks, routing_path, metadata):
+        self.answer_store = answer_store
+        self.completed_blocks = completed_blocks
+        self.routing_path = routing_path
+        self.metadata = metadata
+        self.schema = schema
+
+    def is_section_complete(self, section):
+        return self.get_state_for_section(section) == self.COMPLETED
+
+    def is_group_complete(self, group, group_instance=None):
+        return self.get_state_for_group(group, group_instance=group_instance) == self.COMPLETED
+
+    def is_block_complete(self, location):
+        return location in self.completed_blocks
+
+    def all_sections_complete(self):
+        return all(state in [self.COMPLETED, self.SKIPPED] for state in self._get_all_section_states())
+
+    def any_section_complete(self):
+        return any(state == self.COMPLETED for state in self._get_all_section_states())
+
+    def get_state_for_section(self, section):
+        if isinstance(section, str):
+            # lookup section by section ID
+            section = self.schema.get_section(section)
+
+        group_states = [
+            self.get_state_for_group(group) for group in section['groups']
+        ]
+
+        def eval_state(state_to_compare):
+            return (state == state_to_compare for state in group_states)
+
+        section_state = self.NOT_STARTED
+
+        if group_states:
+            if all(eval_state(self.SKIPPED)):
+                section_state = self.SKIPPED
+
+            elif all(eval_state(self.INVALID)):
+                section_state = self.INVALID
+
+            elif any(eval_state(self.STARTED)):
+                section_state = self.STARTED
+
+            elif all(state in self.COMPLETED_STATES for state in group_states):
+                section_state = self.COMPLETED
+
+        return section_state
+
+    def get_state_for_group(self, group, group_instance=None):
+        if isinstance(group, str):
+            # lookup group by group ID
+            group = self.schema.get_group(group)
+
+        if (QuestionnaireSchema.is_confirmation_group(group) or
+                QuestionnaireSchema.is_summary_group(group)):
+            # summary/confirmations are special cases as we don't want to
+            # render until the whole survey is complete. They're also never
+            # show as complete
+            return self.NOT_STARTED if self.all_sections_complete() else self.SKIPPED
+
+        if self._should_skip(group):
+            return self.SKIPPED
+
+        block_states = [
+            state for location, state in
+            self._get_block_states_for_group(group, group_instance)
+        ]
+
+        def eval_state(state_to_compare):
+            return (state == state_to_compare for state in block_states)
+
+        group_state = self.NOT_STARTED
+
+        if all(eval_state(self.SKIPPED)):
+            group_state = self.SKIPPED
+
+        elif not self.routing_path:
+            group_state = self.NOT_STARTED
+
+        elif all(eval_state(self.INVALID)):
+            group_state = self.INVALID
+
+        elif all(state in self.COMPLETED_STATES for state in block_states):
+            group_state = self.COMPLETED
+
+        elif any(eval_state(self.COMPLETED)):
+            group_state = self.STARTED
+
+        return group_state
+
+    def get_first_incomplete_location_in_survey(self):
+        incomplete_locations = (
+            self.get_first_incomplete_location_in_section(section)
+            for section in self.schema.sections
+        )
+        return next(filter(None, incomplete_locations), None)
+
+    def get_first_incomplete_location_in_section(self, section):
+        incomplete_locations = (
+            self.get_first_incomplete_location_in_group(group)
+            for group in section['groups']
+        )
+        return next(filter(None, incomplete_locations), None)
+
+    def get_first_incomplete_location_in_group(self, group, group_instance=None):
+        if self._should_skip(group):
+            return
+
+        incomplete_locations = (
+            location for location, state in
+            self._get_block_states_for_group(
+                group, group_instance=group_instance)
+            if state not in self.COMPLETED_STATES
+        )
+        return next(incomplete_locations, None)
+
+    def get_latest_location(self):
+        latest_location = self.routing_path[0]
+
+        if self.completed_blocks:
+            latest_location = self.get_first_incomplete_location_in_survey()
+
+            if not latest_location:
+                # survey is complete
+                latest_location = self.routing_path[-1]
+
+        return latest_location
+
+    def _get_block_states_for_group(self, group, group_instance=None):
+        if group_instance is not None:
+            start_instance = max_instance = group_instance
+        else:
+            max_instance = get_number_of_repeats(group, self.schema, self.routing_path, self.answer_store) - 1
+            start_instance = 0
+
+        for current_instance in range(start_instance, max_instance + 1):
+            for block in group['blocks']:
+                location = Location(group['id'], current_instance, block['id'])
+
+                if self._should_skip(block):
+                    state = self.SKIPPED
+                elif self._is_valid_for_completeness(block, location):
+                    state = self.COMPLETED if self.is_block_complete(location) else self.NOT_STARTED
+                else:
+                    # block is not a question block or is not on the routing path
+                    state = self.INVALID
+
+                yield location, state
+
+    def _get_all_section_states(self):
+        return (
+            self.get_state_for_section(section)
+            for section in self._get_all_question_sections()
+        )
+
+    def _get_all_question_sections(self):
+        return (
+            section for section in self.schema.sections
+            if not (
+                QuestionnaireSchema.is_summary_section(section) or
+                QuestionnaireSchema.is_confirmation_section(section))
+        )
+
+    def _should_skip(self, group_or_block):
+        return (
+            'skip_conditions' in group_or_block and
+            evaluate_skip_conditions(group_or_block['skip_conditions'], self.metadata, self.answer_store)
+        )
+
+    def _is_valid_for_completeness(self, block, location):
+        """Returns True if the given block should be checked for completeness
+        """
+        return block['type'] in ('Question', 'ConfirmationQuestion') and location in self.routing_path
--- a/app/questionnaire/navigation.py
+++ b/app/questionnaire/navigation.py
@@ -1,11 +1,10 @@
-import copy
 from collections import defaultdict
 
 from structlog import get_logger
 
+from app.questionnaire.completeness import Completeness
 from app.questionnaire.location import Location
-from app.questionnaire.path_finder import PathFinder
-from app.questionnaire.rules import evaluate_repeat, evaluate_skip_conditions
+from app.questionnaire.rules import evaluate_repeat, get_answer_ids_on_routing_path
 
 logger = get_logger()
 
@@ -15,12 +14,20 @@ class Navigation(object):
     Reads navigation config from the schema and returns collections of dicts that describe
     completion status of each group
     """
-    def __init__(self, schema, answer_store, metadata, completed_blocks, routing_path):
+    def __init__(
+            self,
+            schema,
+            answer_store,
+            metadata,
+            completed_blocks,
+            routing_path,
+            completeness):
         self.schema = schema
         self.metadata = metadata
         self.answer_store = answer_store
         self.completed_blocks = completed_blocks
         self.routing_path = routing_path
+        self.completeness = completeness
 
     def build_navigation(self, current_group_id, current_group_instance):
         """
@@ -34,8 +41,6 @@ class Navigation(object):
         if navigation_block is None or navigation_block.get('visible', True) is False:
             return None
 
-        summary_or_confirmation_blocks = self.schema.get_summary_and_confirmation_blocks()
-
         navigation = []
 
         for section in self.schema.sections:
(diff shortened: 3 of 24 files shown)


app/questionnaire/completeness.py
```diff
@@ -0,0 +1,161 @@
...
+        incomplete_locations = (
+            location for location, is_complete in
+            self._get_block_states_for_group(
+                group, group_instance=group_instance)
+            if not is_complete
+        )
+        return next(incomplete_locations, None)
+
+    def is_block_valid_on_path(self, location, block):
+        if block['type'] == 'Question':
```
jonnyshaw89: if type isn't `Question` this will return `None`?
Should it not return `false`?

app/questionnaire/completeness.py
```diff
@@ -0,0 +1,161 @@
...
+            return location in self.routing_path
+
+    def _get_block_states_for_group(self, group, group_instance=None):
+        if group_instance is not None:
+            current_instance = max_instance = group_instance
+        else:
+            max_instance = self._calculate_no_of_repeats(group) - 1
+            current_instance = 0
+
+        while current_instance <= max_instance:
```
jonnyshaw89: Does this need to be a `while` loop?
we have all the information upfront so it could just be a `for`?

app/questionnaire/completeness.py
```diff
@@ -0,0 +1,161 @@
...
+                location = Location(group['id'], current_instance, block['id'])
+
+                if not self.is_block_valid_on_path(location, block):
+                    continue
+
+                yield location, self.is_block_complete(location)
+
+            current_instance += 1
+
+    def _calculate_no_of_repeats(self, group):
```
jonnyshaw89: This logic is identical to that on `PathFinder` Can't this be reused?

app/questionnaire/navigation.py
```diff
@@ -34,36 +38,40 @@ def build_navigation(self, current_group_id, current_group_instance):
...
 
-            if first_block_id in summary_or_confirmation_blocks \
-                    and self._can_reach_summary_confirmation(first_group_id, first_block_id) is False:
+            if is_summary_or_confirmation_section and \
+                    not self._can_reach_summary_confirmation(section):
                 continue
 
-            repeating_rule = self.schema.get_repeat_rule(
-                self.schema.get_group(first_group_id))
+            # assume that a repeating group will always be the first group in a section
```
jonnyshaw89: This assumption is incorrect for `census_household`
https://github.com/ONSdigital/eq-survey-runner/blob/33d2b28556cb2cd4c3cb20eae178e78a95bd7d72/data/en/census_household.json#L3662

data/en/1_0005.json
```diff
@@ -362,6 +362,20 @@
                                             "voluntary employee contributions to, and annual profit from, profit related pay schemes"
                                         ]
                                     },
+                                    {
```
jonnyshaw89: Is this change related?

jonnyshaw89 (changes requested):

data/en/1_0005.json
nealedj: No - looks like a bad merge. Will fix.

app/questionnaire/completeness.py
nealedj: Not sure - I think might be clearer as a `while` loop in this instance. Doing this as a for loop would be:

`for current_instance in range(start_instance, max_instance + 1):`

That feels less readable to me.

app/questionnaire/completeness.py
nealedj: Done

jonnyshaw89:
@nealedj the PR description says "If is has been completed then the user will link to the final block." I have tested this on `test_navigation_completeness.json` and it takes me to the last block in the section

jonnyshaw89:
When completing `census_household.json` the visitors section appears before I have been asked about visitors. when i click on the visitors link i get redirected to the first incomplete block of the survey and not the 1st block of the visitors section

jonnyshaw89 (changes requested):
A few observations above

dwyeradam (changes requested):
I tried the e-commerce survey as that's got a lot of routing/skip conditions. When you select No to the initial question you get routed around the whole of the rest of the survey straight to the comments section at the end. However when this happens the first Section title goes from the Navigation on the right.

nealedj:
@jonnyshaw89 RE:

> @nealedj the PR description says "If is has been completed then the user will link to the final block." I have tested this on test_navigation_completeness.json and it takes me to the last block in the section

That wasn't quite right - I've updated the description now.

jonnyshaw89:
@nealedj in `census_household.json` the visitors section is still visible when it shouldn't be.

jonnyshaw89:
The pre submission completeness check still checks all block types. if i navigate to another group at the end of a section, when i go to submission page, i am redirected back the the interstitial that was not submitted. only Question pages should be required to be submitted

tests/app/questionnaire/test_completeness.py
```diff
@@ -8,6 +8,16 @@
 
 
 class TestCompleteness(AppContextTestCase): # pylint: disable=too-many-public-methods
+    def test_no_groups(self):
```
jonnyshaw89: `test_section_with_no_groups_not_marked_as_complete_when_.....`??

app/helpers/completeness_helper.py
```diff
@@ -0,0 +1,26 @@
...
+from flask_login import current_user, login_required
+from werkzeug.local import LocalProxy
+
+from app.globals import get_answer_store, get_metadata, get_completed_blocks
+from app.helpers.path_finder_helper import path_finder
+from app.questionnaire.completeness import Completeness
+
+
+@login_required
+def get_completeness():
```
jonnyshaw89: I think this should live in `globals.py` as that is the place we currently house this sort of logic. 
The "helpers" are currently more utility methods or wrappers

jonnyshaw89: This could even be made as a wrapper eg `path_finder_helper.full_routing_path_required`??

app/questionnaire/navigation.py
```diff
@@ -22,6 +22,9 @@ def __init__(self, schema, answer_store, metadata, completed_blocks, routing_pat
         self.completed_blocks = completed_blocks
         self.routing_path = routing_path
 
+        self.completeness = Completeness(
```
jonnyshaw89: Why not use your new get_completeness() method?
I think its off that we have a method that constructs this for us yet we don't consistently use it?
To quote 'O-Town' I think we should have it "All or nothing at all"

app/questionnaire/navigation.py
```diff
@@ -34,147 +37,128 @@ def build_navigation(self, current_group_id, current_group_instance):
...
-                    and self._can_reach_summary_confirmation(first_group_id, first_block_id) is False:
+            if is_summary_or_confirmation_section and \
+                    not self._can_reach_summary_confirmation(section):
                 continue
 
-            repeating_rule = self.schema.get_repeat_rule(
-                self.schema.get_group(first_group_id))
+            # if the first group in a section is a repeating group then repeat the section
+            # navigation link for each repeat instead of rendering the section title
+            repeating_rule = self.schema.get_repeat_rule(non_skipped_groups[0])
```
jonnyshaw89: I think we should also change this to sections that contain one 1 group that repeats.

app/questionnaire/questionnaire_schema.py
```diff
@@ -130,6 +130,42 @@ def get_questions_for_block(block_json):
         for question_json in block_json.get('questions', []):
             yield question_json
 
+    @staticmethod
+    def is_summary_section(section):
+        """Summary section is a section with a single Summary group
+        """
+        if section['groups']:
```
jonnyshaw89: If we change validator to require at least 1 group for a section then we don't need this check.

app/questionnaire/questionnaire_schema.py
```diff
@@ -130,6 +130,42 @@ def get_questions_for_block(block_json):
...
+            return QuestionnaireSchema.is_summary_group(section['groups'][0])
```
jonnyshaw89: I think we should check all groups in the section, as we don't currently restrict a summary to be in its own section. I don't think.

app/questionnaire/questionnaire_schema.py
```diff
@@ -130,6 +130,42 @@ def get_questions_for_block(block_json):
...
+
+        return False
+
+    @staticmethod
+    def is_summary_group(group):
+        """Summary group is a group with a single block of type "Summary"
+        """
+        if group['blocks']:
+            return group['blocks'][0]['type'] == 'Summary'
```
jonnyshaw89: Same 2 comments as above. we shouldn't have a group with no blocks and we don't stop summary being in a group with other blocks so we should check all the blocks

app/questionnaire/questionnaire_schema.py
```diff
@@ -130,6 +130,42 @@ def get_questions_for_block(block_json):
...
+
+        return False
+
+    @staticmethod
+    def is_confirmation_section(section):
+        """Confirmation section is a section with a single confirmation group
+        """
+        if section['groups']:
+            return QuestionnaireSchema.is_confirmation_group(section['groups'][0])
```
jonnyshaw89: same again

app/questionnaire/questionnaire_schema.py
```diff
@@ -130,6 +130,42 @@ def get_questions_for_block(block_json):
...
+
+        return False
+
+    @staticmethod
+    def is_confirmation_group(group):
+        """Confirmation group is a group with a single block of type "Confirmation"
+        """
+        if group['blocks']:
+            return group['blocks'][0]['type'] == 'Confirmation'
```
jonnyshaw89: and again

app/views/questionnaire.py
```diff
@@ -285,9 +287,9 @@ def get_view_submission(eq_id, form_type):  # pylint: disable=unused-argument
     return redirect(url_for('post_submission.get_thank_you', eq_id=eq_id, form_type=form_type))
 
 
-def _redirect_to_latest_location(routing_path, collection_id, eq_id, form_type):
-    latest_location = path_finder.get_latest_location(get_completed_blocks(current_user),
-                                                      routing_path=routing_path)
+def _redirect_to_latest_location(collection_id, eq_id, form_type):
+    latest_location = completeness.get_latest_location()
```
jonnyshaw89: I think if we make `get_completeness` a wrapper then it would be more obvious where `completeness` is coming from.

app/views/questionnaire.py
```diff
@@ -306,7 +308,8 @@ def _render_page(full_routing_path, block, current_location, post_form=None):
 
 
 def _is_valid_location(routing_path, location):
-    return location in routing_path
+    is_survey_complete = not completeness.get_first_incomplete_location_in_survey()
+    return location in routing_path or is_survey_complete
```
jonnyshaw89: Im not sure how this helps?

data/en/census_household.json
```diff
@@ -3668,7 +3668,7 @@
                             "when": [{
                                 "id": "overnight-visitors-answer",
                                 "condition": "equals",
-                                "value": "0"
+                                "value": 0
```
jonnyshaw89, line 3669: 👍

data/en/test_0203.json
```diff
@@ -8,334 +8,346 @@
...
-                                }
-                            ],
-                            "description": "If possible, this should be for the period {{exercise.start_date|format_date}} to {{exercise.end_date|format_date}}.",
-                            "id": "reporting-period-question",
-                            "title": "What are the dates of the sales period you are reporting for?",
-                            "type": "DateRange"
-                        },
-                        {
-                            "guidance": {
+            "id": "default-section",
```
jonnyshaw89: Why have we changed this schema?

data/en/test_0205.json
```diff
@@ -8,361 +8,372 @@
...
-                                    "q_code": "12",
-                                    "label": "Period to",
-                                    "type": "Date",
-                                    "mandatory": true
-                                }
-                            ]
-                        },
-                        {
-                            "guidance": {
+            "id": "default-section",
```
jonnyshaw89: and this one?

tests/app/questionnaire/test_completeness.py
```diff
@@ -0,0 +1,483 @@
...
+            Location('coffee-group', 0, 'coffee'),
+            Location('coffee-group', 0, 'response-yes'),
+        ]
+        progress = Completeness(
+            schema, AnswerStore(), completed_blocks, routing_path, metadata={})
+
+        progress_value = progress.get_state_for_section('coffee-section')
+        self.assertEqual(Completeness.COMPLETED, progress_value)
+
+    def test_all_sections_complete_none(self):
```
jonnyshaw89: The test name doesn't make sense on its own, until you read the subsequent 2 tests

tests/app/questionnaire/test_completeness.py
```diff
@@ -0,0 +1,483 @@
...
+            Location('coffee-group', 0, 'coffee'),
+            Location('coffee-group', 0, 'response-yes'),
+            Location('toast-group', 0, 'toast'),
+        ]
+
+        progress = Completeness(
+            schema, AnswerStore(), completed_blocks, routing_path, metadata={})
+        self.assertTrue(progress.all_sections_complete())
+
+    def test_any_section_complete_none(self):
```
jonnyshaw89, line 115: same comment about the name

tests/app/questionnaire/test_completeness.py
```diff
@@ -0,0 +1,483 @@
...
+            Location('coffee-group', 0, 'coffee'),
+            Location('coffee-group', 0, 'response-yes'),
+            Location('toast-group', 0, 'toast'),
+        ]
+
+        progress = Completeness(
+            schema, AnswerStore(), completed_blocks, routing_path, metadata={})
+        self.assertFalse(progress.get_first_incomplete_location_in_survey())
+
+    def get_first_incomplete_location_in_survey_completed_skipped_interstitials(self):
```
jonnyshaw89: doesn't this test also test everything tested in the test above?

tests/app/questionnaire/test_completeness.py
```diff
@@ -0,0 +1,483 @@
...
+            Location('payment-details', 0, 'security-code'),
+            Location('payment-details', 0, 'security-code-interstitial'),
+            Location('extra-cover-items-group', 0, 'extra-cover-items'),
+            Location('confirmation-group', 0, 'confirmation'),
+        ]
+        progress = Completeness(
+            schema, AnswerStore(), completed_blocks, routing_path, metadata={})
+        self.assertFalse(progress.get_first_incomplete_location_in_survey())
+
+    def test_get_latest_location_no_completed_blocks(self):
```
jonnyshaw89, line 477: I wonder if we should be specifying what we expect on these tests?
`test_get_latest_location_no_completed_blocks_returns_first_location`

tests/app/questionnaire/test_questionnaire_schema.py
```diff
@@ -204,3 +204,67 @@ def test_group_has_questions_returns_false_when_group_doesnt_have_questionnaire_
...
+            }]
+        }
+
+        schema = QuestionnaireSchema(survey_json)
+        self.assertTrue(schema.is_confirmation_section(schema.get_section('section-1')))
+        self.assertTrue(schema.is_confirmation_group(schema.get_group('group-1')))
+        self.assertFalse(schema.is_summary_section(schema.get_section('section-1')))
+        self.assertFalse(schema.is_summary_group(schema.get_group('group-1')))
+
+    def test_is_summary_confirmation_no_groups(self):
```
jonnyshaw89: We shouldn't be able to have this scenario

tests/app/questionnaire/test_questionnaire_schema.py
```diff
@@ -204,3 +204,67 @@ def test_group_has_questions_returns_false_when_group_doesnt_have_questionnaire_
...
+        survey_json = {'sections': [{'id': 'section-1', 'groups': []}]}
+        schema = QuestionnaireSchema(survey_json)
+        self.assertFalse(schema.is_confirmation_section(schema.get_section('section-1')))
+        self.assertFalse(schema.is_summary_section(schema.get_section('section-1')))
+
+    def test_is_summary_confirmation_no_blocks(self):
```
jonnyshaw89: We shouldn't be able to have this scenario

jonnyshaw89 (dismissed):

app/helpers/completeness_helper.py
nealedj: Probably makes more sense as a function in `globals`

app/questionnaire/navigation.py
nealedj: Ha - yep good point. I missed this when putting in that function.

dwyeradam (changes requested):
I have just tried the census household schema. Upon completing the visitors section and pressing continue I get routed back to the first uncompleted block which is correct. However the highlighting on Section navigation isn't working when this happens.

app/questionnaire/questionnaire_schema.py
nealedj: @jonnyshaw89 Given that we now have a section summary. Perhaps we should add this restriction? I think I put in this assumption for that reason.

app/questionnaire/questionnaire_schema.py
nealedj: Good point 👍

app/questionnaire/questionnaire_schema.py
jonnyshaw89: rename it to `is_final_summary_section` and check the block type is `Summary` not `SectionSummary`?

app/views/questionnaire.py
nealedj: I did that for a reason... but I can't remember what it was now. I'll find out and clarify.

app/views/questionnaire.py
nealedj: I did that for a reason... but I can't remember what it was now. I'll find out and clarify.

tests/app/questionnaire/test_completeness.py
nealedj: Yes, true.

tests/app/questionnaire/test_completeness.py
nealedj: I find that can result in really long method names. Prefer that the method name describes the scenario and the assertion should indicate the expected outcome.

Thoughts? @jonnyshaw89

jonnyshaw89: Yeh, i'm not die hard either way.
Our code is scattered with examples of both.

nealedj:
@MIslam92 @shahi645 @MebinAbraham @davecharles 

I've added you all as requested reviewers. Please focus primarily on functionally testing the application with these changes rather than on the code changes.

app/questionnaire/navigation.py
```diff
@@ -34,147 +42,131 @@ def build_navigation(self, current_group_id, current_group_instance):
         if navigation_block is None or navigation_block.get('visible', True) is False:
             return None
 
-        summary_or_confirmation_blocks = self.schema.get_summary_and_confirmation_blocks()
-
         navigation = []
 
         for section in self.schema.sections:
             non_skipped_groups = self._get_non_skipped_groups(section)
             if not non_skipped_groups:
```
dwyeradam, line 48: Why aren't we using the Completeness class for this method? At the moment Navigation still seems to be evaluating skip conditions when that has already been done by the Completeness class. Ideally completeness would be the one stop shop for defining what state a section, group or block is in.

dwyeradam (changes requested):

davecharles (commented):
I've tested your changes with the following schemas and have no observations:
- census_household.json
- test_skip_condition.json
- test_skip_condition_block.json
- test_skip_condition_group.json
- test_routing_on_multiple_select.json
- test_routing_date_*.json

Also checked with labour_force.json and behaviour is as expected.

MIslam92 (commented):
I've tested your changes with the following schemas and seems to behave the way as stated in the PR/Card:

- `1_0005.json` (Including the Introduction in the Nav however does not get marked as complete when you start the survey probs due to its type being 'Introduction'?)
- `census_household.json`
- `e_commerce.json`
- `test_navigation.json`
- `test_navigation_*.json`
- `test_section_summary.json`

MebinAbraham:
@nealedj Can you update your branch, I'm getting behaviours that should no longer happen (ie new PRs that went in since you last updated).

I've merged locally. Will update if I find anything.

MebinAbraham (changes requested):
This PR #1530 seems to cause issues with your section completeness.
Tested schema: `1_0005.json`
If you try to visit a block that should no longer be in the routing path, it currently lets me visit that block and when I try to continue it will throw a 500 error.

If that's not very clear, you can recreate the error following this:
1. Launch `1_0005.json`
2. On `Pay Pattern` section choose 2 options, ie `Weekly` and `Fortnightly`
3. Complete the `Weekly Pay` and `General Comments` section.
4. Go to `Fortnightly Pay` section and copy the URL.
5. Go to `Pay Pattern` section and uncheck `Fortnightly Pay` section and click Continue.
6. Now visit the `Fortnightly` section using the URL.

No. 6 above should take you to 'Summary' since all sections are complete and not the section that shouldn't be in the routing path. If you also click `Continue` after No. 6, it will throw an error 500.

It is working as expected on master. Sorry for making it long-winded.

MIslam92:
@MebinAbraham I just tested your scenario and don't get a 500 error and can submit as normal?

MIslam92:
@MebinAbraham I just re ran your scenario seems to work fine for me and no 500's?

app/questionnaire/navigation.py
nealedj: Have addressed this now.

nealedj:
Good catch @MebinAbraham. I recreated that and fix it in fb5cf509fc7222ba7029cdb485f5990aaea284cc

app/questionnaire/navigation.py
```diff
@@ -34,147 +41,115 @@ def build_navigation(self, current_group_id, current_group_instance):
...
 
     def _get_non_skipped_groups(self, section):
         return [
-            group['id'] for group in self._get_visible_groups_for_section(section)
-            if not self._should_skip_group(group)
+            group for group in section['groups']
+            if self.completeness.get_state_for_group(group) != Completeness.SKIPPED
         ]
 
     def _should_skip_group(self, group):
```
dwyeradam: `_should_skip_group` method is no longer needed. Which will also help your code coverage.

dwyeradam (changes requested):

app/questionnaire/navigation.py
nealedj: Oh I thought I'd deleted that - ta!

app/globals.py
```diff
@@ -67,6 +68,23 @@ def get_completed_blocks(user):
...
+
+    completeness_object = g.get('_completeness')
+
+    if completeness_object is None:
+        metadata = get_metadata(user)
+        answer_store = get_answer_store(user)
+        completed_blocks = get_completed_blocks(user)
+        routing_path = path_finder.get_full_routing_path()
+
+        completeness_object = g.completeness = Completeness(
```
shahi645: should be `_completeness`, after your explanation yesterday?

app/questionnaire/completeness.py
```diff
@@ -0,0 +1,188 @@
...
+
+        def eval_state(state_to_compare):
+            return (state == state_to_compare for state in block_states)
+
+        group_state = self.NOT_STARTED
+
+        if all(eval_state(self.SKIPPED)):
+            group_state = self.SKIPPED
+
+        elif not self.routing_path:
```
shahi645, line 97: A lot of duplication here. Not very familiar with methods within methods. Can these be simplified into the one method?

shahi645 (changes requested):

shahi645:
@nealedj Also flake 8 failing on lint script for an unused import

app/globals.py
nealedj: Oh yeah - well remembered!

app/questionnaire/completeness.py
nealedj: In this case, not that I can think of at the moment. Although the checks look similar the logic for deriving a group's state from its blocks states is different to the logic that derive a section's state from its groups.

nealedj:
@MebinAbraham @dwyeradam @shahi645 ready for re-review

tests/functional/spec/navigation.spec.js
```diff
@@ -200,7 +200,6 @@ describe('Navigation', function() {
     .click(SecurityCodePage.submit())
     .click(SkipInterstitialPage.no())
     .click(SkipInterstitialPage.submit())
-    .then(helpers.isSectionComplete('Payment Details')).should.eventually.be.false
```
dwyeradam, line 203: Really you should have changed this to true as your code changes the behaviour here so Interstitial pages no longer count towards section completeness. Not a blocker though but if any further changes are made it might be worth changing this.

MebinAbraham (commented):
Quick question. If a section has 2 questions and the first question is mandatory. Should you be able to view the Section Summary (SS) without completing all the questions in that section?

Cause at the moment, I am able to view the SS without answering the mandatory question. If I click `Continue` on the SS I would go to the next Section and once that is complete it would route me back to the first mandatory question in the first section.

Just wondering if this is normal or after answering the 2nd question in the first section, I should be routed back to the first mandatory question I bypassed?

nealedj:
@MebinAbraham good call. No, that shouldn't happen. Although it wasn't introduced as part of this PR. It's reasonably minor because the main summary still won't allow the questionnaire to be submitted if there are incomplete blocks.

I'll raise a separate card to address this.

app/questionnaire/completeness.py
dwyeradam: @nealedj, are you sure that we can't reduce the duplication around eval_state. That looks the same for group and block. @shahi645 be strong and force your point.

shahi645: It would save a good few lines of code 😉

dwyeradam: Actually @shahi645 I led us up a garden path. David is correct. There are two subtle differences that I missed. Groups also uses routing path to evaluate NOT_STARTED and an extra step for checking COMPLETED. Lines 97/98, and 103/104 are group only checks.

shahi645: Thats fine! I thought originally we might be able to, was worth asking the question

dwyeradam (approved):
This does conflict slightly with return to Summary after editing a question. In this PR section completeness doesn't rely on SectionSummary and Interstitial pages being complete. However routing back to summary does. Ignoring interstitial in routing back to summary would be an issue if the last block is an interstitial as it would never be seen as you'd skip it straight to summary. Although I don't like tech debt I think this will need more thought and may be better as a PR in it's own right. Current working only gives a bad user experience in an edge case.

MebinAbraham (approved):
Apart from the issues identified that can be raised as separate cards. 
Looks good to me from a functional perspective.
👍

app/questionnaire/navigation.py
```diff
@@ -200,9 +162,58 @@ def _generate_link_names(self, label_answer_ids):
...
+    def _get_location_for_group(self, group, group_instance=None):
+        """Gets the location for a link used for a repeating group
+        """
+        group_completeness = self.completeness.get_state_for_group(group, group_instance=group_instance)
+        first_incomplete = self.completeness.get_first_incomplete_location_in_group(group, group_instance=group_instance)
+
+        return self._get_location_for_completeness(
+            group_completeness,
+            group,
+            group,
```
shahi645, line 188: Should this be reapeated_group or something?

shahi645:
@nealedj LGTM apart from that little comment i made but otherwise 👍

app/questionnaire/navigation.py
nealedj: You mean because the same value is passed twice? No, in this case that's intentional. This method gets a link for a repeating group, all of the possible locations are within the same group so the value passed for `first_group` and `last_group` are both the same.

shahi645: Ahhh ok yes that was it, in which case this is G2G

shahi645 (approved):

PR #1480, opened 2018-03-05
Removed logic from templates

jonnyshaw89:
Constructed a view model will all information required to render the page

The model passed to the templates is now a dict and contains all of the information required to render the page.
Also implemented is the ability to get to json back from any questionnaire endpoint, by only specifying in the request that the client accepts json

### How to review 
Does everything continue to work.
When adding the accepts `application/json` header then all of the context is returned as json

### DIFF
--- a/app/helpers/form_helper.py
+++ b/app/helpers/form_helper.py
@@ -31,7 +31,7 @@ def get_form_for_location(schema, block_json, location, answer_store, disable_ma
 
         data = deserialise_composition_answers(answers)
 
-        return generate_household_composition_form(schema, block_json, data), None
+        return generate_household_composition_form(schema, block_json, data)
 
     elif location.block_id in ['relationships', 'household-relationships']:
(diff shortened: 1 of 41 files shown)


nealedj (approved):
Nicely done. `request_wants_json` is a good addition - plus it made it easier to verify the context.

app/helpers/form_helper.py
```diff
@@ -31,7 +31,7 @@ def get_form_for_location(schema, block_json, location, answer_store, disable_ma
 
         data = deserialise_composition_answers(answers)
 
-        return generate_household_composition_form(schema, block_json, data), None
+        return generate_household_composition_form(schema, block_json, data), {}
```
dwyeradam: Can we remove the return of the empty dict here. Also the same on line 65. Only line 46 returns values and should be made part of the form.

dwyeradam (changes requested):

dwyeradam (approved):

PR #1481, opened 2018-03-05
Runner to use CDN

saedwards:
How to test

Requires thorough manual regression testing by a few people. All variations of the user interface need to be tested. Any deviation in changes to the following media types need to be considered when testing:
- Confirm all script behaviour is correct and in line with behaviour seen in the pattern library.
- Confirm all css styles are correct and in line with the pattern library.
- Confirm all font types are still working.
- Confirm all images are referenced correctly from CSS and HTML.
- Confirm all existing EQ themes work.
- Confirm all favicons are referenced correctly.


jonnyshaw89:
👍 Loving the amount of files deleted.

app/setup.py
```diff
@@ -43,9 +43,9 @@
 }
 
 CSP_POLICY = {
-    'default-src': ["'self'", ],
-    'script-src': ["'self'", 'https://www.google-analytics.com', ],
-    'img-src': ["'self'", 'data:', 'https://www.google-analytics.com', 'https://cdn.ons.gov.uk']
+    'default-src': ["'self'", 'https://cdn.ons.gov.uk', ],
```
nealedj, line 46: This will require the `test_enforces_secure_headers()` test to be updated.

nealedj (dismissed):
Nicely done!

app/templates/layouts/_base.html
```diff
@@ -114,6 +121,8 @@
 
     <!--[if gte IE 9]><!-->
     {% block scripts %}
+      {% set scripts_cdn = cdn_url_prefix~"/scripts/bundle.js" %}
```
jonnyshaw89: We should make sure that we are using the minified assets
eg `bundle.min.js`

jonnyshaw89 (changes requested):

MebinAbraham:
The console is picking up quite a few of these.
```Refused to load the font '<URL>' because it violates the following Content Security Policy directive: "default-src 'self' <URL>". Note that 'font-src' was not explicitly set, so 'default-src' is used as a fallback.```

The cdn has already been added to the CSP header?

Edit:
This fixes it:
ie. Adding `font-src`
```
CSP_POLICY = {
    'default-src': ["'self'", 'https://cdn.ons.gov.uk', ],
    'font-src': ["'self'", 'data:', 'https://cdn.ons.gov.uk', ],
    'script-src': ["'self'", 'https://www.google-analytics.com', 'https://cdn.ons.gov.uk', ],
    'img-src': ["'self'", 'data:', 'https://www.google-analytics.com', 'https://cdn.ons.gov.uk', ]
}
```

**Comment by armstrongb** [comment]
When adding another member of the household in Census Household the following error is reported in console:

`main.js:1460 Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive: "script-src 'self' https://www.google-analytics.com https://cdn.ons.gov.uk 'nonce-UHrZw2zNBt8J3kr9'".`

`value @ main.js:1460
(anonymous) @ main.js:1538`

**Comment by armstrongb** [comment]
Should it be that the footer is still using `footer2` as a class? In the pattern library it is no longer using this class and has been re-factored.

**Comment by shahi645** [comment]
@saedwards could you do a rebase, looking to check a couple of things that need matching with master

**Review by MIslam92 — COMMENTED** [review]
Picked up in logs to do with CSP just noting it here will need to add to CSP in separate PR.

`Refused to connect to 'https://www.google-analytics.com/r/collect' because it violates the following Content Security Policy directive: "default-src 'self' https://cdn.ons.gov.uk". Note that 'connect-src' was not explicitly set, so 'default-src' is used as a fallback.`

**Comment by MebinAbraham** [comment]
To add to @armstrongb's comment, quite a few of these are shown for the census.
```
value @ VM4010 bundle.min.js:1
(anonymous) @ VM4010 bundle.min.js:1
2VM4053 bundle.min.js:1 Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive: "script-src 'self' https://www.google-analytics.com https://cdn.ons.gov.uk 'nonce-mVfKcYcGqNNAloIj'".
```
If you run the functional tests for the census and perserve the console logs, you can see all these.

**Review by MIslam92 — APPROVED** [review]

File: app/setup.py [hunk 3]
```diff
@@ -43,9 +43,10 @@
...
+    'font-src': ["'self'", 'data:', 'https://cdn.ons.gov.uk', ],
+    'script-src': ["'self'", 'https://www.google-analytics.com', 'https://cdn.ons.gov.uk', ],
+    'img-src': ["'self'", 'data:', 'https://www.google-analytics.com', 'https://cdn.ons.gov.uk', ]
 }
```
nealedj: We're also missing a `connect-src` policy here with `'https://www.google-analytics.com'`.

nealedj (changes requested):
The unsafe-eval issue raised by @armstrongb can be fixed by correcting a bug in the `HouseholdMember` module.

https://github.com/ONSdigital/eq-survey-runner/blob/master/app/assets/js/app/modules/household.js#L62

This line should be:

`window.setTimeout(() -> this.node.classList.remove('is-hidden'), 100)`

app/templates/layouts/_base.html
```diff
@@ -114,6 +121,8 @@
 
     <!--[if gte IE 9]><!-->
     {% block scripts %}
+      {% set scripts_cdn = cdn_url_prefix~"/scripts/bundle.min.js" %}
+      <script src="{{ scripts_cdn }}"></script>
       <script src="{{ url_for('static', filename='js/bundle.js') }}"></script>
```
jonnyshaw89, line 126: Do we have 2 called `bundle`? one on the CDN and one with the app?

app/setup.py
saedwards: Changes for the above are now available to review.

app/templates/layouts/_base.html
saedwards: Changes for the above are now available to review.

saedwards:
@armstrongb The footer2 component change is a separate change to the pattern library integration. Require refactoring the HTML to match the pattern library.

app/setup.py
saedwards: As discussed with Sam the `connect-src` for `https://www.google-analytics.com` for be a separate change for an existing issue.

saedwards: @nealedj I certainly think you've found found a bug in the household module. The execution of this expression does need to be deferred but it appears the issue is related to the content security policy. This is an existing issue that already appears in census pages on preprod: https://preprod-surveys-launch.eq.ons.digital/

app/templates/layouts/_base.html
saedwards: We certainly do
...for now

saedwards: The same goes for styles

nealedj (approved):

saedwards:
@shahi645 Rebase done

saedwards:
@armstrongb unused eval problem for household is fixed and ready for review.

armstrongb (approved):
Census household issue appears to be resolved.

jonnyshaw89:
When unselecting a checkbox the focus remains on the outer box. Im not sure if this is new from this work or an existing issue?
screen shot 2018-03-08 at 08 46 07

jonnyshaw89:
I think the above may be an existing issue. Maybe something to now fix in the pattern library?

MebinAbraham (approved):

saedwards:
@jonnyshaw89 I don't think this is an issue at all. If you unselect a checkbox, you are still focused on the checkbox, think this is an unrelated discussion. The aforementioned behaviour already exists in preprod.

jonnyshaw89 (approved):

shahi645 (approved):
Seems all good to me. Good work @saedwards !!! 🍾

bennpearson (changes requested):
Lato font missing in ie8. This is an issue with the Pattern Lib though. Missing `.eot?#iefix") format("embedded-opentype")` from typography.scss.

Happy to get this merged and add separate issue for the Pattern Lib

hamishtaplin (approved):

bennpearson (approved):

dwyeradam (approved):

PR #1484, opened 2018-03-06
Routing based on dates

dwyeradam:
As a user
I want to only answer questions which are applicable to me based on a date I have provided
So that I can skip any questions to reduce burden and to ensure I am able to answer them

Date compared with previous answer, metadata or hard coded date including today.
Offset_by allows for you to compare against a date in the past or future. Offset must allow for days, months and years.

### How to review 
test schemas:
test_routing_date_equals.json
test_routing_date_greater_than.json
test_routing_date_less_than.json
test_routing_date_not_equals.json

### DIFF
--- a/app/questionnaire/rules.py
+++ b/app/questionnaire/rules.py
@@ -1,5 +1,6 @@
 import logging
-
+from datetime import datetime
+from dateutil.relativedelta import relativedelta
 MAX_REPEATS = 25
 
 logger = logging.getLogger(__name__)
@@ -34,6 +35,62 @@ def evaluate_rule(when, answer_value):
         result = True
 
     return result
+
+
+def evaluate_date_rule(when, answer_store, group_instance, metadata, answer_value):
+    result = False
+
+    date_comparison = when['date_comparison']
+
+    answer_value = convert_to_datetime(answer_value)
+    match_value = get_date_match_value(date_comparison, answer_store, group_instance, metadata)
+    condition = when.get('condition')
+
+    if not answer_value or not match_value or not condition:
+        return False
+
+    if condition == 'equals' and match_value == answer_value:
+        result = True
+    elif condition == 'not equals' and match_value != answer_value:
+        result = True
+    elif condition == 'greater than' and answer_value > match_value:
+        result = True
+    elif condition == 'less than' and answer_value < match_value:
+        result = True
+
+    return result
+
+
+def get_date_match_value(date_comparison, answer_store, group_instance, metadata):
+    match_value = None
+
+    if 'value' in date_comparison:
+        if date_comparison['value'] == 'now':
+            match_value = datetime.utcnow().strftime('%Y-%m-%d')
+        else:
+            match_value = date_comparison['value']
+    elif 'id' in date_comparison:
+        match_value = get_answer_store_value(date_comparison['id'], answer_store, group_instance)
+    elif 'meta' in date_comparison:
+        match_value = get_metadata_value(metadata, date_comparison['meta'])
+
+    match_value = convert_to_datetime(match_value)
+
+    if 'offset_by' in date_comparison and match_value:
+        offset = date_comparison['offset_by']
+        match_value = match_value + relativedelta(days=offset.get('days', 0),
+                                                  months=offset.get('months', 0),
+                                                  years=offset.get('years', 0))
+
+    return match_value
+
+
+def convert_to_datetime(value):
+    date_format = '%Y-%m-%d'
+    if value and len(value) == 7:
+        date_format = '%Y-%m'
+
+    return datetime.strptime(value, date_format) if value else None
 
 
 def evaluate_goto(goto_rule, metadata, answer_store, group_instance):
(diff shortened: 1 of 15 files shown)


scripts/test_schemas.sh
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
-docker pull onsdigital/eq-schema-validator
-validator="$(docker run -d -p 5001:5000 onsdigital/eq-schema-validator)"
+docker pull onsdigital/eq-schema-validator:eq-1500-routing-based-on-dates
+validator="$(docker run -d -p 5001:5000 onsdigital/eq-schema-validator:eq-1500-routing-based-on-dates)"
```
dwyeradam: This will need to be reverted before merging.

MebinAbraham:
FYI: The schema titles don't match the schema itself.

Everything else seems to be working as expected.

app/questionnaire/rules.py
```diff
@@ -13,7 +14,8 @@ def evaluate_rule(when, answer_value):
     :return:
     """
     match_value = when['value'] if 'value' in when else None
-    condition = when['condition']
```
nealedj: The rest of the method depends on the `condition` variable being defined. Possibly better to use `when.get('condition')`?

app/questionnaire/rules.py
```diff
@@ -36,6 +38,53 @@ def evaluate_rule(when, answer_value):
...
+                                                  years=offset.get('years', 0))
+
+    answer_value = convert_to_datetime(answer_value)
+    condition = when['condition']
+
+    if condition == 'equals' and match_value == answer_value:
+        result = True
+    elif condition == 'not equals' and match_value != answer_value:
+        result = True
+    elif condition == 'greater than' and isinstance(answer_value, datetime) and answer_value > match_value:
```
nealedj: Should `answer_value` not always be a `datetime`? If it isn't then perhaps it'd be better to throw an error than to silently return `False`.

app/questionnaire/rules.py
```diff
@@ -36,6 +38,53 @@ def evaluate_rule(when, answer_value):
...
+
+    return result
+
+
+def convert_to_datetime(value):
+    date_format = '%Y-%m-%d'
+    if value and len(value) == 7:
+        date_format = '%Y-%m'
+
+    return datetime.strptime(value, date_format) if value else None
```
nealedj, line 93: Perhaps consider using `dateutil.parser.parse` here.

data/en/test_routing_date_equals.json
```diff
@@ -0,0 +1,199 @@
...
+                        "description": "",
+                        "id": "comparison-date-question",
+                        "title": "",
+                        "type": "General",
+                        "guidance": {
+                            "content": [{
+                                "title": "If you enter 31/03/2018 the following dates will be valid",
+                                "list": [
+                                    "Yesterday 30/03/2018",
+                                    "Today 28/03/2018",
```
nealedj: Today would be 31/03/2018

app/questionnaire/rules.py
```diff
@@ -36,6 +38,53 @@ def evaluate_rule(when, answer_value):
...
+        else:
+            match_value = date_comparison['value']
+    elif 'id' in date_comparison:
+        match_value = get_answer_store_value(date_comparison['id'], answer_store, group_instance)
+    elif 'meta' in date_comparison:
+        match_value = get_metadata_value(metadata, date_comparison['meta'])
+
+    match_value = convert_to_datetime(match_value)
+
+    if 'offset_by' in date_comparison and match_value:
```
nealedj, line 79: If `match_value` is `None` then this will error if the code tries to compare it against `answer_value` using `<` or `>` operators.

Consider returning immediately if `match_value` is `None`.

app/questionnaire/rules.py
```diff
@@ -36,6 +38,53 @@ def evaluate_rule(when, answer_value):
     return result
 
 
+def evaluate_date_rule(when, answer_store, group_instance, metadata, answer_value):
```
nealedj, line 40: Consider refactoring the logic to work out `match_value` into a separate function.

nealedj (changes requested):

app/questionnaire/rules.py
shahi645, line 40: bit picky but should be called `evalute_date_comparison_rule`

app/questionnaire/rules.py
dwyeradam: condition is a required value for when in schema validator so this should never be none so i'll actually remove the 'if'.

app/questionnaire/rules.py
shahi645: Agree with this, then only have evaluation in this method

app/questionnaire/rules.py
dwyeradam: This was to deal with a None answer value. Now i'm just going to return early for either answer_value, match_value or condition being None. This will satisfy your other comment above.

app/questionnaire/rules.py
```diff
@@ -120,25 +169,33 @@ def evaluate_when_rules(when_rules, metadata, answer_store, group_instance):
     :return: True if the when condition has been met otherwise False
     """
     for when_rule in when_rules:
+        value = None
```
shahi645: This is unused

app/questionnaire/rules.py
dwyeradam: It is simpler but when a yyyy-mm style string (2018-02) is passed in it would return todays day and not 01.

shahi645 (changes requested):

data/en/test_routing_date_equals.json
```diff
@@ -0,0 +1,199 @@
...
+                                    "Tomorrow 01/04/2018",
+                                    "Last Month 28/02/2018 (28th as no 31st February)",
+                                    "Next Month 30/04/2018 (30th as no 31st April)",
+                                    "Last Year 31/03/2017",
+                                    "Next Year 31/02/2019"
```
shahi645: 31/03/2019

app/questionnaire/rules.py
dwyeradam: UPDATE see below and return early on no comment.

app/questionnaire/rules.py
dwyeradam: Possibly not as it's no different from evaluate_rule being a comparison of answers/values.

dwyeradam:
@nealedj, @shahi645  Ready for re-review. I've added some comments as well for you to agree/disagree with.

nealedj (approved):

yoganandkunche (approved):
LGTM 👍

shahi645 (approved):

PR #1485, opened 2018-03-07
Eq 1868 validate sum of grouped answers against total

shahi645:
Added a feature for question level validation. This feature validates a group of numeric answers in a single block against a given total. 
Also added a feature to group the answers and gain an average to validate against a total.
New type of question --> `Calculated`
Validation can allow 3 conditions:
- Equal to 
- Less than
- Less than or Equal to

### How to review 
Test out Schemas:
- `test_sum_equal_validation_against_total.json`
- `test_sum_equal_or_less_validation_against_total.json`
- `test_sum_less_validation_against_total.json`
- `test_average_equal_validation_against_total.json`
- `test_average_less_validation_against_total.json`
- `test_average_equal_or_less_validation_against_total.json`

Will also need to run https://github.com/ONSdigital/eq-schema-validator/tree/eq-1868-validate-sum-against-total against SR here to test.


app/validation/error_messages.py
```diff
@@ -11,6 +11,9 @@
     'NUMBER_TOO_LARGE': 'Enter an answer less than or equal to %(max)s.',
     'NUMBER_TOO_SMALL_EXCLUSIVE': 'Enter an answer more than %(min)s.',
     'NUMBER_TOO_LARGE_EXCLUSIVE': 'Enter an answer less than %(max)s.',
+    'SUM_NOT_EQUAL_TO_TOTAL': 'Enter answers that add up to the %(label)s, %(total)s.',
+    'SUM_NOT_EQUAL_TO_OR_LESS_THAN_TOTAL': 'Enter answers that add up to or are less than the %(label)s, %(total)s.',
+    'SUM_NOT_LESS_THAN_TOTAL': 'Enter answers that are less than the %(label)s, %(total)s.',
```
MebinAbraham: Are using words like `"add up to"` in the error message correct for the average questions? It could easily be misinterpreted. Best to have something like `"Enter answers that average to"` etc. Because my numbers add up to ie 7 but don't average to 7.

MebinAbraham (changes requested):

app/validation/error_messages.py
shahi645: So average right now is just a feature, and wont be used until its wanted, at which point new error messages should be implemented also

MebinAbraham (commented):
FYI: Schemas are failing, I assume because you haven't pointed to your schema-validator branch.
They pass locally.

But I am getting a 500 Error on submission.

app/forms/questionnaire_form.py
```diff
@@ -31,24 +36,68 @@ def validate(self):
...
-        for question_id, period_from_id, period_to_id in self.date_ranges:
-            if period_from_id not in self.errors and period_to_id not in self.errors:
-                period_from = getattr(self, period_from_id)
-                period_to = getattr(self, period_to_id)
-                validator = DateRangeCheck()
-
-                # Check every field on each form has populated data
-                populated = all([f.data for f in period_from] + [f.data for f in period_to])
+        for question_id, question_calculated, target_label in self.calculated_question:
+            if not self.match_ids_to_errors(question_calculated['answers_to_group']) and valid_form:
```
nealedj: Just a stylistic thing. If you're checking multiple conditions in an expression then it's normally easier to read if the negated ones come last.

e.g. `if valid_form and not self.match_ids_to_errors(question_calculated['answers_to_group']):`

app/forms/questionnaire_form.py
```diff
@@ -1,20 +1,25 @@
...
 from flask_wtf import FlaskForm
 from werkzeug.datastructures import MultiDict
 
 from app.forms.fields import get_field
-from app.validation.validators import DateRangeCheck
+from app.validation.validators import DateRangeCheck, SumCheck
 
 logger = logging.getLogger(__name__)
 
+DEFAULT_VALUE = 'default'
```
nealedj: Slight improvement here. Use an object to compare instead of a string.

`DEFAULT_VALUE = new object()`

Also, maybe change the name of the `DEFAULT_VALUE` attribute to `SENTINEL` as this better describes what the value is there for.

app/forms/questionnaire_form.py
```diff
@@ -31,24 +36,68 @@ def validate(self):
         valid_form = True
```
nealedj: Given that `valid_form` always has to be `True` for this method to return `True` I'd suggest removing this variable and simply returning `False` in the places where this is set.

Having multiple return statements is common in python and if it makes the code clearer it's preferable.

E.g. 
```
for question_id, question_calculated, target_label in self.calculated_question:
    if not (
            self.match_ids_to_errors(question_calculated['answers_to_group']) and
            self._validate_calculated_question(question_id, question_calculated, target_label):
        return False
```

app/forms/questionnaire_form.py
```diff
@@ -31,24 +36,68 @@ def validate(self):
...
 
         return valid_fields and valid_form
 
+    def _validate_date_range_question(self, question_id, period_from_id, period_to_id):
+        period_from = getattr(self, period_from_id)
+        period_to = getattr(self, period_to_id)
+        validator = DateRangeCheck()
+
+        # Check every field on each form has populated data
+        populated = all([f.data for f in period_from] + [f.data for f in period_to])
```
nealedj: Consider using `all(field.data for field in itertools.chain(period_from, period_to))`.

app/forms/questionnaire_form.py
```diff
@@ -31,24 +36,68 @@ def validate(self):
...
+        answers_to_group = calculated_question.get('answers_to_group')
+        calculation_type = calculated_question.get('calculated_type')
+        target_total = self.answer_store.filter(answer_ids=[calculated_question.get('answer_id')]).values()[0]
+        condition = calculated_question.get('condition')
+        validator = SumCheck()
+
+        if calculation_type == 'sum':
+            calculation = sum
+        elif calculation_type == 'average':
+            calculation = mean
```
nealedj: Add an `else` block and raise an exception with a meaningful message like `"Invalid calculation_type: {}"`.

app/forms/questionnaire_form.py
```diff
@@ -31,24 +36,68 @@ def validate(self):
...
+            try:
+                validator(self, period_from, period_to)
+            except validators.ValidationError as e:
+                self.question_errors[question_id] = str(e)
+                return False
+
+        return True
+
+    def _validate_calculated_question(self, question_id, calculated_question, target_label):
+        answers_to_group = calculated_question.get('answers_to_group')
```
nealedj: If you use `.get()` on a dictionary then be prepared to handle missing values. If you know that the key will always be there then just index it directly.

app/forms/questionnaire_form.py
```diff
@@ -31,24 +36,68 @@ def validate(self):
...
+        calculation_type = calculated_question.get('calculated_type')
+        target_total = self.answer_store.filter(answer_ids=[calculated_question.get('answer_id')]).values()[0]
```
nealedj: Lots of assumptions running up to this line.

- assuming that 'answers_to_group', 'calculated_type' and 'answer_id' are all present on `calculated_question` and not `None`
- assuming that the answer pointed to by `calculated_question['answer_id']` exists and that it has exactly one value (or that only the first value is relevant)

At the very least, the code should assert that all of these things are true. That means if they aren't true in future (for any reason) it won't be as tricky to work out from staring at a single line of code that makes so many assumptions about its input.

app/forms/questionnaire_form.py
```diff
@@ -134,18 +186,23 @@ def generate_form(schema, block_json, data, answer_store):
 
     class DynamicForm(QuestionnaireForm):
```
nealedj, line 197: I know this was already here, but there's no need to create a new type when generating a form.

Add `date_ranges` and `calculated_question` to the `QuestionnaireForm` type.

```

class QuestionnaireForm(FlaskForm):
    date_ranges = []
    calculated_question = []

    def __init__(self, schema, block_json, answer_store, formdata=None, **kwargs):
        self.schema = schema
        self.block_json = block_json
        self.answer_store = answer_store
        self.question_errors = {}
        self.options_with_children = {}

        self.date_ranges = kwargs.pop('date_ranges', [])
        self.calculated_question = kwargs.pop('calculated_question', [])

        if formdata:
            super().__init__(formdata=formdata, **kwargs)
        else:
            super().__init__(**kwargs)

...
Then you'd need to pass the `date_ranges` and `calculated_question` arguments through to the `QuestionnaireForm` on creation and delete the `DynamicForm`.
```

app/validation/validators.py
```diff
@@ -235,3 +235,33 @@ def __call__(self, form, from_field, to_field):
 
         if from_date == to_date or from_date > to_date:
             raise validators.ValidationError(self.messages['INVALID_DATE_RANGE'])
+
+
+class SumCheck(object):
+    def __init__(self, messages=None, currency=None):
+        if not messages:
```
nealedj: Nitpick - since this is checked if the value was passed this check would be better checking `None` directly. `if messages is not None`.

app/validation/validators.py
```diff
@@ -235,3 +235,33 @@ def __call__(self, form, from_field, to_field):
...
+        elif condition == 'less-than-equal-to':
+            if total > target_total:
+                raise validators.ValidationError(self.messages['SUM_NOT_EQUAL_TO_OR_LESS_THAN_TOTAL'] %
+                                                 {'label': label,
+                                                  'total': self.format_min_max(target_total)})
+        elif condition == 'less-than':
+            if total >= target_total:
+                raise validators.ValidationError(self.messages['SUM_NOT_LESS_THAN_TOTAL'] %
+                                                 {'label': label,
+                                                  'total': self.format_min_max(target_total)})
```
nealedj: Missing `else` block. Raise an exception if the condition is invalid.

data/en/test_average_equal_or_less_validation_against_total.json
```diff
@@ -0,0 +1,93 @@
...
+                        "id": "total-question",
+                        "title": "Total",
+                        "type": "General",
+                        "answers": [{
+                            "id": "total-answer",
+                            "label": "Total",
+                            "mandatory": true,
+                            "type": "Number",
+                            "decimal_places": 2,
+                            "max_value": {
```
nealedj: I don't think there's any need to specify a max_value here? This isn't what we're testing in this schema.

data/en/test_average_equal_or_less_validation_against_total.json
```diff
@@ -0,0 +1,93 @@
...
+                    "title": "Calculated Total",
+                    "id": "breakdown-block",
+                    "questions": [{
+                        "id": "breakdown-question",
+                        "title": "Breakdown",
+                        "type": "Calculated",
+                        "calculated": {
+                            "calculated_type": "average",
+                            "answer_id": "total-answer",
+                            "answers_to_group": [
```
nealedj: I think using the word "group" should be avoided in here. 1) we aren't actually grouping the answers and 2) we've already given the word "group" a meaning within the application.

Perhaps change to "answers_to_aggregate"?

data/en/test_average_equal_or_less_validation_against_total.json
```diff
@@ -0,0 +1,93 @@
...
+                {
+                    "type": "Question",
+                    "title": "Calculated Total",
+                    "id": "breakdown-block",
+                    "questions": [{
+                        "id": "breakdown-question",
+                        "title": "Breakdown",
+                        "type": "Calculated",
+                        "calculated": {
+                            "calculated_type": "average",
```
nealedj: Could make this more specific and call it "aggregate_function". It's more specific but the terminology is less accessible to a general audience. Not sure that matters though. @dwyeradam you might have thoughts on this?

data/en/test_average_less_validation_against_total.json
```diff
@@ -0,0 +1,93 @@
...
+                                "breakdown-4"
+                            ],
+                            "condition": "less-than"
+                        },
+                        "answers": [{
+                                "id": "breakdown-1",
+                                "label": "Breakdown 1",
+                                "mandatory": false,
+                                "decimal_places": 2,
+                                "type": "Number"
```
nealedj: Consider using answer type="Currency" for at least one of these test schemas. Currency fields are the primary use case.

tests/app/validation/test_sum_check_validator.py
```diff
@@ -0,0 +1,69 @@
...
+        label = 'Total'
+
+        with self.assertRaises(ValidationError) as ite:
+            validator(mock_form, condition, calculation_total, target_total, label)
+
+        self.assertEqual(error_messages['SUM_NOT_LESS_THAN_TOTAL'] % {'label': label,
+                                                                      'total': self.format_min_max(target_total)},
+                         str(ite.exception))
+
+    @staticmethod
```
nealedj: This method already exists in `app.validation.validators`. Duplicating it here doesn't assert anything.

nealedj (changes requested):

MebinAbraham (dismissed):
Clicking on the error message should focus in on the field that is causing the mismatch?

nealedj:
@MebinAbraham that's an outstanding issue with the other feature that runs validation over multiple inputs (the date range). I've added this as a task to fix within that other issue.

MebinAbraham:
@nealedj Ah right ok, but it works with the `total_breakdown.json` schema?. Which is a similar concept to this one?

nealedj:
@MebinAbraham similar in how the validation is done. But different on what the validation is defined against.

In total_breakdown the validation runs on the total answer (and references the other answers). If there's a validation error then the message is attached to that total answer field.

This is the other way around. The validation runs against the N breakdown fields (and references the total answer on a previous page). If there's a validation error the message isn't attached to any of the breakdown inputs because it's logically an error with all of them. In that way it's similar to the date range validation - if the date is wrong then it's going to be a problem with both dates.

app/forms/questionnaire_form.py
shahi645: @nealedj So these things are mandatory and there is a test in schema-validator to raise errors if the answer does not exist. So these potential issues are stopped at schema level as apposed to living in SR code.

MebinAbraham:
@nealedj Ah right I see. Thanks for explaining.

app/forms/questionnaire_form.py
nealedj: Yep - ideally we should never run an invalid schema. But given that schema-validator is a separate app we should write our code in such a way that if a erroneous schema gets run it fails in a meaningful way. Too many assumptions on a single line makes this sort of thing a pain to debug.

app/forms/questionnaire_form.py
dwyeradam: To further this would it be worth returning early if valid_fields is False? Seems pointless validating the form if the fields are invalid.

`if not FlaskForm.validate(self):
     return False`

app/validation/validators.py
dwyeradam: Worse actually is the fact that both messages and currency are never set so we either don't need them or we need the ability to set them.

data/en/test_average_equal_or_less_validation_against_total.json
dwyeradam: In this instance I actually think readability is more important. As we have less and less involvement in creating the schema then understanding of the terms in them will be a greater issue.

dwyeradam (changes requested):
I'll do a fuller review upon addressing David's comments.

data/en/test_average_equal_or_less_validation_against_total.json
shahi645: how about `answers_to_calculate`?

nealedj: Nice - let's go with that.

shahi645: I will need to raise a PR in schema validator for this too as the change has been merged earlier

app/validation/validators.py
shahi645: Is it worth this being part of some tech debt as the whole file observes this behaviour

nealedj: Yeah I think that's fine. Would you mind adding a card for it?

app/forms/questionnaire_form.py
```diff
@@ -134,18 +202,22 @@ def generate_form(schema, block_json, data, answer_store):
...
         date_ranges = []
+        calculated_question = []
```
jonnyshaw89: These 2 lines blow my mind!
We are setting these for every form, just because.

app/forms/questionnaire_form.py
```diff
@@ -69,9 +134,12 @@ def map_errors(self):
     def answer_errors(self, input_id):
         return [error[1] for error in self.map_errors() if input_id == error[0]]
 
-    def get_data(self, answer_id):
+    def get_data(self, answer_id, default=SENTINEL):
```
jonnyshaw89: I don't understand what SENTINEL is??

jonnyshaw89 (changes requested):

app/forms/questionnaire_form.py
dwyeradam: What Jon meant to say was that since we have the block_json available to us at validate then do we need to create the date_ranges and calculated_question objects here?

shahi645: thanks adam

app/forms/questionnaire_form.py
shahi645: its a flag value, it helps when there is `none` data for a value, i.e stops the loop

nealedj: Yeah it's just so you're able to pass `(default=None)` as an argument. It'd work to have a `default='some_other_value'` but sentinel is a posher way to define it 😃

app/forms/questionnaire_form.py
nealedj: We've got a separate card open now to get rid of `DynamicForm` anyway. Better to allow `QuestionnaireForm` to take these sort of values in its cctor and set its own bounds at that point.

app/forms/questionnaire_form.py
jonnyshaw89: Why cant we just default it to none?
`def get_data(self, answer_id, default=None):`

app/validation/error_messages.py
davecharles: `'Enter answers that add up to the %(label)s, %(total)s.'` uses old style formatting, have you considered something like:

`'Enter answers that add up to the {label}, {total}.'`
then something like
```
self.messages['SUM_NOT_EQUAL_TO_OR_LESS_THAN_TOTAL'].format(**{'label': foo, 'total': bar})
```
more readable/neater?

charlotte-hirst (commented):
@shahi645 let me know when ready for review, only i tried this afternoon and get error when trying to look at schema in docker version but I know you are re-working bits so just give me a shout when ready

shahi645:
@charlotte-hirst Just fixing some of the dynamic form stuff should be ready this morning,  will let you know ASAP

app/validation/error_messages.py
nealedj: Agree with this re. formatting. We have a refactoring ticket for this whole area, let's do it in there rather than blocking this PR.

nealedj: @shahi645 I think we should either not support average checks or we need to provide descriptive validation messages for when those checks fail. Otherwise this is a bug report waiting to happen.

shahi645: @nealedj I can add the a average error messages, but the content needs to be approved and supplied

jonnyshaw89: @shahi645 just don't support them, We have the ability to easily add them, but they are not a requirement so get rid for now to make this PR simpler

app/forms/questionnaire_form.py
```diff
@@ -28,26 +34,85 @@ def validate(self):
...
+        except validators.ValidationError as e:
+            self.question_errors[question_id] = str(e)
+            return False
+
+        return True
+
+    @staticmethod
+    def _get_calculation_type(calculation_type):
+        if calculation_type == 'sum':
+            calculation = sum
```
jonnyshaw89: just `return sum`?

app/forms/questionnaire_form.py
```diff
@@ -28,26 +34,85 @@ def validate(self):
...
-                period_to = getattr(self, period_to_id)
-                validator = DateRangeCheck()
+            period_ids = [period_from_id, period_to_id]
+            if not (
+                    self.answers_all_valid(period_ids) and
+                    self._validate_date_range_question(question_id, period_from_id, period_to_id)):
+                return False
+
+        for question, target_answer in self.calculated_question:
+            target_total = self.answer_store.filter(answer_ids=[question['calculated'].get('answer_id')]).values()[0]
```
jonnyshaw89: you are assuming there is 1, and only 1?

jonnyshaw89: you should be able to get the answer directly
https://github.com/ONSdigital/eq-survey-runner/blob/master/app/data_model/answer_store.py#L63

app/forms/questionnaire_form.py
```diff
@@ -28,26 +34,85 @@ def validate(self):
...
+            if 'currency' in target_answer:
+                currency = target_answer['currency']
```
jonnyshaw89: currency = target_answer.get('currency', None)

app/forms/questionnaire_form.py
```diff
@@ -134,18 +202,23 @@ def generate_form(schema, block_json, data, answer_store):
...
         date_ranges = []
+        calculated_question = []
 
     for question in schema.get_questions_for_block(block_json):
         if question['type'] == 'DateRange':
             DynamicForm.date_ranges += [(question['id'], question['answers'][0]['id'], question['answers'][1]['id'])]
+            DynamicForm.date_ranges += [(question['id'], question['answers'][0]['id'], question['answers'][1]['id'])]
+        elif question['type'] == 'Calculated':
+            target_answer = schema.get_answer(question['calculated']['answer_id'])
+            DynamicForm.calculated_question += [(question, target_answer)]
```
dwyeradam: I thought we were getting rid of all this logic here and moving it to `validate(self)`

app/validation/validators.py
```diff
@@ -235,3 +232,39 @@ def __call__(self, form, from_field, to_field):
 
         if from_date == to_date or from_date > to_date:
             raise validators.ValidationError(self.messages['INVALID_DATE_RANGE'])
+
+
+# class SumCheck(object):
```
dwyeradam: You've pushed this commented out. This is no longer a working application.

dwyeradam (changes requested):

app/forms/questionnaire_form.py
shahi645: It breaks the dynamic form stuff, theres a card in our backlog to fix, so for now staying here

yoganandkunche:
@charlotte-hirst is it intentional to have a `,` in the validation error message Enter answers that add up to the Total, 12.12.

yoganandkunche:
@shahi645 sorry for being picky but I think the validation message will look better for `less than or equal` if we do not have `for` :  Enter answers that add up to or **are** less than the Total, 9,999,999,999.

yoganandkunche:
@shahi645 @charlotte-hirst I think for less than scenario it should read: Enter answers that **add up to**  less than the Total, £9,999,999,998.00.

app/validation/validators.py
```diff
@@ -235,3 +232,39 @@ def __call__(self, form, from_field, to_field):
...
+
+class SumCheck(object):
+    def __init__(self, messages=None, currency=None):
+        if not messages:
+            messages = error_messages
+        self.messages = messages
+        self.currency = currency
+
+    def __call__(self, form, conditions, total, target_total, label):
+        if len(conditions) > 1:
```
dwyeradam: As discussed this doesn't allow for easy addition of new conditions as your if statement is assuming only a combination or equals or less than.

app/validation/validators.py
```diff
@@ -235,3 +232,39 @@ def __call__(self, form, from_field, to_field):
 
         if from_date == to_date or from_date > to_date:
             raise validators.ValidationError(self.messages['INVALID_DATE_RANGE'])
+
+
+class SumCheck(object):
+    def __init__(self, messages=None, currency=None):
+        if not messages:
+            messages = error_messages
+        self.messages = messages
```
dwyeradam: we've still not got a way to use bespoke error messages as messages can not be set.

dwyeradam (changes requested):

jonnyshaw89 (changes requested):

app/forms/questionnaire_form.py
```diff
@@ -28,26 +34,85 @@ def validate(self):
         Validate this form as usual and check for any form-level date-range validation errors
         :return:
         """
-        valid_form = True
-        valid_fields = FlaskForm.validate(self)
+        if not FlaskForm.validate(self):
+            return False
```
jonnyshaw89: This isn't idea as it doesn't show all of the errors the 1st time around. Which can be annoying for the respondent.

jonnyshaw89 (changes requested):

app/forms/questionnaire_form.py
shahi645: I think in the previous version respondent only sees field errors first anyways, will double check

shahi645: on testing it returns both field and form validation. @nealedj what sort of behaviour do we want? i.e just field validations first or both at same time

jonnyshaw89: On PreProd it returns both straight away.
Run `test_dates.json`, enter a from date after the to date, don't answer any other questions and you will see field errors and range errors.

yoganandkunche: It has always been field validations first.

jonnyshaw89: Yes, but with this change, if the range is invalid and another question has a field error then the range validation doesn't show.

shahi645: ok have reverted back to valid fields notion, so exerts same behaviour as before

app/validation/validators.py
shahi645: @dwyeradam Have added it to https://trello.com/c/YDXDfWm7/1927-refactor-of-validatorspy-and-questionnaireformpy as AC. Or if we really need it for now i can put it in to questionnaire form to pass it on to validators

app/forms/questionnaire_form.py
jonnyshaw89: `date_ranges` and `calculated_question` don't need to be on the form, they can be calculated form the `block_json` when validating

app/forms/questionnaire_form.py
jonnyshaw89: We don't need to get rid of `DynamicForm` but we can get rid of the variables as we can work them out on validation

dwyeradam: Yep, having tried getting rid of `DynamicForm` here isn't as easy as I expected. However removing the `date_ranges` and `calculated_question` objects took 15 minutes.

app/validation/validators.py
dwyeradam: All our other validation allows for bespoke messages while this currently doesn't so it needs to.

app/forms/questionnaire_form.py
```diff
@@ -25,29 +31,82 @@ def __init__(self, schema, block_json, formdata=None, **kwargs):
...
+        for question, target_answer in self.calculated_question:
+            target_total = self.answer_store.filter(answer_ids=[question['calculated'].get('answer_id')]).values()[0]
+            currency = target_answer.get('currency')
+            if not (
+                    self.answers_all_valid(question['calculated']['answers_to_calculate']) and
+                    self._validate_calculated_question(question['id'], question['calculated'], target_answer['label'],
+                                                       target_total, currency)):
+                return False
+
+        return True and valid_fields
```
dwyeradam: `return True and valid_fields` is the same as `return valid_fields`. It might be worth putting back in the concept of `valid_form` and set those above rather than returning early. You'd then `return valid_fields and valid_form`

app/validation/validators.py
dwyeradam: All our other validation allows for bespoke messages while this currently doesn't so it needs to.

dwyeradam (changes requested):

app/forms/questionnaire_form.py
nealedj: Because the current behaviour is to raise an `AttributeError` if `get_data` is called without a default. Changing that would change its behaviour and require all callers to handle `None` values.

It's just a pattern to maintain a sensible interface and not allow errors to cascade to other areas in the system.

app/forms/questionnaire_form.py
```diff
@@ -25,29 +31,88 @@ def __init__(self, schema, block_json, formdata=None, **kwargs):
...
+        if 'validation' in question:
+            message = question['validation'].get('messages')
+
+        validator = SumCheck(messages=message, currency=target_answer.get('currency'))
+
+        calculation = self._get_calculation_type(calculated_question['calculated_type'])
+
+        formatted_values = self._get_formatted_calculation_values(calculated_question['answers_to_calculate'])
+
+        calculation_total = Decimal('0')
```
nealedj: This line is redundant isn't it? Could just be

```
calculation_total = self._get_calculation_total(calculation, formatted_values)
```

app/forms/questionnaire_form.py
```diff
@@ -25,29 +31,88 @@ def __init__(self, schema, block_json, formdata=None, **kwargs):
...
+                self.question_errors[question_id] = str(e)
+                return False
+
+        return True
+
+    def _validate_calculated_question(self, question, target_answer, target_total):
+        calculated_question = question['calculated']
+        message = None
+        if 'validation' in question:
+            message = question['validation'].get('messages')
```
nealedj: This returns a list so it'd sense to rename this variable to `messages`

app/forms/questionnaire_form.py
```diff
@@ -25,29 +31,88 @@ def __init__(self, schema, block_json, formdata=None, **kwargs):
...
+
+        return True
+
+    @staticmethod
+    def _get_calculation_type(calculation_type):
+        if calculation_type == 'sum':
+            return sum
+        else:
+            error = 'Invalid calculation_type: {}'.format(type)
+            raise validators.StopValidation(error)
```
nealedj: I don't think this is what `StopValidation` is intended for. I believe this raises a validation error and simply prevents any other validators from running.

In reality this is a critical break that should never happen - I think it'd be best to throw an `Exception`.

app/validation/validators.py
```diff
@@ -235,3 +232,40 @@ def __call__(self, form, from_field, to_field):
 
         if from_date == to_date or from_date > to_date:
             raise validators.ValidationError(self.messages['INVALID_DATE_RANGE'])
+
+
+class SumCheck(object):
```
nealedj, line 237: New suggestion:

```
class SumCheck(object):
    def __init__(self, messages=None, currency=None):
        if not messages:
            messages = error_messages
        self.messages = messages
        self.currency = currency

    def __call__(self, form, conditions, total, target_total):
        message_tmpl = 'TOTAL_SUM_NOT_{}'

        if len(conditions) > 1:
            try:
                conditions.remove('equals')
            except ValueError:
                raise Exception(
                    'There are multiple conditions, but equals is not one of them. '
                    'We only support <= and >=')

            condition = '{} or equals'.format(conditions[0])
        else:
            condition = conditions[0]

        is_valid, message_end = self._is_valid(condition, total, target_total)

        if not is_valid:
            raise validators.ValidationError(
                self.messages[message_tmpl.format(message_end)] %
                dict(total=format_playback_value(target_total, self.currency)))

    def _is_valid(self, condition, total, target_total)
        for condition in conditions:
            if condition == 'equals':
                return total == target_total, 'EQUALS'
            elif condition == 'less than':
                return total < target_total, 'LESS_THAN'
            elif condition == 'greater than or equals':
                return total > target_total, 'GREATER_THAN_OR_EQUALS'
            elif condition == 'less than or equals':
                return total > target_total, 'LESS_THAN_OR_EQUALS'
```

nealedj (changes requested):

app/validation/validators.py
```diff
@@ -235,3 +232,52 @@ def __call__(self, form, from_field, to_field):
...
+    def __init__(self, messages=None, currency=None):
+        if not messages:
+            messages = error_messages
+        self.messages = messages
+        self.currency = currency
+
+    def __call__(self, form, conditions, total, target_total):
+        message_tmpl = 'TOTAL_SUM_NOT_{}'
```
jonnyshaw89: I think we should get rid of this and be explicit about the error message we return in `_is_valid`
so instead of returning `EQUALS` we should just return `TOTAL_SUM_NOT_EQUALS` as there is no need to have to stitch this together

app/forms/questionnaire_form.py
jonnyshaw89: I don't like it, i think it makes the code hard to read.
I feel it compromises code readability.

app/validation/validators.py
```diff
@@ -130,7 +127,7 @@ def __init__(self, max_decimals=0, messages=None):
         self.messages = messages
 
     def __call__(self, form, field):
-        data = field.raw_data[0].replace(numbers.get_group_symbol(DEFAULT_LOCALE), '')
+        data = field.raw_data[0].replace(numbers.get_group_symbol(DEFAULT_LOCALE), '').replace(' ', '')
```
jonnyshaw89, line 130: Why are we removing spaces?

app/validation/validators.py
jonnyshaw89: Boy scout rule?

tests/app/forms/test_questionnaire_form.py
```diff
@@ -89,15 +102,207 @@ def test_date_range_to_precedes_from_raises_question_error(self):
...
+            }
+
+            form = generate_form(schema, block_json, data, store)
+
+            with self.assertRaises(wtf_validators.StopValidation):
+                with patch('app.questionnaire.questionnaire_schema.QuestionnaireSchema.get_questions_for_block',
+                           return_value=[question_json]):
+                    form.validate()
+
+    def test_bespoke_message(self):
```
jonnyshaw89: `test_bespoke_message_for_sum_validation`??

app/forms/questionnaire_form.py
dwyeradam: Also since `calculation` and `formatted_values` is only used in this method this would be easier to understand if it was created in `_get_calculation_total`

tests/app/validation/test_sum_check_validator.py
```diff
@@ -0,0 +1,133 @@
...
+
+        with self.assertRaises(ValidationError) as ite:
+            validator(mock_form, conditions, calculation_total, target_total)
+
+        self.assertEqual(error_messages['TOTAL_SUM_NOT_EQUALS'] % dict(total=format_playback_value(target_total,
+                                                                                                   currency='EUR')),
+                         str(ite.exception))
+
+    def test_bespoke_message_playback(self):
+        message = {'TOTAL_SUM_NOT_EQUALS': 'Test'}
```
jonnyshaw89: We should be testing the interpolation of the total as well

tests/app/validation/test_sum_check_validator.py
```diff
@@ -0,0 +1,133 @@
...
+        conditions = ['equals']
+        calculation_total = 10
+        target_total = 11.5
+
+        with self.assertRaises(ValidationError) as ite:
+            validator(mock_form, conditions, calculation_total, target_total)
+
+        self.assertEqual('Test', str(ite.exception))
+
+    def test_invalid_multiple_conditions(self):
```
jonnyshaw89: should probably also add `test_valid_multiple_conditions`

tests/integration/mci/test_clear_value.py
```diff
@@ -44,7 +44,7 @@ def test_clear_value(self):
 
         # We submit the form
         self.post(form_data)
-        self.assertInPage(error_messages['INVALID_DATE_RANGE'])
+        self.assertInPage("Enter a 'period to' date later than the 'period from' date.")
```
jonnyshaw89: Why?

jonnyshaw89 (changes requested):

app/validation/validators.py
shahi645: if the user has entered a space by accident

dwyeradam (changes requested):

app/forms/questionnaire_form.py
shahi645: I disagree

app/validation/validators.py
jonnyshaw89: Is that a requirement?
If it is shouldn't we do that for all inputs?

jonnyshaw89: If so it should probably be a documented feature in its own PR?

app/validation/validators.py
shahi645: Theres a card ^^^^^

app/forms/questionnaire_form.py
shahi645: needs it as breakdown fields can be not mandatory hence user wont enter anything

app/validation/validators.py
jonnyshaw89: I would accept that if it was to fix an existing issue, but this is something we are introducing? We are paying off tech debt at the moment so why introduce more?

app/validation/validators.py
shahi645: if its not there and user enters a space i believe they just get a 5xx, we wouldn't want that

app/forms/questionnaire_form.py
dwyeradam: @shahi645 I don't understand this either so could you explain it to us please.

tests/integration/mci/test_clear_value.py
shahi645: was not passing previously

jonnyshaw89: Tests currently work in master don't they?

shahi645: Yeah they could be reverted now, i think it was the if not fields validare thing but as thats been reverted

shahi645: done

app/validation/validators.py
nealedj: Yeah I think it's reasonable to strip spaces when parsing a numeric field. Even python does this:

```
>>> int(' 2     ')
2
```

nealedj (approved):

app/validation/validators.py
jonnyshaw89: If python does it why do we need to?

shahi645: So if you run `test_numbers` and enter a figure with 2 DP into the decimal field and put a trailing space you get the Decimal error, this just eliminates that for our user

app/forms/questionnaire_form.py
jonnyshaw89: @shahi645 I changed this to the following and it seems to work
```
    def get_data(self, answer_id, default=None):
        attr = getattr(self, answer_id)
        value = attr.raw_data[0] if attr.raw_data else ''
        if default is not None and not value:
            return default
        return value
```

shahi645: I'll remove the SENTINEL value 😖

tests/app/forms/test_questionnaire_form.py
```diff
@@ -97,7 +101,233 @@ def test_date_range_to_precedes_from_raises_question_error(self):
...
+                'breakdown-1': '3',
+                'breakdown-2': '5',
+            }
+
+            form = generate_form(schema, block_json, data, store)
+
+            with self.assertRaises(Exception):
+                with patch('app.questionnaire.questionnaire_schema.QuestionnaireSchema.get_questions_for_block',
+                           return_value=[question_json]):
+                    form.validate()
```
dwyeradam: This shouldn't just check the raising of an exception. It should also check it's of the . correct type with the correct message.

dwyeradam (changes requested):

tests/functional/spec/features/grouped_validation/sum/grouped_validation_sum.spec.js
```diff
@@ -0,0 +1,299 @@
...
+  });
+
+   describe('Equal to', function () {
+    describe('Given I start a grouped answer validation survey and enter 5 into the total', function() {
+
+      var TotalAnswerPage = require('../../../../pages/features/grouped_validation/sum/equal-to/total-block.page');
+      var BreakdownAnswerPage = require('../../../../pages/features/grouped_validation/sum/equal-to/breakdown-block.page');
+      var SummaryPage = require('../../../../pages/features/grouped_validation/sum/equal-to/summary.page');
+
+      before(function() {
```
dwyeradam: There is a lot of duplication here. The first `describe('Equal to', function () {` on line 4 should wrap all the others and that is where you declare the vars and before function. Then further down this is the same for less than etc.

dwyeradam (changes requested):

tests/functional/spec/features/grouped_validation/sum/grouped_validation_sum.spec.js
shahi645: done

jonnyshaw89: Not done.

jonnyshaw89: Still lots of duplication, the before functions should be shared

dwyeradam: @shahi645 You've not removed the duplication of `before` or `var` declaration. That can all be done within the initial `describe('Equal to', function () {` which then wraps the preceding describes. At the moment you still have a before and variable declaration for each describe.

shahi645: surely now i get a tick

jonnyshaw89 (approved):

dwyeradam (approved):

PR #1486, opened 2018-03-12
Move dynamodb session setup into per-request lifecycle

nealedj:
This removes the setup of the dynamodb session from the app setup as this is recommended to me per thread.

### How to review 
- Check that question playback still functions

### DIFF
--- a/app/data_model/submitted_responses.py
+++ b/app/data_model/submitted_responses.py
@@ -1,37 +1,52 @@
 from datetime import datetime
 from botocore.exceptions import BotoCoreError, ConnectionError as BotoConnectionError
+from flask import current_app
 from structlog import get_logger
+
+from app.globals import get_dynamodb
+
 
 logger = get_logger()
 
 
-class SubmittedResponses(object):
+def put_item(item):
+    """Insert an item to table"""
+    table = _get_table()
+    if not table:
+        return
 
-    def __init__(self, table):
-        self.table = table
+    if item and item['valid_until']:
+        item['valid_until'] = int(item['valid_until'].strftime('%s'))  # convert to epoch
 
-    def put_item(self, item):
-        """Insert an item to table"""
-        if item and item['valid_until']:
-            item['valid_until'] = int(item['valid_until'].strftime('%s'))  # convert to epoch
+    try:
+        response = table.put_item(Item=item)['ResponseMetadata']['HTTPStatusCode']
+        return response == 200
 
-        try:
-            response = self.table.put_item(Item=item)['ResponseMetadata']['HTTPStatusCode']
-            return response == 200
+    except (BotoCoreError, BotoConnectionError, ConnectionError):
+        logger.warn('Could not load data into DynamoDB')
+        return False
 
-        except (BotoCoreError, BotoConnectionError, ConnectionError):
-            logger.warn('Could not load data into DynamoDB')
-            return False
 
-    def get_item(self, tx_id):
-        """ Get an item given its key """
-        try:
-            response = self.table.get_item(Key={'tx_id': tx_id})
-            item = response.get('Item', None)
-            if item:
-                item['valid_until'] = datetime.utcfromtimestamp(item['valid_until'])
-            return item
+def get_item(tx_id):
+    """ Get an item given its key """
+    table = _get_table()
+    if not table:
+        return
 
-        except (BotoCoreError, BotoConnectionError, ConnectionError):
-            logger.warn('Could not get data from DynamoDB')
-            return None
+    try:
+        response = table.get_item(Key={'tx_id': tx_id})
+        item = response.get('Item', None)
+        if item:
+            item['valid_until'] = datetime.utcfromtimestamp(item['valid_until'])
+        return item
+
+    except (BotoCoreError, BotoConnectionError, ConnectionError):
+        logger.warn('Could not get data from DynamoDB')
+        return None
+
+
+def _get_table():
+    dynamodb = get_dynamodb()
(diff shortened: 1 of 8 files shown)


app/globals.py
```diff
@@ -62,3 +65,23 @@ def get_answer_store(user):
...
+
+    db = getattr(g, '_dynamodb', None)
+    if db is None:
+        # Number of additional connection attempts
+        config = Config(retries={'max_attempts': 1})
+
+        dynamodb = boto3.resource(
+            'dynamodb', endpoint_url=current_app.config['EQ_DYNAMODB_ENDPOINT'], config=config)
+
+        db = g._dynamodb = dynamodb
```
jonnyshaw89: Would `return g._dynamodb = dynamodb` work?

tests/integration/views/test_view_submission.py
```diff
@@ -47,18 +44,20 @@ def test_view_submission(self):
...
 
         # Submit answers
-        self.table.put_item = Mock(return_value={'ResponseMetadata': {'HTTPStatusCode': 200}})
-        self.post(action=None)
+        with mock.patch('app.data_model.submitted_responses.put_item',
+                        return_value={'ResponseMetadata': {'HTTPStatusCode': 200}}) as m_put_item:
+            self.post(action=None)
 
-        item = self.table.put_item.call_args[1]
+            item = m_put_item.call_args[0][0]
```
jonnyshaw89: anything wrong with just `m_put_item` i see the intention behind it but it brakes consistency

jonnyshaw89 (changes requested):

app/globals.py
nealedj: Yes, but that line only gets reached if the dynamo resource object wasn't already on the context object. So we'd have two returns:

```
db = getattr(g, '_dynamodb', None)
if db is None:
   dynamodb = stuff()
   return g._dynamodb = dynamodb
return db
```

tests/integration/views/test_view_submission.py
nealedj: If we use `m_put_item` outside of the patch context manager then we'll need to indent the following patch inside of it.

I.e.
```
with mock.patch('put_item') as m_put_item:
    ...stuff
    with mock.patch('get_item', return_value=item):
        ...stuff
```

app/globals.py
jonnyshaw89: Ah yeh I suppose its tidier. Just seems odd to set a variable that we don't need.

tests/integration/views/test_view_submission.py
jonnyshaw89: i'm not sure I follow.

jonnyshaw89 (approved):

app/data_model/submitted_responses.py
```diff
@@ -1,37 +1,44 @@
...
-                item['valid_until'] = datetime.utcfromtimestamp(item['valid_until'])
-            return item
+def _get_table():
+    dynamodb = get_dynamodb()
+    table_name = current_app.config['EQ_SUBMITTED_RESPONSES_TABLE_NAME']
 
-        except (BotoCoreError, BotoConnectionError, ConnectionError):
-            logger.warn('Could not get data from DynamoDB')
-            return None
+    return dynamodb.Table(table_name)
```
dwyeradam: Wouldn't this fail if get_dynamodb() returns None? I realise that currently everything checks that dynamodb is enabled before calling put_item or get_item but should we have some checks within submitted_responses.py to ensure this? 
See: https://github.com/ONSdigital/eq-survey-runner/pull/1486/files#diff-a160493e40efb08443761eff2a589ea2R72

dwyeradam (changes requested):

app/data_model/submitted_responses.py
nealedj: Good catch!

dwyeradam (changes requested):
Happy with the code. These changes have dropped the code coverage though so the tests may need to be reworked to hit more.

tests/app/data_model/test_submitted_responses.py
```diff
@@ -20,36 +17,58 @@ def setUp(self):
...
+    @mock.patch('app.data_model.submitted_responses._get_table', return_value=None)
+    def test_no_table(self, _get_table):
+        self.assertFalse(submitted_responses.put_item(self.item))
+        self.assertFalse(submitted_responses.get_item('test_tx_id'))
+
+
+class SubmittedResponsesGetTableTest(unittest.TestCase):
+    @mock.patch('app.data_model.submitted_responses.get_dynamodb')
+    @mock.patch('app.data_model.submitted_responses.current_app', **{'config.new': {'EQ_SUBMITTED_RESPONSES_TABLE_NAME': 'foo'}})
+    def test_get_table(self, get_dynamodb, current_app): # pylint: disable=unused-argument
```
jonnyshaw89, line 73: what is this actually doing?

nealedj: Testing that an object is returned by `_get_table` if `get_dynamodb()` has returned the dynamo resource and there is a `table_name` defined. Which is basically all this method does.

Also just makes sure the function is run as part of the test suite.

dwyeradam (approved):

dwyeradam:
@nealedj can be squashed and we'll merge it after.

jonnyshaw89:
@shahi645 we don't use the "update" button as it causes messy commit history

shahi645:
ahh okay, my bad

PR #1487, opened 2018-03-12
Split QBS 2_0001 Schema into two blocks

dwyeradam:
Schema: 2_0001.json
Total employees question first
No single employee split answer can not be greater than total

### DIFF
--- a/data/en/2_0001.json
+++ b/data/en/2_0001.json
@@ -60,44 +60,16 @@
                     }]
                 },
                 {
-                    "id": "number-of-employees",
+                    "id": "number-of-employees-total-block",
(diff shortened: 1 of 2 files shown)


jonnyshaw89:
If I answer the 1st question as 100 and then the 2nd page questions with appropriate answers, if i go back and edit the 1st question to 1 and jump back to the summary page.
I am able to submit the answers in an invalid state.
I think if we modify an answer that is depended on by validation then we should invalidate the answers for the dependent block.

jonnyshaw89 (approved):

MIslam92 (approved):

Issue #1499, opened 2018-03-13
Missing closing tags in 1_0005 schema


nealedj:
Introduction of fortnightly pay section is missing a closing `</ul></p>`.

PR #1513, opened 2018-03-19
Added Zero routing to MWSS

jonnyshaw89:
Added Zero routing around employee breakdown questions

Fixes https://github.com/ONSdigital/eq-survey-runner/issues/1499


shahi645:
Yarn Format

data/en/test_sum_equal_or_less_validation_against_total.json
```diff
@@ -90,4 +90,4 @@
             ]
         }]
     }]
-}
+}
```
shahi645, line 93: why are these committed?

jonnyshaw89: yarn formatted them. can only assume they slipped through before

data/en/1_0005.json
```diff
@@ -1850,6 +2017,54 @@
                         "title": "Five Weekly Pay",
                         "type": "Question"
                     },
+                    {
+                        "type": "Question",
+                        "title": "Five Weekly Pay",
+                        "id": "confirm-five-weekly-pay-gross-pay",
+                        "skip_conditions": [{
+                            "when": [{
+                                "id": "four-five-weekly-gross-pay-answer",
```
shahi645: delete four

shahi645 (changes requested):

MIslam92 (approved):

data/en/1_0005.json
```diff
@@ -1850,6 +2017,54 @@
                         "title": "Five Weekly Pay",
                         "type": "Question"
                     },
+                    {
+                        "type": "Question",
+                        "title": "Five Weekly Pay",
+                        "id": "confirm-five-weekly-pay-gross-pay",
+                        "skip_conditions": [{
+                            "when": [{
+                                "id": "five-weekly-gross-pay-answer",
```
shahi645: My bad, you also need to add a `pay` here:
`five-weekly-pay-gross-pay-answer`

shahi645 (changes requested):
If you want to rebase and squash same time, its pretty much there

data/en/1_0005.json
```diff
@@ -385,6 +380,54 @@
...
+                                        "value": "No",
+                                        "id": "confirm-weekly-pay-gross-pay-answer",
+                                        "condition": "equals"
+                                    }],
+                                    "block": "weekly-pay-gross-pay"
+                                }
+                            },
+                            {
+                                "goto": {
+                                    "group": "fortnightly-pay"
```
dwyeradam: @charlotte-hirst. Although the slide shows route to end is there an argument that we should actually route to the significant change question?

dwyeradam (changes requested):

data/en/1_0005.json
shahi645: I was thinking that earlier

dwyeradam:
@charlotte-hirst, @will-madebymoon. Nothing to do with this PR but we could easily fix it here. 
On the **Are these all of the pay patterns used?** question the No option states:
**No, I would like to add another pay pattern**
This is fine if you've not selected a pay pattern but doesn't work if you've selected too many. Should it actually be **No, I would like to change the pay patterns**?

data/en/1_0005.json
charlotte-hirst: it should go to significant changes to explain why its zero

charlotte-hirst:
Yes @dwyeradam  i agree it should be  No, I would like to change the pay patterns?

dwyeradam (approved):
Without me complaining we would have had incorrect routing and q-codes as well.

dwyeradam:
@shahi645 can you just check you're happy.

shahi645 (approved):
Just needs that special `Yarn format` but everything else is fine

PR #1514, opened 2018-03-19
Apply breakdown sum validation to QBS

dwyeradam:
As a respondent
I want my answers validated before I submit
So that ONS does not contact me after submission

### How to review 
Enter a value for total number of employees.
The sum of the breakdown answers must then total less than or equal to that value. Otherwise a correct error message should be displayed.

### DIFF
--- a/data/en/2_0001.json
+++ b/data/en/2_0001.json
@@ -154,7 +154,23 @@
                 },
                 {
                     "id": "number-of-employees-split-block",
+                    "title": "Quarterly Business Survey",
+                    "type": "Question",
                     "questions": [{
+                        "calculated": {
+                            "calculated_type": "sum",
+                            "answer_id": "number-of-employees-total",
+                            "answers_to_calculate": [
+                                "number-of-employees-male-more-30-hours",
+                                "number-of-employees-male-less-30-hours",
+                                "number-of-employees-female-more-30-hours",
+                                "number-of-employees-female-less-30-hours"
+                            ],
+                            "conditions": [
+                                "less than",
+                                "equals"
+                            ]
+                        },
                         "answers": [{
                                 "id": "number-of-employees-male-more-30-hours",
                                 "label": "Number of male employees working more than 30 hours per week",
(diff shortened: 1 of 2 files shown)


dwyeradam:
@shahi645 is there a reason the label isn't piped into the error message anymore? I have used a bespoke message to the same effect.

jonnyshaw89 (approved):

shahi645:
@dwyeradam We went through the content me Charlotte and Will, we agreed to take it out

data/en/2_0001.json
```diff
@@ -154,7 +154,28 @@
...
+                            ],
+                            "conditions": [
+                                "less than",
+                                "equals"
+                            ]
+                        },
+                        "validation": {
+                            "messages": {
+                                "TOTAL_SUM_NOT_LESS_THAN_OR_EQUALS": "Enter answers that add up to or are less than the Total number of employees, %(total)s"
+                            }
```
shahi645: I think we just want to use the standard error messages @charlotte-hirst can you confirm?

data/en/2_0001.json
```diff
@@ -154,7 +154,28 @@
...
+                            "answer_id": "number-of-employees-total",
+                            "answers_to_calculate": [
+                                "number-of-employees-male-more-30-hours",
+                                "number-of-employees-male-less-30-hours",
+                                "number-of-employees-female-more-30-hours",
+                                "number-of-employees-female-less-30-hours"
+                            ],
+                            "conditions": [
+                                "less than",
+                                "equals"
```
shahi645, line 171: Also are we going with less than or equals?

data/en/2_0001.json
charlotte-hirst: yes standard message without piping of answer label please

data/en/2_0001.json
charlotte-hirst: its less than or equal for employment

data/en/2_0001.json
jonnyshaw89: @shahi645 @charlotte-hirst we need to make sure that offline conversations are updated in the specs

dwyeradam:
@shahi645 Bespoke error message removed so ready for review please.

shahi645 (approved):
LGTM

PR #1517, opened 2018-03-21
Pipe difference between two dates

MebinAbraham:
Added new Jinja filters to calculate and pipe the difference between two dates.
Now also supports `MonthYearDate` ranges.
[Trello Card](https://trello.com/c/iUoHqZDx/1931-pipe-difference-between-two-dates)

### How to review
1. For the new schemas, ensure the playback value for the difference in years is valid.

Schemas:
```
test_difference_in_years.json
test_difference_in_years_month_year.json
test_difference_in_years_month_year_range.jso
test_difference_in_years_range.json
```

Ensure that all tests pass and validate all the schemas are correct and the structure is valid.

### DIFF
--- a/app/jinja_filters.py
+++ b/app/jinja_filters.py
@@ -5,13 +5,14 @@ import string
 
 from datetime import datetime
 
+from dateutil import relativedelta
 import flask
 
 from jinja2 import Markup, escape, evalcontextfilter
 from babel import units, numbers
 
 from app.settings import DEFAULT_LOCALE
-
+from app.questionnaire.rules import convert_to_datetime
 blueprint = flask.Blueprint('filters', __name__)
 
 
@@ -66,8 +67,12 @@ def format_multilined_string(context, value):
 
 
 @blueprint.app_template_filter()
-def format_date(value, date_format='%-d %B %Y'):
-    return "<span class='date'>{date}</span>".format(date=datetime.strptime(value, '%Y-%m-%d').strftime(date_format))
+def format_date(value):
+    date_format = '%B %Y'
+    if value and re.match(r'\d{4}-\d{2}-\d{2}', value):
+        date_format = '%-d %B %Y'
+
+    return "<span class='date'>{date}</span>".format(date=convert_to_datetime(value).strftime(date_format))
 
 
 @blueprint.app_template_filter()
@@ -98,6 +103,19 @@ def format_conditional_date(date1=None, date2=None):
         raise Exception('No valid dates passed to format_conditional_dates filter')
 
     return format_date(date)
+
+
+@blueprint.app_template_filter()
+def calculate_years_difference(from_str, to_str):
+    if from_str is None or to_str is None:
+        raise Exception('Valid date(s) not passed to calculate_years_difference filter')
+
+    to_date = datetime.now() if to_str == 'now' else convert_to_datetime(to_str)
+    from_date = convert_to_datetime(from_str)
+    difference = relativedelta.relativedelta(to_date, from_date)
+    year_string = '{} year' if difference.years == 1 else '{} years'
+
+    return year_string.format(difference.years)
 
 
 def format_date_range(start_date, end_date=None):
--- a/app/questionnaire/rules.py
+++ b/app/questionnaire/rules.py
@@ -1,4 +1,5 @@
 import logging
+import re
 from datetime import datetime
 from dateutil.relativedelta import relativedelta
 MAX_REPEATS = 25
@@ -79,9 +80,9 @@ def get_date_match_value(date_comparison, answer_store, group_instance, metadata
 
 
 def convert_to_datetime(value):
-    date_format = '%Y-%m-%d'
-    if value and len(value) == 7:
-        date_format = '%Y-%m'
+    date_format = '%Y-%m'
+    if value and re.match(r'\d{4}-\d{2}-\d{2}', value):
+        date_format = '%Y-%m-%d'
 
     return datetime.strptime(value, date_format) if value else None
 
--- a/data/en/test_difference_in_years_month_year.json
+++ b/data/en/test_difference_in_years_month_year.json
@@ -1,0 +1,85 @@
+{
+    "mime_type": "application/json/ons/eq",
+    "schema_version": "0.0.1",
+    "data_version": "0.0.2",
+    "survey_id": "023",
+    "title": "Difference between two dates",
+    "description": "A test schema for calculate age from date",
+    "theme": "default",
+    "legal_basis": "StatisticsOfTradeAct",
+    "sections": [{
+        "id": "default-section",
+        "groups": [{
+            "id": "dates",
+            "title": "Date Difference",
+            "blocks": [{
+                    "type": "Question",
+                    "id": "age-block",
+                    "title": "Calculate Difference",
+                    "questions": [{
+                        "id": "date-pipe-question",
+                        "title": "When did you last go on holiday?",
+                        "type": "General",
+                        "answers": [{
+                            "alias": "date_dob",
+                            "id": "date-dob",
+                            "label": "For example March 1990",
+                            "mandatory": true,
+                            "type": "MonthYearDate"
+                        }]
+                    }]
+                },
+                {
+                    "type": "ConfirmationQuestion",
+                    "id": "age-test",
+                    "title": "Confirm Duration",
+                    "questions": [{
+                        "id": "confirm-dob-question",
+                        "title": "It has been {{ calculate_years_difference (answers.date_dob, 'now') }} since you last went on holiday. Is this correct?",
+                        "type": "General",
+                        "answers": [{
+                            "id": "date-test-answer",
+                            "label": "Piped dates",
+                            "mandatory": true,
+                            "type": "Radio",
+                            "options": [{
+                                    "label": "Yes",
+                                    "value": "Yes"
+                                },
+                                {
+                                    "label": "No",
+                                    "value": "No"
+                                }
+                            ]
+                        }]
+                    }],
+                    "routing_rules": [{
+                            "goto": {
+                                "block": "age-block",
+                                "when": [{
+                                    "id": "date-test-answer",
+                                    "condition": "equals",
+                                    "value": "No"
+                                }]
+                            }
+                        },
+                        {
+                            "goto": {
+                                "block": "summary",
+                                "when": [{
+                                    "id": "date-test-answer",
+                                    "condition": "equals",
+                                    "value": "Yes"
+                                }]
+                            }
+                        }
+                    ]
+                },
+                {
+                    "type": "Summary",
+                    "id": "summary"
+                }
+            ]
+        }]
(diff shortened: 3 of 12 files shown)


scripts/test_schemas.sh
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
-docker pull onsdigital/eq-schema-validator
-validator="$(docker run -d -p 5001:5000 onsdigital/eq-schema-validator)"
+docker pull onsdigital/eq-schema-validator:support-for-month-year-date-range
+validator="$(docker run -d -p 5001:5000 onsdigital/eq-schema-validator:support-for-month-year-date-range)"
```
MebinAbraham: This will be reverted once validator PR is merged.

app/jinja_filters.py
```diff
@@ -67,7 +68,9 @@ def format_multilined_string(context, value):
 
 @blueprint.app_template_filter()
 def format_date(value, date_format='%-d %B %Y'):
-    return "<span class='date'>{date}</span>".format(date=datetime.strptime(value, '%Y-%m-%d').strftime(date_format))
+    if value and len(value) == 7:
```
jonnyshaw89: I would probably use a regex rather then count the chars

app/jinja_filters.py
```diff
@@ -100,6 +103,28 @@ def format_conditional_date(date1=None, date2=None):
     return format_date(date)
 
 
+@blueprint.app_template_filter()
+def calculate_years_difference(date_1, date_2):
```
jonnyshaw89: I would make these `(from, to)` rather than 1 and 2

app/jinja_filters.py
```diff
@@ -100,6 +103,28 @@ def format_conditional_date(date1=None, date2=None):
...
+
+    from_date = convert_to_datetime(date_1)
+
+    if date_2 == 'now':
+
```
jonnyshaw89: why all the whitespace?

app/jinja_filters.py
```diff
@@ -100,6 +103,28 @@ def format_conditional_date(date1=None, date2=None):
...
+        to_date = datetime.now()
+
+    else:
+
+        to_date = convert_to_datetime(date_2)
+
+    difference = relativedelta.relativedelta(to_date, from_date)
+    year_string = 'year' if difference.years == 1 else 'years'
```
jonnyshaw89: year_string = '{} year' if difference.years == 1 else '{} years'
Would make the following more efficient

app/jinja_filters.py
```diff
@@ -100,6 +103,28 @@ def format_conditional_date(date1=None, date2=None):
...
+
+    if from_date is None or to_date is None:
+        raise Exception('Valid date(s) not passed to calculate_years_difference filter')
+
+    return '{} {}'.format(difference.years, year_string)
```
jonnyshaw89: `return year_string.format(difference.years)`
as you are creating 1 less string

app/jinja_filters.py
```diff
@@ -100,6 +103,28 @@ def format_conditional_date(date1=None, date2=None):
...
+        to_date = datetime.now()
+
+    else:
+
+        to_date = convert_to_datetime(date_2)
+
+    difference = relativedelta.relativedelta(to_date, from_date)
+    year_string = 'year' if difference.years == 1 else 'years'
+
+    if from_date is None or to_date is None:
```
jonnyshaw89: Best to make this the 1st thing you do in the method

jonnyshaw89 (changes requested):

tests/app/validation/test_date_range_validator.py
```diff
@@ -16,17 +16,17 @@ def test_date_range_matching_dates(self):
 
         validator = DateRangeCheck()
 
-        period_from = Mock()
+        period_from = MagicMock()
```
shahi645, line 19: @MebinAbraham just out of interest why are we using `MagicMock` instead of `Mock`

app/jinja_filters.py
MebinAbraham: Reused logic from the `convert_to_datetime` method. Shall modify that method to match?

tests/app/validation/test_date_range_validator.py
MebinAbraham: @shahi645`Mock` isn't iterable and throws an error due to this: 
```
from_day = int(from_field.day.data) if 'day' in from_field.data else 1
to_day = int(to_field.day.data) if 'day' in to_field.data else 1
```

data/en/test_difference_in_years_month_year.json
```diff
@@ -0,0 +1,87 @@
...
+    "survey_id": "023",
+    "title": "Difference between two dates",
+    "description": "A test schema for calculate age from date",
+    "theme": "default",
+    "legal_basis": "StatisticsOfTradeAct",
+    "sections": [{
+        "id": "default-section",
+        "groups": [{
+            "id": "dates",
+            "title": "",
```
MIslam92: Should not have empty titles in schemas (Applies to all other instances throughout the schemas)

data/en/test_difference_in_years_month_year.json
```diff
@@ -0,0 +1,87 @@
...
+            "blocks": [{
+                    "type": "Question",
+                    "id": "age-block",
+                    "description": "",
+                    "title": "Calculate Difference",
+                    "questions": [{
+                        "description": "",
```
MIslam92: You can remove these empty description fields if your not using them? (Applies to all other instances throughout the schemas)

MIslam92 (changes requested):

app/jinja_filters.py
```diff
@@ -66,8 +67,11 @@ def format_multilined_string(context, value):
 
 
 @blueprint.app_template_filter()
-def format_date(value, date_format='%-d %B %Y'):
-    return "<span class='date'>{date}</span>".format(date=datetime.strptime(value, '%Y-%m-%d').strftime(date_format))
+def format_date(value, date_format='%B %Y'):
+    if value and re.match(r'\d{4}-\d{2}-\d{2}', value):
+        date_format = '%-d %B %Y'
```
jonnyshaw89: You are reassigning a parameter. bad practice

app/jinja_filters.py
```diff
@@ -66,8 +67,11 @@ def format_multilined_string(context, value):
 
 
 @blueprint.app_template_filter()
-def format_date(value, date_format='%-d %B %Y'):
-    return "<span class='date'>{date}</span>".format(date=datetime.strptime(value, '%Y-%m-%d').strftime(date_format))
+def format_date(value, date_format='%B %Y'):
```
jonnyshaw89: is this really a parameter? we wont honour it if you pass a full date in.

app/jinja_filters.py
```diff
@@ -100,6 +104,23 @@ def format_conditional_date(date1=None, date2=None):
     return format_date(date)
 
 
+@blueprint.app_template_filter()
+def calculate_years_difference(date_from, date_to):
+    if date_from is None or date_to is None:
+        raise Exception('Valid date(s) not passed to calculate_years_difference filter')
+
+    if date_to == 'now':
+        to_date = datetime.now()
```
jonnyshaw89: `date_to` and `to_date` as too similar its confusing

app/questionnaire/rules.py
```diff
@@ -78,10 +79,9 @@ def get_date_match_value(date_comparison, answer_store, group_instance, metadata
     return match_value
 
 
-def convert_to_datetime(value):
-    date_format = '%Y-%m-%d'
-    if value and len(value) == 7:
-        date_format = '%Y-%m'
+def convert_to_datetime(value, date_format='%Y-%m'):
+    if value and re.match(r'\d{4}-\d{2}-\d{2}', value):
+        date_format = '%Y-%m-%d'
```
jonnyshaw89: same comment above re re-assigning parameter

app/jinja_filters.py
```diff
@@ -66,8 +67,11 @@ def format_multilined_string(context, value):
 
 
 @blueprint.app_template_filter()
-def format_date(value, date_format='%-d %B %Y'):
-    return "<span class='date'>{date}</span>".format(date=datetime.strptime(value, '%Y-%m-%d').strftime(date_format))
+def format_date(value, date_format='%B %Y'):
+    if value and re.match(r'\d{4}-\d{2}-\d{2}', value):
```
jonnyshaw89: Do we need any of this logic here if it is done in `convert_to_datetime`?

jonnyshaw89 (changes requested):

app/jinja_filters.py
MebinAbraham: Here it is used to just define the `date_format`, the other ones is used to convert to a datetime object aswell. The `date_format` differs between the two.

jonnyshaw89: Ah yes, same logic but different formats

app/jinja_filters.py
```diff
@@ -100,6 +105,20 @@ def format_conditional_date(date1=None, date2=None):
     return format_date(date)
 
 
+@blueprint.app_template_filter()
+def calculate_years_difference(date_from, date_to):
```
jonnyshaw89: if you make these parametrs `(from, to)`
then you don't need to suffiix `_converted` which looks messy

tests/app/test_jinja_filters.py
```diff
@@ -158,6 +170,36 @@ def test_format_conditional_date(self):
...
+                     ('2016-02-29', '2017-02-28', '1 year'),
+                     ('2016-02-29', '2020-02-28', '3 years'),
+                     (ten_years_ago, 'now', '10 years')]
+
+        for dates in date_list:
+            start_date = dates[0]
+            end_date = dates[1]
+
+            # When
+            calculated_values = calculate_years_difference(start_date, end_date)
```
jonnyshaw89: minor, but this should be `calculated_value`. plural implies that it is a list

jonnyshaw89 (changes requested):
Looking good, Almost done.

tests/app/test_jinja_filters.py
MebinAbraham: Was a list at one point. Whoops.

tests/app/test_jinja_filters.py
```diff
@@ -116,6 +119,15 @@ def test_format_date(self):
 
         self.assertEqual(format_value, "<span class='date'>1 January 2017</span>")
 
+    def test_format_date_month_year(self):
+        # Given
+        date = '2017-01'
+
+        # When
+        format_value = format_date(date)
+
```
MIslam92, line 128: Forgot to add # Then here

MIslam92 (changes requested):

jonnyshaw89 (approved):

MIslam92 (approved):

MIslam92:
@MebinAbraham "Squishy Squashy"

PR #1520, opened 2018-03-22
Allow mulitple OR calculations to be done on Question validation

dwyeradam:
Allow for multiple calculation validation to be performed on a single question
Include value as an option to use instead of answer_id
Apply change to QBS (2_0001.json)

### DIFF
--- a/tests/app/forms/test_questionnaire_form.py
+++ b/tests/app/forms/test_questionnaire_form.py
@@ -124,8 +124,8 @@ class TestQuestionnaireForm(AppContextTestCase):
             question_json = {
                 'id': 'breakdown-question',
                 'type': 'Calculated',
-                'calculated': {
-                    'calculated_type': 'subtraction',
+                'calculations': [{
+                    'calculation_type': 'subtraction',
                     'answer_id': 'total-answer',
                     'answers_to_calculate': [
                         'breakdown-1',
@@ -134,7 +134,7 @@ class TestQuestionnaireForm(AppContextTestCase):
                     'conditions': [
                         'equals'
                     ]
-                },
+                }],
                 'answers': [{
                     'id': 'breakdown-1',
                     'label': 'Breakdown 1',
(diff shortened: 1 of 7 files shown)


scripts/test_schemas.sh
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
-docker pull onsdigital/eq-schema-validator
-validator="$(docker run -d -p 5001:5000 onsdigital/eq-schema-validator)"
+docker pull onsdigital/eq-schema-validator:multi-rule-calculation
```
jonnyshaw89: This needs revert.

tests/app/forms/test_questionnaire_form.py
```diff
@@ -124,8 +124,8 @@ def test_invalid_calculation_type(self):
             question_json = {
                 'id': 'breakdown-question',
                 'type': 'Calculated',
-                'calculated': {
-                    'calculated_type': 'subtraction',
+                'calculations': [{
+                    'calculation_type': 'subtraction',
```
jonnyshaw89, line 128: Whats this?

jonnyshaw89 (changes requested):

tests/app/forms/test_questionnaire_form.py
dwyeradam: So it appears @shahi645 added tests for scenarios he doesn't allow at a schema level and we let it through. @shahi645 do we need to add this at schema level or remove this test until we have a full set of tests for it?

dwyeradam: Ignore this. The test is actually ensuring it can handle an invalid calculation type.

shahi645: it is at schema level but Its good to test it here too, it was a PR change requested

jonnyshaw89 (approved):

shahi645 (approved):
LGTM 
Only thing is the order of multi rules need to be explicit, maybe something we want to look at in the future

PR #1521, opened 2018-03-23
Date range validation

shahi645:
- Added date range validation for a period between to dates, allows for a minimum limit or/and a maximum limit to be set 
- Added period validation given a referenced date for single dates
- Runtime validation for invalid schemas with date ranges

### How to review 
- open `test_date_range_period_validation.json` and test out the question level validation
- also open `test_single_date_period_validation.json` and test out single date period validation
- Schema-validator PR: https://github.com/ONSdigital/eq-schema-validator/pull/44

### DIFF
--- a/app/forms/date_form.py
+++ b/app/forms/date_form.py
@@ -1,20 +1,27 @@
 import logging
 import calendar
 
+from datetime import datetime
+
+from dateutil.relativedelta import relativedelta
+
 from wtforms import Form, SelectField, StringField
 
-from app.validation.validators import DateCheck, OptionalForm, DateRequired, MonthYearCheck
+from app.questionnaire.rules import get_metadata_value, get_answer_store_value, convert_to_datetime
+from app.validation.validators import DateCheck, OptionalForm, DateRequired, MonthYearCheck, SingleDatePeriodCheck
 
 logger = logging.getLogger(__name__)
 
 
-def get_date_form(answer=None, error_messages=None):
+def get_date_form(answer_store, metadata, answer=None, error_messages=None):
     """
     Returns a date form metaclass with appropriate validators. Used in both date and
     date range form creation.
 
     :param error_messages: The messages during validation
     :param answer: The answer on which to base this form
+    :param answer_store: The current answer store
+    :param metadata: metadata for reference meta dates
     :return: The generated DateForm metaclass
     """
     class DateForm(Form):
@@ -23,10 +30,10 @@ def get_date_form(answer=None, error_messages=None):
 
     validate_with = [OptionalForm()]
 
-    if not error_messages:
-        date_messages = {}
-    else:
+    if error_messages:
         date_messages = error_messages.copy()
+    else:
+        date_messages = {}
 
     if answer['mandatory'] is True:
         if 'validation' in answer and 'messages' in answer['validation'] \
@@ -39,7 +46,14 @@ def get_date_form(answer=None, error_messages=None):
             and 'INVALID_DATE' in answer['validation']['messages']:
         date_messages['INVALID_DATE'] = answer['validation']['messages']['INVALID_DATE']
 
-    validate_with += [DateCheck(date_messages['INVALID_DATE'])]
+    validate_with.append(DateCheck(date_messages['INVALID_DATE']))
+
+    if 'minimum' in answer or 'maximum' in answer:
+        messages = None
+        if 'validation' in answer:
+            messages = answer['validation'].get('messages')
+        minimum_date, maximum_date = get_dates_for_single_date_period_validation(answer, answer_store, metadata)
+        validate_with.append(SingleDatePeriodCheck(messages=messages, minimum_date=minimum_date, maximum_date=maximum_date))
 
     # Set up all the calendar month choices for select
     month_choices = [('', 'Select month')] + [(str(x), calendar.month_name[x]) for x in range(1, 13)]
@@ -84,3 +98,64 @@ def get_month_year_form(answer, error_messages):
     MonthYearDateForm.month = SelectField(choices=month_choices, default='', validators=validate_with)
 
     return MonthYearDateForm
+
+
+def get_dates_for_single_date_period_validation(answer, answer_store, metadata):
+    """
+    Gets attributes within a minimum or maximum of a date field and validates that the entered date
+    is valid.
+
+    :param answer: The answer which contains the minimum or maximum
+    :param answer_store: The current answer store
+    :param metadata: metadata for reference meta dates
+    :return: attributes
+    """
+    minimum_referenced_date, maximum_referenced_date = None, None
+
+    if 'minimum' in answer:
+        minimum_referenced_date = get_referenced_offset_value(answer['minimum'], answer_store, metadata)
+    if 'maximum' in answer:
+        maximum_referenced_date = get_referenced_offset_value(answer['maximum'], answer_store, metadata)
+
+    # Extra runtime validation that will catch invalid schemas
+    # Similar validation in schema validator
+    if minimum_referenced_date and maximum_referenced_date:
+        if minimum_referenced_date > maximum_referenced_date:
+            raise Exception('The minimum offset date is greater than the maximum offset date for {}.'.format(answer['id']))
+
+    return minimum_referenced_date, maximum_referenced_date
+
+
+def get_referenced_offset_value(answer_min_or_max, answer_store, metadata):
+    """
+    Gets value of the referenced date type, whether it is a value,
+    id of an answer or a meta date. Then adds/subtracts offset from that value and returns
+    the new offset value
+
+    :param answer_min_or_max: The minimum or maximum object which contains
+    the referenced value.
+    :param answer_store: The current answer store
+    :param metadata: metadata for reference meta dates
+    :return: date value
+    """
+    value = None
+
+    if 'value' in answer_min_or_max:
+        if answer_min_or_max['value'] == 'now':
+            value = datetime.utcnow().strftime('%Y-%m-%d')
+        else:
+            value = answer_min_or_max['value']
+    elif 'meta' in answer_min_or_max:
+        value = get_metadata_value(metadata, answer_min_or_max['meta'])
+    elif 'answer_id' in answer_min_or_max:
+        value = get_answer_store_value(answer_min_or_max['answer_id'], answer_store, group_instance=0)
+
+    value = convert_to_datetime(value)
+
+    if 'offset_by' in answer_min_or_max:
+        offset = answer_min_or_max['offset_by']
+        value += relativedelta(years=offset.get('years', 0),
+                               months=offset.get('months', 0),
+                               days=offset.get('days', 0))
+
+    return value
--- a/app/forms/fields.py
+++ b/app/forms/fields.py
@@ -15,16 +15,17 @@ MAX_DECIMAL_PLACES = 6
 logger = get_logger()
 
 
-def get_field(answer, label, error_messages, answer_store):
+def get_field(answer, label, error_messages, answer_store, metadata):
     guidance = answer.get('guidance', '')
 
     if answer['type'] in ['Number', 'Currency', 'Percentage', 'Unit']:
         field = get_number_field(answer, label, guidance, error_messages, answer_store)
+    elif answer['type'] == 'Date':
+        field = get_date_field(answer, label, guidance, error_messages, answer_store, metadata)
     else:
         field = {
             'Radio': get_select_field,
             'Checkbox': get_select_multiple_field,
-            'Date': get_date_field,
             'MonthYearDate': get_month_year_field,
             'TextArea': get_text_area_field,
             'TextField': get_string_field,
@@ -103,10 +104,10 @@ def get_text_area_field(answer, label, guidance, error_messages):
     )
 
 
-def get_date_field(answer, label, guidance, error_messages):
+def get_date_field(answer, label, guidance, error_messages, answer_store, metadata):
 
     return FormField(
-        get_date_form(answer, error_messages=error_messages),
+        get_date_form(answer_store=answer_store, metadata=metadata, answer=answer, error_messages=error_messages),
         label=label,
         description=guidance,
     )
--- a/app/forms/questionnaire_form.py
+++ b/app/forms/questionnaire_form.py
@@ -1,13 +1,17 @@
 import logging
+from datetime import datetime, timedelta
 
 from decimal import Decimal
 
 import itertools
+
+from dateutil.relativedelta import relativedelta
 from wtforms import validators
 from flask_wtf import FlaskForm
 from werkzeug.datastructures import MultiDict
 
 from app.forms.fields import get_field
+from app.forms.date_form import get_dates_for_single_date_period_validation
 from app.validation.validators import DateRangeCheck, SumCheck
 
 logger = logging.getLogger(__name__)
@@ -15,10 +19,11 @@ logger = logging.getLogger(__name__)
 
 class QuestionnaireForm(FlaskForm):
 
-    def __init__(self, schema, block_json, answer_store, formdata=None, **kwargs):
+    def __init__(self, schema, block_json, answer_store, metadata, formdata=None, **kwargs):
         self.schema = schema
         self.block_json = block_json
         self.answer_store = answer_store
+        self.metadata = metadata
         self.question_errors = {}
         self.options_with_children = {}
 
@@ -33,19 +38,43 @@ class QuestionnaireForm(FlaskForm):
         :return:
         """
         valid_fields = FlaskForm.validate(self)
-        valid_form = True
+        valid_date_range_form = True
+        valid_calculated_form = True
 
         for question in self.schema.get_questions_for_block(self.block_json):
             if question['type'] == 'DateRange':
-                period_from_id = question['answers'][0]['id']
-                period_to_id = question['answers'][1]['id']
-                if not (self.answers_all_valid([period_from_id, period_to_id]) and
-                        self._validate_date_range_question(question['id'], period_from_id, period_to_id)):
-                    valid_form = False
+                valid_date_range_form = self.validate_date_range_question(question)
             elif question['type'] == 'Calculated':
-                valid_form = self.validate_calculated_question(question)
+                valid_calculated_form = self.validate_calculated_question(question)
 
-        return valid_form and valid_fields
+        return valid_calculated_form and valid_date_range_form and valid_fields
+
+    def validate_date_range_question(self, question):
+        date_from = question['answers'][0]
+        date_to = question['answers'][1]
+        if self._has_min_and_max_single_dates(date_from, date_to):
+            # Work out the largest possible range, for date range question
+            period_range = self._get_period_range_for_single_date(date_from, date_to)
+            if 'period_limits' in question:
+                self.validate_date_range_with_period_limits_and_single_date_limits(question['id'],
+                                                                                   question['period_limits'],
+                                                                                   period_range)
+            else:
+                # Check every field on each form has populated data
+                if self.is_date_form_populated(getattr(self, date_from['id']), getattr(self, date_to['id'])):
+                    self.validate_date_range_with_single_date_limits(question['id'], period_range)
+        period_from_id = date_from['id']
+        period_to_id = date_to['id']
+        messages = None
+        if 'validation' in question:
+            messages = question['validation'].get('messages')
+        if not (
+                self.answers_all_valid([period_from_id, period_to_id]) and
+                self._validate_date_range_question(question['id'], period_from_id, period_to_id, messages,
+                                                   question.get('period_limits'))):
+            return False
+
+        return True
 
     def validate_calculated_question(self, question):
         for calculation in question['calculations']:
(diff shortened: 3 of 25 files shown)


data/en/test_date_range_period_validation.json
```diff
@@ -0,0 +1,53 @@
...
+            "blocks": [{
+                    "type": "Question",
+                    "id": "date-range-block",
+                    "title": "Date Range",
+                    "questions": [{
+                            "id": "date-range-question",
+                            "title": "Enter Date Range",
+                            "type": "DateRange",
+                            "period_limits": {
+                              "min_value": 23,
```
dwyeradam: 23 what? Years, Months, Days, Minutes, Seconds, Flamingos? Check the format of offset on date routing.

dwyeradam (changes requested):

data/en/test_date_range_period_validation.json
dwyeradam: Also does min_value define what this is? Is min_range clearer?

shahi645: its days, and min_limit perhaps ?

shahi645: ok so i think `minimum_day_limit` will clear it up a little ?

dwyeradam: My point is that we shouldn't limit ourselves to just days. Years, Months, Weeks and Days could all be valid requirements in the future.

shahi645: This is a bit like the average stuff in the calculated validation, should i add the ability for it when there are no business requirements for it? And then a few days down the line im backing the changes out because its not needed. The card only asks for period limits to be set in days. I am more than happy to add the ability for other units of time but if it is reuested

app/forms/questionnaire_form.py
```diff
@@ -39,9 +39,13 @@ def validate(self):
             if question['type'] == 'DateRange':
                 period_from_id = question['answers'][0]['id']
                 period_to_id = question['answers'][1]['id']
+                messages = None
+                if 'validation' in question:
+                    messages = question['validation'].get('messages')
```
dwyeradam: Good addition.

data/en/test_date_range_period_validation.json
dwyeradam: Consider using dateutil.relativedelta in your date comparison to help compare difference in years, months, days etc.

app/forms/questionnaire_form.py
```diff
@@ -104,10 +104,10 @@ def _validate_calculated_question(self, question, target_answer, target_total):
     def _get_period_limits(limits):
         minimum, maximum = None, None
         if limits:
-            if 'min_value' in limits:
-                minimum = limits['min_value']
-            if 'max_value' in limits:
-                maximum = limits['max_value']
+            if 'minimum_days_limit' in limits:
+                minimum = limits['minimum_days_limit']
+            if 'maximum_days_limit' in limits:
```
dwyeradam: This hasn't addressed my previous comment. "We shouldn't limit ourselves to just days. Years, Months, Weeks and Days could all be valid requirements in the future." This can quite easily be done by having a structure as below and using **dateutil.relativedelta**.
```
"min": {
   "days": 23
}

```
or 

```
"min": {
   "months": 3
   "years": 1
}
```

dwyeradam (changes requested):

app/forms/questionnaire_form.py
shahi645: @dwyeradam I did reply on your last change. There is no requirements for month and year in between a date range. I have double checked with Donna also everything upcoming is for days in between, the trello card also describes days as the only requirement. Also months/years can be representing in days also. 
In regard the `relativedelta` I am just in process of changing over but its causing some issues so still working througth that

jonnyshaw89: @shahi645 it may not be required now, but if it is in the future then we are going to end up having a messy schema.
This pattern has already been introduced so it seems sensible to me to keep this kind of definition consistent across the schema.

dwyeradam: @shahi645. I agree with Jon. We're building a feature not a survey. That feature would be better written allowing for more than just days. Also the code required for that does not add any additional complication. As I keep saying if you research the use of relativedelta you'd realise days, months, years, weeks can all be done in a single line.

shahi645: Yes i don't disagree with that i have looked in to relativedelta its simple enough im working on those changes currently. In regards to the days/weeks/months there is no future requirement for it. This is the exact same thing that was said for average on the calculated question type and it just got backed out in the end.

app/helpers/form_helper.py
```diff
@@ -11,14 +11,15 @@
 logger = get_logger()
 
 
-def get_form_for_location(schema, block_json, location, answer_store, disable_mandatory=False):
+def get_form_for_location(schema, block_json, location, answer_store, metadata, disable_mandatory=False):  # pylint: disable=too-many-locals
     """
     Returns the form necessary for the location given a get request, plus any template arguments
 
+    :param schema: schema
```
MIslam92, line 18: Might want to expand on this docstring i.e. the current schema json or something

app/forms/date_form.py
```diff
@@ -84,3 +98,61 @@ class MonthYearDateForm(Form):
...
+        if minimum_referenced_date > maximum_referenced_date:
+            raise Exception('The minimum offset date is greater than the maximum offset date for {}.'.format(answer['id']))
+
+    return minimum_referenced_date, maximum_referenced_date
+
+
+def get_referenced_offset_value(answer_min_or_max, answer_store, metadata):
+    """
+    Gets value of the referenced date type, whether it is a value,
+    id of an answer or a meta date. Then adds/subtracts offset from that value and
```
MIslam92: Did you have a stroke midway and forget to finish this off?

app/validation/error_messages.py
```diff
@@ -22,4 +22,8 @@
     'MAX_LENGTH_EXCEEDED': 'Your answer is too long, it has to be less than %(max)d characters.',
     'INVALID_DATE': 'Enter a valid date.',
     'INVALID_DATE_RANGE': "Enter a 'period to' date later than the 'period from' date.",
+    'DATE_PERIOD_TOO_SMALL': 'Enter a reporting period greater than %(min)s.',
+    'DATE_PERIOD_TOO_BIG': 'Enter a reporting period less than %(max)s.',
```
MIslam92: Might want to use `DATE_PERIOD_TOO_LARGE` as this is the terminology we use in runner

MIslam92 (changes requested):

app/forms/date_form.py
shahi645: 🌚

app/forms/date_form.py
```diff
@@ -41,6 +48,13 @@ class DateForm(Form):
 
     validate_with += [DateCheck(date_messages['INVALID_DATE'])]
 
+    if 'minimum' in answer or 'maximum' in answer:
+        messages = None
+        if 'validation' in answer:
+            messages = answer['validation'].get('messages')
+        minimum_date, maximum_date = get_min_max_dates_for_single_date_period_validation(answer, answer_store, metadata)
+        validate_with += [SingleDatePeriodCheck(messages=messages, minimum_date=minimum_date, maximum_date=maximum_date)]
```
nealedj: Minor point - but let's not create a list just to append a single item to a list. Use `validate_with.append()`

Whilst there - maybe also change the same thing on line 49.

app/forms/date_form.py
```diff
@@ -84,3 +98,62 @@ class MonthYearDateForm(Form):
...
+    :return: attributes
+    """
+    minimum_referenced_date, maximum_referenced_date = None, None
+
+    if 'minimum' in answer:
+        minimum_referenced_date = get_referenced_offset_value(answer['minimum'], answer_store, metadata)
+    if 'maximum' in answer:
+        maximum_referenced_date = get_referenced_offset_value(answer['maximum'], answer_store, metadata)
+
+    if minimum_referenced_date and maximum_referenced_date:
```
nealedj: Could you please add a little comment to mention that this should never happen as validator should prevent it.

app/forms/date_form.py
```diff
@@ -84,3 +98,62 @@ class MonthYearDateForm(Form):
...
+
+    if 'value' in answer_min_or_max:
+        if answer_min_or_max['value'] == 'now':
+            value = datetime.utcnow().strftime('%Y-%m-%d')
+        else:
+            value = answer_min_or_max['value']
+    elif 'meta' in answer_min_or_max:
+        value = get_metadata_value(metadata, answer_min_or_max['meta'])
+    elif 'answer_id' in answer_min_or_max:
+        value = get_answer_store_value(answer_min_or_max['answer_id'], answer_store, group_instance=0)
```
nealedj, line 151: This would fail in a repeating group wouldn't it? As it would always validate against the value of the answer in the first group instance, not the current group instance.

app/forms/date_form.py
```diff
@@ -84,3 +98,62 @@ class MonthYearDateForm(Form):
...
+
+    if 'offset_by' in answer_min_or_max:
+        offset = answer_min_or_max['offset_by']
+        value = convert_to_datetime(value) + relativedelta(years=offset.get('years', 0),
+                                                           months=offset.get('months', 0),
+                                                           days=offset.get('days', 0))
+    else:
+        value = convert_to_datetime(value)
```
nealedj: `value = convert_to_datetime(value)` is repeated. This could be simplified to:

```
value = convert_to_datetime(value)

if 'offset_by' in answer_min_or_max:
    offset = answer_min_or_max['offset_by']
    value += relativedelta(years=offset.get('years', 0),
                           months=offset.get('months', 0),
                           days=offset.get('days', 0))
```

app/forms/date_form.py
```diff
@@ -84,3 +98,62 @@ class MonthYearDateForm(Form):
...
+        else:
+            value = answer_min_or_max['value']
+    elif 'meta' in answer_min_or_max:
+        value = get_metadata_value(metadata, answer_min_or_max['meta'])
+    elif 'answer_id' in answer_min_or_max:
+        value = get_answer_store_value(answer_min_or_max['answer_id'], answer_store, group_instance=0)
+
+    if 'offset_by' in answer_min_or_max:
+        offset = answer_min_or_max['offset_by']
+        value = convert_to_datetime(value) + relativedelta(years=offset.get('years', 0),
```
nealedj: Is there any reason that `relativedelta` is being used rather than the basic, builtin `datetime.timedelta`?

app/forms/questionnaire_form.py
```diff
@@ -105,6 +111,16 @@ def _validate_calculated_question(self, calculation, question, target_total, cur
 
         return True
 
+    @staticmethod
+    def _get_period_limits(limits):
```
nealedj: Could be simplified to:

```
limits = limits or {}
return limits.get('minimum'), limits.get('maximum')
````

File: app/validation/error_messages.py [hunk 13]
```diff
@@ -22,4 +22,8 @@
     'MAX_LENGTH_EXCEEDED': 'Your answer is too long, it has to be less than %(max)d characters.',
     'INVALID_DATE': 'Enter a valid date.',
     'INVALID_DATE_RANGE': "Enter a 'period to' date later than the 'period from' date.",
+    'DATE_PERIOD_TOO_SMALL': 'Enter a reporting period greater than %(min)s.',
```
nealedj: I'm sure this conversation has already happened - but why are we hard-coding a specific error message about reporting periods when the validation is generic date range validation?

app/validation/validators.py
```diff
@@ -73,9 +75,10 @@ def __init__(self, minimum=None, minimum_exclusive=False,
         self.maximum = maximum
         self.minimum_exclusive = minimum_exclusive
         self.maximum_exclusive = maximum_exclusive
-        if not messages:
-            messages = error_messages
-        self.messages = messages
+        if messages:
```
nealedj: Could be simplified to:

`self.messages = messages or error_messages`

app/validation/validators.py
```diff
@@ -212,31 +216,93 @@ def __call__(self, form, field):
...
 
-    def __call__(self, form, from_field, to_field):
+    def __call__(self, form, field):
+        date_str = format_date_string(form)
+        date = convert_to_datetime(date_str)
+
+        if self.minimum_date:
+            if date < self.minimum_date:
+                raise validators.ValidationError(self.messages['SINGLE_DATE_PERIOD_TOO_EARLY'] %
+                                                 dict(min=self._format_playback_date(self.minimum_date + relativedelta(days=-1))))
```
nealedj: Same comment again about `relativedelta` - is it giving us anything that we don't already have with `datetime.timedelta`?

app/validation/validators.py
```diff
@@ -283,3 +349,10 @@ def format_playback_value(value, currency=None):
     if currency:
         return format_currency(value, currency)
     return format_number(value)
+
+
+def format_date_string(field):
+    day = int(field.data['day']) if 'day' in field.data else 1
```
nealedj: Simplified:

`day = int(field.data.get('day', 1))`

nealedj (changes requested):
Looks good - nearly all my comments are minor.

The only big thing might be whether this validation will work within a repeating group.

app/forms/date_form.py
shahi645: relativedelta deals with months and year and timedelta only days

app/validation/error_messages.py
shahi645: Speaking to Will he just said to use this for now, and then we can change as we need

app/forms/date_form.py
shahi645: Ticket raised in planning backlog to address this, ass per Davids request

app/validation/validators.py
shahi645: nice

app/validation/validators.py
shahi645: answered above, actually looking at it again, it only deals with days, second, microseconds weirdly 😞

nealedj (approved):

app/forms/date_form.py
dwyeradam: @nealedj, @shahi645. Why are we introducing tech debt? In my opinion if we know this feature will fail in certain conditions it shouldn't be put in. I wouldn't class this as an edge case as we already know we're likely to need this in repeating groups for both Census and LFS.

nealedj: I agree that it's not an edge case and that it's not good to introduce tech debt.

But on balance I don't think we gain anything by holding up this feature before fixing it given the fact that it's not a bug that will manifest itself in any current questionnaire.

It's also a fix that needs to address number range validation under the same conditions.

dwyeradam: Having spoken with @nealedj, I realise this is the same format as used by numbers so a separate PR will be needed for that either way. To get this PR through then i'm happy for this to be addressed at the same time and in the same fashion as will be needed for numbers.

app/forms/fields.py
```diff
@@ -15,16 +15,17 @@
...
     if answer['type'] in ['Number', 'Currency', 'Percentage', 'Unit']:
         field = get_number_field(answer, label, guidance, error_messages, answer_store)
+    elif answer['type'] == 'Date':
+        field = get_date_field(answer, label, guidance, error_messages, answer_store, metadata)
     else:
         field = {
             'Radio': get_select_field,
             'Checkbox': get_select_multiple_field,
-            'Date': get_date_field,
             'MonthYearDate': get_month_year_field,
```
dwyeradam, line 29: Does `MonthYearDate` not also need to allow for date validation? Please don't say this isn't a future requirement unless you tell me the winning numbers for the next lottery draw.

app/forms/questionnaire_form.py
```diff
@@ -66,10 +85,40 @@ def _get_target_total_and_currency(self, calculation, question):
         target_answer = self.schema.get_answer(calculation['answer_id'])
         return self.answer_store.filter(answer_ids=[target_answer['id']]).values()[0], target_answer.get('currency')
 
-    def _validate_date_range_question(self, question_id, period_from_id, period_to_id):
+    def validate_date_range_with_period_limits_and_single_date_limits(self, question_id, period_limits, date_from, date_to):  # pylint: disable=too-many-locals
```
dwyeradam: Remove `# pylint: disable=too-many-locals`. It will pass without this. If it didn't pass Lint you should really address the code rather than add an ignore.

app/forms/questionnaire_form.py
```diff
@@ -66,10 +85,40 @@ def _get_target_total_and_currency(self, calculation, question):
...
+        if self._has_min_and_max_single_dates(date_from, date_to):
```
dwyeradam: This would be clearer if it was after [`if question['type'] == 'DateRange':`](https://github.com/ONSdigital/eq-survey-runner/pull/1521/files?diff=unified#diff-c5e43bea5b19fce9acef957e0a446c22R44) as both the methods that are then called below start with that check. You can then also remove it from  [`validate_date_range_with_single_date_limits`](https://github.com/ONSdigital/eq-survey-runner/pull/1521/files?diff=unified#diff-c5e43bea5b19fce9acef957e0a446c22R106). 
It's better practice to only call a method when needed and not to call it regardless and let the method decide if it should have been called.

dwyeradam (changes requested):
A couple of little early points while I continue to review.

app/forms/fields.py
shahi645: I don't think so as they work on a referenced date, so wouldn't be required

dwyeradam: I'm pretty sure the LFS stuff @davecharles was working on had Month Year dates for when you arrived in the UK that couldn't be before dob. We may not need date range but are we sure we don't need single validation setting min and max? I could see a future need there but @nealedj what are your thoughts? @charlotte-hirst are you expecting this for all date types?

shahi645: Can you advise @charlotte-hirst also. As you are in contact with Social & Census

app/forms/questionnaire_form.py
```diff
@@ -37,10 +42,26 @@ def validate(self):
 
         for question in self.schema.get_questions_for_block(self.block_json):
             if question['type'] == 'DateRange':
-                period_from_id = question['answers'][0]['id']
-                period_to_id = question['answers'][1]['id']
-                if not (self.answers_all_valid([period_from_id, period_to_id]) and
-                        self._validate_date_range_question(question['id'], period_from_id, period_to_id)):
+                date_from = question['answers'][0]
+                date_to = question['answers'][1]
```
dwyeradam: There is quite a bit of code in this if statement now. It may be worth considering moving it into a `self.validate_daterange_question(.....)` method as it does below for calculation and returning the result.

app/forms/questionnaire_form.py
```diff
@@ -37,10 +42,26 @@ def validate(self):
...
+                messages = None
+                if 'validation' in question:
+                    messages = question['validation'].get('messages')
+                if not (
+                        self.answers_all_valid([period_from_id, period_to_id]) and
+                        self._validate_date_range_question(question['id'], period_from_id, period_to_id, messages,
+                                                           question.get('period_limits'))):
                     valid_form = False
             elif question['type'] == 'Calculated':
                 valid_form = self.validate_calculated_question(question)
```
dwyeradam: You haven't added this issue in this PR but applying "Boy Scout" rule. If we ever got a block with a date range and a calculation on it (we shouldn't do but I don't think there is anything to stop that possibility) the calculation validation will override the setting of valid_form.

dwyeradam (changes requested):
Two little things but looking good. Only real issue we still need to discuss is the fact that no validation is available for MonthYear dates.

app/forms/fields.py
nealedj: I think we us to consider date range validation "done" then we should support all types of date question. However, I don't think there's any reason to block this PR from going in at the moment. The code changes are becoming pretty large and it's becoming exponentially more difficult to review.

I'd be in favour of closing this off as it adds much-needed value to master. The job of applying the validation to `MonthYearDate` could be considered a separate piece of work - or Part 2 (the validation strikes back).

Let's keep this sort of thing in check though. In an ideal world this PR would cover all use-cases.

app/forms/questionnaire_form.py
nealedj: Good catch.

app/forms/fields.py
shahi645: OK will put a ticket into backlog with some fleshed out AC

MIslam92 (approved):
LGTM

app/validation/error_messages.py
```diff
@@ -22,4 +22,8 @@
...
+    'DATE_PERIOD_TOO_LARGE': 'Enter a reporting period less than %(max)s.',
```
MebinAbraham: The error messages are wrong.
```
"period_limits": {
	"minimum": {
		"days": 31
     },
	"maximum": {
		"days": 31
	}
}
```
If you use the above settings and if the period difference is 32, the error message shown is:
`1) Enter a reporting period less than 31 days.`
It should be: `less than 32 days` or `less than or equal to 31 days`

If the period difference is 30, the error message shown is:
`1) Enter a reporting period greater than 31 days.`
It should be: `greater than 30 days` or `greater than or equal to 31 days`

The range itself is fine, just need to tweak the error messages. This will apply to `days`, `months` and `years`.

MebinAbraham (changes requested):

app/validation/error_messages.py
shahi645: Schema validator wont allow same period limits but error messages are a bit strange there i will dig around a little and get back

shahi645: Just looking at these again the error messages are correct

MebinAbraham: Minimum and maximum are generally inclusive. But also your point about schema validator not allowing?. For me, it says it is a valid schema?

Also, what am I missing?. How is it a valid error message?. `31` is a valid range but the error message would say ie less than `31`.?

dwyeradam (approved):

dwyeradam (changes requested):
On `test_single_date_period_validation.json` if you don't enter anything on the second page it throws a 500 error. If it's not Mandatory we wouldn't want to 500 although in this instance it still does. Edge case but at minimum a new card. The issue is that no valid range can be entered, not entering anything however would be valid.

MebinAbraham (approved):
The error messages has been sorted.
I think Adam point can be raised as a new card and this can go in.
There are certain edge cases we can improve but for now, looks good 👍 
I think the other cards raised from this PR should be moved to the top of ready for dev, to be picked up after MBS goes in.

dwyeradam (approved):
I relent, lets get this bad boy in. 
It would help to add some guidance into `test_single_date_period_validation.json` to help guide what date to enter that won't 500.

dwyeradam:
@shahi645, squishy, squashy, good commit messagey.

shahi645:
@dwyeradam Ill add the guidance, thats a good point. Update with aster. Then squash

PR #1527, opened 2018-03-27
Fix date range check tests

MebinAbraham:
A quick change to fix tests that are passing but are not actually valid as they do not replicate the behaviour of runner.
Changes went in this PR: #1517 

### How to review 
1. Ensure all tests actually pass and are valid tests.
2. Ensure any functionality related to `DateRange` is still working as expected.

### DIFF
--- a/tests/app/validation/test_date_range_validator.py
+++ b/tests/app/validation/test_date_range_validator.py
@@ -1,5 +1,5 @@
 import unittest
-from unittest.mock import MagicMock
+from unittest.mock import Mock
 from wtforms.validators import ValidationError
 
 from app.validation.error_messages import error_messages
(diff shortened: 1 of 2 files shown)


tests/app/validation/test_date_range_validator.py
```diff
@@ -58,19 +65,69 @@ def test_date_range_valid(self):
...
+            'month': '01',
+            'year': '2017'
+        }
+
+        mock_form = Mock()
+
+        try:
+            validator(mock_form, period_from, period_to)
+        except ValidationError:
+            self.fail('Valid date raised ValidationError')
```
jonnyshaw89: why are we not asserting the exception here?

jonnyshaw89 (changes requested):

tests/app/validation/test_date_range_validator.py
MebinAbraham: @jonnyshaw89 Because we are not expecting it to fail?

tests/app/validation/test_date_range_validator.py
```diff
@@ -58,19 +65,69 @@ def test_date_range_valid(self):
...
+
+    def test_date_range_missing_day(self):
```
shahi645: needs rename

shahi645 (approved):

jonnyshaw89 (approved):

PR #1529, opened 2018-03-28
Modified filter to output correct time  with UTC

MIslam92:
Problem displaying times as UTC so now displaying as GMT with the +1

Card: https://trello.com/c/zYD6SlAX/1983-provide-locale-specific-time-for-users-when-server-response-is-in-utc-s

### How to review 
Check that on the view submission page and the after submission page show GMT(+1) on any survey/test surveys.

### DIFF
--- a/app/jinja_filters.py
+++ b/app/jinja_filters.py
@@ -1,15 +1,15 @@
 # coding: utf-8
-
 import re
 import string
 
 from datetime import datetime
+from dateutil import relativedelta, tz
 
-from dateutil import relativedelta
 import flask
 
 from jinja2 import Markup, escape, evalcontextfilter
 from babel import units, numbers
+
 
 from app.settings import DEFAULT_LOCALE
 from app.questionnaire.rules import convert_to_datetime
@@ -82,7 +82,11 @@ def format_month_year_date(value, date_format='%B %Y'):
 
 @blueprint.app_template_filter()
 def format_datetime(value, date_format='%d %B %Y at %H:%M'):
-    return "<span class='date'>{date}</span>".format(date=datetime.strptime(value, '%Y-%m-%dT%H:%M:%S.%f').strftime(date_format))
+    london = datetime.strptime(value, '%Y-%m-%dT%H:%M:%S.%f').replace(tzinfo=tz.gettz('UTC'))\
+        .astimezone(tz.gettz('Europe/London'))
+    date_time = london.strftime(date_format)
+    timezone = '(GMT+1)' if london.dst() else '(GMT)'
+    return "<span class='date'>{date} {timezone}</span>".format(date=date_time, timezone=timezone)
 
 
 @blueprint.app_template_filter()
--- a/tests/app/test_jinja_filters.py
+++ b/tests/app/test_jinja_filters.py
@@ -9,9 +9,9 @@ from mock import Mock
 
 from app.jinja_filters import format_date, format_conditional_date, format_currency, get_currency_symbol, \
     format_multilined_string, format_percentage, format_date_range, format_household_member_name, \
-    format_month_year_date, format_number_to_alphabetic_letter, \
-    format_unit, format_currency_for_input, format_number, format_unordered_list, format_household_member_name_possessive, \
-    concatenated_list, calculate_years_difference
+    format_month_year_date, format_datetime, format_number_to_alphabetic_letter, \
+    format_unit, format_currency_for_input, format_number, format_unordered_list, \
+    format_household_member_name_possessive, concatenated_list, calculate_years_difference
 
 
 class TestJinjaFilters(TestCase):  # pylint: disable=too-many-public-methods
@@ -128,6 +128,28 @@ class TestJinjaFilters(TestCase):  # pylint: disable=too-many-public-methods
 
         # Then
         self.assertEqual(format_value, "<span class='date'>January 2017</span>")
+
+    def test_format_date_time_in_bst(self):
+        # Given
+        date_time = '2018-03-29T11:59:13.528680'
+        date_format = '%d %B %Y at %H:%M'
+
+        # When
+        format_value = format_datetime(date_time, date_format)
+
+        # Then
+        self.assertEqual(format_value, "<span class='date'>29 March 2018 at 12:59 (GMT+1)</span>")
+
+    def test_format_date_time_in_gmt(self):
+        # Given
+        date_time = '2018-10-28T11:59:13.528680'
+        date_format = '%d %B %Y at %H:%M'
+
+        # When
+        format_value = format_datetime(date_time, date_format)
+
+        # Then
+        self.assertEqual(format_value, "<span class='date'>28 October 2018 at 11:59 (GMT)</span>")
 
     def test_format_conditional_date_not_date(self):
         # Given       no test for integers this check was removed from jinja_filters


app/jinja_filters.py
```diff
@@ -82,7 +81,9 @@ def format_month_year_date(value, date_format='%B %Y'):
 
 @blueprint.app_template_filter()
 def format_datetime(value, date_format='%d %B %Y at %H:%M'):
-    return "<span class='date'>{date}</span>".format(date=datetime.strptime(value, '%Y-%m-%dT%H:%M:%S.%f').strftime(date_format))
+    if datetime.now().hour > datetime.utcnow().hour:
+        return "<span class='date'>{date} GMT(+1)</span>".format(date=datetime.strptime(value, '%Y-%m-%dT%H:%M:%S.%f').strftime(date_format))
+    return "<span class='date'>{date} GMT</span>".format(date=datetime.strptime(value, '%Y-%m-%dT%H:%M:%S.%f').strftime(date_format))
```
jonnyshaw89: Too much duplication here. all you really need to do is work out if we are BST and add the `+1`

tests/app/test_jinja_filters.py
```diff
@@ -129,6 +129,16 @@ def test_format_date_month_year(self):
         # Then
         self.assertEqual(format_value, "<span class='date'>January 2017</span>")
 
+    def test_format_date_time(self):
+        # Given
+        date_time = '2018-03-28T13:59:00.659262'
+
+        # When
+        format_value = format_datetime(date_time, date_format='%d %B %Y at %H:%M')
```
jonnyshaw89: What happens on 28th October when BST ends?

jonnyshaw89 (dismissed):

tests/app/test_jinja_filters.py
hamishtaplin: Clocks go back 🙄

jonnyshaw89: and the evaluation of `if datetime.now().hour > datetime.utcnow().hour:` changes

jonnyshaw89: @hamishtaplin also, they go back ;)

hamishtaplin: Not in my house.

hamishtaplin: @jonnyshaw89 Yeah I know, that's what I wrote and I didn't just edit it.

app/jinja_filters.py
```diff
@@ -82,7 +81,9 @@ def format_month_year_date(value, date_format='%B %Y'):
 
 @blueprint.app_template_filter()
 def format_datetime(value, date_format='%d %B %Y at %H:%M'):
-    return "<span class='date'>{date}</span>".format(date=datetime.strptime(value, '%Y-%m-%dT%H:%M:%S.%f').strftime(date_format))
+    if datetime.now().hour > datetime.utcnow().hour:
+        return "<span class='date'>{date} GMT(+1)</span>".format(date=datetime.strptime(value, '%Y-%m-%dT%H:%M:%S.%f').strftime(date_format))
```
dwyeradam: Don't let our users do the maths. This is still displaying GMT.

dwyeradam (changes requested):

app/jinja_filters.py
```diff
@@ -82,7 +85,15 @@ def format_month_year_date(value, date_format='%B %Y'):
...
 @blueprint.app_template_filter()
 def format_datetime(value, date_format='%d %B %Y at %H:%M'):
-    return "<span class='date'>{date}</span>".format(date=datetime.strptime(value, '%Y-%m-%dT%H:%M:%S.%f').strftime(date_format))
+    date = datetime.strptime(value, '%Y-%m-%dT%H:%M:%S.%f')
+    tz_time = pytz.timezone('UTC').localize(date)
+    london_tz = pytz.timezone('Europe/London')
+    london_time = tz_time.astimezone(london_tz)
+    time = london_time.strftime(date_format)
+    if london_time.tzinfo._utcoffset == timedelta(hours=1):
+        return "<span class='date'>{date} GMT (+1)</span>".format(date=datetime.strptime(time, date_format))
```
nealedj: AFAIK technically this would either be UTC+1 or BST.

app/jinja_filters.py
```diff
@@ -82,7 +85,15 @@ def format_month_year_date(value, date_format='%B %Y'):
 
 @blueprint.app_template_filter()
 def format_datetime(value, date_format='%d %B %Y at %H:%M'):
-    return "<span class='date'>{date}</span>".format(date=datetime.strptime(value, '%Y-%m-%dT%H:%M:%S.%f').strftime(date_format))
+    date = datetime.strptime(value, '%Y-%m-%dT%H:%M:%S.%f')
+    tz_time = pytz.timezone('UTC').localize(date)
+    london_tz = pytz.timezone('Europe/London')
+    london_time = tz_time.astimezone(london_tz)
+    time = london_time.strftime(date_format)
+    if london_time.tzinfo._utcoffset == timedelta(hours=1):
```
nealedj: The tzinfo object already implements a `dst()` method to work this out.

nealedj (changes requested):

app/jinja_filters.py
dwyeradam: We believed people were more likely to understand GMT compared to UTC. Also GMT+1 would be more likely understood for foreign users than BST.

armstrongb:
Does runner (or other downstream systems) have any checks in place about started time of survey vs submitted time? If someone happened to be doing this on Oct 28th they could submit before they’d started…

dwyeradam:
@armstrongb we store all dates as UTC which isn't affected by GMT/BST switches. Any checks would be carried out against the stored UTC time and not a modified local time. This change merely takes the stored UTC time and modifies it for the screen.

app/jinja_filters.py
nealedj: Yep okay - whether a timezone name is understandable or not is very subjective so this seems like a reasonable take on it.

dwyeradam:
If you've added a new dependency you need to also add it to the Pipfile
`pipenv install tzlocal`

app/jinja_filters.py
dwyeradam: Should we use the `if` to create a variable called something like timezone and then pass that as another format variable. Then we're left with a single return and lessen duplication.
`return "<span class='date'>{date} {timezone}</span>".format(date=datetime.strptime(time, date_format), timezone=timezone)`

dwyeradam: I also believe the correct format for time zones would be (GMT) or (GMT+1). In other words surround the whole timezone by the bracket not just the +1.

dwyeradam (changes requested):

tests/app/test_jinja_filters.py
```diff
@@ -129,6 +129,28 @@ def test_format_date_month_year(self):
...
+        date_time = '2018-03-29T11:59:13.528680'
+        date_format = '%d %B %Y at %H:%M'
+
+        # When
+        format_value = format_datetime(date_time, date_format)
+
+        # Then
+        self.assertEqual(format_value, "<span class='date'>29 March 2018 at 12:59 (GMT+1)</span>")
+
+    def test_format_date_time_in_non_dst(self):
```
dwyeradam: Sorry petty but:
If it wasn't for the fact i'm leaving another comment then I would have let this go but I don't like the use of "non_bst". This is GMT. 
`test_format_datetime_gmt`.
Equally in the method above it's not really dst it's BST or GMT+1
`test_format_datetime_bst`
We're setting London time then using American terminology. UTC and non UTC would also have worked in reverse.

app/jinja_filters.py
```diff
@@ -82,7 +83,14 @@ def format_month_year_date(value, date_format='%B %Y'):
...
 def format_datetime(value, date_format='%d %B %Y at %H:%M'):
-    return "<span class='date'>{date}</span>".format(date=datetime.strptime(value, '%Y-%m-%dT%H:%M:%S.%f').strftime(date_format))
+    date = datetime.strptime(value, '%Y-%m-%dT%H:%M:%S.%f')
+    tz_time = pytz.timezone('UTC').localize(date)
+    london_tz = pytz.timezone('Europe/London')
+    london_time = tz_time.astimezone(london_tz)
+    time = london_time.strftime(date_format)
+    timezone = '(GMT+1)' if datetime.dst(london_time) == timedelta(hours=1) else '(GMT)'
+    return "<span class='date'>{date} {timezone}</span>".format(date=datetime.strptime(time, date_format).
+                                                                strftime(date_format), timezone=timezone)
```
dwyeradam: `time` is already in the string format you need so why are we converting it back to a date and then converting it back into a string. Just pass `time` in here but rename it `date_time` as a truer reflection of what it is.

dwyeradam (changes requested):

dwyeradam (approved):

app/jinja_filters.py
```diff
@@ -82,7 +83,13 @@ def format_month_year_date(value, date_format='%B %Y'):
 
 @blueprint.app_template_filter()
 def format_datetime(value, date_format='%d %B %Y at %H:%M'):
-    return "<span class='date'>{date}</span>".format(date=datetime.strptime(value, '%Y-%m-%dT%H:%M:%S.%f').strftime(date_format))
+    date = datetime.strptime(value, '%Y-%m-%dT%H:%M:%S.%f')
+    tz_time = pytz.timezone('UTC').localize(date)
+    london_tz = pytz.timezone('Europe/London')
+    london_time = tz_time.astimezone(london_tz)
+    date_time = london_time.strftime(date_format)
+    timezone = '(GMT+1)' if datetime.dst(london_time) == timedelta(hours=1) else '(GMT)'
```
nealedj: Could be simplified to:

```
timezone = '(GMT+1)' if london_time.dst() else '(GMT)'
```

app/jinja_filters.py
```diff
@@ -82,7 +83,13 @@ def format_month_year_date(value, date_format='%B %Y'):
 
 @blueprint.app_template_filter()
 def format_datetime(value, date_format='%d %B %Y at %H:%M'):
-    return "<span class='date'>{date}</span>".format(date=datetime.strptime(value, '%Y-%m-%dT%H:%M:%S.%f').strftime(date_format))
+    date = datetime.strptime(value, '%Y-%m-%dT%H:%M:%S.%f')
```
nealedj: Can parse the datetime as an aware datetime object like this:

```
tz_time = datetime.strptime(value, '%Y-%m-%dT%H:%M:%S.%f').replace(tzinfo=datetime.timezone.utc)
```

nealedj (changes requested):

nealedj (approved):

shahi645 (approved):

PR #1530, opened 2018-03-29
Uncomplete blocks dependent on an answer that changes

dwyeradam:
If I answer the 1st question as 100 and then the 2nd page questions with appropriate answers, if i go back and edit the 1st question to 1 and jump back to the summary page.
I am able to submit the answers in an invalid state.
I think if we modify an answer that is depended on by validation then we should invalidate the answers for the dependent block.

### DIFF
--- a/app/questionnaire/questionnaire_schema.py
+++ b/app/questionnaire/questionnaire_schema.py
@@ -1,5 +1,5 @@
-import collections
 import itertools
+from collections import OrderedDict, defaultdict
 from app.validation.error_messages import error_messages
 
 
@@ -42,6 +42,12 @@ class QuestionnaireSchema(object):  # pylint: disable=too-many-public-methods
 
     def get_answer(self, answer_id):
         return self._answers_by_id.get(answer_id)
+
+    def get_answer_dependencies_by_id(self, answer_id):
+        return self._answer_dependencies.get(answer_id, [])
+
+    def get_group_and_block_id_by_answer_id(self, answer_id):
+        return self._answer_with_block_and_group_ids.get(answer_id)
 
     def get_groups_that_repeat_with_answer_id(self, answer_id):
         for group in self.groups:
@@ -138,9 +144,11 @@ class QuestionnaireSchema(object):  # pylint: disable=too-many-public-methods
         self._answers_by_id = get_nested_schema_objects(self._questions_by_id, 'answers')
         self.error_messages = self._get_error_messages()
         self.aliases = self._get_aliases()
+        self._answer_dependencies = self._get_answer_dependencies()
+        self._answer_with_block_and_group_ids = self._get_answer_block_group_ids()
 
     def _get_sections_by_id(self):
-        return collections.OrderedDict(
+        return OrderedDict(
             (section['id'], section)
             for section in self.json.get('sections', [])
         )
@@ -167,6 +175,42 @@ class QuestionnaireSchema(object):  # pylint: disable=too-many-public-methods
 
         return aliases
 
+    def _get_answer_dependencies(self):
+        dependencies = defaultdict(list)
+        # Answer level dependencies
+        for answer in self.answers:
+            dependency_id = answer['id']
+            if 'min_value' in answer and 'answer_id' in answer['min_value']:
+                answer_id = answer['min_value']['answer_id']
+                dependencies[answer_id].append(dependency_id)
+            if 'max_value' in answer and 'answer_id' in answer['max_value']:
+                answer_id = answer['max_value']['answer_id']
+                dependencies[answer_id].append(dependency_id)
+
+        # Question level dependencies
+        for question in self.questions:
+            for calculation in question.get('calculations', []):
+                answer_id = calculation.get('answer_id')
+                if answer_id:
+                    for dependency_id in calculation['answers_to_calculate']:
+                        dependencies[answer_id].append(dependency_id)
+
+        return dependencies
+
+    def _get_answer_block_group_ids(self):
+        answers_blocks_groups = {}
+        for group in self.groups:
+            for block in group['blocks']:
+                block_group = {
+                    'block': block['id'],
+                    'group': group['id']
+                }
+                answer_ids = self.get_answer_ids_for_block(block['id'])
+                for answer_id in answer_ids:
+                    answers_blocks_groups[answer_id] = block_group
+
+        return answers_blocks_groups
+
 
 def get_nested_schema_objects(parent_object, list_key):
     """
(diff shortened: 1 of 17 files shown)


yoganandkunche (approved):

app/questionnaire/questionnaire_schema.py
```diff
@@ -167,6 +175,42 @@ def _get_aliases(self):
 
         return aliases
 
+    def get_answer_dependencies(self):
```
jonnyshaw89: theses should be private `_get_answer_dependencies(self):`

app/questionnaire/questionnaire_schema.py
```diff
@@ -167,6 +175,42 @@ def _get_aliases(self):
...
+        for question in self._questions_by_id.values():
+            for calculation in question.get('calculations', []):
+                answer_id = calculation.get('answer_id')
+                if answer_id:
+                    for dependency_id in calculation['answers_to_calculate']:
+                        dependencies[answer_id] = dependencies.get(answer_id, []) + [dependency_id]
+
+        return dependencies
+
+    def get_answer_block_group_ids(self):
```
jonnyshaw89: theses should be private `_get_answer_block_group_ids(self):`

app/views/questionnaire.py
```diff
@@ -566,12 +566,24 @@ def update_questionnaire_store_with_form_data(questionnaire_store, location, ans
                 _remove_answer_from_questionnaire_store(answer_id, questionnaire_store)
 
             if answer:
+                latest_answer_store_hash = questionnaire_store.answer_store.get_hash()
                 questionnaire_store.answer_store.add_or_update(answer)
+                _remove_dependent_answers_from_completed_blocks(answer_id, questionnaire_store, latest_answer_store_hash)
```
jonnyshaw89: i think that `if latest_answer_store_hash != questionnaire_store.answer_store.get_hash():` should wrap this function call.
otherwise it hides that is it not always run

jonnyshaw89 (dismissed):

app/views/questionnaire.py
dwyeradam: I did that originally but it would fail lint as too complex. So I cheated.

jonnyshaw89: tut tut

dwyeradam: when are you going?

dwyeradam:
@nealedj Could you review this please and if you're happy I've addressed his comments dismiss Jon's review.

app/questionnaire/questionnaire_schema.py
```diff
@@ -167,6 +175,42 @@ def _get_aliases(self):
 
         return aliases
 
+    def _get_answer_dependencies(self):
+        dependencies = {}
+        # Answer level dependencies
+        for answer in self._answers_by_id.values():
```
nealedj: Might be cleaner to make use of the `answers` property.

```
for answer in self.answers:
    ...
```

app/questionnaire/questionnaire_schema.py
```diff
@@ -167,6 +175,42 @@ def _get_aliases(self):
...
+            dependency_id = answer['id']
+            if 'min_value' in answer and 'answer_id' in answer['min_value']:
+                answer_id = answer['min_value']['answer_id']
+                dependencies[answer_id] = dependencies.get(answer_id, []) + [dependency_id]
+            if 'max_value' in answer and 'answer_id' in answer['max_value']:
+                answer_id = answer['max_value']['answer_id']
+                dependencies[answer_id] = dependencies.get(answer_id, []) + [dependency_id]
+
+        # Question level dependencies
+        for question in self._questions_by_id.values():
```
nealedj: Same here.

```
for question in self.questions:
    ...
```

app/questionnaire/questionnaire_schema.py
```diff
@@ -167,6 +175,42 @@ def _get_aliases(self):
...
         return aliases
 
+    def _get_answer_dependencies(self):
+        dependencies = {}
+        # Answer level dependencies
+        for answer in self._answers_by_id.values():
+            dependency_id = answer['id']
+            if 'min_value' in answer and 'answer_id' in answer['min_value']:
+                answer_id = answer['min_value']['answer_id']
+                dependencies[answer_id] = dependencies.get(answer_id, []) + [dependency_id]
```
nealedj: Consider using a `defaultdict(list)` for this instead.

```
dependencies = defaultdict(list)
if 'min_value' in answer and 'answer_id' in answer['min_value']:
    answer_id = answer['min_value']['answer_id']
    dependencies[answer_id].append(dependency_id)
```

app/views/questionnaire.py
```diff
@@ -542,36 +542,48 @@ def update_questionnaire_store_with_form_data(questionnaire_store, location, ans
...
+    if not answer_value_empty(answer_value):
+        # Dates are comprised of 3 string values
+        formatted_answer_value = _format_answer_value(answer_value)
+        if formatted_answer_value:
+            return formatted_answer_value
+
+    return None
+
+
+def _remove_dependent_answers_from_completed_blocks(answer_id, questionnaire_store):
```
nealedj, line 577: This removes the blocks containing dependent answers but doesn't remove the answers. Is that expected?

Could we add a docstring to this function to make it clear what the intention is?

app/questionnaire/questionnaire_schema.py
nealedj: I don't think this solution currently identifies dependent questions which are dependent due to piped values. Is that in scope for this PR?

nealedj (changes requested):

app/views/questionnaire.py
dwyeradam: We wouldn't want to remove their answers as they could still be correct. This just means that block has to be revisited and therefore validated.

app/questionnaire/questionnaire_schema.py
dwyeradam: Jon and I did discuss this quite a bit but maybe we need to discuss as a whole. Piping alone would not result in any new validation being done on dependent answers and so forcing users to revisit the block would have far less benefit. We also felt that if an answer did affect another then that answer should be in a max/min type check.

app/questionnaire/questionnaire_schema.py
dwyeradam: Great thanks. I was trying to find an easy way to initialise the key on first use and this is a far better way.

app/views/questionnaire.py
dwyeradam: I'll add the docstring though.

dwyeradam:
@nealedj Ready for re-review.

nealedj (approved):
Looks good!

nealedj:
@dwyeradam LGTM - ready to squash

PR #1531, opened 2018-04-03
Eq 1945 social alpha

davecharles:
Author a first cut survey schema for Labour Force Survey.  This PR adds a new schema for LFS to `data/en`, there are no code changes.

Only the first path of the LFM design has been implemented (Socio Demographic), the remainder (Labour Market) will be implemented in a subsequent iteration.  Issues and omissions have been documented on the related [Trello card](https://trello.com/c/eBUSUODf).

### How to review 
- Check schema is valid using the schema validator i.e:
   `pipenv run ./scripts/test_schemas.sh data/en/labour_force.json`
- Run the survey [with this url](https://preprod-surveys-launch.eq.ons.digital/quick-launch?url=https://raw.githubusercontent.com/ONSdigital/eq-survey-runner/eq-1945-social-alpha/data/en/labour_force.json).

### DIFF
--- a/data/en/labour_force.json
+++ b/data/en/labour_force.json
@@ -1,0 +1,2881 @@
+{
+    "mime_type": "application/json/ons/eq",
+    "schema_version": "0.0.1",
+    "data_version": "0.0.2",
+    "survey_id": "lfs ",
+    "title": "Labour Force Survey (Alpha)",
+    "description": "Labour Force Survey Schema (Alpha)",
+    "theme": "default",
+    "legal_basis": "Voluntary",
+    "eq_id": "1",
+    "form_type": "1",
+    "navigation": {
+        "visible": true
+    },
+    "sections": [{
+            "id": "household-section",
+            "title": "About the household",
+            "groups": [{
+                    "id": "about-household-group",
+                    "title": "About the household",
+                    "blocks": [{
+                            "type": "Interstitial",
+                            "id": "about-household",
+                            "title": "About you and people who live at {{ respondent.ru_name }}",
+                            "description": "In this section, we’re going to ask about you, and the people that live with you.",
+                            "content": [{
+                                "title": "Information you need",
+                                "list": [
+                                    "Names of the people living in your household.",
+                                    "Their gender.",
+                                    "How they are related to you and one another."
+                                ]
+                            }]
+                        },
+                        {
+                            "id": "household-composition",
+                            "type": "Question",
+                            "questions": [{
+                                "id": "household-composition-question",
+                                "title": "What are the names of everyone who lives in the <em>{{ respondent.ru_name }}</em> household?",
+                                "guidance": {
+                                    "content": [{
+                                        "title": "Include",
+                                        "list": [
+                                            "Yourself",
+                                            "Students and, or school children who live away from home during term time",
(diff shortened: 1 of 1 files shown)


dwyeradam:
A `yarn format` should fix your linting issue.

data/en/labour_force.json
```diff
@@ -0,0 +1,3201 @@
...
+                                    "mandatory": false,
+                                    "type": "MonthYearDate"
+                                },
+                                {
+                                    "id": "arrive-in-uk-alt-answer",
+                                    "alias": "arrived_alt",
+                                    "mandatory": false,
+                                    "type": "Radio",
+                                    "options": [{
+                                            "label": "Don&apos;t know",
```
MIslam92: Should we be allowing the respondent to give a date then answer "I don't know"/"Prefer not to say " at this point? I know these are essentially two different answers and the mutually exclusive answers might overcome this.

data/en/labour_force.json
```diff
@@ -0,0 +1,3201 @@
...
+                    "hide_in_navigation": true,
+                    "routing_rules": [{
+                        "repeat": {
+                            "type": "answer_count_minus_one",
+                            "answer_id": "first-name"
+                        }
+                    }],
+                    "blocks": [{
+                        "type": "Question",
+                        "id": "household-relationships",
```
MIslam92, line 142: Should this not route to asking the relationship question when the household members are equal to two as well? It takes this into consideration in the JSON but doesn't match what the docs are saying in the card?

MIslam92 (changes requested):
Just some observations in the comments

data/en/labour_force.json
davecharles: Agreed with @charlotte-hirst to remove the DK/PNTS radios for now.

data/en/labour_force.json
davecharles: The routing design is based on a separate "first page" to get info on first two household occupants and their relationship.  We couldn't do this with the existing household-composition mechanics from census so @charlotte-hirst agreed we could duck this in the alpha.

MebinAbraham (commented):
Shouldn't this section be selected in the navigation?
image

MebinAbraham (changes requested):
If I click `Continue` on the proxy check question. I get the following.
image

MebinAbraham (changes requested):
Ignore this. :point_down:

shahi645:
@MebinAbraham The above is because there is piping in the answer value

davecharles:
I can get around this by rephrasing the option labels but it seems to be an issue with functionality (i.e. this should work). @shahi645 sounds like this is a known issue?

MebinAbraham (approved):
Need to raise a card for the issue above, but apart from that, looks good.

nealedj:
Note that https://github.com/ONSdigital/eq-survey-runner/pull/1533 changes the way that values are referenced when piping. If it gets merged first then this PR will need updating -> or vice versa.

davecharles:
@MebinAbraham raised #1538

MIslam92 (approved):

shahi645:
@davecharles Rebase and Squash and we can merge this one?

PR #1533, opened 2018-04-06
Fix for piped answers in repeating groups

nealedj:
This PR is to fix an issue where answers within repeating groups could not be referenced within piped strings. As a side effect of this fix I have also removed the answer alias functionality which provided little value other than to allow piped answer values to be referenced with underscored attribute names.

### How to review 
- Run test_repeating_household schema and assert that the new confirmation block displays the correct answers that we added within the repeating group
- Check all other schemas that used piped values to be sure that they still operate correctly

PRs for publisher and validator also incoming.


nealedj:
When this is going to be merged we need to also update the [census schemas](github.com/ONSdigital/eq-census-schemas).

app/questionnaire/questionnaire_schema.py
```diff
@@ -217,11 +215,12 @@ def get_nested_schema_objects(parent_object, list_key):
     :param parent_object: dict containing a list
     :param list_key: key of the nested list to extract
     """
```
shahi645, line 241: Probably needing some better docstrings here

app/templating/schema_context.py
```diff
@@ -14,33 +15,33 @@ def build_schema_context(metadata, aliases, answer_store, answer_ids_on_path, gr
...
 
-        for answer in answers:
-            if answer['answer_id'] in answer_ids_on_path:
-                safe_answer = json_and_html_safe(answer['value'])
-                matching_answers.append(safe_answer)
+        matching_answers = answer_store.filter(
+            answer_ids=[answer_id], limit=True)
 
-        number_of_matches = len(matching_answers)
+        value = ''
```
shahi645: Would this be better served as an empty list?

shahi645 (changes requested):
Just one or two minor things, but LGTM in general. And also a good deficit or 11 lines of code 😉

app/templating/schema_context.py
nealedj: The default state for the value is an empty string. It's only if we're in a repeating group or it's a repeating answer that the value should ever be a list.

app/questionnaire/questionnaire_schema.py
nealedj: Good call - I added some more words.

shahi645:
screen shot 2018-04-12 at 10 53 18

Getting a 5xx when navigating to summary on MWSS

app/templating/schema_context.py
```diff
@@ -14,33 +15,33 @@ def build_schema_context(metadata, aliases, answer_store, answer_ids_on_path, gr
...
+        if is_repeating_answer or answer_is_in_repeating_group:
+            value = [
```
shahi645: @nealedj Is this why we are getting the 5xx's? should these be normal parentheses

shahi645: On second thoughts i have had a little mess around and it is definitely not this

nealedj: The 500s were caused by an incorrect answer ID being piped in the MWSS schema - I'd missed one when I changed them all.

shahi645 (dismissed):
Still getting 5xx's as the jinja filter is getting an undefined object instead of a value
screen shot 2018-04-13 at 09 42 43

nealedj:
Thanks @shahi645 - good catch! Have fixed that issue in https://github.com/ONSdigital/eq-survey-runner/pull/1533/commits/e4e41e03f11fe3481b725cd18243e7f6b69f3abc

app/questionnaire/questionnaire_schema.py
```diff
@@ -232,14 +230,20 @@ def _get_block_section_group_ids(self):
...
-    )
+    nested_objects = OrderedDict()
+
+    for parent_id, child_object in parent_object.items():
+        for child_list_object in child_object.get(list_key, []):
+            # patch the ID of the parent onto the object
+            child_list_object['parent_id'] = parent_id
+            nested_objects[child_list_object['id']] = child_list_object
+
+    return nested_objects
```
dwyeradam, line 250: Nice addition. With the `parent_id` attribute `self._answer_with_block_and_group_id` and `self._block_with_group_and_section_id` are probably no longer needed as you can more easily track back up the hierarchy now. An argument here that the Boy Scout Rule should apply.

data/en/test_multiple_piping.json
```diff
@@ -117,14 +107,14 @@
...
-                        "title": "Does <em>{{[answers.first_text, answers.second_text] | format_household_name}}</em> live at <em>{{answers.address_line_1}}</em>",
+                        "title": "Does <em>{{[answers['first-text'], answers['second-text']] | format_household_name}}</em> live at <em>{{answers['address-line-1']}}</em>",
                         "type": "General",
                         "answers": [{
                             "id": "multiple-piping-answer",
                             "type": "Radio",
                             "mandatory": false,
                             "options": [{
-                                    "label": "{{[answers.address_building, answers.address_line_1, answers.address_line_2, answers.address_line_3, answers.address_town_city, answers.address_county, answers.address_postcode, answers.address_country] | concatenated_list}}",
+                                    "label": "{{[answers.building, answers['address-line-1'], answers['address-line-2'], answers['address-line-3'], answers['town-city'], answers['county'], answers['postcode'], answers['country']] | concatenated_list}}",
```
dwyeradam: This PR is only changing piping where the answer id is hyphenated but do we want these two standards or try and follow the same format of answers['answerID'] regardless of hyphenation? If we had a single rule it would make it easier if we ever want to search a schema for the use of answers in piping as we'd only ever have to search the one format.

dwyeradam (dismissed):
Considerations to be discussed rather than blockers. Also worth sorting out the conflicts before much more review time is spent.

data/en/test_multiple_piping.json
nealedj: Hmm, possibly. Although I believe there's already another card being worked on by @MebinAbraham that will take both formats into account.

MebinAbraham: @nealedj @dwyeradam So my change can work whether we use dot notation or not, but in my opinion, I think it would be nice if we just keep it consistent with 1 standard. Since dot notation doesn't allow hyphens and the other one does, why not change I guess?

Apart from the cost of changing these, does dot notation provide anything the other one can't?

dwyeradam: Personally I like to chose a style and stick to it. We know `answers['answer-id']` will work for anything and so that should be our standard. As I said, not a blocker by any measure but simplifies if we say we define answer piping as this. Rather than we define answer piping as this except in this situation where it could also be this but that will not work for this.

nealedj: I agree - let's standardise. I'll update.

data/en/census_household.json
```diff
@@ -3636,7 +3631,7 @@
                         "id": "employers-business",
                         "questions": [{
                             "id": "employers-business-question",
-                            "title": "What is the main activity of <em>{{answers.company_name}}</em>?",
+                            "title": "What is the main activity of <em>{{answers['business-name-answer']}}</em>?",
                             "description": "<p>Enter the main activity of your employer or business.</p><p>Another way of describing main activity is the industry that your employer or business works in.</p><p>If you are not sure of the main activity of your employer or business please give the best information you can.</p><p>If you are employed through an agency please enter the main activity of the business you are working for, not the agency.</p><p>For example, <b>primary education, repairing cars, contract catering, computer servicing</b></p>",
```
AndrewTorrance: Does not Render Correctly . Shows 
What is the main activity of [Markup('SomeCompanyNameIAdded')]?

data/en/census_household.json
```diff
@@ -3636,7 +3631,7 @@
                         "id": "employers-business",
                         "questions": [{
                             "id": "employers-business-question",
-                            "title": "What is the main activity of <em>{{answers.company_name}}</em>?",
+                            "title": "What is the main activity of <em>{{answers['business-name-answer']}}</em>?",
```
AndrewTorrance: Also For repeating groups it seems to retain a reference to the first answer. In this case A spouse seemed to have picked up a partners details 
What is the main activity of [Markup('SomeCompanyNameIAdded'), Markup('McDonalds')]?

AndrewTorrance: Child also seems to be retaining parent info ? What is the main activity of [Markup('SomeCompanyNameIAdded'), Markup('McDonalds'), Markup('aldi')]?

AndrewTorrance (dismissed):
Could not test Labour force changes via go launch survey . Census household issues mentioned inline

data/en/census_household.json
nealedj: Good catch! This required some changes to the implementation.

data/en/census_household.json
nealedj: Two issues here:

- the census schema wasn't defining which group instance to get the business name from
- the updated schema context builder wasn't forcing the instance values to be put in the correct position in their list. This wouldn't have been a problem if completing the survey linearly but if you jumped around it would fail

app/questionnaire/questionnaire_schema.py
nealedj: Done

nealedj:
@shahi645 could you re-review please?

AndrewTorrance (approved):

app/templating/schema_context.py
```diff
@@ -66,3 +69,49 @@ def json_and_html_safe(data):
...
+def _get_is_repeating_answer(answer_id):
+    return g.schema.is_repeating_answer_type(answer_id)
+
+
+def _get_answer_is_in_repeating_group(answer_id):
+    return g.schema.answer_is_in_repeating_group(answer_id)
+
+
+def _create_answers_list(answers, index_key):
+    """Creates a list of repeating answer values in the same index position
```
MIslam92, line 83: Minor nitpick but I'm just looking at how in Runner we use docstrings and I think structure wise it should be along these lines for consistency sake: 
https://github.com/ONSdigital/eq-survey-runner/blob/15ff9877b96a06681a8b263a6ea629b88f265579/app/helpers/form_helper.py#L16

MIslam92 (changes requested):

app/templating/schema_context.py
```diff
@@ -66,3 +69,49 @@ def json_and_html_safe(data):
...
+    as that specified by their `group_instance` or `answer_instance`
+
+    Args:
+    :param answers: list of answer dicts
+    :param index_key: key of index value. 'group_instance' or 'answer_instance'
+
+    Returns:
+    List of values. Empty values are padded.
```
MIslam92: Minor comment again your returning a list of values yet called the list 'items'. Would it not be better to call it 'values' instead for semantic purposes?

MIslam92 (changes requested):

app/templating/schema_context.py
nealedj: 👍 have updated

MIslam92 (approved):
"Squishy Squashy"

nealedj:
The following PRs should all be merged together:

https://github.com/ONSdigital/eq-census-schemas/pull/2
https://github.com/ONSdigital/eq-publisher/pull/58
https://github.com/ONSdigital/eq-schema-validator/pull/43

MIslam92 (changes requested):
Sorry! Can't merge this as the commit message is abit to long and not meaningful.
Heres the standard we follow: https://chris.beams.io/posts/git-commit/
For some examples just skim down our commits to find some in master.

MIslam92 (approved):

PR #1534, opened 2018-04-06
Superseded by  1535.- Employees per period mandate to non zero and change error message 

AndrewTorrance:
Superseded by  1535.

MWSS - (1_005)

Would allow a pay period to have zero employees. Added a min_value of 1 to the schema so that if a pay period is selected it must have non zero employees .
If no entry or zero pounds chosen for the pay period the question in the confirmation page always referred to the last week of the period. This was changed to match the slides for monthly, 4 weekly and 5 weekly periods
How to review
Run an mwss survey . And validate that zero employees is not valid and that the confirmation question is correct .

### DIFF
--- a/data/en/1_0005.json
+++ b/data/en/1_0005.json
@@ -302,7 +302,10 @@
                                 "label": "Total number of weekly paid employees",
                                 "mandatory": true,
                                 "q_code": "40",
-                                "type": "Number"
+                                "type": "Number",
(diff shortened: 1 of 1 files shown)


MIslam92 (changes requested):
This commit message needs changing as it does not provide any context to what work has been done and just is not in line with eq standards

PR #1537, opened 2018-04-09
Go to relevant Summary if available after edit

dwyeradam:
Upon returning to a question in any fashion (via edit link, using back buttons or entering question url) when submitting that question if a Summary page (Final or Section) is still available (no change in routing or dependencies) for it then it should return to the summary page.

### How to review 
Use existing test_section_summary.json for testing of returns to both section and final summaries. Final summary can be made unavailable with a change of routing.
Use test_dependencies schemas to show behaviour when summary is unavailable due to a dependency change.

### DIFF
--- a/app/questionnaire/path_finder.py
+++ b/app/questionnaire/path_finder.py
@@ -1,6 +1,5 @@
 import copy
 from structlog import get_logger
-
 from app.questionnaire.location import Location
 from app.questionnaire.rules import evaluate_goto, evaluate_repeat, evaluate_skip_conditions, is_goto_rule
 
@@ -9,10 +8,11 @@ logger = get_logger()
 
 class PathFinder:
 
-    def __init__(self, schema, answer_store, metadata):
+    def __init__(self, schema, answer_store, metadata, completed_blocks):
         self.answer_store = answer_store
         self.metadata = metadata
         self.schema = schema
+        self.completed_blocks = completed_blocks
         self._answer_store_hash = self.answer_store.get_hash()
         self._full_routing_path = None
 
@@ -190,6 +190,7 @@ class PathFinder:
     def get_next_location(self, current_location):
         """
         Returns the next 'location' to visit given a set of user answers
+        If Summary or SectionSummary is available then next location will be those.
         :param current_location:
         :return: The next location as a dict
         """
@@ -198,7 +199,46 @@ class PathFinder:
         current_location_index = PathFinder._get_current_location_index(routing_path, current_location)
 
         if current_location_index is not None and current_location_index < len(routing_path) - 1:
+            if self._is_survey_completed(routing_path):
+                return routing_path[-1]     # Go to Summary location
+
+            section_summary_location = self._get_valid_section_summary(current_location.block_id, routing_path)
+            if section_summary_location:
+                return section_summary_location     # Go to SectionSummary location
+
             return routing_path[current_location_index + 1]
+
+    def _is_survey_completed(self, routing_path):
+        # Check all blocks on routing path are complete
+        for location in routing_path[:-1]:  # Don't evaluate end block (Summary or Confirmation)
+            if location not in self.completed_blocks:
+                return False
+
+        return True
+
+    def _get_valid_section_summary(self, current_block_id, routing_path):
+        """
+        Check only the section you are on and that all blocks in that section are complete
+        If the section has a summary section then pass that back as the next location
+        The SectionSummary block itself does not need to be completed
+        :param current_block_id: Id of block you're routing from
+        :param routing_path: routing path
+        :return: location or None
+        """
+        current_section = self.schema.get_section_and_group_id_by_block_id(current_block_id)['section']
+
+        if self.schema.get_block(current_block_id)['type'] in ['Summary', 'SectionSummary']:
+            return None
+
+        for location in routing_path:
+            location_section = self.schema.get_section_and_group_id_by_block_id(location.block_id)['section']
+            if location_section == current_section:
+                block_type = self.schema.get_block(location.block_id)['type']
+                if block_type == 'SectionSummary':
+                    return location
+
+                if location not in self.completed_blocks:
+                    return None
 
     def get_previous_location(self, current_location):
         """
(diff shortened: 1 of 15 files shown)


app/questionnaire/path_finder.py
```diff
@@ -198,8 +200,49 @@ def get_next_location(self, current_location):
...
+            section_summary_location = self._get_valid_section_summary(current_location.block_id, routing_path, completed_blocks)
+            if section_summary_location:
+                return section_summary_location     # Go to SectionSummary location
+
             return routing_path[current_location_index + 1]
 
+    @staticmethod
+    def _is_survey_completed(routing_path, completed_blocks):
+        # Check all blocks on routing path are complete
+        for location in routing_path[:-1]:  # Don't evaluate end block (Summary or Confirmation)
```
shahi645: We are assuming here that all schemas will either finish with summary or confirmation. Just want to check that will always be the case?

app/questionnaire/questionnaire_schema.py
```diff
@@ -47,7 +47,10 @@ def get_answer_dependencies_by_id(self, answer_id):
         return self._answer_dependencies.get(answer_id, [])
 
     def get_group_and_block_id_by_answer_id(self, answer_id):
-        return self._answer_with_block_and_group_ids.get(answer_id)
+        return self._answer_with_block_and_group_id.get(answer_id)
+
+    def get_section_and_group_id_by_block_id(self, block_id):
```
nealedj, line 52: The `piped-answers-in-repeating-groups-fix` branch patches a `parent_id` onto blocks and groups. Once that's merged that may by a better solution for finding the parent section.

I.e. (in `PathFinder.get_next_location`)
```
location_group = self.schema.get_group(current_block_id['parent_id'])
location_section = self.schema.get_section(location_group['parent_id'])
```

If we merge this first then consider altering this in the other branch. @dwyeradam

app/questionnaire/path_finder.py
```diff
@@ -198,8 +200,49 @@ def get_next_location(self, current_location):
         current_location_index = PathFinder._get_current_location_index(routing_path, current_location)
 
         if current_location_index is not None and current_location_index < len(routing_path) - 1:
+            completed_blocks = get_completed_blocks(current_user)
+            if self._is_survey_completed(routing_path, completed_blocks):
+                return routing_path[-1]     # Go to Summary location
```
MIslam92: Looking at how you returned the section_summary_location below and to make it more readable/consistent  would it be worth declaring a variable called summary_location and returning that instead of `routing_path[-1]` plus you could remove the comments as it would be self explanatory from the name alone

app/questionnaire/path_finder.py
nealedj: For every schema currently in the repo this is true. I think we depend on this being the case in other places as well as this.

Let's check that validator enforces this.

nealedj: I just checked - [it does](https://github.com/ONSdigital/eq-schema-validator/blob/master/app/validation/validator.py#L212).

app/questionnaire/path_finder.py
```diff
@@ -198,8 +200,49 @@ def get_next_location(self, current_location):
         current_location_index = PathFinder._get_current_location_index(routing_path, current_location)
 
         if current_location_index is not None and current_location_index < len(routing_path) - 1:
+            completed_blocks = get_completed_blocks(current_user)
```
nealedj: Rather than introducing a dependency on a global (that then needs to be mocked) could we pass `completed_blocks` in the `PathFinder` constructor?

app/questionnaire/path_finder.py
shahi645: Makes sense just wanted to check that we should enforce this

MIslam92 (changes requested):
Minor nit pick?

nealedj (changes requested):

app/questionnaire/path_finder.py
dwyeradam: @nealedj, Yes, or it may be better to actually replace answer_store with questionnaire_store as we can get both completed_blocks and answer_store from that.

app/questionnaire/path_finder.py
dwyeradam: Yes the last block should be a summary or confirmation page. Regardless the last block will never be completed at this point otherwise the questionnaire would already have been submitted.

app/questionnaire/questionnaire_schema.py
dwyeradam: I'll have a look at that then.

app/questionnaire/path_finder.py
dwyeradam: I don't mind doing that but i'm not sure how more readable that is. 
@nealedj, @shahi645  thumbs up for change, thumbs down for not required please.
It would never be the same as the declaration of variable `section_summary_location` as that is set by a method. That method could then return None and if so I wouldn't return `section_summary_location` at all. Here we would always return the summary location so it's just creating a variable to immediately return it.

app/questionnaire/path_finder.py
nealedj: Yeah that's fair - I guess the only disadvantage might be ease of testing. I figure that the `Navigation` object accepts those arguments separately for that reason.

app/questionnaire/path_finder.py
nealedj: I'll say no in this case. Although assigning values to variables in order to help add context is a valid point.

It's a reasonable implicit bit of logic that if the survey is complete then the next location must be the last (final) location.

app/questionnaire/path_finder.py
dwyeradam: Actually I think i'll follow `Navigation's` lead and pass it in. Simpler all round as easier for testing and checking use of those variables across path_finder.

shahi645 (approved):
Good feature we are adding 👍

nealedj (approved):

dwyeradam:
@MIslam92, can you have another look please and approve if happy. Or raise a new comment if unhappy. Despite the fact I have two reviews don't just dismiss yours as you're committed to a review now.

MIslam92 (approved):
👍 "Squishy Squashy"

PR #1539, opened 2018-04-11
Added group breakdown validations

AndrewTorrance:
**What is the context of this PR?**
Currently MWSS pay groups are not validated against the total. So it is possible that the sum of each could exceed the total. 

**Added Functionality**
For Weekly and Fortnightly the sum can be less than or equal to the total. For monthly, 4 weekly and 5 weekly the sum must be less than the total.

**Changes**
MWSS Adding group validation for pay period so that combined values are validated against already entered total.
Added extra validation for group in the schema , also had to change the validation error messages which where incorrect

**How to test**
launch mwss survey. select all pay periods. Enter a total of £100 for each pay period. For the breakdown validate that individual entries greater than the total cause validation errors on the specific field, then validate that with all fields less than the total but the combined value greater than the total that there is the correct group validation error . For weekly and fortnightly validate that no errors if combined value less than OR equal. For Monthly , 4 weekly and 5 weekly validate that no errors if less than , but there are errors on equal or greater summed value compared to total.

Trello Ticket: https://trello.com/c/xBnknDl9


shahi645 (approved):
👍

MIslam92 (approved):
@AndrewTorrance This is ready but you'll need to change the commit message as its not capitalised

PR #1543, opened 2018-04-16
Ability to pass the current date

MIslam92:
Card: https://trello.com/c/dmXOKqFA/2037-pipe-in-current-date

### How to review 
Try calling the filter in a schema and check if the current date is passed to it

### DIFF
--- a/app/jinja_filters.py
+++ b/app/jinja_filters.py
@@ -15,6 +15,8 @@ from app.settings import DEFAULT_LOCALE
 from app.questionnaire.rules import convert_to_datetime
 blueprint = flask.Blueprint('filters', __name__)
 
+DATE_FORMAT = '%-d %B %Y'
+
 
 @blueprint.app_template_filter()
 def format_number(value):
@@ -53,6 +55,10 @@ def format_unit(unit, value=''):
     return units.format_unit(value=value, measurement_unit=unit, length='short', locale=DEFAULT_LOCALE)
 
 
+def as_london_tz(value):
+    return value.replace(tzinfo=tz.gettz('UTC')).astimezone(tz.gettz('Europe/London'))
+
+
 @evalcontextfilter
 @blueprint.app_template_filter()
 def format_multilined_string(context, value):
@@ -67,10 +73,16 @@ def format_multilined_string(context, value):
 
 
 @blueprint.app_template_filter()
+def get_current_date():
+    now = as_london_tz(datetime.utcnow()).strftime(DATE_FORMAT)
+    return "<span class='date'>{date}</span>".format(date=now)
+
+
+@blueprint.app_template_filter()
 def format_date(value):
     date_format = '%B %Y'
     if value and re.match(r'\d{4}-\d{2}-\d{2}', value):
-        date_format = '%-d %B %Y'
+        date_format = DATE_FORMAT
 
     return "<span class='date'>{date}</span>".format(date=convert_to_datetime(value).strftime(date_format))
 
@@ -81,9 +93,8 @@ def format_month_year_date(value, date_format='%B %Y'):
 
 
 @blueprint.app_template_filter()
-def format_datetime(value, date_format='%d %B %Y at %H:%M'):
-    london_date_time = datetime.strptime(value, '%Y-%m-%dT%H:%M:%S.%f').replace(tzinfo=tz.gettz('UTC'))\
-        .astimezone(tz.gettz('Europe/London'))
+def format_datetime(value, date_format=DATE_FORMAT + ' at %H:%M'):
+    london_date_time = as_london_tz(datetime.strptime(value, '%Y-%m-%dT%H:%M:%S.%f'))
     return "<span class='date'>{date}</span>".format(date=london_date_time.strftime(date_format))
 
 
--- a/data/en/test_routing_date_less_than.json
+++ b/data/en/test_routing_date_less_than.json
@@ -23,7 +23,7 @@
                         }],
                         "description": "",
                         "id": "date-questions",
-                        "title": "Enter a date less than today",
+                        "title": "Enter a date less than {{ get_current_date() }}",
                         "type": "General"
                     }],
                     "title": "",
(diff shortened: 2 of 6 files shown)


shahi645 (changes requested):
are we not implementing this somewhere in a test schema?

MIslam92:
@shahi645 We do not at the moment but I don't see the point creating a whole new test schema to just check the filter works? Better just to call it in any schema and check that it is rendered there?

shahi645:
@MIslam92 if we're adding a feature need to implement in a feature schema and put in functional tests

MIslam92:
@shahi645 Yeah I added the test for the filter now but we do not add a test schema as we do not do that now anyway for every Jinja filter so point doing it for this. Easier to just call the filter and check if it works.

shahi645:
It's a new feature so needs a test schema and functional tests, we cant just add something without demonstrating it works

MIslam92:
@shahi645 Doesn't need to be done as we only unit test for filters? Which alone demonstrates there functionality

shahi645:
I disagree, take a look here: 
https://github.com/ONSdigital/eq-survey-runner/pull/1373/files

data/en/test_routing_date_less_than.json
```diff
@@ -23,7 +23,7 @@
                         }],
                         "description": "",
                         "id": "date-questions",
-                        "title": "Enter a date less than today",
+                        "title": "Enter a date less than",
```
shahi645: shouldnt this have now formatted?

shahi645 (dismissed):
We also need to add a functional test

data/en/test_routing_date_less_than.json
dwyeradam: Half a job Sam. Although I struggle to see much benefit of a new functional test. You might as well add a line in the existing functional for this schema though.

tests/functional/spec/features/routing/date.spec.js
```diff
@@ -186,8 +189,10 @@ describe('Feature: Routing on a Date', function () {
           .getUrl().should.eventually.contain(CorrectAnswerPage.pageName);
       });
 
-      it('When I enter a date greater than or equal to today, Then I should be routed to the incorrect page', function () {
+      it('When I enter a date greater than or equal to today, Then I should be routed to the incorrect page @watch', function () {
         return browser
```
MebinAbraham: Don't forget to remove the `@watch` before merging.
And NEEDS a yarn format.

MebinAbraham (approved):

MebinAbraham (approved):

tests/app/test_jinja_filters.py
```diff
@@ -110,13 +112,24 @@ def test_format_multilined_string_auto_escape(self):
 
         self.assertEqual(str(format_value), '<p>&lt;</p>')
 
+    def test_get_current_date(self):
+        # Given
+        date_format = '%d %B %Y'
+        format_value = get_current_date(date_format)
+        current_date = datetime.utcnow().replace(tzinfo=tz.gettz('UTC')).astimezone(tz.gettz('Europe/London')).strftime\
+            (date_format)
```
MebinAbraham: `# When`?

app/jinja_filters.py
```diff
@@ -66,6 +66,12 @@ def format_multilined_string(context, value):
     return result
 
 
+@blueprint.app_template_filter()
+def get_current_date(date_format='%d %B %Y'):
+    now = datetime.utcnow().replace(tzinfo=tz.gettz('UTC')).astimezone(tz.gettz('Europe/London')).strftime(date_format)
```
nealedj: Could we abstract this bit out into a separate method (`as_london_tz()`) and also call it from `format_datetime()`?

`.replace(tzinfo=tz.gettz('UTC')).astimezone(tz.gettz('Europe/London'))`

nealedj (changes requested):
LGTM - just a tiny refactor comment.

app/jinja_filters.py
dwyeradam: Wouldn't `datetime.now().astimezone(tz.gettz('Europe/London'))` be easier. We don't need to get UTC time and convert it. We just need local London time.

shahi645:
LGTM as it is, just got the minor comments by David and Adam, and remove the `@watch`. 
So won't approve just yet

app/jinja_filters.py
```diff
@@ -66,6 +70,12 @@ def format_multilined_string(context, value):
     return result
 
 
+@blueprint.app_template_filter()
+def get_current_date(date_format='%d %B %Y'):
```
AndrewTorrance: %d will give a leading zero for days < 10 ? perhaps %-d may be used (haven't tested this)

nealedj: Good catch.

Looks like that's also incorrect on `format_datetime()`

The correct format is being used in `format_date()`

Please move these formats into variables to help prevent these sort of inconsistencies occurring.

nealedj (changes requested):

app/jinja_filters.py
```diff
@@ -66,6 +70,12 @@ def format_multilined_string(context, value):
     return result
 
 
+@blueprint.app_template_filter()
+def get_current_date(date_format='%-d %B %Y'):
```
nealedj: As discussed - could you also pull these 3 instances of this date format out into a variable?

nealedj (changes requested):

AndrewTorrance (approved):

app/jinja_filters.py
```diff
@@ -15,6 +15,8 @@
 from app.questionnaire.rules import convert_to_datetime
 blueprint = flask.Blueprint('filters', __name__)
 
+common_date_format = '%-d %B %Y'
```
nealedj: By convention, constants such as this would be defined in upper case. It's probably worth just calling it `DATE_FORMAT`

app/jinja_filters.py
```diff
@@ -53,6 +55,10 @@ def format_unit(unit, value=''):
     return units.format_unit(value=value, measurement_unit=unit, length='short', locale=DEFAULT_LOCALE)
 
 
+def as_london_tz(time):
```
nealedj: Let's not call this parameter `time` - it's a datetime. Maybe just call it `value`.

nealedj (changes requested):
Couple more things - sorry!

nealedj (approved):

nealedj:
@MIslam92 okay to squash 👍

PR #1545, opened 2018-04-19
Specify metadata in survey JSON

MebinAbraham:
Refactor how runner makes use of metadata. Check trello [card](https://trello.com/c/LIK77JbM/1909-specify-metadata-in-the-survey-json-l) for specs.
These changes ensure that runner has the required metadata to successfully launch and submit a survey. The idea is that we should fail earlier if we do not have all the required details.

Each schema must now define the metadata it requires in the `metadata` field.
By default all schemas must contain the following metadata:
1. `user_id`
2. `period_id`

The validator [PR](https://github.com/ONSdigital/eq-schema-validator/pull/47) will enforce the above.

The `exercise` and `respondent` metadata are now under one single context var (`metadata`).

This change affects the downstream data as well, therefore, need to ensure that everything is still functioning as expected.

All metadata used within the current schemas has been documented in the `metadata` field.
As we are using bracket notation for piping answers in the schemas, I have also changed the dot notation for metadata piping ie `metadata.ru_name` -> `metadata['ru_name']`.

### How to review 
1. Ensure all tests pass.
2. Check downstream metadata still works correctly.
3. Check all previous references to `exercise.` and `respondent.` is correctly configured to `metadata.`
4. Check that piping is still working as expected.
5. Ensure all piped metadata in schemas are documented in `required_metadata` field.


app/storage/metadata_parser.py
```diff
@@ -8,7 +8,7 @@
 logger = get_logger()
 
 
-def iso_8601_data_parser(iso_8601_string):
+def iso_8601_date_parser(iso_8601_string):
```
nealedj, line 10: Good catch

app/storage/metadata_parser.py
```diff
@@ -20,74 +20,45 @@ def id_generator():
...
-    'case_ref': MetadataField(mandatory=False),
-    'account_service_url': MetadataField(mandatory=False),
-}
-
-
-def parse_metadata(metadata_to_check):
-    parsed = {}
+def parse_metadata(decrypted_token, schema_metadata):
+
+    validators = {
```
nealedj: Consider moving this mapping to a module-level constant

app/storage/metadata_parser.py
```diff
@@ -20,74 +20,45 @@ def id_generator():
...
-                attr_value = metadata_to_check[key]
-                field.validate(attr_value)
-                logger.debug('parsing metadata', key=key, value=attr_value)
-            else:
+        for key, field in mandatory_metadata.items():
+            if key in claims:
+                value = claims[key]
+                validators[field](value)
+                logger.debug('parsing metadata', key=key, value=value)
+            elif key == 'tx_id':
```
nealedj: Is there any need to include `tx_id` in `mandatory_metadata` if it has a totally different code path within this loop?

app/storage/metadata_parser.py
```diff
@@ -20,74 +20,45 @@ def id_generator():
...
         logger.error('unable to parse metadata', exc_info=e)
         raise InvalidTokenException('incorrect data in token')
-    return parsed
 
+    return claims
+
+
+def _validate_mandatory_schema_metadata(metadata, schema_metadata):
+
+    for key in schema_metadata.keys():
```
nealedj: This method is quite difficult to read.

I guess it could be rewritten as:

```
for key in schema_metadata.keys():
    if key == 'trad_as_or_ru_name' and not (
            'trad_as' in metadata or 'ru_name' in metadata):
        raise InvalidTokenException('Missing key/value for trad_as_or_ru_name')

    if not metadata.get(key):
        raise InvalidTokenException('Missing key/value for {}'.format(key))
```

I'm not sold on the above either - open to suggestions.

app/templating/metadata_context.py
```diff
@@ -4,58 +4,37 @@
...
-        'return_by': metadata['return_by'],
-        'start_date': metadata['ref_p_start_date'],
-        'end_date': metadata['ref_p_end_date'],
-        'employment_date': metadata['employment_date'],
-        'region_code': json_and_html_safe(metadata['region_code']) if 'region_code' in metadata else None,
-        'period_str': json_and_html_safe(metadata['period_str']),
         'eq_id': json_and_html_safe(metadata['eq_id']),
-        'collection_id': json_and_html_safe(metadata['collection_exercise_sid']),
         'form_type': json_and_html_safe(metadata['form_type']),
+        'ref_p_start_date': metadata.get('ref_p_start_date'),
```
nealedj, line 13: Is there a reason that we're still hard-coding these keys into runner? Shouldn't we be able to iterate through the keys in the survey `required_metadata` and put those values into the context?

app/templating/schema_context.py
```diff
@@ -43,26 +42,21 @@ def _map_alias_to_answers(aliases, answer_store, answer_ids_on_path):
...
-        'region_code': json_and_html_safe(metadata['region_code']),
-    }
-
-
-def _build_respondent(metadata):
-    return {
-        'ru_name': json_and_html_safe(metadata['ru_name']),
-        'trad_as': json_and_html_safe(metadata['trad_as']),
-        'trad_as_or_ru_name': json_and_html_safe(metadata['trad_as'] or metadata['ru_name']),
+        'ref_p_start_date': metadata.get('ref_p_start_date'),
```
nealedj, line 50: Same here - we should be able to build this up from the `required_metadata` in the schemas.

I don't think we should need to change the code to accept new schema-specific metadata. Ideally we should just be able to add the key to the `required_metadata`.

app/views/session.py
```diff
@@ -52,20 +52,21 @@ def login():
...
 
-    metadata = parse_metadata(decrypted_token)
-    eq_id = metadata['eq_id']
-    form_type = metadata['form_type']
+    eq_id = decrypted_token.get('eq_id')
+    form_type = decrypted_token.get('form_type')
+    language_code = decrypted_token.get('language_code', 'en')
+
+    g.schema = load_schema_from_params(eq_id, form_type, language_code)
+    schema_metadata = g.schema.json['required_metadata']
```
nealedj: Not sure it makes sense to call this `required_metadata`. There isn't any such thing as optional metadata so I'd say it's better to just just call it `metadata`.

app/storage/metadata_parser.py
```diff
@@ -20,74 +20,45 @@ def id_generator():
...
-    'roles': MetadataField(mandatory=False),
-    'case_id': MetadataField(mandatory=False),
-    'case_ref': MetadataField(mandatory=False),
-    'account_service_url': MetadataField(mandatory=False),
-}
-
-
-def parse_metadata(metadata_to_check):
-    parsed = {}
+def parse_metadata(decrypted_token, schema_metadata):
```
nealedj: Could you add a docstring to this function to explain that it's extracting values from the claims that are defined in the schema's metadata?

app/templates/introduction.html
```diff
@@ -25,14 +25,14 @@
             <div class="grid__col col-7@m">
 
                 <h2 class="saturn">You are completing this for
-                    <span>{{ meta.respondent.name}}</span>
-                    {% if meta.respondent.trading_as %}
-                        <span class="trading_as">(trading as&nbsp;<span class="trad_as">{{meta.respondent.trading_as }}</span>)</span>
+                    <span>{{ metadata.name}}</span>
```
nealedj: I think this should be `{{ metadata.ru_name }}` ?

app/templates/partials/topbar.html
```diff
@@ -2,8 +2,8 @@
     <div class="bar__inner container">
         <div class="bar__title venus">{{ survey_title }}
             <span class="mars">
-            {% if meta and meta.respondent.name %}
-                for&nbsp;<span>{{ meta.respondent.name }}</span>
+            {% if meta and metadata.name %}
```
nealedj: Another instance where it should be `ru_name`

nealedj (changes requested):
Looks really good. A few inline comments. The main one is around the metadata keys still being hardcoded into the context rather than being dynamically added based on the schema's metadata definition.

app/storage/metadata_parser.py
MebinAbraham: Well, its required by runner but it may be provided via the token and if not we want to generate it.

Edit: Removed the check in the loop and moved it above to generate if not present.

app/templating/metadata_context.py
MebinAbraham: To be honest, I just wasn't quite sure the purpose of this card was to also decouple runner from these metadata. I thought it was to ensure we have what the required metadata to launch and submit the survey successfully.

But that is definitely a good idea and I'll make the changes soon. Since `required_metadata` only contains the metadata for the schema and not things like `eq_id`, shall I handle this the same way I did for `parse_metadata`?
Also, would it be inefficient to just make use of the `parse_metadata` method again to return all the key/values instead of iterating through again?

app/templating/schema_context.py
MebinAbraham: For the schema_context, I'll just iterate through the `required_meatadata` as there are no eq specific keys here.

app/templates/introduction.html
MebinAbraham: Good catch.

MebinAbraham:
Closing for now as additional functionality has been discovered that were not outlined in the spec.

PR #1556, opened 2018-04-26
Added mbs_0251, 0201 and 0202

AndrewTorrance:
Added support for 3 variants of mbs 

Compare to slides
Note: Was created based on Master moved to mbs-feature after pr raised the only files the original commit had where mbs_0201, mbs_0202 and mbs_0251


AndrewTorrance:
Changes needed due to NI wording

PR #1571, opened 2018-05-02
Specify metadata in survey JSON

MebinAbraham:
Refactor how runner makes use of metadata. Check trello [card](https://trello.com/c/LIK77JbM/1909-specify-metadata-in-the-survey-json-l) for specs.
These changes ensure that runner has the required metadata to successfully launch and submit a survey. The idea is that we should fail earlier if we do not have all the required details.

Each schema must now define the metadata it requires in the `metadata` field.
By default all schemas must contain the following metadata:
1. `user_id`
2. `period_id`

The validator [PR](https://github.com/ONSdigital/eq-schema-validator/pull/50) will enforce the above.

The `exercise` and `respondent` metadata are now under one single context var (`metadata`).

This change affects the downstream data as well, therefore, need to ensure that everything is still functioning as expected.

All metadata used within the current schemas has been documented in the `metadata` field.
As we are using bracket notation for piping answers in the schemas, I have also changed the dot notation for metadata piping ie `metadata.ru_name` -> `metadata['ru_name']`, both ways are valid.

### How to review 
1. Ensure all tests pass.
2. Check downstream metadata still works correctly.
3. Check all previous references to `exercise.` and `respondent.` is correctly configured to `metadata.`
4. Check that piping is still working as expected.
5. Ensure all piped metadata in schemas are documented in `metadata` field.

### DIFF
--- a/app/authentication/authenticator.py
+++ b/app/authentication/authenticator.py
@@ -3,7 +3,6 @@ from uuid import uuid4
 from blinker import ANY
 from flask import session as cookie_session, current_app
 from flask_login import LoginManager, user_logged_out
-from sdc.crypto.exceptions import InvalidTokenException
 from sdc.crypto.decrypter import decrypt
 from structlog import get_logger
 
@@ -12,7 +11,7 @@ from app.authentication.user import User
 from app.data_model.session_data import SessionData
 from app.globals import get_questionnaire_store, get_session_store, create_session_store
 from app.keys import KEY_PURPOSE_AUTHENTICATION
-from app.storage.metadata_parser import is_valid_metadata
+
 from app.settings import EQ_SESSION_ID, USER_IK
 
 logger = get_logger()
@@ -114,18 +113,14 @@ def store_session(metadata):
 
 
 def decrypt_token(encrypted_token):
-    logger.debug('decrypting token')
-    if not encrypted_token or encrypted_token is None:
+    if not encrypted_token:
         raise NoTokenException('Please provide a token')
 
+    logger.debug('decrypting token')
     decrypted_token = decrypt(token=encrypted_token,
                               key_store=current_app.eq['key_store'],
                               key_purpose=KEY_PURPOSE_AUTHENTICATION,
                               leeway=current_app.config['EQ_JWT_LEEWAY_IN_SECONDS'])
-
-    valid, field = is_valid_metadata(decrypted_token)
-    if not valid:
-        raise InvalidTokenException('Missing value {}'.format(field))
 
     logger.debug('token decrypted')
     return decrypted_token
--- a/app/storage/metadata_parser.py
+++ b/app/storage/metadata_parser.py
@@ -1,6 +1,5 @@
 import uuid
 from datetime import datetime
-
 from structlog import get_logger
 
 from sdc.crypto.exceptions import InvalidTokenException
@@ -8,7 +7,7 @@ from sdc.crypto.exceptions import InvalidTokenException
 logger = get_logger()
 
 
-def iso_8601_data_parser(iso_8601_string):
+def iso_8601_date_parser(iso_8601_string):
     return datetime.strptime(iso_8601_string, '%Y-%m-%d')
 
 
@@ -16,90 +15,95 @@ def uuid_4_parser(plain_string):
     return str(uuid.UUID(plain_string))
 
 
+def clean_leading_trailing_spaces(metadata):
+    for key, value in metadata.items():
+        if isinstance(value, str):
+            metadata[key] = value.strip()
+
+    return metadata
+
+
 def id_generator():
     return str(uuid.uuid4())
 
 
-class MetadataField(object):
-    def __init__(self, mandatory=True, validator=None, generator=None):
-        # the function to convert the value from the jwt into the required data type
-        self._validator = validator
-        # the function to execute if the value should be auto-generated
-        self._generator = generator
-        # flag to indicate if the value must exist in the jwt token
-        self.mandatory = mandatory
+VALIDATORS = {
+    'date': iso_8601_date_parser,
+    'uuid': uuid_4_parser,
+    'string': lambda *args: None,
+    'object': lambda *args: None
+}
 
-    def validate(self, original_value):
-        if self._validator:
-            # The parser methods throw exceptions on incorrect data
-            self._validator(original_value)
-
-    def generate(self):
-        if self._generator:
-            return self._generator()
-
-        return None
-
-
-metadata_fields = {
-    'user_id': MetadataField(),
-    'ru_ref': MetadataField(),
-    'ru_name': MetadataField(),
-    'eq_id': MetadataField(),
-    'collection_exercise_sid': MetadataField(),
-    'period_id': MetadataField(),
-    'period_str': MetadataField(),
-    'ref_p_start_date': MetadataField(validator=iso_8601_data_parser),
-    'ref_p_end_date': MetadataField(mandatory=False, validator=iso_8601_data_parser),
-    'form_type': MetadataField(),
-    'survey_url': MetadataField(mandatory=False),
-    'return_by': MetadataField(validator=iso_8601_data_parser),
-    'trad_as': MetadataField(mandatory=False),
-    'employment_date': MetadataField(mandatory=False, validator=iso_8601_data_parser),
-    'region_code': MetadataField(mandatory=False),
-    'language_code': MetadataField(mandatory=False),
-    'tx_id': MetadataField(mandatory=False, validator=uuid_4_parser, generator=id_generator),
-    'variant_flags': MetadataField(mandatory=False),
-    'roles': MetadataField(mandatory=False),
-    'case_id': MetadataField(mandatory=False),
-    'case_ref': MetadataField(mandatory=False),
-    'account_service_url': MetadataField(mandatory=False),
+MANDATORY_METADATA = {
+    'eq_id': {
+        'validator': 'string'
+    },
+    'form_type': {
+        'validator': 'string'
+    },
+    'ru_ref': {
+        'validator': 'string'
+    },
+    'collection_exercise_sid': {
+        'validator': 'string'
+    },
+    'tx_id': {
+        'validator': 'uuid'
+    }
 }
 
 
-def parse_metadata(metadata_to_check):
-    parsed = {}
+def parse_runner_claims(claims):
+    cleaned_claims = clean_leading_trailing_spaces(claims.copy())
+    cleaned_claims['tx_id'] = claims.get('tx_id', id_generator())  # Generate tx_id if not preset in claims
+    validate_metadata(cleaned_claims, MANDATORY_METADATA)
+
+    return cleaned_claims
+
+
+def validate_metadata(claims, required_metadata):
+    _validate_mandatory_metadata(claims, required_metadata)
+    _validate_metadata_values(claims, required_metadata)
+
+
+def _validate_mandatory_metadata(metadata, required_metadata):
+    """
+    Validate that JWT claims contain the required metadata
+    :param metadata:
+    :param required_metadata:
+    """
+    for key in required_metadata.keys():
+        if key == 'variant_flags':
+            # variant flags is empty by default
+            valid = 'variant_flags' in metadata
+        elif key == 'trad_as_or_ru_name':
+            # either of 'trad_as' or 'ru_name' is required
+            valid = 'trad_as' in metadata or 'ru_name' in metadata
+        else:
+            valid = bool(metadata.get(key))
+
+        if not valid:
+            raise InvalidTokenException('Missing key/value for {}'.format(key))
+
+
+def _validate_metadata_values(claims, validators):
+    """
+    Validate metadata from the JWT claims that are required by the schema/runner.
+    Ensures that the values adhere to the expected format.
+    :param claims:
+    :param validators:
+    """
     try:
-        for key, field in metadata_fields.items():
-            if key in metadata_to_check:
-                attr_value = metadata_to_check[key]
-                field.validate(attr_value)
-                logger.debug('parsing metadata', key=key, value=attr_value)
-            else:
-                logger.debug('generating metadata value', key=key)
-                attr_value = field.generate()
+        for key, field in validators.items():
+            if claims.get(key):
+                value = claims[key]
+                VALIDATORS[field['validator']](value)
+                logger.debug('parsing metadata', key=key, value=value)
 
-            parsed[key] = attr_value
-    except (RuntimeError, ValueError, TypeError) as e:
-        logger.error('unable to parse metadata', exc_info=e)
-        raise InvalidTokenException('incorrect data in token')
-
-    return clean_leading_trailing_spaces(parsed)
-
-
-def is_valid_metadata(metadata):
-    for key, field in metadata_fields.items():
-        if field.mandatory and key not in metadata:
-            return False, key
-    return True, ''
-
-
-def clean_leading_trailing_spaces(metadata):
-    clean_metadata = {}
-    for key, value in metadata.items():
-        if isinstance(value, str):
-            value = value.strip()
-
-        clean_metadata[key] = value
-
-    return clean_metadata
+    except (RuntimeError, ValueError, TypeError) as error:
+        logger.error('unable to parse metadata', exc_info=error)
+        raise InvalidTokenException('incorrect data in token') from error
+    except KeyError as key_error:
+        error_msg = 'Invalid validator for schema metadata - {}'.format(key_error.args[0])
+        logger.error(error_msg, exc_info=key_error)
+        raise KeyError(error_msg) from key_error
--- a/app/templates/view-submission.html
+++ b/app/templates/view-submission.html
@@ -11,8 +11,8 @@
 
 <div class="u-mb-s">
     <div class="panel panel--simple panel--success panel--spacious">
-        <p class="mars">Your answers were submitted for <b>{{meta.respondent.ru_name}}</b> on {{meta.survey.submitted_time|format_datetime}}</p>
-        <p class="mars">Transaction ID: <b>{{meta.respondent.tx_id}}</b></p>
+        <p class="mars">Your answers were submitted for <b>{{metadata.ru_name}}</b> on {{metadata.submitted_time|format_datetime}}</p>
+        <p class="mars">Transaction ID: <b>{{metadata.tx_id}}</b></p>
     </div>
 </div>
 
--- a/app/templating/metadata_context.py
+++ b/app/templating/metadata_context.py
@@ -1,61 +1,38 @@
 from app.libs.utils import convert_tx_id
 from app.templating.schema_context import json_and_html_safe
+from app.templating.schema_context import build_schema_metadata
 
 
 def build_metadata_context(metadata):
     """
-    :param metadata: user metadata
+    Builds the metadata context using eq claims and schema defined metadata
+    :param metadata: token metadata
     :return: metadata context
     """
-    render_data = {
-        'survey': _build_survey_meta(metadata),
-        'respondent': _build_respondent_meta(metadata),
-    }
-    return render_data
-
-
-def _build_respondent_meta(metadata):
-    respondent_id = None
-    name = None
-    trading_as = None
-
-    if metadata:
-        respondent_id = metadata['ru_ref']
-        name = metadata['ru_name']
-        trading_as = metadata['trad_as']
-
-    respondent_meta = {
-        'tx_id': convert_tx_id(metadata['tx_id']),
-        'respondent_id': json_and_html_safe(respondent_id),
-        'name': json_and_html_safe(name),
-        'trading_as': json_and_html_safe(trading_as),
-    }
-    return respondent_meta
-
-
-def _build_survey_meta(metadata):
-    return {
-        'return_by': metadata['return_by'],
-        'start_date': metadata['ref_p_start_date'],
-        'end_date': metadata['ref_p_end_date'],
-        'employment_date': metadata['employment_date'],
-        'region_code': json_and_html_safe(metadata['region_code']) if 'region_code' in metadata else None,
-        'period_str': json_and_html_safe(metadata['period_str']),
+    eq_context = {
         'eq_id': json_and_html_safe(metadata['eq_id']),
         'collection_id': json_and_html_safe(metadata['collection_exercise_sid']),
         'form_type': json_and_html_safe(metadata['form_type']),
+        'ru_ref': json_and_html_safe(metadata['ru_ref']),
+        'tx_id': json_and_html_safe(metadata['tx_id'])
     }
+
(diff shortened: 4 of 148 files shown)


app/storage/metadata_parser.py
```diff
@@ -20,74 +19,69 @@ def id_generator():
...
+    :return: Validated claims
+    """
 
-def is_valid_metadata(metadata):
-    for key, field in metadata_fields.items():
-        if field.mandatory and key not in metadata:
-            return False, key
-    return True, ''
+    required_schema_metadata = _validate_mandatory_schema_metadata(decrypted_token, schema_metadata)
+    MANDATORY_METADATA.update(required_schema_metadata)
```
nealedj: This will update the module attribute which will get added to and won't be reset between requests.

Can fix by just switching round the variable used for checking metadata:

```
...
required_schema_metadata.update(MANDATORY_METADATA)
...
        for key, field in required_schema_metadata.items():
```

app/storage/metadata_parser.py
```diff
@@ -20,74 +19,69 @@ def id_generator():
...
+    try:
+        for key, field in MANDATORY_METADATA.items():
+            if key in claims:
+                value = claims[key]
+                VALIDATORS[field](value)
+                logger.debug('parsing metadata', key=key, value=value)
+
+    except (RuntimeError, ValueError, TypeError) as error:
+        logger.error('unable to parse metadata', exc_info=error)
+        raise InvalidTokenException('incorrect data in token')
```
nealedj: Consider making use of exception chaining here.

```
raise InvalidTokenException('incorrect data in token') from error
```

app/storage/metadata_parser.py
```diff
@@ -20,74 +19,69 @@ def id_generator():
...
+    except KeyError as key_error:
+        error_msg = 'Invalid validator for schema metadata - {}'.format(key_error.args[0])
+        logger.error(error_msg, exc_info=key_error)
+        raise KeyError(error_msg)
```
nealedj: and here

```
raise KeyError(error_msg) from key_error
```

app/storage/metadata_parser.py
```diff
@@ -20,74 +19,69 @@ def id_generator():
...
+    :param metadata:
+    :param schema_metadata:
+    :return:
+    """
+    for key in schema_metadata.keys():
+
+        if key == 'variant_flags':
+            # variant flags is empty by default
+            valid = 'variant_flags' in metadata
+        elif key == 'trad_as_or_ru_name':
```
shahi645: @MebinAbraham Just to confirm, we cant have both `variant_flags` and `ru_name` or `trad_as` together as mandatory metadata?

app/templates/view-submission.html
```diff
@@ -11,8 +11,8 @@ <h1 class="saturn">Submitted answers</h1>
 
 <div class="u-mb-s">
     <div class="panel panel--simple panel--success panel--spacious">
-        <p class="mars">Your answers were submitted for <b>{{meta.respondent.ru_name}}</b> on {{meta.survey.submitted_time|format_datetime}}</p>
-        <p class="mars">Transaction ID: <b>{{meta.respondent.tx_id}}</b></p>
+        <p class="mars">Your answers were submitted for <b>{{metadata.ru_name}}</b> on {{metadata.submitted_time|format_datetime}}</p>
```
shahi645, line 14: Need to edit the contact-us pages, as that contains `ru-ref` is user is authenticated

app/templating/metadata_context.py
```diff
@@ -1,61 +1,37 @@
 from app.libs.utils import convert_tx_id
 from app.templating.schema_context import json_and_html_safe
+from app.templating.schema_context import build_schema_metadata
 
 
 def build_metadata_context(metadata):
     """
-    :param metadata: user metadata
+    :param metadata: token metadata
```
shahi645: Should probably add better docstring here above this param, something like, `eq-context metadata`

scripts/test_schemas.sh
```diff
@@ -1,9 +1,15 @@
 #!/usr/bin/env bash
 
-docker pull onsdigital/eq-schema-validator
-validator="$(docker run -d -p 5001:5000 onsdigital/eq-schema-validator)"
+run_docker=true
+if [ "$1" == "-local" ] || [ "$2" == "-local" ]; then
```
shahi645: Would have been better to keep these changes in a separate PR, just for tracking purposes

shahi645 (changes requested):
Just a couple things when looking over code, currently testing

app/authentication/authenticator.py
```diff
@@ -114,18 +115,20 @@ def store_session(metadata):
...
+    logger.debug('decrypting token')
     decrypted_token = decrypt(token=encrypted_token,
                               key_store=current_app.eq['key_store'],
                               key_purpose=KEY_PURPOSE_AUTHENTICATION,
                               leeway=current_app.config['EQ_JWT_LEEWAY_IN_SECONDS'])
 
-    valid, field = is_valid_metadata(decrypted_token)
-    if not valid:
-        raise InvalidTokenException('Missing value {}'.format(field))
+    for claim in mandatory_claims:
```
nealedj: Consider displaying all missing claims in error:

```
missing_claims = [k for k in mandatory_claims if not decrypted_token.get(k)]
if missing_claims:
    raise InvalidTokenException('Missing mandatory key/value(s) in claims - {}'.format(', '.join(missing_claims)))
```

app/storage/metadata_parser.py
nealedj: No - it's valid to have both. Bear in mind that this is iterating over the `key` so it's just switching the logic for how to validate the metadata based on what key it is. `variant_flags` and `trad_as_or_ru_name` are just separate special cases.

nealedj (changes requested):
Looking really good. A few minor points.

Main point is that eq-publisher needs to have some way to generate this metadata for the schema before we can merge this.

app/templates/view-submission.html
dwyeradam: @shahi645, I might have misunderstood but doesn't the contact-us page use the session data for this and not metadata? And then only displays ruref when session_data is available.

scripts/test_schemas.sh
dwyeradam: He just won't be told Shahi.

app/templates/view-submission.html
MebinAbraham: :point_up: Contact page is using session.

scripts/test_schemas.sh
MebinAbraham: Ill add it to a separate PR. Adam was in on it with me 😆

app/storage/metadata_parser.py
MebinAbraham: Thanks, also now used `.copy()` when resolving `required_schema_metadata` so its not the same reference.

MebinAbraham:
@nealedj @shahi645 Ready for re-review. Thanks

app/storage/metadata_parser.py
```diff
@@ -20,78 +19,75 @@ def id_generator():
...
+
+    except (RuntimeError, ValueError, TypeError) as error:
+        logger.error('unable to parse metadata', exc_info=error)
+        raise InvalidTokenException('incorrect data in token') from error
+    except KeyError as key_error:
+        error_msg = 'Invalid validator for schema metadata - {}'.format(key_error.args[0])
+        logger.error(error_msg, exc_info=key_error)
+        raise KeyError(error_msg) from key_error
+
+    return clean_leading_trailing_spaces(claims)
```
dwyeradam: We should clean the leading/trailing spaces as part of string validation. Otherwise we could end up with an empty string that shouldn't have passed validation if empty.

app/authentication/authenticator.py
```diff
@@ -114,18 +115,20 @@ def store_session(metadata):
...
                               key_store=current_app.eq['key_store'],
                               key_purpose=KEY_PURPOSE_AUTHENTICATION,
                               leeway=current_app.config['EQ_JWT_LEEWAY_IN_SECONDS'])
 
-    valid, field = is_valid_metadata(decrypted_token)
-    if not valid:
-        raise InvalidTokenException('Missing value {}'.format(field))
+    missing_claims = [claim for claim in mandatory_claims if not decrypted_token.get(claim)]
+    if missing_claims:
+        raise InvalidTokenException('Missing mandatory key/value(s) in claims - {}'.format(', '.join(missing_claims)))
```
dwyeradam: Ideally we would have a one stop shop for metadata validation. Here we're checking some hard coded runner type mandatory keys and then later checking specific schema defined keys. Ideally we would have a metadata validation method that checks both kind of defined "mandatory".

app/storage/metadata_parser.py
```diff
@@ -20,78 +19,75 @@ def id_generator():
...
+MANDATORY_METADATA = {
+    'tx_id': {
+        'validator': 'uuid'
+    }
 }
 
 
-def parse_metadata(metadata_to_check):
-    parsed = {}
+def parse_metadata(decrypted_token, schema_metadata):
```
dwyeradam: Should we rename this? It's now more a validator than a parser. As per my other comment this would be the best place for all types of metadata validation.

dwyeradam (changes requested):
Good work. A few points and may be worth @nealedj seeing if he agrees with some/all of them.

app/storage/metadata_parser.py
nealedj: Agree - strip and then validate.

app/storage/metadata_parser.py
nealedj: Good point. It parses dates but only for the purposes of validation, it doesn't return the dates as parsed.

`validate_metadata` may be more descriptive?

app/authentication/authenticator.py
nealedj: I think I'm okay with this as it is. My reasoning is that we're actually conflating two very different types of data in our metadata:

1) Data required for survey runner to function. They're essentially the arguments that are passed to the application and all schemas will always need them. These will always be part of the JWT payload (Long term I think `eq_id` and `form_type` should be replaced by a URI to a schema). Strictly speaking I'd say that this isn't metadata. It's more of a survey manifest.
2) Data required to generate and render a questionnaire. This is data this is being relied upon within the JSON schema but that (ideally) the application code will have no need to have knowledge of. In the future I see these values being returned from an API call from runner. This is actual metadata.

dwyeradam: I agree they are two different reasons for doing the validation but at the same time if I was asked where do we validate the metadata I wouldn't think in the token decryption. I'm just worried this type of validation could get lost if we leave it where it is.
For instance we all missed the fact we have duplicate checking here:
https://github.com/ONSdigital/eq-survey-runner/blob/master/app/authentication/user_id_generator.py#L64
Also we don't check for empty spaces on these if only check is here.

nealedj: Oh yeah. In that case there's no point in validating these values in authenticator.py anyway is there?

dwyeradam: Yes and no. If eq_id and form_type isn't there then we'd fail to load the schema. I've been talking through this with Mebin quite a bit and i'm hoping we've come with a decent compromise where we have split the two types of validation so they're properly defined while also reusing the code.

MebinAbraham:
Making some changes to the `parse_metadata`, will update on Monday as I'm working on the dashboard.

app/templates/view-submission.html
shahi645: Yep my mistake 👍

shahi645:
@MebinAbraham You have a failing unit test, but otherwise, I'm happy with this work! 👍

app/storage/metadata_parser.py
```diff
@@ -1,105 +1,123 @@
...
+
+def metadata_validator(claims, validators=None):
+    """
+    Validate metadata from the JWT claims that are required by the schema/runner.
+    Ensures that the values adhere to the expected format.
+    :param claims:
+    :param validators:
+    :return: Validated claims
+    """
+    if not validators:
```
nealedj: Consider a one-liner.

`validators = validators or MANDATORY_METADATA`

app/storage/metadata_parser.py
```diff
@@ -1,105 +1,123 @@
...
-                logger.debug('parsing metadata', key=key, value=attr_value)
-            else:
-                logger.debug('generating metadata value', key=key)
-                attr_value = field.generate()
 
-            parsed[key] = attr_value
-    except (RuntimeError, ValueError, TypeError) as e:
-        logger.error('unable to parse metadata', exc_info=e)
-        raise InvalidTokenException('incorrect data in token')
+def parse_runner_claims(claims):
```
nealedj, line 56: This function doesn't actually parse anything does it? It just adds the `tx_id` and then calls the validator.

This might be a more appropriate place to call `clean_leading_trailing_spaces`.

app/authentication/authenticator.py
```diff
@@ -114,18 +114,18 @@ def store_session(metadata):
...
+    logger.debug('decrypting token')
     decrypted_token = decrypt(token=encrypted_token,
                               key_store=current_app.eq['key_store'],
                               key_purpose=KEY_PURPOSE_AUTHENTICATION,
                               leeway=current_app.config['EQ_JWT_LEEWAY_IN_SECONDS'])
 
-    valid, field = is_valid_metadata(decrypted_token)
-    if not valid:
-        raise InvalidTokenException('Missing value {}'.format(field))
+    cleaned_claims = clean_leading_trailing_spaces(decrypted_token)
```
nealedj: I don't think the authenticator module should be taking responsibility for cleaning any of the token data. That seems more appropriate for the metadata_parser module.

app/storage/metadata_parser.py
```diff
@@ -1,105 +1,123 @@
...
-def iso_8601_data_parser(iso_8601_string):
+def iso_8601_date_parser(iso_8601_string):
     return datetime.strptime(iso_8601_string, '%Y-%m-%d')
 
 
 def uuid_4_parser(plain_string):
     return str(uuid.UUID(plain_string))
 
 
+def clean_leading_trailing_spaces(metadata):
```
nealedj, line 18: I don't think there's any need to build a new object in this case:

```
for key, value in metadata.items():
    if isinstance(value, str):
        metadata[key] = value.strip()

return metadata
```

nealedj (changes requested):

app/storage/metadata_parser.py
MebinAbraham: I should call it in both this and `validate_schema_metadata` then?. Otherwise its assuming that this one will take care of it? And in terms of tests I would have to call both functions.

app/storage/metadata_parser.py
MebinAbraham: Existing code, also thought it was bad practice to modify passed in arguments?.

MebinAbraham:
@nealedj @dwyeradam @shahi645 Could you re-review when you can please.
@AndrewTorrance @davecharles Could you also review this please even if it is from a functional perspective as it affect author and downstream. So even checking if piping is working as expected etc. Thanks

shahi645 (approved):
LGTM, code is good, functionally works as expected. 👍 👍 👍

app/storage/metadata_parser.py
dwyeradam: I would think you would call or merge the method clean_leading_trailing_spaces(claims.copy()) here instead of passing in a cleaned claims? Simplifies session.py a bit then.

MebinAbraham: Moved it to session for some reason but can't remember the reason so 👍

app/storage/metadata_parser.py
nealedj: It's a judgement call.

Probably a bad idea in most cases but I think it's pretty explanatory here.

app/templating/schema_context.py
```diff
@@ -46,27 +47,27 @@ def _build_answers(answer_store, answer_ids_on_path):
...
-        'start_date': metadata['ref_p_start_date'],
-        'end_date': metadata['ref_p_end_date'],
-        'period_str': json_and_html_safe(metadata['period_str']),
-        'employment_date': metadata['employment_date'],
-        'return_by': metadata['return_by'],
-        'region_code': json_and_html_safe(metadata['region_code']),
-    }
+def build_schema_metadata(metadata):
 
+    schema_metadata = load_schema_from_params(metadata['eq_id'], metadata['form_type']).json['metadata']
```
dwyeradam: You might as well just use `g.schema` rather than `load_schema_from_params(metadata['eq_id'], metadata['form_type'])`.

dwyeradam (changes requested):
Small thing. Looking good though but still checking.

scripts/test_schemas.sh
```diff
@@ -6,8 +6,8 @@ if [ "$1" == "--local" ] || [ "$2" == "--local" ]; then
 fi
 
 if [ "$run_docker" == true ]; then
-    docker pull onsdigital/eq-schema-validator
-    validator="$(docker run -d -p 5001:5000 onsdigital/eq-schema-validator)"
+    docker pull onsdigital/eq-schema-validator:specify-required-metadata-in-schema
+    validator="$(docker run -d -p 5001:5000 onsdigital/eq-schema-validator:specify-required-metadata-in-schema)"
```
dwyeradam: Will need to be put back to latest.

tests/app/forms/test_fields.py
```diff
@@ -15,7 +15,15 @@ class TestFields(unittest.TestCase):
 
     def setUp(self):
         self.answer_store = AnswerStore()
-        self.metadata = parse_metadata({
+        self.schema_metadata = {
```
dwyeradam: You don't actually need `schema_metadata` for this test so it may confuse thing. It might be worth just removing the `parse_metadata` step from `self.metadata`.

dwyeradam (changes requested):

tests/app/forms/test_fields.py
MebinAbraham: Sorted 👍 Ready for more :)

app/views/session.py
```diff
@@ -43,40 +43,45 @@ def login():
...
-    logger.bind(eq_id=eq_id, form_type=form_type, tx_id=tx_id, ru_ref=ru_ref)
-    logger.info('decrypted token and parsed metadata')
+    eq_id = claims.get('eq_id')
+    form_type = claims.get('form_type')
+    language_code = claims.get('language_code', 'en')
 
-    if not eq_id or not form_type:
-        logger.error('missing eq id or form type in jwt')
-        raise NotFound
+    g.schema = load_schema_from_params(eq_id, form_type, language_code)
```
nealedj: This prevents surveys from being loaded by `survey_url`.

```
  File "/Users/nealed/.local/share/virtualenvs/eq-survey-runner-dcmWSl7V/lib/python3.4/site-packages/flask/app.py", line 1612, in full_dispatch_request
    rv = self.dispatch_request()
  File "/Users/nealed/.local/share/virtualenvs/eq-survey-runner-dcmWSl7V/lib/python3.4/site-packages/flask/app.py", line 1598, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/Users/nealed/source/eq-survey-runner/app/views/session.py", line 54, in login
    g.schema = load_schema_from_params(eq_id, form_type, language_code)
  File "/Users/nealed/.local/share/virtualenvs/eq-survey-runner-dcmWSl7V/lib/python3.4/site-packages/flask_caching/__init__.py", line 674, in decorated_function
    rv = f(*args, **kwargs)
  File "/Users/nealed/source/eq-survey-runner/app/utilities/schema.py", line 29, in load_schema_from_params
    return QuestionnaireSchema(_load_schema_file('{}_{}.json'.format(eq_id, form_type), language_code))
  File "/Users/nealed/source/eq-survey-runner/app/utilities/schema.py", line 54, in _load_schema_file
    raise e
  File "/Users/nealed/source/eq-survey-runner/app/utilities/schema.py", line 49, in _load_schema_file
    with open(schema_path, encoding='utf8') as json_data:
FileNotFoundError: [Errno 2] No such file or directory: 'data/en/1_1.json'
```

nealedj (changes requested):

app/storage/metadata_parser.py
```diff
@@ -1,105 +1,119 @@
...
 
-    return clean_metadata
+    try:
+        for key, field in validators.items():
+            if claims.get(key):
+                value = claims[key]
+                VALIDATORS[field['validator']](value)
+                logger.debug('parsing metadata', key=key, value=value)
+            elif key in MANDATORY_METADATA.keys():
+                raise InvalidTokenException('Missing key/value for {}'.format(field))
```
dwyeradam: This shouldn't really be done here. Consider making the method `_validate_mandatory_schema_metadata(metadata, schema_metadata):` more generic and calling it in `parse_runner_claims(claims)`.

Add a test that will throw an exception in your new generic method being called from `parse_runner_claims(claims)`.

dwyeradam (changes requested):

app/storage/metadata_parser.py
MebinAbraham: Done

app/views/session.py
MebinAbraham: @nealedj Should be sorted now. Loads now using metadata which uses `survey_url`

nealedj (approved):
🎉

app/storage/metadata_parser.py
```diff
@@ -1,105 +1,117 @@
...
-        logger.error('unable to parse metadata', exc_info=e)
-        raise InvalidTokenException('incorrect data in token')
+def parse_runner_claims(claims):
 
-    return clean_leading_trailing_spaces(parsed)
+    cleaned_claims = clean_leading_trailing_spaces(claims.copy())
+    cleaned_claims['tx_id'] = claims.get('tx_id', id_generator())  # Generate tx_id if not preset in claims
 
+    _validate_mandatory_schema_metadata(cleaned_claims, MANDATORY_METADATA)
+    metadata_validator(cleaned_claims)
```
dwyeradam: If you just passed `MANDATORY_METADATA` here you wouldn't need line 81 (`validators = validators or MANDATORY_METADATA`) and you can remove the None default.

Then these two lines are the same as the whole of method  `validate_schema_metadata`. It might therefore be worth considering just calling a generic `validate_metadata` and just call it twice, once for each of the two types of metadata that you're validating. 

If you do that it might also be worth considering renaming method `metadata_validator` to `_validate_metadata_values`

```
def validate_metadata(claims, required_metadata):
    _validate_mandatory_metadata(claims, required_metadata)
    _validate_metadata_values(claims, required_metadata)
```

dwyeradam (approved):
I'm happy to approve as the code it fine. I've left a comment to ponder though as we have some minor code duplication that could be cleaned up.

app/storage/metadata_parser.py
MebinAbraham: Could you have a final check please. Thanks

PR #1586, opened 2018-05-08
Allow date validation for MM/YYYY

dwyeradam:
We want to be able to validate a period between 2 month/year dates in a date range.

We also want to be able to say on a single date whether the month/year date to be post/pre a certain date

###
Test schema `test_date_validation_mm_yyyy_combined.json`

### DIFF
--- a/data/en/test_date_validation_mm_yyyy_combined.json
+++ b/data/en/test_date_validation_mm_yyyy_combined.json
@@ -1,0 +1,77 @@
+{
+    "mime_type": "application/json/ons/eq",
+    "schema_version": "0.0.1",
+    "data_version": "0.0.2",
+    "survey_id": "023",
+    "title": "Date formats",
+    "description": "A test schema for different date formats",
+    "theme": "default",
+    "legal_basis": "StatisticsOfTradeAct",
+    "variables": {
+        "period": "{{ format_conditional_date (answers.period_from, exercise.start_date)}} to {{ format_conditional_date (answers.period_to, exercise.end_date)}}"
+    },
+    "sections": [{
+        "id": "default-section",
+        "groups": [{
+            "id": "dates",
+            "title": "Date Range Validation",
+            "blocks": [{
+                    "type": "Question",
+                    "id": "date-range-block",
+                    "title": "Date Range",
+                    "questions": [{
+                        "id": "date-range-question",
+                        "title": "Enter Date Range",
+                        "type": "DateRange",
+                        "guidance": {
+                            "content": [{
+                                "list": [
+                                    "Dates between 2 and 3 months apart",
+                                    "Period from date greater than 1 month before {{exercise.start_date|format_date}}",
+                                    "Period to date no greater than 3 months after {{exercise.end_date|format_date}}"
+                                ]
+                            }]
+                        },
+                        "period_limits": {
+                            "minimum": {
+                                "months": 2
+                            },
+                            "maximum": {
+                                "months": 3
+                            }
+                        },
+                        "answers": [{
+                                "id": "date-range-from",
+                                "label": "Period from",
+                                "mandatory": true,
+                                "type": "MonthYearDate",
+                                "minimum": {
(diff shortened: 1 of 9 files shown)


scripts/test_schemas.sh
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
-docker pull onsdigital/eq-schema-validator
-validator="$(docker run -d -p 5001:5000 onsdigital/eq-schema-validator)"
+docker pull onsdigital/eq-schema-validator:eq-2051-date-validation-mm-yyyy
+validator="$(docker run -d -p 5001:5000 onsdigital/eq-schema-validator:eq-2051-date-validation-mm-yyyy)"
```
dwyeradam: Will need to be put back before merging. Dependent on [schema validator PR](https://github.com/ONSdigital/eq-schema-validator/pull/51)

data/en/test_date_validation_mm_yyyy_combined.json
```diff
@@ -0,0 +1,77 @@
+{
+    "mime_type": "application/json/ons/eq",
+    "schema_version": "0.0.1",
+    "data_version": "0.0.2",
+    "survey_id": "023",
+    "title": "Date formats",
+    "description": "A test schema for different date formats",
+    "theme": "default",
+    "legal_basis": "StatisticsOfTradeAct",
+    "variables": {
```
shahi645, line 10: @dwyeradam this can be removed as part of the metadata work later on

tests/app/forms/test_questionnaire_form.py
```diff
@@ -283,6 +283,99 @@ def test_date_combined_range_too_large_validation(self):
                              schema.error_messages['DATE_PERIOD_TOO_LARGE'] % dict(max='50 days'))
 
 
+    def test_date_mm_yyyy_combined_single_validation(self):
+        with self.test_request_context():
```
shahi645, line 287: I know we have a test for invalid period limits in a combined schema, should we add one for `mm-yyyy` also?

dwyeradam: @shahi645 Do you mean like `test_date_mm_yyyy_combined_range_too_large_validation` and `test_date_mm_yyyy_combined_range_too_large_validation` below?

shahi645: No as in the 5xx when the schema is invalid in the combined case.

shahi645: like `test_invalid_minimum_period_limit_and_single_date_periods`

shahi645:
@dwyeradam quick question about when an `answer_id` is used in the single date validation. Can it be `mm-yyyy` type answer?

tests/app/forms/test_questionnaire_form.py
dwyeradam: Probably for the best. Strictly speaking this hasn't changed the way that works  but lets be sure it hasn't.

dwyeradam:
@shahi645 I think I may write another schema that mixes and matches different date types. They should all work together as when they come to be used they should all have a day component. It already tests dd/mm/yyyy being used for mm/yyyy validation but i'll try the other way around.

shahi645:
Good to approve once extra tests and test schema introduced. Like what you have done with `date_form` also. Much clearer

dwyeradam:
@shahi645, Reviewing the code again I don't really see the benefit of all the extra tests. Now that I have removed the duplication of code within date_form.py early on both date types go through the rules.py method `convert_to_datetime`. This means that from that point on both date types are the same as far as the code goes. Therefore my tests would be pretty much identical.

This also proves the compatibility between using a mix and match range of date types for validation.

shahi645 (approved):
LGTM

data/en/test_date_validation_mm_yyyy_combined.json
dwyeradam: Done

nealedj (approved):

PR #1588, opened 2018-05-09
Add month year formatter to template renderer

davecharles:
LFS requires that month-year date answers are rendered for display.  This change updates and existing the existing (unused) `format_month_year_date` to work correctly, adds it to the template renderer and adds/updates tests.  This change takes into consideration the following:
- There is an existing `format_month_year_date` filter defined but it is not initialised in the template renderer class.
- There is an existing `format_date` filter that could render month-year dates but currently is only used to format metadata (e.g. exercise start date) and fails to parse answers (e.g. in the form `[Markup('2000-02')]').  This could have been modified but ideally `format_date` filter should be singular in purpose and not a catch-all date formatter.
- Additionally the existing `format_date` filter does not have an overridable format parameter.
- This update takes the opportunity to improve the layout of affected code without refactoring the source too extensively.

### How to review
- Examine the code and tests, run the tests.
- Try the formatter in a schema (e.g. `labour_force.json`, `continuous-in-uk-question` and format the `arrive-in-uk-answer` month-year answer.

### DIFF
--- a/app/templating/template_renderer.py
+++ b/app/templating/template_renderer.py
@@ -5,52 +5,58 @@ import json
 import re
 from jinja2 import Environment
 
-from app.jinja_filters import format_date, format_household_member_name, format_currency, format_number,\
-    get_currency_symbol, format_household_summary, format_conditional_date, format_unordered_list, \
-    format_date_range, format_household_member_name_possessive, concatenated_list, calculate_years_difference, \
-    get_current_date
+import app.jinja_filters as filters
 
 
 class TemplateRenderer:
     def __init__(self):
-        self.environment = Environment(autoescape=True)
-
-        self.environment.filters['format_date'] = format_date
-        self.environment.filters['format_household_name'] = format_household_member_name
-        self.environment.filters['format_household_name_possessive'] = format_household_member_name_possessive
-        self.environment.filters['format_household_summary'] = format_household_summary
-        self.environment.filters['format_unordered_list'] = format_unordered_list
-        self.environment.globals['format_conditional_date'] = format_conditional_date
-        self.environment.filters['format_currency'] = format_currency
-        self.environment.filters['format_number'] = format_number
-        self.environment.filters['get_currency_symbol'] = get_currency_symbol
-        self.environment.globals['format_date_range'] = format_date_range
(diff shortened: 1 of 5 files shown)


dwyeradam:
@davecharles. Devils advocate here. Would it be better to expand the existing format_date to allow for date format to be passed into it. It will then work on both date types and be a bit more flexible.

app/templating/template_renderer.py
```diff
@@ -5,52 +5,73 @@
...
-        self.environment.globals['format_conditional_date'] = format_conditional_date
-        self.environment.filters['format_currency'] = format_currency
-        self.environment.filters['format_number'] = format_number
-        self.environment.filters['get_currency_symbol'] = get_currency_symbol
-        self.environment.globals['format_date_range'] = format_date_range
-        self.environment.filters['concatenated_list'] = concatenated_list
-        self.environment.globals['calculate_years_difference'] = calculate_years_difference
-        self.environment.globals['get_current_date'] = get_current_date
+        env = Environment()
+        env.filters['format_date'
```
nealedj: Ick - I may need to argue with the linter over this. Splitting square brackets looks really odd to me.

Ideally I'd say this is a good example of where the 80 char limit makes things unnecessarily difficult.

Possible alternative:
```
env.filters['format_household_name_possessive'] = (
    filters.format_household_member_name_possessive
)
```

Or a bit more left-of-field:

```
included_filters = ['format_date', 'format_month_year_date', 'etc...']
for name in included_filters:
  env.filters[name] = getattr(filters, name)

included_globals = ['format_conditional_date', 'format_date_range', 'etc...']
for name in included_globals:
  env.globals[name] = getattr(filters, name)
```

nealedj (changes requested):

app/templating/template_renderer.py
davecharles: Agree it looks odd.  I very much like the latter (left-of-field) - that's a good steer! Alas the filter names don't all match their filter module method names :(

shahi645 (commented):
Thinking shall we put this into a schema, like `test_dates` and add a functional test?

nealedj (approved):

shahi645 (approved):
@davecharles LGTM, squishy-squashy

PR #1594, opened 2018-05-11
Enable autoescaping on TemplateRenderer

nealedj:
This enables autoescape on our `TemplateRenderer` object. This object is used for piping values into the JSON blocks in a schema. We already santise and escape inputs but this should help to avoid accidentally introducing XSS vectors in future. Codacy is also warning us about it.

A number of jinja filters and functions needed to be updated as they were returning markup but were not marking the output as a `Markup` object which is needed to avoid autoescaping.

### How to review 
- Ensure that piped values are still correctly escaped
- Ensure that markup within schemas is still rendered and not escaped
- Ensure that the output of filters and functions used within schemas (I.e. `format_household_name`) is not escaped but that any values _within_ that output are escaped.

### DIFF
--- a/app/jinja_filters.py
+++ b/app/jinja_filters.py
@@ -7,7 +7,7 @@ from dateutil import relativedelta, tz
 
 import flask
 
-from jinja2 import Markup, escape, evalcontextfilter
+from jinja2 import Markup, escape, evalcontextfilter, evalcontextfunction
 from babel import units, numbers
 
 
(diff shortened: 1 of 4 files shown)


davecharles:
Given these inputs:
screen shot 2018-05-11 at 13 31 17

The escaping results are inconsistent, this is what I'd expect:
screen shot 2018-05-11 at 13 31 25

But not this:
screen shot 2018-05-11 at 13 31 39

Or this:
screen shot 2018-05-11 at 13 32 05

davecharles (commented):
Looks good but I'm not sure the escaping results is what we want but also not sure if it can be avoided.

nealedj:
Ah, that's a fairly good example of where a filter isn't currently marking up a string that it's returning. `format_household_member_name` takes an already escaped set of names and then concatenates them. But the resulting string isn't marked as safe so it escapes them again.

I'll adjust that filter to be a bit less naive.

davecharles (approved):
LGTM

shahi645 (approved):

shahi645:
Needs rebasing then squishy-squashy

PR #1596, opened 2018-05-15
Ability to validate schemas against local schema-validator

MebinAbraham:
Just a quick PR to add the ability to validate schemas against `eq_schema_validator` locally instead of using docker.

You can now pass `--local` as the 1st or 2nd argument to run it locally.
- `/scripts/test_schemas.sh --local` to run all schemas locally.
- `./scripts/test_schemas.sh data/en/0_star_wars.json --local` to run a specific schema

Note: It will run against what ever is running on port 5001 so, if your docker `eq-schema_validator` is running, it will run against that.

### How to review 
1. `run_app` on `eq-schema_validator`
2. Run `test_schemas` script with and without `--local` flag.

Schemas should be validated as expected.

### DIFF
--- a/scripts/test_schemas.sh
+++ b/scripts/test_schemas.sh
@@ -1,9 +1,15 @@
 #!/usr/bin/env bash
 
-docker pull onsdigital/eq-schema-validator
-validator="$(docker run -d -p 5001:5000 onsdigital/eq-schema-validator)"
+run_docker=true
+if [ "$1" == "--local" ] || [ "$2" == "--local" ]; then
+    run_docker=false
+fi
 
-sleep 3
+if [ "$run_docker" == true ]; then
+    docker pull onsdigital/eq-schema-validator
+    validator="$(docker run -d -p 5001:5000 onsdigital/eq-schema-validator)"
+    sleep 3
+fi
 
 green="$(tput setaf 2)"
 red="$(tput setaf 1)"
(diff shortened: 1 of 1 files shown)


scripts/test_schemas.sh
```diff
@@ -1,9 +1,15 @@
 #!/usr/bin/env bash
 
-docker pull onsdigital/eq-schema-validator
-validator="$(docker run -d -p 5001:5000 onsdigital/eq-schema-validator)"
+run_docker=true
+if [ "$1" == "-local" ] || [ "$2" == "-local" ]; then
```
nealedj: Could we use double hyphens here? By convention, bash arguments prefixed with a single hypen tend to be single character and can be concatenated (I.e. `ls -al`).

So I'd expect to write `./test_schemas --local`

nealedj (changes requested):

scripts/test_schemas.sh
shahi645: i agree otherwise, this LGTM

shahi645: just gonna test

shahi645 (approved):

scripts/test_schemas.sh
MebinAbraham: Sorted. 👍

nealedj (approved):
Nice - really useful.

PR #1581, opened 2018-05-04
LFS Alpha fix-ups

davecharles:
### What is the context of this PR?
Social Demographic submitted comprehensive comments on the first cut alpha for LFS; this PR implements many of their change requests and also incorporates some fixes.

### How to review 
Run local eq-survey-runner instance using this branch and local go-launcher (LFS already in go-launcher master), or use this [quick launch URL](https://preprod-surveys-launch.eq.ons.digital/quick-launch?url=https://raw.githubusercontent.com/ONSdigital/eq-survey-runner/eq-2072-lfs-alpha-fixups/data/en/labour_force.json).

**Note** there are some oddities and inconsistencies with other eq schemas at the request of SD - any queries please come and look at marked up hard copies I have.

PR #1599, opened 2018-05-17
Add min and max jinja filters

davecharles:
Social Demographic require a page in LFS to render the most recent of two date responses.  This PR implements two general purpose filters `max_value` and `min_value` that fulfil the SD requirement as well as providing a useful feature for other use cases.  Currently blocking #1581.

### How to review 
- Review code in `app/jinja_filters.py` and `app/templating/template_renderer.py`.
- Review and run related tests in `tests/app/templating/test_template_renderer.py` and `tests/app/test_jinja_filters.py`.
- Apply the following patch to `labour_force.json` in `master` and run the survey (rename to `.diff`):
    [lfs.diff](https://github.com/ONSdigital/eq-survey-runner/files/2014939/lfs.diff.txt)

### DIFF
--- a/app/jinja_filters.py
+++ b/app/jinja_filters.py
@@ -8,6 +8,8 @@ from dateutil import relativedelta, tz
 import flask
 
 from jinja2 import Markup, escape, evalcontextfilter, evalcontextfunction
+from jinja2.exceptions import UndefinedError
+
 from babel import units, numbers
 
 
@@ -199,6 +201,26 @@ def format_number_to_alphabetic_letter(number):
     if 0 <= int(number) < 26:
         return string.ascii_lowercase[int(number)]
     return ''
+
+
+@blueprint.app_template_filter()
+def min_value(a, b):
+    try:
+        return min(i for i in [a, b] if i is not None)
+    except (TypeError, UndefinedError):
+        msg = ('Cannot determine minimum of incompatible types '
+               'min({}, {})'.format(a.__class__, b.__class__))
+        raise Exception(msg)
+
+
+@blueprint.app_template_filter()
+def max_value(a, b):
+    try:
+        return max(i for i in [a, b] if i is not None)
+    except (TypeError, UndefinedError):
+        msg = ('Cannot determine maximum of incompatible types '
+               'max({}, {})'.format(a.__class__, b.__class__))
+        raise Exception(msg)
 
 
 @blueprint.app_context_processor
--- a/app/templating/template_renderer.py
+++ b/app/templating/template_renderer.py
@@ -24,6 +24,8 @@ class TemplateRenderer:
         env.filters['concatenated_list'] = filters.concatenated_list
         env.globals['calculate_years_difference'] = filters.calculate_years_difference
         env.globals['get_current_date'] = filters.get_current_date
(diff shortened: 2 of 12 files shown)


shahi645 (changes requested):
Again i think this would definitely benefit if it was in a test_schema of sought accompanied by some functional tests, @nealedj what would your thinking be here?

app/jinja_filters.py
```diff
@@ -201,6 +203,26 @@ def format_number_to_alphabetic_letter(number):
...
+
+
+@blueprint.app_template_filter()
+def max_value(a, b):
+    try:
+        return max(i for i in [a, b] if i is not None)
+    except (TypeError, UndefinedError):
+        msg = ('Cannot determine maximum of incompatible types '
+               'max({}, {})'.format(a.__class__, b.__class__))
+        raise Exception(msg)
```
shahi645, line 223: @davecharles Can we pass in the min or max here then have this all in same method, as the code is very similar

nealedj: In this case I'd leave them separate for readability.

nealedj:
@shahi645 In this case I don't think that functional tests provide any value above what unit tests provide.

A functional test could test the combination of the filter and the jinja renderer. But the unit test for `TemplateRenderer` already covers that. Beyond that I can't see what extra scenarios are added.

I don't want to start adding functional tests for every change.

Martin Fowler puts it better:

> I always argue that high-level tests are there as a second line of test defence. If you get a failure in a high level test, not just do you have a bug in your functional code, you also have a missing or incorrect unit test. Thus I advise that before fixing a bug exposed by a high level test, you should replicate the bug with a unit test. Then the unit test ensures the bug stays dead.

https://martinfowler.com/bliki/TestPyramid.html

nealedj (approved):

dwyeradam:
@nealedj a date for the diary as I agree with @shahi645. I agree we may don't want to add a load of functional test but I would like to see an implementation in a test schema otherwise we're not testing the implementation of this filter. Also our test schemas make it simpler to see how we would implement this when needed in a real survey. Maybe just add to an existing test_dates.json etc.

nealedj:
Yep just chatted about this a little more. There's value in exposing the new filters in at least one test schema. Otherwise there's no way to test it (even manually) - and we don't want to rely on live surveys to be the sole consumer of functionality.

@davecharles would you mind adding a use of the filter to the test_dates, test_numbers and test_textfield schemas?

shahi645 (approved):
LGTM. Quick comment for line: https://github.com/ONSdigital/eq-survey-runner/pull/1599/commits/6150e4e4d49126f892af8441f492bed3dc84f429#diff-ed4511c7a48d60265a6bfa720e219c00R15
Unrelated to this work, but as we are editing should probably add some guidance like; 'Please enter a minimum value between 0 and 1000 and a maximum greater than 1001'.

davecharles:
@shahi645 good point - will add to backlog

shahi645:
@davecharles was thinking as its just a line to a schema adding it here, but i dont mind either way, just a suggestion. Can merge this in now and come back to it later, add a ticket that covers all test schemas?

PR #1600, opened 2018-05-21
2112 address codacy warning about using jinja2 rendering without autoescape

nealedj:
Defect fix stemming from #1594

Fixes another Jinja filter that wasn't returning a string marked as safe

### How to review 
Test MWSS (1_0005.json) and assert that the period as shown in the header is not escaped.
Test one of the RSI schemas (E.g. 1_0205.json) and assert the same. These schemas make use of `format_conditional_date` which should benefit from the same fix.

### DIFF
--- a/app/jinja_filters.py
+++ b/app/jinja_filters.py
@@ -145,9 +145,12 @@ def calculate_years_difference(from_str, to_str):
 @evalcontextfunction
 def format_date_range(context, start_date, end_date=None):
     if end_date:
-        return '{from_date} to {to_date}'.format(from_date=format_date(context, start_date),
-                                                 to_date=format_date(context, end_date))
-    return format_date(context, start_date)
+        result = '{from_date} to {to_date}'.format(from_date=format_date(context, start_date),
+                                                   to_date=format_date(context, end_date))
+    else:
+        result = format_date(context, start_date)
+
+    return mark_safe(context, result)
 
 
 @blueprint.app_template_filter()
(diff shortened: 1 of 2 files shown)


shahi645 (approved):
You were getting a flake 8 failure so i pushed the tiny fix. Otherwise all looks good

app/jinja_filters.py
```diff
@@ -145,9 +145,12 @@ def calculate_years_difference(from_str, to_str):
 @evalcontextfunction
 def format_date_range(context, start_date, end_date=None):
     if end_date:
-        return '{from_date} to {to_date}'.format(from_date=format_date(context, start_date),
-                                                 to_date=format_date(context, end_date))
-    return format_date(context, start_date)
+        result = '{from_date} to {to_date}'.format(from_date=format_date(context, start_date),
+                                                   to_date=format_date(context, end_date))
+    else:
+        result = format_date(context, start_date)
```
AndrewTorrance, line 151: format date ends in a call to mark safe . I can see it both ways . At least its explicit here under all circumstances and it will be a trivial cost to leaving it this way - explicit over implicit and all that ....

AndrewTorrance (approved):
One comment on explicit vs implicit call to mark_safe.

app/jinja_filters.py
nealedj: Good point. Let's just pretend that I intended to do it this way and it wasn't just a happy oversight :)

PR #1601, opened 2018-05-22
DynamoDB spike

nealedj:
This PR is only here for the purposes getting the build working on travis.


AndrewTorrance (commented):
Is this still a spike ? It looks more like an implementation ? If its a spike then I would say it is at a level where we can make an informed decision as to if we should proceed. If its an implementation then I think it should have more integration/chimp tests e.g  when around having sessions in both rds and dynamo concurrently ?

nealedj:
Yeah it's become quite a feature-full spike.

There's still some gap for this to be considered production-ready:

- setting `created_at` and `updated_at` timestamps
- adding tests for RBS fallback
- consider allowing data storage to be set on a per-table basis (or at least allow us to still deploy with the submission table in Dynamo and the other three still in RDS)
- Investigate the extra 40-60ms latency that seems to be exhibited on each request when dynamodb is enabled

nealedj:
Closing for now as spike is on hold.

PR #1602, opened 2018-05-22
Correct employment date on 0867 and 0873

nealedj:
Fixes piped employment date in 0867 and 0873.

### How to review 
In the "Of the X total employees employed on Y, how many male and female employees worked the following hours?" question ensure that the value passed as "employment date" is shown in Y.


shahi645 (approved):
oh yes, we flagged this before but must not have raised a ticket

MebinAbraham (approved):

PR #1605, opened 2018-05-25
Pay pattern confirmation in MWSS missing default routing rule

shahi645:
Schema validator wasn't working for routing rules properly, so a fix revealed invalid schema in `1_0005.json`. All that was missing was a default routing rule which has been added. 

### How to review 
MWSS should still work functionally as it was

### DIFF
--- a/data/en/1_0005.json
+++ b/data/en/1_0005.json
@@ -275,6 +275,11 @@
                                         "value": "Five weekly"
                                     }]
                                 }
+                            },
+                            {
+                                "goto": {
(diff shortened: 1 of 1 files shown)


nealedj (approved):
LGTM - this can never be hit given the current options. However it's a sensible place to default to since we need a default.

shahi645:
yeah thats exactly it @nealedj it makes no difference apart from its technically an invalid schema

davecharles:
LGTM - squishy squashy

davecharles (approved):

PR #1606, opened 2018-05-25
Allow for different versions of a question title depending on a conditional question

shahi645:
We have a use case in SR where the user answers a proxy question, e.g. "Who are you answering on behalf of?". This means that we want subsequent questions to be asked differently depending on the users answers. Previously this meant using routing rules and having a lot of excess schema.
This PR introduces a `titles` object to question, this can be used instead of a single `title` value. The titles object consists of a `default value` with is a single title, with optional extra `titles` that are accompanied with a `when` rule (`when` functionality already exists in routing rules, so we have re-used that code). When the question title is rendered each `when clause` is evaluated and if true returns its given `title`. If no `when clauses` evaluate true then the `default title` is returned and rendered.

**UPDATE**
We have added changes that now mean all getting of question titles are handled by our templating utils and template logic on selecting the question title is handled in our jinja filters.

***NOTE***
Currently we have a bug with repeating groups that are as follows: 
- The summary only renders the latest group instance of a repeating group. (existing bug)

- When evaluating a `when` clause in a repeating group greater than `group_instance` **0** it defaults to 0 `group_instance` for summary rendering, as summary has no knowledge of which group instance the question/answer has come from.

### How to review 
Run our test_schemas created and see if the changes are correct. 
Also does the code changes make sense as to where they are.

### DIFF
--- a/app/forms/questionnaire_form.py
+++ b/app/forms/questionnaire_form.py
@@ -12,6 +12,7 @@ from werkzeug.datastructures import MultiDict
 
 from app.forms.fields import get_field
 from app.forms.date_form import get_dates_for_single_date_period_validation
+from app.templating.utils import get_question_title
 from app.validation.validators import DateRangeCheck, SumCheck
 
 logger = logging.getLogger(__name__)
@@ -259,7 +260,7 @@ class QuestionnaireForm(FlaskForm):
         return None
 
 
-def get_answer_fields(question, data, error_messages, answer_store, metadata):
+def get_answer_fields(question, data, error_messages, answer_store, metadata, group_instance):
     answer_fields = {}
     for answer in question['answers']:
         if 'parent_answer_id' in answer and answer['parent_answer_id'] in data and \
@@ -267,7 +268,7 @@ def get_answer_fields(question, data, error_messages, answer_store, metadata):
             answer['mandatory'] = \
                 next(a['mandatory'] for a in question['answers'] if a['id'] == answer['parent_answer_id'])
 
-        name = answer.get('label') or question.get('title')
+        name = answer.get('label') or get_question_title(question, answer_store, metadata, group_instance)
         answer_fields[answer['id']] = get_field(answer, name, error_messages, answer_store, metadata)
     return answer_fields
 
@@ -298,12 +299,12 @@ def map_child_option_errors(errors, answer_json):
     return child_errors
 
 
-def generate_form(schema, block_json, data, answer_store, metadata):
+def generate_form(schema, block_json, data, answer_store, metadata, group_instance):
     answer_fields = {}
 
     class DynamicForm(QuestionnaireForm):
         for question in schema.get_questions_for_block(block_json):
-            answer_fields.update(get_answer_fields(question, data, schema.error_messages, answer_store, metadata))
+            answer_fields.update(get_answer_fields(question, data, schema.error_messages, answer_store, metadata, group_instance))
 
     for answer_id, field in answer_fields.items():
         setattr(DynamicForm, answer_id, field)
--- a/app/jinja_filters.py
+++ b/app/jinja_filters.py
@@ -7,7 +7,7 @@ from dateutil import relativedelta, tz
 
 import flask
 
-from jinja2 import Markup, escape, evalcontextfilter, evalcontextfunction
+from jinja2 import Markup, escape, evalcontextfilter, evalcontextfunction, contextfunction
 from jinja2.exceptions import UndefinedError
 
 from babel import units, numbers
@@ -226,6 +226,47 @@ def max_value(a, b):
         msg = ('Cannot determine maximum of incompatible types '
                'max({}, {})'.format(a.__class__, b.__class__))
         raise Exception(msg)
+
+
+@contextfunction
+@blueprint.app_template_filter()
+def get_question_title(context, question_id):
+    """Return the value that should be used as the title to a question
+    May be from question.title or question_titles in the context"""
+    context = context.parent
+    question = context['question']
+
+    if question_id == question['id']:
+        if question.get('title') is not None:
+            return question['title']
+        question_title = context['content']['question_titles']
+        return question_title[question_id]
+
+
+@contextfunction
+@blueprint.app_template_filter()
+def get_answer_label(context, answer_id, question_id):
+    """Return the value that should be used as the answer label tries
+    answer.label first then resorts to question title"""
+    parent_context = context.parent
+    question = parent_context['question']
+    answers = question['answers']
+
+    for answer in answers:
+        if answer_id == answer['id']:
+            if answer.get('label') is not None:
+                return answer['label']
+            return get_question_title(context, question_id)
+
+
+@blueprint.app_context_processor
+def get_question_title_processor():
+    return dict(get_question_title=get_question_title)
+
+
+@blueprint.app_context_processor
+def get_answer_label_processor():
+    return dict(get_answer_label=get_answer_label)
 
 
 @blueprint.app_context_processor
--- a/app/questionnaire/questionnaire_schema.py
+++ b/app/questionnaire/questionnaire_schema.py
@@ -1,6 +1,7 @@
 import itertools
-from collections import OrderedDict, defaultdict
+from collections import OrderedDict
 from app.validation.error_messages import error_messages
+from app.questionnaire.answer_dependencies import get_dependencies
 
 
 class QuestionnaireSchema(object):  # pylint: disable=too-many-public-methods
@@ -28,6 +29,10 @@ class QuestionnaireSchema(object):  # pylint: disable=too-many-public-methods
     def answers(self):
         return self._answers_by_id.values()
 
+    @property
+    def dependencies(self):
+        return self._answer_dependencies
+
     def get_section(self, section_id):
         return self._sections_by_id.get(section_id)
 
@@ -42,9 +47,6 @@ class QuestionnaireSchema(object):  # pylint: disable=too-many-public-methods
 
     def get_answer(self, answer_id):
         return self._answers_by_id.get(answer_id)
-
-    def get_answer_dependencies_by_id(self, answer_id):
-        return self._answer_dependencies.get(answer_id, [])
 
     def get_section_by_block_id(self, block_id):
         block = self.get_block(block_id)
@@ -191,7 +193,7 @@ class QuestionnaireSchema(object):  # pylint: disable=too-many-public-methods
         self._questions_by_id = get_nested_schema_objects(self._blocks_by_id, 'questions')
         self._answers_by_id = get_nested_schema_objects(self._questions_by_id, 'answers')
         self.error_messages = self._get_error_messages()
-        self._answer_dependencies = self._get_answer_dependencies()
+        self._answer_dependencies = get_dependencies(self)
 
     def _get_sections_by_id(self):
         return OrderedDict(
@@ -206,27 +208,17 @@ class QuestionnaireSchema(object):  # pylint: disable=too-many-public-methods
 
         return messages
 
-    def _get_answer_dependencies(self):
-        dependencies = defaultdict(list)
-        # Answer level dependencies
-        for answer in self.answers:
-            dependency_id = answer['id']
-            if 'min_value' in answer and 'answer_id' in answer['min_value']:
-                answer_id = answer['min_value']['answer_id']
-                dependencies[answer_id].append(dependency_id)
-            if 'max_value' in answer and 'answer_id' in answer['max_value']:
-                answer_id = answer['max_value']['answer_id']
-                dependencies[answer_id].append(dependency_id)
+    def _get_answers_by_id_for_question(self, question_id):
+        question = self.get_question(question_id)
 
-        # Question level dependencies
-        for question in self.questions:
-            for calculation in question.get('calculations', []):
-                answer_id = calculation.get('answer_id')
-                if answer_id:
-                    for dependency_id in calculation['answers_to_calculate']:
-                        dependencies[answer_id].append(dependency_id)
+        return {
+            answer['id']: answer
+            for answer in question['answers']
+        }
 
-        return dependencies
+    def get_answer_ids_for_question(self, question_id):
+        """ get answer ids associated with a specific question_id """
+        return list(self._get_answers_by_id_for_question(question_id).keys())
 
 
 def get_nested_schema_objects(parent_object, list_key):
--- a/app/questionnaire/rules.py
+++ b/app/questionnaire/rules.py
@@ -174,7 +174,10 @@ def evaluate_when_rules(when_rules, metadata, answer_store, group_instance):
     """
     for when_rule in when_rules:
         if 'id' in when_rule:
-            value = get_answer_store_value(when_rule['id'], answer_store, group_instance)
+            when_specified_group_instance = when_rule.get('group_instance')
+            instance = when_specified_group_instance or group_instance
+
+            value = get_answer_store_value(when_rule['id'], answer_store, instance)
         elif 'meta' in when_rule:
             value = get_metadata_value(metadata, when_rule['meta'])
         else:
--- a/app/templating/summary/question.py
+++ b/app/templating/summary/question.py
@@ -1,16 +1,19 @@
 import collections
 
 from app.templating.summary.answer import Answer
+from app.templating.utils import get_question_title
 
 
 class Question:
 
-    def __init__(self, question_schema, answer_store):
+    def __init__(self, question_schema, answer_store, metadata):
         self.id = question_schema['id']
         self.type = question_schema['type']
 
         answer_schemas = iter(question_schema['answers'])
-        self.title = question_schema['title'] or question_schema['answers'][0]['label']
+
+        # Using group instance as 0 for now as summary rendering context only has knowledge of current locations group instance (i.e. 0)
+        self.title = get_question_title(question_schema, answer_store, metadata, group_instance=0) or question_schema['answers'][0]['label']
         self.number = question_schema.get('number', None)
         self.answers = self._build_answers(answer_store, question_schema, answer_schemas)
 
@@ -31,7 +34,8 @@ class Question:
                 continue
             if question_schema['type'] == 'RepeatingAnswer':
                 for answer_value in self._get_answers(answer_store, answer_schema['id']):
-                    answer = self._build_answer(answer_store, question_schema, answer_schema, answer_schemas, answer_value)
+                    answer = self._build_answer(answer_store, question_schema, answer_schema, answer_schemas,
+                                                answer_value)
                     summary_answers.append(Answer(answer_schema, answer))
             else:
                 answer_value = self._get_answer(answer_store, answer_schema['id'])
--- a/app/templating/utils.py
+++ b/app/templating/utils.py
@@ -1,0 +1,22 @@
+from app.questionnaire.rules import evaluate_when_rules
+
+
+def get_question_title(question_schema, answer_store, metadata, group_instance):
+    """Return the value that should be used as the title to a question
+    May be from question.title or question.titles"""
+
+    if question_schema.get('title') is not None:
+        return question_schema['title']
+
+    titles = question_schema.get('titles')
+
+    return get_title_from_titles(metadata, answer_store, titles, group_instance)
+
+
+def get_title_from_titles(metadata, answer_store, titles, group_instance):
+    """returns a title from titles available by evaluating the when rules , if all fail returns default"""
+
+    for when, value in ((title['when'], title['value']) for title in titles if 'when' in title):
+        if evaluate_when_rules(when, metadata, answer_store, group_instance):
+            return value
+    return titles[-1]['value']
--- a/app/templating/view_context.py
+++ b/app/templating/view_context.py
@@ -5,10 +5,10 @@ from app.globals import is_dynamodb_enabled
 from app.helpers.form_helper import get_form_for_location
 from app.templating.summary_context import build_summary_rendering_context
 from app.templating.template_renderer import renderer
+from app.templating.utils import get_title_from_titles
 
 
 def build_view_context(block_type, metadata, answer_store, schema_context, rendered_block, current_location, form):
-
     variables = None
     if g.schema.json.get('variables'):
         variables = renderer.render(g.schema.json.get('variables'), **schema_context)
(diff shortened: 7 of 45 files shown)


shahi645:
please note travis build wont pass until https://github.com/ONSdigital/eq-survey-runner/pull/1605 is merged

app/questionnaire/questionnaire_schema.py
```diff
@@ -206,8 +206,27 @@ def _get_error_messages(self):
 
         return messages
 
+    def _get_answer_ids_in_question(self, question_id):
```
nealedj: There's already a `get_answers_by_id_for_block` method. Could we follow the convention and rename this to `get_answer_ids_for_question`?

Following the convention you'd have

```
    def get_answers_by_id_for_question(self, question_id):
        question = self.get_question(question_id)
        if question:
            return {
                answer['id']: answer
                for answer in question['answers']
            }

        return {}

    def get_answer_ids_for_question(self, question_id):
        return list(self.get_answers_by_id_for_question(question_id).keys())

    def get_answers_for_question(self, question_id):
        return list(self.get_answers_by_id_for_question(question_id).values())
```

app/questionnaire/questionnaire_schema.py
```diff
@@ -226,7 +246,30 @@ def _get_answer_dependencies(self):
                     for dependency_id in calculation['answers_to_calculate']:
                         dependencies[answer_id].append(dependency_id)
 
-        return dependencies
+    def _set_titles_dependencies_for_all_questions(self, dependencies):
+        """ evaluate a questions dependencies via when clauses on titles """
+        for question in self.questions:
+            self._set_titles_dependencies_for_question(dependencies, question)
+
+    def _set_titles_dependencies_for_question(self, dependencies, question):
```
nealedj: Lots of nesting here. Making it quite tricky to read.

Possible refactor:

```
        for title in question.get('titles', []):
            if 'when' not in title:
                continue

        when_clauses = title.get('when')

        for when_clause in when_clauses:
            for answer_id in answer_ids:
                when_id = when_clause.get('id')
                if answer_id not in dependencies[when_id]:
                    dependencies[when_id].append(answer_id)
```

app/questionnaire/questionnaire_schema.py
```diff
@@ -206,8 +206,27 @@ def _get_error_messages(self):
...
+        question = self.get_question(question_id)
+        for answer in question['answers']:
+            if not answer['id'] in answer_ids:
+                answer_ids.append(answer['id'])
+        return answer_ids
+
     def _get_answer_dependencies(self):
         dependencies = defaultdict(list)
+
+        self._get_answer_level_dependencies(dependencies)
```
nealedj: Some of these methods are called `_set_xxx` and others `_get_xxx` but it's not clear why. As they're all mutating the `dependencies` object `set` might be more appropriate.

app/questionnaire/questionnaire_schema.py
```diff
@@ -206,8 +206,27 @@ def _get_error_messages(self):
...
+        """ get answer ids associated with a specific question_id """
+        answer_ids = []
+        question = self.get_question(question_id)
+        for answer in question['answers']:
+            if not answer['id'] in answer_ids:
+                answer_ids.append(answer['id'])
+        return answer_ids
+
     def _get_answer_dependencies(self):
         dependencies = defaultdict(list)
```
nealedj: Consider making this a `defaultdict(set)` so that you can add to it without the need to check if a value is already present.

E.g.
```
>>> question_id, answer_id = 'foo', 'bar'
>>> dependencies = defaultdict(set)
>>> dependencies[question_id].add(answer_id)
>>> dependencies[question_id].add(answer_id)
{'foo': {'bar'}}
```

app/questionnaire/rules.py
```diff
@@ -172,7 +172,10 @@ def evaluate_when_rules(when_rules, metadata, answer_store, group_instance):
     """
     for when_rule in when_rules:
         if 'id' in when_rule:
-            value = get_answer_store_value(when_rule['id'], answer_store, group_instance)
+            when_specified_group_instance = when_rule.get('group_instance')
+            instance = when_specified_group_instance if when_specified_group_instance is not None else group_instance
```
nealedj: Could be reduced to:
`instance  = when_specified_group_instance or group_instance`

app/templating/summary/question.py
```diff
@@ -1,19 +1,37 @@
...
 
         answer_schemas = iter(question_schema['answers'])
-        self.title = question_schema['title'] or question_schema['answers'][0]['label']
+        self.title = self.get_question_title(question_schema, answer_store, metadata)
         self.number = question_schema.get('number', None)
         self.answers = self._build_answers(answer_store, question_schema, answer_schemas)
 
+    @classmethod
+    def get_title_from_titles(cls, metadata, answer_store, titles, group_instance):
+        when_rules = [(title['when'], title['value']) for title in titles if 'when' in title]
```
nealedj: Nitpick - could be a generator:

```
when_rules = ((title['when'], title['value']) for title in titles if 'when' in title)
```

app/templating/summary/question.py
```diff
@@ -1,19 +1,37 @@
...
+        for when in when_rules:
```
nealedj: Indexing a tuple rather than unpacking to variables.

Possibly more readable:

```
for when, value in when_rules:
    if evaluate_when_rules(when, metadata, answer_store, group_instance):
        return value
```

app/templating/summary/question.py
```diff
@@ -1,19 +1,37 @@
...
+            if evaluate_when_rules(when[0], metadata, answer_store, group_instance):
+                return when[1]
+        return titles[-1]['value']
+
+    @classmethod
+    def get_question_title(cls, question_schema, answer_store, metadata):
+        if question_schema.get('title'):
```
nealedj: What's the expected behaviour if the `title` is defined but == `""`?

Given that the templates are using the `is defined` jinja operator I'd say that for consistency we should treat empty strings as a defined title.

So this could become `if 'title' in question_schema:`

app/templating/summary/question.py
```diff
@@ -1,19 +1,37 @@
...
+            return question_schema['title']
+
+        titles = question_schema.get('titles')
+
+        return cls.get_title_from_titles(metadata, answer_store, titles, 0) if titles else question_schema['answers'][0]['label']
```
nealedj: Suggest that this is opened onto separate lines for readability:

```
if 'titles' in question_schema:
    return cls.get_title_from_titles(metadata, answer_store, titles, 0)
else:
    return question_schema['answers'][0]['label']
```

nealedj: We're explicitly setting the `group_instance` to `0` here but without explanation.

We know it's because summaries don't yet support repeating groups - could we put a comment to explain?

data/en/test_different_question_titles.json
```diff
@@ -0,0 +1,204 @@
...
+                },
+                {
+                    "type": "Question",
+                    "id": "who-is-answering-block",
+                    "questions": [{
+                        "id": "behalf-of-question",
+                        "title": "Who are you answering on behalf of?",
+                        "guidance": {
+                            "content": [{
+                                "description": "The answer you choose will have an affect on question titles in next question"
```
nealedj: effect ;)

scripts/test_schemas.sh
```diff
@@ -6,8 +6,8 @@ if [ "$1" == "--local" ] || [ "$2" == "--local" ]; then
 fi
 
 if [ "$run_docker" == true ]; then
-    docker pull onsdigital/eq-schema-validator
-    validator="$(docker run -d -p 5001:5000 onsdigital/eq-schema-validator)"
+    docker pull onsdigital/eq-schema-validator:eq-1990-different-versions-of-question-titles
```
nealedj: Note to remove this before merge.

tests/app/forms/test_questionnaire_form.py
```diff
@@ -869,6 +869,40 @@ def test_multi_calculation(self):
             self.assertEqual(form.question_errors['breakdown-question'],
                              schema.error_messages['TOTAL_SUM_NOT_EQUALS'] % dict(total='10'))
 
+    def test_generate_form_with_titles(self):
```
nealedj: I don't understand what this is testing.

app/templating/view_context.py
```diff
@@ -53,13 +53,18 @@ def build_view_context_for_question(current_location, variables, rendered_block,
             'answer_errors': {},
             'data': {},
             'other_answer': {},
-            'fields': {}
+            'fields': {},
+            'question_titles': {}
```
nealedj: I don't think the `question_titles` should be nested inside the `form` object. They don't belong to it.

In the absence of a proper view model I'd suggest patching the title directly onto the `question` objects within each `rendered_block`.

```
for question in rendered_block.get('questions'):
    question['question_title`] = get_the_title()
```

Then the templates wouldn't need any logic for working out where to get the title from.

app/templating/view_context.py
```diff
@@ -5,10 +5,10 @@
 from app.helpers.form_helper import get_form_for_location
 from app.templating.summary_context import build_summary_rendering_context
 from app.templating.template_renderer import renderer
+from app.templating.summary.question import Question
```
nealedj: This is being imported so that the `get_title_from_titles` function can be called. I think that implies that the function is in the wrong place because it isn't specific to the summary.

Consider adding a `utils` module to the `templating` package for now and add the function there instead.

nealedj (changes requested):
Well done - loads of good stuff here. Lots of little comments but nothing fundamental. The main thing that I think could be improved is where the titles are rendered into the view context.

General point - could the test schemas be named such that they're grouped together? Might be useful for these bigger features that require multiple schemas.

E.g. test_titles.json, test_titles_within_repeating_blocks.json, test_titles_radio_checkbox

app/questionnaire/questionnaire_schema.py
AndrewTorrance: Done

app/questionnaire/questionnaire_schema.py
AndrewTorrance: Did a different implementation using list comprehensions.

app/questionnaire/questionnaire_schema.py
AndrewTorrance: That one does not directly set the self.dependencies , it returns dependencies which other code assigns to the self.dependencies so I think its  valid as it does not off itself affect the object state.

nealedj: The two `_get` methods and the `_set` method both mutate the `dependencies` argument that gets passed to them - I'm not why they employ different verbs.

Generally (subjectively) I think I'd expect a method start with `get` to not mutate any state. A `set` method might mutate the state of the object.

In this case perhaps `build` might be a better word to use?

app/questionnaire/questionnaire_schema.py
AndrewTorrance: I like the idea , but the dependencies are used by existing code , so I'd rather leave as is for now.

app/questionnaire/rules.py
AndrewTorrance: good catch

app/templating/summary/question.py
AndrewTorrance: Changed , I was toying with a named tuple , but no need for 1 use simple unpacking is perfectly good

app/templating/summary/question.py
AndrewTorrance: Its is nitpicky , but valid , changed

app/templating/summary/question.py
AndrewTorrance: Its not that straightforward . The pre existing logic was if title =='' then use the value defined in the answer label as the title. So we changed the logic to
if schema[title] return schema[title]
if valid title in titles return that title
else return answer label for question 0 

I think should be revisited when looking at repeating groups .

app/templating/summary/question.py
AndrewTorrance: mmm com si com sa , changed it without the else

AndrewTorrance: Added

data/en/test_different_question_titles.json
AndrewTorrance: :-)

scripts/test_schemas.sh
AndrewTorrance: Added comment to script. PR also has do not merge on it

app/templating/view_context.py
AndrewTorrance: We raised something similar as a possibility at the beginning but we shied away from that after sharing the ideas because of the need to know if the question was inside a repeating group / summary . We can do it but it would mean a significant rewrite .

app/templating/view_context.py
AndrewTorrance: Your spot on. Adding it to view context caused circular dependencies . Moving out to a utils file

shahi645:
@nealedj ready for re-review

app/forms/questionnaire_form.py
```diff
@@ -298,12 +299,12 @@ def map_child_option_errors(errors, answer_json):
     return child_errors
 
 
-def generate_form(schema, block_json, data, answer_store, metadata):
+def generate_form(schema, block_json, data, answer_store, metadata, group_instacne):
```
davecharles: Bug in waiting... `group_instacne`

app/jinja_filters.py
```diff
@@ -226,6 +226,47 @@ def max_value(a, b):
         raise Exception(msg)
 
 
+@contextfunction
+@blueprint.app_template_filter()
+def get_question_title(context, question_id):
```
davecharles, line 233: Existing filters have adopted the convention of `format_` - could we call theses `format_question_title` etc?

app/jinja_filters.py
```diff
@@ -226,6 +226,47 @@ def max_value(a, b):
...
+
+    for answer in answers:
+        if answer_id == answer['id']:
+            if answer.get('label') is not None:
+                return answer['label']
+            return get_question_title(context, question_id)
+
+
+@blueprint.app_context_processor
+def get_question_title_check():
```
davecharles: Maybe `question_title_processor` for consistency?

app/templating/utils.py
```diff
@@ -0,0 +1,22 @@
+from app.questionnaire.rules import evaluate_when_rules
```
davecharles, line 1: Don't understand why the separate utils module.  It's very 'titles' centric so not really utils. Less cognitive load for the reader if it's just in `view_context.py` - unless there's a reason I'm missing (e.g. avoiding circ import).

app/views/questionnaire.py
```diff
@@ -703,14 +703,19 @@ def _get_front_end_navigation(answer_store, current_location, metadata, routing_
     return None
 
 
-def get_page_title_for_location(schema, current_location):
+def _get_question_title_for_page(first_question, context):
+    return first_question['title'] if first_question.get('title') is not None else context['question_titles'][first_question['id']]
```
davecharles: I think this is a bit more readable:
```python
if first_question.get('title') is not None:
    return first_question['title']
return context['question_titles'][first_question['id']]
```
Do you explicitly want to test for None and not empty str as well?

tests/app/forms/test_household_composition.py
```diff
@@ -20,7 +20,7 @@ def _error_exists(answer_id, msg, mapped_errors):
 
     def test_get_name_form_generates_name_form(self):
         with self.test_request_context():
-            form = generate_household_composition_form(self.schema, self.block_json, {})
+            form = generate_household_composition_form(self.schema, self.block_json, {}, metadata={}, group_instance=0)
```
davecharles, line 23: The function def only has positional args so technically you don't need to name `metadata` and `group_instance` keywords. Not a big issue but I'm wondering if it says something about `generate_household_composition_form` args - do we need a `*` arg list delimiter for example?

davecharles (changes requested):

app/jinja_filters.py
shahi645: i think in this instance its better as a get, we're not mutating anything here we're just getting a value based on the context

app/templating/utils.py
shahi645: its called in several places, so we agreed in the utils would be best served

tests/app/forms/test_household_composition.py
shahi645: i just named them for clarity, i guess thats a better standard but not sure myself. Raises food for thought though, all tutorials i have watched have mentioned to name args regardless

app/jinja_filters.py
shahi645: yeah can do, the others have check but you're right processor probably sounds better

app/views/questionnaire.py
shahi645: its tested in `test_questionnaire`

app/questionnaire/questionnaire_schema.py
AndrewTorrance: I can see why its confusing . They are internal functions , none of which change the dependencies as stored on self. I have changed them all to gets as they get the dependencies . The actual set is done in the parse

tests/functional/spec/features/conditional_question_title/conditional_titles_in_repeating_groups.spec.js
```diff
@@ -1,6 +1,6 @@
 const helpers = require('../../../helpers');
 
-describe('Feature: Use of Titles in Repeating blocks', function() {
+describe('@watch Feature: Use of Titles in Repeating blocks', function() {
```
davecharles: Note to remove this before merge

tests/functional/spec/features/conditional_question_title/conditional_titles_in_repeating_groups.spec.js
```diff
@@ -0,0 +1,40 @@
+const helpers = require('../../../helpers');
+
+describe('@watch Feature: Use of Titles in Repeating blocks', function() {
```
davecharles: Need to remove this @watch

AndrewTorrance: Done

davecharles (approved):

tests/app/forms/test_household_composition.py
nealedj: As a rule I never name positional args and always name kwargs. TBH I've never questioned it but all python devs I've worked with have done the same.

app/templating/utils.py
nealedj: Ideally this logic would form part of a `Question` view model of some type. But we don't have that concept and it's quite a big one to introduce.

Separate module is because it's used by both the context builder module and the summary view models.

And yep, there would have been a circular import if it was defined in `view_context.py` since that imports the summary view model which depends on this function.

app/views/questionnaire.py
```diff
@@ -714,14 +714,21 @@ def _get_front_end_navigation(answer_store, current_location, metadata, routing_
     return None
 
 
-def get_page_title_for_location(schema, current_location):
+def _get_question_title_for_page(first_question, context):
+    if first_question.get('title') is not None:
```
nealedj: This might make more sense as `if 'title' in first_question`. My reasoning being that as a rule if `title` is defined on a schema then it should take precedence over `titles` - even if `title` is an empty string.

All this check does is adds implicit behaviour if somebody were to define `"title": null` (which validator wouldn't allow anyway).

nealedj (approved):
Really good work chaps. I've left one minor comment to consider but I'm not blocking it based on that.

dwyeradam:
Apparently I'm the Simon Cowell of eQ reviews and the designated merger here. Just doing a pass of the code and i'll merge it when i'm done.

app/questionnaire/questionnaire_schema.py
```diff
@@ -206,9 +206,33 @@ def _get_error_messages(self):
...
+
+    def _get_answers_dependencies(self):
         dependencies = defaultdict(list)
+
+        dependencies.update(self._get_answer_level_dependencies())
+
+        dependencies.update(self._get_question_level_dependencies())
+
+        dependencies.update(self._get_titles_dependencies_for_all_questions())
+
```
dwyeradam: This is a type of question level dependency so rather than loop through the questions twice it's worth calling `_get_titles_dependencies_for_question(...)` within `_get_question_level_dependencies()`. 
It may also be worth moving the existing code in `_get_question_level_dependencies()` into it's own method, something like ``_get_calculations_dependencies_for_question(...)``

app/questionnaire/questionnaire_schema.py
```diff
@@ -217,17 +241,44 @@ def _get_answer_dependencies(self):
...
+        return dependencies
 
+    def _get_titles_dependencies_for_all_questions(self):
+        """ evaluate a questions dependencies via when clauses on titles """
+        dependencies = defaultdict(list)
+        for question in self.questions:
+            self._get_titles_dependencies_for_question(dependencies, question)
         return dependencies
 
+    def _get_titles_dependencies_for_question(self, dependencies, question):
```
dwyeradam: I'm not a fan of mutating parameters. We'd have to test if we can safely merge the dictionary though if keys already exist. We may already have an issue here.

dwyeradam (changes requested):

app/questionnaire/dependencies.py
```diff
@@ -0,0 +1,39 @@
...
+    set used to prevent duplicates
+
+    To add a dependency either call add with a single dependency
+    or a+=b where both a and b are Dependencies objects.
+
+    To access dependencies for a question id simply index
+    i.e  a['question_id']
+    indexing is read only
+    """
+    def __init__(self):
```
nealedj: Instead of holding a defaultdict internally maybe consider extending a `defaultdict` for this class.

```
class Dependencies(defaultdict):
  def __init__(self):
    super().__init__(set)

  def update(*args, **kwargs):    
    for other in args:  
      self._merge(other)
    if kwargs:
      self._merge(kwargs)

  def _merge(self, other):
    for key, dependencies in other.items():
      self[key] |= dependencies
```

(untested)

app/questionnaire/dependencies.py
```diff
@@ -0,0 +1,39 @@
+from collections import defaultdict
+
+
+class Dependencies:
+    """represents the dependencies for a group of answers
+    held internally as a dict(string,set) where set contains dependent ids
+    set used to prevent duplicates
+
+    To add a dependency either call add with a single dependency
+    or a+=b where both a and b are Dependencies objects.
```
nealedj: I'm not sure it's a good idea to implement the `+` operator for the `Dependencies` object when the `Dependencies` object itself is largely acting as a dict. Dicts can't be added together, that's what `update()` is for, so I'd suggest implementing that instead.

app/questionnaire/dependency_builder.py
```diff
@@ -0,0 +1,74 @@
+from app.questionnaire.dependencies import Dependencies
+
+
+class DependenciesBuilder:
```
nealedj: Since the dependencies state is held and managed by the `Dependencies` type I don't think there's much value in this being a class itself.

In a static language this would be a static class - it holds no state. In python I'd suggest that it's more pythonic to define the methods as functions on the module instead.

nealedj (changes requested):

dwyeradam (approved):

nealedj (approved):

PR #1607, opened 2018-05-29
Consider group instance when removing answers

davecharles:
While testing LFS we were getting strange routing behaviour and answer store being updated incorrectly.  This update resolves that `app/views/questionnaire.py:_remove_answer_from_questionnaire_store` didn't consider group instance ID when updating an answer, meaning an answer in a repeating group always updated the first group instance's answer.

### How to review 
- Run the tests
- Run LFS
-  Add min 2 household occupants.
- Answer "Yes" to "Is that everyone"
- Continue
- Answer all questions for first occupant
- Begin answers for second occupant but when asked DoB skip (i.e. click Save and Continue)
- Observe that navigation still shows Occupant-1 as completed.

### DIFF
--- a/data/en/test_repeating_household_routing.json
+++ b/data/en/test_repeating_household_routing.json
@@ -1,0 +1,253 @@
+{
+    "mime_type": "application/json/ons/eq",
+    "schema_version": "0.0.1",
+    "data_version": "0.0.2",
+    "survey_id": "0",
+    "title": "Test repeating household routing",
+    "description": "Tests a repeating group with alternate paths based on an optional answer",
+    "theme": "census",
+    "legal_basis": "Voluntary",
+    "eq_id": "1",
+    "form_type": "1",
+    "metadata": {
+        "user_id": {
+            "validator": "string"
+        },
+        "period_id": {
+            "validator": "string"
+        }
+    },
+    "navigation": {
+        "visible": true
+    },
+    "sections": [{
+            "id": "household-section",
+            "title": "About the household",
+            "groups": [{
+                "id": "about-household-group",
+                "title": "About the household",
+                "blocks": [{
+                        "id": "household-composition",
+                        "type": "Question",
+                        "questions": [{
+                            "id": "household-composition-question",
+                            "title": "What are the names of everyone who lives in your household?",
+                            "type": "RepeatingAnswer",
+                            "answers": [{
+                                    "id": "first-name",
+                                    "label": "First name",
+                                    "mandatory": true,
+                                    "type": "TextField",
+                                    "validation": {
+                                        "messages": {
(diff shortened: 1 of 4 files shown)


tests/integration/views/test_questionnaire.py
```diff
@@ -535,3 +535,42 @@ def test_updating_questionnaire_store_removes_completed_block_for_calculation_de
             update_questionnaire_store_with_form_data(self.question_store, calculation_answer_location, calculation_answer_data)
 
         self.assertNotIn(dependent_location, self.question_store.completed_blocks)
+
+    def test_updating_questionnaire_store_specific_group(self):
+        g.schema = load_schema_from_params('labour', 'force')
```
nealedj: I don't think it's a good idea to tie our tests to live questionnaire schemas. Let's keep them restricted to test schemas.

Perhaps test_repeating_household could be augmented to include this scenario?

If you need to you can create a minimal test schema inline for the test and create a `QuestionnaireSchema` from it. See `test_only_question_blocks_counted_for_completeness` in `test_completeness.py` for an example. I think augmenting `test_repeating_household` is probably better in this case though.

nealedj (changes requested):

tests/integration/views/test_questionnaire.py
davecharles: Good points - will take another look

davecharles: Looked at augmenting `test_repeating_household` schema but the conditions that give rise to the issue arise specifically when you have a question that can be skipped and if so route to a question that is normally routed "around".  So rather than a fair amount of fiddling with it and obscuring the purpose of tests that currently use it, I think it would be better to use a separate test schema and add a new module `integration/household_composition/` using new schema.  Objections/thoughts?

data/en/test_repeating_household_routing.json
```diff
@@ -0,0 +1,245 @@
...
+    "mime_type": "application/json/ons/eq",
+    "schema_version": "0.0.1",
+    "data_version": "0.0.2",
+    "survey_id": "0",
+    "title": "Test repeating household routing",
+    "description": "Tests a repeating group with alternate paths based on an optional answer",
+    "theme": "census",
+    "legal_basis": "Voluntary",
+    "eq_id": "1",
+    "form_type": "1",
```
shahi645, line 11: @davecharles when you update from master, will need to add required metadata here:
```
"metadata": {
        "user_id": {
            "validator": "string"
        },
        "period_id": {
            "validator": "string"
        }
    },
```

shahi645 (approved):
Just the one comment left about the metadata, but all looks good to me

nealedj (approved):

PR #1611, opened 2018-05-31
Mutually Exclusive Checkbox

dwyeradam:
**As a** user
**I want** mutually exclusive checkbox options
**So that** I cannot select multiple options which contradict each other

### How to review 
New MutuallyExclusiveCheckbox answer type
OR option defined as last option in schema
Server side validation of mutually exclusive checkbox

test schema: **test_checkbox_mutually_exclusive.json**

### DIFF
--- a/app/forms/fields.py
+++ b/app/forms/fields.py
@@ -6,7 +6,7 @@ from structlog import get_logger
 
 from app.forms.custom_fields import MaxTextAreaField, CustomIntegerField, CustomDecimalField
 from app.forms.date_form import get_date_form, get_month_year_form
-from app.validation.validators import NumberCheck, NumberRange, ResponseRequired, DecimalPlaces
+from app.validation.validators import NumberCheck, NumberRange, ResponseRequired, DecimalPlaces, MutuallyExclusive
 
 MAX_LENGTH = 2000
 MAX_NUMBER = 9999999999
@@ -24,10 +24,11 @@ def get_field(answer, label, error_messages, answer_store, metadata):
         field = get_date_field(answer, label, guidance, error_messages, answer_store, metadata)
     elif answer['type'] == 'MonthYearDate':
         field = get_month_year_field(answer, label, guidance, error_messages, answer_store, metadata)
+    elif answer['type'] in ['Checkbox', 'MutuallyExclusiveCheckbox']:
+        field = get_select_multiple_field(answer, label, guidance, error_messages)
     else:
         field = {
             'Radio': get_select_field,
-            'Checkbox': get_select_multiple_field,
             'TextArea': get_text_area_field,
             'TextField': get_string_field,
             'Dropdown': get_dropdown_field,
@@ -124,6 +125,11 @@ def get_month_year_field(answer, label, guidance, error_messages, answer_store, 
 
 def get_select_multiple_field(answer, label, guidance, error_messages):
     validate_with = get_mandatory_validator(answer, error_messages, 'MANDATORY_CHECKBOX')
+
+    if answer['type'] == 'MutuallyExclusiveCheckbox':
+        validate_with.append(
+            MutuallyExclusive(answer.get('options'), error_messages)
+        )
 
     return SelectMultipleField(
         label=label,
--- a/app/submitter/convert_payload_0_0_1.py
+++ b/app/submitter/convert_payload_0_0_1.py
@@ -26,9 +26,10 @@ def convert_answers_to_payload_0_0_1(answer_store, schema, routing_path):
             value = answer['value']
 
             if answer_schema is not None and value is not None and 'parent_answer_id' not in answer_schema:
-                if 'q_code' in answer_schema and (answer_schema['type'] != 'Checkbox' or any('q_code' not in option for option in
-                                                                                             answer_schema['options'])):
 
+                is_single_value_answer = answer_schema['type'] not in ['Checkbox', 'MutuallyExclusiveCheckbox']
+                all_options_have_q_codes = all('q_code' in option for option in answer_schema.get('options', []))
+                if 'q_code' in answer_schema and (is_single_value_answer or not all_options_have_q_codes):
                     answer_data = _get_answer_data(data, answer_schema['q_code'], value)
                     if answer_data is not None:
                         data[answer_schema['q_code']] = _format_downstream_answer(answer_schema['type'], answer['value'], answer_data)
--- a/app/templates/partials/widgets/multiple_choice_widget.html
+++ b/app/templates/partials/widgets/multiple_choice_widget.html
@@ -15,8 +15,13 @@
   {% endif %}
 
   {% for option in form.fields[answer.id] %}
+
+    {% if exclusive and loop.last %}
+      <div class="field__label u-mt-s venus" aria-hidden="true">Or</div>
+    {% endif %}
+
     {% set input = {
-      "class": "input input--" ~ widget_type ~ " js-focusable",
+      "class": "input input--" ~ widget_type ~ " js-focusable " ~ exclusive ~ "",
       "type": widget_type,
       "value": option.data,
       "name": option.name,
@@ -35,12 +40,20 @@
     <div class="field__item js-focusable-box">
       <input {{input|xmlattr}}>
       <label {{label|xmlattr}}>
-        {{option.label.text}}
+          {% if not exclusive or (exclusive and not loop.last) %}
+            {{option.label.text}}
+          {% endif %}
         {% if answer.options[loop.index0].description %}
           <br />
           <span class="label__description label__inner pluto">{{answer.options[loop.index0].description}}</span>
         {% endif %}
+          {% if exclusive and loop.last %}
+            <span class="u-vh">Or, </span>{{option.label.text}}<span class="u-vh">. Selecting this will uncheck all other checkboxes</span>
+          {% endif %}
       </label>
+        {% if exclusive and loop.last %}
+            <span class="js-exclusive-checkbox-alert u-vh" role="alert" aria-live="polite" data-adjective="deselected"></span>
+        {% endif %}
 
       {% if answer.id in form.other_answer and form.other_answer[answer.id][loop.index0] != None %}
         {% set other_answer = form.other_answer[answer.id][loop.index0] %}
--- a/app/validation/error_messages.py
+++ b/app/validation/error_messages.py
@@ -26,4 +26,5 @@ error_messages = {
     'DATE_PERIOD_TOO_LARGE': 'Enter a reporting period less than or equal to %(max)s.',
     'SINGLE_DATE_PERIOD_TOO_EARLY': 'Enter a date after %(min)s.',
     'SINGLE_DATE_PERIOD_TOO_LATE': 'Enter a date before %(max)s.',
+    'MUTUALLY_EXCLUSIVE': 'Uncheck %(non_exclusives)s or "%(exclusive)s" to continue',
 }
--- a/app/validation/validators.py
+++ b/app/validation/validators.py
@@ -14,7 +14,7 @@ from app.questionnaire.rules import convert_to_datetime
 logger = get_logger()
 
 
-class NumberCheck(object):
+class NumberCheck:
     def __init__(self, message=None):
         if message:
             self.message = message
@@ -31,7 +31,7 @@ class NumberCheck(object):
             raise validators.StopValidation(self.message)
 
 
-class ResponseRequired(object):
+class ResponseRequired:
     """
     Validates that input was provided for this field. This is a copy of the
     InputRequired validator provided by wtforms, which checks that form-input data
@@ -55,7 +55,7 @@ class ResponseRequired(object):
             raise validators.StopValidation(self.message)
 
 
-class NumberRange(object):
+class NumberRange:
     """
     Validates that a number is of a minimum and/or maximum value, inclusive.
     This will work with any comparable number type, such as floats and
@@ -113,7 +113,7 @@ class NumberRange(object):
         return None
 
 
-class DecimalPlaces(object):
+class DecimalPlaces:
     """
     Validates that an input has less than or equal to a
     set number of decimal places
@@ -135,7 +135,7 @@ class DecimalPlaces(object):
                 raise validators.ValidationError(self.messages['INVALID_DECIMAL'] % dict(max=self.max_decimals))
 
 
-class OptionalForm(object):
+class OptionalForm:
     """
     Allows completely empty form and stops the validation chain from continuing.
     Will not stop the validation chain if any one of the fields is populated.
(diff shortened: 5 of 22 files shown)


scripts/test_schemas.sh
```diff
@@ -6,8 +6,8 @@ if [ "$1" == "--local" ] || [ "$2" == "--local" ]; then
 fi
 
 if [ "$run_docker" == true ]; then
-    docker pull onsdigital/eq-schema-validator
-    validator="$(docker run -d -p 5001:5000 onsdigital/eq-schema-validator)"
+    docker pull onsdigital/eq-schema-validator:eq-2095-mutually-exclusive-checkboxes
+    validator="$(docker run -d -p 5001:5000 onsdigital/eq-schema-validator:eq-2095-mutually-exclusive-checkboxes)"
```
dwyeradam: Will need to be reverted before merge.

shahi645: can be reverted now @dwyeradam

app/forms/fields.py
```diff
@@ -125,6 +126,11 @@ def get_month_year_field(answer, label, guidance, error_messages, answer_store,
 def get_select_multiple_field(answer, label, guidance, error_messages):
     validate_with = get_mandatory_validator(answer, error_messages, 'MANDATORY_CHECKBOX')
 
+    if answer['type'] == 'MutuallyExclusiveCheckbox':
+        validate_with = validate_with + [
```
nealedj: Could use this to avoid creating a list:

```
validate_with.append(
    MutuallyExclusive(answer.get('options'), error_messages)
)
```

app/submitter/convert_payload_0_0_1.py
```diff
@@ -26,8 +26,9 @@ def convert_answers_to_payload_0_0_1(answer_store, schema, routing_path):
             value = answer['value']
 
             if answer_schema is not None and value is not None and 'parent_answer_id' not in answer_schema:
-                if 'q_code' in answer_schema and (answer_schema['type'] != 'Checkbox' or any('q_code' not in option for option in
-                                                                                             answer_schema['options'])):
+                if 'q_code' in answer_schema and (
```
nealedj: This is pretty tricky to read. Perhaps break out this conditions into variables to add some meaning?

```
is_single_value_answer = answer_schema['type'] not in ['Checkbox', 'MutuallyExclusiveCheckbox']
all_options_have_q_code = all('q_code' in option for option in answer_schema['options'])
if 'q_code' in answer_schema and (
            is_single_value_answer or not all_options_have_q_code):
    ...
```

I also inverted the `q_code` in options check.

app/validation/validators.py
```diff
@@ -345,3 +345,37 @@ def format_date_string(field):
...
+                separator = ', ' if index < len(values) - 1 else ' and '
+                non_exclusives = non_exclusives + separator
+
+            non_exclusives = non_exclusives + '"' + self.value_labels[additional] + '"'
+
+        return non_exclusives
+
+    @staticmethod
+    def _get_value_labels(options):
+        value_labels = {}
```
nealedj: Worth considering using a dict comprehension for this?

```
return {option['value']: option['label'] for option in options}
```

app/validation/validators.py
```diff
@@ -345,3 +345,37 @@ def format_date_string(field):
...
+
+class MutuallyExclusive(object):
+    def __init__(self, options, messages=None):
+        self.exclusive_value = options[-1]['value']
+        self.value_labels = self._get_value_labels(options)
+        self.messages = messages or error_messages
+
+    def __call__(self, form, field):
+        values = field.data
+        if (len(values) > 1) and (self.exclusive_value in values):
```
nealedj: Neither sets of these brackets are really necessary.

app/validation/validators.py
```diff
@@ -345,3 +345,37 @@ def format_date_string(field):
...
+            non_exclusives = self._build_non_exclusives(values.copy())
+            raise validators.ValidationError(self.messages['MUTUALLY_EXCLUSIVE'] %
+                                             dict(non_exclusives=non_exclusives, exclusive=self.value_labels[self.exclusive_value]))
+
+    def _build_non_exclusives(self, values):
+        values.remove(self.exclusive_value)
```
nealedj: Instead of copying 'values', then passing it, and then removing the `exclusive_value` - consider just checking if the value in the loop.

Or even check it before the iteration in a generator:

```
non_exclusive_values = (value for value in values if value != self.exclusive_value)
for index, additional in enumerate(non_exclusive_values):
    ...
```

app/validation/validators.py
```diff
@@ -345,3 +345,37 @@ def format_date_string(field):
...
+        non_exclusives = ''
+        for index, additional in enumerate(values):
```
nealedj: Might be able to reduce this logic to:

```
non_exclusive_values = [
    '"{}"'.format(self.value_labels[value])
    for value in values
    if value != self.exclusive_value
]
non_exclusives = "{} and {}".format(", ".join(non_exclusive_values[:-1]),  non_exclusive_values[-1])
```

tests/app/forms/test_fields.py
```diff
@@ -304,12 +304,61 @@ def test_checkbox_field(self):
...
+            ],
+            'type': 'MutuallyExclusiveCheckbox'
+        }
+
+        unbound_field = get_field(checkbox_json, checkbox_json['label'], error_messages, self.answer_store,
+                                  self.metadata)
+
+        expected_choices = [(option['value'], option['label']) for option in checkbox_json['options']]
+
+        self.assertTrue(unbound_field.field_class == SelectMultipleField)
```
nealedj: Why not `assertEqual`?

data/en/test_checkbox_mutually_exclusive.json
```diff
@@ -0,0 +1,96 @@
+{
```
nealedj, line 1: 👍 to **not** calling this mutually_exclusive_checkbox.json

nealedj (changes requested):

app/templates/partials/widgets/multiple_choice_widget.html
```diff
@@ -35,12 +40,20 @@
...
-        {{option.label.text}}
+          {% if not exclusive or (exclusive and not loop.last) %}
+            {{option.label.text}}
+          {% endif %}
         {% if answer.options[loop.index0].description %}
           <br />
           <span class="label__description label__inner pluto">{{answer.options[loop.index0].description}}</span>
         {% endif %}
+          {% if exclusive and loop.last %}
+            <span class="u-vh">Or,</span> {{option.label.text}}<span class="u-vh">. Selecting this will uncheck all other checkboxes</span>
```
nealedj: Space could probably be within the `<span u-vh`. `<span class="u-vh">Or, </span>{{option.label.text}}`

nealedj (changes requested):

tests/app/forms/test_fields.py
dwyeradam: That was a copy and paste without thinking. Changed quite a few additional ones in the rest of the file while i'm at it.

data/en/test_checkbox_mutually_exclusive.json
dwyeradam: I don't think it reads as well but I also like that it's testing a checkbox that happens to be mutually exclusive. It's far better to bunch test schemas together based on the feature/components it's for at the loss of a nicer name.

app/validation/validators.py
dwyeradam: The `non_exclusives` wouldn't work where only one non-exclusive value is selected but I do prefer the building of `non_exclusive_values`. I'll find some middle ground.

app/validation/validators.py
```diff
@@ -345,3 +345,34 @@ def format_date_string(field):
     return '{}-{:02d}-{:02d}'.format(int(field.data['year']),
                                      int(field.data['month']),
                                      day)
+
+
+class MutuallyExclusive(object):
```
davecharles: Unless we still plan to run with python 2.x there's no reason to derive from `object` in python 3.x

app/validation/validators.py
```diff
@@ -345,3 +345,34 @@ def format_date_string(field):
...
+    def __init__(self, options, messages=None):
+        self.exclusive_value = options[-1]['value']
```
davecharles: Assuming we're guaranteed a list of options here if schema valid and no need to guard against `options==[]`

data/en/test_checkbox_mutually_exclusive.json
```diff
@@ -0,0 +1,96 @@
...
+                                    {
+                                        "label": "Other",
+                                        "q_code": "6",
+                                        "description": "Choose any other topping",
+                                        "value": "Other",
+                                        "child_answer_id": "other-answer-mandatory"
+                                    },
+                                    {
+                                        "label": "No extra toppings",
+                                        "value": "None",
```
davecharles: May value of 'No extra toppings` for consistency?

davecharles (changes requested):

app/validation/validators.py
dwyeradam: Yes, otherwise it's not a valid checkbox and wouldn't get past validation.

data/en/test_checkbox_mutually_exclusive.json
dwyeradam: I wanted to differentiate it from the label as if they are different (allowed) it's important that the error message shows the label. The validation however is done on the value. I'll change the name from None though as that's confusing during debug.

dwyeradam:
@nealedj and @davecharles ready for re-review.

app/templates/partials/widgets/multiple_choice_widget.html
```diff
@@ -35,12 +40,20 @@
...
       <label {{label|xmlattr}}>
-        {{option.label.text}}
+          {% if not exclusive or (exclusive and not loop.last) %}
+            {{option.label.text}}
+          {% endif %}
         {% if answer.options[loop.index0].description %}
           <br />
           <span class="label__description label__inner pluto">{{answer.options[loop.index0].description}}</span>
         {% endif %}
+          {% if exclusive and loop.last %}
```
shahi645, line 50: @dwyeradam A lot of logic being put in the template, similar to what we talked about the other day, are we wanting it to stay like this, or maybe look to addd a jinja filter?

app/validation/validators.py
```diff
@@ -345,3 +345,34 @@ def format_date_string(field):
     return '{}-{:02d}-{:02d}'.format(int(field.data['year']),
                                      int(field.data['month']),
                                      day)
+
+
+class MutuallyExclusive():
```
shahi645: Even though its pretty self explanatory, would be nice to have a doc string here, like some of the others do

data/en/test_checkbox_mutually_exclusive.json
```diff
@@ -0,0 +1,96 @@
...
+        "id": "default-section",
+        "groups": [{
+            "blocks": [{
+                    "type": "Question",
+                    "id": "mandatory-checkbox",
+                    "description": "",
+                    "questions": [{
+                        "answers": [{
+                                "id": "mandatory-checkbox-answer",
+                                "label": "",
```
shahi645: I think Label (or legend) isn't needed here, it will use the question title

data/en/test_checkbox_mutually_exclusive.json
```diff
@@ -0,0 +1,96 @@
...
+            "validator": "string"
+        }
+    },
+    "sections": [{
+        "id": "default-section",
+        "groups": [{
+            "blocks": [{
+                    "type": "Question",
+                    "id": "mandatory-checkbox",
+                    "description": "",
```
shahi645: drop description

data/en/test_checkbox_mutually_exclusive.json
```diff
@@ -0,0 +1,96 @@
...
+                    "questions": [{
+                        "answers": [{
+                                "id": "mandatory-checkbox-answer",
+                                "label": "",
+                                "mandatory": true,
+                                "options": [{
+                                        "label": "Cheese",
+                                        "value": "Cheese",
+                                        "q_code": "1"
```
shahi645, line 34: probably get rid of q_codes and change to data_version 0.0.2 for simplicity

app/validation/error_messages.py
```diff
@@ -26,4 +26,5 @@
     'DATE_PERIOD_TOO_LARGE': 'Enter a reporting period less than or equal to %(max)s.',
     'SINGLE_DATE_PERIOD_TOO_EARLY': 'Enter a date after %(min)s.',
     'SINGLE_DATE_PERIOD_TOO_LATE': 'Enter a date before %(max)s.',
+    'MUTUALLY_EXCLUSIVE': 'Uncheck %(non_exclusives)s or "%(exclusive)s" to continue',
```
shahi645, line 29: In the original trello car, this says `Deselect` ... https://trello.com/c/N74J2IcB/1775-mutually-exclusive-check-boxes

shahi645 (changes requested):

app/validation/error_messages.py
shahi645: Ahh the AC however says Uncheck so ignore this

shahi645 (approved):
just a bunch of tiny suggestions, but the everything else looks good.

data/en/test_checkbox_mutually_exclusive.json
dwyeradam: I've added the q-codes as this PR required a change to convert_payload_0_0_1.py.

davecharles (approved):

app/templates/partials/widgets/multiple_choice_widget.html
dwyeradam: Slightly different context I think. The logic is around wether something should be shown on the page or not. A filter would be used to change what is show not decide if it should be shown or not. We did consider having a separate widget for mutually exclusive but this would have led to a lot of duplication so a bit of logic here seemed the best option.

app/validation/validators.py
```diff
@@ -14,7 +14,7 @@
 logger = get_logger()
 
 
-class NumberCheck(object):
+class NumberCheck():
```
nealedj: Now this _really_ is a nit pick.

You don't need the `()` to define a class that inherits directly from `object`.

```
class NumberCheck:
   ...
```

app/templates/partials/widgets/multiple_choice_widget.html
nealedj: Good point to raise. I agree that this is probably a fair balance in this case though.

If the template were to become more complex we could look to refactor later but as it is I don't think it's holding too much logic nor does it take responsibility for any logic that it shouldn't be.

app/validation/validators.py
nealedj: Good compromise 👍

nealedj (approved):
I added one nitpick comment that you could address before you squash if you feel like it. But not change-request worthy.

Issue #1613, opened 2018-06-04
Fix metadata validation for trad_as_or_ru_name


dwyeradam:
### Expected behaviour
If you pass in empty values for both Trad As and Ru Name it should throw a 403 InvlaidTokenException.

### Actual behaviour
If you pass in empty values for both Trad As and Ru Name it will still pass validation. The current validation only checks the keys for this and not the value.

### Steps to reproduce the behaviour
Delete Trad As and RU Name from go launcher and launch any business survey.

### Technical information
This [line](https://github.com/ONSdigital/eq-survey-runner/blob/3adc7490ae100ac3fc37897da57dc5543cc77cb7/app/storage/metadata_parser.py#L81) in metadata_parser.py should also check the key has a value. 
`valid = bool(metadata.get('trad_as')) or bool(metadata.get('ru_name'))`

PR #1614, opened 2018-06-04
Fix validation for trad_as or ru_name

MebinAbraham:
Fixes [1613](https://github.com/ONSdigital/eq-survey-runner/issues/1613)

### DIFF
--- a/app/storage/metadata_parser.py
+++ b/app/storage/metadata_parser.py
@@ -78,7 +78,7 @@ def _validate_mandatory_metadata(metadata, required_metadata):
             valid = 'variant_flags' in metadata
         elif key == 'trad_as_or_ru_name':
             # either of 'trad_as' or 'ru_name' is required
-            valid = 'trad_as' in metadata or 'ru_name' in metadata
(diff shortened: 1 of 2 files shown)


shahi645 (approved):

app/storage/metadata_parser.py
```diff
@@ -78,7 +78,7 @@ def _validate_mandatory_metadata(metadata, required_metadata):
             valid = 'variant_flags' in metadata
         elif key == 'trad_as_or_ru_name':
             # either of 'trad_as' or 'ru_name' is required
-            valid = 'trad_as' in metadata or 'ru_name' in metadata
+            valid = bool(metadata.get('trad_as') or metadata.get('ru_name'))
```
dwyeradam, line 81: Looks good but we should probably have a test that returns false.

dwyeradam (changes requested):

app/storage/metadata_parser.py
shahi645: good point

MebinAbraham: Sorted 👍

dwyeradam (approved):

PR #1615, opened 2018-06-04
RU Name mandatory in metadata for Default/NI Themes

dwyeradam:
The introduction page expects RU Name to be present for `default` and `northernireland` themes. 

This PR adds ru_name as required metadata in the required schemas.


scripts/test_schemas.sh
```diff
@@ -6,8 +6,8 @@ if [ "$1" == "--local" ] || [ "$2" == "--local" ]; then
 fi
 
 if [ "$run_docker" == true ]; then
-    docker pull onsdigital/eq-schema-validator
-    validator="$(docker run -d -p 5001:5000 onsdigital/eq-schema-validator)"
```
dwyeradam: Will need reverting once [validator PR](https://github.com/ONSdigital/eq-schema-validator/pull/55) has been approved. Both should go in together.

MebinAbraham (changes requested):
Shouldn't need this [line](https://github.com/ONSdigital/eq-survey-runner/blob/eq-2164-mandatory-metadata-based-on-theme/app/templating/schema_context.py#L58) anymore. As it would be created on line 51 for the schemas that require it.

MebinAbraham (approved):
👍

shahi645 (approved):

PR #1620, opened 2018-06-07
Remove group_id and block_id from Answers

davecharles:
All answers with the exception of those of type `RepeatingAnswer` store the group_id and block_id against them.  This is inconsistent and we don't actually need these attributes on the answers anymore.

This PR removes storage of group and block ID attrs on an answer, and updates all the tests that did the same. 

### How to review 
- Run the tests.
- Run a survey and  perform a dump submission e.g. using a url like: `http://localhost:5000/dump/submission`.
- Examine the results and verify the answers no longer have group and block id attrs, e.g. look something like this:
```
{
  "answers": [
    {
      "answer_id": "address-check-answer", 
      "answer_instance": 0, 
      "group_instance": 0, 
      "value": "Yes"
    },
```
And not like this:

```
{
  "answers": [
    {
      "answer_id": "address-check-answer", 
      "answer_instance": 0, 
      "block_id": "address-check-block", 
      "group_id": "about-household-group", 
      "group_instance": 0, 
      "value": "Yes"
    },
```

### DIFF
--- a/app/data_model/answer_store.py
+++ b/app/data_model/answer_store.py
@@ -5,13 +5,10 @@ from jinja2 import escape
 import simplejson as json
 
 
-class Answer(object):
-    def __init__(self, answer_id, value, group_instance=0, answer_instance=0, group_id=None, block_id=None):
+class Answer:
+    def __init__(self, answer_id, value, group_instance=0, answer_instance=0):
         if answer_id is None or value is None:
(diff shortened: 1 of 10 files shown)


dwyeradam (approved):

shahi645 (approved):

shahi645:
@davecharles can be rebased and squashed

PR #1621, opened 2018-06-07
Year Entry Field

dwyeradam:
Format YYYY only
Includes Min/Max on single entry
Includes Min/Max on date range

As a user
I want to be able to enter a date in a year only format
So that I can give the correct information required by the survey

### How to review 
Test schema: `test_date_validation_yyyy_combined.json`

### DIFF
--- a/app/forms/date_form.py
+++ b/app/forms/date_form.py
@@ -7,8 +7,9 @@ from dateutil.relativedelta import relativedelta
 
 from wtforms import Form, SelectField, StringField
 
+from app.forms.custom_fields import CustomIntegerField
 from app.questionnaire.rules import get_metadata_value, get_answer_store_value, convert_to_datetime
-from app.validation.validators import DateCheck, OptionalForm, DateRequired, MonthYearCheck, SingleDatePeriodCheck
+from app.validation.validators import DateCheck, OptionalForm, DateRequired, MonthYearCheck, YearCheck, SingleDatePeriodCheck
 
 logger = logging.getLogger(__name__)
 
@@ -30,19 +31,12 @@ def get_date_form(answer_store, metadata, answer=None, error_messages=None):
 
     validate_with = [OptionalForm()]
 
-    if error_messages:
-        date_messages = error_messages.copy()
-    else:
-        date_messages = {}
-
     if answer['mandatory'] is True:
         validate_with = validate_mandatory_date(error_messages, answer)
 
-    if 'validation' in answer and 'messages' in answer['validation'] \
-            and 'INVALID_DATE' in answer['validation']['messages']:
-        date_messages['INVALID_DATE'] = answer['validation']['messages']['INVALID_DATE']
+    error_message = get_bespoke_message(answer, 'INVALID_DATE') or error_messages['INVALID_DATE']
 
-    validate_with.append(DateCheck(date_messages['INVALID_DATE']))
+    validate_with.append(DateCheck(error_message))
 
     if 'minimum' in answer or 'maximum' in answer:
         min_max_validation = validate_min_max_date(answer, answer_store, metadata, 'yyyy-mm-dd')
@@ -71,12 +65,8 @@ def get_month_year_form(answer, answer_store, metadata, error_messages):
     if answer['mandatory'] is True:
         validate_with = validate_mandatory_date(error_messages, answer)
 
-    if 'validation' in answer and 'messages' in answer['validation'] \
-            and 'INVALID_DATE' in answer['validation']['messages']:
-        error_message = answer['validation']['messages']['INVALID_DATE']
-        validate_with.append(MonthYearCheck(error_message))
-    else:
-        validate_with.append(MonthYearCheck())
+    error_message = get_bespoke_message(answer, 'INVALID_DATE')
+    validate_with.append(MonthYearCheck(error_message))
 
     if 'minimum' in answer or 'maximum' in answer:
         min_max_validation = validate_min_max_date(answer, answer_store, metadata, 'yyyy-mm')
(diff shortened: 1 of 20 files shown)


scripts/test_schemas.sh
```diff
@@ -6,8 +6,8 @@ if [ "$1" == "--local" ] || [ "$2" == "--local" ]; then
 fi
 
 if [ "$run_docker" == true ]; then
-    docker pull onsdigital/eq-schema-validator
-    validator="$(docker run -d -p 5001:5000 onsdigital/eq-schema-validator)"
+    docker pull onsdigital/eq-schema-validator:eq-2089-year-entry-field
+    validator="$(docker run -d -p 5001:5000 onsdigital/eq-schema-validator:eq-2089-year-entry-field)"
     sleep 3
```
dwyeradam: Revert when [validator PR](https://github.com/ONSdigital/eq-schema-validator/pull/57) is approved.

app/forms/date_form.py
```diff
@@ -87,6 +88,44 @@ class MonthYearDateForm(Form):
...
+
+    class YearDateForm(Form):
+        year = None
+
+    validate_with = [OptionalForm()]
+
+    if answer['mandatory'] is True:
+        validate_with = validate_mandatory_date(error_messages, answer)
+
+    if 'validation' in answer and 'messages' in answer['validation'] \
```
nealedj: There's a few instances of this code littered around this module. Perhaps consider some way of abstracting the ability to fetch an error message with given key from an `answer` dict?

app/forms/date_form.py
```diff
@@ -113,6 +152,10 @@ def validate_min_max_date(answer, answer_store, metadata, date_format):
         display_format = '%B %Y'
         minimum_date = minimum_date.replace(day=1) if minimum_date else None    # First day of Month
         maximum_date = maximum_date + relativedelta(day=31) if maximum_date else None   # Last day of month
+    elif date_format == 'yyyy':
+        display_format = '%Y'
+        minimum_date = minimum_date.replace(month=1, day=1) if minimum_date else None    # January 1st
+        maximum_date = maximum_date + relativedelta(month=12, day=31) if maximum_date else None   # Last day of december
```
nealedj: Could this not be `maximum_date = maximum_date.replace(month=12, day=31)`?

I don't think there's anything relative about this date calculation.

app/jinja_filters.py
```diff
@@ -23,7 +23,7 @@
 @blueprint.app_template_filter()
 def format_number(value):
     if value or value is 0:
-        return numbers.format_number(value, locale=DEFAULT_LOCALE)
+        return numbers.format_decimal(value, locale=DEFAULT_LOCALE)
```
dwyeradam, line 26: Full transparency. This isn't related to this PR but seemed too small for a PR in it's own right. I noticed during testing that format_number has now been deprecated by Babel and format_decimal is to be used in it's place.

data/en/test_dates.json
```diff
@@ -113,4 +126,4 @@
             "title": ""
         }]
     }]
-}
\ No newline at end of file
+}
```
nealedj: `yarn format`

app/forms/date_form.py
```diff
@@ -56,7 +57,7 @@ class DateForm(Form):
 
 def get_month_year_form(answer, answer_store, metadata, error_messages):
     """
-    Returns a month year form metaclass with appropriate validators. Used in both date and
+    Returns a year form metaclass with appropriate validators. Used in both date and
```
shahi645: @dwyeradam This has been changed wrongly?

app/forms/date_form.py
```diff
@@ -87,6 +88,44 @@ class MonthYearDateForm(Form):
     return MonthYearDateForm
 
 
+def get_year_form(answer, answer_store, metadata, error_messages, label, guidance):
+    """
+    Returns a month year form metaclass with appropriate validators. Used in both date and
```
shahi645: this should be year form

app/forms/date_form.py
```diff
@@ -87,6 +88,44 @@ class MonthYearDateForm(Form):
...
+    class YearDateForm(Form):
+        year = None
+
+    validate_with = [OptionalForm()]
+
+    if answer['mandatory'] is True:
+        validate_with = validate_mandatory_date(error_messages, answer)
+
+    if 'validation' in answer and 'messages' in answer['validation'] \
+            and 'INVALID_DATE' in answer['validation']['messages']:
```
shahi645: This code is used in each `get_form` list, probably worth putting in a method at bottom and call, `is_bespoke_message` or something

app/forms/date_form.py
```diff
@@ -87,6 +88,44 @@ class MonthYearDateForm(Form):
...
+        error_message = answer['validation']['messages']['INVALID_DATE']
+        validate_with.append(YearCheck(error_message))
+    else:
+        validate_with.append(YearCheck())
+
+    if 'minimum' in answer or 'maximum' in answer:
```
shahi645: same as above comment, maybe `has_period_limits`

app/jinja_filters.py
shahi645: good catch

shahi645 (changes requested):
Looks good and sleek so far, just left some refactoring comments

app/forms/date_form.py
dwyeradam: oops, noticed after and deleted the wrong one.

app/validation/validators.py
```diff
@@ -201,6 +206,16 @@ def __call__(self, form, field):
             raise validators.ValidationError(self.message)
 
 
+class YearCheck:
+    def __init__(self, message=None):
+        self.message = message or error_messages['INVALID_DATE']
+
+    def __call__(self, form, field):
+        # Must be 4 digits long
```
nealedj, line 217: This is fine unless we ever want to allow people to enter dates between 0 and 999 AD... which I suspect isn't a requirement 👍

app/forms/date_form.py
dwyeradam: I've put the bespoke message get into it's own method but i'll leave this as it is. Calling it's own method wouldn't actually reduce the lines of code as I would still need to evaluate an if on the return of that method.

app/validation/validators.py
```diff
@@ -171,9 +173,12 @@ def __call__(self, form, field):
         if hasattr(form, 'day'):
```
nealedj: Consider a refactor to avoid numerous points at which the same exception is raised:

```
    invalid_year = not form.year.data
    invalid_month = hasattr(form, 'month') and not form.month.data
    invalid_day = hasattr(form, 'day') and not form.day.data

    if invalid_year or invalid_month or invalid_year:
```

app/forms/date_form.py
AndrewTorrance: Not all params in DocString

nealedj (changes requested):

app/forms/date_form.py
AndrewTorrance: Not all params in doc string

app/forms/date_form.py
dwyeradam: Yes, sorry it should be, I don't know what happened there. Looks like a cut and paste without thinking from the mm-yyyy equivalent.

app/validation/validators.py
dwyeradam: To be honest I was tempted to ask if I could limit it to dates after 1900 but chose against that.

shahi645 (approved):
Happy from my POV, some changes to be considered by Adam

app/forms/date_form.py
```diff
@@ -87,16 +77,55 @@ class MonthYearDateForm(Form):
...
-    if 'validation' in answer and 'messages' in answer['validation']:
-        if 'MANDATORY_DATE' in answer['validation']['messages']:
-            error_message = answer['validation']['messages']['MANDATORY_DATE']
+    error_message = get_bespoke_message(answer, 'MANDATORY_DATE') or error_messages['MANDATORY_DATE']
 
     validate_with = [DateRequired(message=error_message)]
     return validate_with
 
 
+def get_bespoke_message(answer, message_type):
```
nealedj, line 121: Nice 👍

scripts/test_schemas.sh
nealedj: Can be removed now

app/validation/validators.py
dwyeradam: @nealedj So this refactor wouldn't quite work. This check is to check you've entered at least one field in a mandatory date. Not if it's a valid date. The DateCheck below checks the format. I've come up with a cleaner way to do this though so will push that soon.

nealedj (approved):

PR #1625, opened 2018-06-11
Added trading as to thank you and view submission

AndrewTorrance:
There is a requirement to show trading as on the thank you and view submission page . 

### How to review 
Launch any survey from go launcher with 2 test cases 1 with a trading as string set to something different to ru , and with zero length string as trading as . if there is a trading as it will be displayed on the thank you page after ru name in parenthesis . If there is an empty string then there should be no 
parenthesis. Repeat for a survey that show view submission and validate same cases in the view submission page. e.g test_view_submitted_response

### DIFF
--- a/app/data_model/session_data.py
+++ b/app/data_model/session_data.py
@@ -1,4 +1,4 @@
-class SessionData(object):      # pylint: disable=too-many-locals
+class SessionData(object):
 
     def __init__(self,
                  tx_id,
@@ -14,7 +14,7 @@ class SessionData(object):      # pylint: disable=too-many-locals
                  account_service_url=None,
                  submitted_time=None,
                  trad_as=None,
-                 **_):
+                 **_):                  # pylint: disable=too-many-locals
         self.tx_id = tx_id
         self.eq_id = eq_id
         self.form_type = form_type
--- a/app/data_model/session_data.py
+++ b/app/data_model/session_data.py
@@ -1,4 +1,4 @@
-class SessionData(object):
+class SessionData(object):      # pylint: disable=too-many-locals
 
     def __init__(self,
                  tx_id,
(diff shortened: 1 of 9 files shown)


tests/app/data_model/test_session_store.py
```diff
@@ -203,10 +203,10 @@ def __init__(self,
                 self.case_ref = case_ref
                 self.account_service_url = account_service_url
 
-        def old_json_loader(raw, object_hook):
+        def old_json_loader(raw, object_hook):  # pylint: disable=unused-argument
             """replacement for json.loads to decode to old format ( no trading as) """
 
-            old_data = original_loads_func(raw, object_hook=lambda d: OldSessionData(**d))   # call original json loads but with a hook pointing to an old class
+            old_data = original_loads_func(raw, object_hook=lambda d: OldSessionData(**d))   # call json.loads ,hook pointing to an old class
```
shahi645: remove space before comma, and add one after

app/data_model/session_data.py
```diff
@@ -14,7 +14,7 @@ def __init__(self,
                  account_service_url=None,
                  submitted_time=None,
                  trad_as=None,
-                 **_):
+                 **_):                  # pylint: disable=too-many-locals
```
shahi645: Curious question...are these all required for session data?

shahi645 (approved):
Tested and verified. left a couple of minor questions and I think the `trading_as` name looks a bit strange not being bold, but the design matches

app/data_model/session_data.py
AndrewTorrance: They all look as if they are useful. But I have not tracked the usage of all the params .

davecharles:
LGTM

davecharles (approved):
LGTM

PR #1626, opened 2018-06-11
Don't show checkbox as a list on summary if only one box is checked

kierenbeckett:
On the summary page for a set of checkboxes we no longer want to show it as a list if only one box is checked

### How to review 
Check one item, see that there's not a bullet point next to it, check two items, see that there is a bullet point next to each.

### DIFF
--- a/app/templates/partials/summary/checkbox.html
+++ b/app/templates/partials/summary/checkbox.html
@@ -1,18 +1,26 @@
 {% set not_answered = 'No answer provided' %}
-<ul class="u-m-no">
-      {% for selected_answer in answer.value %}
-        <li>{{selected_answer.label}}
-            {%- if answer.child_answer_value is not none and selected_answer.should_display_other %}
-            <ul>
-                <li>
-                    {%- if answer.child_answer_value == ''  -%}
-                        {{not_answered}}
-                    {%- else -%}
-                        {{ answer.child_answer_value }}
-                    {%- endif -%}
-                </li>
-            </ul>
+
+{% macro answer_with_child(selected_answer) %}
+    {{selected_answer.label}}
+    {%- if answer.child_answer_value is not none and selected_answer.should_display_other %}
+    <ul>
+        <li>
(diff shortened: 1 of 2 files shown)


shahi645 (approved):
LGTM, using a lot of logic in the template though, might want to consider using a jinja filter to handle this

AndrewTorrance (changes requested):
All thats there is ok,  but I think it needs a test ?

nealedj:
Change looks reasonable. I don't think this case warrants a jinja filter. 👍 to making sure this is covered by a functional test.

kierenbeckett:
@shahi645 cool, will do.

@AndrewTorrance I've added a couple of tests

app/templates/partials/summary/checkbox.html
```diff
@@ -1,18 +1,34 @@
...
+                        {%- endif -%}
+                    </li>
+                </ul>
+                {%- endif -%}
+            </li>
+          {% endfor %}
+    </ul>
+{%- else -%}
+    {% set selected_answer = answer.value[0] %}
+    {{selected_answer.label}}
```
jonnyshaw89: This block seems to be duplicated? can we set this as a block and reuse?

jonnyshaw89 (changes requested):

app/templates/partials/summary/checkbox.html
shahi645: i tried fiddling with this yday, and could't get it to work perfectly, unless we use a filter but yeah i agree if possible we should do.

kierenbeckett: Couldn't set this as a block as the duplicated section contains a local variable.  I'm not sure a filter is quite the right tool either as it's mostly display logic.  I've factored it out into a macro.

shahi645 (approved):
Im good with this. Just tested and works fine

jonnyshaw89 (approved):

AndrewTorrance (approved):

PR #1629, opened 2018-06-12
Apply new summary design to surveys

MebinAbraham:
Updates pattern library to [v1.2.0](https://github.com/ONSdigital/sdc-global-design-patterns/releases/tag/v1.2.0) and fixes the breaking change to summary pages.
[Card](https://trello.com/c/KCYOLcBk/1297-apply-new-summary-design-surveys-without-sections-m) for context.
[Summary Design - Pattern Library](https://sdc-global-design-patterns.netlify.com/components/detail/summary.html)
This takes in all changes from pattern library since 5th of March till `v1.2.0`

### How to review 
1. Make sure that summary looks and behaves as expected.
2. Ensure the templating is correct.
3. Ensure all assets and styles are as expected.

### DIFF
--- a/app/jinja_filters.py
+++ b/app/jinja_filters.py
@@ -67,8 +67,7 @@ def format_multilined_string(context, value):
     escaped_value = escape(value)
     new_line_regex = r'(?:\r\n|\r|\n)+'
     value_with_line_break_tag = re.sub(new_line_regex, '<br>', escaped_value)
-    result = '<p>{}</p>'.format(value_with_line_break_tag)
-
+    result = '{}'.format(value_with_line_break_tag)
     return mark_safe(context, result)
 
 
@@ -168,6 +167,18 @@ def format_household_member_name_possessive(names):
         return name + '’'
 
     return name + '’s'
+
+
+@blueprint.app_template_filter()
+def answer_is_type(answer, target_answer_type):
+    """
+    :param answer:
+    :param target_answer_type:
+    :return: true if the answer type matches given input
+    'RepeatingAnswer' question type return 'answers' as an object, and dict for all other question types.
+    """
+    answer_type = answer['type'] if isinstance(answer, dict) else answer.type
+    return answer_type == target_answer_type.lower()
 
 
 @evalcontextfilter
@@ -272,6 +283,11 @@ def get_question_title_processor():
 @blueprint.app_context_processor
 def get_answer_label_processor():
     return dict(get_answer_label=get_answer_label)
+
+
+@blueprint.app_context_processor
+def answer_is_type_processor():
+    return dict(answer_is_type=answer_is_type)
 
 
 @blueprint.app_context_processor
--- a/app/templates/partials/summary/summary.html
+++ b/app/templates/partials/summary/summary.html
@@ -1,53 +1,24 @@
-{% set not_answered = 'No answer provided' %}
+<div class="summary">
+	{%- for group in content.summary.groups -%}
+        {%- if group.title -%}
+            <h2 class="summary__title neptune" id="{{ group.id }}">{{ group.title }}</h2>
+        {%- endif %}
 
-<div>
-  <div class="summary">
+        <div class="summary__block u-mb-l">
+            {%- for block in group.blocks -%}
+                {%- for question in block.questions -%}
 
-    {% for group in content.summary.groups %}
-      {% if group.title %}
-        <h2 class="summary__title saturn" id="{{group.id}}">{{group.title}}</h2>
-      {% endif %}
-      <div class="summary__block">
+                    {% set has_multiple_answers = question.answers|length > 1 %}
 
-        {% for block in group.blocks %}
-
-          <div class="summary__items">
-
-            {% for question in block.questions %}
-
-              <div class="summary__question" id="{{question.id}}" data-qa="summary-question-title">
-                {{question.number or ''}} {{question.title|striptags}}
-              </div>
-
-              {% for answer in question.answers %}
-                {% if question.answers|length > 1 and answer.label %}
-                  <div class="summary__question summary__question--sub">
-                    {{answer.label|striptags}}
-                  </div>
-                {% endif %}
-
-                <div class="summary__answer">
-                  <div class="summary__answer-text" id="{{answer.id}}-answer" data-qa="{{answer.id}}-answer">
-                    {%- if answer.value is none or answer.value == '' -%}
-                      {{not_answered}}
+                    {%- if has_multiple_answers -%}
+                        {%- include theme(['partials/summary/question-multiple-answers.html']) -%}
                     {%- else -%}
-                      {% include theme(['partials/summary/' ~ answer.type ~ '.html', 'partials/summary/default.html']) %}
-                     {%- endif -%}
-                   </div>
+                        {%- include theme(['partials/summary/question-single-answer.html']) -%}
+                    {%- endif -%}
 
-                  {% if content.summary.answers_are_editable %}
-                    <div class="summary__edit">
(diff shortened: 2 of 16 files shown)


shahi645:
Couple of tests failing

app/jinja_filters.py
```diff
@@ -170,6 +170,13 @@ def format_household_member_name_possessive(names):
     return name + '’s'
 
 
+@blueprint.app_template_filter()
+def has_answer_type(questions, answer_type):
+    if questions.get('answers'):
+        return any(answer for answer in questions['answers'] if answer['type'] == answer_type.lower())
```
shahi645: Good use of any 👍

app/templates/partials/summary/summary.html
```diff
@@ -1,53 +1,59 @@
...
+                    <div class="summary__item {{ 'summary__item--multiple' if has_multiple_answers }}">
+                        <div class="grid">
+                            <div class="grid__col {{ 'col-12@m' if has_multiple_answers else 'col-12@xs' }} {{ 'col-10@m' if has_textarea else 'col-6@m' }}">
+                                <div class="summary__question mars" id="{{ question.id }}" data-qa="summary-question-title">{{ question.number or ''}} {{question.title|striptags }}</div>
+                            </div>
+                    {{ '</div>' if has_multiple_answers }}
+
+                    {% for answer in question.answers %}
+
+                        {% set is_textarea = has_answer_type(answer, 'textarea') %}
```
shahi645: this is confusing, I'm guessing your just trying to get the answer_type retuned when it matches 'textarea'. In which case, should update the Jinja filter input to say `question_or_answer`, and add a docstring to explain

shahi645 (changes requested):
Looks good, just small changes

app/templates/partials/summary/summary.html
MebinAbraham: Will separate this into 2 functions to make it clear what it is doing.

app/jinja_filters.py
```diff
@@ -170,6 +170,24 @@ def format_household_member_name_possessive(names):
     return name + '’s'
 
 
+@blueprint.app_template_filter()
+def question_has_answer_type(question, answer_type):
+    # returns true if any answer type matches input
```
shahi645: @MebinAbraham use the docstrings for consistency with rest of file

app/jinja_filters.py
```diff
@@ -170,6 +170,32 @@ def format_household_member_name_possessive(names):
...
+@blueprint.app_template_filter()
+def question_has_answer_type(question, answer_type):
+    """
+    :param question:
+    :param answer_type:
+    :return: true if any answer type matches given input
+    'RepeatingAnswer' question type return 'answers' as an object, and dict for all other question types.
+    """
+    if question['type'] == 'RepeatingAnswer':
+        return any(answer for answer in question['answers'] if answer.type == answer_type.lower())
```
nealedj: Maybe put `answer_type = answer_type.lower()` at the top of the function so you don't have to call it every time you iterate through each answer.

app/jinja_filters.py
```diff
@@ -170,6 +170,32 @@ def format_household_member_name_possessive(names):
...
+
+@blueprint.app_template_filter()
+def answer_is_type(answer, answer_type):
+    """
+    :param answer:
+    :param answer_type:
+    :return: true if the answer type matches given input
+    'RepeatingAnswer' question type return 'answers' as an object, and dict for all other question types.
+    """
+    if not isinstance(answer, dict):
```
nealedj: Another way to skin it:

```
def answer_is_type(answer, target_answer_type):
    answer_type = answer['type'] if isinstance(answer, dict) else answer.type
    return answer_type == target_answer_type.lower()
```

app/templates/partials/summary/summary.html
```diff
@@ -1,53 +1,59 @@
...
+
+                    {% set has_textarea = question_has_answer_type(question, 'textarea') %}
+                    {% set has_multiple_answers = question.answers|length > 1 %}
+
+                    <div class="summary__item {{ 'summary__item--multiple' if has_multiple_answers }}">
+                        <div class="grid">
+                            <div class="grid__col {{ 'col-12@m' if has_multiple_answers else 'col-12@xs' }} {{ 'col-10@m' if has_textarea else 'col-6@m' }}">
+                                <div class="summary__question mars" id="{{ question.id }}" data-qa="summary-question-title">{{ question.number or ''}} {{question.title|striptags }}</div>
+                            </div>
+                    {{ '</div>' if has_multiple_answers }}
```
nealedj: I'd avoid putting any conditionals on single opening or closing tags. Easy route to ending up with invalid HTML at some point.

I've seen it in lots of PHP... but nowhere else.

app/templates/partials/summary/summary.html
```diff
@@ -1,53 +1,59 @@
...
+
+                    {% for answer in question.answers %}
+
+                        {% set is_textarea = answer_is_type(answer, 'textarea') %}
+                        {% set has_answer = answer.value is none or answer.value == '' %}
```
nealedj: 1. If `answer.value` is none then surely `has_answer` should be false?
2. Is this variable used? If it's setting a variable for the purposes of a child partial template then I'd suggest declaring it as close to its use as possible.

app/templates/partials/summary/summary.html
```diff
@@ -1,53 +1,59 @@
...
+
+                        {% if has_multiple_answers %}
+                            <div class="grid">
+                                <div class="grid__col col-12@xs {{ 'col-10@m' if is_textarea else 'col-6@m' }}">
+                                    <div class="summary__label mars">{{ answer.label|striptags }}</div>
+                                </div>
+                        {% endif %}
```
nealedj: Try and keep the content of each conditional as valid HTML. This intentionally leaves `<div class="grid">` open because it gets closed later - but that's an _really_ easy condition to break.

nealedj (changes requested):
I've pushed a suggested change to `eq-1297-apply-new-summary-design-v2--david-suggestion` for you to take a look at.

Basically I just split the existing logic for rendering questions with multiple answers and those with single answers into separate partials. That removes the use of isolated conditional closing tags.

app/templates/partials/summary/summary.html
nealedj: I addressed this in my branch.

app/templates/partials/summary/summary.html
nealedj: I addressed this in my branch.

app/templates/partials/summary/summary.html
nealedj: I addressed this in my branch.

MebinAbraham: Initial plan was to have the 2 separate partials like you suggested but then tried to minimise duplication but ya I get why this can cause unnecessary headaches 👍

app/templates/partials/summary/summary.html
MebinAbraham: Its existing logic to display necessary answer. Ya it shouldn't be in this file, forgot to remove as its moved code that needed it to `answer.html`

nealedj (changes requested):
Looks good but I don't think there's enough space between each group. On the pattern library there's 2rem between each example summary - worth checking if this was intended to be used between groups also.

summary - section summary_ - http___localhost_5000_questionnaire

MebinAbraham (commented):
@nealedj Spoke with Benn and added a `margin-top` class to group titles to get the same look as `collator-items` is pattern lib specific.

But it would only apply if there is actually a group title.
Will see what Benn says regarding the other cases.

MebinAbraham (commented):
Okay spoke to Benn again and solved the issue. Added `margin-bottom` to space it out regardless of whether titles are present. Removed a redundant class and `p` tags also.

screen shot 2018-06-18 at 13 19 36

shahi645 (approved):
LGTM, everything works and matches design. Probably needs a FE review

bennpearson (approved):
LGTM 👍  First PR using the latest pattern library! 🥇  Nice one @MebinAbraham

nealedj (approved):

nealedj:
@MebinAbraham good to squash

PR #1630, opened 2018-06-13
Bump chromedriver version to work with latest chrome

kierenbeckett:
The chromedriver version set in chimp no longer works against the latest version of chrome


samiwel:
@kierenbeckett we had issues getting chimp to play nicely with Chrome 65 previously. In the end we pinned Chrome to version 34 in the concourse pipeline. See https://github.com/ONSdigital/eq-smoke-tests/pull/12 for context.

shahi645 (commented):
would we need to change any of the travis config for this?

kierenbeckett:
As this is more complicated than I assumed I've created a ticket.

Issue #1627, opened 2018-06-12
Completion status incorrectly cleared


davecharles:
### Expected behaviour
When using question `titles` feature in a repeating group, questions answered for the first repeating group instance should still be marked as complete when you navigate to questions for the next group instance.

### Actual behaviour
When using question `titles` feature in a repeating group, navigating to questions for a second group instance causes answers from a previous group instance to be considered incomplete.

### Steps to reproduce the behaviour
Run the refactored `lms_1.json` survey in eq-2174-lms-proxy-titles branch. Add two occupants, answer questions for the first and notice that completion for first occupant is cleared when navigating to questions for the second occupant.

PR #1631, opened 2018-06-13
Eq 2177 completion status incorrectly cleared

AndrewTorrance:
When a conditional when was evaluated inside a repeating group and the value it based the decision on was from another conditional group then the existing code was removing the data from the first iteration , making it never ending

### How to review 
Run test_titles_conditional_within_repeating_block.json from go launcher and enter 3names . When asked if you are that person say no and the next question should say what is <names>'s religion? On the last one say yes and the religion title should be what is your religion


tests/functional/spec/features/conditional_question_title/conditional_within_repeating_block.spec.js
```diff
@@ -0,0 +1,44 @@
...
+        .click(HouseholdMemberCompletedPage.submit())
+        .click(ProxyCheckPage.proxy())
+        .click(ProxyCheckPage.submit())
+        .getText(ReligionPage.questionText()).should.eventually.contain("What is Mary’s religion?")
+        .click(ReligionPage.jedi())
+        .click(ReligionPage.submit())
+        .click(HouseholdMemberCompletedPage.submit())
+        .click(ProxyCheckPage.noProxy())
+        .click(ProxyCheckPage.submit())
+        .getText(ReligionPage.questionText()).should.eventually.contain("What is your religion?")
```
shahi645, line 42: missing a semi-colon at the end here

shahi645 (dismissed):
Changes look good to me, just tiny error

app/views/questionnaire.py
```diff
@@ -576,7 +576,7 @@ def _return_date_answer_value(answer_value):
     return None
 
 
-def _remove_dependent_answers_from_completed_blocks(answer_id, questionnaire_store):
+def _remove_dependent_answers_from_completed_blocks(answer_id, group_instance, questionnaire_store):
```
dwyeradam, line 580: What happens when an answer within a repeating group is dependent on an answer not in that repeating group?

dwyeradam (changes requested):

tests/functional/spec/features/conditional_question_title/conditional_within_repeating_block.spec.js
```diff
@@ -0,0 +1,44 @@
+const helpers = require('../../../helpers');
+
+describe('@watch Feature: Use of conditional Titles in Repeating blocks with condition dependant on answer changing within block', function() {
```
kierenbeckett: Left a watch annotation on this

kierenbeckett (dismissed):
Just for my understanding, the new test is checking the question text of the repeating block shows the correct name, but that was already working on master?  Whereas this PR fixes the looping issue, but the new test doesn't cover that?

app/views/questionnaire.py
jonnyshaw89: What he said ^^^
We should add a test for this other scenario while we are here.

AndrewTorrance: We test that in test titles within repeating blocks

dwyeradam: Thanks, will have a look.

AndrewTorrance: If not in a repeating group the group instance is 0 .

tests/functional/spec/features/conditional_question_title/conditional_within_repeating_block.spec.js
AndrewTorrance: Not quite . On Master the second iteration will invalidate the answer from the first iteration so we would never get to the third . Am changing the test however to explicitly go right through to the confirmation page

tests/functional/spec/features/conditional_question_title/conditional_within_repeating_block.spec.js
AndrewTorrance: fair'nuff , changing

app/views/questionnaire.py
dwyeradam: @AndrewTorrance sorry the test_titles_within_repeating_blocks.json doesn't cover the scenario i'm concerned with. The dependency here is the `confirm-answer` is dependent on the `behalf-of-answer` which are both within the repeating group.

AndrewTorrance: Added new test

AndrewTorrance:
@kierenbeckett . The reason it works on Master is that master is not using titles , it uses the title instead. So you dont see the issue. The test you mention was only checking a dependency between a repeating group and a repeating answer which is the use case we have in the LMS survey.

app/views/questionnaire.py
AndrewTorrance: Now we can have a dependency in a repeating group on a non repeating independent variable. Note this was an issue that not only affects titles , but any routing via when .

nealedj:
Fixes #1627

app/questionnaire/rules.py
```diff
@@ -170,14 +170,16 @@ def evaluate_when_rules(when_rules, metadata, answer_store, group_instance):
     :param metadata: metadata for evaluating rules with metadata conditions
     :param answer_store: store of answers to evaluate
     :param group_instance: when evaluating a when rule for a repeating group
+           this holds the group instance of the repeating group
     :return: True if the when condition has been met otherwise False
     """
     for when_rule in when_rules:
         if 'id' in when_rule:
             when_specified_group_instance = when_rule.get('group_instance')
```
nealedj: This could be:

```
target_group_instance = when_rule.get('group_instance', group_instance)
```

(renamed to `target_group_instance` to be explicit)

or

```
if 'group_instance' in when_rule:
    group_instance = when_rule['group_instance']
```

The latter may be clearer. I don't think we should need to explictly check for `None` as the only way this variable could hold that value is if group_instance wasn't defined for the when_rule.

app/questionnaire/rules.py
```diff
@@ -170,14 +170,16 @@ def evaluate_when_rules(when_rules, metadata, answer_store, group_instance):
...
-            instance = when_specified_group_instance or group_instance
+            if when_specified_group_instance is not None:   # 0 is a valid group instance
```
dwyeradam: If we used `g.schema.answer_is_in_repeating_group(when_rule['id'])` it would uniform this with your dependency changes. Since 'group_instance' is only in the one test schema we currently don't use it.

app/views/questionnaire.py
```diff
@@ -586,15 +586,22 @@ def _remove_dependent_answers_from_completed_blocks(answer_id, questionnaire_sto
...
 
-        location = Location(block['parent_id'], 0, block['id'])
-        if location in questionnaire_store.completed_blocks:
-            questionnaire_store.completed_blocks.remove(location)
+        if dependency_in_repeating_group and not answer_in_repeating_group:
+            questionnaire_store.remove_completed_by_group_and_block(block['parent_id'], block['id'])
+        else:
+            location = Location(block['parent_id'], group_instance, block['id'])
+            if location in questionnaire_store.completed_blocks:
+                questionnaire_store.completed_blocks.remove(location)
```
nealedj: Since we now have an interface for removing completed blocks from the questionnaire store could we please add a second method to remove complete blocks by group id, block id + group instance? Or maybe just have a single `remove_completed_blocks` method that can accept any of these arguments?

This would be a useful refactor as we'd like to gradually try and remove this sort of logic from the views package.

nealedj (changes requested):

app/views/questionnaire.py
```diff
@@ -586,15 +586,22 @@ def _remove_dependent_answers_from_completed_blocks(answer_id, questionnaire_sto
     :param questionnaire_store: holds the completed blocks
     :return: None
     """
+    answer_in_repeating_group = g.schema.answer_is_in_repeating_group(answer_id)
```
dwyeradam, line 590: It might be personal preference but I would rather `answer_is_in_repeating_group` to return a bool.

dwyeradam (changes requested):
More talking points than issues.

app/views/questionnaire.py
```diff
@@ -586,15 +586,22 @@ def _remove_dependent_answers_from_completed_blocks(answer_id, questionnaire_sto
...
         block = g.schema.get_block(question['parent_id'])
 
-        location = Location(block['parent_id'], 0, block['id'])
-        if location in questionnaire_store.completed_blocks:
-            questionnaire_store.completed_blocks.remove(location)
+        if dependency_in_repeating_group and not answer_in_repeating_group:
+            questionnaire_store.remove_completed_blocks(group_id=block['parent_id'], block_id=block['id'])
+        else:
+            location = Location(block['parent_id'], group_instance, block['id'])
+            if location in questionnaire_store.completed_blocks:
```
nealedj, line 604: @AndrewTorrance  Not going to block on something like this... but would it not be better to have this conditional within the `remove_completed_blocks` method itself? It seems like a reasonable thing for the `QuestionnaireSchema` object to manage itself as it's only related to its internal state.

nealedj (approved):

nealedj:
@AndrewTorrance I don't see the test schema listed in go launcher or a PR for it.

app/views/questionnaire.py
AndrewTorrance: Not sure . First of all I thought you were right , but its questionnaire.py which is a bunch of functions that have the business logic that use the schema and utalize the store .  I think we may be better off with a Questionnaire class that has a schema and a store ? I think its worth a talk through this

nealedj:
Sorry I meant questionnaire store, not schema.

questionnaire.py is overloaded IMO. It contains the request handlers and a
whole bunch of logic that would be better off living elsewhere.

app/questionnaire/rules.py
```diff
@@ -175,10 +177,8 @@ def evaluate_when_rules(when_rules, metadata, answer_store, group_instance):
     """
     for when_rule in when_rules:
         if 'id' in when_rule:
-            when_specified_group_instance = when_rule.get('group_instance')
-            if when_specified_group_instance is not None:   # 0 is a valid group instance
-                group_instance = when_specified_group_instance
-
+            if not _answer_is_in_repeating_group(when_rule['id']):
+                group_instance = 0
```
dwyeradam: @AndrewTorrance Is it worth just adding an initial `and` to check that group instance isn't already 0? Just reduces the amount of time that the _answer_is_in_repeating_group(...) need to be evaluated.

app/questionnaire/rules.py
```diff
@@ -170,14 +172,14 @@ def evaluate_when_rules(when_rules, metadata, answer_store, group_instance):
...
     :return: True if the when condition has been met otherwise False
     """
     for when_rule in when_rules:
         if 'id' in when_rule:
-            when_specified_group_instance = when_rule.get('group_instance')
-            instance = when_specified_group_instance or group_instance
-
-            value = get_answer_store_value(when_rule['id'], answer_store, instance)
+            if not _answer_is_in_repeating_group(when_rule['id']):
+                group_instance = 0
```
dwyeradam: @AndrewTorrance Is it worth just adding an initial and to check that group instance isn't already 0? Just reduces the amount of time that the _answer_is_in_repeating_group(...) need to be evaluated. It's not a lot of overhead so just a thought.

AndrewTorrance: Fair comment , doing

dwyeradam (approved):

jonnyshaw89:
@AndrewTorrance Squishy Squashy?

jonnyshaw89:
@AndrewTorrance looks like you have a merge error

PR #1635, opened 2018-06-18
Implement Dynamodb with RDS fallback

kierenbeckett:
We want to store out state in dynamodb table rather than in RDS

### How to review 
Run a survey, check /dump/submission after posting some data to see that it saved correctly, see that it completes successfully


.travis.yml
```diff
@@ -9,6 +9,22 @@ env:
     - EQ_RUN_DOCKER_UP=True
     - EQ_RUN_FUNCTIONAL_TESTS=True
     - EQ_RUN_FUNCTIONAL_TESTS_HEADLESS=True
+    - EQ_DYNAMODB_ENDPOINT=http://localhost:6060
+    - EQ_SUBMITTED_RESPONSES_TABLE_NAME=dev-submitted-responses
+    - EQ_SUBMITTED_RESPONSES_DYNAMO_READ=True
+    - EQ_SUBMITTED_RESPONSES_DYNAMO_WRITE=True
```
jonnyshaw89: submitted responses is only ever dynamo so we probably don't need these?

app/data_model/models.py
```diff
@@ -31,10 +39,17 @@ class EQSession(db.Model):
...
         self.eq_session_id = eq_session_id
         self.user_id = user_id
-        self.session_data = None
+        self.session_data = session_data
+
+    def to_new_model(self):
+        return app_models.EQSession(self.eq_session_id, self.user_id, self.session_data, self.created_at, self.updated_at)
+
+    @classmethod
+    def from_new_model(cls, model):
```
jonnyshaw89: `from_new_model` or `from_model`?

app/settings.py
```diff
@@ -66,10 +66,21 @@ def get_env_or_fail(key):
...
-EQ_DYNAMODB_ENABLED = parse_mode(os.getenv('EQ_DYNAMODB_ENABLED', 'True'))
-if EQ_DYNAMODB_ENABLED:
-    EQ_DYNAMODB_ENDPOINT = os.getenv('EQ_DYNAMODB_ENDPOINT')
-    EQ_SUBMITTED_RESPONSES_TABLE_NAME = get_env_or_fail('EQ_SUBMITTED_RESPONSES_TABLE_NAME')
+EQ_DYNAMODB_ENDPOINT = os.getenv('EQ_DYNAMODB_ENDPOINT')
+EQ_DYNAMODB_MAX_RETRIES = int(os.getenv('EQ_DYNAMODB_MAX_RETRIES', 1))
+EQ_DYNAMODB_MAX_POOL_CONNECTIONS = int(os.getenv('EQ_DYNAMODB_MAX_POOL_CONNECTIONS', 30))
+EQ_SUBMITTED_RESPONSES_TABLE_NAME = os.getenv('EQ_SUBMITTED_RESPONSES_TABLE_NAME')
+EQ_SUBMITTED_RESPONSES_DYNAMO_READ = parse_mode(os.getenv('EQ_SUBMITTED_RESPONSES_DYNAMO_READ', 'True'))
+EQ_SUBMITTED_RESPONSES_DYNAMO_WRITE = parse_mode(os.getenv('EQ_SUBMITTED_RESPONSES_DYNAMO_WRITE', 'True'))
```
jonnyshaw89: I don't think we need these 2

app/storage/data_access.py
```diff
@@ -0,0 +1,188 @@
...
+from app.storage import dynamo_api
+
+logger = get_logger()
+
+
+TABLE_CONFIG = {
+    app_models.SubmittedResponse: {
+        'table_name_key': 'EQ_SUBMITTED_RESPONSES_TABLE_NAME',
+        'dynamo_read_key': 'EQ_SUBMITTED_RESPONSES_DYNAMO_READ',
+        'dynamo_write_key': 'EQ_SUBMITTED_RESPONSES_DYNAMO_WRITE',
```
jonnyshaw89: Just default these to true as they shouldn't be switchable

app/storage/data_access.py
```diff
@@ -0,0 +1,188 @@
...
+
+
+def setup_tables():
+    """Sets up DynamoDB tables
+    """
+    for config in TABLE_CONFIG.values():
+        if _is_dynamodb_read_enabled(config):
+            table_name = _get_table_name(config)
+            if table_name:
+                dynamo_api.create_table(table_name, config['key_field'])
```
jonnyshaw89: Do we want this? as we create our tables with Terraform.

app/storage/data_access.py
```diff
@@ -0,0 +1,188 @@
...
+
+
+def flush_all_data():
+    """Deletes all data in DynamoDB
```
jonnyshaw89: Where do we use this? seems pretty dangerous

app/storage/dynamo_api.py
```diff
@@ -0,0 +1,71 @@
...
+logger = get_logger()
+
+
+def put_item(table_name, item, overwrite=True):
+    """Insert an item into table"""
+    table = _get_table(table_name)
+
+    put_kwargs = {'Item': item}
+    if not overwrite:
+        first_key = next(iter(item.keys()))
```
jonnyshaw89: Can we guarantee that the 1st key is and always will be the table key?

app/storage/dynamo_api.py
```diff
@@ -0,0 +1,71 @@
...
+    """
+    table = _get_table(table_name)
+
+    response = table.delete_item(Key=key)
+    item = response.get('Item', None)
+    return item
+
+
+def delete_all(table_name):
+    """Deletes all items from a table
```
jonnyshaw89: This also seems dangerous to me, why do we need this?

app/storage/dynamo_api.py
```diff
@@ -0,0 +1,71 @@
...
+            key_dict = {k: item[k] for k in keys}
+            batch.delete_item(Key=key_dict)
+
+
+def _get_table(table_name):
+    return current_app.eq['dynamodb'].Table(table_name)
+
+
+def create_table(table_name, key_field):
+    current_app.eq['dynamodb'].create_table(
```
jonnyshaw89: We shouldn't be creating tables in code.
It also wont/shouldn't have permission to do this

app/storage/encrypted_questionnaire_storage.py
```diff
@@ -22,21 +23,18 @@ def add_or_update(self, data, version):
         questionnaire_state = self._find_questionnaire_state()
         if questionnaire_state:
             logger.debug('updating questionnaire data', user_id=self._user_id)
-            questionnaire_state.state = encrypted_data_json
+            questionnaire_state.q_state = encrypted_data_json
```
jonnyshaw89: `q_state`? what does that mean?

app/templates/summary.html
```diff
@@ -37,7 +37,7 @@ <h1 class="saturn">{{content.summary.title}}</h1>
 
 {% include theme('partials/summary/summary.html') %}
 
-{% if content.is_view_submission_response_enabled %}
+{% if content.summary.is_view_submission_response_enabled %}
```
jonnyshaw89: Is this change related?

jonnyshaw89 (changes requested):

app/templates/summary.html
kierenbeckett: No, I can move it into a separate PR or make a ticket?

app/storage/encrypted_questionnaire_storage.py
kierenbeckett: It means state is a reserved keyword in dynamo, short for questionnaire state

app/storage/dynamo_api.py
kierenbeckett: This is to create tables in the dynamodb mock, I'm using moto

app/storage/dynamo_api.py
kierenbeckett: For the tests

app/storage/dynamo_api.py
kierenbeckett: I went back and forth on this one.  We can't, but all we need here is a field that is guaranteed to be in all items.  I'll rejig it to explicitly use the key.

app/storage/data_access.py
kierenbeckett: For the tests

app/storage/data_access.py
kierenbeckett: Creates tables in mock dynamodb

app/data_model/models.py
kierenbeckett: It might be a bit confusing to have `from_model` in a module named `models`, how about `from_app_model`?

jonnyshaw89: yeh, thats better as it matches the naming of `app_models` which is what they are

app/storage/data_access.py
jonnyshaw89: Then it should live in the tests

app/storage/data_access.py
jonnyshaw89: move it into tests then

app/storage/dynamo_api.py
jonnyshaw89: can we use this? https://github.com/ONSdigital/eq-survey-runner/pull/1635/files#diff-53d3d8854e66c3440c718db4d80e0ecfR18

app/storage/dynamo_api.py
jonnyshaw89: same as above. if its for tests it should be in the tests folder

app/storage/dynamo_api.py
jonnyshaw89: same again

app/storage/encrypted_questionnaire_storage.py
jonnyshaw89: We generally don't use abbreviations, as they are not always clear. so this is `questionnaire_state.questionnaire_state` which then seems a bit odd?

app/templates/summary.html
jonnyshaw89: raise a ticket for it. then it doesn't muddy the waters

app/storage/data_access.py
```diff
@@ -0,0 +1,175 @@
...
+            models.db.session.commit()
+
+
+def get_table_name(config):
+    return app.config[config['table_name_key']]
+
+
+def is_dynamodb_read_enabled(config):
+    # If we do not have a config key, assume dynamo is enabled
+    if not config['dynamo_read_key']:
```
jonnyshaw89: Had a tip off on some smelly code
"will eval to true if they key had the value False rather than empty"
maybe should be `if 'dynamo_write_key' not in config:`

app/authentication/jti_claim_storage.py
```diff
@@ -41,10 +45,12 @@ def use_jti_claim(jti_claim):
         raise TypeError
 
     try:
-        jti = UsedJtiClaim(jti_claim)
-        # pylint: disable=maybe-no-member
-        db.session.add(jti)
-        db.session.commit()
-    except IntegrityError as e:
+        used_at = datetime.now(tz=tzutc())
+        valid_until = used_at + timedelta(seconds=current_app.config['EQ_USED_JTI_CLAIM_TTL_SECONDS'])
```
jonnyshaw89: I think we should use the `exp` value in the JWT for this.
http://ons-schema-definitions.readthedocs.io/en/latest/respondent_management_to_electronic_questionnaire.html

app/data_model/app_models.py
```diff
@@ -0,0 +1,95 @@
...
+
+    @post_load
+    def make_model(self, data):
+        return EQSession(**data)
+
+
+class UsedJtiClaimSchema(Schema):
+    jti_claim = fields.Str()
+    used_at = fields.DateTime()
+    valid_until = Timestamp()
```
jonnyshaw89: This technically isn't `valid_until` its more like `invalid_until` as if its in our DB then it can't be used

app/settings.py
```diff
@@ -66,10 +66,20 @@ def get_env_or_fail(key):
...
+EQ_QUESTIONNAIRE_STATE_TABLE_NAME = os.getenv('EQ_QUESTIONNAIRE_STATE_TABLE_NAME')
+EQ_QUESTIONNAIRE_STATE_DYNAMO_READ = parse_mode(os.getenv('EQ_QUESTIONNAIRE_STATE_DYNAMO_READ', 'False'))
+EQ_QUESTIONNAIRE_STATE_DYNAMO_WRITE = parse_mode(os.getenv('EQ_QUESTIONNAIRE_STATE_DYNAMO_WRITE', 'False'))
+EQ_SESSION_TABLE_NAME = os.getenv('EQ_SESSION_TABLE_NAME')
+EQ_SESSION_DYNAMO_READ = parse_mode(os.getenv('EQ_SESSION_DYNAMO_READ', 'False'))
+EQ_SESSION_DYNAMO_WRITE = parse_mode(os.getenv('EQ_SESSION_DYNAMO_WRITE', 'False'))
+EQ_USED_JTI_CLAIM_TABLE_NAME = os.getenv('EQ_USED_JTI_CLAIM_TABLE_NAME')
+EQ_USED_JTI_CLAIM_DYNAMO_READ = parse_mode(os.getenv('EQ_USED_JTI_CLAIM_DYNAMO_READ', 'False'))
+EQ_USED_JTI_CLAIM_DYNAMO_WRITE = parse_mode(os.getenv('EQ_USED_JTI_CLAIM_DYNAMO_WRITE', 'False'))
+EQ_USED_JTI_CLAIM_TTL_SECONDS = int(os.getenv('EQ_USED_JTI_CLAIM_TTL_SECONDS', 900))
```
jonnyshaw89: We should let the calling service define this in the JWT with `exp`

app/settings.py
```diff
@@ -66,10 +66,20 @@ def get_env_or_fail(key):
 EQ_SERVER_SIDE_STORAGE_USER_ID_ITERATIONS = ensure_min(os.getenv('EQ_SERVER_SIDE_STORAGE_USER_ID_ITERATIONS', 10000),
                                                        1000)
 
-EQ_DYNAMODB_ENABLED = parse_mode(os.getenv('EQ_DYNAMODB_ENABLED', 'True'))
-if EQ_DYNAMODB_ENABLED:
-    EQ_DYNAMODB_ENDPOINT = os.getenv('EQ_DYNAMODB_ENDPOINT')
-    EQ_SUBMITTED_RESPONSES_TABLE_NAME = get_env_or_fail('EQ_SUBMITTED_RESPONSES_TABLE_NAME')
+EQ_DYNAMODB_ENDPOINT = os.getenv('EQ_DYNAMODB_ENDPOINT')
+EQ_DYNAMODB_MAX_RETRIES = int(os.getenv('EQ_DYNAMODB_MAX_RETRIES', 1))
```
jonnyshaw89: Im in 2 minds about the retries. Do we want to make the user wait a little bit longer if we hit DB capacity?
Is it better to eventually save the users input rather than error?

app/storage/encrypted_questionnaire_storage.py
```diff
@@ -22,21 +23,18 @@ def add_or_update(self, data, version):
         questionnaire_state = self._find_questionnaire_state()
         if questionnaire_state:
             logger.debug('updating questionnaire data', user_id=self._user_id)
-            questionnaire_state.state = encrypted_data_json
+            questionnaire_state.state_data = encrypted_data_json
```
jonnyshaw89, line 26: the repetition of state seems a bit odd? the fact it is state is implied by the object name?
so maybe just `.data`?

tests/app/app_context_test_case.py
```diff
@@ -11,19 +16,56 @@ class AppContextTestCase(unittest.TestCase):
...
 
     def setUp(self):
+        self._ddb = mock_dynamodb2()
+        self._ddb.start()
+
         setting_overrides = {
             'SQLALCHEMY_DATABASE_URI': 'sqlite://',
             'LOGIN_DISABLED': self.LOGIN_DISABLED,
-            'EQ_DYNAMODB_ENABLED': False
+            'EQ_DYNAMODB_ENDPOINT': None,
```
jonnyshaw89, line 24: Why?

jonnyshaw89 (changes requested):

app/storage/encrypted_questionnaire_storage.py
kierenbeckett: `data` is also a reserved keyword in dynamo

tests/app/app_context_test_case.py
kierenbeckett: moto doesn't mock dynamo if an endpoint is specified

app/settings.py
kierenbeckett: The retry backoff is described here https://github.com/boto/botocore/blob/develop/botocore/data/_retry.json#L108.  So seconds wait between retries would be: `0.05, 0.1, 0.2, 0.4, 0.8, 1.6, 3.2, 6.4, 12.8, 25.6`.  I think we could afford a few retries, say 5?

app/storage/encrypted_questionnaire_storage.py
jonnyshaw89: 🤦‍♂️

app/settings.py
jonnyshaw89: yeh, will give that a go.

Pipfile
```diff
@@ -26,6 +26,8 @@ httmock = "==1.2.6"
 "flake8-print" = "*"
 "flake8-tuple" = "*"
 python-dateutil = "*"
+moto = "*"
+google-compute-engine = "*"
```
jonnyshaw89, line 30: what is this used for?

app/authentication/jti_claim_storage.py
```diff
@@ -40,11 +43,18 @@ def use_jti_claim(jti_claim):
         logger.info('jti claim is invalid', jti_claim=jti_claim)
         raise TypeError
 
+    exp = datetime.utcfromtimestamp(exp_timestamp).replace(tzinfo=tzutc())
+    if exp < datetime.now(tz=tzutc()):
```
jonnyshaw89: So this wasn't being checked?

jonnyshaw89: it shouldn't be park of the JTI logic thou its general login checking

app/authentication/jti_claim_storage.py
```diff
@@ -40,11 +43,18 @@ def use_jti_claim(jti_claim):
...
+
     try:
-        jti = UsedJtiClaim(jti_claim)
-        # pylint: disable=maybe-no-member
-        db.session.add(jti)
-        db.session.commit()
-    except IntegrityError as e:
+        used_at = datetime.now(tz=tzutc())
+        # Make claim expire a little later than exp to avoid race conditions with out of sync clocks
+        expires = exp + timedelta(seconds=60)
```
jonnyshaw89: There is already a leeway built into the `exp`. also DynamoDB takes a minute or so to delete

jonnyshaw89 (changes requested):
Getting there. Don't think we are too far off now.

Pipfile
kierenbeckett: https://github.com/GoogleCloudPlatform/compute-image-packages/issues/262

Travis was failing to install boto into a virtualenv, presumably because it's on google cloud and they're doing something funky to boto in their debian images.  Options were `--use-site-packages`, which I didn't fancy or to add `google-compute-engine` as a dependency.

app/authentication/jti_claim_storage.py
kierenbeckett: As far as I could tell, no.  I'll move the check upwards.

app/authentication/jti_claim_storage.py
kierenbeckett: I was concerned about the following (slightly extreme example):
* The clock in dynamo is 10 seconds ahead of the one in survey runner and the one that sets `exp`
* A request is made, JTI token gets persisted with TTL `exp`
* Dynamo reaper happens to run 9 seconds before `exp` (survey runner clock), deletes JTI token
* Another request is made with JTI token, it's not expired according to survey runner clock and isn't in dynamo, so gets used a second time

app/data_model/app_models.py
```diff
@@ -0,0 +1,95 @@
...
+from dateutil.tz import tzutc
+from marshmallow import Schema, fields, post_load, pre_dump
+
+
+class QuestionnaireState:
+    def __init__(self, user_id, state_data, version, created_at=None, updated_at=None):
+        self.user_id = user_id
+        self.state_data = state_data
+        self.version = version
+        self.created_at = created_at or datetime.datetime.now(tz=tzutc())
```
nealedj: Won't this set the `created_at` timestamp both on the way in and out of the database? This shouldn't be an issue given that all entities will have been saved with these values - but could hide a bug in future?

app/views/session.py
```diff
@@ -73,9 +76,13 @@ def login():
 
 
 def validate_jti(decrypted_token):
+    expires = datetime.utcfromtimestamp(decrypted_token['exp']).replace(tzinfo=tzutc())
```
nealedj, line 79: As discussed - this may be unnecessary as it looks like it's already checked in jwcrypto https://github.com/latchset/jwcrypto/blob/master/jwcrypto/jwt.py#L324

docker-compose.yml
```diff
@@ -75,7 +84,7 @@ services:
 
 
   eq-docker-dynamodb:
-    image: onsdigital/eq-docker-dynamodb:latest
+    image: onsdigital/eq-docker-dynamodb:eq-2120-dynamodb-spike
```
nealedj: Leaving a blocking comment as a reminder to switch this back before merging.

tests/app/app_context_test_case.py
```diff
@@ -11,19 +16,56 @@ class AppContextTestCase(unittest.TestCase):
...
         self._app = create_app(setting_overrides)
 
         self._app.config['SERVER_NAME'] = 'test.localdomain'
         self._app_context = self._app.app_context()
         self._app_context.push()
 
+        setup_tables()
+
     def tearDown(self):
+        flush_all_data()
```
nealedj: Is it necessary to flush the dynamo data now that it's being mocked out?

tests/integration/integration_test_case.py
```diff
@@ -95,6 +85,15 @@ def _set_up_app(self, enable_dynamo_db=False):
 
         self._client = self._application.test_client()
 
+        with self._application.app_context():
+            setup_tables()
+
+    def tearDown(self):
+        with self._application.app_context():
+            flush_all_data()
```
nealedj: As per other comment - do we need to flush data with `moto`?

nealedj (changes requested):

app/views/session.py
kierenbeckett: sdc doesn't invoke this check, doesn't even have the option to.  It's probably worth us doing manually.

tests/app/data_model/test_app_models.py
```diff
@@ -16,10 +16,14 @@ def test_submitted_response(self):
         self._test_model(SubmittedResponse('txid', 'somedata', NOW))
 
     def test_questionnaire_state(self):
-        self._test_model(QuestionnaireState('someuser', 'somedata', 1))
+        new = self._test_model(QuestionnaireState('someuser', 'somedata', 1))
+        self.assertGreaterEqual(new.updated_at, NOW)
+        self.assertGreaterEqual(new.updated_at, NOW)
```
jonnyshaw89: should this line be duplicated?

docker-compose.yml
```diff
@@ -46,8 +46,17 @@ services:
...
+      EQ_SUBMITTED_RESPONSES_TABLE_NAME: dev-submitted-responses
+      EQ_QUESTIONNAIRE_STATE_TABLE_NAME: dev-questionnaire-state
+      EQ_QUESTIONNAIRE_STATE_DYNAMO_READ: "True"
+      EQ_QUESTIONNAIRE_STATE_DYNAMO_WRITE: "True"
+      EQ_SESSION_TABLE_NAME: dev-eq-session
+      EQ_SESSION_DYNAMO_READ: "True"
+      EQ_SESSION_DYNAMO_WRITE: "True"
+      EQ_USED_JTI_CLAIM_TABLE_NAME: dev-used-jti-claim
+      EQ_USED_JTI_CLAIM_DYNAMO_READ: "True"
+      EQ_USED_JTI_CLAIM_DYNAM_WRITE: "True"
```
jonnyshaw89, line 59: You should also update https://github.com/ONSdigital/eq-compose

nealedj (approved):

docker-compose.yml
kierenbeckett: See ONSdigital/eq-compose#14

tests/app/app_context_test_case.py
kierenbeckett: Good point

app/data_model/app_models.py
kierenbeckett: IRL: Agreed it would be neater to get the database layer to do this, however that involves using dynamodb's `update_item` rather than `put_item` .  This gets a bit cumbersome as you have to construct an expression of fields that need updating and a way of specifying fields that should be excludes.  I did start down this path but changed my mind when we hit an error due to one of our fields, `data` on `SubmittedResponse` being a reserved keyword (fine for put apparently, but not to put in an update expression).  I didn't want to get sucked into doing a field migration at the same time as adding the new tables.

jonnyshaw89 (approved):

jonnyshaw89:
Jenkins should be updated to the latest version of survey runner beanstalk so the following non-defaulted env vars are set on the environment before this is merged. otherwise deployments to preprod will fail
https://github.com/ONSdigital/eq-survey-runner/pull/1635/files#diff-cb8b13ea27d26f0e53eb65167566c085R72

PR #1636, opened 2018-06-18
Add playback block to MBS 0111

nealedj:
Adds a playback/confirmation block to MBS 0111.

### How to review 
Run MBS 0111. Check that the question "For the period {{start date}} to {{end date}} ,the value of the total turnover was {total_turnover}, is this correct?" is shown.

If "No" then user should be taken back to the turnover question. If "Yes" they should continue to the changes in total turnover question.

### DIFF
--- a/data/en/mbs_0111.json
+++ b/data/en/mbs_0111.json
@@ -230,6 +230,7 @@
                             "answers": [{
                                 "id": "turnover-answer",
                                 "label": "Total turnover excluding VAT",
+                                "description": "Enter value in full. For example, for 56 thousand you would enter 56,000",
                                 "type": "Currency",
                                 "currency": "GBP",
                                 "decimal_places": 2,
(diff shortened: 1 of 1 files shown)


dwyeradam (changes requested):
Don't forget "Enter value in full. For example, for 56 thousand you would enter 56,000" has been added to the Total Turnover answer.

dwyeradam:
Are we going to add a default of zero to Total Turnover and remove the mandatory or make them answer? This will then be caught by the confirmation rather than an error message.

MebinAbraham (changes requested):
Found a problem. If you type in `0`, then click `Continue`, then `Previous` then `Continue` the piped value is missing.
Existing issue?
screen shot 2018-06-18 at 13 41 22

shahi645 (approved):
I'm ok to approve, the current bug has been brought up

dwyeradam (approved):

MebinAbraham (approved):
👍 Issue fixed in another PR.

PR #1637, opened 2018-06-18
Add test to clarify expected behaviour for multiple sessions on same survey

nealedj:
Integration test to check that multiple flask sessions can be active against a single questionnaire.


shahi645 (approved):
LGTM

jonnyshaw89 (approved):
👍 Seems to cover what our current setup supports

PR #1638, opened 2018-06-18
Fix formatting of zero float values

nealedj:
Fixes formatting of empty float values.

### How to review 
Test MBS 0111 (updated as per #1636), enter a turnover of `0.00` (with decimal point). Confirmation page should show `£0.00` and not an empty string.


MebinAbraham (dismissed):
Working as expected. Tested against MBS.
👍

shahi645 (dismissed):
LGTM

app/jinja_filters.py
```diff
@@ -22,16 +23,18 @@
 
 @blueprint.app_template_filter()
 def format_number(value):
-    if value or value is 0:
+    try:
         return numbers.format_decimal(value, locale=DEFAULT_LOCALE)
-    return ''
+    except (TypeError, decimal.InvalidOperation):
```
dwyeradam: I personally don't like using exceptions to handle expected behaviour. I'm sure this will open a series of debates depending how long you've been doing Python compared to languages like Java.

dwyeradam (dismissed):

app/jinja_filters.py
jonnyshaw89: I think that Exceptions should be used for Exceptions and not as flow control.
We expect this to happen so we should check for it.

nealedj: This isn't flow control...

nealedj: And we don't expect for it to happen. Why would we pass a value to format currency that we didn't expect to get formatted?

jonnyshaw89: If we are not expecting it why are we handing it? surely we should be handing it where it is called then?

nealedj: In this case it'd be in a template where that logic would be cumbersome.

I think this is inline with how I've seen formatter functions written (including in the jinja source code). It's also in line with the "easier to ask forgiveness than permission" python mantra.

I can't think of a more readable/safe way to write this - happy to consider other options though.

kierenbeckett: Why not `value == 0` rather than `value is 0`?

jonnyshaw89: I prefer that if it works?

nealedj: Yeah that'd work.

I find it jarring because this is the third issue I've seen with these filters making unnecessary erroneous checks.

I guess my main issue is that it's simply less readable and provides no performance benefit.

nealedj: Just because I keep questionning myself - this answer is from a long standing Python core dev. https://stackoverflow.com/questions/16138232/is-it-a-good-practice-to-use-try-except-else-in-python

kierenbeckett: FTR in python I think it's fine to use exceptions this way if the alternative is attempting to list all the conditions that might make `format_decimal` fail.  Beyond `None` and `0` though would we rather other invalid values cause an error we can see/investigate or are we happy for them to be silently left blank?

shahi645 (approved):

jonnyshaw89 (approved):

dwyeradam (approved):

PR #1643, opened 2018-06-20
Mark static content for translation

dwyeradam:
Mark any static content that could get rendered as translatable.
This includes html, error messages etc.

### How to review 
Change setting `BABEL_DEFAULT_LOCALE` to `'cy'` in **settings.py**
Run: `pybabel init -i app/translations/messages.pot -d app/translations -l cy`

Swap `app/translations/cy/LC_MESSAGES/messages.po` with this [messages.txt](https://github.com/ONSdigital/eq-survey-runner/files/2120174/messages.txt) and change suffix to **.po** rather than **.txt**. 

Run: `pybabel compile -d app/translations`

Now you're ready to go. Load any schema and all static content should be in Welsh (albeit poorly translated). 

I'm bound to have missed something so try to test as many page types as you can.


app/templates/partials/summary/radio.html
```diff
@@ -1,10 +1,9 @@
-{% set not_answered = 'No answer provided' %}
 {{answer.value}}
 {%- if answer.child_answer_value is not none -%}
     <ul>
         <li>
             {%- if answer.child_answer_value == ''  -%}
-                {{not_answered}}
+                {{ _("No answer provided") }} %}
             {%- else -%}
```
MebinAbraham: Invalid syntax - `{{ _("No answer provided") }} %}`, leftover `%}`

Also: https://github.com/ONSdigital/eq-survey-runner/blob/f634dc2a9d0cf14eb746fc0caaafafe81c8f13c9/app/templates/partials/summary/answer.html#L1

MebinAbraham (commented):
Are Census pages within the scope of this PR?

app/templates/view-submission.html
```diff
@@ -5,20 +5,38 @@
...
 
+{% if metadata.trad_as %}
+    {% set  trading_as = metadata.trad_as %}
+    {% set  trad_as_open = '<span class="trading_as"> (<span class="trad_as">' %}
+    {% set  trad_as_close = '</span>)</span>' %}
+{% endif %}
+
 <div>
   <h1 class="saturn">Submitted answers</h1>
 </div>
```
MebinAbraham: Missed.

MebinAbraham (commented):
Missed.
https://github.com/ONSdigital/eq-survey-runner/blob/558fe198ca7732b4b079c9ef5f07c34a84b125b0/app/templates/layouts/_base.html#L80

MebinAbraham (commented):
Missed.
https://github.com/ONSdigital/eq-survey-runner/blob/558fe198ca7732b4b079c9ef5f07c34a84b125b0/app/templates/static/contact-us.html#L4

dcdarrell9:
Does there need to be translations here?

screen shot 2018-06-22 at 08 34 15

dcdarrell9:
Characters remaining not translated

screen shot 2018-06-22 at 08 58 41

dwyeradam:
@dcdarrell9, this card only does the static content. Your screen print shows schema content which is covered by a separate card. It does make this quite difficult to review sorry.

I'll look at the characters remaining. I thought I had done that but may have been lost in a rebase.

app/templates/static/cookies-privacy.html
```diff
@@ -26,160 +24,153 @@
...
 
-                            <p>The Office for National Statistics (ONS) has created this cookies and privacy
-                                statement in order to demonstrate our firm commitment to privacy.</p>
+                            <p>{{ _("The Office for National Statistics (ONS) has created this cookies and privacy
+                                statement in order to demonstrate our firm commitment to privacy.") }}</p>
 
-                            <p>The following is a description of what data is collected when you use the Survey Data Collection (SDC) service,
-                                how it is stored and what it is used for.</p>
+                            <p>{{ _("The following is a description of what data is collected when you use the Survey Data Collection (SDC) service,
+                                how it is stored and what it is used for") }}</p>
```
dcdarrell9: `.` has been removed

MebinAbraham (commented):
@dwyeradam The character remaining bit is done in JS. You have translated the actual `Characters remaining` static content but in template its just `-` being replaced with the validation dynamically based on the specified length.

In `components.js`
`var charactersRemainingSuffix = " characters remaining";`

app/templates/static/cookies-privacy.html
dwyeradam: Good spot.

dcdarrell9 (approved):

kierenbeckett (approved):

MebinAbraham (commented):
Cookies page:
screen shot 2018-06-27 at 09 45 10

---
Also is the messages file up to date?
On cookies page:
```
The following is a description of what data is collected when you use the Survey Data Collection (SDC) service, how it is stored and what it is used for.
```
Isn't translated but its marked for translation?

MebinAbraham:
Also not sure if we care but translations seem to be adding extra spaces.
English:
`<a href="mailto:surveys@ons.gov.uk" target="_top">surveys@ons.gov.uk</a>`
Welsh:
`<a href="mailto:surveys@ons.gov.uk" target="_top"> surveys@ons.gov.uk </a>`

Makes no difference functionally I believe.

dwyeradam:
@MebinAbraham if it's marked and not translated that's probably a problem with the translations file. It's not completely up to date. Also with regards to the extra spaces that's the Google Translations Toolbar adding them. Something we wouldn't need to worry about as part of this.

MebinAbraham (commented):
Regarding error messages, they don't match up?
screen shot 2018-06-27 at 11 11 43
screen shot 2018-06-27 at 11 10 55
The style is missing **(strong)** and are we not doing the count?

armstrongb:
There is a separate card in the pattern library backlog for dealing with CY strings in .js reliant components like the character count, accordion etc.

@dwyeradam @MebinAbraham @dcdarrell9

MebinAbraham (approved):
Looks good 👍

app/templates/partials/header.html
```diff
@@ -8,14 +8,14 @@
...
+            <h3 class="info__title venus" id="survey-help">{{ _("Survey Help") }}</h3>
             <div class="info__tel">
                 <a class="info__link" href="tel:03001234931" aria-describedby="survey-help">0300 1234 931</a>
             </div>
             {% if meta %}
             <div>
                 <dl class="info__list">
-                    <dt class="info__dt first"><abbr title="Survey identification code">ID:</abbr></dt>
+                    <dt class="info__dt first"><abbr title="Survey identification code">{{ _("ID") }}:</abbr></dt>
                     <dd class="info__dd">{{ survey_id or 'N/A'}}</dd>
```
nealedj: I think "N/A" needs to be translatable.

app/templates/partials/header.html
```diff
@@ -8,14 +8,14 @@
...
                     <dt class="info__dt">Period:</dt>
```
nealedj: Missed string?

app/templates/confirmation.html
```diff
@@ -5,8 +5,8 @@
 {% block submit_button %}
 
 {% if content.summary and content.summary.summary_type == 'SectionSummary' %}
-    <button class="btn btn--primary btn--lg btn--loader js-btn-submit" data-qa="btn-submit">Continue</button>
+    <button class="btn btn--primary btn--lg btn--loader js-btn-submit" data-qa="btn-submit">{{ _("Continue") }}</button>
 {% else  %}
-    <button class="btn btn--primary btn--lg btn--loader js-btn-submit" data-qa="btn-submit" data-loading-msg="Submitting&hellip;" type="submit">Submit answers</button>
+    <button class="btn btn--primary btn--lg btn--loader js-btn-submit" data-qa="btn-submit" data-loading-msg="Submitting&hellip;" type="submit">{{ _("Submit answers") }}</button>
```
nealedj: I think the data loading message needs to be marked up too.

app/templates/partials/timeout.html
```diff
@@ -23,9 +23,9 @@ <h1 class="saturn dialog__title" id="dialog-title" tabindex="0">Your session is
       </div>
 
       <form class="" action="/" method="post">
-        <button class="btn btn--primary btn--loader js-timeout-continue" type="button" data-loading-msg="Continuing&hellip;">Continue survey</button>
+        <button class="btn btn--primary btn--loader js-timeout-continue" type="button" data-loading-msg="Continuing&hellip;">{{ _("Continue survey") }}</button>
```
nealedj: Another data loading message

dwyeradam:
@MebinAbraham The error messages are an oddity with Welsh translation. Try test_dates.json with 3 errors on the page and it works. I don't understand how Welsh pluralisation works but not an issue with how they are marked.

app/validation/validators.py
```diff
@@ -295,23 +296,29 @@ def _is_first_relative_delta_largest(relativedelta1, relativedelta2):
...
         return error_message
 
     @staticmethod
     def return_error_component(error_message, number, unit):
+        unit_plurals = {
+            'year': [_('years'), _('year')],
+            'month': [_('months'), _('month')],
+            'day': [_('days'), _('day')]
+        }
+        plural = unit_plurals[unit][(1 if number == 1 else 0)]
```
nealedj: Welsh has 4 plural forms (`Plural-Forms: nplurals=4; plural=(n==1) ? 0 : (n==2) ? 1 : (n != 8 && n != 11) ? 2 : 3;\n`)

E.g. (from google translate)
0 years: 0 blynedd
1 year: 1 flynedd
2 years: 2 flynedd
3 (and 4 and 6) years: 3 blynedd
5 (and all other years): 5 mlynedd

gettext supports plural forms natively with [ngettext](https://pythonhosted.org/Flask-Babel/#flask.ext.babel.ngettext).

I'm a little concerned with all the string concatenation that we're going to either get untranslated parts of validation error messages or we'll end up preventing a translator from constructing the text correctly.

Here's some english -> welsh examples:

"Enter a reporting period greater than or equal to 0 years." -> "Rhowch gyfnod adrodd yn fwy na neu'n hafal i 0 blynedd."
"Enter a reporting period greater than or equal to 1 years." -> "Rhowch gyfnod adrodd yn fwy nag 1 flwyddyn neu fwy cyfartal."
"Enter a reporting period greater than or equal to 2 years." -> "Rhowch gyfnod adrodd yn fwy na 2 flynedd neu gyfartal."

Note that there's extra differences beyond the word for "year" that get altered depending on the count.

I think the only way we can reasonably do this is to define all of the error messages in their complete forms and accepting only a numeric count as an argument rather than a translated unit.

"Enter a reporting period greater than or equal to %(min)s years."
"Enter a reporting period greater than or equal to %(min)s months."
"Enter a reporting period greater than or equal to %(min)s days."

tests/app/forms/test_questionnaire_form.py
```diff
@@ -14,7 +16,8 @@ class TestQuestionnaireForm(AppContextTestCase):  # noqa: C901  pylint: disable=
 
     @staticmethod
     def _error_exists(answer_id, msg, mapped_errors):
-        return any(a_id == answer_id and msg in ordered_errors for a_id, ordered_errors in mapped_errors)
+        msg = str(msg) if isinstance(msg, LazyString) else msg
```
nealedj: No need to check `isinstance(msg, LazyString)` here is there? Just always run `str(msg)`. It'll just return immediately if `msg` is already a string.

tests/app/forms/test_questionnaire_form.py
```diff
@@ -14,7 +16,8 @@ class TestQuestionnaireForm(AppContextTestCase):  # noqa: C901  pylint: disable=
...
+        return any(a_id == answer_id and str(msg) in ordered_errors for a_id, ordered_errors in mapped_errors)
```
nealedj: Ah - and `str(msg)` is getting run in the loop as well.

tests/integration/integration_test_case.py
```diff
@@ -207,6 +208,8 @@ def getHtmlSoup(self):
 
     # Extra Helper Assertions
     def assertInPage(self, content, message=None):
+        content = str(content) if isinstance(content, LazyString) else content
```
nealedj: Again - possibly no need for `isinstance(msg, LazyString)` checks.

nealedj (changes requested):

app/templates/partials/header.html
MebinAbraham: I've also raised an issue relating to the `if` as `meta` isn't valid anymore since the metadata PR

tests/app/forms/test_questionnaire_form.py
dwyeradam: I wanted to be a bit more explicit as to why i'm converting it but maybe overkill.

app/templates/partials/header.html
dwyeradam: Looking at runner this `header.html` isn't actually used. We should probably get rid of it. Everything now uses header2.html. The only surveys using header.html are census but they use an overwritten one in the census theme.

nealedj: Ah yeah - there's a card for that in the backlog.

app/templates/confirmation.html
dwyeradam: @nealedj How would I identify these? When are they used?

app/validation/validators.py
dwyeradam: @nealedj We would actually need an error message for each combination or year/years, month/months, day/days which is huge. 

Enter a reporting period greater than or equal to %(years)s year.
Enter a reporting period greater than or equal to %(years)s years.
Enter a reporting period greater than or equal to %(years)s year, %(months) month.
Enter a reporting period greater than or equal to %(years)s years, %(months) months.
Enter a reporting period greater than or equal to %(years)s year, %(months) months.
Enter a reporting period greater than or equal to %(years)s years, %(months) month.
Enter a reporting period greater than or equal to %(years)s years, %(months)s months, %(days)s days.
...................

nealedj: I guess this will be a trade off then. Either we stop compiling these messages and add _a lot_ of subtly different error messages. Or we accept that the translation of these validation messages will be limited.

Since I don't speak any of the other languages that we'll need to support I don't really know how significant this is.

dwyeradam:
@nealedj I've hopefully addressed all your comments except those for `validators.py`. We'll need some discussion around that. The issue is we're trying to translate dynamic error messages which we always knew would be an issue.

app/validation/validators.py
nealedj: Yeah let's break that out into a separate job in that case.

I think it's worth looking at the pluralization of year(s), month(s) and day(s) though. We can make that a little more translatable with `ngettext`

app/validation/validators.py
```diff
@@ -296,29 +296,26 @@ def _is_first_relative_delta_largest(relativedelta1, relativedelta2):
...
-            'month': [_('months'), _('month')],
-            'day': [_('days'), _('day')]
-        }
-        plural = unit_plurals[unit][(1 if number == 1 else 0)]
 
         if error_message != '':
             error_message = error_message + ', '
 
-        error_message = error_message + str(number) + ' ' + _(plural)
+        error_message = error_message + str(number) + ' ' + unit
```
nealedj: I'd suggest including the count number with the translation usually instead of concatenating it.

```
error_message += ngettext('%(num)s month', '%(num)s months', number)
```

This approach lets languages do the equivalent of saying "a year" instead of "1 year".

Just something to consider if it can be done without compromising code readability?

nealedj (approved):

nealedj:
@dwyeradam looks good to squash

PR #1624, opened 2018-06-08
Eq 2090 buttons links

boxadesign:
### What is the context of this PR?
Updated runner to use the latest release of the Pattern Library and updated templates to use latest button styles and also all focus elements.

Have also added some accessibility improvements (ARIA etc) to the repeating answers component.

Please note that this breaks the summary page. The summary page required changes are currently in progress. Do not merge this branch until the summary pages work is merged in. (https://trello.com/c/KCYOLcBk/1297-apply-new-summary-design-surveys-without-sections-m)

### How to review 

Check that the survey journey displays correctly.

PR #1649, opened 2018-06-26
Ability to have question definitions

MebinAbraham:
Implements question definitions.
[Card](https://trello.com/c/FOZ9wf4A/1950-definitions-enhanced-hideable-description-m) for context. To match Will's design, you need to update runner to pattern lib [v1.4.1](https://github.com/ONSdigital/sdc-global-design-patterns/releases/tag/v1.4.1)

### How to review 
1. Ensure definitions look and behave as expected.


scripts/test_schemas.sh
```diff
@@ -6,8 +6,8 @@ if [ "$1" == "--local" ] || [ "$2" == "--local" ]; then
 fi
 
 if [ "$run_docker" == true ]; then
-    docker pull onsdigital/eq-schema-validator
-    validator="$(docker run -d -p 5001:5000 onsdigital/eq-schema-validator)"
+    docker pull onsdigital/eq-schema-validator:eq-1950-definitions
+    validator="$(docker run -d -p 5001:5000 onsdigital/eq-schema-validator:eq-1950-definitions)"
     sleep 3
```
MebinAbraham: Reminder to revert once validator PR is in.

data/en/mbs_0205.json
```diff
@@ -504,4 +504,4 @@
             }]
         }
     ]
-}
+}
```
insacuri, line 507: Is there a reason you're removing all the newline characters from each of these files?
If not, re-add it to each of the files you've removed it from.

insacuri (changes requested):

insacuri:
Also, master needs to be merged into this branch

data/en/mbs_0205.json
MebinAbraham: @insacuri We don't use newline characters for our schemas. They get automatically removed if you run `yarn format`.

nealedj (changes requested):
Looks really good.

Only thing I noted was that the :focus style in survey runner is showing a blue background. In pattern library the focus style is an orange outline.

Is this an intentional difference?

data/en/mbs_0205.json
dwyeradam: They shouldn't be in this PR though.

MebinAbraham:
@nealedj That's because I've raised this PR on `v1.2.0` of pattern library since #1624 has not gone in yet. If you change it to `v1.4.1` it should match the pattern library.

data/en/mbs_0205.json
MebinAbraham: I think for minor things like this, does it really matter? It shouldn't have gone into master with those changes in the first place. But I ran `yarn format` to format my json schema but I feel it is counter-intuitive for me to add lines back in because it was missed out before.

I can add the lines back in, I don't mind but we need to either get rid of `yarn format` or make it a necessary check when linting.

nealedj (approved):

data/en/mbs_0205.json
insacuri: If we ran a linter over it and it was supposed to remove them, then it's fine being in this PR.  You're right, they shouldn't have gone into master like this anyway so we might as well sort it out while we're here

insacuri (approved):

MebinAbraham:
https://github.com/ONSdigital/eq-schema-validator/pull/58 PR needs to go in first before this can be merged.

PR #1633, opened 2018-06-14
Pass current date for census intro page

shahi645:
### What is the context of this PR?
Census intro page has a hard coded date for 29 November 2017. We wanted to change that so it is the current date

### How to review 
Launch Census household survey. On the intro page where it states `Fill it in as soon as possible, but answer it about`, is it followed by todays date?

PR #1652, opened 2018-06-28
Update date on Census intro page

kierenbeckett:
We want to hard code the census intro date

### How to review 
Start a census survey, see the date on the intro page is 25 June 2018

### DIFF
--- a/app/jinja_filters.py
+++ b/app/jinja_filters.py
@@ -270,11 +270,6 @@ def get_answer_label(context, answer_id, question_id):
 
 
 @blueprint.app_context_processor
-def get_current_date_processor():
-    return dict(get_current_date=get_current_date)
-
-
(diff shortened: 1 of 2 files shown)


MebinAbraham (approved):

jonnyshaw89 (changes requested):
We should also remove the `get_current_date_processor()` which was added to support what we are reverting here
https://github.com/ONSdigital/eq-survey-runner/pull/1633/files

charlotte-hirst:
is the current date not a useful filter to keep?

jonnyshaw89:
Any code we don't use we shouldn't keep. 
it adds testing overhead. also its easy to add back when we need it

jonnyshaw89 (approved):

PR #1656, opened 2018-06-28
Fix displaying of submitted response copy on summary page

kierenbeckett:
When viewing the summary page of a schema with submitted responses, there's supposed to be a bit of copy telling the user they can print/copy their answers

### How to review 
In `dev_settings.sh` set `EQ_DYNAMODB_ENABLED=True`. Run `test_view_submitted_response.json` and go through to the summary, at the bottom you should see the copy 'You will have the opportunity...'.  Try a different survey and see that the copy isn't there.

### DIFF
--- a/app/templates/summary.html
+++ b/app/templates/summary.html
@@ -37,8 +37,8 @@
 
 {% include theme('partials/summary/summary.html') %}
 
-{% if content.is_view_submission_response_enabled %}
-  <div class="u-mb-s u-mb-m@s">
+{% if content.summary.is_view_submission_response_enabled %}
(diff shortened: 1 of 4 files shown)


dwyeradam (approved):

nealedj (changes requested):
Could we add an assertion to the "Submitted Responses" functional test to make sure this content is getting rendered?

nealedj (approved):

dwyeradam:
@kierenbeckett once the conflicts are resolved this can be squashed and merged.

PR #1657, opened 2018-07-02
Bilingual: set the current locale in survey runner based on language_code

kierenbeckett:
Requires #1660 for the tests to reliably pass.

We want babel to use the language specified in the metadata, rather than a global default

### How to review 
Run `test_language_cy.json`, the template text (not survey question text) should be in English.  Change `language_code` to `cy` in survey runner and run again, the template text should be in Welsh.

### DIFF
--- a/app/authentication/authenticator.py
+++ b/app/authentication/authenticator.py
@@ -2,7 +2,7 @@ from uuid import uuid4
 
 from blinker import ANY
 from flask import session as cookie_session, current_app
-from flask_login import LoginManager, user_logged_out
+from flask_login import LoginManager, user_logged_out, login_user
 from sdc.crypto.decrypter import decrypt
 from structlog import get_logger
 
@@ -110,6 +110,11 @@ def store_session(metadata):
     questionnaire_store = get_questionnaire_store(user_id, user_ik)
     questionnaire_store.metadata = metadata
     questionnaire_store.add_or_update()
+
+    # Set the user in Flask such that anyone calling current_user for the duration of this request doesn't have to
+    # load it from the db.
+    login_user(User(user_id, user_ik))
+
     logger.info('user authenticated')
 
 
--- a/app/authentication/user.py
+++ b/app/authentication/user.py
@@ -8,3 +8,6 @@ class User(UserMixin):
             self.user_ik = user_ik
         else:
             raise ValueError('No user_id or user_ik found in session')
+
+    def get_id(self):
+        return self.user_id
--- a/app/setup.py
+++ b/app/setup.py
@@ -24,6 +24,7 @@ from app.authentication.authenticator import login_manager
 from app.authentication.cookie_session import SHA256SecureCookieSessionInterface
 from app.authentication.user_id_generator import UserIDGenerator
 from app.data_model.models import EQSession, QuestionnaireState, db
+from app.globals import get_session_store
 from app.keys import KEY_PURPOSE_SUBMISSION
 from app.new_relic import setup_newrelic
 from app.secrets import SecretStore, validate_required_secrets
@@ -324,6 +325,20 @@ def setup_secure_cookies(application):
 def setup_babel(application):
     application.babel = Babel(application)
     application.jinja_env.add_extension('jinja2.ext.i18n')
+
+    @application.babel.localeselector
+    def get_locale():  # pylint: disable=unused-variable
+        session = get_session_store()
+
+        if not session:
+            return None
+
+        return session.session_data.language_code
+
+    @application.babel.timezoneselector
+    def get_timezone():  # pylint: disable=unused-variable
+        # For now regardless of locale we will show times in GMT/BST
+        return 'Europe/London'
 
 
 def add_safe_health_check(application):
--- a/app/utilities/schema.py
+++ b/app/utilities/schema.py
@@ -5,13 +5,12 @@ import simplejson as json
 from structlog import get_logger
 from werkzeug.exceptions import NotFound
 
-from app.questionnaire.questionnaire_schema import QuestionnaireSchema
+from app.questionnaire.questionnaire_schema import QuestionnaireSchema, DEFAULT_LANGUAGE_CODE
 from app.setup import cache
 
 logger = get_logger()
 
 DEFAULT_SCHEMA_DIR = 'data'
-DEFAULT_LANGUAGE_CODE = 'en'
 
 
 def load_schema_from_metadata(metadata):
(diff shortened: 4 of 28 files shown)


MebinAbraham (commented):
Is this to get `language_code` from schema metadata or the token?
Dw just got confused by the word metadata. I guess this is for the value in the `jwt token` as there is no reason you would define it in the schema itself.

kierenbeckett:
IRL: yes, we will be getting the language code out of the metadata in the JTI token, we won't define it as a required metadata field in schemas and will default to `en` rather than requiring it to be set.

app/authentication/user.py
```diff
@@ -4,6 +4,7 @@
 class User(UserMixin):
     def __init__(self, user_id, user_ik):
         if user_id and user_ik:
+            self.id = user_id
```
nealedj: What's the reason for this? Is it for equality checking of `User` objects? In that case it might be clearer to override `get_id`?

app/globals.py
```diff
@@ -45,7 +45,7 @@ def create_session_store(eq_session_id, user_id, user_ik, session_data):
 
 
 def get_metadata(user):
-    if user.is_anonymous:
+    if not user or user.is_anonymous:
```
nealedj: Under what conditions will the `user` argument evaluate to false?

nealedj (changes requested):
Just added a couple of points whilst I was passing by.

Could you maybe rebase the changes you made for fixing tests into their own commit with a message explaining what problem they fixed? Just to help with source history as those changes aren't related to the PR itself.

app/authentication/user.py
kierenbeckett: `login_user` calls `get_id` which requires an `id` attribute.  I'll override it instead to make it cleaner.

kierenbeckett:
For whatever reason these changes make the CSRF test failures a more likely. I'll pull the fix on into a separate PR, we'll probably have to get that merged before this can go out.

app/globals.py
kierenbeckett: A whole bunch of tests fail without this and it seemed reasonable to be defensive.  On closer inspection this might be more of a fault with the test setup code than the app code, will see what I can do.

MebinAbraham:
As a future card, we should probably validate the value for `language` against babel and fail gracefully instead of the default flask exception.

MebinAbraham (changes requested):
The thank you page is no longer being translated. And I suspect the answers after submissions as well.

jonnyshaw89:
We should use the `language_code` from session_store as it is available after submission.

.gitignore
```diff
@@ -81,8 +81,6 @@ htmlcov/
 
 *errorShots*
 
-data/cy
```
dwyeradam: Why? Our Welsh translated schemas gets put in here. This risks a poorly translated schema getting into master.

app/setup.py
```diff
@@ -325,6 +326,20 @@ def setup_babel(application):
     application.babel = Babel(application)
     application.jinja_env.add_extension('jinja2.ext.i18n')
 
+    @application.babel.localeselector
+    def get_locale():  # pylint: disable=unused-variable
+        session = get_session_store()
+
+        if not session:
+            return None
```
dwyeradam, line 334: Shall we default to 'en' here?

app/authentication/authenticator.py
```diff
@@ -110,6 +110,9 @@ def store_session(metadata):
     questionnaire_store = get_questionnaire_store(user_id, user_ik)
     questionnaire_store.metadata = metadata
     questionnaire_store.add_or_update()
+
+    login_user(User(user_id, user_ik))
```
dwyeradam: I don't think we need this now we're getting values from session_store.

app/authentication/user.py
```diff
@@ -8,3 +8,6 @@ def __init__(self, user_id, user_ik):
             self.user_ik = user_ik
         else:
             raise ValueError('No user_id or user_ik found in session')
+
+    def get_id(self):
+        return self.user_id
```
dwyeradam, line 13: I don't think we need this anymore but may be good to leave in for consistency across the user_id?

nealedj (approved):

nealedj:
LGTM - @kierenbeckett could you also raise a PR in the go launcher to remove `test_language_cy`?

app/utilities/schema.py
```diff
@@ -26,7 +25,7 @@ def load_schema_from_session_data(session_data):
 
 @cache.memoize()
 def load_schema_from_params(eq_id, form_type, language_code=DEFAULT_LANGUAGE_CODE):
-    return QuestionnaireSchema(_load_schema_file('{}_{}.json'.format(eq_id, form_type), language_code))
+    return QuestionnaireSchema(_load_schema_file('{}_{}.json'.format(eq_id, form_type), language_code), language_code)
```
dwyeradam: My worry here is that we're sometimes getting language_code from metadata. When/If we come to work on the card that allows for swapping of language part way through a survey we would change the session_data language_code to drive this but I wouldn't think we'd want to change the metadata. We could therefore get inconsistency.

We should at least reduce the times that `load_schema_from_metadata` is called and actually using metadata to minimise this. I think we can do this without breaking the tests that this is here to fix but stop the possibility of inconsistency when loading the schema in live conditions.

dwyeradam (changes requested):
I may be confusing matters further but lets discuss.

.gitignore
nealedj: We need this for the `test_language` schema that's use in a functional else. But perhaps it's worth keeping the ignore in place and just forcing that file in as an exception?

kierenbeckett: Will the properly translated schemas not be put here (assuming we've not switched to a registry by then)?  For now the only welsh schema is of `test_language`, which really helps with the testing of the new functionality and doesn't need a perfect translation.

app/setup.py
kierenbeckett: Returning `None` tells babel to use it's global default, which is `en`.

app/authentication/authenticator.py
kierenbeckett: Technically we don't, but this is still worth having as it saves a db call.

app/authentication/user.py
kierenbeckett: Also technically no, but according to the spec Flask user classes are supposed to set `self.id` or implement `get_id`.

app/authentication/authenticator.py
nealedj: Maybe add a comment just to indicate that the reasoning behind this line is non-functional?

app/utilities/schema.py
kierenbeckett: I've removed most usages of `load_schema_from_metadata`. The ones that are left are:

* At initial login, as we need to get the schema before we create the session.  This is ok as at this point the language code on the metadata and session will always match.
* In `/flush` where we don't have a session

.gitignore
dwyeradam: The properly translated schemas will be put here. At the moment however we don't have any properly translated schemas. We do however have some scripts (see translate_schemas.sh called by build.sh) that will create some very poorly translated census schemas that may end up getting into master if we take this ignore out.

kierenbeckett: `translate_schemas.sh` isn't working very well right now.  There's a card to fix it so until then I'll leave `cy/test_language.json` committed for the benefit of the tests and keep the line in `.gitignore`.

dwyeradam (approved):

kierenbeckett:
@MebinAbraham the thank you page is fixed now that we're getting the language code from the session

MebinAbraham (approved):

PR #1660, opened 2018-07-02
Fix CSRF missing errors that happen occasionally in tests

kierenbeckett:
Intermittently some of the star wars integration tests will fail with CSRF missing errors

### How to review 
It's hard to reproduce on master, though happens reliably on #1657.  Maybe just review the code, or add a sleep into the star wars test to make it fail more consistently and see that this change stops it happening.

### DIFF
--- a/tests/integration/session/test_timeout.py
+++ b/tests/integration/session/test_timeout.py
@@ -11,6 +11,11 @@ class TestTimeout(IntegrationTestCase):
         settings.EQ_SESSION_TIMEOUT_GRACE_PERIOD_SECONDS = 0
         super().setUp()
 
+    def tearDown(self):
+        settings.EQ_SESSION_TIMEOUT_SECONDS = 45 * 60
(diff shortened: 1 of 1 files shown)


tests/integration/session/test_timeout.py
```diff
@@ -11,6 +11,9 @@ def setUp(self):
         settings.EQ_SESSION_TIMEOUT_GRACE_PERIOD_SECONDS = 0
         super().setUp()
 
+    def tearDown(self):
+        settings.EQ_SESSION_TIMEOUT_SECONDS = 3600
```
nealedj: Could you call `tearDown` on the base please? Although the base doesn't implement it at the moment the DynamoDB PR introduces an implementation on it.

nealedj (changes requested):

nealedj (approved):

dwyeradam (approved):

PR #1661, opened 2018-07-03
Added JS for mutually exclusive checkboxes

MebinAbraham:
Adds the JS needed for the mutually exclusive checkbox to work. When JS is enabled, server-side validation won't be hit.
[Card](https://trello.com/c/aCpgM0A6/2177-eq-survey-runner-1627-completion-status-incorrectly-cleared-issue) for context.
[Pattern Lib implementation](https://sdc-global-design-patterns.netlify.com/components/detail/mutually-exclusive.html)
Test Schema - `test_checkbox_mutually_exclusive.json`
### How to review 
- Ensure mutually exclusive checkbox works as expected.
- Check both the visible and voice-over functionality works on different browsers, IE 9 and above.


tests/integration/test_checkbox_mutually_exclusive.py
```diff
@@ -0,0 +1,35 @@
+from tests.integration.integration_test_case import IntegrationTestCase
+
+
+class TestCheckboxMutuallyExclusive(IntegrationTestCase):
+
```
nealedj: I think it's worth adding a docstring here to explain what's being tested. Otherwise it isn't clear that this test exists for javascript.

The test file itself doesn't really make sense in the top level integrations directory. I'm not sure where best to put it though. Perhaps either alongside `test_render_percentage_widget.py` and rename widget/ to component/ or in a separate directory - form/?

nealedj (changes requested):

MebinAbraham:
@nealedj Ready for re-review.

tests/functional/spec/components/checkbox-mutually-exclusive.spec.js
```diff
@@ -1,63 +1,85 @@
...
+				.click(MandatoryCheckboxPage.pineapple())
+				.click(MandatoryCheckboxPage.none())
+
+				// Then
+				.isSelected(MandatoryCheckboxPage.none())
+
+				// When
+				.click(MandatoryCheckboxPage.submit())
+
+				// Then
```
jonnyshaw89: Shouldn't it be "given, when, then" not "When, Then, When, Then"

jonnyshaw89 (commented):
Looks good, just the functional tests are a bit confusing to read

nealedj (approved):

MebinAbraham:
@jonnyshaw89 Ready for re-review.

MebinAbraham:
@jonnyshaw89 Ready for re-review again.

tests/functional/spec/components/checkbox-mutually-exclusive.spec.js
```diff
@@ -1,62 +1,94 @@
...
+
       return browser
+        // Given
         .click(MandatoryCheckboxPage.cheese())
         .click(MandatoryCheckboxPage.ham())
         .click(MandatoryCheckboxPage.pineapple())
+
+        // When
         .click(MandatoryCheckboxPage.none())
         .click(MandatoryCheckboxPage.submit())
```
dwyeradam: We should check after the exclusive value is selected that the others are deselected rather than just relying on the Summary page not including them.

dwyeradam (changes requested):
One little change to the functional test but otherwise looking good.

tests/functional/spec/components/checkbox-mutually-exclusive.spec.js
MebinAbraham: @dwyeradam Ready for re-review.

jonnyshaw89 (approved):

dwyeradam (approved):

PR #1662, opened 2018-07-03
Add instructions to readme to downgrade node

insacuri:
Currently, if you follow the README as is, then survey runner won't work properly.  This PR adds the extra step that's needed until survey runner is compatible with the latest version of node.

### How to review 
N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,11 @@ docker build -t onsdigital/eq-survey-runner-unit-tests -f Dockerfile.test .
 docker run onsdigital/eq-survey-runner-unit-tests
 ```
 
+To run the unit tests locally:
+```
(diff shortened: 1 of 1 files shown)


nealedj (approved):

MebinAbraham:
While you are here fancy removing line 75? https://github.com/ONSdigital/eq-survey-runner/blob/4af5a1fb192d47233c741bee6f539fa01a5839e9/README.md#L73-L76

MebinAbraham (approved):
Apart from that optional task that you totally don't have to do but would be nice if you did, looks good 👍 

Can squash also.

nealedj:
@insacuri Good to squash (rebase/squash locally and force push, we don't use github's auto squash)

PR #1663, opened 2018-07-03
Add set condition and unit test

insacuri:
The inverse to 'not set' didn't exist so this creates it so that future routing can check 'did you put anything in the box'.

### How to review 
There aren't any schemas that use this currently so it's just a code quality and unit test check


nealedj (approved):

nealedj (changes requested):
I think we're missing a test schema here.

We have test schemas for various routing rules but nothing for `not set` (it's used in other schemas but they aren't explicitly testing that).

Since this is a feature we need to have it in at least one test schema to verify the functionality.

Could you add `test_routing_set.json` and `test_routing_not_set.json` schemas along with basic functional tests for them?

MebinAbraham:
Apart from David's comment, looks good 👍

MebinAbraham (changes requested):
You will also need to make a validator change to allow `set` as one of the options for `condition` before this can be merged.

https://github.com/ONSdigital/eq-schema-validator/blob/e25aa3a06a693374e1a5a2d37c9d7fd5b8468786/schemas/common_definitions.json#L120

MebinAbraham (approved):
Validator PR merged.

README.md
```diff
@@ -164,6 +164,10 @@ To run a single test, add `@watch` into the name of any `describe` or `it` funct
 
 `yarn test_functional --watch`
 
+An example of adding `@watch` looks like this:  
+`describe('@watch Skip Conditions', function() {...}` or  
+`it('@watch Given this is a test', function() {...}`
+
```
dwyeradam, line 170: 👍

nealedj (approved):
Nicely done. Helpful README addition in there too 👍

PR #1666, opened 2018-07-09
Eq 2174 lms proxy titles

AndrewTorrance:
lms had two identical paths one based on if the person in question was answering , the other if someone was answering on their behalf. This change uses titles in the schema so that we only swap the question title.
When implementing this it was noticed that submitting a schema with titles and more than one answer in a repeating group caused a 500 . The cause was dump did not set g.schema . dump was changed to set g.schema .  

### How to review 
Run lms_1 via go launcher , add in at least 2 people one of which you should be answering on behalf of 
Validate that the question title changes from first person to third person if you are answering on behalf of.
Then before submitting dump the submission ( http://localhost:5000/dump/submission) and validate that valid Json not a 500 is returned )


data/en/lms_1.json
```diff
@@ -835,11 +917,11 @@
                                 "mandatory": true,
                                 "type": "Radio",
                                 "options": [{
-                                        "label": "I&apos;d like to correct their date of birth",
+                                        "label": "I’d like to correct the date of birth",
```
kierenbeckett, line 917: Is this safe for the rest of our stack? e.g. do all browsers correctly treat our responses as utf-8?

app/views/dump.py
```diff
@@ -27,8 +27,8 @@ def dump_answers():
 def dump_submission():
     answer_store = get_answer_store(current_user)
     metadata = get_metadata(current_user)
-    schema = load_schema_from_metadata(metadata)
+    g.schema = load_schema_from_metadata(metadata)
```
kierenbeckett: What's the benefit of doing this?

kierenbeckett, line 30: This works, but it doesn't seem ideal that `get_full_routing_path` relies on request variables that have to be manually set before calling it.  Is it non-trivial to refactor it?  Could `g.schema` perhaps be set more automatically e.g. as part of retrieving a session?

nealedj: The pattern to retrieve any other global is via a `get_x` method in `app.globals`. These access `current_user` as an argument. I think it'd be much safer to only access `g.schema` via one of these methods.

_aside_

The preferable solution would be for the function in question in `rules.py` to have an explicit dependency so `dump_submission` would be able to control directly the schema that's set.

As a rule I don't think it's a good idea for us to use globals outside of view functions - there are only a few exceptions to this which I think should be refactored.

data/en/lms_1.json
nealedj: We're setting charset utf-8 and are using these quotes elsewhere. @armstrongb I'm guessing you aren't aware of any issues with our smart quotes not getting rendered correctly?

armstrongb: I haven't had any reports of them being rendered incorrectly and wouldn't expect them to when using utf-8.

AndrewTorrance:
Closing this as I merged incorrectly into this branch . Its easier to create a new branch and do this walk of shame, than unpick the merges correctly

PR #1668, opened 2018-07-11
Eq 2174 lms using proxy titles

AndrewTorrance:
Modified LMS_1.Json to use titles instead of duplicating two arms of the survey. In doing so it was noticed that dumping did not work . This was fixed with a change to dump.py. This is known not to be ideal ( setting a global for use lower down) However it has been agreed to look at the use of globals as a separate ticket.
(Note this replaces an earlier PR where I messed up a merge.)

### How to review 
Run lms_1 with multiple users and verify functionality. Check submission works prior to submitting

### DIFF
--- a/data/en/lms_1.json
+++ b/data/en/lms_1.json
@@ -394,28 +394,28 @@
                             }]
                         }],
                         "routing_rules": [{
-                                "goto": {
-                                    "block": "date-of-birth",
-                                    "when": [{
-                                        "id": "proxy-check-answer",
-                                        "condition": "equals",
-                                        "value": "proxy"
-                                    }]
-                                }
-                            },
-                            {
-                                "goto": {
-                                    "block": "date-of-birth-no-proxy"
-                                }
+                            "goto": {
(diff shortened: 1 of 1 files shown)


kierenbeckett (approved):

app/views/dump.py
```diff
@@ -28,8 +28,8 @@ def dump_submission():
     answer_store = get_answer_store(current_user)
     metadata = get_metadata(current_user)
     session_data = get_session_store().session_data
-    schema = load_schema_from_session_data(session_data)
+    g.schema = load_schema_from_session_data(session_data)
```
jonnyshaw89: why?

jonnyshaw89: We should make this a separate PR as its unrelated to the change

AndrewTorrance: Agree, changing it

nealedj (approved):

nealedj:
LGTM - @AndrewTorrance could you squash those 3 commits and rebase this branch on master?

PR #1669, opened 2018-07-11
Add translation tags to html

dwyeradam:
The Links and Buttons PR removed some static translation tags from a few conflicting files.

### How to review 
New messages.pot file should include the same static content as before including the newly named Save and continue button.
Compare with https://github.com/ONSdigital/eq-survey-runner/pull/1643

### DIFF
--- a/app/templates/confirmation.html
+++ b/app/templates/confirmation.html
@@ -5,8 +5,8 @@
 {% block submit_button %}
 
 {% if content.summary and content.summary.summary_type == 'SectionSummary' %}
-    <button class="btn btn--loader js-btn-submit" data-qa="btn-submit">Continue</button>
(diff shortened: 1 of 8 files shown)


jonnyshaw89 (approved):

MebinAbraham (commented):
https://github.com/ONSdigital/eq-survey-runner/blob/6819e38c0dca2ae4363ec4c0c7b44b157df6366d/app/templates/interstitial.html#L5

Not sure if it was from before or not.

kierenbeckett (approved):

dwyeradam:
@MebinAbraham, good spot. This was missed from initial translation. However we don't have the concept of section numbers anymore so actually this line will never get hit. We should remove it.

MebinAbraham (approved):

PR #1672, opened 2018-07-16
Format schema validation output in test_schemas.sh

benhowes:
This PR adds a prettifying step on the JSON response from the `eq-schema-validator`, in order to make the output more readable. This has been added to speed up reviewing and fixing schemas (or changing schemas as I was).

Simply makes use of [`json.tool`](https://docs.python.org/2/library/json.html) in the python std library to format the JSON response from the validator. From looking in to it - this works on python2.6+, so it's not expected that there will be dev environments that don't meet this condition!

### How to review 
Try validating an invalid schema - e.g. by making a change to a questionnaire so that it's no longer valid to the schema and running the validator against it. e.g. edit `data/en/test_navigation.json`and run ` ./scripts/test_schemas.sh ./data/en/test_navigation.json` to see the improved output.

### DIFF
--- a/scripts/test_schemas.sh
+++ b/scripts/test_schemas.sh
@@ -52,7 +52,7 @@ for schema in $(find $file_path -name '*.json'); do
 
     result="$(curl -s -w 'HTTPSTATUS:%{http_code}' -X POST -H "Content-Type: application/json" -d @"$schema" http://localhost:5001/validate | tr -d '\n')"
     result_response="${result//*HTTPSTATUS:/}"
-    result_body="${result//HTTPSTATUS:*/}"
+    result_body=$(echo "${result//HTTPSTATUS:*/}" | python -m json.tool)
 
     if [ "$result_response" == "200" ] && [ "$result_body" == "{}" ]; then
         echo -e "${green}$schema - PASSED${default}"


nealedj (approved):

MebinAbraham (commented):
Nice but does it not format the whole block, ie the python dict?.
```
./data/en/0_rogue_one.json - FAILED
HTTP Status @ /validate: [400]
Error: [{
    "errors": {
        "message": "{'type': 'Question', 'id': 'block-1', 'description': '', 'questions': [{'answers': [{'id': 'character-answer', 'label': 'Who do you want to know more about?', 'mandatory': True, 'options': [{'label': 'Jyn Erso', 'value': 'Jyn Erso'}, {'label': 'Cassian Andor', 'value': 'Cassian Andor'}, {'label': 'Bodhi Rook', 'value': 'Bodhi Rook'}, {'label': 'Orson Krennic', 'value': 'Orson Krennic'}], 'q_code': '1', 'type': 'Radio', 'validation': {'messages': {}}}], 'description': '', 'id': 'character-question', 'title': '', 'type': 'General'}], 'title': 'Who do you want to know more about?', 'routing_rules': [{'goto': {'block': 'jyn-erso-like-this-page', 'when': [{'id': 'character-answer', 'condition': 'equals', 'value': 'Jyn Erso'}]}}, {'goto': {'block': 'cassian-andor-like-this-page', 'when': [{'id': 'character-answer', 'condition': 'equals', 'value': 'Cassian Andor'}]}}, {'gotoo': {'block': 'bodhi-rook-like-this-page', 'when': [{'id': 'character-answer', 'condition': 'equals', 'value': 'Bodhi Rook'}]}}, {'goto': {'block': 'orson-krennic-like-this-page', 'when': [{'id': 'character-answer', 'condition': 'equals', 'value': 'Orson Krennic'}]}}]} is not valid under any of the given schemas",
        "path": "deque(['sections', 0, 'groups', 0, 'blocks', 1])"
    }
}]
```

MebinAbraham (approved):

nealedj:
@MebinAbraham the `message` key is a string so it won't get formatted. We could change that in validator.

dwyeradam:
@benhowes Before we merge this can you shorten the commit message so it fits on one line. Any extra information goes in the description within the commit.
Commit message is what you've done. Along the lines of "Format validator output"
Description can add how if relevant. In this case probably not but otherwise "Using json.tool to format message.".

benhowes:
@dwyeradam I've updated that commit message - will keep the rules in mind, thanks!

PR #1675, opened 2018-07-17
Remove empty answer lists to work with new validator changes

jonnyshaw89:
### What is the context of this PR?
Schema changes to support https://github.com/ONSdigital/eq-schema-validator/pull/61

### How to review 
Do all the schemas pass against validator changes?

PR #1679, opened 2018-07-18
Fix keyerror when answers missing from content question.

samiwel:
A change recently went in to publisher to create confirmation pages using a content question type.

This causes a 500 error to be thrown in survey runner due to the question missing the answers key when the dependencies are enumerated within survey runner.

This PR aims to fix this issue by defaulting answers to an empty list if the key does not exist.

### How to review 
Review of code change is required.
All tests and checks should continue to pass.


samiwel:
This should be closed as it is already covered by #1675

PR #1680, opened 2018-07-18
Ignore CVE-2017-18342 for now

kierenbeckett:
There is not a patched version ready yet and we don't load arbitrary YAML files anyway


jonnyshaw89:
@kierenbeckett i think we are going to have to do this as part of https://github.com/ONSdigital/eq-survey-runner/pull/1675 now `eq-schema-validator` change has been merged

PR #1684, opened 2018-07-20
Update various code style config to match

kierenbeckett:
Our lines are 120. json files are intended with 4 spaces, html 2.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -193,7 +193,7 @@ max-nested-blocks=5
 [FORMAT]
 
 # Maximum number of characters on a single line.
-max-line-length=150
+max-line-length=160
 
 # Regexp for a line that is allowed to be longer than the limit.
 ignore-long-lines=^\s*(# )?<?https?://\S+>?$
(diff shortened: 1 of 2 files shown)


benhowes:
@kierenbeckett where did the line lengths of 120 come from? The `.pylintrc` has `max-line-length=150` at the moment

kierenbeckett:
Good point, I was looking at pycharm warnings.  Our linter actually seems to look at `setup.cfg` so I've updated `.pylintrc` to match.

benhowes (approved):
LGTM

.pylintrc
```diff
@@ -193,7 +193,7 @@ max-nested-blocks=5
 [FORMAT]
 
 # Maximum number of characters on a single line.
-max-line-length=150
+max-line-length=160
```
dcdarrell9, line 196: Should we be doing 120, 150, 160? Do we know what we want as a team?

jonnyshaw89 (approved):

jonnyshaw89:
@kierenbeckett update?

.pylintrc
kierenbeckett: 160 is what the travis currently fails at.  Let's start by making everything else consistent with that, we can discuss if we want to change separately.

PR #1687, opened 2018-07-20
support skipping a block in repeating group based on value from an ea…

AndrewTorrance:
Block Skip conditions not working if independent variable is in a repeating group 

For LMS we have to ascertain how many people are in the house , then for each we ask them they their age , and then for each if that age is over 16 we ask them personal questions. Then if they are over 16 we ask  them questions about working. This means that we need to gather the people in a house. (Repeating Answer) then for each we determine their age in a repeating group . Then we have a dependancy on that repeating group from another repeating group. It was this that was not working.

This code also highlighted another issue that is since skip conditions are evaluated on entry , we can have the situation where the first or many group instances may be skipped, so now we cannot assume that the first group instance in a repeating group is zero as that may be skipped.

### How to review 
Run test_skip_conditions_on_blocks_repeating_group using go launcher. Enter 3 people.  When you are asked for date of birth do not enter a value for some people . Validate that you only get asked the 
additional question 'Some question for <name>' for those people who you entered a date of birth for. 
Run this using a value for the first user and not entering a value for the first user. This is because the existing code defaulted to group instance zero . 

### Checklist

* [  ] New static content marked up for translation - NA
* [  ] Newly defined schema content included in eq-translations repo - NA

### DIFF
--- a/app/questionnaire/path_finder.py
+++ b/app/questionnaire/path_finder.py
@@ -45,8 +45,8 @@ class PathFinder:
         first_groups = self._get_first_group_in_section()
 
         for group in self.schema.groups:
+            first_block_in_group = self.schema.get_first_block_id_for_group(group['id'])
             if not this_location:
-                first_block_in_group = self.schema.get_first_block_id_for_group(group['id'])
                 this_location = Location(group['id'], 0, first_block_in_group)
 
             if 'skip_conditions' in group:
@@ -55,15 +55,23 @@ class PathFinder:
 
             no_of_repeats = get_number_of_repeats(group, self.schema, path, self.answer_store)
 
+            first_group_instance_index = None
             for instance_idx in range(0, no_of_repeats):
-                group_blocks = self._build_blocks_for_group(group, instance_idx)
+                group_blocks = list(self._build_blocks_for_group(group, instance_idx))
                 blocks += group_blocks
 
-            if group['id'] in first_groups:
-                this_location = Location(group['id'], 0, group['blocks'][0]['id'])
+                if group_blocks and first_group_instance_index is None:
+                    # get the group instance of the first instance of a block in this group that has not been skipped
+                    first_group_instance_index = group_blocks[0]['group_instance']
 
-            if blocks:
-                path, block_index = self._build_path_within_group(blocks, block_index, this_location, path)
+            all_blocks_skipped = first_group_instance_index is None
+
+            if not all_blocks_skipped:
+                if group['id'] in first_groups:
+                    this_location = Location(group['id'], first_group_instance_index, first_block_in_group)
+
+                if blocks:
+                    path, block_index = self._build_path_within_group(blocks, block_index, this_location, path)
 
         return path
 
@@ -71,7 +79,7 @@ class PathFinder:
         for block in group['blocks']:
             skip_conditions = block.get('skip_conditions')
             if skip_conditions and evaluate_skip_conditions(
-                    skip_conditions, self.metadata, self.answer_store):
+                    skip_conditions, self.metadata, self.answer_store, instance_idx):
                 continue
 
             yield {
--- a/data/en/test_skip_conditions_on_blocks_repeating_group.json
+++ b/data/en/test_skip_conditions_on_blocks_repeating_group.json
@@ -1,0 +1,150 @@
+{
+    "mime_type": "application/json/ons/eq",
+    "schema_version": "0.0.1",
+    "data_version": "0.0.2",
+    "survey_id": "0",
+    "title": "A Test Schema to test skip conditions in a repeating group",
+    "description": "A survey that has an answer in one repeating group being the independent variable for a skip condition on a block in another repeating group ",
+    "theme": "default",
+    "legal_basis": "Voluntary",
+    "form_type": "1",
+    "navigation": {
+        "visible": true
+    },
+    "metadata": {
+        "user_id": {
+            "validator": "string"
+        },
+        "period_id": {
+            "validator": "string"
+        },
+        "ru_name": {
+            "validator": "string"
+        }
+    },
+    "sections": [{
+            "id": "household-section",
+            "title": "About the household",
+            "groups": [{
+                "id": "about-household-group",
+                "title": "About the household",
+                "blocks": [{
+                    "id": "household-composition",
+                    "type": "Question",
+                    "description": "gather names of people",
+                    "questions": [{
+                        "id": "household-composition-question",
+                        "titles": [{
+                            "value": "What are the names of everyone who lives in the household?"
+                        }],
+                        "type": "RepeatingAnswer",
+                        "answers": [{
+                            "id": "first-name",
+                            "label": "First name",
+                            "mandatory": true,
+                            "type": "TextField",
+                            "validation": {
+                                "messages": {
+                                    "MANDATORY_TEXTFIELD": "Please enter a name or remove the person to continue"
+                                }
+                            }
+                        }]
+                    }]
+                }]
+            }]
+        },
+        {
+            "id": "household-members-section",
+            "title_from_answers": [
+                "first-name"
+            ],
+            "groups": [{
+                "id": "household-member-group",
+                "title": "Household Member Details",
+                "routing_rules": [{
+                    "repeat": {
+                        "type": "answer_count",
+                        "answer_id": "first-name"
+                    }
+                }],
+                "blocks": [{
+                    "type": "Question",
+                    "id": "date-of-birth",
+                    "description": "used to gather dob for each person from repeating question",
+                    "questions": [{
+                        "id": "date-of-birth-question",
+                        "titles": [{
+                            "value": "What is <em>{{answers['first-name'][group_instance]}}'s</em> date of birth?"
+                        }],
+                        "type": "General",
+                        "answers": [{
+                            "id": "date-of-birth-answer",
+                            "mandatory": false,
+                            "type": "Date",
+                            "maximum": {
+                                "value": "now"
+                            }
+                        }]
+                    }]
+                }]
+            }]
+        },
+        {
+            "id": "repeating-group",
+            "title_from_answers": [
+                "first-name"
+            ],
+            "groups": [{
+                "id": "dependant-group",
+                "title": " ",
+                "routing_rules": [{
+                    "repeat": {
+                        "type": "answer_count",
+                        "answer_id": "first-name"
+                    }
+                }],
+                "blocks": [{
+                    "id": "additional-question-block",
+                    "type": "Question",
+                    "description": "skip the question if no date of birth set for each person",
+                    "skip_conditions": [{
+                        "when": [{
+                            "id": "date-of-birth-answer",
+                            "condition": "not set"
+                        }]
+                    }],
+                    "questions": [{
+                        "id": "some-additional-question",
+                        "title": "Some question for <em>{{answers['first-name'][group_instance]}}</em>",
+                        "type": "General",
+                        "answers": [{
+                            "id": "some-answer",
+                            "label": "Some answer",
+                            "mandatory": false,
+                            "type": "TextField"
+                        }]
+                    }]
+                }]
+            }]
+        },
+        {
+            "id": "submit-answers",
+            "title": "Submit answers",
+            "groups": [{
+                "id": "questionnaire-completed",
+                "title": "Submit answers",
+                "blocks": [{
+                    "type": "Confirmation",
+                    "id": "confirmation",
+                    "title": "",
+                    "questions": [{
+                        "id": "questionnaire-completed-question",
+                        "title": "Thank you, please submit your answers.",
+                        "type": "Content",
+                        "description": "<p>Please submit your responses by using the <em>Submit answers</em> button below.</p>"
+                    }]
+                }]
+            }]
+        }
+    ]
+}
--- a/tests/app/questionnaire/test_path_finder.py
+++ b/tests/app/questionnaire/test_path_finder.py
@@ -653,6 +653,124 @@ class TestPathFinder(AppContextTestCase):  # pylint: disable=too-many-public-met
 
             self.assertEqual(expected_path, path_finder.get_full_routing_path())
 
+    def test_repeating_group_block_skip_correct_if_independent_answer_is_not_repeating_but_causes_a_skip(self):
+        """ tests blocks are correctly added to path in the case where the blocks in a repeating group have
+        skip conditions and those skip conditions are dependant on a non repeating answer which causes a skip
+        condition to return true.
+        """
+
+        independent_answer = '3'
+
+        expected_path = [
+            Location('about-household-group', 0, 'household-composition'),
+            Location('independent-answer-group', 0, 'non-repeating-question-block'),
+            Location('questionnaire-completed', 0, 'confirmation')
+        ]
+
+        schema = load_schema_from_params('test', 'skip_conditions_from_repeating_group_based_on_non_repeating_answer')
+        answer_store = AnswerStore()
+
+        # Set up some first names
+        answer_store.add(Answer(answer_id='first-name', value='aaa', group_instance=0))
+        answer_store.add(Answer(answer_id='first-name', value='bbb', group_instance=1))
+        answer_store.add(Answer(answer_id='first-name', value='ccc', group_instance=2))
+
+        # Set up the independent answer
+        answer_store.add(Answer(answer_id='an-independent-answer', value=independent_answer, group_instance=0))
+
+        with patch('app.questionnaire.rules._answer_is_in_repeating_group', return_value=False):
+            sut = PathFinder(schema, answer_store=answer_store, metadata={}, completed_blocks=[])
+            sut_path = sut.get_full_routing_path()
+
+            # validate that path returned from path finder matches expected path
+
+            self.assertEqual(expected_path, sut_path)
+
+    def test_repeating_group_block_skip_correct_if_independent_answer_is_not_repeating_but_does_not_cause_a_skip(self):
+        """ tests blocks are correctly added to path in the case where the blocks in a repeating group have
+        skip conditions and those skip conditions are dependant on a non repeating answer which causes a skip
+        condition to return false.
+        """
+
+        independent_answer = 2
+
+        expected_path = [
+            Location('about-household-group', 0, 'household-composition'),
+            Location('independent-answer-group', 0, 'non-repeating-question-block'),
+            Location('dependant-group', 0, 'additional-question-block'),
+            Location('dependant-group', 1, 'additional-question-block'),
+            Location('dependant-group', 2, 'additional-question-block'),
+            Location('questionnaire-completed', 0, 'confirmation')
+        ]
+
+        schema = load_schema_from_params('test',
+                                         'skip_conditions_from_repeating_group_based_on_non_repeating_answer')
+        answer_store = AnswerStore()
+
+        # Set up some first names
+        answer_store.add(Answer(answer_id='first-name', value='aaa', group_instance=0))
+        answer_store.add(Answer(answer_id='first-name', value='bbb', group_instance=1))
+        answer_store.add(Answer(answer_id='first-name', value='ccc', group_instance=2))
+
+        # Set up the independent answer
+        answer_store.add(Answer(answer_id='an-independent-answer', value=independent_answer, group_instance=0))
+
+        with patch('app.questionnaire.rules._answer_is_in_repeating_group', return_value=False):
+            sut = PathFinder(schema, answer_store=answer_store, metadata={}, completed_blocks=[])
+            sut_path = sut.get_full_routing_path()
+
+            # validate that path returned from path finder matches expected path
+
+            self.assertEqual(expected_path, sut_path)
+
+    def test_repeating_group_block_skip_correct_if_independent_answer_is_repeating(self):
+        """ tests blocks are correctly added to path in the case where the blocks have
+        skip conditions and those skip conditions are dependant on a repeating answer
+        tests various combinations of independent answers being set , since we saw different
+        behaviours for group_instance 0 as it was the default"""
+
+        param_list = [(0, 2), (1, 3)]
+
+        with patch('app.questionnaire.rules._answer_is_in_repeating_group', return_value=True):
+            for first_group_instance, second_group_instance in param_list:
+
+                with self.subTest(first_group_instance=first_group_instance,
+                                  second_group_instance=second_group_instance):
+                    expected_path = [
+                        Location('about-household-group', 0, 'household-composition'),
+                        Location('household-member-group', 0, 'date-of-birth'),
+                        Location('household-member-group', 1, 'date-of-birth'),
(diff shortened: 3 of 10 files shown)


tests/functional/spec/features/skipping/skipping_blocks_in_repeating_groups.js
```diff
@@ -0,0 +1,51 @@
...
+        .click(HouseholdCompositionPage.addPerson())
+        .setValue(HouseholdCompositionPage.answer('_2'),'ccc')
+        .click(HouseholdCompositionPage.submit())
+        .setValue(DobPage.day(), 15)
+        .setValue(DobPage.year(), '1962')
+        .selectByValue(DobPage.month(), 2)
+        .click(DobPage.submit())
+        .click(DobPage.submit())
+        .click(DobPage.submit())
+        .getText(AdditionalQuestionPage.questionText()).should.eventually.contain('Some question for aaa');
```
benhowes: You could test that the summary page is the next one after this

app/questionnaire/path_finder.py
```diff
@@ -59,19 +59,32 @@ def build_path(self):
...
-            if blocks:
-                path, block_index = self._build_path_within_group(blocks, block_index, this_location, path)
+                if blocks:
+                    path, block_index = self._build_path_within_group(blocks, block_index, this_location, path)
 
         return path
 
+    @staticmethod
+    def _get_first_group_instance_in_block(group, blocks):
+        """returns the first non skipped group instance of a block else None"""
```
benhowes: "non skipped `group_instance`" (I read this as "an instance of a group" initially, but `group_instance` is used everywhere else for this

app/questionnaire/path_finder.py
```diff
@@ -59,19 +59,32 @@ def build_path(self):
                 group_blocks = self._build_blocks_for_group(group, instance_idx)
                 blocks += group_blocks
 
-            if group['id'] in first_groups:
-                this_location = Location(group['id'], 0, group['blocks'][0]['id'])
+            first_group_instance = self._get_first_group_instance_in_block(group, blocks)
+            if first_group_instance is not None:
```
benhowes: Just `if first_group_instance:` will work nicely here.

app/questionnaire/path_finder.py
```diff
@@ -59,19 +59,32 @@ def build_path(self):
...
+                if group['id'] in first_groups:
+                    this_location = Location(group['id'], first_group_instance, group['blocks'][0]['id'])
```
benhowes: This might be clearer if you just set the location's group_instance rather than make a new location? (unless there's a reason I'm missing!) e.g. `this_location.group_instance = first_group_instance`

app/questionnaire/path_finder.py
```diff
@@ -59,19 +59,32 @@ def build_path(self):
                 group_blocks = self._build_blocks_for_group(group, instance_idx)
                 blocks += group_blocks
```
benhowes: May be worth commenting this `first_group_instance` related block to say why it's required briefly, e.g. "`first_group_instance` is required in the case that the first instance of a repeating group is skipped"?

app/questionnaire/path_finder.py
kierenbeckett: If a block is skipped, do we want to be using the next group instance rather than the next block in the current group instance?

app/questionnaire/path_finder.py
```diff
@@ -59,19 +59,32 @@ def build_path(self):
...
-            if group['id'] in first_groups:
-                this_location = Location(group['id'], 0, group['blocks'][0]['id'])
+            first_group_instance = self._get_first_group_instance_in_block(group, blocks)
+            if first_group_instance is not None:
+                if group['id'] in first_groups:
+                    this_location = Location(group['id'], first_group_instance, group['blocks'][0]['id'])
 
-            if blocks:
-                path, block_index = self._build_path_within_group(blocks, block_index, this_location, path)
+                if blocks:
```
kierenbeckett: Doesn't this still need to be called regardless of whether we have a `first_group_instance`?

app/questionnaire/path_finder.py
```diff
@@ -59,19 +59,32 @@ def build_path(self):
...
 
         return path
 
+    @staticmethod
+    def _get_first_group_instance_in_block(group, blocks):
+        """returns the first non skipped group instance of a block else None"""
+        ret_val = None
+        try:
+            first_block = next(block for block in blocks if block['group_id'] == group['id'])
+            ret_val = first_block['group_instance']
```
kierenbeckett: Maybe just `return` here, no need for `ret_val`

app/questionnaire/path_finder.py
```diff
@@ -59,19 +59,32 @@ def build_path(self):
...
+        except (StopIteration, KeyError):  # Either no group instances (all skipped) or group has no 'id'
+            pass
+        return ret_val
+
     def _build_blocks_for_group(self, group, instance_idx):
         for block in group['blocks']:
             skip_conditions = block.get('skip_conditions')
             if skip_conditions and evaluate_skip_conditions(
-                    skip_conditions, self.metadata, self.answer_store):
+                    skip_conditions, self.metadata, self.answer_store, instance_idx):
```
kierenbeckett, line 82: What happens here if we are trying to skip on an answer that isn't in a repeating group?

data/en/test_skip_conditions_on_blocks_repeating_group.json
```diff
@@ -0,0 +1,151 @@
+{
+    "mime_type": "application/json/ons/eq",
+    "schema_version": "0.0.1",
+    "data_version": "0.0.2",
+    "survey_id": "0",
+    "title": "A Test Schema to test skip conditions in a repeating answer",
+    "description": "Online Household Study Schema (Alpha)",
```
kierenbeckett: Needs updating

data/en/test_skip_conditions_on_blocks_repeating_group.json
```diff
@@ -0,0 +1,151 @@
...
+    "theme": "default",
+    "legal_basis": "Voluntary",
+    "eq_id": "lms",
```
kierenbeckett: Needs updating

kierenbeckett (changes requested):

benhowes (changes requested):
Only minor bits really, however I was wondering if there is a way to do this without the special case of skipping over the first n group_instances. I couldn't see a way in which that could be done though.

app/questionnaire/path_finder.py
AndrewTorrance: I dont think so . If the skip condition points to an answer in a repeating group then the skip condition will need to be evaluated for each group instance, not for all instances within the block. If however it points to an answer not in a repeating group then it will skip or not skip the same for each group instance.

app/questionnaire/path_finder.py
AndrewTorrance: I dont think so  ,what it is trying to do is to not add the path for those blocks unless there is at least one block that is not skipped

app/questionnaire/path_finder.py
AndrewTorrance: Changing

app/questionnaire/path_finder.py
AndrewTorrance: group id will be zero and return the first one

tests/functional/spec/features/skipping/skipping_blocks_in_repeating_groups.js
AndrewTorrance: Its not part of the functionality that is being tested, so I thought it better to leave the last check to be the one I really want to pass.

app/questionnaire/path_finder.py
AndrewTorrance: group_instance is a synonym for a question repeated multiple times based on a value submitted earlier. So If I had 5 people in a household I can repeat asking them their names 5 times . Same group, same block but a different group_instance. Changing doc string

app/questionnaire/path_finder.py
AndrewTorrance: Not quite , the group_instance is badly named and is in fact an integer, and zero is valid. Changing variable names to make it clearer.

app/questionnaire/path_finder.py
AndrewTorrance: see above

app/questionnaire/path_finder.py
AndrewTorrance: Changed function name to make it clearer

app/questionnaire/path_finder.py
benhowes: Ah - of course, you're quite right

tests/functional/spec/features/skipping/skipping_blocks_in_repeating_groups.js
benhowes: I felt it is, because you want to ensure that the other sections are skipped?

app/questionnaire/path_finder.py
benhowes: I think this point is unrelated to the above one

kierenbeckett: Makes sense.  I guess I'm wondering about `group['blocks'][0]['id']`, might block 0 have been skipped?

app/questionnaire/path_finder.py
```diff
@@ -59,19 +59,32 @@ def build_path(self):
...
+                    path, block_index = self._build_path_within_group(blocks, block_index, this_location, path)
 
         return path
 
+    @staticmethod
+    def _get_first_group_instance_index_in_block(group, blocks):
+        """returns the first group instance of a block that does not need to be skipped
+        or else None if all to be skipped"""
+        try:
+            first_block = next(block for block in blocks if block['group_id'] == group['id'])
```
nealedj: Could this be reduced to:

```
first_block = next((block for block in blocks if block['group_id'] == group['id']), None)
if first_block:
    return first_block['group_instance']
```
?

tests/app/questionnaire/test_path_finder.py
```diff
@@ -653,6 +653,124 @@ def test_repeating_groups_conditional_location_path(self):
...
+
+            # Set up some first names
+            answer_store.add(Answer(answer_id='first-name', value='aaa', group_instance=0))
+            answer_store.add(Answer(answer_id='first-name', value='bbb', group_instance=1))
+            answer_store.add(Answer(answer_id='first-name', value='ccc', group_instance=2))
+
+            # Set up the independent answer
+            answer_store.add(Answer(answer_id='an-independent-answer', value=independent_answer, group_instance=0))
+
+            sut = PathFinder(schema, answer_store=answer_store, metadata={}, completed_blocks=[])
```
nealedj: Not sure what `sut` means?

tests/app/questionnaire/test_path_finder.py
```diff
@@ -653,6 +653,124 @@ def test_repeating_groups_conditional_location_path(self):
...
 
+    def test_repeating_group_block_skip_correct_if_independent_answer_is_not_repeating_but_causes_a_skip(self):
+        """ tests blocks are correctly added to path in the case where the blocks in a repeating group have
+        skip conditions and those skip conditions are dependant on a non repeating answer which causes a skip
+        condition to return true.
+        """
+
+        independent_answer = '3'
+
+        with patch('app.questionnaire.rules._answer_is_in_repeating_group', return_value=False):
```
nealedj: Probably worth moving this patch down to only wrap the call to `get_full_routing_path`. Best not to wrap the setup code (not that it'll make any difference in this case).

tests/app/questionnaire/test_path_finder.py
```diff
@@ -653,6 +653,124 @@ def test_repeating_groups_conditional_location_path(self):
...
+
+            expected_path = [
```
nealedj: Nitpick - could this be moved closer to end where it's used in the assertion? Defining it this high mixes it with code that's otherwise engaged in setting up the test.

tests/app/questionnaire/test_path_finder.py
```diff
@@ -653,6 +653,124 @@ def test_repeating_groups_conditional_location_path(self):
...
+        skip conditions and those skip conditions are dependant on a repeating answer
+        tests various combinations of independent answers being set , since we saw different
+        behaviours for group_instance 0 as it was the default"""
+
+        param_list = [(0, 2), (1, 3)]
+
+        with patch('app.questionnaire.rules._answer_is_in_repeating_group', return_value=True):
+            for first_group_instance, second_group_instance in param_list:
+
+                with self.subTest():
```
nealedj: Could you pass the context to sub test so it's included in any test failure message?

```
with self.subTest(first_group_instance=first_group_instance, second_group_instance=second_group_instance):
```

app/questionnaire/path_finder.py
```diff
@@ -59,19 +59,32 @@ def build_path(self):
...
-            if blocks:
-                path, block_index = self._build_path_within_group(blocks, block_index, this_location, path)
+                if blocks:
+                    path, block_index = self._build_path_within_group(blocks, block_index, this_location, path)
 
         return path
 
+    @staticmethod
+    def _get_first_group_instance_index_in_block(group, blocks):
+        """returns the first group instance of a block that does not need to be skipped
```
nealedj: The method code doesn't match the docstring. This method simply returns the group_instance index of the first block belonging to the group. Perhaps this might be better as an inline comment on the calling code to better explain how `first_group_instance_index` is being used?

app/questionnaire/path_finder.py
```diff
@@ -59,19 +59,32 @@ def build_path(self):
                 group_blocks = self._build_blocks_for_group(group, instance_idx)
                 blocks += group_blocks
 
-            if group['id'] in first_groups:
-                this_location = Location(group['id'], 0, group['blocks'][0]['id'])
+            first_group_instance_index = self._get_first_group_instance_index_in_block(group, blocks)
+            if first_group_instance_index is not None:
+                if group['id'] in first_groups:
+                    this_location = Location(group['id'], first_group_instance_index, group['blocks'][0]['id'])
```
nealedj: I know it was already here but could you replace `group['blocks'][0]['id']` with `first_block_in_group`?

```
        for group in self.schema.groups:
            first_block_in_group = self.schema.get_first_block_id_for_group(group['id'])
            if not this_location:
                this_location = Location(group['id'], 0, first_block_in_group)

            ....

            first_group_instance_index = self._get_first_group_instance_index_in_block(group, blocks)
            if first_group_instance_index is not None:
                if group['id'] in first_groups:
                    this_location = Location(group['id'], first_group_instance_index, first_block_in_group)
```

app/questionnaire/path_finder.py
```diff
@@ -59,19 +59,32 @@ def build_path(self):
                 group_blocks = self._build_blocks_for_group(group, instance_idx)
```
nealedj: Might be worth grabbing the `group_instance` of the first block here. Then you wouldn't need to filter the `blocks` list on `group_id` outside the loop.

```
            first_group_instance_index = None
            for instance_idx in range(0, no_of_repeats):
                group_blocks = self._build_blocks_for_group(group, instance_idx)
                blocks += group_blocks

                if group_blocks and first_group_instance_index is None:
                    # get the group instance of the first instance of a block in this group that has not been skipped
                    first_group_instance_index = group_blocks[0]['group_instance']
```

app/questionnaire/path_finder.py
```diff
@@ -59,19 +59,32 @@ def build_path(self):
...
                 blocks += group_blocks
 
-            if group['id'] in first_groups:
-                this_location = Location(group['id'], 0, group['blocks'][0]['id'])
+            first_group_instance_index = self._get_first_group_instance_index_in_block(group, blocks)
+            if first_group_instance_index is not None:
```
nealedj: Might be a case for a variable here to provide some extra context:

```
all_blocks_skipped = first_group_instance_index is None
if not all_blocks_skipped:
    if group['id'] in first_groups:
        ...
```

nealedj (changes requested):
Nicely done - functionally looks good to me. My comments mainly around readability because this touches an area of the codebase which can be fairly impenetrable.

tests/functional/spec/features/skipping/skipping_blocks_in_repeating_groups.js
AndrewTorrance: You know what ben , changed my mind - I think your right. I had the two tests forming a pair , one checking we could skip two iterations and the other testing we can use the 0th group_instance. In that case I should not show a question for another group instance. So adding a check for summary makes this test complete .

app/questionnaire/path_finder.py
AndrewTorrance: In the when group evaluation it checks if the independent answer is in a repeating group, if it is it uses the group_instance of zero . Added tests to show that

app/questionnaire/path_finder.py
AndrewTorrance: I think you are questioning the try catch ? StopIteration is thrown if no blocks match . KeyError thrown if no group id. Both cases should result in no blocks added

tests/app/questionnaire/test_path_finder.py
AndrewTorrance: system under test . Its short hand for the item being tested . https://en.wikipedia.org/wiki/System_under_test

tests/app/questionnaire/test_path_finder.py
AndrewTorrance: will do

tests/app/questionnaire/test_path_finder.py
AndrewTorrance: done

tests/app/questionnaire/test_path_finder.py
AndrewTorrance: Thought about that but  couldn't think of a nice way of setting an array of locations  for the expected_path without making the param_list hard to read. Came down on the side of duplicating code in a test for clarity . Will change if you prefer

app/questionnaire/path_finder.py
AndrewTorrance: Good catch - thanks

app/questionnaire/path_finder.py
AndrewTorrance: Done it , but needed an additional change , to make sure first_block_in_group set on each iteration of groups

tests/app/questionnaire/test_path_finder.py
insacuri: There was a bunch that in secure-messaging and I didn't really understand it there either.  It's easier to actually call the thing what it's called instead of using the generic term for it

app/questionnaire/path_finder.py
AndrewTorrance: Agree , thats clearer.

tests/app/questionnaire/test_path_finder.py
nealedj: Not sure we're talking about the same thing? The subTest() method accepts arguments purely for the purpose of displaying what iteration is currently being tested in case of a failure.

app/questionnaire/path_finder.py
nealedj: In this case I'm not sure it's a good fit for a `try/catch` since the alternative is just to pass `None` to the `next` function.

app/questionnaire/path_finder.py
AndrewTorrance: Doesnt work if first group_instance is skipped . Its not obvious though.

tests/app/questionnaire/test_path_finder.py
AndrewTorrance: No I was talking about almost duplicate tests , I see what you mean , will add message

app/questionnaire/path_finder.py
```diff
@@ -59,19 +59,34 @@ def build_path(self):
...
+    @staticmethod
+    def _get_first_group_instance_index_in_block(group, blocks):
+        # returns the first block of a list of blocks that matches the group id
+
+        try:
+            first_block = next(block for block in blocks if block['group_id'] == group['id'])
+            return first_block['group_instance']
+        except (StopIteration, KeyError):  # Either no group instances (all skipped) or group has no 'id'
+            pass
+        return None
```
kierenbeckett: Pedantic: I wouldn't bother with this line

app/questionnaire/path_finder.py
```diff
@@ -59,19 +59,34 @@ def build_path(self):
...
-            if group['id'] in first_groups:
-                this_location = Location(group['id'], 0, group['blocks'][0]['id'])
+            first_group_instance_index = self._get_first_group_instance_index_in_block(group, blocks)
+            all_blocks_skipped = first_group_instance_index is None
 
-            if blocks:
-                path, block_index = self._build_path_within_group(blocks, block_index, this_location, path)
+            if not all_blocks_skipped:
+                if group['id'] in first_groups:
+                    this_location = Location(group['id'], first_group_instance_index, first_block_in_group)
```
kierenbeckett: Is it possible that `first_block_in_group` might be skipped, does it matter?

kierenbeckett (changes requested):

app/questionnaire/path_finder.py
AndrewTorrance: It was an issue , but I changed the setting of it so that it is reset on every iteration of group. Blocks are mandatory within a group so it will be valid on each iteration.

app/questionnaire/path_finder.py
nealedj: But `group_blocks` doesn't contain skipped instances does it? So it should be safe to get the first one because that's the first one that wasn't skipped.

nealedj: I just changed it and the tests still pass.

app/questionnaire/path_finder.py
AndrewTorrance: I know , but I thought best to make it explicit

kierenbeckett (approved):

app/questionnaire/path_finder.py
AndrewTorrance: It does not matter, but we did not have a test to proove that . Modified an existing test to cover that scenario.

tests/app/questionnaire/test_path_finder.py
benhowes: The rest of that test file just does `path_finder = PathFinder(...`, agree with @insacuri

app/questionnaire/path_finder.py
AndrewTorrance: Not for me .  I get an error because group blocks is not subscriptable.

nealedj: Oh that's just because `_build_blocks_for_group` is a generator.

That line would need to become

```
group_blocks = list(self._build_blocks_for_group(group, instance_idx))
```

Not going to labour the point now but I think it's clearer to do it here rather than extra it from the full block set later.

benhowes (approved):

benhowes:
Checked out, tested the survey schema suggested - all works fine, answers in /dump/answers appear correct

nealedj (approved):

PR #1692, opened 2018-07-24
Make section summary only available if section complete

MebinAbraham:
Previously, section summary was accessible via the URL if a section was incomplete but was once complete.
This change ensures that if you try to access the section summary of a section, it redirects to the first incomplete block in that section if any.

Also started on the new `Router` class to help refactor out some of the logic currently in our views. Others logic still exists that need to be refactored to get the most of this class. https://trello.com/c/R3yLORjb/2244-move-routing-logic-to-a-new-component

***Expected behaviour:***
- If requesting section summary via URL then go to the first incomplete block within that section.
- If requesting final summary via URL then go to the first incomplete block within the survey.

### How to review 
- Ensure the section summary and final summary behave as expected.
- Ensure questionnaire behaves as normal.

### Checklist


app/views/questionnaire.py
```diff
@@ -301,12 +312,6 @@ def get_view_submission(eq_id, form_type):  # pylint: disable=unused-argument
     return redirect(url_for('post_submission.get_thank_you', eq_id=eq_id, form_type=form_type))
 
 
-def _redirect_to_latest_location(collection_id, eq_id, form_type):
-    latest_location = get_completeness(current_user).get_latest_location()
```
kierenbeckett, line 305: Can we remove `get_latest_location` from `completeness`

app/questionnaire/router.py
```diff
@@ -0,0 +1,75 @@
...
+
+        expected_location = self._completeness.get_first_incomplete_location_in_section(self._section)
+
+        if self._block['type'] == 'SectionSummary' and self._current_location != expected_location:
+            self._first_incomplete_block_in_section = expected_location
+
+        return self._first_incomplete_block_in_section
+
+    def _is_skipping_to_survey_end(self):
+        if self._first_incomplete_block_in_survey:
```
kierenbeckett: Discussed IRL, I think we can simplify things here by removing the instance variables and just getting locally as required.

kierenbeckett (changes requested):
I like the interface to `Router`, as discussed IRL we can probably simplify the code inside it a little bit.

app/views/questionnaire.py
MebinAbraham: Yes we can, forgot to do so. Good catch 👍

kierenbeckett (approved):

app/views/questionnaire.py
```diff
@@ -125,15 +126,15 @@ def save_questionnaire_store_wrapper(*args, **kwargs):
 @full_routing_path_required
 def get_block(routing_path, eq_id, form_type, collection_id, group_id, group_instance, block_id):  # pylint: disable=unused-argument,too-many-locals
     current_location = Location(group_id, group_instance, block_id)
+    completeness = get_completeness(current_user)
 
-    if not _is_valid_location(routing_path, current_location):
-        return _redirect_to_latest_location(collection_id, eq_id, form_type)
-
-    block = _get_block_json(current_location)
+    route = Router(g.schema, routing_path, completeness, current_location)
```
jonnyshaw89: I can't see a good reason for calling this `route` instead of `router`?

nealedj (changes requested):
Since `PathFinder.get_latest_location` is only used in `post_block` now could we remove that as well?

app/views/questionnaire.py
jonnyshaw89: the comment applies to other usages too

app/views/session.py
```diff
@@ -70,9 +72,13 @@ def login():
...
 
-    current_location = get_completeness(current_user).get_latest_location()
+    routing_path = path_finder.get_full_routing_path()
+    completeness = get_completeness(current_user)
+    route = Router(g.schema, routing_path, completeness)
 
-    return redirect(current_location.url(claims))
+    next_location = route.get_next_location()
+
+    return redirect(next_location.url(claims))
```
jonnyshaw89: This isn't related to changes in this PR but why do we pass the claims here?

jonnyshaw89 (changes requested):

app/views/questionnaire.py
nealedj: Nitpick - an instance of `Router` is probably still best assigned to variable called `router` rather than `route`. What it returns isn't a route.

app/questionnaire/router.py
```diff
@@ -0,0 +1,52 @@
...
+
+        if self._completeness.completed_blocks:
+            next_location = self._completeness.get_first_incomplete_location_in_survey()
+            if not next_location:
+                next_location = self._routing_path[-1]
+
+        return next_location
+
+    def _get_first_incomplete_block_in_section(self):
+        section = self._schema.get_section_by_block_id(self._current_location.block_id)
```
nealedj: `self._current_location` may be `None` and that'll throw an error here.

I know neither of the `_get_first_incomplete_block_in_x` methods get called if there's no `current_location` because `_is_valid_location` returns `False`. But I'd suggest being defensive here.

MebinAbraham:
@nealedj I assume you are talking about `pathfinder.get_next_location`?, technically can't be removed with the current state of `Router` since Routers `get_next_location` isn't used to move linearly yet, I am not quite sure about the scope of this card. I could remove `get_next_location` from `PathFinder` but I would need to handle that in Router with also the likes of `_get_valid_section_summary` etc. Not sure where to stop.

app/views/questionnaire.py
MebinAbraham: No specific reason just thought it didn't not make sense to call it `route` but ya can change to `router`. 👍

app/questionnaire/router.py
```diff
@@ -0,0 +1,52 @@
...
+        if not self._is_valid_location():
+            next_location = self._get_expected_location()
+            return next_location
+
+        return self._get_first_incomplete_block_in_section() or self._get_first_incomplete_block_in_survey() or self._get_expected_location()
+
+    def _is_valid_location(self):
+        return self._current_location in self._routing_path
+
+    def _get_expected_location(self):
```
nealedj: Took me a while to work out what this method was responsible for. It's designed to work out what location to take a user to if there isn't a valid `current_location` available. I think the word "expected" is quite confusing here.

Perhaps consider calling it `_get_default_location` and then add a decent docstring to explain that it's essentially used for a) the location after login and b) as a fallback if somebody hits a URL they can't reach.

app/views/session.py
MebinAbraham: Was making use of `Router` where it can be used at its current state?
Regarding `claims` which is existing, that is just used to get values from it such as `eq_id` etc for the path?.

app/questionnaire/router.py
```diff
@@ -0,0 +1,52 @@
...
+                next_location = self._routing_path[-1]
+
+        return next_location
+
+    def _get_first_incomplete_block_in_section(self):
+        section = self._schema.get_section_by_block_id(self._current_location.block_id)
+        expected_location = self._completeness.get_first_incomplete_location_in_section(section)
+        block = self._schema.get_block(self._current_location.block_id)
+
+        if block['type'] == 'SectionSummary' and self._current_location != expected_location:
```
nealedj: So this is checking if the user has gone directly to a section summary but there are still incomplete blocks in that section?

Either a comment or an extra variable to add this context would be helpful for future devs.

Maybe something like this?
```
can_access_section_summary = self._current_location == expected_location
if block['type'] == 'SectionSummary' and not can_access_section_summary:
    return expected_location
```

app/questionnaire/router.py
```diff
@@ -0,0 +1,52 @@
...
+        next_location = self._routing_path[0]
+
+        if self._completeness.completed_blocks:
+            next_location = self._completeness.get_first_incomplete_location_in_survey()
+            if not next_location:
+                next_location = self._routing_path[-1]
+
+        return next_location
+
+    def _get_first_incomplete_block_in_section(self):
```
nealedj: This method name doesn't quite match the functionality because the method only returns the first incomplete block if they aren't allowed access the block they're trying to get to.

I don't know if it's much better but perhaps `_get_section_redirect_location`.

Again - needs docstrings. This stuff isn't obvious :)

app/views/session.py
jonnyshaw89: Yeh maybe we should refactor this in another PR, we pass in claims but just use these 2 values. we are currently passing `collection_id` into all of the views and only using it for redirects to other views and never actually using it.

app/questionnaire/router.py
```diff
@@ -0,0 +1,52 @@
...
+        self._current_location = current_location
+
+    def can_access_location(self):
+        if self._is_valid_location() and not self._get_first_incomplete_block_in_section() and not self._get_first_incomplete_block_in_survey():
+            return True
+
+        return False
+
+    def get_next_location(self):
+        if not self._is_valid_location():
```
nealedj: Consider inverting this to:

```
if self._is_valid_location():
    # get next location based on current location
else:
    # get next location by default
```

app/questionnaire/router.py
```diff
@@ -0,0 +1,52 @@
...
+            next_location = self._get_expected_location()
+            return next_location
+
+        return self._get_first_incomplete_block_in_section() or self._get_first_incomplete_block_in_survey() or self._get_expected_location()
+
+    def _is_valid_location(self):
```
nealedj: Since we have a couple of "location" concepts floating around (location that the user is trying to get to, location that we're _actually_ sending them to because we aren't allowing that location, location that we're redirecting them to because they just want to know where to go next...) I think it's worth changing this method name to something like `_is_current_location_valid`.

tests/app/questionnaire/test_router.py
```diff
@@ -0,0 +1,248 @@
+from unittest.mock import MagicMock
+
+from tests.app.app_context_test_case import AppContextTestCase
+from app.questionnaire.router import Router
+from app.questionnaire.location import Location
+from app.questionnaire.completeness import Completeness
+from app.utilities.schema import load_schema_from_params
+
+
+class TestRouter(AppContextTestCase):
```
nealedj: There's no need to inherit from `AppContextTestCase` here. It'll set up the app and DynamoDB for each test but since these tests don't require that you could simply inherit from `TestCase`

app/views/session.py
MebinAbraham: Tbh, not too sure what was within the scope, I just thought I'll refactor the things that Router can currently cope with and do the rest along with https://trello.com/c/R3yLORjb/2244-move-routing-logic-to-a-new-component

Regarding claims Im not too sure, its used in multiple places and we just pass it the metadata/claims and it gets the info it needs from it. Something for another card to change that?

nealedj (changes requested):
This looks really good. I think this nicely encapsulates some important logic that was previously held over some slightly disparate functions.

Most of my comments are to try and make the code more understandable. Things like suggested method names and requests for docstrings.

With the Router component could you add a docstring to every method? I know it sounds like overkill but I think this component is likely to grow so it'd be useful to make it clear what each bit is doing.

nealedj:
@MebinAbraham 

> @nealedj I assume you are talking about pathfinder.get_next_location?, technically can't be removed with the current state of Router since Routers get_next_location isn't used to move linearly yet, I am not quite sure about the scope of this card. I could remove get_next_location from PathFinder but I would need to handle that in Router with also the likes of _get_valid_section_summary etc. Not sure where to stop.

Ah I see - yes this sounds like a good place to stop in that case. Don't worry about bringing in the linear logic. 👍

app/views/session.py
nealedj: Oh yeah good catch. That doesn't make sense. Could one of you raise a card to explicitly pass the bits we need to construct the URL?

insacuri (approved):

app/questionnaire/router.py
MebinAbraham: I changed the names so many times as there are not any great ones, ya ill go with your suggestion and add doc strings to help. When router comes to handle `_get_first_incomplete_location_in_section` this will become confusing. ATM this is only used if its a section summary or an end block, so not quite the same meaning.

app/questionnaire/router.py
MebinAbraham: Its used in the event that current location is invalid or current location is valid but not skipping.

Ill add doc strings and rename method 👍

MebinAbraham:
@jonnyshaw89 @nealedj Ready for re-review.

nealedj (approved):

jonnyshaw89 (approved):

PR #1694, opened 2018-07-25
Fix submission dumping with repeating groups

benhowes:

1. A minor fix to dumping the submission, because an unset global was relied on. Adds a test for this case.
2. As part of 1, reduces (but doesn't entirely remove) use of `g.schema` throughout. This means passing the schema as an argument a lot more, particularly in the templating where this wasn't done before. This appears to be because until repeating groups were added, there was no need for knowledge of the schema in these places. 
3. Due to the reduced use of `g.schema`, `_answer_is_in_repeating_group` has been removed from `rules.py`, which has required modification of a boat load of tests since this was mocked in many places. In general, I've tried to patch a higher level call - i.e. mock out `evaluate_goto` instead of the low level one. This is to try to isolate units more.
4. A couple of tests have been removed, either were they were testing another unit by proxy, and in one case there was an identical test under two names.
5. A couple of tests have been added - these are unrelated to the fix, however this PR was failing on code coverage, due to removing tested code (which increased the percentage of lines which were not tested).

This PR has ended up being bigger than intended. Some of the unit tests in templating have changed more than they strictly had to, however since I needed to change all the call signatures anyway, I wanted to isolate the units more whist I was understanding what each test did.

### How to review 
1. Tests should all pass!
2. Running `census_household.json` or other surveys with repeating groups should allow dumping of the submission on `/dump/submission`. To ensure the fix works, test one is failing on master first (should return a 500 error)

### DIFF
--- a/app/templating/summary/question.py
+++ b/app/templating/summary/question.py
@@ -6,14 +6,14 @@ from app.templating.utils import get_question_title
 
 class Question:
 
-    def __init__(self, question_schema, answer_store, metadata):
+    def __init__(self, question_schema, answer_store, metadata, schema):
         self.id = question_schema['id']
         self.type = question_schema['type']
 
         answer_schemas = iter(question_schema['answers'])
 
         # Using group instance as 0 for now as summary rendering context only has knowledge of current locations group instance (i.e. 0)
-        self.title = get_question_title(question_schema, answer_store, metadata, group_instance=0) or question_schema['answers'][0]['label']
+        self.title = get_question_title(question_schema, answer_store, schema, metadata, group_instance=0) or question_schema['answers'][0]['label']
         self.number = question_schema.get('number', None)
         self.answers = self._build_answers(answer_store, question_schema, answer_schemas)
 
--- a/app/views/questionnaire.py
+++ b/app/views/questionnaire.py
@@ -1,3 +1,4 @@
+from functools import wraps
 import re
 from collections import defaultdict
 from datetime import datetime, timedelta
@@ -18,7 +19,10 @@ from app.data_model.app_models import SubmittedResponse
 from app.globals import get_answer_store, get_completed_blocks, get_metadata, get_questionnaire_store
 from app.helpers.form_helper import post_form_for_location
 from app.helpers.path_finder_helper import path_finder, full_routing_path_required
-from app.helpers import template_helper
+from app.helpers.schema_helpers import with_schema
+from app.helpers.session_helpers import with_answer_store, with_metadata
+from app.helpers.template_helper import (with_session_timeout, with_metadata_context, with_analytics,
+                                         with_questionnaire_url_prefix, with_legal_basis, render_template)
 
 from app.questionnaire.location import Location
 from app.questionnaire.navigation import Navigation
@@ -38,7 +42,7 @@ from app.templating.summary_context import build_summary_rendering_context
 from app.templating.template_renderer import renderer, TemplateRenderer
 from app.templating.view_context import build_view_context
 
-from app.utilities.schema import load_schema_from_session_data, load_schema_from_metadata
+from app.utilities.schema import load_schema_from_session_data
 from app.views.errors import MultipleSurveyError
 from app.authentication.no_token_exception import NoTokenException
 
@@ -86,6 +90,7 @@ def before_post_submission_request():
         raise NoTokenException(401)
 
     session_data = session.session_data
+    g.schema = load_schema_from_session_data(session_data)
 
     logger.bind(tx_id=session_data.tx_id)
 
@@ -109,6 +114,7 @@ def add_cache_control(response):
 
 
 def save_questionnaire_store(func):
+    @wraps(func)
     def save_questionnaire_store_wrapper(*args, **kwargs):
         response = func(*args, **kwargs)
         if not current_user.is_anonymous:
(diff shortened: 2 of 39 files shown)


tests/integration/mci/test_backwards_navigation_after_submission.py
```diff
@@ -1,11 +1,13 @@
 from tests.integration.mci import mci_test_urls
-from .test_happy_path import TestHappyPath
+from tests.integration.integration_test_case import IntegrationTestCase
 
+from .test_happy_path import HappyPathHelperMixin
 
-class TestBackwardsNavigationAfterSubmission(TestHappyPath):
+
+class TestBackwardsNavigationAfterSubmission(IntegrationTestCase, HappyPathHelperMixin):
```
benhowes, line 7: I just wanted to signpost here that the new mixin class is there because previously, when this inherited from `TestHappyPath`, it meant that all the test cases from `TestHappyPath` were also executed here.

jonnyshaw89: Is this similar to this? https://github.com/ONSdigital/eq-survey-runner/blob/master/tests/app/app_context_test_case.py#L39

benhowes: Yes, except that one must be executed a lot more than the one I've fixed

benhowes: If it's okay with you, I'll put through a separate PR which addresses that. Of course it's being collected because pytest is looking for everything starting with `test_`, so I'll just rename that

app/templating/summary/question.py
```diff
@@ -6,14 +6,14 @@
 
 class Question:
 
-    def __init__(self, question_schema, answer_store, metadata):
+    def __init__(self, question_schema, answer_store, metadata, schema):
```
insacuri, line 9: What's the difference between question_schema and schema?

app/views/questionnaire.py
```diff
@@ -258,7 +263,7 @@ def get_thank_you(eq_id, form_type):  # pylint: disable=unused-argument
 def get_view_submission(eq_id, form_type):  # pylint: disable=unused-argument
 
     session_data = get_session_store().session_data
-    g.schema = load_schema_from_session_data(session_data)
+    g.schema = schema = load_schema_from_session_data(session_data)
```
insacuri: Pretty cool, had no idea that you could do chain assignments in python!

insacuri (approved):
LGTM 
EDIT: Though 2 others should look at this cause there's a chance I missed something cause I'm still new to the codebase ^^

app/templating/summary/question.py
benhowes: `question_schema` is a 'fragment' (n.b. contentious term) of the whole `schema`, but there are places that the whole schema is needed to evaluate if something's in a repeating group

insacuri: thats what I thought, just wanted to be sure I got it

benhowes:
I missed a few instances of `_answer_is_in_repeating_group` which appeared after rebasing, working through those now

app/views/questionnaire.py
```diff
@@ -125,6 +125,8 @@ def save_questionnaire_store_wrapper(*args, **kwargs):
 @full_routing_path_required
 def get_block(routing_path, eq_id, form_type, collection_id, group_id, group_instance, block_id):  # pylint: disable=unused-argument,too-many-locals
     current_location = Location(group_id, group_instance, block_id)
+    metadata = get_metadata(current_user)
+    schema = load_schema_from_metadata(metadata)
```
jonnyshaw89: worth considering this as a wrapper? like we do with `@full_routing_path_required`?
given we repeat these lines in most of the views.

jonnyshaw89 (changes requested):
A nice bit of boy scouting done here 👍

app/views/questionnaire.py
benhowes: It is for sure, I've suggested that we use pytest fixtures for most of these repeated helpers here: https://trello.com/c/zuIZkU2N/674-make-more-use-of-pytest-fixtures-for-simplification-and-speeding-up-of-tests

benhowes: But I've not included it here because this was already getting pretty big!

jonnyshaw89: I wasn't thinking specifically about pytest. i was looking more at the repeated lines in every view

benhowes: Shall I put that on the backlog?

jonnyshaw89: Id say it was a change related to the code changed here?

benhowes: Yeah, okay fair. I'll do that.

app/views/questionnaire.py
nealedj: Really useful. Don't make the mistake of assigning the same reference to multiple variables and thinking they're different objects.

Pretty obvious:
```
>>> a_string = b_string = "foo"
>>> a_string = "bar"
>>> a_string
'bar'
>>> b_string
'foo'
```

Sometimes less expected:
```
>>> a_list = b_list = []
>>> a_list.append('foo')
>>> a_list
['foo']
>>> b_list
['foo']
```

jonnyshaw89:
@benhowes can you try and hold off with squashing until you have 2 ticks? it just makes re-reviews a lot easier as we can just re-review the changes and not have to look over all the files again.

benhowes:
I hadn't realised that implication of squashing - sorry!

app/views/questionnaire.py
```diff
@@ -123,155 +129,165 @@ def save_questionnaire_store_wrapper(*args, **kwargs):
 
 @questionnaire_blueprint.route('<group_id>/<int:group_instance>/<block_id>', methods=['GET'])
 @login_required
+@with_answer_store
+@with_metadata
+@with_schema
```
jonnyshaw89, line 134: Is the order important here?
Just a thought, can we make the wrappers pass in the value as a kwarg?

benhowes: Order is very important. can easily pass as kwargs, but I was being consistent with the existing `@full_routing_path_required` decorator

jonnyshaw89: Yeh, just thinking the more we add the more brittle it could become to order, maybe if we make them all use kwarg it would be more resilient?

benhowes: Actually, thinking about it, there would be some interesting errors if they were kwargs and one was missing for some reason, because you can't easily provide valid defaults for those arguments, meaning we'd have to introduce some `if schema is None: ...` for each of the decorators. Unless I'm missing something of course.

jonnyshaw89: would we also get funky errors if they were in the wrong order? so a schema was actually metadata?

benhowes: Yep, we would. it's weird errors either way, both ways have the same downsides I think. What do you reckon @jonnyshaw89

jonnyshaw89 (approved):

kierenbeckett (approved):

benhowes:
@kierenbeckett @jonnyshaw89 since you reviewed, I added some tests and removed some unreachable code in this commit: https://github.com/ONSdigital/eq-survey-runner/pull/1694/commits/80920a80caf634bab0c81e98f366b22a02d2b7b9, can you have a look over to ensure you're happy with the changes?

jonnyshaw89:
👍 Happy

kierenbeckett:
👍

PR #1698, opened 2018-07-26
Fix pipenv lock

insacuri:
### What is the context of this PR?
When trying to rebuild the pipenv lockfile, the following error would be generated 
`Could not find a version that matches astroid<2.0,>=1.6,>=2.0.1`
By updating (and unpinning) a few dependencies, the lockfile regenerates successfully now.

Also by updating the dependencies, pylint had a few more complaints about the code, so this PR fixes those also.

### How to review 
The site and all tests should still work.  No new features were added.

### Checklist

* [x] New static content marked up for translation
* [x] Newly defined schema content included in eq-translations repo

PR #1699, opened 2018-07-26
Unpin all packages

insacuri:
Branched off of https://github.com/ONSdigital/eq-survey-runner/pull/1698. As an experiment, I wanted to see what would happen if I unpinned all the dev package versions.  Turns out, very little changed in the lockfile and after adding/removing a few commas because of the updated flake8, all the tests passed as expected.

This needs to be merged in after the pylint branch as we don't want to change too many dependencies in one go.

### How to review 
Site and tests should work/pass as expected.  No new features added

### Checklist


nealedj (approved):
Looks like the only changes were due to flake8-commas implementing some extra rules - but they all look like good changes to me.

benhowes:
I've checked around and the dependencies are always installed from the lockfile i.e. with `pipenv install --deploy`. This effectively makes the lockfile the method of pinning. 

The main thing that this changes is that when freshly cloning the repo, one probably wants to install with the `--deploy` flag so as to not update the dependencies without realising, because by default a new lockfile will be generated if there are new releases of any of the packages.

benhowes (approved):

PR #1689, opened 2018-07-23
Add a repeating type of `until`

kierenbeckett:
### What is the context of this PR?
Add a repeating type of `until`.  See also ONSdigital/go-launch-a-survey#106 and ONSdigital/eq-schema-validator#67.

### How to review 
Run `test_routing_repeat_until.json` , fill in some names and choose yes/no for the more people question at various stages.  See that you get asked about the sex of all the people's names you entered.  Note that the summary page only displays the first person in the repeating group, fixing this isn't within the scope of this PR.

PR #1700, opened 2018-07-26
Add a repeat type of 'group'

kierenbeckett:

Adding a repeat type of `group` and the concept of group identifiers (which i've named group_instance_id).

* By default group instances get a group_instance_id based on their group name and a UUID
* Added a new repeating type `group` which repeats over a list of groups specified by their group_ids.  These group instances get the group_instance_id of the group instances they are repeating over
* Added a context objects `group_instances` and `group_instance_id` which allows us to get at the answers from the group instances being repeated over
* There is a new field on answers `group_instance_id`, which will show up in the payload

### How to review
Run `test_routing_repeat_until.json`, enter yes when asked if there are any more people such that you enter at least 3 names.  See that you get asked the sex of each person, look in `/dump/submission` and have a look at the `group_instance_id` of each answer to see that the name and sex answers match up.

See also ONSdigital/eq-schema-validator#71

### DIFF
--- a/app/data_model/answer_store.py
+++ b/app/data_model/answer_store.py
@@ -6,11 +6,12 @@ import simplejson as json
 
 
 class Answer:
-    def __init__(self, answer_id, value, group_instance=0, answer_instance=0):
+    def __init__(self, answer_id, value, group_instance_id=None, group_instance=0, answer_instance=0):
         if answer_id is None or value is None:
             raise ValueError("Both 'answer_id' and 'value' must be set for Answer")
 
         self.answer_id = answer_id
+        self.group_instance_id = group_instance_id
         self.group_instance = group_instance
         self.answer_instance = answer_instance
         self.value = value
@@ -18,13 +19,13 @@ class Answer:
     def matches(self, answer):
         """
         Check to see if two answers match.
-        Two answers are considered to match if they share the same block_id, answer, answer_instance, group_id and group_instance.
+        Two answers are considered to match if they share the same answer_id, answer_instance and group_instance_id.
 
         :param answer: An answer to compare
         :return: True if both answers match, otherwise False.
         """
         return self.answer_id == answer.answer_id and \
-            self.group_instance == answer.group_instance and \
+            self.group_instance_id == answer.group_instance_id and \
             self.answer_instance == answer.answer_instance
 
     def matches_dict(self, answer_dict):
@@ -38,6 +39,7 @@ class Answer:
         return self.matches(Answer(
             answer_dict['answer_id'],
             answer_dict['value'],
+            answer_dict['group_instance_id'],
             answer_dict['group_instance'],
             answer_dict['answer_instance'],
         ))
@@ -155,14 +157,14 @@ class AnswerStore:
             escaped.append(answer)
         return self.__class__(existing_answers=escaped)
 
-    def filter(self, answer_ids, group_instance=None, answer_instance=None,
-               limit=None):
+    def filter(self, answer_ids=None, group_instance=None, group_instance_id=None, answer_instance=None, limit=None):
         """
         Find all answers in the answer store for a given set of filter parameter matches.
         If no filter parameters are passed it returns a copy of the instance.
 
         :param answer_ids: The answer ids to filter results by
         :param answer_instance: The answer instance to filter results by
+        :param group_instance_id: The group instance ID to filter results by
         :param group_instance: The group instance to filter results by
         :param limit: True | False Limit the number of answers returned
         :return: Return a new AnswerStore object with filtered answers for chaining
@@ -172,6 +174,7 @@ class AnswerStore:
         filter_vars = {
             'answer_id': answer_ids,
             'answer_instance': answer_instance,
+            'group_instance_id': group_instance_id,
             'group_instance': group_instance,
         }
 
@@ -201,7 +204,7 @@ class AnswerStore:
         :param answer_instance: The answer instance to filter results to remove
         :param group_instance: The group instance to filter results to remove
         """
-        for answer in self.filter(answer_ids, group_instance, answer_instance):
+        for answer in self.filter(answer_ids, group_instance=group_instance, answer_instance=answer_instance):
             self.answers.remove(answer)
 
     def get_hash(self):
--- a/app/helpers/form_helper.py
+++ b/app/helpers/form_helper.py
@@ -1,4 +1,5 @@
 from collections import OrderedDict
+
 from structlog import get_logger
 from werkzeug.datastructures import MultiDict
 
@@ -7,6 +8,7 @@ from app.data_model.answer_store import natural_order
 from app.forms.household_composition_form import generate_household_composition_form, deserialise_composition_answers
 from app.forms.household_relationship_form import build_relationship_choices, deserialise_relationship_answers, generate_relationship_form
 from app.forms.questionnaire_form import generate_form
+from app.helpers.schema_helpers import get_group_instance_id
 
 logger = get_logger()
 
@@ -34,6 +36,8 @@ def get_form_for_location(schema, block_json, location, answer_store, metadata, 
 
         return generate_household_composition_form(schema, block_json, data, metadata, location.group_instance)
 
+    group_instance_id = get_group_instance_id(schema, answer_store, location)
+
     if location.block_id in ['relationships', 'household-relationships']:
         answer_ids = schema.get_answer_ids_for_block(location.block_id)
         answers = answer_store.filter(answer_ids, location.group_instance)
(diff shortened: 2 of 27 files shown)


tests/integration/views/test_questionnaire.py
```diff
@@ -66,12 +68,14 @@ def test_update_questionnaire_store_with_default_value(self):
             'answer': None
         }
 
-        with self._application.test_request_context():
+        with self._application.test_request_context(), patch('app.helpers.form_helper.uuid4') as uuid:
+            uuid.return_value = '1'
```
benhowes: `patch('app.helpers.form_helper.uuid4', return_value='1')` is cleaner way to patch this considering you're not asserting on uuid

tests/integration/views/test_dump.py
```diff
@@ -83,6 +83,10 @@ def test_dump_answers_authenticated_with_role_with_answers(self):
         # Enable full dictionary diffs on test failure
         self.maxDiff = None
 
+        # Cannot assert on randomly generated IDs
```
benhowes: There probably should be something that either asserts these keys are present? __Or__ you could use the `side_effect` of mock/patch to generate a predictable pattern of `group_instance_id`s, e.g. something along the lines of (untested)

```python
def uuid_mock():
    i = 0
    while True:
        yield group-{}".format(i)
        i += 1
# sometime later
patch('....', side_effect=uuid_mock()`
```

tests/app/helpers/test_form_helper.py
```diff
@@ -161,29 +162,29 @@ def test_get_form_for_household_relationship(self):
                 {
                     'group_id': 'who-lives-here-relationship',
```
benhowes: Should probably remove the `group_id` if you're tidying this up

app/data_model/answer_store.py
```diff
@@ -6,25 +6,27 @@
...
         if answer_id is None or value is None:
             raise ValueError("Both 'answer_id' and 'value' must be set for Answer")
 
         self.answer_id = answer_id
+        self.group_instance_id = group_instance_id
         self.group_instance = group_instance
         self.answer_instance = answer_instance
         self.value = value
 
     def matches(self, answer):
```
benhowes, line 19: This is more of a discussion point: Should this still use `group_instance` on the matching, because surely the `group_instance_id` is the only one that matters now?

app/views/questionnaire.py
```diff
@@ -666,13 +668,15 @@ def _get_block_json(current_location, schema, answer_store, metadata):
     return _evaluate_skip_conditions(block_json, current_location, schema, answer_store, metadata)
 
 
-def _get_schema_context(full_routing_path, group_instance, metadata, answer_store, schema):
+def _get_schema_context(full_routing_path, location, metadata, answer_store, schema):
     answer_ids_on_path = get_answer_ids_on_routing_path(schema, full_routing_path)
+    group_instance_id = get_group_instance_id(g.schema, answer_store, location) if location else None
```
benhowes: No `g.schema` please! Schema is passed as an argument in any case

benhowes (changes requested):
Haven't run it locally yet, but going to do that when the launcher build has been deployed.

tests/integration/views/test_dump.py
kierenbeckett: I wasn't convinced I should be using mocks in an integration test, I guess it's still better than not checking those fields.

app/views/questionnaire.py
kierenbeckett: Sorry, I had to merge that PR into this branch, must have missed this one

app/data_model/answer_store.py
benhowes: @nealedj ^^^

tests/integration/views/test_dump.py
benhowes: My take on that is generally no, but for randomness and datetime related stuff it's often the lesser evil compared to not asserting

app/data_model/answer_store.py
kierenbeckett: This PR already updates form helper to fetch answers by `(answer_id,group_instance_id)` so I believe it should be safe to do the same when saving.  I'll have a more thorough read through the code to be sure.

MebinAbraham (changes requested):
`KeyError` on launching the schema.
```
File "/Users/abrahm/Documents/projects/eq-survey-runner/app/questionnaire/rules.py", line 121, in evaluate_repeat
    for group_id in repeat_rule['group_ids']:
KeyError: 'group_ids'
```

kierenbeckett:
Oops, merging went a bit wrong

MebinAbraham (commented):
Still getting errors when launching the schema.

kierenbeckett:
Try now

MebinAbraham (commented):
All added records are meant to show on the summary now?
I added 3 people but only shows record for 2.

Also:
screen shot 2018-07-31 at 12 41 33

MebinAbraham (commented):
Maybe a question regarding the `until` implementation and summary, do we want to show the answer to the repeating question for each member? `Does anyone else live here?`. Should it not inherit the same behaviour as `Confirmation`?

kierenbeckett:
Sorting out the summary is a different card.  I did try making the 'any more people' question a confirmation, it works, but it isn't valid in the schema yet.

benhowes:
@kierenbeckett Needs rebasing now!

kierenbeckett:
@MebinAbraham @benhowes All rebased and passing if you could re-review

MebinAbraham:
@kierenbeckett This existing?
Not finding the group instance?
screen shot 2018-07-31 at 12 41 33

app/helpers/form_helper.py
```diff
@@ -130,10 +135,53 @@ def get_mapped_answers(schema, answer_store, block_id, group_instance):
...
+                                      group_instance_id=group_instance_id):
         answer_id = answer['answer_id']
         answer_id += '_' + str(answer['answer_instance']) if answer['answer_instance'] > 0 else ''
 
         result[answer_id] = answer['value']
 
     return OrderedDict(sorted(result.items(), key=lambda t: natural_order(t[0])))
+
+
+def get_group_instance_id(schema, answer_store, location):
```
benhowes: I was just wondering - is there an argument for making this a method on the schema - I suspect this is going to end up getting used in a lot of places. Sorry I didn't think of this before!

app/questionnaire/rules.py
```diff
@@ -116,11 +116,18 @@ def evaluate_repeat(repeat_rule, answer_store, answer_ids_on_path):
...
+            answer_ids_in_group = schema.get_answer_ids_for_group(group_id)
+            group_instances |= {a['group_instance_id'] for a in answer_store.filter(answer_ids=answer_ids_in_group)}
+
+        no_of_repeats = len(group_instances)
     else:
         repeat_functions = {
             'answer_value': _get_answer_value,
-            'answer_count': lambda f: _get_answer_count(f, repeat_rule),
-            'answer_count_minus_one': lambda f: min(_get_answer_count_minus_one(f), MAX_REPEATS - 1),
+            'answer_count': len,
```
benhowes, line 129: Looking at this again, the `_get_answer_count_minus_one` is capped by `MAX_REPEATS`, but this isn't

benhowes (commented):
Like Mebin, I'm not sure if the summary should be at least showing the name correctly for the first user.

app/helpers/form_helper.py
kierenbeckett: Not sure putting it in `QuestionnaireSchema` is the right place as all the other methods are directly related to schema json, whereas this method is based on a combination of schema and answer data.  Perhaps `form_helper` isn't the ideal location either, there might be a more sensible module to put it in.

app/questionnaire/rules.py
kierenbeckett: The capping happens a few lines down, `answer_count_minus_one` has additional capping to make it one less than the max.  I've really just reverted this code to how it was before #1689 as it turned out we don't need the offset functionality.

app/helpers/form_helper.py
benhowes: That's fair enough! It definitely looks like it should be reusable, happy for it to go somewhere else if that's more appropriate.

app/questionnaire/rules.py
benhowes: Ah right, yes!

kierenbeckett:
The question title uses a new context variable to get the name of the person in the repeating group instance.  I didn't worry about making it work for summaries as there's another card to sort that out.

MebinAbraham (commented):
So the primary question isn’t included as part of the  `MAX_REPEATS`, so if I add 1 + 25 repeats, I should get 26 whats your age questions?

Not sure what the expected behaviour is. If I add 1 primary + 25 repeats, I get asked only 25 whats your age questions, should it not be 26?.

For the repeating group, whats your age question, it looks like `group_instance: 0` is skipped.

benhowes (approved):

kierenbeckett:
The repeat type of group repeats over an arbitrary set of groups, there's no concept of a primary.  The limit should apply to the whole repeat, not individual groups within it.

The groups are repeated over in the order they are in the schema, so as `primary-group` is first it will always be included, it will be `repeating-group` that gets truncated (after 24).

MebinAbraham:
@kierenbeckett When you dump the answer store, `group_instance: 0` is skipped?
I might be wrong/confused, but have a look at this.
https://gist.github.com/MebinAbraham/191101326dd26eb8d983462091bc5eb0

MebinAbraham:
@kierenbeckett Got clarification from Adam, not confused now. Working as expected. 👍
Can rebase btw.

MebinAbraham (approved):

PR #1702, opened 2018-07-26
Introduce AnswerStoreUpdater

nealedj:

### How to review 
All tests should pass
Code coverage should not reduce
No changes to functionality

### Checklist

* [] New static content marked up for translation
* [] Newly defined schema content included in eq-translations repo

### DIFF
--- a/app/data_model/questionnaire_store.py
+++ b/app/data_model/questionnaire_store.py
@@ -53,7 +53,8 @@ class QuestionnaireStore:
            or all group instances within a group and block"""
 
         if 'location' in kwargs:
-            self.completed_blocks.remove(kwargs['location'])
+            if kwargs['location'] in self.completed_blocks:
+                self.completed_blocks.remove(kwargs['location'])
         else:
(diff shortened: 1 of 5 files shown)


benhowes:
@nealedj we're going to have some real fun merging this with: https://github.com/ONSdigital/eq-survey-runner/pull/1694!

jonnyshaw89:
@nealedj Shouldn't we have at least discussed changes like this before working on them? To get the opinion of the team and agree a priority of the work?

nealedj:
@jonnyshaw89 Yes we should have - this came out of my investigation into the relationship storage issue. Sorry about that.

It turns out that the fix is somewhere else.

We can close this PR if that's easier.

nealedj:
@benhowes Nah no big deal merging - 5 mins to change a few `g.schema` refs to `schema`.

nealedj:
Will reopen if somebody decides they want this.

benhowes:
I thought we could talk about this in a tech session?!

PR #1706, opened 2018-07-30
Code Coverage

benhowes:

This PR increases code coverage of the project to 100%. There are no logic changes.

**isnt this just a vanity metric?**: Yes, however it means that unrelated PR's which remove lines of code can fail the code-cov checks

**Oh yeah, like when?**: If you have a file where 98/100 (98%) lines are covered, then you remove 2  lines which (had coverage) by refactoring, but don't make changes affecting the untested lines, then you have 96/98 (97.9%) coverage. Reducing coverage in a PR is not acceptable from a PR.

**How come not only test code has changed then?**: I have had to make a few small refactors to make patching or testing easier - There should be no logic changes!

Where ever I've tested something, I've tried to ensure that I am adding tests which will be useful, rather than just getting coverage for the sake of it! e.g. QuestionnaireStore had really high coverage, but no unit tests.

### How to review 
Everything should be exactly the same as before, test a few surveys, dumping data etc.

### DIFF
--- a/app/data_model/questionnaire_store.py
+++ b/app/data_model/questionnaire_store.py
@@ -5,11 +5,12 @@ from app.questionnaire.location import Location
 
 
 class QuestionnaireStore:
-
     LATEST_VERSION = 1
 
-    def __init__(self, storage, version=LATEST_VERSION):
+    def __init__(self, storage, version=None):
         self._storage = storage
+        if version is None:
+            version = self.get_latest_version_number()
         self.version = version
         self.metadata = {}
         self.answer_store = AnswerStore()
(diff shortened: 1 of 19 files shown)


app/setup.py
```diff
@@ -178,27 +178,28 @@ def setup_secure_headers(application):
         frame_options='DENY')
 
 
-def setup_database(application):
-    application.config['SQLALCHEMY_TRACK_MODIFICATIONS'] = False
+def get_database_uri(application):
+    """ Returns database URI. Prefer SQLALCHEMY_DATABASE_URI over components."""
```
insacuri, line 182: I love it!

kierenbeckett (approved):

app/data_model/questionnaire_store.py
```diff
@@ -21,6 +22,9 @@ def __init__(self, storage, version=LATEST_VERSION):
         if version is not None:
             self.version = version
```
dcdarrell9: Do we need this?

dcdarrell9 (approved):

MebinAbraham:
@benhowes Squash and rebase?

benhowes:
@MebinAbraham will do, I've been trying to get master sorted out first. We're nearly there now.

PR #1707, opened 2018-07-30
Add CalculatedSummary block type

dwyeradam:
As a user
I want my figures I have entered to be presented to me as a total
So that I don't have to provide a total as well as a breakdown and I have the opportunity to check for any mistakes

### How to review 
test_calculated_summary.json

### Checklist

### DIFF
--- a/app/templating/view_context.py
+++ b/app/templating/view_context.py
@@ -1,27 +1,36 @@
 from flask_wtf import FlaskForm
-from flask_babel import gettext as _
-
+from app.jinja_filters import format_currency, format_number, format_unit, format_percentage
 from app.helpers.form_helper import get_form_for_location
 from app.templating.summary_context import build_summary_rendering_context
 from app.templating.template_renderer import renderer
-from app.templating.utils import get_title_from_titles
+from app.templating.utils import get_title_from_titles, get_question_title
+from app.questionnaire.rules import get_answer_store_value
 
 
-def build_view_context(block_type, metadata, schema, answer_store, schema_context, rendered_block, current_location, form):
+def build_view_context(block_type, metadata, schema, answer_store, schema_context, rendered_block, current_location,
+                       form):
     variables = None
     if schema.json.get('variables'):
         variables = renderer.render(schema.json.get('variables'), **schema_context)
 
-    if block_type in ('Summary', 'SectionSummary'):
+    if block_type == 'Summary':
         form = form or FlaskForm()
+        return build_view_context_for_final_summary(metadata, schema, answer_store, schema_context, block_type,
+                                                    variables, form.csrf_token)
+    if block_type == 'SectionSummary':
+        form = form or FlaskForm()
+        return build_view_context_for_section_summary(metadata, schema, answer_store, schema_context, block_type,
+                                                      variables, form.csrf_token, current_location.group_id)
 
-        return build_view_context_for_summary(metadata, schema, answer_store, schema_context, block_type, variables,
-                                              form.csrf_token, current_location)
+    if block_type == 'CalculatedSummary':
+        form = form or FlaskForm()
+        return build_view_context_for_calculated_summary(metadata, schema, answer_store, schema_context, block_type,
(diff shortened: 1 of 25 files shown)


app/templating/view_context.py
```diff
@@ -88,39 +97,148 @@ def build_view_context_for_question(metadata, schema, answer_store, current_loca
...
+
+    reduced_block['questions'] = questions
+
+    return reduced_block
+
+
+def _get_formatted_total(block, answer_store, group_instance, schema):
+    calculated_total = 0
+    for answer in block['calculation']['answers_to_calculate']:
+        if group_instance > 0 and not schema.answer_is_in_repeating_group(answer):
```
benhowes, line 207: Regarding behaviour with repeats:
1) Do we know that this should always be for the current `group_instance`? 
2) This will not return results a value for repeating answers either?

I can't see anything in the ticket/acceptance criteria around this, so thought we should clear up the expected behaviour

benhowes (commented):
Looks good, tested locally with no problems, really obvious test schema! Just one point to clarify which may well not require any test changes.

app/templating/view_context.py
benhowes: As discussed, there's no user need for this, so this is fine until the requirements change, if ever

benhowes (approved):

app/templating/view_context.py
```diff
@@ -88,39 +97,148 @@ def build_view_context_for_question(metadata, schema, answer_store, current_loca
...
+    for block in schema.blocks:
+        answers_in_block = schema.get_answers_by_id_for_block(block['id'])
+        answers_to_calculate = rendered_block['calculation']['answers_to_calculate']
+
+        answer_matches = set(answers_in_block.keys()) & set(answers_to_calculate)
+        if answer_matches:
+            blocks.append(_remove_unwanted_questions_answers(block, answers_in_block, answer_matches))
+
+    section = {
+        'title': '',
```
jonnyshaw89: Why is this empty?

app/templating/view_context.py
```diff
@@ -88,39 +97,148 @@ def build_view_context_for_question(metadata, schema, answer_store, current_loca
...
+        'groups': [
+            {
+                'id': group_id,
+                'title': '',
```
jonnyshaw89: and this?
should either have content or not be present

jonnyshaw89 (changes requested):

app/templating/view_context.py
dwyeradam: group.py requires 'title' key but I don't want to use it on the page. Any text here would show under the Boxed "Please review your answer and confirm these are correct."
I can get rid of the empty section title though.

dwyeradam:
I hope you don't hold this against me and considering it was my retro point but I squashed as I rebased. It was only deleting a single line after all.

app/templating/view_context.py
```diff
@@ -88,39 +97,147 @@ def build_view_context_for_question(metadata, schema, answer_store, current_loca
...
 
+def _build_calculated_summary_section_list(schema, rendered_block, group_id):
+    blocks = []
+    for block in schema.blocks:
+        answers_in_block = schema.get_answers_by_id_for_block(block['id'])
+        answers_to_calculate = rendered_block['calculation']['answers_to_calculate']
+
+        answer_matches = set(answers_in_block.keys()) & set(answers_to_calculate)
+        if answer_matches:
+            blocks.append(_remove_unwanted_questions_answers(block, answers_in_block, answer_matches))
```
dcdarrell9, line 172: is it worth maybe doing something like `blocks = [_remove_unwanted_questions_answers(block, answers_in_block, answer_matches)]`

dcdarrell9 (approved):
LGTM

app/templating/view_context.py
dwyeradam: This is within a loop. If I did that blocks would only contain the last block in the loop.

jonnyshaw89 (approved):

PR #1709, opened 2018-07-31
 Introduce AnswerStoreUpdater

nealedj:

### How to review 
All tests should pass
Code coverage should not reduce
No changes to functionality

### Checklist


nealedj:
Sorry - meant to create this on my fork.

PR #1708, opened 2018-07-31
Updated sdc-cryptography

benhowes:
### What is the context of this PR?
Updates sdc-cryptography, from https://github.com/ONSdigital/sdc-cryptography/pull/24

### How to review 
Nothing should break, it's just a minor version library update :)

PR #1711, opened 2018-08-01
Fix pipenv lockfile, unpin pytest

benhowes:
I pinned pytest yesterday due to a problem with resolving dependencies. An issue I raised about this has had some response now and the provided solution is viable!

This PR changes nothing other than updating pytest in the pipfile and the lockfile

Tidies up hacks from yesterday in https://github.com/ONSdigital/eq-survey-runner/pull/1708

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -4,7 +4,7 @@ name = "pypi"
 verify_ssl = true
 
 [dev-packages]
-pytest = "==3.6.4" # The latest pytest introduces dependency conflicts on the pluggy library, waiting for a fix.
+pytest = "*"
 "pep8" = "*"
 mock = "*"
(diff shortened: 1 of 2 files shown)


Pipfile
```diff
@@ -50,6 +50,7 @@ blinker = "*"
 humanize = "*"
 flask-talisman = "*"
 marshmallow = "*"
+pytest = "*"
```
insacuri: Why are we duplicating pytest?  Shouldn't it only be there as part of the dev packages?

benhowes: Good catch, that was not intended! I've re-pushed

insacuri (approved):

MebinAbraham (approved):

PR #1712, opened 2018-08-01
Update question definitions to support multiple content

MebinAbraham:
Modified definitions structure to allow content so we can display bullet points etc. Needed for: trello.com/c/rC82gUmp/2180-move-significant-changes-to-use-definition-pattern

### How to review 
- Make sure app works as expected.
- Definitions to now allow multiple contents ie multiple descriptions, lists etc.

### Checklist

Related PRs:
1. https://github.com/ONSdigital/eq-schema-validator/pull/72
2. https://github.com/ONSdigital/eq-translations/pull/20

### DIFF
--- a/data/en/test_question_definition.json
+++ b/data/en/test_question_definition.json
@@ -30,11 +30,25 @@
                     "id": "question",
                     "title": "Do you connect a LiFePO4 battery to your <em>photovoltaic system</em> to store surplus energy?",
                     "type": "General",
-                    "description": "",
                     "definitions": [{
-                        "title": "What is a photovoltaic system?",
-                        "description": "A typical photovoltaic system employs solar panels, each comprising a number of solar cells, which generate electrical power. PV installations may be ground-mounted, rooftop mounted or wall mounted. The mount may be fixed, or use a solar tracker to follow the sun across the sky."
-                    }],
+                            "title": "What is a photovoltaic system?",
+                            "content": [{
+                                "description": "A typical photovoltaic system employs solar panels, each comprising a number of solar cells, which generate electrical power. PV installations may be ground-mounted, rooftop mounted or wall mounted. The mount may be fixed, or use a solar tracker to follow the sun across the sky."
+                            }]
+                        },
+                        {
+                            "title": "Why use LiFePO4 batteries?",
+                            "content": [{
+                                "title": "3 Benefits of LifePO4 batteries."
+                            }, {
+                                "list": [
+                                    "LifePO4 batteries have a life span 10 times longer than that of traditional lead acid batteries. This dramatically reduces the need for battery changes.",
+                                    "Lithium iron phosphate batteries operate with much lower resistance and consequently recharge at a faster rate.",
+                                    "LifeP04 lightweight batteries are lighter than lead acid batteries, usually weighing about 1/4 less."
+                                ]
+                            }]
+                        }
+                    ],
                     "answers": [{
                         "type": "Radio",
                         "id": "radio-mandatory-answer",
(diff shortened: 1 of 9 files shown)


benhowes:
@MebinAbraham if I'm not mistaken, there is only one example of using the definition content. That one uses the `description` key, but perhaps it would be good to also use the `title` and `list` keys somewhere? Just to make sure no weird bugs around that creep in later.

Should there not also be at least an e2e test to ensure the content appears and is interactive as expected.

In both cases, that's not because I suspect the improvements don't work now, but just that future changes to other areas may have unintended consequences which go un-noticed without tests :)

data/en/test_question_definition.json
```diff
@@ -30,10 +30,11 @@
                     "id": "question",
                     "title": "Do you connect a LiFePO4 battery to your <em>photovoltaic system</em> to store surplus energy?",
                     "type": "General",
-                    "description": "",
                     "definitions": [{
                         "title": "What is a photovoltaic system?",
-                        "description": "A typical photovoltaic system employs solar panels, each comprising a number of solar cells, which generate electrical power. PV installations may be ground-mounted, rooftop mounted or wall mounted. The mount may be fixed, or use a solar tracker to follow the sun across the sky."
+                        "content": [{
+                            "description": "A typical photovoltaic system employs solar panels, each comprising a number of solar cells, which generate electrical power. PV installations may be ground-mounted, rooftop mounted or wall mounted. The mount may be fixed, or use a solar tracker to follow the sun across the sky."
```
benhowes: What about `titles` and `list`?

benhowes (changes requested):

data/en/test_question_definition.json
MebinAbraham: Fair point, will update tests. Currently, have tests for definition but ya only for description. 👍

tests/functional/spec/question_definitions.spec.js
```diff
@@ -61,5 +61,21 @@ describe('Component: Definition', function() {
...
+      return browser
+        .isVisible(DefinitionPage.definitionContent2()).should.eventually.be.false
+        .isVisible(DefinitionPage.definitionButton2()).should.eventually.be.false
+
+        // When
+        .click(DefinitionPage.definitionTitle2())
+
+        // Then
+        .waitForVisible(DefinitionPage.definitionContent2(), 300)
+        .waitForVisible(DefinitionPage.definitionButton2(), 300);
```
benhowes, line 76: Do these assert? As fas as I can tell they don't fail if the elements are not visible?

MebinAbraham: Yes, they'll fail if the element doesn't become visible within 300ms.

benhowes (approved):

scripts/test_schemas.sh
```diff
@@ -6,8 +6,8 @@ if [ "$1" == "--local" ] || [ "$2" == "--local" ]; then
 fi
 
 if [ "$run_docker" == true ]; then
-    docker pull onsdigital/eq-schema-validator
-    validator="$(docker run -d -p 5001:5000 onsdigital/eq-schema-validator)"
+    docker pull onsdigital/eq-schema-validator:update-definition-requirements
+    validator="$(docker run -d -p 5001:5000 onsdigital/eq-schema-validator:update-definition-requirements)"
     sleep 3
```
MebinAbraham: Reminder to revert before merging.

dcdarrell9 (approved):

insacuri (approved):

PR #1715, opened 2018-08-02
Flatten variant_flags in to top level claims object

benhowes:
Related to https://github.com/ONSdigital/ons-schema-definitions/pull/32#discussion_r207161581
Required changes to schema validation in https://github.com/ONSdigital/eq-schema-validator/pull/73

`variant_flags` was a special case and it was decided that it could be flattened in to the top level object. The knock on effects of this are:
1. Nothing uses the `object` validator for metadata. The actual content of objects was never validated and it seemed to have the potential to cause corner cases, so it's been removed.
2. The `variant_flags` did sometimes contain boolean values, however there was no way to validate these at the top level of the schema.

Some surveys now use `sexual_identity` in the claims rather than `varient_flags.sexual_identity`. there wasn't really any other place where functionality was being enabled/disabled from the metadata.

### How to review 
Nothing should break. Run some surveys from the launcher.

### DIFF
--- a/app/storage/metadata_parser.py
+++ b/app/storage/metadata_parser.py
@@ -15,6 +15,17 @@ def uuid_4_parser(plain_string):
     return str(uuid.UUID(plain_string))
 
 
+def boolean_parser(boolean_value):
+    """Strict matching of boolean values from JWT claims.
+    :param metadata:
+    :raises: TypeError - when a non boolean value has been loaded from the JWT.
+    :returns: A boolean value
+    """
+    if not isinstance(boolean_value, bool):
+        raise TypeError('Claim was not of type `bool`')
+    return boolean_value
+
+
 def clean_leading_trailing_spaces(metadata):
     for key, value in metadata.items():
         if isinstance(value, str):
@@ -26,8 +37,8 @@ def clean_leading_trailing_spaces(metadata):
 VALIDATORS = {
     'date': iso_8601_date_parser,
     'uuid': uuid_4_parser,
+    'boolean': boolean_parser,
     'string': lambda *args: None,
-    'object': lambda *args: None,
 }
 
 MANDATORY_METADATA = [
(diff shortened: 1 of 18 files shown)


jonnyshaw89:
Its used in `census_household.json` `variant_flags.sexual_identity` which should be changed to just `sexual_identity`

benhowes:
@jonnyshaw89 It's used in 2 other surveys too, however I this was intended to be a change which just means that it's no longer mandatory metadata? `variant_flags` does appear in the `metadata` key for each survey it's used in.

MebinAbraham:
If any `variant_flags` are used, it should be defined, ie https://github.com/ONSdigital/eq-survey-runner/blob/06cf7215f9a6799a6ffb20a9e9611e1fc0e0194d/data/en/census_household.json#L13-L25

At the time this is what was agreed upon. We only check that the `variant_flag` key is present, we don't validate the values.

app/storage/metadata_parser.py
```diff
@@ -99,11 +111,11 @@ def _validate_metadata_values_are_valid(claims, required_metadata):
             name = metadata_field['name']
             claim = claims.get(name)
             if claim:
-                VALIDATORS[metadata_field['validator']](claim)
                 logger.debug('parsing metadata', key=name, value=claim)
+                VALIDATORS[metadata_field['validator']](claim)
```
jonnyshaw89, line 115: 👍

app/storage/metadata_parser.py
```diff
@@ -99,11 +111,11 @@ def _validate_metadata_values_are_valid(claims, required_metadata):
...
 
     except (RuntimeError, ValueError, TypeError) as error:
-        logger.error('unable to parse metadata', exc_info=error)
+        logger.error('Unable to parse metadata for key `{}`'.format(name), exc_info=error)
```
jonnyshaw89: We should maintain structured logging format
`logger.error('Unable to parse metadata', key=name, exc_info=error)`

data/en/census_household.json
```diff
@@ -20,8 +20,8 @@
         "name": "region_code",
         "validator": "string"
     }, {
-        "name": "variant_flags",
-        "validator": "object"
+        "name": "sexual_identity",
+        "validator": "boolean"
```
jonnyshaw89, line 24: 👍

scripts/test_schemas.sh
```diff
@@ -6,8 +6,8 @@ if [ "$1" == "--local" ] || [ "$2" == "--local" ]; then
 fi
 
 if [ "$run_docker" == true ]; then
-    docker pull onsdigital/eq-schema-validator
-    validator="$(docker run -d -p 5001:5000 onsdigital/eq-schema-validator)"
+    docker pull onsdigital/eq-schema-validator:eq-2224-update-metadata-requirements
+    validator="$(docker run -d -p 5001:5000 onsdigital/eq-schema-validator:eq-2224-update-metadata-requirements)"
```
jonnyshaw89: remember to switch back

jonnyshaw89 (changes requested):

dcdarrell9:
Launching census_household / census_individual breaks for me...
@jonnyshaw89 is working on refactoring the go-launch, but this is where it's being set
https://github.com/ONSdigital/go-launch-a-survey/blob/1914c3d4d6c61b5346c2510018ed2f303d759746/authentication/auth.go#L136

jonnyshaw89 (approved):

dcdarrell9 (approved):

dwyeradam (approved):
Worthwhile PR with recent metadata changes. Ready to squash. I'd be happy to merge even though current Census schemas are affected.

PR #1717, opened 2018-08-02
Add started_at to final payload

insacuri:
https://trello.com/c/HGU9OpOe/2173-add-dates-to-metadata-for-downstream-systems

Needed to add the started at time to a survey as the downstream systems needed it.  It gets added to the metadata and written in the same place as 'submitted_at' in the final payload

### How to review 
Submit any survey and check that started_at appears in the payload at the top level of the json

### Checklist

### DIFF
--- a/app/submitter/converter.py
+++ b/app/submitter/converter.py
@@ -36,6 +36,7 @@ def convert_answers(metadata, schema, answer_store, routing_path, flushed=False)
           'instrument_id': 'yui789',
           'period': '2016-02-01'
         },
+        'started_at': '2016-03-06T15:28:05Z',
         'submitted_at': '2016-03-07T15:28:05Z',
         'metadata': {
           'user_id': '789473423',
@@ -57,6 +58,8 @@ def convert_answers(metadata, schema, answer_store, routing_path, flushed=False)
         'collection': _build_collection(metadata),
         'metadata': _build_metadata(metadata),
     }
+    if 'started_at' in metadata and metadata['started_at']:
+        payload['started_at'] = metadata['started_at']
     if 'case_id' in metadata and metadata['case_id']:
         payload['case_id'] = metadata['case_id']
     if 'case_ref' in metadata and metadata['case_ref']:
@@ -69,7 +72,7 @@ def convert_answers(metadata, schema, answer_store, routing_path, flushed=False)
     else:
         raise DataVersionError(schema.json['data_version'])
 
-    logger.debug('converted answer ready for submission')
+    logger.info('converted answer ready for submission')
     return payload
 
 
--- a/app/views/questionnaire.py
+++ b/app/views/questionnaire.py
@@ -1,7 +1,7 @@
 from functools import wraps
 import re
 from collections import defaultdict
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, timezone
 import humanize
 import simplejson as json
 from dateutil.tz import tzutc
@@ -177,6 +177,7 @@ def post_block(routing_path, schema, metadata, answer_store, eq_id, form_type, c
         return _save_sign_out(routing_path, current_location, form, schema, answer_store, metadata)
 
     if form.validate():
+        _set_started_at_metadata_if_required(form, metadata)
         _update_questionnaire_store(current_location, form, schema)
         next_location = path_finder.get_next_location(current_location=current_location)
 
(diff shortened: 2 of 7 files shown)


app/views/questionnaire.py
```diff
@@ -177,6 +177,7 @@ def post_block(routing_path, schema, metadata, answer_store, eq_id, form_type, c
         return _save_sign_out(routing_path, current_location, form, schema, answer_store, metadata)
 
     if form.validate():
+        _set_started_at_metadata_if_required(form, metadata)
```
benhowes, line 180: I only thought of this because another PR got reverted for breaking existing sessions - this will currently incorrectly set the `started_at` time to be the time of the first form submission after this gets deployed - it may be better to set this in the setup of new sessions, e.g. around here: https://github.com/ONSdigital/eq-survey-runner/blob/master/app/authentication/authenticator.py#L110?

benhowes (changes requested):
Not tested locally yet

app/submitter/converter.py
```diff
@@ -53,6 +53,7 @@ def convert_answers(metadata, schema, answer_store, routing_path, flushed=False)
         'origin': 'uk.gov.ons.edc.eq',
         'survey_id': survey_id,
         'flushed': flushed,
+        'started_at': metadata['started_at'],
```
benhowes: You told me to tell you that this is going to `KeyError` on existing surveys

app/views/questionnaire.py
benhowes: As discussed, this is not a problem!

benhowes (approved):

benhowes (changes requested):
Need to fix the key error.

benhowes (changes requested):
@insacuri it would be good to have a test for when you call `convert_answers` without a `metadata['started_at']`. Just to make sure that someone doesn't refactor later because it looks like it will always be present.

tests/app/submitter/test_converter.py
```diff
@@ -104,6 +104,18 @@ def test_convert_answers_flushed_flag_overriden_to_true(self):
...
 
+    def test_started_at_should_be_set_in_payload(self):
+        with self._app.test_request_context():
+
+            questionnaire = {
+                'survey_id': '021',
+                'data_version': '0.0.2'
+            }
+
+            answer_object = convert_answers(self.metadata, QuestionnaireSchema(questionnaire), AnswerStore(), {})
```
benhowes, line 116: This is a good test - it doesn't load a whole survey to use those couple of fields!

It would be helpful to add another test that deletes the `started_at` field before calling convert_answers. The benefit of this would be that future refactors or changes where someone wants to make it rely on having `started_at` in the metadata would fail, leading them to see your test and realise it's like that for a reason!

benhowes (changes requested):

tests/app/submitter/test_converter.py
insacuri: If `started_at` isn't in the metadata then it won't fall over, it just won't be included as I don't want to tank submissions that are in flight.  I'll still do a test that shows that if it's not in the metadata then it wont appear in the final submission.

benhowes (approved):

app/views/questionnaire.py
jonnyshaw89: Ad discussed where with who?

app/views/questionnaire.py
```diff
@@ -328,6 +329,13 @@ def get_view_submission(schema, eq_id, form_type):  # pylint: disable=unused-arg
     return redirect(url_for('post_submission.get_thank_you', eq_id=eq_id, form_type=form_type))
 
 
+def _set_started_at_metadata_if_required(form, metadata):
+    questionnaire_store = get_questionnaire_store(current_user.user_id, current_user.user_ik)
+    if not questionnaire_store.answer_store.answers and len(form.data) > 1:
+        logger.info('first answer about to be stored. writing started_at time to metadata')
```
jonnyshaw89: any value in logging the start time?

tests/app/submitter/test_converter.py
jonnyshaw89: I like the consideration of "in flight" data

app/views/questionnaire.py
insacuri: whatcha mean sorry?  You mean is it worth logging this line?  or are you saying I should include the started_at time to the log line itself?

jonnyshaw89: as in should we also log the time?

app/views/questionnaire.py
insacuri: Between Ben and I, and it spawned the next comment in the converter about the KeyError (as that's where the actual problem was)

app/views/questionnaire.py
insacuri: Yea sure, I'll get that done in a bit.

insacuri: Done

jonnyshaw89 (approved):

PR #1718, opened 2018-08-03
Expose schema endpoint

jonnyshaw89:
Add an endpoint to get the schema json

### How to review 
Does it return the schema json?

### DIFF
--- a/app/views/schema.py
+++ b/app/views/schema.py
@@ -1,0 +1,25 @@
+from os import listdir
+from os.path import isfile, join
+
+from flask import Blueprint, jsonify
+from app.utilities.schema import load_schema_from_params
+
+schema_blueprint = Blueprint('schema', __name__)
+
+
+@schema_blueprint.route('/schemas/<eq_id>/<form_type>', methods=['GET'])
+def get_schema_json(eq_id, form_type):
+    try:
+        schema = load_schema_from_params(eq_id, form_type)
+
+        return jsonify(schema.json)
+    except FileNotFoundError:
+        return 'Schema Not Found', 404
+
+
+@schema_blueprint.route('/schemas', methods=['GET'])
+def list_schemas():
+    schema_path = 'data/en'
+    onlyfiles = [f for f in listdir(schema_path) if isfile(join(schema_path, f)) and f.endswith('.json')]
+
+    return jsonify(onlyfiles)
--- a/tests/integration/views/test_schema.py
+++ b/tests/integration/views/test_schema.py
@@ -1,0 +1,26 @@
+import json
+
+from tests.integration.integration_test_case import IntegrationTestCase
+
+
+class TestSchema(IntegrationTestCase):
+    def test_get_schema_json(self):
+        self.get('/schemas/test/textfield')
+        response = self.getResponseData()
+        parsed_json = json.loads(response)
+
+        self.assertIn('title', parsed_json)
+        self.assertEqual(parsed_json['title'], 'Other input fields')
+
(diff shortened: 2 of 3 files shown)


app/views/schema.py
```diff
@@ -0,0 +1,22 @@
...
+
+from flask import Blueprint, jsonify
+from app.utilities.schema import load_schema_from_params
+
+schema_blueprint = Blueprint('schema', __name__)
+
+
+@schema_blueprint.route('/schemas/<eq_id>/<form_type>', methods=['GET'])
+def get_schema_json(eq_id, form_type):
+    schema = load_schema_from_params(eq_id, form_type)
```
benhowes: Could this be wrapped in something so that it can 404 rather than 500 if the schema isn't there?

app/views/schema.py
```diff
@@ -0,0 +1,22 @@
...
+
+    return jsonify(schema.json)
+
+
+@schema_blueprint.route('/schemas', methods=['GET'])
+def list_schemas():
+    schema_path = "data/en"
+    onlyfiles = [f for f in listdir(schema_path) if isfile(join(schema_path, f)) and f.endswith(".json")]
```
benhowes: Given they're accessed by `/<eq_id>/<form_type>`, I think we should abstract that here. What we make here should be a good start for a schema store API too right? Perhaps a response like:

```javascript
{
   "results": [ // results key to make adding metadata easier
      {
        "eq_id": "test",
        "form_type": "metadata_validation",
        // "date-changed": "2019-10-15T00:00:00Z" - Other things could go here later without breaking anythign
      }
   ]
   "pages": 1, // So you can add paging later
}
```

tests/integration/views/test_schema.py
```diff
@@ -0,0 +1,22 @@
+import json
+
+from tests.integration.integration_test_case import IntegrationTestCase
+
+
+class TestSchema(IntegrationTestCase):
+    def test_get_schema_json(self):
```
benhowes, line 7: Suggest testing an invalid get too :)

benhowes (changes requested):

app/views/schema.py
jonnyshaw89: We already have something like that, its just a case of actually moving the schemas out. 
https://github.com/ONSdigital/eq-survey-register
Im wary of over-engineering given that its increasingly likely that we will move the schemas out into a private repo in the near future.

tests/integration/views/test_schema.py
jonnyshaw89: 👍

app/views/schema.py
benhowes: Okay, how about returning the link HATEOAS stylie, like you do here then: https://github.com/ONSdigital/eq-survey-register/blob/master/api.go#L62. The survey-launcher shouldn't have the logic for parsing the filenames was my main point, but the response structure got mixed in with it.

jonnyshaw89: Yeh I did think about HATEOAS but then that leads to launcher using the url to launch the surveys, rather than `eq_id` and `form_type` which wont work as the schemas in runner don't define `eq_id` and `form_type` on them

insacuri (approved):

app/views/schema.py
benhowes: Okay. I guess my point rested on making this a longer term solution, but if it's not possible to know what that's going to look like, it's fine as it is :+1:

benhowes (approved):

PR #1720, opened 2018-08-06
Added accordion to summary page

dcdarrell9:
- Added accordions to the summary page
- [Card](https://trello.com/c/aRJ2ZW4M/2077-apply-new-final-summary-design-surveys-with-sections)

### How to review 
- Run alongside schema-validator (onsdigital/eq-schema-validator:eq-2077-apply-summary-add-collapsible-key)
- Run test_section_summary and check section summary and summary are complete, section summary displays without accordion, and summary with accordion.
- Add a collapsible key to 1_0005 schema to test that accordions work.
- Run some other surveys to see that the accordions aren't there.
- Check if any designs are wrong

##
- *Save and Complete button will look wrong until moved to header*
##
### Checklist


app/templates/partials/summary/summary.html
```diff
@@ -1,28 +1,30 @@
 <div class="summary">
-	{%- for group in content.summary.groups -%}
-        {%- if group.title -%}
-            <h2 class="summary__title neptune" id="{{ group.id }}">{{ group.title }}</h2>
-        {%- endif %}
+    {%- if metadata.form_type == '0005' and content.summary.summary_type == 'Summary' -%}
```
benhowes: Is there no better way to decide if the `final-summary.html` should be used?

benhowes:
@dcdarrell9 this should result in some changes to e2e tests. If nothing else a test should ensure that the accordions can be rendered and open/closed. As per, that's not to check that it works now, but to ensure that other changes in future don't accidentally break it!

app/templates/partials/summary/summary.html
MebinAbraham: We shouldn't be hardcoding survey specific logic. Why do you need the `== '0005'`?. `Summary` should be the final section and `SectionSummary` for the individual sections.

What am I missing?

dcdarrell9: The reason this is here for now is because the card says we only wanted it for mwss, and if it was used as is on the other surveys it would break, I'd say the 'Summary' type would probably be changed and then we can change some of the logic then. But yeah I agree I don't like it done this way, couldn't think of another for now

jonnyshaw89: In the schema i think you just remove the block type for summary and use a final summary block type?

app/templates/partials/accordion.html
```diff
@@ -1,14 +1,18 @@
...
-        </div>
+  <div class="accordion__controls">
+    <button class="btn btn--secondary btn--small accordion__control js-accordion-toggle-all u-wa--@xs" data-ga="click" data-ga-category="Preview Survey" data-ga-action="Show all" data-close-all-label="Hide all" data-open-all-label="Show all" data-open="false" aria-hidden="true">Show all</button>
+  </div>
+  <div class="js-accordion-content">
+    {% for item in accordion_data %}
+    <h3 class="accordion__title js-accordion-title u-cf icon--accordion" data-js-accordion-event-label="{{item.title}}">
+      <span class="accordion__title-text">{{item.title}}</span>
+      <button class="btn btn--secondary btn--small accordion__title-right accordion-unhide@m js-accordion-toggle u-wa--@xs" data-close-label="Hide" data-open-label="Show" aria-hidden="true">Show</button>
+    </h3>
```
MebinAbraham: You need to mark all new static content for translation and update the `messages.pot` file.

MebinAbraham (changes requested):

MebinAbraham:
@dcdarrell9 This change is to both update accordion's design as well as to implement it for summaries?

app/templates/partials/introduction/preview.html
```diff
@@ -18,18 +18,19 @@ <h3 class="venus u-mt-s">{{ content.title }}</h3>
...
-        <div class="mars accordion__controls">
-            <a class="accordion__control js-accordion-open-all u-hidden" data-ga="click" data-ga-category="Preview Survey" data-ga-action="Show all" tabindex="0" aria-hidden="true">{{ _("Show all") }}</a>
-            <a class="accordion__control js-accordion-close-all u-hidden" data-ga="click" data-ga-category="Preview Survey" data-ga-action="Hide all" tabindex="0" aria-hidden="true">{{ _("Hide all") }}</a>
+        <div class="accordion__controls">
+            <button class="btn btn--secondary btn--small accordion__control js-accordion-toggle-all u-wa--@xs" data-ga="click" data-ga-category="Preview Survey" data-ga-action="Show all" data-close-all-label="{{ _("Hide all") }}" data-open-all-label="{{ _("Show all") }}" data-open="false" aria-hidden="true">{{ _("Show all") }}</button>
         </div>
-        <dl class="js-accordion-content">
+
+        <div class="js-accordion-content">
             {% for question in intro.questions %}
```
MebinAbraham, line 26: I know its existing and you are updating the markup, but can we not use the `partials/accordion.html` here to reduce duplication?

app/templates/partials/summary/final-summary.html
```diff
@@ -0,0 +1,44 @@
+<div class="accordion js-accordion">
```
MebinAbraham: Since we didn't like the term `final-summary` because essentially they are all summaries, can we now not name this `collapsed-summary`, since that is the key we use to display this?

AndrewTorrance (changes requested):
I think it needs a test that the individual section showing and hiding works . The tests we have use only the show all button. So if a change elsewhere causes this to not work we would not currently detect an error.

app/templating/view_context.py
```diff
@@ -141,6 +142,7 @@ def build_view_context_for_calculated_summary(metadata, schema, answer_store, sc
                                                         metadata, current_location.group_instance, formatted_total),
         'title': get_question_title(rendered_block, answer_store, schema, metadata,
                                     current_location.group_instance) % dict(total=formatted_total),
+        'collapsible': _get_collapsible_final_summary(rendered_block),
```
dwyeradam: I don't think that the Calculated Summary should ever be collapsed so should be taken out.

tests/app/templating/test_summary_context.py
```diff
@@ -91,10 +97,11 @@ def test_build_view_context_for_summary(self):
...
 
         context = build_view_context_for_final_summary(self.metadata, self.schema, self.answer_store,
-                                                       self.schema_context, self.block_type, variables, csrf_token)
+                                                       self.schema_context, self.block_type, variables, csrf_token,
+                                                       self.rendered_block)
 
         self.check_context(context)
-        self.assertEqual(len(context['summary']), 4)
+        self.assertEqual(len(context['summary']), 5)
         self.assertTrue('is_view_submission_response_enabled' in context['summary'])
```
dwyeradam, line 105: We should also add a test to ensure new context is here.
`self.assertTrue('collapsible' in context['summary'])`

app/templating/view_context.py
```diff
@@ -98,14 +98,15 @@ def build_view_context_for_question(metadata, schema, answer_store, current_loca
...
-                                         block_type, variables, csrf_token):
+                                         block_type, variables, csrf_token, rendered_block):
     section_list = renderer.render(schema.json, **schema_context)['sections']
 
     context = build_view_context_for_summary(schema, section_list, answer_store, metadata,
                                              csrf_token, block_type, variables)
 
     context['summary'].update({
         'is_view_submission_response_enabled': is_view_submitted_response_enabled(schema.json),
+        'collapsible': _get_collapsible_final_summary(rendered_block),
```
dwyeradam: Do we really need a separate method for this?

`'collapsible': rendered_block.get('collapsible', False),`

tests/functional/spec/features/calculated_summary.spec.js
```diff
@@ -13,7 +13,7 @@ const NumberTotalPlaybackPage = require('../../pages/features/calculated_summary
 const SummaryPage = require('../../pages/summary.page.js');
 const ThankYouPage = require('../../pages/thank-you.page.js');
 
-describe('@watch Feature: Calculated Summary', function() {
+describe('Feature: Calculated Summary', function() {
```
dwyeradam: Ooops. Thank you. 👍

dwyeradam (changes requested):
Couple of little points.

app/templating/view_context.py
dwyeradam: Also strictly speaking we don't need to pass `block_type` now as we can get that from the `rendered_block['type']`.

data/en/1_0005.json
```diff
@@ -2125,6 +2125,7 @@
         "groups": [{
             "blocks": [{
                 "type": "Summary",
+                "collapsible": true,
```
dwyeradam: Have we had business approval for this?

dwyeradam (dismissed):

AndrewTorrance (approved):

app/assets/favicons/browserconfig.xml
```diff
@@ -2,7 +2,7 @@
 <browserconfig>
     <msapplication>
         <tile>
-            <square150x150logo src="https://cdn.ons.gov.uk/sdc/v1.4.4/favicons/mstile-150x150.png"/>
+            <square150x150logo src="https://cdn.ons.gov.uk/sdc/v1.5.0/favicons/mstile-150x150.png"/>
```
insacuri: Does this need to be 1.5.1? Not sure if the other header/footer branch updated this file

dcdarrell9: Yeah, you're right, basically anything with 1.5.0 should be 1.5.1 👍

MebinAbraham (commented):
Updated `message.pot` file?

MebinAbraham (commented):
Since a final section should only ever be collapsed if there are any section summaries, is it not worthwhile to enforce if the summary is marked as collapsible then ensure there is actually at least 1 section summary?.

Happy with everything else. Also ideally, we should have a review on this from a front-end dev as well.

MebinAbraham (approved):
Approving because the suggestion above would be enforced in eq-schema-validator anyway.

insacuri (approved):

bitdivision:
@dcdarrell9 Do you know what's going on with Codacy?

PR #1721, opened 2018-08-06
Ensure session_data is available for language_code

dwyeradam:
We had a 500 error in Production where session was available but session_data was not. This stops that scenario from happening again.

### DIFF
--- a/app/setup.py
+++ b/app/setup.py
@@ -335,10 +335,10 @@ def setup_babel(application):
     def get_locale():  # pylint: disable=unused-variable
         session = get_session_store()
 
-        if not session:
-            return None
+        if session and session.session_data:
+            return session.session_data.language_code
 
-        return session.session_data.language_code
+        return None
 
     @application.babel.timezoneselector
     def get_timezone():  # pylint: disable=unused-variable


dwyeradam:
If anyone can write a test to recreate the 500 then they get a bar of chocolate.

benhowes:
@dwyeradam, I've been having a look at this - was wondering if you have a traceback?

I was wondering if the following line is ever logged too: https://github.com/ONSdigital/eq-survey-runner/blob/master/app/data_model/session_store.py#L76?

Probably not on to anything, but trying to find ways in which the user could have a valid cookie, but no session data. 

It does look like something should be raised if the session data is missing though, since they need to login again to create a new populated session right?

benhowes:
Jon pointed me in the direction of the stack trace.

I think I might have found something though, though it doesn't look like the cause of this issue, I'll raise a new PR.

The save method on the session does not return:
https://github.com/ONSdigital/eq-survey-runner/blob/7f076204594992a005e3414636b941b840802210/app/data_model/session_store.py#L35-L43

However the return value is used as an assignment
https://github.com/ONSdigital/eq-survey-runner/blob/7f076204594992a005e3414636b941b840802210/app/globals.py#L45

This might not be a problem now, but it would cause interesting issues when dynamodb is the main store, or if read replica's are used with RDS.

app/setup.py
```diff
@@ -335,10 +335,10 @@ def setup_babel(application):
     def get_locale():  # pylint: disable=unused-variable
         session = get_session_store()
 
-        if not session:
-            return None
+        if session and session.session_data:
+            return session.session_data.language_code
 
-        return session.session_data.language_code
+        return None
```
benhowes, line 341: Ideally this would fallback to using HTTP `Accept-Language`, or some other flask-bable tool i think? because it looks like this would always render the error in english.

benhowes (changes requested):

app/setup.py
jonnyshaw89: I believe the intention is to default to english

benhowes: Oh okay, if that is the case I can approve this.

insacuri (approved):

app/setup.py
benhowes: Might need changing in future if non english error pages are required for logged out users.

benhowes (approved):

app/setup.py
dwyeradam: Yes, expected behaviour is to default to English. Not sure if we would ever change that.

PR #1722, opened 2018-08-07
Session store was being reloaded from the database after creation

benhowes:
This change populates `g._session_store` when the session is created - it looks that this was always the intended behaviour.

There's no easy way to test this, since the moto library does not simulate eventual consistency (I checked through the code)

The problem would be that in some cases the document would be written, but might not be available for 'a bit' of time, if syncing between multiple instances is occurring.

### How to review 
The session should not be read from the database immediately after creation.

### DIFF
--- a/app/data_model/session_store.py
+++ b/app/data_model/session_store.py
@@ -42,6 +42,8 @@ class SessionStore:
 
             data_access.put(self._eq_session)
 
+        return self
+
     def delete(self):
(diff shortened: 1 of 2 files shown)


jonnyshaw89 (approved):

MebinAbraham (approved):
Nothing seems to have broke :)

PR #1725, opened 2018-08-07
Hide personal data from page titles

### LINKED ISSUE #1723: Page Title Including Answers when getting Titles
### Expected behaviour
Answers should be replaced with ... for the page title. 

### Actual behaviour
Actual answers are already rendered in title so not being replaced. 

Answer test_titles_conditional_within_repeating_block.json. Add at least one person and when in that person's section select that you're answering on their behalf.

### Technical information
End point for error is in questionnaire.py and _get_question_title_for_page(.....)

#### Screenshot
screen shot 2018-08-07 at 11 59 27

benhowes:
Closes #1723

This is because page titles are sent to analytics services, such as google analytics.

No page title's should contain answers, or metadata anymore, but should show `…` instead. This should not affect the page content though,

### How to review 
Run a survey which uses titles containing templating, e.g. `test_titles_conditional_within_repeating_block.json`. You should see something like this in the tab title:

screenshot from 2018-08-07 17-13-41

But the page content should still show:

screenshot from 2018-08-07 17-13-28

Net code removal :fist_oncoming:

### DIFF
--- a/app/helpers/template_helper.py
+++ b/app/helpers/template_helper.py
@@ -40,7 +40,7 @@ def with_metadata_context(func):
         metadata = get_metadata(current_user)
         metadata_context = build_metadata_context(metadata)
 
-        return func(*args, metadata=metadata_context, **kwargs)
+        return func(*args, metadata_context=metadata_context, **kwargs)
 
     return metadata_context_wrapper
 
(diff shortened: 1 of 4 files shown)


MebinAbraham (approved):
Works fine. Maybe we should consider better naming for metadata/metadata_context/claims etc in the future.

AndrewTorrance (approved):

PR #1727, opened 2018-08-09
Add repeating over multiple groups

benhowes:
For the most part, this is the same as #1700, however it's been modified to add an upgrade for the answer store and testing of that. 

### How to review 
For the answer_store upgrade:
1. In docker-compose, run a survey which contains repeating groups from the master branch. Complete some answers so that you have multiple `group_instance` values.
2. Dump those answers and keep them somewhere for later comparison (e.g. keep the tab open!)
3. Then stop docker-compose, check this branch out and re-start docker-compose
4. Dump the answers again - they should have been upgraded to have `group_instance_id` added.

I've not changed anything relating to the addition of the group iteration from the previous PR, so the testing from that should still be valid, however check the original PR for testing that.

### DIFF
--- a/app/data_model/answer_store.py
+++ b/app/data_model/answer_store.py
@@ -3,14 +3,18 @@ import re
 from datetime import datetime
 from jinja2 import escape
 import simplejson as json
+from structlog import get_logger
+
+logger = get_logger()
 
 
 class Answer:
-    def __init__(self, answer_id, value, group_instance=0, answer_instance=0):
+    def __init__(self, answer_id, value, group_instance_id=None, group_instance=0, answer_instance=0):
         if answer_id is None or value is None:
             raise ValueError("Both 'answer_id' and 'value' must be set for Answer")
 
         self.answer_id = answer_id
+        self.group_instance_id = group_instance_id
         self.group_instance = group_instance
         self.answer_instance = answer_instance
         self.value = value
@@ -18,12 +22,13 @@ class Answer:
     def matches(self, answer):
         """
         Check to see if two answers match.
-        Two answers are considered to match if they share the same block_id, answer, answer_instance, group_id and group_instance.
+        Two answers are considered to match if they share the same answer_id, answer_instance and group_instance_id.
 
         :param answer: An answer to compare
         :return: True if both answers match, otherwise False.
         """
         return self.answer_id == answer.answer_id and \
+            self.group_instance_id == answer.group_instance_id and \
             self.group_instance == answer.group_instance and \
             self.answer_instance == answer.answer_instance
 
@@ -38,6 +43,7 @@ class Answer:
         return self.matches(Answer(
             answer_dict['answer_id'],
             answer_dict['value'],
+            answer_dict['group_instance_id'],
             answer_dict['group_instance'],
             answer_dict['answer_instance'],
         ))
@@ -155,14 +161,14 @@ class AnswerStore:
             escaped.append(answer)
         return self.__class__(existing_answers=escaped)
 
-    def filter(self, answer_ids, group_instance=None, answer_instance=None,
-               limit=None):
+    def filter(self, answer_ids=None, group_instance=None, group_instance_id=None, answer_instance=None, limit=None):
         """
         Find all answers in the answer store for a given set of filter parameter matches.
         If no filter parameters are passed it returns a copy of the instance.
 
         :param answer_ids: The answer ids to filter results by
         :param answer_instance: The answer instance to filter results by
+        :param group_instance_id: The group instance ID to filter results by
         :param group_instance: The group instance to filter results by
         :param limit: True | False Limit the number of answers returned
         :return: Return a new AnswerStore object with filtered answers for chaining
@@ -172,6 +178,7 @@ class AnswerStore:
         filter_vars = {
             'answer_id': answer_ids,
             'answer_instance': answer_instance,
+            'group_instance_id': group_instance_id,
             'group_instance': group_instance,
         }
 
@@ -201,7 +208,7 @@ class AnswerStore:
         :param answer_instance: The answer instance to filter results to remove
         :param group_instance: The group instance to filter results to remove
         """
-        for answer in self.filter(answer_ids, group_instance, answer_instance):
+        for answer in self.filter(answer_ids, group_instance=group_instance, answer_instance=answer_instance):
             self.answers.remove(answer)
 
     def get_hash(self):
@@ -213,21 +220,18 @@ class AnswerStore:
         return hash(json.dumps(self.answers, sort_keys=True))
 
     def upgrade(self, current_version, schema):
+        """
+            Upgrade the answer_store to the latest version
 
-        # Upgrade from version 0 to version 1
-        if current_version == 0:
-            # Update Date formats
-            for answer in self.answers:
-                answer_schema = schema.get_answer(answer['answer_id'])
+            :param current_version: The current version integer of the answer store
+            :param schema: The new schema
+        """
+        desired_version = len(UPGRADE_TRANSFORMS)
 
-                if answer_schema:
-                    if answer_schema['type'] == 'Date':
-                        answer['value'] = datetime.strptime(answer['value'], '%d/%m/%Y').strftime('%Y-%m-%d')
-                        continue
-
-                    if answer_schema['type'] == 'MonthYearDate':
-                        answer['value'] = datetime.strptime(answer['value'], '%m/%Y').strftime('%Y-%m')
-                        continue
+        for version in range(current_version, desired_version):
+            transform = UPGRADE_TRANSFORMS[version]
+            logger.info('Upgrading answer store version', current_version=current_version, new_version=version, transform=transform.__name__)
+            transform(self, schema)
 
 
 def number_else_string(text):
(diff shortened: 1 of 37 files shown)


app/submitter/convert_payload_0_0_2.py
```diff
@@ -1,3 +1,5 @@
+
```
insacuri, line 1: Super minor, but are the 2 extra lines needed?

benhowes: Huh, that's slipped through from the original PR, I'll sort it out and push it with any other changes which get requested

insacuri (dismissed):
LGTM, on a code level anyway.  The solution to make it backwards compatible looks reasonable, but someone more familiar might see something I didn't ^^

jonnyshaw89:
@benhowes I think it might be good to squash all of your commits together and leave the original commit (so there will be 2 commits) that way i can see that changes you have made on top of the original work?

dwyeradam (commented):
This is from the original PR but there seems to be inconsistency about what gets the group_instance_id. For example if you start LMS_1 in this branch everything but the names will get a group_instance_id. If you get a partially completed survey back then the upgrade just gives it to everything. 
So my question is should everything get a group_instance_id or should we only need it for repeats. A survey without repeat such as MBS gets it for everything which seems overkill.

app/data_model/answer_store.py
```diff
@@ -216,21 +219,18 @@ def get_hash(self):
...
-                        continue
+            :param current_version: The current version integer of the answer store
+            :param schema: The new schema
+        """
+        desired_version = len(UPGRADE_TRANSFORMS)
 
-                    if answer_schema['type'] == 'MonthYearDate':
-                        answer['value'] = datetime.strptime(answer['value'], '%m/%Y').strftime('%Y-%m')
-                        continue
+        for version in range(current_version, desired_version):
```
jonnyshaw89, line 231: 👍

app/data_model/answer_store.py
```diff
@@ -245,3 +245,47 @@ def natural_order(key):
...
+        location = Location(
+            group_id=group_id,
+            group_instance=answer['group_instance'],
+            block_id=block_id,
+        )
+        # `get_group_instance_id` handles providing a consistent group_instance_id
+        answer['group_instance_id'] = get_group_instance_id(schema, answer_store, location)
+
+
+UPGRADE_TRANSFORMS = (
```
jonnyshaw89: This is good, my only worry is that we access these by index, would it not be better to explicitly define them by index? i know this will make the code less "pure" but might add to readability? open for comment....

benhowes:
The upgrade I added does just add `group_instance_id` to everything - I was under the impression that this was the desired functionality now, but looking at it, the code is selective about what gets a `group_instance_id` and when.

This code (and similar things in a couple of other places) only add `group_instance_id`s when explicitly required. 

https://github.com/ONSdigital/eq-survey-runner/blob/9337df97fea6b7c6dc530ec52280caaf2618732e/app/helpers/form_helper.py#L81-L84

app/globals.py
```diff
@@ -18,6 +19,10 @@ def get_questionnaire_store(user_id, user_ik):
         storage = EncryptedQuestionnaireStorage(user_id, user_ik, pepper)
         store = g._questionnaire_store = QuestionnaireStore(storage)
 
+    if store.metadata:  # When creating the questionnaire storage, there is no metadata for upgrading
```
jonnyshaw89, line 22: I don't understand what this is for?

jonnyshaw89 (changes requested):

benhowes:
As discussed in the tech session, I'll implement something which allows only populating the `group_instance_id` when the answer is to be used in a repeating group.

app/globals.py
benhowes: When this is called: https://github.com/ONSdigital/eq-survey-runner/blob/7846f721c89ec3aa0cc27c4544188c1796b3539f/app/authentication/authenticator.py#L109 it appeared that metadata had not been populated, so it was unable to fetch the schema

app/data_model/answer_store.py
benhowes: My thought was to try and do the most simple thing for now. I think that longer term a means to upgrade other parts of the questionnaire store is required too - potentially the upgrading should happen in the `QuestionnaireStore`, since that's where the version number is stored, and it has access to the other bits of data which **could** need updating at some point (e.g. the metadata would need upgrading if you change a type or add a new claim).

I didn't want to drag us down that rabbit hole as part of this though, I felt this was the minimum required to make it obvious what the upgrades are and will make it easy to patch in a couple more updates if needed (without a refactor), but it might be worth looking at refactoring the store upgrader a bit sooner

app/globals.py
jonnyshaw89: Maybe `schema = load_schema_from_metadata(store.metadata)` should be within the if at 
https://github.com/ONSdigital/eq-survey-runner/pull/1727/files/283be2434be900ecbaeea3e758300e5f815f6a8b..9337df97fea6b7c6dc530ec52280caaf2618732e#diff-dc061d8bc41d7eabbb038d35ad5ae467R32

otherwise we are loading the schema when we don't need it.

jonnyshaw89 (changes requested):

app/questionnaire/rules.py
```diff
@@ -116,11 +116,18 @@ def evaluate_repeat(repeat_rule, answer_store, answer_ids_on_path):
...
+            answer_ids_in_group = schema.get_answer_ids_for_group(group_id)
+            group_instances |= {a['group_instance_id'] for a in answer_store.filter(answer_ids=answer_ids_in_group)}
+
+        no_of_repeats = len(group_instances)
     else:
         repeat_functions = {
             'answer_value': _get_answer_value,
-            'answer_count': lambda f: _get_answer_count(f, repeat_rule),
-            'answer_count_minus_one': lambda f: min(_get_answer_count_minus_one(f), MAX_REPEATS - 1),
+            'answer_count': len,
```
insacuri, line 129: How is this working?  There's no variable called `len` in this module

jonnyshaw89: there is a function in python called `len` thou

tests/integration/census/test_census_individual_submission_data.py
```diff
@@ -21,122 +21,122 @@ def get_expected_submission_data():
             {'answer_id': 'middle-names',
              'answer_instance': 0,
              'group_instance': 0,
-             'group_instance_id': 'household-member-3',
+             'group_instance_id': None,
```
jonnyshaw89: Should these be `None` or should we just omit the key?

benhowes: They are set to None by https://github.com/ONSdigital/eq-survey-runner/blob/c86bb548a507694ff2c7efb26075f57327528e91/app/data_model/answer_store.py#L17, however this could be changed?

app/questionnaire/rules.py
insacuri: Yea, but when you call it you have to give it an object don't you? unless there's some hidden 'when called without arguments, it counts the length of the last functions return' or something?

benhowes: That's just a map of funtions, it's called by reference later on https://github.com/ONSdigital/eq-survey-runner/blob/c86bb548a507694ff2c7efb26075f57327528e91/app/questionnaire/rules.py#L141

tests/integration/census/test_census_individual_submission_data.py
benhowes: I spoke with @dwyeradam about this and it will make the answer comparison more complex if the `group_instance_id` key/attr is missing some of the time, so suggest leaving it in. Could filter it out in submission dumping/flushing perhaps?

app/globals.py
benhowes: I figured that it's only fetching a reference to the cached schema? I didn't want to put the logic for fetching the schema inside the QuestionnaireStore unit particuarly

jonnyshaw89:
Ive run through `census_household` when completing household composition and subsequent repeating group the `group_instance_id` isn't set.
```
{
"answer_id": "first-name",
"answer_instance": 0,
"group_instance": 0,
"group_instance_id": null,
"value": "Jon"
},
{
"answer_id": "middle-names",
"answer_instance": 0,
"group_instance": 0,
"group_instance_id": null,
"value": ""
},
{
"answer_id": "last-name",
"answer_instance": 1,
"group_instance": 0,
"group_instance_id": null,
"value": "Shaw"
},
```
```
{
"answer_id": "details-correct-answer",
"answer_instance": 0,
"group_instance": 0,
"group_instance_id": null,
"value": "Yes, this is my full name"
},
{
"answer_id": "over-16-answer",
"answer_instance": 0,
"group_instance": 0,
"group_instance_id": null,
"value": "Yes"
}
```

benhowes:
@jonnyshaw89 yesterday we said that we'd add group_instance_id's where they are needed - which is for anything which has a repeat type `group`. Answers which increment `answer_instance` do not require it to work properly. 

If you were to use a repeating group instead of a repeating answer, then the `group_instance_id` would be set.

I did investigate if it was needed for RepeatingAnswers, but it doesn't need to be there at the moment - didn't want to over engineer it.

jonnyshaw89:
I though that would have included the census repeating household?
I can see in the schema that this is based on an `answer_count` but i thought that half of the reason behind this change was to make those links more explicit bu using the id rather than the position in a list to link the names and the repeating group?

app/questionnaire/questionnaire_schema.py
```diff
@@ -135,17 +140,25 @@ def get_parent_options_for_block(self, block_id):
...
+
+    def get_group_dependencies_block_drivers(self):
+        return self.group_dependencies['block_drivers']
+
+    def location_requires_group_instance(self, location):
+        return bool(self.group_dependencies.get(location.group_id) or
+                    location.group_id in self.get_group_dependencies_group_drivers() or
+                    location.block_id in self.get_group_dependencies_block_drivers())
+
+    def is_block_relationship_type(self, block_id):
```
jonnyshaw89: Maybe we can make this more generic, eg `is_block_of_type(self, block_id, type)`
otherwise we can end up with lots of functions

app/helpers/schema_helpers.py
```diff
@@ -27,30 +27,42 @@ def wrapped_function(*args, **kwargs):
...
-    return '-'.join([location.group_id, str(uuid4())])
+    return str(uuid4())
+
+
+def _get_dependent_group_instance(schema, dependent_drivers, answer_store, group_instance):
+    group_instance_ids = []
+    for driver_id in dependent_drivers:
+        if driver_id in schema.get_group_dependencies_group_drivers():
+            driver_answer_ids = schema.get_answer_ids_for_group(driver_id)
+            group_instance_ids += _get_group_instance_ids_for_group(answer_store, driver_answer_ids)
```
jonnyshaw89: `group_instance_ids.append(_get_group_instance_ids_for_group(answer_store, driver_answer_ids))`?

jonnyshaw89 (changes requested):

app/questionnaire/questionnaire_schema.py
dwyeradam: This might be a bit misleading. This checks that the question within a block are of type relationship. Not the block['type'] itself.
I just hated this `if location.block_id in ['relationships', 'household-relationships']:` to signify that it's a relationship block.

Maybe change to `has_block_question_type(self, block_id, question_type)`.

jonnyshaw89: `block_has_question_type(self, block_id, question_type)`?

app/helpers/schema_helpers.py
dwyeradam: I need extend but yes, will change that now. Otherwise i'll get a list of lists.

jonnyshaw89:
👍 Seems to work as expected!

bitdivision:
Have tested on all the repeating groups I can find. Not seen any issues :+1:

AndrewTorrance (commented):
Tested on LMS (Social) OK👍

insacuri:
Upgrade on mbs surveys (tried 2) seem to upgrade just fine

jonnyshaw89 (approved):

insacuri (approved):

AndrewTorrance (approved):

bitdivision (approved):

Issue #1731, opened 2018-08-16
started_at wiped from metadata when a questionnaire is re-launched


jonnyshaw89:
### Expected behaviour
When a questionnaire is re-launched the `started_at` date should remain

The `started_at` value should be set when the 1st page is posted (generally introduction)

### Actual behaviour
When a questionnaire is re-launched the metadata is overwritten which wipes the `started_at` date.
Question:  Should our metadata be immutable?

The `started_at` value is set when the 1st answer is sent, rather than when the "Start survey" button is pressed

### Steps to reproduce the behaviour
Launch from launcher and name a note of `RU Ref` and `Collection Exercise SID`
Answers some questions
Launch again from launcher using the same `RU Ref` and `Collection Exercise SID`

bitdivision:
Can we rely on the start survey button always being present?

jonnyshaw89:
@bitdivision no, so the 1st form submit

PR #1732, opened 2018-08-16
Calculate reference dates

bitdivision:
This is currently blocked on #1731 (Which may cause significant changes)

We need to add a few reference date ranges, which all amount to dates which are a constant offset from the previous Sunday (not including today if it is Sunday).

There is additional functionality in the formatting of these date ranges. If the range spans a single month and year, then the first month / year should not be shown. e.g. `Thursday 16 to Friday 17 August 2018` rather than `Thursday 16 August 2018 to Friday 17 August 2018`. I've assumed this change is required going from the examples, but let me know if that's incorrect.

### How to review 

* The test schema contains a few examples as per the examples we were given. If testing manually with go-launcher, remove `started_at` from metadata to default to `utcnow()`.
* Ensure these examples match the expected ranges from trello.

### Notable changes
Look over the following code in particular:

* `started_at` changed from `now(timezone.utc)` to `utcnow()`. Due to discrepancies between these, I think we should standardise `utcnow()`.
* `metadata_parser.py` there is a special case here for validating started_at for the first question when it isn't present. This also defaults to `utcnow()` . See more detailed comment in trello.
* I've added a dependency on `freezegun` in order to pin `started_at` after the first question has been answered in integration tests. I can't see a way to inject any metadata for the second question, but if anyone knows of a way to do this nicely I'd love to remove this.

I'll need to setup signed commits before this is merged.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -24,6 +24,7 @@ httmock = "*"
 "flake8-tuple" = "*"
 moto = "*"
 google-compute-engine = "*"
+freezegun = "*"
 
 [packages]
 colorama = "*"
--- a/app/jinja_filters.py
+++ b/app/jinja_filters.py
@@ -113,6 +113,16 @@ def format_date(context, value):
 
 @evalcontextfilter
 @blueprint.app_template_filter()
+def format_date_custom(context, value, date_format='EEEE d MMMM YYYY'):
+
+    london_date = datetime.strptime(value, '%Y-%m-%d')
+    result = "<span class='date'>{date}</span>".format(date=flask_babel.format_datetime(london_date,
+                                                                                        format=date_format))
+    return mark_safe(context, result)
+
+
+@evalcontextfilter
+@blueprint.app_template_filter()
 def format_datetime(context, value, date_format="d MMMM YYYY 'at' HH:mm"):
 
     london_date_time = datetime.strptime(value, '%Y-%m-%dT%H:%M:%S.%f')
@@ -143,6 +153,45 @@ def format_conditional_date(context, date1=None, date2=None):
 
 
 @blueprint.app_template_filter()
+def calculate_offset_from_weekday_in_last_whole_week(input_datetime, offset, day_of_week='MO'):
+    """
+    Offset a date from a particular day of the week in the previous week.
+
+    Intended to be used to generate reference date ranges around a particular date
+    The offsets will always be based on one day of the week in the previous Mon-Sun
+
+    :param (str) input_datetime: The datetime (or date) to offset. Defaults to now
+    :param (int) offset: The offset dictionary, can contain days, weeks, or years
+    :param (str) day_of_week: The day of the previous week to offset from (two letter abbreviation)
+    :returns (str): The offset date
+    """
+    if input_datetime:
+        parsed_datetime = datetime.strptime(input_datetime.split('T')[0], '%Y-%m-%d')
+    else:
+        parsed_datetime = datetime.utcnow()
+
+    offset_days = offset.get('days', 0)
+    offset_weeks = offset.get('weeks', 0)
+    offset_years = offset.get('years', 0)
+
+    offset = relativedelta.relativedelta(days=offset_days, weeks=offset_weeks, years=offset_years)
+
+    weekdays = ('MO', 'TU', 'WE', 'TH', 'FR', 'SA', 'SU')
+
+    try:
+        day_of_week_index = weekdays.index(day_of_week)
+    except ValueError:
+        raise Exception('Invalid day of week passed to calculate_offset_from_weekday_in_last_whole_week')
+
+    day_of_week_offset = relativedelta.relativedelta(days=(-parsed_datetime.weekday() - (7 - day_of_week_index)))
+    day_of_last_week = parsed_datetime + day_of_week_offset
+
+    offset_output = day_of_last_week + offset
+
+    return datetime.strftime(offset_output, '%Y-%m-%d')
+
+
+@blueprint.app_template_filter()
 def calculate_years_difference(from_str, to_str):
     if from_str is None or to_str is None:
         raise Exception('Valid date(s) not passed to calculate_years_difference filter')
@@ -164,6 +213,53 @@ def format_date_range(context, start_date, end_date=None):
         result = format_date(context, start_date)
 
     return mark_safe(context, result)
+
+
+@evalcontextfunction
+def format_date_range_no_repeated_month_year(context, start_date, end_date, date_format='d MMMM YYYY'):
+    """
+    Format a date range, ensuring months and years are not repeated.
+
+    If the dates are in the same year, the first year (YYYY) will be removed.
+    If the dates are in the same month and year, the first year (YYYY) and month will be removed.
+
+    e.g. Friday 1 to Sunday 3 October
+    or   Thursday 30 September to Sunday 3 October
+
+    Assumptions:
+        - The date format uses space as a seperator
+        - The date format can have leading and trailing whitespace stripped
+
+    :param (jinja2.nodes.EvalContext) context: Evaluation context.
+    :param (str) start_date : The date format that should be used for output. MMMM, YYYY will be removed if necessary
+    :param (str) end_date: The date format that should be used for output. MMMM, YYYY will be removed if necessary
+    :param (str) date_format: The date format that should be used for output. MMMM, YYYY will be removed if necessary
+    :returns (str): The formatted range.
+    """
+    start_datetime = convert_to_datetime(start_date)
+    end_datetime = convert_to_datetime(end_date)
+
+    first_date_format = date_format
+
+    if start_datetime.year == end_datetime.year:
+        first_date_format = date_format.replace('YYYY', '')
+
+        if start_datetime.month == end_datetime.month:
+            first_date_format = first_date_format.replace('MMMM', '')
+
+    # Cleanup any extra spaces in the new format string
+    first_date_format = first_date_format.replace('  ', ' ').strip()
+
+    if not first_date_format:
+        # If the date format was entirely removed, leave it alone
+        first_date_format = date_format
+
+    output = '{} to {}'.format(
+        format_date_custom(context, start_date, first_date_format),
+        format_date_custom(context, end_date, date_format),
+    )
+
+    return mark_safe(context, output)
 
 
 @blueprint.app_template_filter()
--- a/app/submitter/converter.py
+++ b/app/submitter/converter.py
@@ -1,4 +1,4 @@
-from datetime import datetime, timezone
+from datetime import datetime
 from structlog import get_logger
 from app.submitter.convert_payload_0_0_1 import convert_answers_to_payload_0_0_1
 from app.submitter.convert_payload_0_0_2 import convert_answers_to_payload_0_0_2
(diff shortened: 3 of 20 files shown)


Pipfile
```diff
@@ -50,6 +50,7 @@ blinker = "*"
 humanize = "*"
 flask-talisman = "*"
 marshmallow = "*"
+freezegun = "*"
```
jonnyshaw89: shouldn't this be in `[dev-packages]`?

app/jinja_filters.py
```diff
@@ -142,6 +150,43 @@ def format_conditional_date(context, date1=None, date2=None):
...
+    :param (int) offset_days: The number of days to offset this date`
+    :param (str) day_of_week: The day of the previous week to offset from (two letter abbreviation)
+    :returns (str): The offset date
+    """
+    if 'T' in input_datetime:
+        if '.' in input_datetime:
+            parsed_datetime = datetime.strptime(input_datetime, '%Y-%m-%dT%H:%M:%S.%f')
+        else:
+            # Freezegun doesn't support microseconds.
+            parsed_datetime = datetime.strptime(input_datetime, '%Y-%m-%dT%H:%M:%S')
```
jonnyshaw89: is this just here to support tests?

app/submitter/converter.py
```diff
@@ -46,7 +46,7 @@ def convert_answers(metadata, schema, answer_store, routing_path, flushed=False)
       }
     """
     survey_id = schema.json['survey_id']
-    submitted_at = datetime.now(timezone.utc)
+    submitted_at = datetime.utcnow()
```
jonnyshaw89, line 50: 👍

jonnyshaw89 (changes requested):

Pipfile
bitdivision: Yikes. Now sure how that happened.

app/jinja_filters.py
bitdivision: Yes, and it's horrendous. I'd ideally like to get rid of freezegun, but couldn't see a way to inject metadata here.

bitdivision:
Now that #1731 has been sorted out, this should be ready to go in.

app/jinja_filters.py
```diff
@@ -142,6 +152,42 @@ def format_conditional_date(context, date1=None, date2=None):
...
+    :param (int) offset_days: The number of days to offset this date`
+    :param (str) day_of_week: The day of the previous week to offset from (two letter abbreviation)
+    :returns (str): The offset date
+    """
+    if input_datetime:
+        if 'T' in input_datetime:
+            parsed_datetime = datetime.strptime(input_datetime, '%Y-%m-%dT%H:%M:%S.%f')
+        else:
+            parsed_datetime = datetime.strptime(input_datetime, '%Y-%m-%d')
+    else:
```
dwyeradam: Not sure I like this. Possibly use result of `input_datetime.split("T")[0]` instead.
```
if input_datetime:
   parsed_datetime = datetime.strptime(input_datetime.split("T")[0], '%Y-%m-%d')
else:
```

dwyeradam (changes requested):

app/storage/metadata_parser.py
```diff
@@ -92,6 +92,9 @@ def _validate_metadata_is_present(metadata, required_metadata):
         if name == 'trad_as_or_ru_name':
             # either of 'trad_as' or 'ru_name' is required
             valid = bool(metadata.get('trad_as') or metadata.get('ru_name'))
+        elif name == 'started_at':
+            # started_at will not normally be passed to us, so we can't really validate immediately.
+            valid = True
```
bitdivision: I previously had a line here:
```	
            if not metadata.get('started_at'): 
                metadata['started_at'] = datetime.utcnow().isoformat()
```
I don't think this is needed anymore, but please correct me if you think it is.

app/jinja_filters.py
```diff
@@ -142,6 +152,39 @@ def format_conditional_date(context, date1=None, date2=None):
     return format_date(context, date)
 
 
+@blueprint.app_template_filter()
+def calculate_offset_date(input_datetime, offset_days, day_of_week='SU'):
```
dwyeradam: I've changed my mind. As initially discussed I think this should accept a dictionary rather than only support days.
So along the lines of:
```
calculate_offset_date(input_datetime, offset, day_of_week='SU'):

   offset_days = offset.get('days', 0)
   offset_weeks = offset.get('weeks', 0)
   offset_years = offset.get('years', 0)
```

The filter would then look like this:
`calculate_offset_date(collection_metadata['started_at'], {'weeks'-4, 'days':1})`

dwyeradam (changes requested):

dwyeradam (changes requested):
We also need a change in https://github.com/ONSdigital/go-launch-a-survey
At the moment if someone launches this with `"started_at"` in the metadata (only a problem with our launcher) it will 500. This should default to either blank or a useable date format.

app/jinja_filters.py
bitdivision: Fixed now.

bitdivision:
Good point about `started_at` in the launcher. We could actually remove it from metadata in that schema since it is now in collection_metadata, but I think it is useful for the purposes of testing.

Will make a PR for go-launcher to sort this out @dwyeradam

bitdivision:
I've changed the default day of week to Monday, and updated the test schema to match as per @dwyeradam's suggestion. By using the Monday most of the time, and changing to Sunday where necessary, we avoid using strange numbers in the offset.

bitdivision:
One more change has gone in to remove the handling of in-flight `started_at` in metadata. 

This is because the `started_at` in production currently is defective, and defaulting to now is not going to make any real difference. Additionally, the questionnaires that use `started_at` are not currently in production.

Since `started_at` has been removed from the schema, this PR becomes more difficult to test manually, and the unit / integration tests should be testing all edge cases.

bitdivision:
@AndrewTorrance I've just double checked that babel does format days of the week in welsh:
```
>>> format_datetime(d, locale='cy', format='full')
'Dydd Iau, 23 Awst 2018 am 13:42:12 Amser Cyffredniol Cydlynol'
>>> format_datetime(d, locale='en', format='full')
'Thursday, August 23, 2018 at 1:42:12 PM Coordinated Universal Time'
```
I assume that's correct.

I can try to add a test for this if you think its needed, but I'm not sure it's really worth it?

bitdivision:
Need to make a change to schema validator to fix travis.

tests/integration/mci/test_date_reference_ranges.py
```diff
@@ -0,0 +1,38 @@
+from freezegun import freeze_time
+from tests.integration.integration_test_case import IntegrationTestCase
+
+@freeze_time('2018-10-08')
```
dwyeradam, line 4: I've never used this but I will in future 👍.

dwyeradam (changes requested):
Functional Tests?

app/jinja_filters.py
```diff
@@ -142,6 +152,45 @@ def format_conditional_date(context, date1=None, date2=None):
     return format_date(context, date)
 
 
+@blueprint.app_template_filter()
+def calculate_offset_date(input_datetime, offset, day_of_week='MO'):
```
dwyeradam: This is really hard to name but I think we may need to be clearer about the fact that this offset starts from a specified day in the last whole week. We'll need to buy a hat to put suggestions in.

bitdivision:
Functional tests present issues in terms of mocking started_at now that we've removed it from metadata.

I'll have a think about potential solutions, but I'd really rather avoid reproducing the offset code in the tests.

app/jinja_filters.py
bitdivision: `calculate_date_offset_from_last_weekday` - I think that works?

bitdivision:
The only obvious way I can see to mock collection_metadata for functional tests is to pick it up from the jwt if it's available.

This would still be adding code that's only really used in the tests though. @jonnyshaw89 @dwyeradam any thoughts?

app/helpers/template_helper.py
```diff
@@ -34,6 +34,17 @@ def session_wrapper(*args, **kwargs):
     return session_wrapper
 
 
+def with_collection_metadata_context(func):
+    @wraps(wraps)
+    def collection_metadata_context_wrapper(*args, **kwargs):
```
AndrewTorrance: Looks to be duplicating existing decorator ?

AndrewTorrance (commented):
Looks good , one question regarding the new decorator. Not putting it in as a blocker as I will not be here to take it off next week.

bitdivision:
Absolutely right about the decorator. I've removed it entirely now.

bitdivision:
@jonnyshaw89 @dwyeradam I'm fairly confident that this is finally ready to go in :)

tests/functional/spec/date_reference_ranges.spec.js
```diff
@@ -0,0 +1,41 @@
...
+  // We have no way to mock out started_at, the unit and integration tests cover the date values
+  // this test is only to validate the general questionnaire functionality and to make sure
+  // the date format is correct.
+  it('Given we start the survey, when we click the button, the dates should be in the correct format...', function() {
+    return helpers.startQuestionnaire('test_date_reference_ranges.json').then(() => {
+        return browser
+          // Should show a date range in the right format
+
+          // Matches `in the week Friday 24 August 2018 to`. The day of the month can be 1 or two characters
+          .getText(SummaryPage.questionText()).should.eventually.match(new RegExp('in the week [a-zA-Z]+ \\d(?:\\d)? [a-zA-Z]+ \\d{4} to'))
```
jonnyshaw89, line 18: 👍

data/en/test_date_reference_ranges.json
```diff
@@ -0,0 +1,106 @@
...
+    }, {
+        "name": "ru_name",
+        "validator": "string"
+    }],
+    "sections": [{
+        "id": "default-section",
+        "groups": [{
+            "blocks": [{
+                "type": "Introduction",
+                "id": "introduction",
```
jonnyshaw89: I would probably drop the introduction from this test schema?

jonnyshaw89 (approved):
Looks good. I would just drop the intro block from the test schemas as I cant see it adds anything.

tests/functional/spec/date_reference_ranges.spec.js
```diff
@@ -0,0 +1,41 @@
...
+const DateSeparateBlockPage = require('../pages/features/date_reference_ranges/date-separate-block.page.js');
+const DateRangeBlockPage = require('../pages/features/date_reference_ranges/date-range-block.page.js');
+const SummaryPage = require('../pages/features/date_reference_ranges/summary.page.js');
+
+describe('Reference Date Range Checks', function() {
+  // We have no way to mock out started_at, the unit and integration tests cover the date values
+  // this test is only to validate the general questionnaire functionality and to make sure
+  // the date format is correct.
+  it('Given we start the survey, when we click the button, the dates should be in the correct format...', function() {
+    return helpers.startQuestionnaire('test_date_reference_ranges.json').then(() => {
```
jonnyshaw89: I think you need to change this to `openQuestionnaire` as it doesn't have intro.

bitdivision: I did run the tests after the change, and it did work. I'll switch it over though.

bitdivision: Apparently I didn't :)

dwyeradam (approved):

PR #1736, opened 2018-08-17
Use labels instead of questions within the total playback summary type

insacuri:
https://trello.com/c/smEWR8IV/2337-use-labels-instead-of-questions-within-the-total-playback-summary-type

### How to review 
Fill out the `test_calculated_summary` survey.  In the calculated summary pages (not the final summary), the single answers should end with 'label' where they used to end with title.

The final summary screen should be unaffected by this change.

### Checklist

### DIFF
--- a/app/templates/partials/summary/display-answer-label.html
+++ b/app/templates/partials/summary/display-answer-label.html
@@ -1,0 +1,1 @@
+<div class="summary__label mars" id="{{ answer.id }}-label" data-qa="{{ answer.id }}-label">{{ answer.label|striptags }}</div>
--- a/data/en/test_calculated_summary.json
+++ b/data/en/test_calculated_summary.json
@@ -59,7 +59,7 @@
                         "type": "General",
                         "answers": [{
                                 "id": "second-number-answer",
-                                "label": "Second answer in currency total",
+                                "label": "Second answer in currency label",
                                 "mandatory": true,
                                 "type": "Currency",
                                 "currency": "GBP",
--- a/tests/functional/generate_pages.py
+++ b/tests/functional/generate_pages.py
@@ -116,6 +116,10 @@ SUMMARY_TITLE_GETTER = Template(r"""  ${group_id_camel}Title() { return '#${grou
 
 """)
 
+CALCULATED_SUMMARY_LABEL_GETTER = Template(r"""  ${answerName}Label() { return '#${answerId}-label'; } 
+
+""")
+
 CONSTRUCTOR = Template(r"""  constructor() {
     super('${block_id}');
   }
@@ -231,6 +235,17 @@ def process_question(question, page_spec, num_questions, page_name):
     for answer in question['answers']:
         process_answer(question_type, answer, page_spec, long_names, page_name)
 
+def process_calculated_summary(answers, page_spec):
+    for answer in answers:
+        answer_name = generate_pascal_case_from_id(answer)
+        answer_context = {
+            'answerName': camel_case(answer_name),
+            'answerId': answer
+        }
+
+        page_spec.write(SUMMARY_ANSWER_GETTER.substitute(answer_context))
+        page_spec.write(SUMMARY_ANSWER_EDIT_GETTER.substitute(answer_context))
+        page_spec.write(CALCULATED_SUMMARY_LABEL_GETTER.substitute(answer_context))
 
 def process_summary(schema_data, page_spec):
     for section in schema_data['sections']:
@@ -317,6 +332,11 @@ def process_block(block, dir_out, schema_data, spec_file, relative_require='..')
         base_page = 'QuestionPage'
         base_page_file = 'question.page'
 
+        if block['type'] == 'CalculatedSummary':
+            base_page = 'CalculatedSummaryPage'
+            base_page_file = 'calculated-summary.page'
+            relative_require = relative_require.replace('/surveys', '')
+
         block_context = {
             'pageName': page_name,
             'basePage': base_page,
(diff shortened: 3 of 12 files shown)


data/en/test_calculated_summary.json
```diff
@@ -59,22 +59,22 @@
...
+                                "label": "Second answer in currency label",
                                 "mandatory": true,
                                 "type": "Currency",
                                 "currency": "GBP",
                                 "decimal_places": 2
                             },
                             {
                                 "id": "second-number-answer-unit-total",
-                                "label": "Second answer label in unit total",
+                                "label": "Second answer label in unit label",
```
dwyeradam: Petty but the label wording is now confusing. It was probably clearer before. This is the label of the second answer used in the Unit Total. Few confusing labels like this.

tests/integration/views/test_questionnaire.py
```diff
@@ -538,6 +538,66 @@ def test_build_view_context_for_question(self):
         # Then
         self.assertEqual(question_view_context['question_titles']['single-title-question'], 'How are you feeling??')
 
+    def test_calculated_summary_answer_label(self):
```
dwyeradam: I would have expected a change to the functional test for this, not an integration test. This is a html change of what the user sees not the context that is passed. So essentially this test in `calculated_summary.spec.js`

dwyeradam (changes requested):
A couple of small things.

tests/integration/views/test_questionnaire.py
insacuri: I tried, but the label field doesn't have an id (and thus no webdriver object).  So I thought an integration test was the only way to do it?
From what you said though, I'm going to assume there is.  Can you let me know what that is exactly?

data/en/test_calculated_summary.json
insacuri: That's a point fair.  It's got the word label in the title so I don't need to change it again at the end

tests/integration/views/test_questionnaire.py
dwyeradam: @insacuri you're quite right there isn't but nothing stopping you adding one to `display-answer-label.html`

dwyeradam: Or the very catchy selector of
`#main > div.summary > div > div:nth-child(2) > div:nth-child(1) > div.grid__col.col-12\40 xs.col-6\40 m > div` but I wouldn't recommend using that.

data/en/test_calculated_summary.json
```diff
@@ -59,7 +59,7 @@
                         "type": "General",
                         "answers": [{
                                 "id": "second-number-answer",
-                                "label": "Second answer in currency total",
+                                "label": "Second answer in currency label",
```
jonnyshaw89, line 62: I don't understand why this is changed?

app/templates/partials/summary/display-answer-label.html
```diff
@@ -0,0 +1 @@
+<div class="summary__label mars" id="{{ answer.id }}-label" data-qa="{{ answer.id }}-label">{{ answer.label|striptags }}</div>
```
dwyeradam, line 1: Do we need a `data-qa` and `id` when they are both the same? One or the other but both seems overkill.

dwyeradam (changes requested):

app/templates/partials/summary/display-answer-label.html
insacuri: I was copying display-answer.html.  Is that one wrong also?  If so then I can make a trello card to remove one or the other if you want.

data/en/test_calculated_summary.json
insacuri: To make it super obvious that it's the label of the field.  The whole point of my changes was to make sure it was the label of the question, rather then the title that was being displayed on the calculated summary screen.  
This way I can make it explicit that this bit of text is definitely the label and test for it.
That seem reasonable?

jonnyshaw89: but "Second answer in currency total" was a unique value on that page? It doesn't really matter, i just couldn't see the reason behind it.

app/templates/partials/summary/display-answer-label.html
dwyeradam: Keep it then if used elsewhere. Seems excessive rather than wrong.

bitdivision:
I've tested and looked through the changes. Can't see any issues.

dwyeradam (dismissed):

tests/functional/pages/features/calculated_summary/currency-total-playback.page.js
```diff
@@ -11,46 +11,68 @@ class CurrencyTotalPlaybackPage extends QuestionPage {
 
   firstNumberAnswerEdit() { return '[data-qa="first-number-answer-edit"]'; }
 
+  firstNumberAnswerLabel() { return '[data-qa="first-number-answer-label"]'; } // Field not generated with generated with generate_pages.py script
```
jonnyshaw89: I feel this is a slippery path to head down.
These files are generated, as soon as we start making changes that are not generated then we are opening ourselves up for a world of pain?

jonnyshaw89 (dismissed):

tests/functional/generate_pages.py
```diff
@@ -317,6 +332,11 @@ def process_block(block, dir_out, schema_data, spec_file, relative_require='..')
         base_page = 'QuestionPage'
         base_page_file = 'question.page'
 
+        if block['type'] == 'CalculatedSummary':
+            base_page = 'CalculatedSummaryPage'
+            base_page_file = 'calculated-summary.page'
+            relative_require = relative_require.replace('/surveys', '')
```
dwyeradam, line 338: 👍

dwyeradam (approved):
That's what I call team work.

bitdivision (approved):

jonnyshaw89:
@insacuri Squash?

PR #1737, opened 2018-08-17
Make `case_id` Mandatory

jonnyshaw89:
`case_id` is mandatory downstream so we should make it mandatory upstream also

### How to review 
Launch with and without `case_id` to test that it validates the mandatoriness of the claim

### DIFF
--- a/tests/integration/create_token.py
+++ b/tests/integration/create_token.py
@@ -42,6 +42,7 @@ class TokenGenerator:
         payload_vars['iat'] = time.time()
         payload_vars['exp'] = payload_vars['iat'] + float(3600)  # one hour from now
         payload_vars['jti'] = str(uuid4())
+        payload_vars['case_id'] = str(uuid4())
 
         for key, value in extra_payload.items():
             payload_vars[key] = value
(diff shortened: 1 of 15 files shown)


bitdivision (approved):
Looks good.

tests/integration/create_token.py
```diff
@@ -42,6 +42,7 @@ def _get_payload_with_params(form_type_id, eq_id, survey_url=None, **extra_paylo
         payload_vars['iat'] = time.time()
         payload_vars['exp'] = payload_vars['iat'] + float(3600)  # one hour from now
         payload_vars['jti'] = str(uuid4())
+        payload_vars['case_id'] = str(uuid4())
```
dcdarrell9, line 45: The `collection_exercise_sid` is in the PAYLOAD above, should it be? Or should the case_id be there? Okus could maybe have a test similar to the jti test, but with case_id instead

```
    def test_login_without_case_id_in_token_is_unauthorised(self):
        # Given
        token = self.token_generator.create_token_without_case_id('0205', '1')
        self.get('/session?token=' + token)

        # Then
        self.assertStatusForbidden()

    def create_token_without_case_id(self, form_type_id, eq_id, **extra_payload):
        payload_vars = self._get_payload_with_params(form_type_id, eq_id, None, **extra_payload)
        del payload_vars['case_id']

        return self.generate_token(payload_vars)

```

jonnyshaw89: If we put `case_id` in the `PAYLOAD` var then it would be the same for every launch.
I can add those tests though

dcdarrell9: Thinking on that, should `collection_exercise` be different every time?

jonnyshaw89: 🤷‍♂️ i don't know how much it matters

dcdarrell9 (approved):

PR #1738, opened 2018-08-17
Add collection_metadata, make metadata immutable

### LINKED ISSUE #1731: started_at wiped from metadata when a questionnaire is re-launched

bitdivision:
[Trello](https://trello.com/c/ET4X7qU4/2343-eq-survey-runner-1731-startedat-wiped-from-metadata-when-a-questionnaire-is-re-launched-issue)

Closes #1731 

Metadata is passed in and shouldn't be changed. It should be immutable.

We have metadata around the collection of the questionnaire which can't be immutable and doesn't fit in with the rest of `metadata` (started_at). Store it in `collection_metadata`.

`MappingProxyType` is used to make `metadata` a read-only view of `_metadata`. 

### How to review 
- Launch a questionnaire taking note of `RU Ref` and `Collection Exercise SID`
- Tap start survey and ensure `started_at` has been set.
- Launch with the same `RU Ref` and `Collection Exercise SID` and ensure `started_at` doesn't change.

### Checklist

### DIFF
--- a/app/submitter/converter.py
+++ b/app/submitter/converter.py
@@ -15,10 +15,11 @@ class DataVersionError(Exception):
         return 'Data version {} not supported'.format(self.version)
 
 
-def convert_answers(metadata, schema, answer_store, routing_path, flushed=False):
+def convert_answers(metadata, collection_metadata, schema, answer_store, routing_path, flushed=False):
     """
     Create the JSON answer format for down stream processing
     :param metadata: metadata for the questionnaire
+    :param collection_metadata: Any metadata about the collection of the questionnaire
     :param schema: QuestionnaireSchema class with populated schema json
     :param answer_store: the users answers
     :param routing_path: the path followed by the user when answering the questionnaire
@@ -59,9 +60,13 @@ def convert_answers(metadata, schema, answer_store, routing_path, flushed=False)
         'metadata': _build_metadata(metadata),
         'case_id': metadata['case_id'],
     }
-    if 'started_at' in metadata and metadata['started_at']:
+    if collection_metadata.get('started_at'):
+        payload['started_at'] = collection_metadata['started_at']
+    elif metadata.get('started_at'):
         payload['started_at'] = metadata['started_at']
-    if 'case_ref' in metadata and metadata['case_ref']:
+    if metadata.get('case_id'):
+        payload['case_id'] = metadata['case_id']
+    if metadata.get('case_ref'):
         payload['case_ref'] = metadata['case_ref']
 
     if schema.json['data_version'] == '0.0.2':
--- a/app/views/questionnaire.py
+++ b/app/views/questionnaire.py
@@ -1,7 +1,7 @@
 from functools import wraps
 import re
 from collections import defaultdict
-from datetime import datetime, timedelta, timezone
+from datetime import datetime, timedelta
 import humanize
 import simplejson as json
 from dateutil.tz import tzutc
@@ -16,11 +16,12 @@ from structlog import get_logger
 from app.globals import get_session_store, get_completeness
 from app.data_model.answer_store import Answer, AnswerStore
 from app.data_model.app_models import SubmittedResponse
-from app.globals import get_answer_store, get_completed_blocks, get_metadata, get_questionnaire_store
+from app.globals import (get_answer_store, get_completed_blocks, get_metadata, get_questionnaire_store,
+                         get_collection_metadata)
 from app.helpers.form_helper import post_form_for_location
 from app.helpers.path_finder_helper import path_finder, full_routing_path_required
 from app.helpers.schema_helpers import with_schema
-from app.helpers.session_helpers import with_answer_store, with_metadata
+from app.helpers.session_helpers import with_answer_store, with_metadata, with_collection_metadata
 from app.helpers.template_helper import (with_session_timeout, with_metadata_context, with_analytics,
                                          with_questionnaire_url_prefix, with_legal_basis, render_template)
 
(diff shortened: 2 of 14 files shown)


app/questionnaire/completeness.py
```diff
@@ -15,13 +15,16 @@ class Completeness:
     # one item is complete
     COMPLETED_STATES = (COMPLETED, INVALID, SKIPPED)
 
-    def __init__(self, schema, answer_store, completed_blocks, routing_path, metadata):
+    def __init__(self, schema, answer_store, completed_blocks, routing_path, metadata, collection_metadata=None):
```
jonnyshaw89: Why are we passing this into here if we don't use it?

app/submitter/converter.py
```diff
@@ -58,11 +59,11 @@ def convert_answers(metadata, schema, answer_store, routing_path, flushed=False)
         'collection': _build_collection(metadata),
         'metadata': _build_metadata(metadata),
     }
-    if 'started_at' in metadata and metadata['started_at']:
-        payload['started_at'] = metadata['started_at']
-    if 'case_id' in metadata and metadata['case_id']:
+    if collection_metadata.get('started_at'):
+        payload['started_at'] = collection_metadata['started_at']
+    if metadata.get('case_id'):
```
jonnyshaw89: 👍

jonnyshaw89 (changes requested):
Looks good, have you tested launching before this change then switching to this code in flight?
to mimic a deployment with in-flight users?
also we need to test having this new code and then not. as when deploying both versions of the app are "in" at the same time. therefore someone could have `COLLECTION_METADATA` set on the old code?

app/questionnaire/completeness.py
```diff
@@ -22,6 +22,9 @@ def __init__(self, schema, answer_store, completed_blocks, routing_path, metadat
         self.metadata = metadata
         self.schema = schema
 
+        if collection_metadata is None:
+            collection_metadata = {}
+
```
jonnyshaw89: psst :point_up:

bitdivision: The shame...

bitdivision:
Very good points about in-flight data.

I've added some logic to pull in the `started_at` time from metadata if it exists. This will need to be removed after deployment. I'll raise a ticket to do that. I've tested that this works by starting a survey on master, switching to this branch, and checking that the payload includes the original started at date.

In terms of starting a questionnaire on this branch and moving back to master, `started_at` also would have been incorrect, since it `metadata` would be used. To avoid this, I've added a couple of nasty bits of code to write started_at to `metadata` as well as `collection_metadata` for the moment. This should all be reverted after deployment.

I've tested moving back to master on the latest version and `started_at` remained set to the original value.

If anyone can see a nicer method of handling in-flight data I'd be all ears.

@jonnyshaw89 Could you take another look?

jonnyshaw89:
@bitdivision looking good, just the codecov now.

bitdivision:
Hopefully fixed now. I'll raise a ticket to remove the following code after this is deployed:

- tests/integration/introduction/test_introduction: test_started_at_retrieved_f
rom_metadata_if_available()

- tests/app/submitter/test_converter.py: test_started_at_comes_from_metadata_if
_necessary()

- app/views/questionnaire.py: _set_started_at_metadata_if_required()
Remove indicated lines.

- app/data/model_questionnair_store.py: QuestionnaireStore.set_metadata_key
This method can be removed. It is used to set data on the readonly metadata sto
re.

I suppose the other thing to think about is how long we need to wait after deployment before we can be sure collection_metadata is the proper source of truth.

jonnyshaw89 (approved):

tests/integration/introduction/test_introduction.py
```diff
@@ -42,3 +43,33 @@ def test_intro_basis_for_completion_not_displayed(self):
...
+        # Given survey with a start survey button
+        self.launchSurvey('1', '0112', roles=['dumper'])
+
+        self.post(action='start_questionnaire')
+
+        # When start survey button is pressed,
+        # Then started_at should be set in collection_metadata (and payload)
+        actual = self.dumpSubmission()['submission']
+
+        self.assertIsNotNone(actual['started_at'])
```
AndrewTorrance: Shouldn't this check if the value of started at is a valid date format , it could be an empty string ?

AndrewTorrance (changes requested):
I think there are a few more test cases possibly needed ? Launch a survey twice . We should retain the start date of the first . (not sure if we need variants on this with start date passed in / calculated ?) . Also one test is only testing that start date exists , not that it is a valid date

tests/integration/introduction/test_introduction.py
bitdivision: Yep, I'll sort that out.

insacuri (approved):
Tested it, looks good

app/views/questionnaire.py
```diff
@@ -330,13 +332,25 @@ def get_view_submission(schema, eq_id, form_type):  # pylint: disable=unused-arg
...
-    questionnaire_store = get_questionnaire_store(current_user.user_id, current_user.user_ik)
-    if not questionnaire_store.answer_store.answers and len(form.data) > 1:
-        started_at = datetime.now(timezone.utc).isoformat()
-        logger.info('first answer about to be stored. writing started_at time to metadata',
+def _set_started_at_metadata_if_required(form, collection_metadata, metadata):
+    if not collection_metadata.get('started_at') and form.data:
+        started_at = datetime.utcnow().isoformat()
+
+        # To avoid loss of in-flight data while collection_metadata is deployed.
+        # After deployment of collection metdata, remove from here to the next comment.
```
insacuri, line 340: Loving this.  Decent instructions on how to remove code after it's not useful anymore is welcome to see!

bitdivision:
@AndrewTorrance Could you take a look at the test in test_multiple_login? I think that resolves your point.

If this is okay, I'll squash and sign.

AndrewTorrance (approved):

bitdivision:
@jonnyshaw89 @dwyeradam Any objections to my merging this?

jonnyshaw89:
@bitdivision yeh, don't merge your own.

PR #1739, opened 2018-08-20
Answers not on routing path excluded from CalculatedSummary

dwyeradam:
The summary context already excludes blocks not on the path.
This uses the values in that context to create the total instead.

Skipped answers were still be added to the total of the calculated summary. The answers themselves were however not shown on the page. 

### How to review 
The option to skip the fourth block is now part of the calculated summary test schema.

### DIFF
--- a/app/templating/view_context.py
+++ b/app/templating/view_context.py
@@ -4,7 +4,6 @@ from app.helpers.form_helper import get_form_for_location
 from app.templating.summary_context import build_summary_rendering_context
 from app.templating.template_renderer import renderer
 from app.templating.utils import get_title_from_titles, get_question_title
-from app.questionnaire.rules import get_answer_store_value
 
 
 def build_view_context(block_type, metadata, schema, answer_store, schema_context, rendered_block, current_location,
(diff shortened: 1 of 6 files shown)


dwyeradam:
This [PR](https://github.com/ONSdigital/eq-survey-runner/pull/1736) should go in first as will be needed for QCAS. There will likely be some merge conflicts between the two.

AndrewTorrance (changes requested):
Functionally fine , just one minor code styling comment

bitdivision:
@dwyeradam Was just testing the functionality here. 

Noticed that if you initially skip the fourth block, then fill in fifth and sixth, go back to fourth, don't skip and fill answers, it goes straight to the summary and doesn't move through the fifth and sixth blocks again.

I'm unsure if this is expected or not, but thought it was worth raising.

app/templating/view_context.py
```diff
@@ -201,25 +200,32 @@ def _remove_unwanted_questions_answers(block, answers_in_block, answer_ids_to_ke
...
+        for block in group['blocks']:
+            for question in block['questions']:
+                for answer in question['answers']:
+                    if not answer_format:
+                        answer_format = {
+                            'type': answer['type'],
+                            'unit': answer.get('unit'),
+                            'currency': answer.get('currency'),
+                        }
+                    answer_value = answer.get('value') or 0
```
AndrewTorrance, line 216: Perhaps answer.get('value', 0 ) is easier to read ?

AndrewTorrance (changes requested):

app/templating/view_context.py
dwyeradam: That's what I originally had but doesn't work if answer['value'] == None. This way of doing it not only ensure the key is there but also that a value is assigned against that key.

dwyeradam:
@bitdivision. This is a runner feature. Once you get to the summary page as long as it stays valid any change will return you to that summary page. We have an answer dependency class that will uncomplete blocks if an answer changes. At the moment routing doesn't affect that though.

bitdivision:
Right, that makes sense. Thanks for the explanation.

AndrewTorrance (approved):

bitdivision (approved):

insacuri (approved):

PR #1741, opened 2018-08-21
Allow for no answers being available for CalculatedSummary

dwyeradam:
There could be a situation where every answer in a CalculatedSummary is skipped or routed past. This would currently cause a 500 error.

### How to review 
In `test_calculated_summary.json` for the `currency-total-playback` block delete all but the fourth numbers so that the answers to calculate looks like this
```
"answers_to_calculate": [
   "fourth-number-answer",
   "fourth-number-answer-also-in-total"
],
```
The when asked to skip that block do so. When you get to the CalculatedSummary the total should be zero and no answers are displayed.

### DIFF
--- a/app/templating/view_context.py
+++ b/app/templating/view_context.py
@@ -202,12 +202,12 @@ def _remove_unwanted_questions_answers(block, answers_in_block, answer_ids_to_ke
 
 def _get_formatted_total(groups):
     calculated_total = 0
-    answer_format = None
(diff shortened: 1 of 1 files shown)


bitdivision (approved):
Reproduced the issue and confirmed that this branch fixes the problem.

MebinAbraham (approved):
Tested and working as expected. 👍

PR #1746, opened 2018-08-23
LMS 2

AndrewTorrance:
Adding the business section to lms as a new survey lms_2 

### How to review 
Currently difficult to review. No clear complete documentation available. There is a hardcopy visio document , and a pdf of it on the ticket. There is also a 'final outline spec' which is useful to understand content.
( I have also added notes on the trello ticket. to help explain)  

The easiest way is to go through the documented paths . Note that even the visio printout does not contain anything regarding proxy questions.  And also very few of the answers are mandatory.

known issues:- Date times need reformatting , after adding ability to route based on answer value comparison then routing to change. Totaliser use of labels has meant changing answer labels. 
We are going to add extra changes through additional PRS.

### DIFF
--- a/data/en/lms_2.json
+++ b/data/en/lms_2.json
@@ -1,0 +1,4970 @@
+{
+    "mime_type": "application/json/ons/eq",
+    "schema_version": "0.0.1",
+    "data_version": "0.0.2",
+    "survey_id": "lms",
+    "title": "Online Household Study (Alpha)",
+    "description": "Online Household Study Schema (Alpha)",
+    "theme": "default",
+    "legal_basis": "Voluntary",
+    "eq_id": "lms",
+    "form_type": "1",
+    "navigation": {
+        "visible": true
+    },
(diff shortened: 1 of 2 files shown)


bitdivision:
@AndrewTorrance Did you get a chance to make the changes that came up in planning yesterday?

bitdivision:
Also, I'm wondering whether I should be suggesting changes like validating the `Confirm email` question, or adding max validation on number of hours per week to 7*24?

insacuri:
Is the first page meant to be a 'save and continue' button page? When you launch the survey, are you meant to be in the middle of something?

insacuri:
Does there need to be a 'display_address' field in the metadata like in lms_1?  I fired up the survey with default data and I see a 'Your address: ' which looks odd

jonnyshaw89 (approved):

dwyeradam (approved):

PR #1751, opened 2018-08-29
Address playback

dcdarrell9:
- Playback address from metadata on separate lines instead of on one.

### How to review 
- Check that address comes gets displayed as it should.
- Not sure how the address should display in a summary, so possibly need some guidance on that

### Checklist

### DIFF
--- a/app/jinja_filters.py
+++ b/app/jinja_filters.py
@@ -6,7 +6,7 @@ from datetime import datetime
 import flask
 import flask_babel
 from dateutil import relativedelta, tz
-from jinja2 import Markup, contextfunction, escape, evalcontextfilter, evalcontextfunction
+from jinja2 import Markup, contextfunction, escape, evalcontextfilter, evalcontextfunction, Undefined
 from jinja2.exceptions import UndefinedError
 
 from babel import units, numbers
@@ -56,6 +56,30 @@ def format_currency_for_input(value, decimal_places=0):
 @blueprint.app_template_filter()
 def format_percentage(value):
     return '{}%'.format(value)
+
+
+@blueprint.app_template_filter()
+def format_address_list(user_entered_address=None, metadata_address=None):
+    """
+        This function format_address_list accepts two addresses, a user submitted address and a metadata address
+        If all the items in address list are 'Undefined' (nothing in answer_store) then we retrieve the metadata.
+
+        :param user_entered_address user entered address which is a list of answer fields
+        :param metadata_address metadata address which is a list of answer fields passed to us
+        :return: the value of the address to be piped
+    """
+    if all(isinstance(field, Undefined) for field in user_entered_address) or \
+       all(field == '' for field in user_entered_address):
+        address = metadata_address
+    else:
+        address = user_entered_address
+
+    address_list = concatenated_list(list_items=address, delimiter='<br />')
+
+    if not address_list:
+        raise Exception('No valid address passed to format_address_list filter')
+
+    return address_list
 
 
 def format_unit(unit, value=''):
--- a/tests/app/test_jinja_filters.py
+++ b/tests/app/test_jinja_filters.py
@@ -18,7 +18,7 @@ from app.jinja_filters import (
     calculate_years_difference, get_current_date, as_london_tz, max_value,
     min_value, get_question_title, get_answer_label,
     format_duration, calculate_offset_from_weekday_in_last_whole_week, format_date_custom,
-    format_date_range_no_repeated_month_year, format_repeating_summary)
+    format_date_range_no_repeated_month_year, format_repeating_summary, format_address_list)
 from tests.app.app_context_test_case import AppContextTestCase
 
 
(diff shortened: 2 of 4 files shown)


app/jinja_filters.py
```diff
@@ -52,6 +52,29 @@ def format_percentage(value):
     return '{}%'.format(value)
 
 
+@blueprint.app_template_filter()
+def format_address_list(address1=None, address2=None):
+    """
+        This function format_address_list accepts two address', a user submitted address and a metadata address
```
insacuri: Addresses instead of address'

app/jinja_filters.py
```diff
@@ -52,6 +52,29 @@ def format_percentage(value):
     return '{}%'.format(value)
 
 
+@blueprint.app_template_filter()
+def format_address_list(address1=None, address2=None):
```
insacuri: I think instead of address1 and address2, it should be something like `user_entered_address` and `metadata_address` maybe?  Just something more descriptive.

app/jinja_filters.py
```diff
@@ -52,6 +52,29 @@ def format_percentage(value):
...
+    """
+        This function format_address_list accepts two address', a user submitted address and a metadata address
+        If all the items in address list are 'Undefined' (nothing in answer_store) then we retrieve the metadata.
+
+        :param address1 user entered date
```
insacuri: This isn't a date is it?  Is this a copy/pasta error? ^^
(Same comment with the line below)

insacuri (changes requested):

app/jinja_filters.py
```diff
@@ -52,6 +52,29 @@ def format_percentage(value):
     return '{}%'.format(value)
 
 
+@blueprint.app_template_filter()
+def format_address_list(user_entered_address=None, metadata_address=None):
+    """
+        This function format_address_list accepts two addresses, a user submitted address and a metadata address
+        If all the items in address list are 'Undefined' (nothing in answer_store) then we retrieve the metadata.
+
+        :param user_entered_address user entered date
```
insacuri: Still says date

tests/app/test_jinja_filters.py
```diff
@@ -859,3 +859,27 @@ def test_format_repeated_summaries_delimiters(self):
...
+
+    def test_format_address_list_no_metadata(self):
+        address1 = ['44', 'Testing', 'Gardens', 'Swansea', 'SA1 1AA']
+        address2 = []
+        self.assertEqual('44<br />Testing<br />Gardens<br />Swansea<br />SA1 1AA',
+                         format_address_list(address1, address2))
+
+    def test_format_address_list(self):
+        address1 = ['44', 'Testing', 'Gardens', 'Swansea', 'SA1 1AA']
+        address2 = ['123', 'Testy', 'Place', 'Newport', 'NP5 7AR']
```
jonnyshaw89: Can you create a test where you have missing values from both that don't match?
eg 
```
address1 = ['44', None, 'Gardens', None, 'SA1 1AA']
address2 = [None, 'Testy', None, 'Newport', None]
```

jonnyshaw89 (changes requested):

tests/app/test_jinja_filters.py
jonnyshaw89: Or would the `None`s be `''`?

app/jinja_filters.py
dcdarrell9: I'm so sure I changed that 🤣

tests/app/test_jinja_filters.py
dcdarrell9: Both come back as `''` unless all the answers the answers were `None` or `''` then an exception would be raised

dcdarrell9:
If you leave the answers blank and submit then an exception is raised, we need to know whether we have an `at least one` of a certain amount of answers, so this probably needs some other type of testing/analysis/research or something along those lines to work out what the expected functionality is.

app/jinja_filters.py
```diff
@@ -52,6 +52,29 @@ def format_percentage(value):
...
+    """
+    if all(isinstance(a, Undefined) for a in user_entered_address):
+        address = metadata_address
+    else:
+        address = user_entered_address
+
+    address_list = Markup('<br />').join([i for i in address if i])
+
+    if not address_list:
+        raise Exception('No valid address passed to format_address_list filter')
```
jonnyshaw89: I wonder if exception it a bit harsh?

app/jinja_filters.py
```diff
@@ -52,6 +52,29 @@ def format_percentage(value):
...
+def format_address_list(user_entered_address=None, metadata_address=None):
+    """
+        This function format_address_list accepts two addresses, a user submitted address and a metadata address
+        If all the items in address list are 'Undefined' (nothing in answer_store) then we retrieve the metadata.
+
+        :param user_entered_address user entered address
+        :param metadata_address is metadata address
+        :return: the value of the address to be piped
+    """
+    if all(isinstance(a, Undefined) for a in user_entered_address):
```
insacuri: Sorry to be super picky, but could we rename `a` to `field` or something like that?  Makes it easier at a glance that it's a list of fields

app/jinja_filters.py
```diff
@@ -52,6 +52,29 @@ def format_percentage(value):
     return '{}%'.format(value)
 
 
+@blueprint.app_template_filter()
+def format_address_list(user_entered_address=None, metadata_address=None):
+    """
+        This function format_address_list accepts two addresses, a user submitted address and a metadata address
+        If all the items in address list are 'Undefined' (nothing in answer_store) then we retrieve the metadata.
+
+        :param user_entered_address user entered address
```
insacuri: More pickiness sorry, but could you expand these a bit more and say that they're lists of address fields.  Makes it easier to figure out what's going on at a glance

app/jinja_filters.py
dcdarrell9: That's what I'm thinking, if it's required to have at least one of these, then exception would be right, but if not then I'm not too sure, possible to redo this so that if the `user_entered_address` is empty then it defaults back to the `metadata_address` I think that way may be better?

jonnyshaw89: You need to check the behaviour with Donna

bitdivision (commented):
Looks good. Happy to approve once existing comments are addressed.

insacuri (approved):

app/jinja_filters.py
bitdivision: Did you resolve this @dcdarrell9 ?

app/jinja_filters.py
```diff
@@ -52,6 +52,29 @@ def format_percentage(value):
...
+        :param user_entered_address user entered address which is a list of answer fields
+        :param metadata_address metadata address which is a list of answer fields passed to us
+        :return: the value of the address to be piped
+    """
+    if all(isinstance(field, Undefined) for field in user_entered_address):
+        address = metadata_address
+    else:
+        address = user_entered_address
+
+    address_list = Markup('<br />').join([i for i in address if i])
```
bitdivision: Could you use concatenated_list here to keep things consistent?

dcdarrell9: Didn't see that, looks like it works, so gonna test a little 👍

bitdivision: Not a big deal really, just a little nicer.

tests/app/test_jinja_filters.py
```diff
@@ -859,3 +859,39 @@ def test_format_repeated_summaries_delimiters(self):
...
+                         format_address_list(user_entered_address, metadata_address))
+
+    def test_format_address_list_None_value(self):
+        user_entered_address = [None, None, None, None, None]
+        metadata_address = [None, None, None, None, None]
+        with self.assertRaises(Exception):
+            format_address_list(user_entered_address, metadata_address)
+
+    def test_format_address_list_no_values_in_answer(self):
+        user_entered_address = ['', '', '', '', '']
```
MebinAbraham, line 882: Not sure if this conversation happened, but is it really wrong to use metadata in this scenario?

bitdivision: I think that is probably what should happen, but needs to be confirmed with Donna.

jonnyshaw89:
@dcdarrell9 still waiting on confirmation about this https://github.com/ONSdigital/eq-survey-runner/pull/1751#discussion_r214864482

dcdarrell9:
@jonnyshaw89 There's a possibility of having 2 types of question...
1. At least one of need to be answered.
2. If none are answered, default to metadata

jonnyshaw89 (dismissed):

app/jinja_filters.py
```diff
@@ -52,6 +52,31 @@ def format_percentage(value):
...
+
+        :param user_entered_address user entered address which is a list of answer fields
+        :param metadata_address metadata address which is a list of answer fields passed to us
+        :return: the value of the address to be piped
+    """
+    if all(isinstance(field, Undefined) for field in user_entered_address):
+        address = metadata_address
+    elif all(field == '' for field in user_entered_address):
+        address = metadata_address
+    else:
```
dwyeradam: If you do find issues @jonnyshaw89 can you add that this `if and elif` combo should just be an `or` as it's doing the same thing.

dcdarrell9: Yeah, I thought about that, but then thought it looks easier to read like this, happy to change

bitdivision: Definitely better as an `or`. 
Otherwise you're using an if statement purely as a way to split up lines, which isn't what anyone would expect.

jonnyshaw89 (approved):

PR #1753, opened 2018-08-30
Fix legal basis not rendering on introduction page

dcdarrell9:
- Issue where legal basis was not being rendered

### How to review 
- Check that legal basis is rendered on introduction for different surveys (MWSS, MBS, Northern Ireland) and does not render if not in schema (census, lms)

### Checklist


tests/integration/introduction/test_introduction.py
```diff
@@ -57,3 +57,19 @@ def test_start_survey_sets_started_at(self):
         started_at_datetime = datetime.strptime(actual['started_at'], '%Y-%m-%dT%H:%M:%S.%f')
 
         self.assertIsNotNone(started_at_datetime)
+
+    def test_legal_basis_should_be_visible(self):
+        # Given survey with legal_basis for completion
+        self.launchSurvey('1', '0112')
```
dwyeradam: We don't use live surveys in tests. Can you change this to one of our test schemas please. Same with mbs below.

dwyeradam (approved):
Small change in the test but otherwise good.

dwyeradam (changes requested):
Sorry shouldn't have been approved.

insacuri (approved):

tests/integration/introduction/test_introduction.py
dcdarrell9: Shall I change the other ones in that file? I think there's 5 other ones there testing live surveys

jonnyshaw89 (changes requested):
I think we should add a functional test to check its being displayed

tests/integration/introduction/test_introduction.py
dwyeradam: Yes please. Boy scout rule.

dwyeradam: test_introduction.json should include any thing that would need testing.

tests/functional/spec/introduction.spec.js
```diff
@@ -0,0 +1,20 @@
...
+describe('@watch Introduction page', function() {
+
+  const introduction_schema = 'test_introduction.json';
+
+  it('Given I start a survey, When I view the introduction page then I should be able to see introduction information', function() {
+    return helpers.openQuestionnaire(introduction_schema).then(() => {
+      return browser
+        .getUrl().should.eventually.contain(IntroductionPage.pageName)
+        .click(IntroductionPage.showAllButton())
+        .isVisible(IntroductionPage.useOfData().should.eventually.be.true)
```
dwyeradam: Close brackets are in the wrong place.
Should be .isVisible(IntroductionPage.useOfData()`)`.should.eventually.be.true

tests/functional/pages/landing.page.js
```diff
@@ -10,6 +10,26 @@ class LandingPage extends BasePage {
     return '.qa-btn-get-started';
   }
 
+  useOfInformation() {
```
dwyeradam: Why aren't we using an `introduction.page.js`?

tests/functional/pages/landing.page.js
```diff
@@ -10,6 +10,26 @@ class LandingPage extends BasePage {
...
+    return '#use-of-information';
+  }
+
+  useOfData() {
+    return '#how-we-use-your-data';
+  }
+
+  legalResponse() {
+      return '#legal-response';
```
dwyeradam: I don't think this 'id' exists on the introduction page. Might be worth adding 'id's of 'data-qa's to the html to make testing easier.

dwyeradam (changes requested):

tests/functional/pages/landing.page.js
dcdarrell9: Arent' the landing/introduction pages the same thing?

dwyeradam: Yes really but since we don't have a concept of a landing page I would just rename `landing.page.js` to `introduction.page.js` to make it clear.
Don't forget to change it here as well `super('landing-page');`

tests/functional/spec/introduction.spec.js
```diff
@@ -0,0 +1,20 @@
...
+        .isVisible(IntroductionPage.useOfInformation().should.eventually.be.true)
+        .isVisible(IntroductionPage.legalResponse().should.be.eventually.true)
```
dwyeradam: We should check the content rather than it just being visible.

dwyeradam (changes requested):

app/themes/northernireland/templates/introduction.html
```diff
@@ -51,7 +51,7 @@ <h2 class="saturn">You are completing this for
             {% endif %}
 
             {%- if legal_basis -%}
-                <h2 class="neptune u-mb-xs">Your response is legally required</h2>
+                <h2 class="neptune u-mb-xs" id="legal-response">Your response is legally required</h2>
                 <p class="mars u-mb-m">{{ legal_basis }}</p>
```
dwyeradam: I would add an 'id' to `legal_basis` and then test the content of that in the functional test. This is the variable context we're actually trying to test here.

dwyeradam (changes requested):

app/templates/introduction.html
```diff
@@ -59,8 +59,8 @@ <h2 class="saturn">{{ _("You are completing this for <span>%(ru_name)s</span>%(t
             {% endif %}
 
             {%- if legal_basis -%}
-                <h2 class="neptune u-mb-xs">{{ _("Your response is legally required") }}</h2>
-                <p class="mars u-mb-m">{{ legal_basis }}</p>
+                <h2 class="neptune u-mb-xs" id="legal-response">{{ _("Your response is legally required") }}</h2>
```
jonnyshaw89: if these ids are just for the tests we should probably use `data-qa="legal-response"`

jonnyshaw89 (changes requested):

app/templates/introduction.html
dwyeradam: We currently have two 'id's that are the same. In basic.html and preview.html. This would cause issues as it's possible that they are both called so at least `data-qa` would negate that.

dwyeradam: Example of selector with data-qa rather than id:
`return '[data-qa="legal-response"]'`

app/templates/partials/introduction/basic.html
```diff
@@ -7,7 +7,7 @@ <h2 class="saturn">{{ intro.title }}</h2>
     <h3 class="venus">{{ content.title }}</h3>
     {% endif %}
     {% if content.description %}
-    <p class="mars qa-intro-description">
+    <p class="mars qa-intro-description" data-qa="intro-description">
```
dwyeradam: If you're going to use these then it's still best to uniquely name them. Currently same as intro-description on preview.html. `into-basic-description` and `intro-preview-description`???

dwyeradam (changes requested):

dwyeradam (approved):

jonnyshaw89 (approved):

jonnyshaw89:
@dcdarrell9 rebase and squash, without deleting changes?

jonnyshaw89:
@dcdarrell9 without deleting things? your tests failed.

PR #1755, opened 2018-08-30
Change country_code to country in LMS

jonnyshaw89:
RH is going to be sending `country` not `country_code`

### How to review 
Try launching with `country` not `country_code`

### DIFF
--- a/data/en/lms_1.json
+++ b/data/en/lms_1.json
@@ -21,7 +21,7 @@
         "name": "ru_name",
         "validator": "string"
     }, {
-        "name": "country_code",
+        "name": "country",
         "validator": "string"
     }, {
         "name": "display_address",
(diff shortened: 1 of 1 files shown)


insacuri:
There needs to be a corresponding PR in go-launch-a-survey.  Otherwise this looks fine

bitdivision (approved):
The only thing I can think of is in-flight data, but I don't think that would actually cause any issues if someone has `metadata['country_code']==W` while this goes out.

Is LMS part 1 live and in use?

jonnyshaw89:
@insacuri Here you go https://github.com/ONSdigital/go-launch-a-survey/pull/116

jonnyshaw89:
@bitdivision LMS is not "live"

dwyeradam (approved):

PR #1757, opened 2018-08-30
Allow comparison of answer values in when clause

bitdivision:
[Trello](https://trello.com/c/Lzc6bQVK/2323-route-based-comparing-one-answer-to-another-s-m)

[Validator PR](https://github.com/ONSdigital/eq-schema-validator/pull/83)

We need to be able to compare numeric answers against other numeric answers when doing routing. 
This PR adds a generic way of doing this that should work in all `when` clauses except titles.

It adds a `comparison_id` to the `when` clause. This is used to refer to an answer_id, for which the answer value is used in the comparison. Since this is in the when clause, it means comparisons can also be used for skip conditions, and repeating groups.

I've disallowed the use of `comparison_id` in titles (since there's no user need and it could cause issues with dependencies).

I've removed a couple of skip conditions which aren't needed in `1_0005.json` which refer to invalid answer ids. They were breaking the new validation.
### How to review 
There are test schemas for these when clauses when used for routing, skipping, and repeats.

Check `test_repeating_answer_comparison.json`, `test_routing_answer_comparison.json` and `test_skip_condition_answer_comparison.json`

#### test_repeating_answer_comparison.json
This test should repeat until the questions match

#### test_routing_answer_comparison.json
This test should route around an interstitial depending on the two answers.

#### test_skip_condition_answer_comparison.json
This test should skip interstitials depending on the two answers

### Check this code in particular
- `evaluate_repeat()` in `rules.py`
- The removed skip conditions in `1_00005.json`
### Checklist

### DIFF
--- a/app/forms/date_form.py
+++ b/app/forms/date_form.py
@@ -11,6 +11,7 @@ from babel.dates import get_month_names
 from app.forms.custom_fields import CustomIntegerField
 from app.questionnaire.rules import get_metadata_value, get_answer_store_value, convert_to_datetime
 from app.validation.validators import DateCheck, OptionalForm, DateRequired, MonthYearCheck, YearCheck, SingleDatePeriodCheck
+from app.utilities.schema import load_schema_from_metadata
 
 logger = logging.getLogger(__name__)
 
@@ -282,7 +283,8 @@ def get_referenced_offset_value(answer_min_or_max, answer_store, metadata):
     elif 'meta' in answer_min_or_max:
         value = get_metadata_value(metadata, answer_min_or_max['meta'])
     elif 'answer_id' in answer_min_or_max:
-        value = get_answer_store_value(answer_min_or_max['answer_id'], answer_store, group_instance=0)
+        schema = load_schema_from_metadata(metadata)
+        value = get_answer_store_value(answer_min_or_max['answer_id'], answer_store, schema, group_instance=0)
 
     value = convert_to_datetime(value)
 
--- a/app/questionnaire/rules.py
+++ b/app/questionnaire/rules.py
@@ -8,39 +8,63 @@ MAX_REPEATS = 25
 logger = logging.getLogger(__name__)
 
 
+def evaluate_comparison_rule(when, answer_value, comparison_value):
+    """
+    Determine whether a comparison rule will be satisfied based on an
+    answer value, and a value to compare it to.
+    :param when: The when clause to evaluate
+    :param answer_value: The value of the answer
+    :param comparison_value: The value to compare the answer to.
+    :return (bool): The result of the evaluation
+    """
+    condition = when['condition']
+
+    return evaluate_condition(condition, answer_value, comparison_value)
+
+
 def evaluate_rule(when, answer_value):
     """
     Determine whether a rule will be satisfied based on a given answer
-    :param when:
-    :param answer_value:
-    :return:
+    :param when: The when clause to evaluate
+    :param answer_value: The value of the answer
+    :return (bool): The result of the evaluation
     """
+
     match_value = when['value'] if 'value' in when else None
+
     condition = when['condition']
 
     # Evaluate the condition on the routing rule
-    return evaluate_condition(condition, match_value, answer_value)
+    return evaluate_condition(condition, answer_value, match_value)
 
 
-def evaluate_date_rule(when, answer_store, group_instance, metadata, answer_value):
+def evaluate_date_rule(when, answer_store, schema, group_instance, metadata, answer_value):
     date_comparison = when['date_comparison']
 
     answer_value = convert_to_datetime(answer_value)
-    match_value = get_date_match_value(date_comparison, answer_store, group_instance, metadata)
+    match_value = get_date_match_value(date_comparison, answer_store, schema, group_instance, metadata)
     condition = when.get('condition')
 
     if not answer_value or not match_value or not condition:
         return False
 
     # Evaluate the condition on the routing rule
-    return evaluate_condition(condition, match_value, answer_value)
+    return evaluate_condition(condition, answer_value, match_value)
 
 
-def evaluate_condition(condition, match_value, answer_value):
+def evaluate_condition(condition, answer_value, match_value):
+    """
+    :param condition: string representation of comparison operator
+    :param answer_value: the left hand operand in the comparison
+    :param match_value: the right hand operand in the comparison
+    :return: value of comparing lhs and rhs using the specified operator
+    """
     result = False
-    if condition == 'equals' and match_value == answer_value:
+    answer_and_match = answer_value is not None and match_value is not None
+
+    if condition == 'equals' and answer_value == match_value:
         result = True
-    elif condition == 'not equals' and match_value != answer_value:
+    elif condition == 'not equals' and answer_value != match_value:
         result = True
     elif condition == 'contains' and isinstance(answer_value, list) and match_value in answer_value:
         result = True
@@ -50,15 +74,15 @@ def evaluate_condition(condition, match_value, answer_value):
         result = answer_value is not None
     elif condition == 'not set':
         result = answer_value is None
-    elif condition == 'greater than' and answer_value and answer_value > match_value:
+    elif condition == 'greater than' and answer_and_match and answer_value > match_value:
         result = True
-    elif condition == 'less than' and answer_value and answer_value < match_value:
+    elif condition == 'less than' and answer_and_match and answer_value < match_value:
         result = True
 
     return result
 
 
-def get_date_match_value(date_comparison, answer_store, group_instance, metadata):
+def get_date_match_value(date_comparison, answer_store, schema, group_instance, metadata):
     match_value = None
 
     if 'value' in date_comparison:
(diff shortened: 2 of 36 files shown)


bitdivision:
I probably need to add functional tests for this. I'll do that now, but the rest should be reviewable now.

app/questionnaire/rules.py
```diff
@@ -8,18 +8,26 @@
...
     :param when:
     :param answer_value:
+    :param comparison_id_value , answer value used for comparison
     :return:
     """
-    match_value = when['value'] if 'value' in when else None
+
+    if 'comparison_id' in when:
+        if comparison_id_value is None:
+            return False
```
dwyeradam: We need to consider why this is None? An answer that hasn't been answered would be none and therefore you may want to evaluate it against another answer that also hasn't been set. However if that answer doesn't exist this would also be None and you probably would want it to return False. It may be worth considering splitting this logic to a different method.

dwyeradam (changes requested):

app/questionnaire/rules.py
bitdivision: @dwyeradam - I've sorted this out now I think, although the git history has some major weirdness in it. Sorry about that.

app/questionnaire/rules.py
```diff
@@ -222,13 +285,21 @@ def get_answer_store_value(answer_index, answer_store, group_instance, group_ins
     return filtered[0]['value'] if filtered.count() == 1 else None
 
 
+def get_answer_store_value_ignoring_group_instance_id(answer_index, answer_store, group_instance):
```
bitdivision: I'm not particularly happy with this solution, does anyone have any suggestions?

app/questionnaire/rules.py
```diff
@@ -8,18 +8,34 @@
 logger = logging.getLogger(__name__)
 
 
+def evaluate_comparison_rule(when, answer_value, comparison_value):
+    """
```
AndrewTorrance, line 12: Minor point . Perhaps a name such as evaluate_answer_comparison_rule may be more explicit since evaluate_comparison_rule could mean comparing answers to constants ?

AndrewTorrance (changes requested):

app/questionnaire/rules.py
bitdivision: Not averse to changing, but I think it could be used in both cases?

AndrewTorrance (dismissed):

app/questionnaire/rules.py
dwyeradam: Can we simplify this with `schema.answer_is_in_repeating_group(answer_id)` to decide if we need to include group_instance_id?

bitdivision: @dwyeraadam - I had been under the impression that answer_store.filter actually filtered on None values. As Jon said, it doesn't need to, and on further inspection, ignores None values.

So I think this stuff can be completely removed :)

bitdivision: I've pulled that out. Ready for review I think

app/questionnaire/rules.py
```diff
@@ -156,6 +182,19 @@ def _get_answer_count_minus_one(filtered_answers):
...
+        Gets the value of an answer specified as an operand in a comparator ( i.e right hand argument ,
+        or rhs in if lhs>rhs ) In a when clause
+        Supports lhs being in a repeating group and rhs can be either in a similar sized repeating group or
+        not in a repeating group
+    """
+
+    if 'comparison_id' in when_rule:
+        return get_answer_store_value(answer_index=when_rule['comparison_id'], answer_store=answer_store,
+                                      group_instance=group_instance, group_instance_id=group_instance_id)
```
dwyeradam: I'm not sure this will always work if the comparison answer doesn't have the same group instance and group_instance_id as the group you're currently in. For instance if you're in a repeating group but the answer is outside of that repeating group then group_instance and group_instance_id may well not match. Alternatively if you're in a repeating group made up of more than one group driver then the group_instance wouldn't be correct.
We need some logic to work out where the comparison_id sits in relation to the group you are currently in.

dwyeradam (changes requested):

app/questionnaire/rules.py
bitdivision: Ah, that sounds right. I'll write some tests for that and fix it up.

dcdarrell9:
test_repeating_answer_comparison.json:
- Not sure what expected behaviour for this is, works as I would think it should, but if you give a non matching answer first (4, 5) no you then answer a matching number (6, 6) then you'll go to the summary which gives the answers (4, 5)... 
- If you then change that answer, you'll always be returned to the summary page where the value does change, but you're out of the repeating answer (again, not sure what expected functionality is, just wanted to write something down)

test_routing_answer_comparison.json:
- Same thing with above, works fine but once you get to summary it ignores everything and just returns.

test_skip_condition_answer_comparison.json:
- Works fine 👍

bitdivision:
@dcdarrell9 Good spots but I think this behaviour is as it was before for repeating groups. I think @dwyeradam is working on fixing the summaries. Please correct me if I'm wrong @dwyeradam .

You can make sure it's not a regression by changing the `condition_id` key to a `value` key and testing. I've double checked on repeating_answer_comparison, and see the same behaviour that you noted.

app/forms/date_form.py
```diff
@@ -282,7 +284,9 @@ def get_referenced_offset_value(answer_min_or_max, answer_store, metadata):
     elif 'meta' in answer_min_or_max:
         value = get_metadata_value(metadata, answer_min_or_max['meta'])
     elif 'answer_id' in answer_min_or_max:
-        value = get_answer_store_value(answer_min_or_max['answer_id'], answer_store, group_instance=0)
+        session_data = get_session_store().session_data
+        schema = load_schema_from_session_data(session_data)
```
MebinAbraham: Just a suggestion, at this point you already have metadata, therefore, you can make use of `load_schema_from_metadata` instead of getting the session from globals to get the metadata.

bitdivision: Thanks! That makes things a little nicer.

MebinAbraham:
@bitdivision @dcdarrell9 Pretty sure its going to summary after the first time because of how we calculate `completeness`. Interstitials are by default completed, hence you only see it the one time before clicking Continue.

MebinAbraham (commented):
Tested some of the schemas and the behaviour seems to be as expected. After you `complete` the blocks, the pages that you skip/route etc does not get shown because of `completeness`, so I think we should look to fix this behaviour soon. (Need to mark the blocks as a dependency when values or changes?)

tests/functional/spec/features/repeating/answer_comparison_repeating_groups.spec.js
```diff
@@ -0,0 +1,31 @@
...
+        return browser
+          .setValue(RepeatingComparison1BlockPage.answer(), 5)
+          .click(RepeatingComparison1BlockPage.submit())
+          .setValue(RepeatingComparison2BlockPage.answer(), 6)
+          .click(RepeatingComparison2BlockPage.submit())
+          .getText(RepeatingComparison2BlockPage.questionText()).should.eventually.contain('Enter a number');
+        });
+    });
+
+  it('@watch Given we open the repeating comparison test When we enter the same numbers, Then the question should not repeat', function() {
```
jonnyshaw89: remove `@watch`

tests/functional/spec/features/repeating/answer_comparison_repeating_groups.spec.js
```diff
@@ -0,0 +1,31 @@
+const helpers = require('../../../helpers');
+
+const RepeatingComparison1BlockPage = require('../../../pages/features/answer_comparison/repeating_groups/repeating-comparison-1-block.page.js');
+const RepeatingComparison2BlockPage = require('../../../pages/features/answer_comparison/repeating_groups/repeating-comparison-2-block.page.js');
+const SummaryPage = require('../../../pages/features/answer_comparison/repeating_groups/summary.page.js');
+
+describe('Test repeating with answer comparisons', function() {
+
+  it('Given we open the repeating comparison test When we enter different numbers, Then the question should repeat', function() {
+    return helpers.openQuestionnaire('test_repeating_answer_comparison.json').then(() => {
```
jonnyshaw89: if you move the `helpers.openQuestionnaire` into a before each then you don't need to repeat it in each test. that applies for all, as its 1 schema per test

jonnyshaw89 (changes requested):

jonnyshaw89 (approved):

app/templating/view_context.py
```diff
@@ -44,7 +45,7 @@ def build_view_context_for_non_question(variables, csrf_token, rendered_block):
     }
 
 
-def build_view_context_for_question(metadata, schema, answer_store, current_location, variables, rendered_block, form):  # noqa: C901, E501  pylint: disable=too-complex,line-too-long
+def build_view_context_for_question(metadata, schema, answer_store, current_location, variables, rendered_block, form):  # noqa: C901, E501  pylint: disable=too-complex,line-too-long,too-many-locals
```
dwyeradam, line 48: cheat.

dwyeradam (approved):
@bitdivision squash away

PR #1758, opened 2018-09-03
Use xdist to speed up tests in travis

bitdivision:
This may speed up the travis pipeline a little by parallelizing the tests (using pytest-xdist)

### Checklist


jonnyshaw89:
I like it but I also don't.
Im worried we are covering up the issue that runner has got a lot slower

bitdivision:
Agreed, it's just a bandage. 

It probably won't make a massive difference to Travis since I think that only gives you two cores. But it drastically improves performance on my machine, thought everyone should at least be aware.

Maybe should discuss in tech session.

bitdivision:
I think the other 'easy' thing we could do is to split things up into different build paths so that travis could parallelize itself.

jonnyshaw89:
Yeh we used to use matrix in travis, and a few months back we merged it back into 1

bitdivision:
@jonnyshaw89 Seems like removing all the census tests has dropped it from 29-30 minutes to ~24 minutes.

Definitely helps.

jonnyshaw89:
You could potentially save another `54.26s` by not building the runner container twice?

Pipfile
```diff
@@ -22,9 +22,9 @@ httmock = "*"
 "flake8-mock" = "*"
 "flake8-print" = "*"
 "flake8-tuple" = "*"
-moto = "*"
-google-compute-engine = "*"
+moto = "==1.3.4"
```
bitdivision, line 25: This had to be done because moto doesn't support the latest boto, so they pinned to an old version. We can't run on that version so had to pin boto as well.

jonnyshaw89 (approved):

MebinAbraham (commented):
Tested and seems to work as expected. @dwyeradam Opinions?

MebinAbraham (approved):

PR #1761, opened 2018-09-03
Support negative values for QCAS

dwyeradam:
Support Negative Values for QCAS


dcdarrell9 (approved):
Changes LGTM. 
But I don't know whether all or some of the questions are supposed to handle negative (I'm assuming they all are)

dcdarrell9:
Just one question, why is it 10 characters here `-999999999` but with SDX it's 11 characters `99999999999` ?

MebinAbraham (approved):

PR #1763, opened 2018-09-04
Wrap Currency values in span to ensure negative values wrap properly

dwyeradam:
Now that QCAS supports negative values it was noticed when the formatted currency value was displayed on the confirmation page it wrapped on the negative symbol rather than defining the whole thing as a single entity.

### How to review 
Start QCAS and enter a negative value. The corresponding confirmation page should have the complete formatted entered value in a span.

### Checklist

### DIFF
--- a/app/jinja_filters.py
+++ b/app/jinja_filters.py
@@ -25,8 +25,14 @@ def format_number(value):
     return ''
 
 
-@blueprint.app_template_filter()
-def format_currency(value, currency='GBP'):
+@evalcontextfunction
+def format_currency(context, value, currency='GBP'):
+    currency_value = get_formatted_currency(value, currency)
+    result = "<span class='date'>{currency_value}</span>".format(currency_value=currency_value)
+    return mark_safe(context, result)
+
+
+def get_formatted_currency(value, currency='GBP'):
     if value or value == 0:
         return numbers.format_currency(number=value, currency=currency, locale=DEFAULT_LOCALE)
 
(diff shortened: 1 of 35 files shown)


app/jinja_filters.py
```diff
@@ -25,8 +25,15 @@ def format_number(value):
     return ''
 
 
+@evalcontextfunction
 @blueprint.app_template_filter()
-def format_currency(value, currency='GBP'):
+def format_currency(context, value, currency='GBP'):
+    currency_value = get_formatted_currency(value, currency)
+    result = "<span class='date'>{currency_value}</span>".format(currency_value=currency_value)
```
dwyeradam: This will use the `date` class in the span for now. A new card exists to use a new utility class that has been added to the pattern library. https://github.com/ONSdigital/sdc-global-design-patterns/pull/161

bitdivision:
Slight problem with the context processor, it's still being used as a filter in the schemas. I guess it needs to be defined as both in template renderer, or the schemas need to be updated.

bitdivision (changes requested):
Need to modify filter use in qcas (and many others) since it currently 500.

dwyeradam:
@bitdivision, my bad. So we can support other currencies in the future I felt it best to do schema changes. Easier to see changes in separate commits so worked out for the best.

app/jinja_filters.py
```diff
@@ -25,8 +25,15 @@ def format_number(value):
     return ''
 
 
+@evalcontextfunction
 @blueprint.app_template_filter()
```
bitdivision: This can be removed now I think?

bitdivision (approved):
Minor point on extra decorator. Happy otherwise.

MebinAbraham (commented):
Existing, but I can't see where this is even used?.
https://github.com/ONSdigital/eq-survey-runner/blob/9a7db32bb0880bf114e088f60689a1553f11e9bf/app/jinja_filters.py#L47-L53

MebinAbraham (approved):

PR #1768, opened 2018-09-05
Added language switcher

jonnyshaw89:
Allows language to be switched for social surveys
Currently hard coded to en/cy

### How to review 
load `test_language.json` and use the switcher in the top right

### DIFF
--- a/app/jinja_filters.py
+++ b/app/jinja_filters.py
@@ -403,6 +403,11 @@ def max_value(a, b):
         raise Exception(msg)
 
 
+@blueprint.app_template_filter()
+def language_urls(languages, current_language):
+    return [language for language in languages if language[0] != current_language]
+
+
 @contextfunction
 @blueprint.app_template_filter()
 def get_question_title(context, question_id):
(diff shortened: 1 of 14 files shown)


jonnyshaw89:
@boxadesign can you have a look at the styles here after the new header is in?

dcdarrell9:
Will this need to be able to switch at different points? Like introduction/summary, if so would it be worth adding to the test schema

jonnyshaw89:
@dcdarrell9 you mean I should test that the switcher works on each page?

app/jinja_filters.py
```diff
@@ -373,6 +373,17 @@ def max_value(a, b):
         raise Exception(msg)
 
 
+@blueprint.app_template_filter()
+def language_urls(languages, current_language):
+    switcher = []
+    for language in languages:
+        if language[0] == current_language:
+            continue
```
bitdivision: Not a big deal, but could this be made simpler?
```
return [l for l in languages if l[0] != current_language]
```

bitdivision (commented):
Looks good, will test in a moment.

app/jinja_filters.py
jonnyshaw89: ah yeh, I refactored some logic out and was lazy

bitdivision (approved):
Just need to sort out `ru_name`

dcdarrell9 (approved):

PR #1771, opened 2018-09-07
Add collapsible summary to MWSS and update pattern library to 1.5.6

dcdarrell9:
- Add collapsible summary to MWSS

### How to review 
- Check MWSS summary contains accordions and is correct

### Checklist

### DIFF
--- a/app/assets/favicons/browserconfig.xml
+++ b/app/assets/favicons/browserconfig.xml
@@ -2,7 +2,7 @@
 <browserconfig>
     <msapplication>
         <tile>
-            <square150x150logo src="https://cdn.ons.gov.uk/sdc/v1.5.2/favicons/mstile-150x150.png"/>
+            <square150x150logo src="https://cdn.ons.gov.uk/sdc/v1.5.6/favicons/mstile-150x150.png"/>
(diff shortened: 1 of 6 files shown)


insacuri (dismissed):

jonnyshaw89 (dismissed):
Doesn't scroll properly

jonnyshaw89:
@boxadesign It appears as though the summary scrolls independently of the page. This make the page behave very strangely when scrolling

boxadesign:
@jonnyshaw89 @dcdarrell9 I have fixed the issue with the summary component. To do this I had to also bring runner up to the latest version of the pattern library (v1.5.5). Version 1.5.3 had a small breaking change in it for the character counter in the textarea template (which I've fixed in my last commit)

bitdivision (approved):
Looks as I would expect.

MebinAbraham (approved):
Changes look fine, text area and collapsable summary working as expected. We should upgrade to cdn ver `1.5.6`, since that has just been released and it has no breaking changes.

bitdivision:
Looks good still @dcdarrell9 .

PR #1772, opened 2018-09-07
Able to return to ras from eq via 'My account' link in header

insacuri:
https://trello.com/c/uTTLYYwX/2017-able-to-return-to-ras-from-eq

This adds the 'My account' link to the header of every page. 

This shouldn't be merged yet as I've not yet made the account_url mandatory and got the tests working at the same time.

### How to review 
- Open any survey, in the header should be 'My account' on every page throughout the survey.
- After launching a survey, add an extra slash and some garbage e.g., `/blah` to fire up an error page.  My account link should still appear if it was in the original payload when the survey was launched.
- Go to `localhost:5000` (or whatever port runner is on), an error page should appear WITHOUT a my account link in the header

### Checklist


jonnyshaw89:
Can you update PR title?

app/helpers/template_helper.py
```diff
@@ -78,6 +78,13 @@ def legal_basis_wrapper(*args, **kwargs):
     return legal_basis_wrapper
 
 
+def get_account_url():
+    metadata = get_metadata(current_user)
+    if metadata.get('account_url') not in ['', None]:
```
MebinAbraham: Can't this just be `if metadata.get('account_url'):`?

app/templates/session-expired.html
```diff
@@ -12,7 +12,8 @@ <h1 class="saturn" {{helpers.track('visible', 'Timeout', 'Timeout expired')}}>{{
   <div class="panel panel--simple panel--success panel--spacious u-mb-s">
     <p>{{ _("To help protect your information we have signed you out") }}.</p>
   </div>
-
+  {% if respondent_account_url is not none %}
```
MebinAbraham: Here also same as above, this would be true for an empty string, so just `{% if respondent_account_url %}` would cover both cases.

app/templates/signed-out.html
```diff
@@ -9,6 +9,8 @@
     <h1 class="saturn">{{ _("Your survey answers have been saved. You are now signed out") }}.</h1>
 </div>
 
-<a href="{{respondent_account_url}}">{{ _("Return to your account") }}</a>
+{% if respondent_account_url is not none %}
```
MebinAbraham: Same as above.

app/views/questionnaire.py
```diff
@@ -341,6 +342,12 @@ def get_view_submission(schema, eq_id, form_type):  # pylint: disable=unused-arg
     return redirect(url_for('post_submission.get_thank_you', eq_id=eq_id, form_type=form_type))
 
 
+def _get_account_url(metadata_context):
```
MebinAbraham: Can we not reuse this method?. The same thing is in multiple files?

app/templates/partials/header.html
```diff
@@ -12,6 +12,13 @@
                     </div>
                 {% endblock logo %}
                 </div>
+                <div class="grid__col col-6@s">
+                    <nav class="nav nav--horizontal  u-d-no--@xs@s u-fr" aria-label="Site services menu">
+                        <ul class="nav__list" role="menubar" aria-label="Site services menu">
+                            <li class="nav__item" role="menuitem"><a href="{{respondent_account_url}}" id="my-account" class="nav__link">My account</a></li>
```
MebinAbraham: You need to mark any new static content for translation. ie {{ _(My text) }}

MebinAbraham (changes requested):

app/views/questionnaire.py
insacuri: I don't really want to because various ones have different places that it looks for its information.  Some are looking at metadata, others are looking at the session.
One thing that was confusing throughout this was what was available to me when dealing with each page.  Having a copy in a file makes that slightly clearer

app/templates/partials/header.html
insacuri: Good point, good spot!

dcdarrell9: ++ Generate new message.pot file

dcdarrell9 (dismissed):

app/views/session.py
```diff
@@ -69,8 +70,8 @@ def login():
     cookie_session['theme'] = g.schema.json['theme']
     cookie_session['survey_title'] = g.schema.json['title']
 
-    if 'account_service_url' in claims and claims.get('account_service_url'):
-        cookie_session[ACCOUNT_URL] = claims.get('account_service_url')
+    if 'account_url' in claims and claims.get('account_url'):
```
MebinAbraham: I know its just renamed from previous but we don't need the first check so can we remove it?. So just `if claims.get('account_url'):`

MebinAbraham (commented):
Do you know if the default value for this is meant to be this:
https://github.com/ONSdigital/eq-survey-runner/blob/851e909d86fb1c2875ae5a1a5c957b69e8d7ef50/app/settings.py#L83

Because that link doesn't seem to work and says the new one is: `https://surveys.ons.gov.uk/`

app/views/questionnaire.py
```diff
@@ -259,7 +259,7 @@ def post_household_composition(routing_path, schema, metadata, answer_store, **k
 @login_required
 @with_metadata
 @with_schema
-def get_thank_you(schema, metadata, eq_id, form_type):  # pylint: disable=unused-argument
```
bitdivision, line 262: :+1:

bitdivision (dismissed):

app/authentication/authenticator.py
```diff
@@ -78,7 +78,7 @@ def _create_session_data_from_metadata(metadata):
         case_id=metadata.get('case_id'),
         case_ref=metadata.get('case_ref'),
         trad_as=metadata.get('trad_as'),
-        account_service_url=metadata.get('account_service_url'),
+        account_url=metadata.get('account_url'),
```
jonnyshaw89: Why are we changing this without updating the spec?
https://ons-schema-definitions.readthedocs.io/en/latest/respondent_management_to_electronic_questionnaire.html#required-fields

app/helpers/template_helper.py
```diff
@@ -78,6 +78,13 @@ def legal_basis_wrapper(*args, **kwargs):
     return legal_basis_wrapper
 
 
+def get_account_url():
+    metadata = get_metadata(current_user)
+    if metadata.get('account_url'):
```
jonnyshaw89: could this be?
`return metadata.get('account_url', current_app.config['RESPONDENT_ACCOUNT_URL'])`

app/views/errors.py
```diff
@@ -89,10 +89,9 @@ def get_tx_id():
 
 
 def get_account_url():
-    account_url = current_app.config['RESPONDENT_ACCOUNT_URL']
-    if ACCOUNT_URL in cookie_session:
-        account_url = cookie_session[ACCOUNT_URL]
-    return account_url
+    if cookie_session.get('account_url'):
```
jonnyshaw89: `return cookie_session.get('account_url', current_app.config['RESPONDENT_ACCOUNT_URL'])`

jonnyshaw89 (changes requested):

dcdarrell9:
So, these are the comments that Adam left in the chat about the `account_service_url`, but there were no replies..

Is there a difference between the `account_url` found in the metadata and the `account_service_url` found in the session data?

For the thank you page, there isn't metadata anymore, but there is session data (which is confusingly name metadata....but that's for another time).  In session data there is an `account_service_url` that never seems to be written too and never seems to be used by anything.  There IS however in the metadata/claims (I think the terms are interchangeable but correct me if I'm wrong) an `account_url` that gets passed in, but never gets written to the session.

I *think* the right thing to do is to rename the `account_service_url` to `account_url` so that it gets written to the session, so that when you're on the thank you page, you have access to it even while you're signed out

ajmaddaford:
I think we should discuss this pull request in the tech session today. We should also consider removing the fallback to a config RESPONDENT_ACCOUNT_URL as this was only implemented for RRM and should always be passed by new clients.

app/views/errors.py
```diff
@@ -104,5 +97,5 @@ def render_template(template_name):
                                  analytics_ua_id=current_app.config['EQ_UA_ID'],
                                  ua=user_agent,
                                  tx_id=tx_id,
-                                 respondent_account_url=get_account_url(),
+                                 respondent_account_url=cookie_session.get('account_service_url'),
```
insacuri: Need to change key to 'account_service_url'

tests/functional/spec/my_account_header_link.js
```diff
@@ -0,0 +1,26 @@
+const helpers = require('../helpers');
+
+const IntroductionPage = require('../pages/introduction.page');
+const GenericPage = require('../pages/surveys/generic.page');
+const SummaryPage = require('../pages/summary.page');
+const ThankYouPage = require('../pages/thank-you.page');
+
+describe('Introduction page', function() {
+
+  it('@watch Given I start a survey, When I go through every page then I should see the "My account" button on each page', function() {
```
insacuri: Need to remove watch

insacuri:
The comment about having the account_url in the page url isn't relevant now.  It gets the url out of the cookie, and if it's not in the cookie then the my account link doesn't appear.

MebinAbraham:
@insacuri Is there an updated spec?, PR says: display on every page, Trello card says: return back to 'My account' page from the start of a survey and the completion page.

Am I correct in thinking, we want to display on every page if we have the data?

MebinAbraham (commented):
I've just launched MWSS (1_0005) and I don't see the My Account link?

edit: Ok working with GO launch PR.

MebinAbraham (changes requested):
This doesn't display the `My Account` link for other themes such as `Northern Ireland`.

insacuri:
@MebinAbraham Link should be appearing now on northern ireland themed surveys

boxadesign:
@insacuri There is some additional elements required for it to work on mobile. Are you happy for me to dive in and do the work?

insacuri:
@boxadesign go for it!

app/views/questionnaire.py
```diff
@@ -280,6 +280,7 @@ def get_thank_you(schema, metadata, eq_id, form_type):  # pylint: disable=unused
                                      survey_title=TemplateRenderer.safe_content(schema.json['title']),
                                      is_view_submitted_response_enabled=is_view_submitted_response_enabled(schema.json),
                                      view_submission_url=view_submission_url,
+                                     account_service_url=metadata_context.get('account_service_url'),
```
jonnyshaw89: I thought we were going to always use session?

tests/functional/pages/introduction.page.js
```diff
@@ -6,6 +6,10 @@ class IntroductionPage extends BasePage {
     super('introduction-page');
   }
 
+  myAccountLink() {
```
jonnyshaw89, line 9: you could push this to `BasePage` so it doesn't need to be duplicated?

jonnyshaw89 (changes requested):

app/views/questionnaire.py
insacuri: oh, good spot!  Thought I'd updated them all

tests/functional/pages/introduction.page.js
insacuri: It's not there on the error page.  Is it okay to move it there if it's not on literally every page?

app/helpers/template_helper.py
```diff
@@ -86,6 +86,7 @@ def render_template(template, **kwargs):
         theme,
         template,
         survey_title=TemplateRenderer.safe_content(g.schema.json['title']),
+        account_service_url=get_metadata(current_user).get('account_service_url'),
```
jonnyshaw89: Is there a reason we didn't use session here?

jonnyshaw89 (changes requested):

jonnyshaw89 (approved):

app/views/session.py
```diff
@@ -69,8 +70,8 @@ def login():
     cookie_session['theme'] = g.schema.json['theme']
     cookie_session['survey_title'] = g.schema.json['title']
 
-    if 'account_service_url' in claims and claims.get('account_service_url'):
-        cookie_session[ACCOUNT_URL] = claims.get('account_service_url')
+    if 'account_service_url' in claims:
```
bitdivision: It can never be empty or None in claims?

bitdivision: Or, to clarify, previously we didn't set the url if it was None in claims. Now we will. Just want to make sure that's expected.

insacuri: Thats a good point...I'll put that back to what it was.  There's no adverse effects from this, but there's no point in adding keys with empty values in the session if we can avoid it.

bitdivision (approved):
Tested and looked through the code fairly carefully.

jonnyshaw89:
@MebinAbraham on you

MebinAbraham (approved):

PR #1774, opened 2018-09-10
Don't error on missing member_label

jonnyshaw89:
Schemas currently error when `member_label` isn't defined on relationships
This allows them to not be present and will default to answer value

### How to review 
Launch `lms_2.json` and see if relationships render


dwyeradam (approved):
Fixes the issue with LMS and anything else that may not have member_label. I do wonder if it might be better to enforce `member_label` for relationships in validator though.

bitdivision (approved):

PR #1776, opened 2018-09-11
Replace ru_name with display_adderss

insacuri:
https://trello.com/c/iugfWgRR/2392-add-address-fields-to-lms-schema

Just like the lms_1 survey, lms_2 needs a display address so that people can confirm their address.

### How to review 
`display_address` should appear as a metadata field when you open it in launcher.  The the address questions should contain that `display_address` that was passed through.

### Checklist

### DIFF
--- a/data/en/lms_2.json
+++ b/data/en/lms_2.json
@@ -27,6 +27,10 @@
         {
             "name": "country",
             "validator": "string"
+        },
+        {
+            "name": "display_address",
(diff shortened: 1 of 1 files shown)


insacuri:
Added don't merge label while I make sure that this is the correct solution

insacuri:
Got the requirements, this needs to be changed a bit so closing the PR

PR #1781, opened 2018-09-12
Add answer comparison routing LMS2

bitdivision:
LMS 2 has two totalisers with answer comparison routing.

### How to review 
In the spec, look at TOTAC1 and TOTAC2 questions. These are asked when the user has one or two jobs respectively. The various rules defined in red for each question show how routing should be done going out of the totaliser.

There is a proxy answer for self employed, so both need to be checked. We also do not have a `greater than or equal` condition, so have to use `>` and `=` separately. This gives 12 conditions for each question.

Testing would consist of checking each of the >,=,< paths while choosing employed and self employed and checking while proxying.

### Checklist

### DIFF
--- a/data/en/lms_2.json
+++ b/data/en/lms_2.json
@@ -4020,9 +4020,9 @@
                                             "value": "employee"
                                         },
                                         {
-                                            "comparison_id": "two-jobs-j1-usual-hours-answer",
-                                            "condition": "less than",
-                                            "id": "two-jobs-j1-actual-hours-answer"
+                                            "id": "two-jobs-j1-actual-hours-answer",
+                                            "condition": "greater than",
+                                            "comparison_id": "two-jobs-j1-usual-hours-answer"
                                         }
                                     ]
                                 }
@@ -4036,9 +4036,9 @@
                                             "value": "employee"
                                         },
                                         {
-                                            "comparison_id": "two-jobs-j1-usual-hours-answer",
-                                            "condition": "less than",
-                                            "id": "two-jobs-j1-actual-hours-answer"
+                                            "id": "two-jobs-j1-actual-hours-answer",
+                                            "condition": "greater than",
+                                            "comparison_id": "two-jobs-j1-usual-hours-answer"
                                         }
                                     ]
                                 }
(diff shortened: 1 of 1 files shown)


insacuri (dismissed):
LGTM, tested a bunch of scenarios and it all seemed to work fine.

data/en/lms_2.json
```diff
@@ -3774,7 +3773,234 @@
...
+                                "goto": {
+                                    "block": "reason-why-fewer-hours-than-usual-self-employed-w14-semp",
+                                    "when": [{
+                                            "condition": "equals",
+                                            "id": "one-job-emp-status-emp-or-self-answer",
+                                            "value": "Self-employed"
+                                        },
+                                        {
+                                            "id": "one-job-usual-hours-answer",
+                                            "condition": "less than",
```
dcdarrell9: As one respondent with a full-time self-employed job, I put usual hours as 55 and actual hours as 77 and I got to this question `reason-why-fewer-hours-than-usual-self-employed-w14-semp` Shouldn't it be the other way around?

dcdarrell9 (changes requested):

dcdarrell9:
- The question `reason-why-fewer-hours-than-usual-self-employed-w14-semp` kind of doesn't make much sense, put my usual hours on two jobs as 20+20 and my actual as 10+40 and it asks the 'what is the main reason worked fewer', when actually worked more?

dcdarrell9:
screen shot 2018-09-14 at 10 36 42

dcdarrell9:
screen shot 2018-09-14 at 10 54 04

- I think one-job-usual-hours-w12 needs to have a label for hours like one-job-actual-hours-w12

bitdivision:
> * I think `one-job-usual-hours-w12` needs to have a label for hours like `one-job-actual-hours-w12`
> * The question `reason-why-fewer-hours-than-usual-self-employed-w14-semp` kind of doesn't make much sense, put my usual hours on two jobs as 20+20 and my actual as 10+40 and it asks the 'what is the main reason worked fewer, when actually worked more?

This is partially since we don't support sums in answer comparisons. So they can't use both answers, they can only use the main job.

I did double check this with them and they're fine with only using the main job

data/en/lms_2.json
bitdivision: Absolutely right. All the conditions are the wrong way around.

Good spot!

dcdarrell9:
Not sure if this is to do with this or not...
- As a respondent with 2 self-employed jobs... If I select anything plus casual working in `working-hours-flex10` I only ever get asked the casual hours question then the survey finishes.

bitdivision:
With respect of the title including reference dates, that is seperate to this PR. I'll raise another one that replaces with ... as per the other titles.

bitdivision:
Regarding casual hours, I don't think that is to do with this, but AFAICS, you should be routed to `casual-hours-worked-casac`, or `overtime-w6`

dcdarrell9 (approved):

jonnyshaw89 (approved):
LGTM 👍

PR #1786, opened 2018-09-13
LMS Change order of birth country depending on metadata

bitdivision:
[Trello](https://trello.com/c/LeyGkUPV/2376-lms-re-ordering-answers-depending-on-meta-data-s)

Different order of countries of birth depending on the `country` that is passed from upstream.

### How to review 
Select different country codes in launches and ensure that the order changes appropriately

Valid codes: `S`: Scotland, `N`: Scotland, `E`: England, `W`: Wales
### Checklist

### DIFF
--- a/data/en/lms_2.json
+++ b/data/en/lms_2.json
@@ -662,18 +662,20 @@
                                     "label": "Please enter nationality"
                                 }
                             ]
-                        }],
-                        "routing_rules": [{
-                            "goto": {
-                                "block": "country-of-birth"
-                            }
                         }]
                     },
                     {
                         "type": "Question",
-                        "id": "country-of-birth",
+                        "id": "country-of-birth-wales",
+                        "skip_conditions": [{
+                            "when": [{
+                                "meta": "country",
+                                "condition": "not equals",
+                                "value": "W"
+                            }]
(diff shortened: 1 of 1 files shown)


dcdarrell9 (dismissed):
Tested with all different countries and worked 👍

data/en/lms_2.json
```diff
@@ -687,15 +687,311 @@
                         }],
                         "routing_rules": [{
                             "goto": {
-                                "block": "country-of-birth"
+                                "block": "country-of-birth-scotland",
```
jonnyshaw89: Would it be more concise / appropriate to use `skip_conditions` for this scenario.
As it should remove the amount of `routing_rules`

jonnyshaw89 (changes requested):

data/en/lms_2.json
bitdivision: Yeah, makes sense

bitdivision: Fixed now. Looks much better.

jonnyshaw89 (approved):
Tidier 👍

dcdarrell9 (approved):
👍

PR #1787, opened 2018-09-13
Add address fields to lms schema

insacuri:
https://trello.com/c/iugfWgRR/2392-add-address-fields-to-lms-schema

The individual address fields needed to be passed in, but addresses being what they are, can have empty fields.  This PR adds those fields and allows metadata to be labelled as optional also.

### How to review 
Open the lms_2.json schema in launcher.  Remove the value of any of the 5 individual address fields.  It should launch the survey fine.  Every other field not labelled with optional: true should still fail validation if it's empty.

### Checklist

### DIFF
--- a/app/storage/metadata_parser.py
+++ b/app/storage/metadata_parser.py
@@ -88,6 +88,10 @@ def _validate_metadata_is_present(metadata, required_metadata):
     :param required_metadata:
     """
     for metadata_field in required_metadata:
+        # If the field is optional then skip validation on field
+        if metadata_field.get('optional') is True:
+            continue
+
         name = metadata_field['name']
         if name == 'trad_as_or_ru_name':
             # either of 'trad_as' or 'ru_name' is required
--- a/data/en/lms_2.json
+++ b/data/en/lms_2.json
@@ -29,6 +29,31 @@
             "validator": "string"
         },
         {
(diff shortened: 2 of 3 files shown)


bitdivision (changes requested):
I don't think it's strictly necessary, but I think adding some optional fields to an integration test would be nice.

Additionally, this will need a change to validator.

app/storage/metadata_parser.py
```diff
@@ -88,6 +88,10 @@ def _validate_metadata_is_present(metadata, required_metadata):
     :param required_metadata:
     """
     for metadata_field in required_metadata:
```
jonnyshaw89, line 90: technically this is not correctly named now as its no longer a list of `required_metadata`

data/en/lms_2.json
```diff
@@ -27,6 +27,35 @@
         {
             "name": "country",
             "validator": "string"
+        },
+        {
+            "name": "address_line1",
```
jonnyshaw89, line 32: Im confused as to why we are defining these in metadata if we are not using them?

jonnyshaw89 (changes requested):

bitdivision:
Closing until further work as discussed.

PR #1788, opened 2018-09-14
Fix safe_content regex for jinja filters in titles

bitdivision:
@dcdarrell9 noticed that some titles on LMS2 were including jinja filters.

The regex was not matching anything that contained curly braces (which we use for date offsets).

### How to review 
Test LMS2 and make sure reference date filters etc. are not shown in titles.

### Checklist

### DIFF
--- a/app/templating/template_renderer.py
+++ b/app/templating/template_renderer.py
@@ -61,7 +61,7 @@ class TemplateRenderer:
         """
         if content is not None:
             # Replace piping with ellipsis
-            content = re.sub(r'{{[^}]+}}', '…', content)
+            content = re.sub(r'{{.*?}}', '…', content)
             # Strip HTML Tags
             content = re.sub(r'</?[^>]+>', '', content)
         return content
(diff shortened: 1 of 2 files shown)


dcdarrell9 (approved):

app/templating/template_renderer.py
```diff
@@ -61,7 +61,7 @@ def safe_content(content):
         """
         if content is not None:
             # Replace piping with ellipsis
-            content = re.sub(r'{{[^}]+}}', '…', content)
+            content = re.sub(r'{{.*?}}', '…', content)
```
jonnyshaw89, line 64: Testy McTest Face?

jonnyshaw89 (changes requested):

app/templating/template_renderer.py
bitdivision: Donezy McDone Face.

bitdivision:
I guess the only thing this won't handle currently is nested dictionaries in a filter argument e.g.

`{{ some_jinja_filter(arg, {'a': 'test', 'b': {1:1}}) }}`

I can't really see how to handle that without building a stack, which I don't think we can do in a regex.

jonnyshaw89 (approved):

PR #1789, opened 2018-09-14
Display repeating groups in summary

### LINKED ISSUE #1765: Totalizer type uses values from an earlier persons answers if value is missing
### Expected behaviour
If we are in a repeating group , if an answer was not entered for an answer in a specific iteration then its value should not appear in the totalised answer

### Actual behaviour
In this case the repeat is over people. If answers regarding hours worked are entered for person A , but not person B then the totaliser displays and uses the answers from person A

LMS_2 . Enter two people. For first enter 2 jobs and enter worked overtime. When asked enter hours actual and overtime. Totaliser question shows correct answer. For second person enter 1 job and no overtime. Totaliser page shows data from person A for those answers not given by person B

### Screenshot
In this case this is person B who worked no overtime and had one job. The overtime answers were from person A
screen shot 2018-09-05 at 10 45 08

dwyeradam:
Allows for all answers for each repeating groups to be shown on summary pages. 
On calculated summaries it ensures the answers are from the correct repeat (Fixes #1765).

### How to review 
Test existing schemas with repeating groups. 
New schema `test_repeat_until_summaries.json` includes a repeat until that has section/calculated and final summaries within.

### Checklist

### DIFF
--- a/app/helpers/schema_helpers.py
+++ b/app/helpers/schema_helpers.py
@@ -49,8 +49,9 @@ def get_group_instance_id(schema, answer_store, location, answer_instance=0):
         existing_answers = answer_store.filter(answer_ids=block_answer_ids, answer_instance=answer_instance)
 
     # If there are existing answers with a group_instance_id
-    if existing_answers and [x for x in existing_answers if x.get('group_instance_id')]:
-        return existing_answers[0]['group_instance_id']
+    existing_answers_with_group_instance_id = [answer for answer in existing_answers if answer.get('group_instance_id')]
+    if existing_answers_with_group_instance_id:
+        return existing_answers_with_group_instance_id[0]['group_instance_id']
 
     return str(uuid4())
 
--- a/app/templates/partials/summary/question-single-answer.html
+++ b/app/templates/partials/summary/question-single-answer.html
@@ -4,7 +4,7 @@
         {% set is_textarea = answer_is_type(answer, 'textarea') %}
 
         <div class="grid__col col-12@xs mars {{ 'col-10@m' if is_textarea else 'col-6@m' }}">
-            {% if content.summary.summary_type == 'CalculatedSummary' and answer.label %}
+            {% if (answer.type == 'relationship' or content.summary.summary_type == 'CalculatedSummary') and answer.label %}
                 {%- include theme(['partials/summary/display-answer-label.html']) -%}
             {% else %}
                 {%- include theme(['partials/summary/display-question.html']) -%}
--- a/app/templates/partials/summary/summary.html
+++ b/app/templates/partials/summary/summary.html
@@ -5,9 +5,9 @@
         {%- for group in content.summary.groups -%}
             {%- if group.title -%}
                 <h2 class="summary__title neptune" id="{{ group.id }}">{{ group.title }}</h2>
+                <div class="summary__block u-mb-l">
             {%- endif -%}
 
-            <div class="summary__block u-mb-l">
                 {%- for block in group.blocks -%}
                     {%- for question in block.questions -%}
 
@@ -24,7 +24,9 @@
                 {%- if content.summary.summary_type == 'CalculatedSummary' -%}
                     {%- include theme(['partials/summary/calculated-summary-answer.html']) -%}
                 {%- endif -%}
-            </div>
+            {%- if loop.last or loop.nextitem.title-%}
+                </div>
+            {%- endif -%}
         {%- endfor -%}
     {%- endif -%}
 </div>
--- a/app/templating/summary/answer.py
+++ b/app/templating/summary/answer.py
@@ -1,7 +1,7 @@
 class Answer:
 
-    def __init__(self, answer_schema, answer, child_answer_value=None):
-        self.id = answer_schema['id']
+    def __init__(self, answer_schema, answer, group_instance, child_answer_value=None):
+        self.id = answer_schema['id'] + '-' + str(group_instance)
         self.label = answer_schema.get('label')
         self.value = answer
         self.type = answer_schema['type'].lower()
--- a/app/templating/summary/group.py
+++ b/app/templating/summary/group.py
@@ -1,29 +1,63 @@
+from copy import deepcopy
+
 from app.templating.summary.block import Block
+from app.helpers.schema_helpers import get_group_instance_id
+from app.questionnaire.location import Location
+from app.templating.template_renderer import renderer
 
 
 class Group:
 
-    def __init__(self, group_schema, path, answer_store, metadata, schema):
-        self.id = group_schema['id']
-        self.title = group_schema.get('title')
-        self.blocks = self._build_blocks(group_schema, path, answer_store, metadata, schema)
+    def __init__(self, group_schema, path, answer_store, metadata, schema, group_instance, schema_context):
+        self.id = group_schema['id'] + '-' + str(group_instance)
+        self.schema_context = schema_context
+
+        location = Location(group_schema['id'], group_instance, group_schema['blocks'][0]['id'])
+        self.group_instance_id = get_group_instance_id(schema, answer_store, location)
+
+        self.title = self._get_title(group_schema, schema, answer_store, group_instance)
+
+        self.blocks = self._build_blocks(group_schema, path, answer_store, metadata, schema, group_instance)
 
     @staticmethod
-    def _build_blocks(group_schema, path, answer_store, metadata, schema):
+    def _build_blocks(group_schema, path, answer_store, metadata, schema, group_instance):
         blocks = []
 
-        block_ids_on_path = [location.block_id for location in path]
+        block_ids_on_path = [location.block_id for location in path if location.group_id == group_schema['id'] and location.group_instance == group_instance]
 
         for block in group_schema['blocks']:
             if block['id'] in block_ids_on_path and \
                     block['type'] == 'Question':
-                blocks.extend([Block(block, group_schema['id'], answer_store, metadata, schema).serialize()])
+                blocks.extend([Block(block, group_schema['id'], answer_store, metadata, schema, group_instance).serialize()])
 
         return blocks
 
+    def _get_title(self, group_schema, schema, answer_store, group_instance):
+        section = schema.get_section(schema.get_group(group_schema['id'])['parent_id'])
+        answer_values = []
+        title_answer_ids = section.get('title_from_answers', [])
+
+        for answer_id in title_answer_ids:
+            for answer in answer_store.filter(answer_ids=[answer_id],
+                                              group_instance_id=self.group_instance_id).escaped():
+                if answer['value']:
+                    answer_values.append(answer['value'])
+
+        if answer_values:
+            return ' '.join(answer_values)
+
+        if group_instance == 0:
+            return group_schema.get('title')
+
     def serialize(self):
-        return {
-            'id': self.id,
-            'title': self.title,
-            'blocks': self.blocks,
-        }
+        schema_context_with_group_instance_id = deepcopy(self.schema_context)
+        schema_context_with_group_instance_id['group_instance_id'] = self.group_instance_id
+
+        return renderer.render(
+            {
+                'id': self.id,
+                'title': self.title,
+                'blocks': self.blocks,
+            },
+            **schema_context_with_group_instance_id
+        )
--- a/app/templating/summary/question.py
+++ b/app/templating/summary/question.py
@@ -1,63 +1,81 @@
 import collections
 
+from app.forms.household_relationship_form import build_relationship_choices
 from app.templating.summary.answer import Answer
 from app.templating.utils import get_question_title
 
 
 class Question:
 
-    def __init__(self, question_schema, answer_store, metadata, schema):
-        self.id = question_schema['id']
+    def __init__(self, question_schema, answer_store, metadata, schema, group_instance):
+        self.id = question_schema['id'] + '-' + str(group_instance)
         self.type = question_schema['type']
-
+        self.schema = schema
         answer_schemas = iter(question_schema['answers'])
 
-        # Using group instance as 0 for now as summary rendering context only has knowledge of current locations group instance (i.e. 0)
-        self.title = get_question_title(question_schema, answer_store, schema, metadata, group_instance=0) or question_schema['answers'][0]['label']
+        self.title = (get_question_title(question_schema, answer_store, schema, metadata, group_instance=group_instance) or
+                      question_schema['answers'][0]['label'])
         self.number = question_schema.get('number', None)
-        self.answers = self._build_answers(answer_store, question_schema, answer_schemas)
+        self.answers = self._build_answers(answer_store, question_schema, answer_schemas, group_instance)
 
     @staticmethod
-    def _get_answers(answer_store, answer_id):
-        return answer_store.filter(answer_ids=[answer_id]).escaped().values()
+    def _get_answers(answer_store, answer_id, group_instance):
+        answers = answer_store.filter(answer_ids=[answer_id], group_instance=group_instance).values()
 
-    def _get_answer(self, answer_store, answer_id):
-        try:
-            return self._get_answers(answer_store, answer_id)[0]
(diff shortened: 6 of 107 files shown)


bitdivision:
Not sure if you want comments on this yet, but it looks like if I add three people, I get an extra (empty) summary block at the bottom.

summary - test repeat until

dwyeradam:
@bitdivision I'm always open to comments but it's still a work in progress. I noticed this yesterday and I think I know what to fix.

dcdarrell9:
- Added 26 members and it auto cuts off 👍 
- 26th member doesn't get asked the value question, but shows in the summary 👎 
- Previous link doesn't show on first question for each individual (seems right to me?)

app/templating/summary_context.py
```diff
@@ -4,7 +4,7 @@
 from app.templating.summary.group import Group
 
 
-def build_summary_rendering_context(schema, sections, answer_store, metadata):
+def build_summary_rendering_context(schema, sections, answer_store, metadata, schema_context):
```
bitdivision, line 7: Update docstring

app/templating/summary/question.py
```diff
@@ -1,63 +1,84 @@
...
         self.number = question_schema.get('number', None)
-        self.answers = self._build_answers(answer_store, question_schema, answer_schemas)
+        self.answers = self._build_answers(answer_store, question_schema, answer_schemas, group_instance)
 
     @staticmethod
-    def _get_answers(answer_store, answer_id):
-        return answer_store.filter(answer_ids=[answer_id]).escaped().values()
+    def _get_answers(answer_store, answer_id, group_instance):
+        answers = answer_store.filter(answer_ids=[answer_id], group_instance=group_instance).values()
+        if not answers:
```
bitdivision: Extremely minor, could be `return answers or [None]`

app/helpers/schema_helpers.py
```diff
@@ -49,8 +49,9 @@ def get_group_instance_id(schema, answer_store, location, answer_instance=0):
         existing_answers = answer_store.filter(answer_ids=block_answer_ids, answer_instance=answer_instance)
 
     # If there are existing answers with a group_instance_id
-    if existing_answers and [x for x in existing_answers if x.get('group_instance_id')]:
-        return existing_answers[0]['group_instance_id']
+    existing_answers_with_group_instance_id = [x for x in existing_answers if x.get('group_instance_id')]
```
jonnyshaw89: `x`?

bitdivision:
I'll start updating the rest of the repeating group functional tests to account for the summaries.

app/helpers/schema_helpers.py
insacuri: Feel free to ignore this, but could you change x to something more meaningful?

app/templating/summary/answer.py
```diff
@@ -1,7 +1,7 @@
 class Answer:
 
-    def __init__(self, answer_schema, answer, child_answer_value=None):
-        self.id = answer_schema['id']
+    def __init__(self, answer_schema, answer, group_instance, child_answer_value=None):
+        self.id = answer_schema['id'] + '-' + str(group_instance)
```
insacuri, line 4: Maybe a silly question, but can group instance ever be None or ''?  If so, are we okay with an id of `answer_id-`

jonnyshaw89: `group_instance` defaults to 0

app/templating/summary/question.py
```diff
@@ -1,63 +1,82 @@
...
+                      question_schema['answers'][0]['label'])
         self.number = question_schema.get('number', None)
-        self.answers = self._build_answers(answer_store, question_schema, answer_schemas)
+        self.answers = self._build_answers(answer_store, question_schema, answer_schemas, group_instance)
 
     @staticmethod
-    def _get_answers(answer_store, answer_id):
-        return answer_store.filter(answer_ids=[answer_id]).escaped().values()
+    def _get_answers(answer_store, answer_id, group_instance):
+        answers = answer_store.filter(answer_ids=[answer_id], group_instance=group_instance).values()
```
insacuri, line 23: Why don't we escape these anymore?

insacuri:
The `test_repeat_until` summary has the 'does anyone else live here?' question in the summary (so if you add 3 extra people, you see 3 instances of that question).  Is that an issue?

app/templating/summary/answer.py
bitdivision: I can't see an obvious path for `group_instance` to be None. If it is, then I think we're going to have issues elsewhere as well.

I double checked this by checking that group_instance was never set to None in the debugger while running most of the unit / integration tests.

app/templating/summary/question.py
bitdivision: I'm not sure...

Tests still pass with `escaped()` . I can't see why it was removed. @jonnyshaw89 Any ideas?

app/helpers/schema_helpers.py
bitdivision: Will do.

app/templating/summary/group.py
```diff
@@ -1,29 +1,63 @@
...
-        self.title = group_schema.get('title')
-        self.blocks = self._build_blocks(group_schema, path, answer_store, metadata, schema)
+    def __init__(self, group_schema, path, answer_store, metadata, schema, group_instance, schema_context):
+        self.id = group_schema['id'] + '-' + str(group_instance)
+        location = Location(group_schema['id'], group_instance, group_schema['blocks'][0]['id'])
+        self.schema_context = schema_context
+        self.schema_context['group_instance_id'] = get_group_instance_id(schema, answer_store, location)
+
+        self.open_summary_block = False
+        self.close_summary_block = False
```
jonnyshaw89: Are these 2 more display logic? rather than the model?

jonnyshaw89: We could use`{{loop.first}}` and `{{loop.last}}` in the templates?
http://jinja.pocoo.org/docs/2.10/templates/

jonnyshaw89 (changes requested):

app/templates/partials/summary/summary.html
```diff
@@ -7,7 +7,9 @@
                 <h2 class="summary__title neptune" id="{{ group.id }}">{{ group.title }}</h2>
             {%- endif -%}
 
-            <div class="summary__block u-mb-l">
+            {%- if group.open_summary_block -%}
```
insacuri: I think this should be changed (sorry for not including this comment in the first pass!).
It looked odd to me when I first saw it but didn't look into it too deeply.  But after looking, there isn't a single instance where one is true and the other is false.
Also, theres a chance of an open without a close which would produce invalid HTML.

Is there a reason for this that I'm missing?

insacuri (changes requested):

app/templating/summary/group.py
bitdivision: That's really nice!

app/templates/partials/summary/summary.html
bitdivision: There is one instance where they will be different.

https://github.com/ONSdigital/eq-survey-runner/pull/1789/files/8e821331540515439c1bcdf7e5d87f4c058450cb#diff-41b3143304060f6f5310f1cdb4d18153L7

If we use a single flag, then we get a double line:
summary - test repeat until 1
Versus the current functionality:
summary - test repeat until 2

I do agree that it's not great, will investigate if there's a more robust method.

app/templating/summary/group.py
bitdivision: I can't see an obvious way to replicate the logic in build_summary_rendering_context without changing the context drastically. Thoughts?

jonnyshaw89: Because we only want to close/open when the title changes?
`{{loop.changed(group.title)}}`
`True if previously called with a different value (or not called at all).`

app/templates/partials/summary/summary.html
```diff
@@ -24,7 +24,9 @@ <h2 class="summary__title neptune" id="{{ group.id }}">{{ group.title }}</h2>
                 {%- if content.summary.summary_type == 'CalculatedSummary' -%}
                     {%- include theme(['partials/summary/calculated-summary-answer.html']) -%}
                 {%- endif -%}
-            </div>
+            {%- if loop.last or loop.nextitem.title-%}
```
jonnyshaw89, line 27: 👍

jonnyshaw89 (dismissed):

app/templates/partials/summary/question-single-answer.html
```diff
@@ -4,7 +4,7 @@
         {% set is_textarea = answer_is_type(answer, 'textarea') %}
 
         <div class="grid__col col-12@xs mars {{ 'col-10@m' if is_textarea else 'col-6@m' }}">
-            {% if content.summary.summary_type == 'CalculatedSummary' and answer.label %}
+            {% if answer.type == 'relationship' or content.summary.summary_type == 'CalculatedSummary' and answer.label %}
```
ajmaddaford: I think `answer.type == 'relationship'` can be `True` and `answer.label` not set as the `and` takes precedence over the `or` (https://docs.python.org/3/reference/expressions.html#operator-precedence).

bitdivision:
@insacuri Could you take another look at this?

app/templating/summary/group.py
```diff
@@ -1,29 +1,57 @@
...
+            link_names = defaultdict(list)
+            for answer_id in title_answer_ids:
+                for answer in answer_store.filter(answer_ids=[answer_id]).escaped():
+                    if answer['value']:
+                        link_names[answer['group_instance_id']].append(answer['value'])
+
+            for link_name in link_names:
+                link_names[link_name] = ' '.join(link_names[link_name])
+
+            return link_names[self.schema_context['group_instance_id']]
```
ajmaddaford: If we only ever return the title for the given `group_instance_id`, can we simplify this by filtering the answer store by that?
```python
answer_store.filter(
    answer_ids=[answer_id], 
    group_instance_id=self.schema_context['group_instance_id'])
```

app/templating/summary/question.py
```diff
@@ -1,63 +1,82 @@
...
+    def __init__(self, question_schema, answer_store, metadata, schema, group_instance):
+        self.id = question_schema['id'] + '-' + str(group_instance)
         self.type = question_schema['type']
-
+        self.schema = schema
         answer_schemas = iter(question_schema['answers'])
 
         # Using group instance as 0 for now as summary rendering context only has knowledge of current locations group instance (i.e. 0)
-        self.title = get_question_title(question_schema, answer_store, schema, metadata, group_instance=0) or question_schema['answers'][0]['label']
+        self.title = (get_question_title(question_schema, answer_store, schema, metadata, group_instance=group_instance) or
```
ajmaddaford, line 16: The comment is not needed anymore

app/templating/summary/question.py
```diff
@@ -1,63 +1,82 @@
...
+
+        repeat_rule = group['routing_rules'][0]['repeat']
+        if 'answer_ids' in repeat_rule:
+            for answer_id in repeat_rule['answer_ids']:
+                answer_ids.append(answer_id)
+        if 'answer_id' in repeat_rule:
+            answer_ids.append(repeat_rule['answer_id'])
+
+        choice = build_relationship_choices(answer_ids, answer_store, group_instance, member_label)[index]
+        label = answer_label % dict(current_person=choice[0], other_person=choice[1])
```
insacuri, line 62: I won't block the PR on this, but is there a reason we're doing string format using reasonably obscure operators?  I had to look up what % did in this context because I'd not come across it before

insacuri (approved):

insacuri:
Just needs merging and squashing then it'll be good to go!

app/templating/summary/group.py
```diff
@@ -1,29 +1,57 @@
...
 
-    def __init__(self, group_schema, path, answer_store, metadata, schema):
-        self.id = group_schema['id']
-        self.title = group_schema.get('title')
-        self.blocks = self._build_blocks(group_schema, path, answer_store, metadata, schema)
+    def __init__(self, group_schema, path, answer_store, metadata, schema, group_instance, schema_context):
+        self.id = group_schema['id'] + '-' + str(group_instance)
+        location = Location(group_schema['id'], group_instance, group_schema['blocks'][0]['id'])
+        self.schema_context = schema_context
+        self.schema_context['group_instance_id'] = get_group_instance_id(schema, answer_store, location)
```
ajmaddaford: This use of the schema context seems wrong. The group instance id on the schema context is the group_instance_id of the current location. Where is this overridden group_instance_id being used?

app/templating/summary_context.py
```diff
@@ -25,8 +26,18 @@ def build_summary_rendering_context(schema, sections, answer_store, metadata):
...
 
     for group in itertools.chain.from_iterable(group_lists):
-        if group['id'] in group_ids_on_path \
-                and schema.group_has_questions(group['id']):
-            groups.extend([Group(group, path, answer_store, metadata, schema).serialize()])
+        if (group['id'] in group_ids_on_path and schema.group_has_questions(group['id'])):
+            no_of_repeats = _number_of_repeats(group, path)
+            repeating_groups = []
+            for instance_idx in range(0, no_of_repeats):
+                summary_group = Group(group, path, answer_store, metadata, schema, instance_idx, schema_context).serialize()
```
ajmaddaford, line 33: Won't this break if a repeating group is removed?

ajmaddaford (changes requested):

app/templating/summary_context.py
bitdivision: It will break. I guess the question is whether we want to address this now, or do it when we start implementing deleting repeating groups.
There are going to be a lot of changes once we start removing them, and it probably depends on the removal of `group_instance`. That will presumably entail creating reusable patterns for this kind of stuff.

I'm tempted to leave it for now, but happy to discuss.

app/templating/summary/group.py
bitdivision: As discussed, I've removed the modification of schema_context here since it should refer to current location.

Since `schema_context` is passed into render, that group_instance_id will also represent the current location. It's not currently used though.

app/templating/summary_context.py
ajmaddaford: Fair point. I've raised a card in the backlog about migrating to using `group_instance_id` rather than `group_instance` when we're referring to a group of answers.

jonnyshaw89:
👍

ajmaddaford (approved):

app/templating/summary/group.py
bitdivision: @ajmaddaford I could have sworn I searched all the templates for 'group_instance_id' and didn't find any, but apparently not. My bad.

`group_instance_id` is used in the test_repeat_until_summary schema in a number of places. The most problematic of these is in the question titles. Of course currently, these work properly on the question itself, but don't display on the summary. 

The easy fix is to override group_instance_id in the call to render. Thoughts?

ajmaddaford: I thought I'd searched the templates before I made the comment as well! I agree, for now we should override it next to the render call. I'll raise a card for reviewing how we use the schema context.

jonnyshaw89:
We should add a test for the issue that was fixed

bitdivision:
If @jonnyshaw89 can bear the horror of `../summary/..` then I think this is ready.

bitdivision:
The final fix was to avoid rendering the section in the view-submission view, and just pass in the schema section directly.
This means that view-submission now shows the names in questions (tested in functional tests).

app/views/questionnaire.py
```diff
@@ -319,8 +319,8 @@ def get_view_submission(schema, eq_id, form_type):  # pylint: disable=unused-arg
             routing_path = PathFinder(schema, answer_store, metadata, []).get_full_routing_path()
 
             schema_context = _get_schema_context(routing_path, None, metadata, collection_metadata, answer_store, schema)
-            rendered_schema = renderer.render(schema.json, **schema_context)
```
bitdivision, line 322: This was removed since it is rendered further down.

PR #1791, opened 2018-09-18
Route around totaliser if only one value

bitdivision:
[Trello](https://trello.com/c/NJQXrMl2/2398-routing-issue-hours-worked-totaliser-is-showing-if-only-1-answer-entered)

We want to skip the totaliser if there is only one item to be shown. In this case, that means skipping if they only have one job and they have not worked any overtime.

Since the totaliser currently includes the routing around actual hours and usual hours, that needs to be duplicated onto the previous question and its confirmation question.

As mentioned by @jonnyshaw89 yesterday, this would be much nicer if we could use skip conditions instead of routing rules on the totaliser. I currently can't see a way to do this nicely since we have 4 complex paths coming out of the totaliser. If I've missed a better solution let me know.

The new routing rules are copied from the one job totaliser but include a guard on overtime being 'No overtime'.

~~Since some bugs were found with routing after the respondent answers that they did not work any days this week, I've updated the schema to comply with the new finalised spec. This means that if the respondent answers that they did not work this week, they will be asked overtime, and then taken directly to the why-fewer-hours question. This removes two extra blocks that were added to handle intermediate routing here.~~

### How to review 
Ensure that if you only have one job with no overtime, the totaliser is not displayed.

In any other situation, the totaliser should be displayed.

Ensure all questions between overtime and why-fewer-hours-than-usual are skipped if the respondent did not work any days that week.

I have fixed routing on YSTART18 (not-be-able-to-start-reason-nw6) and LKTIMA (how-long-have-you-been-looking-dur1) and CASWRK18 (not-working-casual-work-nw2).

The routing from EVEROT to YLESS has changed fairly significantly and should be double checked against the spec.

### Checklist

### DIFF
--- a/data/en/lms_2.json
+++ b/data/en/lms_2.json
@@ -705,6 +705,16 @@
                             }]
                         }],
                         "routing_rules": [{
+                                "goto": {
+                                    "block": "nationality",
+                                    "when": [{
+                                        "id": "dob-age-answer",
+                                        "condition": "not set"
+                                    }]
+                                }
+
+                            },
+                            {
 
                                 "goto": {
                                     "block": "nationality",
@@ -737,7 +747,14 @@
                                     }]
                                 }
 
+                            },
+                            {
+                                "goto": {
+                                    "block": "marital-status"
+                                }
+
                             }
+
                         ]
                     },
 
@@ -1557,6 +1574,11 @@
                                         "condition": "not set"
                                     }]
                                 }
+                            },
+                            {
+                                "goto": {
+                                    "block": "national-identity"
+                                }
                             }
                         ]
                     },
@@ -2153,6 +2175,17 @@
                                 }
                             }, {
                                 "goto": {
+                                    "block": "household-member-completed",
+                                    "when": [{
+                                        "id": "dob-age-answer",
+                                        "condition": "not set"
+                                    }, {
+                                        "id": "date-of-birth-answer",
+                                        "condition": "not set"
+                                    }]
+                                }
+                            }, {
+                                "goto": {
                                     "block": "religion",
                                     "when": [{
                                         "id": "date-of-birth-answer",
@@ -2257,6 +2290,17 @@
                                         "value": 16
                                     }]
                                 }
+                            }, {
+                                "goto": {
+                                    "block": "household-member-completed",
+                                    "when": [{
+                                        "id": "dob-age-answer",
+                                        "condition": "not set"
+                                    }, {
+                                        "id": "date-of-birth-answer",
+                                        "condition": "not set"
+                                    }]
+                                }
                             },
                             {
                                 "goto": {
@@ -2337,6 +2381,17 @@
                                         "id": "dob-age-answer",
                                         "condition": "less than",
                                         "value": 16
+                                    }]
+                                }
+                            }, {
+                                "goto": {
+                                    "block": "household-member-completed",
+                                    "when": [{
+                                        "id": "dob-age-answer",
+                                        "condition": "not set"
+                                    }, {
+                                        "id": "date-of-birth-answer",
+                                        "condition": "not set"
                                     }]
                                 }
                             },
@@ -2425,6 +2480,17 @@
                                         "value": 16
                                     }]
                                 }
+                            }, {
+                                "goto": {
+                                    "block": "household-member-completed",
+                                    "when": [{
+                                        "id": "dob-age-answer",
+                                        "condition": "not set"
+                                    }, {
+                                        "id": "date-of-birth-answer",
+                                        "condition": "not set"
+                                    }]
+                                }
                             },
                             {
                                 "goto": {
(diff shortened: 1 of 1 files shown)


jonnyshaw89 (dismissed):
Seems to work for me 👍

MebinAbraham (commented):
As discussed its broke.

bitdivision:
I've noticed that the complete tick appears on the wrong person in navigation as well while it's being completed.

bitdivision:
I think you found two problems @MebinAbraham 

1. When marking 'Did not work this week' a routing rule on usual hours kicks in to take you straight to the totaliser. I'll need to add some logic there too to avoid that.

2. Strange routing with two people. This was caused by the lack of a default routing rule on the totaliser I think. This meant that the routing_path was being cut off after the first totaliser, marking all later blocks as `INVALID` which are treated as complete. This was causing some really strange behaviour.

jonnyshaw89:
@bitdivision i thought there was logic in validator to stop you from not having a default routing rule, this should have caught that?

bitdivision:
Yeah, seems to be a problem there. I'll fix it.

bitdivision:
@jonnyshaw89 I've made a card to fix validator as it does require quite a few schema changes and should be done seperately. https://trello.com/c/DDwSA4qj/2414-schema-validator-does-not-validate-default-routing-rules

bitdivision:
The above commit should fix both the default route and the bug with "Did not work this week"

MebinAbraham (dismissed):
Right so as discussed again, the routing path seems to not be working as expected. One of the scenarios: Going to an Incomplete section when you click 'Change' link. You are manually able to access the location the Change link is trying to take you but not when it evaluates the routing path.

bitdivision:
@MebinAbraham This is all fixed in Adam's branch. I've confirmed that it works as expected with that code.

jonnyshaw89:
@bitdivision why is this blocked?

bitdivision:
Blocked since the point Mebbin raised in review can only be tested once Adam's branch has gone in. I don't think it's critical to testing the main functionality though, so I'll remove.

MebinAbraham (commented):
As discussed, primary job only > Did not work any days of the week > Worked overtime, displays totaliser with 1 total.

bitdivision:
Spec has now been finalised to skip everything after the overtime question if they answer that they did not work any days that week.

bitdivision:
Looks as though spec is going to change again regarding the above routing.

bitdivision:
I am now restoring the routing as per the spec.

bitdivision:
Just realised that we need to double up all of the 'employee' routing rules since we have to treat not set on employee / self employed as employee for routing.

bitdivision:
Only 410 lines added for that.

jonnyshaw89:
Seems to be ok given the combination of routes I took. but @MebinAbraham seems to be the pro at finding routing errors 👍

jonnyshaw89 (approved):

insacuri (dismissed):
Routing seems to work just fine, Tried a bunch of combinations (including seeing the totaliser, going back and saying no overtime, then going forward) and all of them were as I was expecting.

bitdivision:
The bug darrell found (causing a 500) should be fixed in https://github.com/ONSdigital/eq-survey-runner/pull/1808

insacuri:
@bitdivision I just found a bug, not sure if it's the same thing. First person had a job, for the second person I said I didn't have a job but I did have casual work, then it blew up.
I think it's probably going to be fixed by that other PR.

bitdivision:
Yep, that's the one that's fixed.

insacuri:
@bitdivision still broken.  Same as before (first person 2 jobs, second person casual job), it gets to the 'what type of work is it' screen.  If you select 'none of the above' it breaks.  Not sure if it'll do the same thing if you select any of the other options

MebinAbraham:
Completeness seems to be broke but hopefully will be fixed with the new changes you are making.

bitdivision:
Looks as though this was just a case of two questions being out of order.

MebinAbraham (commented):
```
Did you have a paid job, either as an employee or self-employed, in the week Monday 24 to Sunday 30 September 2018?
```
> No
```
Did you do any casual work for payment, even for an hour, in the week Monday 24 to Sunday 30 September 2018?
```
> Yes
```
On which of the following days did you do any work in your paid job or business?
```
> Did not work this week

Enter you `Actual Hours`, then it displays on the `CalculatedSummary` with 1 answer.

MebinAbraham (commented):
So as discussed, still some odd behaviour, with routing/completeness.
Waiting also answer from Donna regarding the routing for casual work + overtime.

dcdarrell9 (approved):

dcdarrell9 (changes requested):
"It does not matter how slowly you go as long as you do not stop." - Confucius

dcdarrell9 (approved):

PR #1792, opened 2018-09-20
Add years label to textfield and evaluate condition

dcdarrell9:
- Added years label to age question.
- Added condition for evaluation.

### How to review 
- Test lms routing is still consistent.

### Checklist

### DIFF
--- a/app/questionnaire/rules.py
+++ b/app/questionnaire/rules.py
@@ -294,6 +294,9 @@ def get_answer_store_value(answer_id, answer_store, schema, group_instance, grou
         raise Exception('Multiple answers ({:d}) found evaluating when rule for answer ({})'
                         .format(filtered.count(), answer_id))
 
+    if filtered.answers and isinstance(filtered.answers[0]['value'], dict):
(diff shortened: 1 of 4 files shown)


app/questionnaire/rules.py
```diff
@@ -294,6 +294,9 @@ def get_answer_store_value(answer_id, answer_store, schema, group_instance, grou
         raise Exception('Multiple answers ({:d}) found evaluating when rule for answer ({})'
                         .format(filtered.count(), answer_id))
 
+    if filtered.answers and isinstance(filtered.answers[0]['value'], dict):
+        return filtered.answers[0]['value']['years']
```
dcdarrell9: Not sure how/what way to make this more generic.. Thought about something like below
```
    if filtered.answers and isinstance(filtered.answers[0]['value'], dict):
        for value in filtered.answers[0]['value'].values():
            return value
```

dcdarrell9:
Being merged with 2415-add-hours as the change will be the same

PR #1794, opened 2018-09-21
Allow for long/short and single/plural labels for answer fields

insacuri:
https://trello.com/c/KYe8MSnc/2415-hours-unit-for-lms-should-be-hours-not-hrs

The 'hrs' field in LMS needs to be 'hours'.  In order to do this, we had to add extra metadata to the unit type so that length and plurality could be specified.

### How to review 
All the `test_unit_patterns.json` measurements. They should all be in short form apart from the time one, which should be long form (hours).

The time entry fields in `lms_2.json` have 'hours' in the label.
Test any/all of the following that they look as expected (short, singular form units)
`mbs_0203.json`
`mbs_0204.json`
`mbs_0253.json`
`test_calculated_summary.json`
`test_summary.json`
`test_view_submitted_response.json`

### Checklist

### DIFF
--- a/app/jinja_filters.py
+++ b/app/jinja_filters.py
@@ -82,8 +82,21 @@ def format_address_list(user_entered_address=None, metadata_address=None):
     return address_list
 
 
-def format_unit(unit, value=''):
-    return units.format_unit(value=value, measurement_unit=unit, length='short', locale=DEFAULT_LOCALE)
+def format_unit(unit, value, length='short'):
+    return units.format_unit(value=value, measurement_unit=unit, length=length, locale=DEFAULT_LOCALE)
+
+
+def format_unit_input_label(unit, unit_length='short'):
+    """
+    This function is used to only get the unit of measurement text.  If the unit_length
+    is long then only the plural form of the word is returned (e.g., Hours, Years, etc).
+
+    :param (str) unit unit of measurement
+    :param (str) unit_length length of unit text, can be one of short/long/narrow
+    """
+    if unit_length == 'long':
+        return units.format_unit(value=2, measurement_unit=unit, length=unit_length, locale=DEFAULT_LOCALE).replace('2 ', '')
+    return units.format_unit(value='', measurement_unit=unit, length=unit_length, locale=DEFAULT_LOCALE).strip()
 
 
 def format_duration(value):
@@ -471,6 +484,11 @@ def conditional_dates_check():
 @blueprint.app_context_processor
 def format_unit_processor():
     return dict(format_unit=format_unit)
+
+
+@blueprint.app_context_processor
+def format_unit_input_label_processor():
+    return dict(format_unit_input_label=format_unit_input_label)
 
 
 @blueprint.app_context_processor
--- a/data/en/lms_2.json
+++ b/data/en/lms_2.json
@@ -466,7 +466,9 @@
                             "type": "General",
                             "answers": [{
                                 "id": "dob-age-answer",
-                                "type": "Number",
+                                "unit": "duration-year",
+                                "type": "Unit",
+                                "unit_length": "long",
                                 "mandatory": false
                             }]
                         }],
@@ -3255,6 +3257,7 @@
                                 "mandatory": false,
                                 "type": "Unit",
                                 "unit": "duration-hour",
+                                "unit_length": "long",
                                 "label": "Usual hours"
                             }]
                         }],
@@ -3308,6 +3311,7 @@
                                 "mandatory": false,
                                 "type": "Unit",
                                 "unit": "duration-hour",
+                                "unit_length": "long",
                                 "label": "Usual hours"
                             }]
                         }]
@@ -3335,6 +3339,7 @@
                                 "mandatory": false,
                                 "type": "Unit",
                                 "unit": "duration-hour",
+                                "unit_length": "long",
                                 "label": "Usual hours"
                             }]
                         }],
@@ -3390,6 +3395,7 @@
                                 "mandatory": false,
                                 "type": "Unit",
                                 "unit": "duration-hour",
+                                "unit_length": "long",
                                 "label": "Unpaid overtime hours"
                             }]
                         }],
@@ -3440,6 +3446,7 @@
                                 "mandatory": false,
                                 "type": "Unit",
                                 "unit": "duration-hour",
+                                "unit_length": "long",
                                 "label": "Paid overtime hours"
                             }]
                         }]
@@ -3469,6 +3476,7 @@
                                 "mandatory": false,
                                 "type": "Unit",
                                 "unit": "duration-hour",
+                                "unit_length": "long",
                                 "label": "Usual hours"
                             }],
                             "guidance": {
@@ -3518,6 +3526,7 @@
                                 "mandatory": false,
                                 "type": "Unit",
                                 "unit": "duration-hour",
+                                "unit_length": "long",
                                 "label": "Actual hours"
                             }]
                         }],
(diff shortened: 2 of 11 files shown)


app/jinja_filters.py
```diff
@@ -82,8 +82,21 @@ def format_address_list(user_entered_address=None, metadata_address=None):
...
+    """
+    This function is used to only get the unit of measurement text for answer fields.
+
+    :param (str) unit unit of measurement
+    :param (str) length length of unit  can be one of short/long/narrow
+    :param (bool) use_plural_for_label if true, will return the plural version of the unit.
+    """
+    if use_plural_for_label:
+        return units.format_unit(value=2, measurement_unit=unit, length=length, locale=DEFAULT_LOCALE).replace('2 ', '')
+    return units.format_unit(value='', measurement_unit=unit, length=length, locale=DEFAULT_LOCALE).replace(' ', '', 1)
```
jonnyshaw89: Does the current implementation have a space before the unit then?

app/jinja_filters.py
```diff
@@ -82,8 +82,21 @@ def format_address_list(user_entered_address=None, metadata_address=None):
     return address_list
 
 
-def format_unit(unit, value=''):
-    return units.format_unit(value=value, measurement_unit=unit, length='short', locale=DEFAULT_LOCALE)
+def format_unit(unit, value='', length='short'):
+    return units.format_unit(value=value, measurement_unit=unit, length=length, locale=DEFAULT_LOCALE)
+
+
+def format_unit_input_label(unit, length='short', use_plural_for_label=False):
```
jonnyshaw89: Should we not default `use_plural_for_label` to `True`?

jonnyshaw89: and probably rename to `use_plural` given the context of label is in the function name?

data/en/lms_2.json
```diff
@@ -466,7 +466,10 @@
                             "type": "General",
                             "answers": [{
                                 "id": "dob-age-answer",
-                                "type": "Number",
+                                "unit": "duration-year",
+                                "type": "Unit",
+                                "length": "long",
+                                "use_plural_for_label": true,
```
jonnyshaw89: if default was true this would be less verbose

data/en/mbs_0203.json
```diff
@@ -151,6 +151,7 @@
                         "mandatory": true,
                         "q_code": "110",
                         "type": "Unit",
+                        "length": "short",
```
jonnyshaw89: The default is `short` do want to have to specify this on everything?
we need to be consistent, as for `use_plural_for_label` we are not specifying it everywhere but for `length` we are.

data/en/test_unit_patterns.json
```diff
@@ -30,35 +30,71 @@
...
+                        },
+                        {
+                            "id": "duration-year",
+                            "description": "",
+                            "label": "Years",
+                            "mandatory": false,
+                            "type": "Unit",
+                            "unit": "duration-year",
+                            "length": "long",
+                            "use_plural_for_label": true
```
jonnyshaw89: Do we have an example where this is set to false?

jonnyshaw89 (changes requested):

app/jinja_filters.py
insacuri: Yea, we just didn't notice it because it gets ignored when it's rendered.
There weren't any unit tests around using `format_unit` for labels before, but I've added them now so little surprises like this shouldn't happen again

app/jinja_filters.py
insacuri: To your first question, I don't know.  What are we more likely to want?  Everything so far has been singular til now so I assumed that should be the default, though it's not much effort to change it.

To your second point, that's a good point!

app/jinja_filters.py
jonnyshaw89: We should probably just use trim then?

app/jinja_filters.py
```diff
@@ -82,8 +82,21 @@ def format_address_list(user_entered_address=None, metadata_address=None):
     return address_list
 
 
-def format_unit(unit, value=''):
-    return units.format_unit(value=value, measurement_unit=unit, length='short', locale=DEFAULT_LOCALE)
+def format_unit(unit, value='', length='short'):
```
jonnyshaw89: If this is no longer used for labels, do we need the value default?

app/jinja_filters.py
insacuri: On second thought, I think the default should be short/singular, because of the overwhelming amount of stuff that has its labels in that form.  Do you reckon that'll change?  If not, then I think we should leave it as is

data/en/mbs_0203.json
insacuri: This was to fix a bug that I bumped into.  Originally, because it was short by default, I wasn't defining this.  It then rendered the unit name instead of the unit because in the jinja template, when resolving answer.length, it was passing through None or '' (I didn't check), but it was overriding the default.

Is there a way to get around this?

data/en/lms_2.json
insacuri: Then all the other places it's used, it would be just as verbose?

data/en/test_unit_patterns.json
insacuri: Added this just before I saw this message ^^

app/jinja_filters.py
insacuri: Agreed, removed it

jonnyshaw89 (changes requested):
A common theme that i have spotted is that when `length: long` then `plural` is always true.
I wonder if this will always be the case? as this can make things a lot cleaner in the schema.

insacuri:
@jonnyshaw89 what are you actually asking for here?  

It's like this because up until this point we've ONLY had the short/singular label to deal with.  Now we've got a combination of short/singular and long/plural.  The way this PR is structured, if/when we eventually have a long/singular or short/plural form need to accommodate for, it'll just work without having to do any extra work outside of tweaking the schema.

It looks like the schema is less clean because one of the keys is kinda long because it needs to describe a very specific thing.  But I'd rather it be present for all of them then only present when it's true/false (depending on how we evaluated it) as nowhere else do we have keys in the schema that are sometimes there (that I could see anyway, could be wrong on that).

jonnyshaw89:
@insacuri I think that i'm trying to question is, will there ever be a logical scenario when we would want to display the long version not as a plural?

insacuri:
@jonnyshaw89 Fair enough, who can answer that question for us?

jonnyshaw89:
@insacuri charlotte or donna i guess?

insacuri:
@jonnyshaw89 Talked to Will, he said there's basically no chance that we'll ever use the other 2 variations.  I'll get around to simplifying this.

bitdivision (approved):
👍

scripts/test_schemas.sh
```diff
@@ -7,7 +7,7 @@ fi
 
 if [ "$run_docker" == true ]; then
     docker pull onsdigital/eq-schema-validator
-    validator="$(docker run -d -p 5001:5000 onsdigital/eq-schema-validator)"
+    validator="$(docker run -d -p 5001:5000 onsdigital/eq-schema-validator:2415-hours-unit-for-lms-should-be-hours-not-hrs)"
```
bitdivision: reminder to remove.

jonnyshaw89 (approved):

jonnyshaw89:
@insacuri You need to remove the tag in `test_schemas.sh`

PR #1795, opened 2018-09-24
Fix translation issues

ajmaddaford:
Tidies up some existing translatable strings and adds the translate call to some places it was missing.

### How to review 
- Run the app and verify the amended screens work as before
- Follow the new translation instructions to verify they work


app/templates/introduction.html
```diff
@@ -20,20 +20,18 @@
...
             <div class="grid__col col-7@m">
 
-                <h2 class="saturn">{{ _("You are completing this for <span>%(ru_name)s</span>%(trad_as_open)s%(trad_as)s%(trad_as_close)s",
-                    ru_name = metadata.ru_name,
-                    trad_as_open = trad_as_open,
-                    trad_as = trading_as,
-                    trad_as_close = trad_as_close) }}
+                <h2 class="saturn">
+                    {% if metadata.trad_as %}
+                        {{ _("You are completing this for <span>%(ru_name)s</span> (%(trading_as_name)s)",
```
jonnyshaw89, line 28: we are loosing `<span class="trading_as">` with this. not sure if thats intentional?

app/templates/view-submission.html
```diff
@@ -5,24 +5,23 @@
...
+            {% if metadata.trad_as %}
+                {{ _("Your answers were submitted for <span>%(ru_name)s</span> (%(trading_as_name)s) on %(submitted_time)s",
+                ru_name = metadata.ru_name,
+                trading_as_name = metadata.trad_as,
+                submitted_time=metadata.submitted_time|format_datetime) }}
+            {% else %}
+                {{ _("Your answers were submitted for <span>%(ru_name)s</span> on %(submitted_time)s",
+                ru_name = metadata.ru_name,
+                submitted_time=metadata.submitted_time|format_datetime) }}
+            {% endif %}
```
jonnyshaw89, line 24: This logic is duplicated, should we make it common?

app/translations/messages.pot
```diff
@@ -598,715 +622,678 @@ msgstr ""
...
-msgid ""
-"When a user closes the browser or when their cookie\n"
-"                                            cache is cleared"
+#: app/templates/static/cookies-privacy.html:173
+#: app/templates/static/cookies-privacy.html:180
+#: app/templates/static/cookies-privacy.html:187
+#: app/templates/static/cookies-privacy.html:208
+#: app/templates/static/cookies-privacy.html:215
+#: app/templates/static/cookies-privacy.html:222
+msgid "When a user closes the browser or when their cookie cache is cleared"
```
jonnyshaw89, line 851: I wonder how this page will make sense with each paragraph translated out of context? maybe the page need to be translated as a whole?

jonnyshaw89 (changes requested):

README.md
```diff
@@ -231,6 +231,27 @@ To create the gaelic language files, use the following:
...
+```
+po2csv app/translations/cy/LC_MESSAGES/messages.po cy.csv
+```
+
+To convert back to a .po file:
+```
+csv2po cy.csv app/translations/cy/LC_MESSAGES/messages.po
+```
+
+*Important:* There are some encoding issues when opening the .csv file in Excel. For now, opening in Google sheets and saving as a .xslx file is the least work solution.
```
insacuri: 'is the least work solution' doesn't make much sense. Could you rephrase it?

app/jinja_filters.py
```diff
@@ -153,11 +153,15 @@ def format_date_custom(context, value, date_format='EEEE d MMMM YYYY'):
...
-def format_datetime(context, value, date_format="d MMMM YYYY 'at' HH:mm"):
+def format_datetime(context, value):
 
     london_date_time = datetime.strptime(value, '%Y-%m-%dT%H:%M:%S.%f')
-    result = "<span class='date'>{date}</span>".format(date=flask_babel.format_datetime(london_date_time,
-                                                                                        format=date_format))
+    formatted_date = flask_babel.format_date(london_date_time, format='d MMMM YYYY')
+    formatted_time = flask_babel.format_time(london_date_time, format='HH:mm')
+
+    result = "<span class='date'>{date}</span>".format(
```
insacuri: Not sure if this is in the scope of this PR, but hardcoding "_date_ at _time_" won't translate.  
It's not enough to block this PR because it's how it was beforehand, but this should probably be addressed somewhere

app/templates/introduction.html
ajmaddaford: It was. It's not used by any tests and doesn't appear in any css or javascript.

app/templates/view-submission.html
ajmaddaford: I did think about it, but went with http://wiki.c2.com/?ThreeStrikesAndYouRefactor

app/translations/messages.pot
ajmaddaford: I agree, but that's not possible with our current translation approach. Also, the lines will all appear together in the csv/spreadsheet that's generated from the .pot file. We've sent that file to the translators to get some feedback on if it will work for them.

README.md
ajmaddaford: Updated

jonnyshaw89 (approved):

insacuri (approved):

PR #1796, opened 2018-09-24
Enforce expiry on eq_session

MebinAbraham:
We do not need to keep all the session data in DynamoDB. If a session hasn't been updated for a set period the session is no longer valid and can be deleted from our table. Currently session is not updated per request hence all users are signed out after the preset 45 minutes.

Issues being addressed:
- `updated_at` time not being updated hence session times out after 45 minutes (current timeout length), regardless of whether you were active or not.
- browser cookie never expires - current expiry time `(1969-12-31T23:59:59.000Z)`

This PR introduces a new column `expires_at` which is used to determine whether a user has a valid session. This column has been introduced to both Postgres and Dynamo to keep the data consistent. For Postgres, a DB migration is required, which will be done through another PR. 

Once the `expires_at` time has passed, the user will be signed out when making the next request, or automatically via the timeout modal (JS) feature. This value is also used to configure [TTL](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/TTL.html) within DynamoDB, which will delete the data within 48hrs of expiry.

---

To test DynamoDB, provision Dynamo using eq-terraform using the branch `eq-session-ttl-enabled`. Make sure you change the env variable in runner to match your environment eg 
`dev-eq-session` - > `<name>-eq-session`
For context here are my current env settings: https://gist.github.com/MebinAbraham/11160b4daa9163aec1b65ee9cdc46d9f
Toggle: `EQ_SESSION_DYNAMO_WRITE=True` to test between Dynamo/SQLite

To test the changes for Postgres (SQLite locally), you need to alter the table manually. Use contents from PR #1797 to add a new column in SQLite.
`cat database-scripts/pr-1797-apply.sql | sqlite3 /tmp/questionnaire.db`
or just remove the tmp db. `rm /tmp/questionnaire.db`

You can test the expiry by using the `test_timeout.json` schema or by changing these in `settings.py`:
```
EQ_SESSION_TIMEOUT_SECONDS = int(os.getenv('EQ_SESSION_TIMEOUT_SECONDS', str(45 * 60)))
EQ_SESSION_TIMEOUT_GRACE_PERIOD_SECONDS = int(os.getenv('EQ_SESSION_TIMEOUT_GRACE_PERIOD_SECONDS', '5'))
EQ_SESSION_TIMEOUT_PROMPT_SECONDS = int(os.getenv('EQ_SESSION_TIMEOUT_PROMPT_SECONDS', '5'))
```

### How to review 
Test all scenarios that you can think of.
- Ensure user is signed out when their session has expired.
- Ensure the `updated_at` and `expires_at` time is modified with each request.
- Use chrome dev tool to inspect the session cookie to ensure the expiry time is as expected. `(Inspect Element > Application > Cookies > http://localhost:5000 > session > Expires/Max-Age)`

### Expected behaviour
- All new sessions are given an expiration time.
- Once the expiration time has passed, the user is signed out.
- In-flight sessions will not be given an expiration time. (This can be discussed, currently, it won't since the `expires_at` is retrieved from the cookie).
- All requests should be backwards compatible.

Note: I have still have some test to write for the different scenarios but wanted to get some eyes on this.

### Checklist

### DIFF
--- a/app/authentication/authenticator.py
+++ b/app/authentication/authenticator.py
@@ -1,6 +1,8 @@
+from datetime import datetime, timedelta
 from uuid import uuid4
 
 from blinker import ANY
+from dateutil.tz import tzutc
 from flask import session as cookie_session, current_app
 from flask_login import LoginManager, user_logged_out, login_user
 from sdc.crypto.decrypter import decrypt
@@ -11,7 +13,6 @@ from app.authentication.user import User
 from app.data_model.session_data import SessionData
 from app.globals import get_questionnaire_store, get_session_store, create_session_store
 from app.keys import KEY_PURPOSE_AUTHENTICATION
-
 from app.settings import EQ_SESSION_ID, USER_IK
 
 logger = get_logger()
@@ -40,14 +41,43 @@ def when_user_logged_out(sender_app, user):  # pylint: disable=unused-argument
     cookie_session.pop(USER_IK, None)
 
 
+def _extend_session_expiry(session_store):
+    """
+    Extends the expiration time of the session
+    :param session_store:
+    """
+    session_timeout = cookie_session.get('expires_in')
+    if session_timeout:
+        session_store.expiration_time = datetime.now(tz=tzutc()) + timedelta(seconds=session_timeout)
+        session_store.save()
+
+        logger.debug('session expiry extended')
+
+
+def _is_session_valid(session_store):
+    """
+    Checks that the user's session has not expired
+    :param session_store:
+    :return: True if the session is valid else False
+    """
+
+    if session_store.expiration_time and \
+            session_store.expiration_time < datetime.now(tz=tzutc()):
+        return False
+
+    return True
+
+
 def load_user():
     """
     Checks for the present of the JWT in the users sessions
     :return: A user object if a JWT token is available in the session
     """
     session_store = get_session_store()
-    if session_store and session_store.user_id:
+
+    if session_store and session_store.user_id and _is_session_valid(session_store):
         logger.debug('session exists')
+
         user_id = session_store.user_id
         user_ik = cookie_session.get(USER_IK)
         user = User(user_id, user_ik)
@@ -55,9 +85,13 @@ def load_user():
         if session_store.session_data.tx_id:
             logger.bind(tx_id=session_store.session_data.tx_id)
 
+        _extend_session_expiry(session_store)
+
         return user
 
     logger.info('session does not exist')
+
+    cookie_session.pop(USER_IK, None)
     return None
 
 
--- a/app/data_model/app_models.py
+++ b/app/data_model/app_models.py
@@ -1,4 +1,4 @@
-import datetime
+from datetime import datetime
 
 from dateutil.tz import tzutc
 from marshmallow import Schema, fields, post_load, pre_dump
@@ -9,17 +9,24 @@ class QuestionnaireState:
         self.user_id = user_id
         self.state_data = state_data
         self.version = version
-        self.created_at = datetime.datetime.now(tz=tzutc())
-        self.updated_at = datetime.datetime.now(tz=tzutc())
+        self.created_at = datetime.now(tz=tzutc())
+        self.updated_at = datetime.now(tz=tzutc())
 
 
 class EQSession:
-    def __init__(self, eq_session_id, user_id, session_data=None):
+    def __init__(self, eq_session_id, user_id, session_data=None, expires_at=None):
         self.eq_session_id = eq_session_id
         self.user_id = user_id
         self.session_data = session_data
-        self.created_at = datetime.datetime.now(tz=tzutc())
-        self.updated_at = datetime.datetime.now(tz=tzutc())
+        self.created_at = datetime.now(tz=tzutc())
+        self.updated_at = datetime.now(tz=tzutc())
+        self.expires_at = expires_at
+
+        # Needed only when data is read from Postgres.
+        # The Timestamp() class already handles this when reading from Dynamo.
+        # Can be removed once only Dynamo is used.
+        if expires_at:
+            self.expires_at = expires_at.replace(tzinfo=tzutc())
 
 
 class UsedJtiClaim:
@@ -39,10 +46,14 @@ class SubmittedResponse:
 # pylint: disable=no-self-use
 class Timestamp(fields.Field):
     def _serialize(self, value, attr, obj):
-        return int(value.strftime('%s'))
+        if value:
+            # Timezone aware datetime to timestamp
+            return int(value.replace(tzinfo=tzutc()).strftime('%s'))
 
     def _deserialize(self, value, attr, data):
-        return datetime.datetime.utcfromtimestamp(value).replace(tzinfo=tzutc())
+        if value:
+            # Timestamp to timezone aware datetime
+            return datetime.utcfromtimestamp(value).replace(tzinfo=tzutc())
 
 
 class DateTimeSchemaMixin:
@@ -51,7 +62,7 @@ class DateTimeSchemaMixin:
 
     @pre_dump
     def set_date(self, data):
-        data.updated_at = datetime.datetime.now(tz=tzutc())
+        data.updated_at = datetime.now(tz=tzutc())
         return data
 
 
@@ -74,6 +85,7 @@ class EQSessionSchema(Schema, DateTimeSchemaMixin):
     eq_session_id = fields.Str()
     user_id = fields.Str()
     session_data = fields.Str()
+    expires_at = Timestamp(allow_none=True)  # To cater in flight data (Should never actually be None)
 
     @post_load
     def make_model(self, data):
--- a/app/data_model/models.py
+++ b/app/data_model/models.py
@@ -2,6 +2,7 @@ import datetime
 
 from flask_sqlalchemy import SQLAlchemy
 from structlog import get_logger
+
 from app.data_model import app_models
 
 logger = get_logger()
@@ -41,21 +42,23 @@ class EQSession(db.Model):
     session_data = db.Column('session_data', db.String)
     created_at = db.Column('created_at', db.DateTime, default=datetime.datetime.utcnow)
     updated_at = db.Column('updated_at', db.DateTime, default=datetime.datetime.utcnow, onupdate=datetime.datetime.utcnow)
+    expires_at = db.Column('expires_at', db.DateTime)
 
-    def __init__(self, eq_session_id, user_id, session_data=None):
+    def __init__(self, eq_session_id, user_id, session_data=None, expires_at=None):
         self.eq_session_id = eq_session_id
         self.user_id = user_id
         self.session_data = session_data
+        self.expires_at = expires_at
 
     def to_app_model(self):
-        model = app_models.EQSession(self.eq_session_id, self.user_id, self.session_data)
+        model = app_models.EQSession(self.eq_session_id, self.user_id, self.session_data, self.expires_at)
         model.created_at = self.created_at
         model.updated_at = self.updated_at
         return model
 
     @classmethod
     def from_app_model(cls, model):
-        return cls(model.eq_session_id, model.user_id, model.session_data)
+        return cls(model.eq_session_id, model.user_id, model.session_data, model.expires_at)
 
 
 # pylint: disable=maybe-no-member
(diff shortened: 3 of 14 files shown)


app/authentication/authenticator.py
```diff
@@ -40,23 +41,57 @@ def when_user_logged_out(sender_app, user):  # pylint: disable=unused-argument
...
 
 
+def _extend_session_expiry(session_store):
+    session_timeout = cookie_session.get('expires_at')
+    if session_store and session_store.expiration_time and session_timeout:
+
+        session_store.expiration_time = datetime.utcnow() + timedelta(seconds=session_timeout)
+        session_store.save()
+
+        logger.info('session expiry extended')
```
jonnyshaw89: This is going to get a bit loggy?

app/helpers/template_helper.py
```diff
@@ -38,6 +33,10 @@ def with_metadata_context(func):
     @wraps(wraps)
     def metadata_context_wrapper(*args, **kwargs):
         metadata = get_metadata(current_user)
+
+        if not metadata:
+            raise Unauthorized
```
jonnyshaw89: Is this not handled elsewhere?
can you get this far without having metadata?

app/settings.py
```diff
@@ -51,9 +51,9 @@ def get_env_or_fail(key):
 EQ_RABBITMQ_QUEUE_NAME = os.getenv('EQ_RABBITMQ_QUEUE_NAME', 'submit_q')
 EQ_RABBITMQ_ENABLED = parse_mode(os.getenv('EQ_RABBITMQ_ENABLED', 'True'))
 EQ_NEW_RELIC_CONFIG_FILE = os.getenv('EQ_NEW_RELIC_CONFIG_FILE', './newrelic.ini')
-EQ_SESSION_TIMEOUT_SECONDS = int(os.getenv('EQ_SESSION_TIMEOUT_SECONDS', str(45 * 60)))
-EQ_SESSION_TIMEOUT_GRACE_PERIOD_SECONDS = int(os.getenv('EQ_SESSION_TIMEOUT_GRACE_PERIOD_SECONDS', '30'))
-EQ_SESSION_TIMEOUT_PROMPT_SECONDS = int(os.getenv('EQ_SESSION_TIMEOUT_PROMPT_SECONDS', '120'))
+EQ_SESSION_TIMEOUT_SECONDS = int(os.getenv('EQ_SESSION_TIMEOUT_SECONDS', str(10)))
+EQ_SESSION_TIMEOUT_GRACE_PERIOD_SECONDS = int(os.getenv('EQ_SESSION_TIMEOUT_GRACE_PERIOD_SECONDS', '5'))
+EQ_SESSION_TIMEOUT_PROMPT_SECONDS = int(os.getenv('EQ_SESSION_TIMEOUT_PROMPT_SECONDS', '5'))
```
jonnyshaw89: Are these changes just you testing?

app/views/questionnaire.py
```diff
@@ -454,6 +449,16 @@ def _store_viewable_submission(answers, metadata, submitted_time):
...
+#
+# def update_session_expiry(eq_session_id):
+#     """
+#     Extends the session expiry time
+#     """
+#     _eq_session = data_access.get_by_key(EQSession, eq_session_id)
+#     _eq_session.expires_at = datetime.utcnow() + timedelta(seconds=get_session_timeout_in_seconds(g.schema))
+#
+#     data_access.put(_eq_session)
+
```
jonnyshaw89: This is elsewhere now so can go?

jonnyshaw89 (changes requested):

app/settings.py
MebinAbraham: Yes, will revert before.

app/helpers/template_helper.py
MebinAbraham: I added this in when going to feedback page without a session threw a 500, but it seems to magically work now. I'll remove.

app/settings.py
jonnyshaw89: why not now?

app/authentication/authenticator.py
```diff
@@ -40,23 +41,57 @@ def when_user_logged_out(sender_app, user):  # pylint: disable=unused-argument
...
         user = User(user_id, user_ik)
 
         if session_store.session_data.tx_id:
             logger.bind(tx_id=session_store.session_data.tx_id)
 
+        _extend_session_expiry(session_store)
+
         return user
 
+    if session_store:
```
bitdivision: You're checking this in `_is_valid_session` too. 

I don't know why, but I found the control flow here a little bit difficult to grok. Personally I would go for:
```
if not session_store:
    return None

if _is_valid_session(session_store) and session_store.user_id:
    ...
    return user
else:
    session_store.delete()
    ...
```

bitdivision (commented):
Haven't done any testing yet, but code looks good so far.

app/authentication/authenticator.py
MebinAbraham: Fair point. Updated. Didn't use `else` because of linting ` Unnecessary "else" after "return" (no-else-return)`.

app/data_model/app_models.py
```diff
@@ -1,4 +1,4 @@
-import datetime
+from datetime import datetime
```
insacuri, line 1: 👍

app/data_model/session_store.py
```diff
@@ -1,5 +1,5 @@
-from structlog import get_logger
 import simplejson as json
+from structlog import get_logger
```
insacuri, line 2: Not used now so this can be removed

insacuri: Ignore that sorry! I thought this was being added and I couldn't see any new log lines so I thought it wasn't being used.

insacuri:
Code looks good but I've not tested it yet.

jonnyshaw89:
Ive tested expiring and deleting the token from the DB and it kicks me out 👍

MebinAbraham:
The expiry and updated time can also be seen updating in Dynamo/SQLite when you run the functional tests.

MebinAbraham:
Okay I broke it for in-flight sessions for Dynamo. 

Edit: This should never happen but added a fix anyway.

app/data_model/models.py
```diff
@@ -40,22 +42,25 @@ class EQSession(db.Model):
     user_id = db.Column('user_id', db.String, primary_key=True)
     session_data = db.Column('session_data', db.String)
     created_at = db.Column('created_at', db.DateTime, default=datetime.datetime.utcnow)
-    updated_at = db.Column('updated_at', db.DateTime, default=datetime.datetime.utcnow, onupdate=datetime.datetime.utcnow)
+    updated_at = db.Column('updated_at', db.DateTime)
```
jonnyshaw89: How does the `updated_at` get updated now?

bitdivision: In `__init__()`

jonnyshaw89: so we set updated when we load from the DB?

bitdivision: Indeed. I was a little skeptical too.

MebinAbraham: Trying to keep it similar, but I can also do `on_update` since `expires_at` would be updated.

jonnyshaw89 (approved):

app/globals.py
```diff
@@ -42,12 +44,35 @@ def get_session_store():
...
+    if with_grace_period:
+        grace_period = current_app.config['EQ_SESSION_TIMEOUT_GRACE_PERIOD_SECONDS']
+        default_session_timeout += grace_period
+
+        if schema_session_timeout:
+            schema_session_timeout += grace_period
+
+    return schema_session_timeout if schema_session_timeout and \
+                                     schema_session_timeout < default_session_timeout else \
+                                     default_session_timeout
```
ajmaddaford: It's confusing to add the grace period in two places, this would read more easily as:

```
default_session_timeout = current_app.config['EQ_SESSION_TIMEOUT_SECONDS']
schema_session_timeout = schema.json.get('session_timeout_in_seconds')
timeout = schema_session_timeout if schema_session_timeout and \
          schema_session_timeout < default_session_timeout else \
          default_session_timeout

if with_grace_period:
    grace_period = current_app.config['EQ_SESSION_TIMEOUT_GRACE_PERIOD_SECONDS']
    timeout += grace_period

return timeout
```

app/authentication/authenticator.py
```diff
@@ -40,24 +41,53 @@ def when_user_logged_out(sender_app, user):  # pylint: disable=unused-argument
...
+        session_store.save()
+
+        logger.debug('session expiry extended')
+
+
+def _is_valid_session(session_store):
+    """
+    checks that the user has a valid session
+    """
+    if session_store and session_store.user_id:
```
ajmaddaford: When is `user_id` not set on `session_store`?

app/authentication/authenticator.py
```diff
@@ -40,24 +41,53 @@ def when_user_logged_out(sender_app, user):  # pylint: disable=unused-argument
     cookie_session.pop(USER_IK, None)
 
 
+def _extend_session_expiry(session_store):
+    session_timeout = cookie_session.get('expires_in')
+    if session_store and session_store.expiration_time and session_timeout:
```
ajmaddaford: Is the check for `session_store.expiration_time` only required whilst we have in flight sessions? If so it may be worth highlighting it so we know to remove it.

app/data_model/app_models.py
```diff
@@ -9,17 +9,21 @@ def __init__(self, user_id, state_data, version):
...
 class EQSession:
-    def __init__(self, eq_session_id, user_id, session_data=None):
+    def __init__(self, eq_session_id, user_id, session_data=None, expires_at=None):
         self.eq_session_id = eq_session_id
         self.user_id = user_id
         self.session_data = session_data
-        self.created_at = datetime.datetime.now(tz=tzutc())
-        self.updated_at = datetime.datetime.now(tz=tzutc())
+        self.created_at = datetime.now(tzutc())
+        self.updated_at = datetime.now(tzutc())
```
ajmaddaford: These are inconsistent with the `created_at` and `updated_at` above that use `tz=tzutc()`

tests/app/data_model/test_app_models.py
```diff
@@ -16,14 +15,17 @@ def test_submitted_response(self):
         self._test_model(SubmittedResponse('txid', 'somedata', NOW))
 
     def test_questionnaire_state(self):
-        new = self._test_model(QuestionnaireState('someuser', 'somedata', 1))
-        self.assertGreaterEqual(new.created_at, NOW)
-        self.assertGreaterEqual(new.updated_at, NOW)
+        new_model = self._test_model(QuestionnaireState('someuser', 'somedata', 1))
+
+        self.assertGreaterEqual(new_model.created_at, NOW)
```
bitdivision, line 20: We do have `freezegun` in the dependencies now, which you could use here. Not particularly important in this case, but worth bearing in mind if you're writing these kinds of tests.

bitdivision:
I've tested the basics, and I'm happy with the code. 
I'm still getting weird terraform errors, which seem to be particular to my env. I'll try to resolve those tomorrow and get this merged.

dcdarrell9 (commented):
LGTM 👍  Had no issues

ajmaddaford (changes requested):

app/authentication/authenticator.py
MebinAbraham: Yes it is. Will add a comment.

app/data_model/app_models.py
MebinAbraham: Will make it consistent.

app/globals.py
MebinAbraham: Sweet, will change.

MebinAbraham:
@ajmaddaford Have made the changes and ready for re-review.
@jonnyshaw89 Could you make sure you are still happy with the new changes.
Thanks

tests/app/data_model/test_app_models.py
MebinAbraham: Will have a look into it. Don't see why we have it in the dependencies since we don't seem to use it anywhere currently?.

bitdivision: We use it in `test_date_reference_ranges.py`

bitdivision: https://github.com/ONSdigital/eq-survey-runner/blob/master/tests/integration/mci/test_date_reference_ranges.py#L4

ajmaddaford (approved):
Looks good. Tested in both Postgres and DynamoDB in AWS.

PR #1798, opened 2018-09-25
Repeat rule check if answer instance is on path

jonnyshaw89:
Previously it was only checking if an answer id was on the routing path. Now we check if that instance of an answer is on the path

### How to review 
run `test_routing_repeat_until.json` add 2 additional people then press the browser back and say no additional people to essentially orphan the last person in the answer store.
In the next group it should not ask the sex of the orphaned person.

### Checklist

### DIFF
--- a/app/questionnaire/rules.py
+++ b/app/questionnaire/rules.py
@@ -3,6 +3,8 @@ import re
 from datetime import datetime
 from dateutil.relativedelta import relativedelta
 
+from app.questionnaire.location import Location
+
 MAX_REPEATS = 25
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


app/questionnaire/rules.py
```diff
@@ -164,6 +169,14 @@ def evaluate_repeat(schema, repeat_rule, answer_store, answer_ids_on_path):
 
         answers = list(answer_store.filter(answer_ids=repeat_indexes))
 
+        # Get answer block_id
+        # Use answer group_instance
+        # is block_id / group_instance on routing path?
```
insacuri: This block of comments still needed?  If so, putting them above the line that does the action would be more useful

insacuri (approved):
Tested it and works as expected.  Code looks fine 👍

MebinAbraham (changes requested):
As discussed, if you add multiple people, ie 4, then you navigate back using the browser and answer `No` to the driver question for the eg 2nd person, it repeats the `driver` question for the other people you added and keeps their instances in the submission. Hope this makes sense.

jonnyshaw89:
@MebinAbraham i believe that issue is fixed now.
It was a similar case of repeats on the number of answers and not checking those answers are still on the path

MebinAbraham (approved):
Seem to be working as expected. 👍

PR #1802, opened 2018-09-28
Ensure not set and set conditions work on empty lists

bitdivision:
LMS does routing based on a MutuallyExclusive answer not being set.

MutuallyExclusive uses SelectMultipleField from wtforms, and that will return an empty list if no options are select.

The `not set` and `set` rules currently compare to `None` which won't work for these questions.

We could alternatively override the SelectMultipleField from wtforms, but checking for an empty list seems better for the moment.

### How to review 

test_routing_checkbox_set_not_set will skip interstitials after each question depending on whether you answer the question.

### Checklist

### DIFF
--- a/app/questionnaire/rules.py
+++ b/app/questionnaire/rules.py
@@ -73,9 +73,9 @@ def evaluate_condition(condition, answer_value, match_value):
     elif condition == 'not contains' and isinstance(answer_value, list) and match_value not in answer_value:
         result = True
     elif condition == 'set':
-        result = answer_value is not None
+        result = answer_value is not None and answer_value != []
     elif condition == 'not set':
-        result = answer_value is None
+        result = answer_value is None or answer_value == []
     elif condition == 'greater than' and answer_and_match and answer_value > match_value:
         result = True
     elif condition == 'less than' and answer_and_match and answer_value < match_value:
--- a/tests/functional/spec/features/routing/checkbox_set_not_set_routing.spec.js
+++ b/tests/functional/spec/features/routing/checkbox_set_not_set_routing.spec.js
@@ -1,0 +1,53 @@
+const helpers = require('../../../helpers');
+
+const ToppingCheckboxPage = require('../../../pages/features/routing/set_not_set/checkbox/topping-checkbox.page.js');
+const ToppingInterstitialNotSetPage = require('../../../pages/features/routing/set_not_set/checkbox/topping-interstitial-not-set.page.js');
+const ToppingInterstitialSetPage = require('../../../pages/features/routing/set_not_set/checkbox/topping-interstitial-set.page.js');
+const OptionalMutuallyExclusivePage = require('../../../pages/features/routing/set_not_set/checkbox/optional-mutually-exclusive.page.js');
+const CheeseInterstitialNotSetPage = require('../../../pages/features/routing/set_not_set/checkbox/cheese-interstitial-not-set.page.js');
+const CheeseInterstitialSetPage = require('../../../pages/features/routing/set_not_set/checkbox/cheese-interstitial-set.page.js');
+const SummaryPage = require('../../../pages/features/routing/set_not_set/checkbox/summary.page.js');
+
+describe('Test routing using not set and set conditions on checkboxes', function() {
+
+  beforeEach(function() {
+    return helpers.openQuestionnaire('test_routing_checkbox_set_not_set.json');
+  });
+
+  it('Given a user sets a topping and a cheese, they should see an interstitial for each saying that they were set', function() {
+      return browser
+        .click(ToppingCheckboxPage.cheese())
+        .click(ToppingCheckboxPage.submit())
+
+        .getUrl().should.eventually.contain(ToppingInterstitialSetPage.pageName)
+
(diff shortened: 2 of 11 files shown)


tests/functional/spec/components/checkbox-mutually-exclusive.spec.js
```diff
@@ -108,4 +116,29 @@ describe('Component: Mutually Exclusive Checkbox', function() {
     });
   });
 
+  describe('Given the user does selects any options on the optional page', function() {
```
ajmaddaford: Should this be "Given the user selects an option on the optional page"?

dcdarrell9:
So, when I select one of the answers in the second question, the interstitial is shown, and if i choose 'no cheese' the interstitial is skipped, but on the summary page it looks a bit weird.
screen shot 2018-09-28 at 13 01 16

bitdivision:
Cheers @dcdarrell9 . I'll take a look now

insacuri:
I noticed the same thing, the submit buttons are skewed to the side

data/en/test_checkbox_mutually_exclusive.json
```diff
@@ -20,61 +20,101 @@
...
-            }, {
-                "type": "Summary",
-                "id": "summary"
-            }],
+                    "title": "Optional question"
+                },
+                {
+                    "id": "cheese-interstitial",
+                    "title": "Cheese Interstitial Page",
+                    "description": "You Didn't select an option",
```
insacuri: This description should probably be updated.  I had to look at the schema to figure out what this page was actually testing

bitdivision:
Looks like the summary layout is broken in master too. I have a funny feeling this was my fault.

bitdivision:
Sorry about the bad description @insacuri . That confused me for a minute too.

I've created another PR to fix the alignment on summary.

app/questionnaire/rules.py
```diff
@@ -71,9 +71,9 @@ def evaluate_condition(condition, answer_value, match_value):
     elif condition == 'not contains' and isinstance(answer_value, list) and match_value not in answer_value:
         result = True
     elif condition == 'set':
-        result = answer_value is not None
+        result = answer_value is not None and answer_value != []
     elif condition == 'not set':
-        result = answer_value is None
+        result = answer_value is None or answer_value == []
```
ajmaddaford, line 78: I think this is fine for now, but we need to look into why we're storing an empty list for unanswered checkboxes. I've raised a card in the backlog to investigate.

tests/functional/spec/components/checkbox-mutually-exclusive.spec.js
```diff
@@ -108,4 +116,29 @@ describe('Component: Mutually Exclusive Checkbox', function() {
...
+    it('They should not see the interstitial page', function() {
+      return browser
+        .click(MandatoryCheckboxPage.noExtra())
+        .isSelected(MandatoryCheckboxPage.noExtra()).should.eventually.be.true
+        .click(MandatoryCheckboxPage.submit())
+        .click(OptionalCheckboxPage.submit())
+        .getUrl().should.eventually.contain(SummaryPage.pageName)
+    });
+  });
+
```
ajmaddaford: Is the behaviour we're testing specific to a mutually exclusive checkbox? If not, I don't think it should be added to the mutually exclusive schema and associated tests.

ajmaddaford (changes requested):

tests/functional/spec/components/checkbox-mutually-exclusive.spec.js
bitdivision: Fair point, I'll move it to a seperate schema.

app/questionnaire/rules.py
bitdivision: :+1:

tests/functional/spec/components/checkbox-mutually-exclusive.spec.js
bitdivision: I've moved it to a test_routing_checkbox_set_not_set.json schema

ajmaddaford (approved):

tests/functional/spec/features/routing/checkbox_set_not_set_routing.spec.js
```diff
@@ -0,0 +1,53 @@
...
+        .click(OptionalMutuallyExclusivePage.submit())
+
+        .getUrl().should.eventually.contain(CheeseInterstitialSetPage.pageName)
+
+        .click(CheeseInterstitialSetPage.submit())
+
+        .getUrl().should.eventually.contain(SummaryPage.pageName);
+  });
+
+  it('Given a user does not set a topping and does not set a cheese, they should see an interstitial for each saying that they were not set', function() {
```
MebinAbraham, line 36: `does not set a cheese` :satisfied:
Maybe just `Given a user does not select an answer`?. Don't mind.

MebinAbraham (changes requested):
Need to change your `data_version` in your schema to `0.0.2`. `0.0.1` requires `q_code` hence breaks on submission. Everything else looks good 👍

MebinAbraham (approved):

PR #1804, opened 2018-10-01
Apply address playback to lms schema

dcdarrell9:
- Applied address playback to lms.

### How to review 
- Check that spec matches.

- [Schema validator pr](https://github.com/ONSdigital/eq-schema-validator/pull/92)
- [Go launcher pr](https://github.com/ONSdigital/go-launch-a-survey/pull/118)
### Checklist

### DIFF
--- a/app/storage/metadata_parser.py
+++ b/app/storage/metadata_parser.py
@@ -16,14 +16,21 @@ def uuid_4_parser(plain_string):
 
 
 def boolean_parser(boolean_value):
-    """Strict matching of boolean values from JWT claims.
-    :param metadata:
-    :raises: TypeError - when a non boolean value has been loaded from the JWT.
-    :returns: A boolean value
-    """
     if not isinstance(boolean_value, bool):
         raise TypeError('Claim was not of type `bool`')
     return boolean_value
+
+
+def string_parser(string_value):
+    if not isinstance(string_value, str) or string_value == '':
+        raise TypeError('Claim not a valid string value')
+    return string_value
+
+
+def optional_string_parser(optional_string_value):
+    if not isinstance(optional_string_value, str):
+        raise TypeError('Claim not a valid/empty string')
+    return optional_string_value
 
 
 def clean_leading_trailing_spaces(metadata):
@@ -38,7 +45,8 @@ VALIDATORS = {
     'date': iso_8601_date_parser,
     'uuid': uuid_4_parser,
     'boolean': boolean_parser,
-    'string': lambda *args: None,
+    'string': string_parser,
+    'optional_string': optional_string_parser,
 }
 
 MANDATORY_METADATA = [
@@ -77,30 +85,7 @@ def parse_runner_claims(claims):
 
 
 def validate_metadata(claims, required_metadata):
-    _validate_metadata_is_present(claims, required_metadata)
     _validate_metadata_values_are_valid(claims, required_metadata)
-
-
-def _validate_metadata_is_present(metadata, required_metadata):
-    """
-    Validate that JWT claims contain the required metadata (both eq-runner and survey specific required metadata).
-    :param metadata:
-    :param required_metadata:
-    """
-    for metadata_field in required_metadata:
-        name = metadata_field['name']
-        if name == 'trad_as_or_ru_name':
-            # either of 'trad_as' or 'ru_name' is required
-            valid = bool(metadata.get('trad_as') or metadata.get('ru_name'))
-        else:
-            # Validate that the value is one of:
-            # a) A boolean value
-            # b) A non empty value
-            value = metadata.get(name)
-            valid = isinstance(value, bool) or bool(value)
-
-        if not valid:
-            raise InvalidTokenException('Missing key/value for {}'.format(name))
 
 
 def _validate_metadata_values_are_valid(claims, required_metadata):
@@ -114,13 +99,17 @@ def _validate_metadata_values_are_valid(claims, required_metadata):
         for metadata_field in required_metadata:
             name = metadata_field['name']
             claim = claims.get(name)
-            if claim:
-                logger.debug('parsing metadata', key=name, value=claim)
-                VALIDATORS[metadata_field['validator']](claim)
+            if name == 'trad_as_or_ru_name':
+                claim = claims.get('ru_name') or claims.get('trad_as')
+            elif name not in claims:
+                raise InvalidTokenException('Missing required key {} from claims'.format(name))
+
+            logger.debug('parsing metadata', key=name, value=claim)
+            VALIDATORS[metadata_field['validator']](claim)
 
     except (RuntimeError, ValueError, TypeError) as error:
         logger.error('Unable to parse metadata', key=name, value=claim, exc_info=error)
-        raise InvalidTokenException('incorrect data in token') from error
+        raise InvalidTokenException('incorrect data in token for {}'.format(name)) from error
     except KeyError as key_error:
         error_msg = 'Invalid validator for schema metadata - {}'.format(key_error.args[0])
         logger.error(error_msg, exc_info=key_error)
--- a/data/en/lms_2.json
+++ b/data/en/lms_2.json
@@ -21,11 +21,31 @@
             "validator": "string"
         },
         {
-            "name": "ru_name",
+            "name": "country",
             "validator": "string"
         },
         {
-            "name": "country",
+            "name": "address_line1",
+            "validator": "optional_string"
+        },
+        {
+            "name": "address_line2",
+            "validator": "optional_string"
+        },
+        {
+            "name": "locality",
+            "validator": "optional_string"
+        },
+        {
+            "name": "town_name",
+            "validator": "optional_string"
+        },
+        {
+            "name": "postcode",
+            "validator": "optional_string"
+        },
+        {
+            "name": "display_address",
             "validator": "string"
         }
     ],
@@ -38,7 +58,7 @@
                     "blocks": [{
                             "type": "Interstitial",
                             "id": "about-household",
-                            "title": "About you and people who live at {{ metadata['ru_name'] }}",
+                            "title": "About you and people who live at {{ metadata['display_address'] }}",
                             "description": "In this section, we’re going to ask about you, and the people that live with you.",
                             "content": [{
                                 "title": "Information you need",
@@ -51,16 +71,15 @@
                         {
                             "type": "Question",
                             "id": "address-check-block",
-                            "title": "Your address:",
-                            "description": "{{metadata['ru_name']}}",
+                            "description": "",
                             "questions": [{
                                 "id": "address-check-question",
-                                "title": "Is the address above your main residence?",
-                                "description": "",
+                                "title": "Is this address your main residence?",
+                                "description": "{{ format_address_list ([metadata['address_line1'], metadata['address_line2'], metadata['locality'], metadata['town_name'], metadata['postcode']]) }}",
                                 "type": "General",
                                 "answers": [{
                                     "id": "address-check-answer",
-                                    "mandatory": false,
+                                    "mandatory": true,
                                     "options": [{
                                             "label": "Yes",
                                             "value": "Yes",
@@ -104,7 +123,7 @@
                         {
                             "type": "Question",
                             "id": "address-type-check-block",
-                            "title": "You said {{ metadata['ru_name'] }} is not your main residence",
+                            "title": "You said {{ metadata['display_address'] }} is not your main residence",
                             "description": "",
                             "questions": [{
                                 "id": "address-type-check-question",
@@ -112,7 +131,7 @@
                                 "type": "General",
                                 "answers": [{
                                         "id": "address-type-check-answer",
-                                        "mandatory": false,
+                                        "mandatory": true,
                                         "type": "Radio",
                                         "options": [{
                                                 "label": "A previous address from which mail has been redirected",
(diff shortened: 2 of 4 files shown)


app/storage/metadata_parser.py
```diff
@@ -88,6 +88,10 @@ def _validate_metadata_is_present(metadata, required_metadata):
     :param required_metadata:
     """
     for metadata_field in required_metadata:
+        # If the field is optional then skip validation on field
+        if metadata_field.get('optional') is True:
```
MebinAbraham: No need for `is True`. Prob out of scope but could do validation on things such as postcode etc. Can't quite remember, but there was conversations arround address validator?.

app/storage/metadata_parser.py
```diff
@@ -88,6 +88,10 @@ def _validate_metadata_is_present(metadata, required_metadata):
     :param required_metadata:
     """
     for metadata_field in required_metadata:
```
jonnyshaw89: These technically are no longer `required_metadata`?

jonnyshaw89 (changes requested):

app/storage/metadata_parser.py
dcdarrell9: `metadata` and `required_metadata` get passed through to us, so maybe we should look at doing something with optional/required? Or should we simply just be renaming?

app/storage/metadata_parser.py
dcdarrell9: May be, but will have a look 👍

scripts/test_schemas.sh
```diff
@@ -7,7 +7,7 @@ fi
 
 if [ "$run_docker" == true ]; then
     docker pull onsdigital/eq-schema-validator
-    validator="$(docker run -d -p 5001:5000 onsdigital/eq-schema-validator)"
+    validator="$(docker run -d -p 5001:5000 onsdigital/eq-schema-validator:eq-2392-add-default-values-for-address-playback-metadata)"
```
MebinAbraham: Reminder to revert.

app/storage/metadata_parser.py
jonnyshaw89: `schema_metadata` maybe?

jonnyshaw89: i think we need to look at the naming of a lot of the params in this file

tests/app/parser/test_metadata_parser.py
```diff
@@ -308,6 +308,26 @@ def test_invalid_schema_metadata(self):
 
         self.assertEqual('Invalid validator for schema metadata - invalidValidator', ite.exception.args[0])
 
+    def test_optional_metadata(self):
+        metadata = {
```
MebinAbraham: `claims` or `metadata`? 😂

dcdarrell9: 🤷‍♂️

tests/app/parser/test_metadata_parser.py
```diff
@@ -308,6 +308,26 @@ def test_invalid_schema_metadata(self):
...
+            'ru_ref': '2016-04-04',
+            'period_str': 'May 2016',
+            'case_id': str(uuid.uuid4())
+        }
+
+        self.schema_metadata.append({'name': 'address_line1', 'validator': 'string', 'optional': True})
+        validate_metadata(metadata, self.schema_metadata)
+
+        metadata['address_line1'] = '123 awesome place'
+        validate_metadata(metadata, self.schema_metadata)
```
insacuri: You're not actually asserting anything in this whole test.  Probably want to actually test something :P

insacuri (changes requested):

app/storage/metadata_parser.py
```diff
@@ -81,22 +81,24 @@ def validate_metadata(claims, required_metadata):
     _validate_metadata_values_are_valid(claims, required_metadata)
 
 
-def _validate_metadata_is_present(metadata, required_metadata):
+def _validate_metadata_is_present(claims, required_metadata):
```
insacuri: I think for this PR it's probably worth leaving the names the same until we can make them all consistent as seeing claims and metadata in the same function is confusing

tests/app/parser/test_metadata_parser.py
dcdarrell9: Is this a bad time to say this was the one you wrote 👀

insacuri: I looked at the file again, theres another test that does it in the same way.  Feel free to leave it as is, but I think validate_metadata returns a True/False, might be a simple as adding assertTrue/False to it

MebinAbraham: It not raising an exception is the test.
You could just add an optional metadata into `self.schema_metadata`. Then it would be tested as part of:
https://github.com/ONSdigital/eq-survey-runner/blob/496d70d4cc9e1c9fc3f9348098edd10486c3289e/tests/app/parser/test_metadata_parser.py#L94-L96

app/storage/metadata_parser.py
```diff
@@ -81,22 +81,24 @@ def validate_metadata(claims, required_metadata):
...
     """
     Validate that JWT claims contain the required metadata (both eq-runner and survey specific required metadata).
-    :param metadata:
+    :param claims:
     :param required_metadata:
     """
+    # Remove optional metadata from required_metadata
+    required_metadata = [field for field in required_metadata if 'optional' not in field]
```
MebinAbraham: This technically isn't the best as it allows you to override the `MANDATORY_METADATA` I think.
If you are keeping this comprehension it should be something like this?
```
required_metadata = [field
  for field in required_metadata
  if field not in MANDATORY_METADATA and
  if 'optional' not in field
]
```

dcdarrell9: Would `MANDATORY_METADATA` ever have optional fields? Then they wouldn't be mandatory

tests/app/parser/test_metadata_parser.py
dcdarrell9: Good shout 👍

app/storage/metadata_parser.py
MebinAbraham: Forgot that mandatory metadata was checked, then the schema metadata, not together. No mandatory wouldn't have optional. Realised the code above wouldn't actually work for mandatory metadata.

So fine as it is. 👍 
Not sure if it is better to do `if then continue` instead of looping twice essentially. 🤷‍♂️

jonnyshaw89: @dcdarrell9 is this change still required given they always pass it?

dcdarrell9: Yeah, this is checking that we have a value for `required_metadata`, so the optional ones may not be there which would give us the `missing key/value` exception

insacuri:
Just tested it.  If you provide all/some of the address fields, it looks and functions correctly.
If you provide none of the fields, it goes bang on the page where the individual fields are used

insacuri (approved):
Tested it and it works fine except for the situation where you get no data.  They shouldn't passing us address fields where they're all empty so I'm okay approving it as there's a card in the backlog about what to do in this situation.

tests/app/parser/test_metadata_parser.py
MebinAbraham: You don't need this test?. The behaviour is already checked as long as you add an optional metadata into `self`.

MebinAbraham (dismissed):
Working as expected. As discussed other changes can be done part of the new card raised.

app/storage/metadata_parser.py
jonnyshaw89: Im thinking maybe the optional flag is no longer relevant? given the key is always present. if they are passing an empty string then the string validator should still work? if they are not passing a value, maybe we need a new validator that allows no value?

dcdarrell9: That's a good point actually, so what I can work out from RH is that we should get all keys, but some with a null value ...
```
"attributes": {
    "ADDRESS_LINE1": "Tredegar House",
    "ADDRESS_LINE2": null,
    "LOCALITY": null,
    "TOWN_NAME": "Newport",
    "POSTCODE": "NV184QG"
  }
```

app/storage/metadata_parser.py
```diff
@@ -39,6 +39,7 @@ def clean_leading_trailing_spaces(metadata):
     'uuid': uuid_4_parser,
     'boolean': boolean_parser,
     'string': lambda *args: None,
+    'optional_string': lambda *args: None,
```
bitdivision: I don't like this. It's not immediately obvious what's happening. I would move to a seperate def with a reasonable name.

tests/app/parser/test_metadata_parser.py
```diff
@@ -134,13 +142,13 @@ def test_missing_required_metadata_user_id_in_token(self):
         with self.assertRaises(InvalidTokenException) as ite_key:
             validate_metadata(metadata, self.schema_metadata)
 
-        self.assertEqual('Missing key/value for user_id', str(ite_key.exception))
+        self.assertEqual('incorrect data in token', str(ite_key.exception))
```
jonnyshaw89: It would be good to keep the existing error message as it was very helpful the other day when trying to see which item of the claims was preventing RH from launching

app/storage/metadata_parser.py
```diff
@@ -16,16 +16,23 @@ def uuid_4_parser(plain_string):
...
-    :raises: TypeError - when a non boolean value has been loaded from the JWT.
-    :returns: A boolean value
-    """
     if not isinstance(boolean_value, bool):
         raise TypeError('Claim was not of type `bool`')
     return boolean_value
 
 
+def string_parser(string_value):
+    if not string_value:
```
jonnyshaw89: This doesn't validate that it is also a string

app/storage/metadata_parser.py
```diff
@@ -16,16 +16,23 @@ def uuid_4_parser(plain_string):
...
+        raise TypeError('Claim not a valid string value')
+    return string_value
+
+
+def optional_string_parser(optional_string_value):
+    if not isinstance(optional_string_value, str):
```
jonnyshaw89, line 31: What if it isn't there?

jonnyshaw89 (changes requested):

app/jinja_filters.py
```diff
@@ -352,7 +352,8 @@ def format_unordered_list(context, list_items):
 
 @blueprint.app_template_filter()
 def concatenated_list(list_items, delimiter=', '):
-    return Markup(delimiter).join(item.strip() for item in list_items if item)
+    if list_items:
+        return Markup(delimiter).join(item.strip() for item in list_items if item)
```
jonnyshaw89: Why have you made this change?

jonnyshaw89 (changes requested):

app/jinja_filters.py
dcdarrell9: So, with the `format_address_list` jinja filter we get an error when all of the fields are empty strings. So if we have an empty list in this `concatenated_list` function it will error, I was just having a play about and accidentally added it, I can remove this and change format_address_list or add handling here? But it should be there in case there's an error

data/en/lms_2.json
```diff
@@ -21,11 +21,31 @@
...
             "validator": "string"
         },
         {
-            "name": "country",
+            "name": "address_line1",
+            "validator": "optional_string"
+        },
+        {
+            "name": "address_line2",
+            "validator": "optional_string"
```
ajmaddaford, line 33: Isn't `optional` defined in the schema as a new property (https://github.com/ONSdigital/eq-schema-validator/blob/master/schemas/questionnaire_v1.json#L78), so would look like
```
{
            "name": "address_line2",
            "validator": "string"
			"optional": true
}
```
in a survey schema?

dcdarrell9: Yeah, so this was what we had originally (I added the `"optional": true`) but after a couple of discussions we went with creating `"validator": "optional_string"`

https://github.com/ONSdigital/eq-schema-validator/pull/93

insacuri: If you can remember, Why did you go with this instead of the first version?

app/jinja_filters.py
jonnyshaw89: i think `concatenated_list` is used by many other filters and there are no tests to support this.

jonnyshaw89 (changes requested):

data/en/lms_2.json
dcdarrell9: Instead of adding that additional field, it was added to the existing list of types, there's validation that's done in metadata_parser.py(~L44) which does validation on each of these types, so it made sense to add it to there

app/storage/metadata_parser.py
```diff
@@ -114,13 +99,17 @@ def _validate_metadata_values_are_valid(claims, required_metadata):
         for metadata_field in required_metadata:
             name = metadata_field['name']
             claim = claims.get(name)
-            if claim:
-                logger.debug('parsing metadata', key=name, value=claim)
-                VALIDATORS[metadata_field['validator']](claim)
+            if name == 'trad_as_or_ru_name':
+                claim = claims.get('ru_name') or claims.get('trad_as')
+            if name != 'trad_as_or_ru_name' and name not in claims:
```
jonnyshaw89: if you use `elif` you can get rid of the `name != 'trad_as_or_ru_name' and`?

jonnyshaw89 (changes requested):
Last one i think

jonnyshaw89 (approved):

PR #1808, opened 2018-10-02
Fix summary context repeating groups

bitdivision:
We currently don't add any group to the summary context if it doesn't have any blocks in it. 

This was to avoid showing and empty group on the summary. e.g if you only add one person on test_repeat_until_summaries.json. it would show the `Other Household Members` title without any content.

Not adding groups to the summary context causes issues in LMS for instance, when we have one person with 1 job and another with 2 jobs, the first group would not be added to the context (since they don't have any of the `two-job` questions, yet we still reference the context by the group_instance. So in this case there would only be one item in context['summary']['groups'], but we're trying to reference it as if there were 2 items.

I've added a condition to the summary template to avoid showing anything if there are no blocks in the group. 

### How to review 
Check test_repeat_until_summaries does not include any extra titles etc.

### Checklist

### DIFF
--- a/app/templates/partials/summary/summary.html
+++ b/app/templates/partials/summary/summary.html
@@ -2,7 +2,7 @@
     {%- if content.summary.collapsible -%}
         {%- include 'partials/summary/collapsible-summary.html' -%}
     {%- else -%}
-        {%- for group in content.summary.groups -%}
+        {%- for group in content.summary.groups if group.blocks -%}
             {%- if group.title -%}
                 <h2 class="summary__title neptune" id="{{ group.id }}">{{ group.title }}</h2>
                 <div class="summary__block u-mb-l">
@@ -26,7 +26,7 @@
                 {%- if content.summary.summary_type == 'CalculatedSummary' -%}
                     {%- include theme(['partials/summary/calculated-summary-answer.html']) -%}
                 {%- endif -%}
-            {%- if loop.last or loop.nextitem.title-%}
+            {%- if loop.last or loop.nextitem.title -%}
                 </div>
             {%- endif -%}
         {%- endfor -%}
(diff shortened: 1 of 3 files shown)


app/templating/summary_context.py
```diff
@@ -31,8 +31,7 @@ def build_summary_rendering_context(schema, sections, answer_store, metadata, sc
             repeating_groups = []
             for instance_idx in range(0, no_of_repeats):
                 summary_group = Group(group, path, answer_store, metadata, schema, instance_idx, schema_context).serialize()
-                if summary_group['blocks']:
-                    repeating_groups.extend([summary_group])
+                repeating_groups.extend([summary_group])
```
jonnyshaw89: could this be `repeating_groups.append(summary_group)` ?

jonnyshaw89 (changes requested):

bitdivision:
Fixed @jonnyshaw89

jonnyshaw89 (approved):

bitdivision:
We could have had a situation when loop.nextitem was an empty group, so I've filtered in the for loop to avoid that. Nicer anyway.

app/templates/partials/summary/summary.html
```diff
@@ -2,8 +2,8 @@
     {%- if content.summary.collapsible -%}
         {%- include 'partials/summary/collapsible-summary.html' -%}
     {%- else -%}
-        {%- for group in content.summary.groups -%}
-            {%- if group.title -%}
+        {%- for group in content.summary.groups if group.blocks-%}
+            {%- if group.title and group.blocks-%}
```
jonnyshaw89: is ` and group.blocks` redundant as it cant get in here if no blocks?

jonnyshaw89 (changes requested):

app/templates/partials/summary/summary.html
bitdivision: My bad.

jonnyshaw89 (approved):

app/templates/partials/summary/summary.html
```diff
@@ -2,7 +2,7 @@
     {%- if content.summary.collapsible -%}
         {%- include 'partials/summary/collapsible-summary.html' -%}
     {%- else -%}
-        {%- for group in content.summary.groups -%}
+        {%- for group in content.summary.groups if group.blocks-%}
```
dcdarrell9: Space at the end? :trollface:

dcdarrell9 (approved):

PR #1810, opened 2018-10-03
Fix casual working routing for LMS.

bitdivision:
[trello](https://trello.com/c/I7bckAlk/2451-add-casac-block-and-associated-routing-to-lms)

When you select casual working, the routing around usual hours / totaliser was a little broken.

### How to review 
- You should never be asked usual hours if you are a casual worker. 
- You should be routed to the CASAC question when asked for actual hours.
- The totaliser should still be skipped if there would only be one entry

### Checklist

### DIFF
--- a/data/en/lms_2.json
+++ b/data/en/lms_2.json
@@ -3455,6 +3455,75 @@
                             },
                             {
                                 "goto": {
+                                    "block": "unpaid-overtime-w10",
+                                    "when": [{
+                                        "id": "not-working-casual-work-answer",
+                                        "condition": "equals",
+                                        "value": "Yes"
+                                    }, {
+                                        "id": "overtime-answer",
+                                        "condition": "equals",
+                                        "value": "Paid and unpaid overtime"
+                                    }]
+                                }
+                            },
+                            {
+                                "goto": {
+                                    "block": "paid-overtime-w-11",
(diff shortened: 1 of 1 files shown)


data/en/lms_2.json
```diff
@@ -3711,13 +3780,8 @@
                             },
                             {
                                 "goto": {
-
                                     "block": "casual-hours-worked-casac",
-                                    "when": [{
```
bitdivision: Pretty sure this was redundant.

MebinAbraham (commented):
Happy to approve as CASAC seems to work fine but noticed that for the non-casual route, if you say you didn't work any days of the week it still asks you about `Actual hours` but it shouldn't?. Make a new card or?.

bitdivision:
Think we should create a new card for any other issues.

Makes it easier for other people to know what issues we have found / are fixing and minimises merge conflicts with any other LMS changes.

MebinAbraham (approved):
As discussed, CASAC is working.
Need confirmation from Donna for the other missing routing, But another card for that is in Trello.

bitdivision:
Going to rebase as I don't think anyone else has started reviewing

jonnyshaw89 (approved):

PR #1812, opened 2018-10-04
LMS - Who lives here section

jonnyshaw89:
Have updated the Who lives here section to meet the slides

Also fixes navigation to work with the multiple repeating groups

### DIFF
--- a/app/questionnaire/navigation.py
+++ b/app/questionnaire/navigation.py
@@ -1,7 +1,6 @@
-from collections import defaultdict
-
 from structlog import get_logger
 
+from app.helpers.schema_helpers import get_group_instance_id
 from app.questionnaire.completeness import Completeness
 from app.questionnaire.location import Location
 from app.questionnaire.rules import evaluate_repeat
@@ -85,12 +84,10 @@ class Navigation:
 
         repeating_nav = []
         if repeating_rule['type'] == 'answer_count':
-            link_names = self._generate_link_names(section['title_from_answers'])
 
             for group in groups:
                 is_current_group = group['id'] == current_group_id
-                repeating_nav += self._generate_repeated_items(
-                    link_names, group, no_of_repeats, is_current_group, current_group_instance)
+                repeating_nav += self._generate_repeated_items(section['title_from_answers'], group, no_of_repeats, is_current_group, current_group_instance)
 
         elif no_of_repeats > 0:
             target_location = self._get_location_for_section(section, groups)
@@ -105,7 +102,7 @@ class Navigation:
         is_completed = self.completeness.is_section_complete(section)
         return self._generate_item(section['title'], is_completed, target_location, is_highlighted)
 
-    def _generate_repeated_items(self, link_names, group, no_of_repeats, is_current_group,
+    def _generate_repeated_items(self, title_from_answers, group, no_of_repeats, is_current_group,
                                  current_group_instance):
         """
         Generates a lists of navigation items
@@ -126,8 +123,10 @@ class Navigation:
             is_current_group_instance = is_current_group and i == current_group_instance
             is_completed = self.completeness.is_group_complete(group, group_instance=i)
 
+            link_name = self._generate_link_name(title_from_answers, location)
+
             nav_item = self._generate_item(
-                name=link_names.get(i),
+                name=link_name,
                 completed=is_completed,
                 location=location,
                 highlight=is_current_group_instance,
@@ -146,18 +145,17 @@ class Navigation:
             'repeating': repeating,
         }
 
-    def _generate_link_names(self, label_answer_ids):
-        link_names = defaultdict(list)
+    def _generate_link_name(self, label_answer_ids, location):
+        link_names = []
+
+        group_instance_id = get_group_instance_id(self.schema, self.answer_store, location)
 
         for answer_id in label_answer_ids:
-            for answer in self.answer_store.filter(answer_ids=[answer_id]).escaped():
+            for answer in self.answer_store.filter(answer_ids=[answer_id], group_instance_id=group_instance_id).escaped():
                 if answer['value']:
-                    link_names[answer['answer_instance']].append(answer['value'])
+                    link_names.append(answer['value'])
 
-        for link_name in link_names:
-            link_names[link_name] = ' '.join(link_names[link_name])
-
-        return link_names
+        return ' '.join(link_names)
 
     def _get_location_for_section(self, section, groups):
         section_completeness = self.completeness.get_state_for_section(section)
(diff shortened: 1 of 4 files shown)


data/en/lms_2.json
```diff
@@ -33,94 +33,103 @@
...
-                                            "description": "The address above is my main residence"
+                                            "label": "Yes, the above address is my main residence",
+                                            "value": "Yes"
                                         },
                                         {
-                                            "label": "No",
+                                            "label": "No, the above address is not my main residence",
                                             "value": "No",
-                                            "description": "The address above is not my main residence.  For example it is my old address, second home, holiday home or business address"
+                                            "description": "For examples it is my old address, second home, holiday home, business address or is not familiar to me"
```
bitdivision, line 101: For example

bitdivision: I've commented on the spec, Donna will need to change it.

data/en/lms_2.json
```diff
@@ -3412,196 +3575,147 @@
...
+                                            "id": "working-days-answer",
+                                            "condition": "contains",
+                                            "value": "Did not work this week"
+                                        }
+                                    ]
                                 }
                             },
                             {
                                 "goto": {
-                                    "block": "unpaid-overtime-w10",
```
bitdivision, line 3477: This is just a bad diff I think?

bitdivision (commented):
Obviously address piping and navigation aren't in yet. 

I've done a little bit of testing on the routing and flow and it all seems to be as expected.

MebinAbraham (commented):
Person icon needs to be added instead of bullets for listing peoples names?

data/en/lms_2.json
```diff
@@ -53,93 +53,103 @@
...
                         },
                         {
                             "type": "Question",
                             "id": "address-check-block",
-                            "description": "",
+                            "description": "{{metadata['ru_name']}}",
                             "questions": [{
                                 "id": "address-check-question",
-                                "title": "Is this address your main residence?",
+                                "title": "Is this address your main residence",
```
MebinAbraham: Does it not need the `?`

data/en/lms_2.json
```diff
@@ -53,93 +53,103 @@
...
+                                        "type": "TextField"
+                                    }
                                 ]
                             }]
                         },
                         {
                             "type": "Question",
                             "id": "address-check-block",
-                            "description": "",
+                            "description": "{{metadata['ru_name']}}",
```
MebinAbraham: Bad merge?. Not meant to be using `ru_name` anymore?

MebinAbraham (commented):
Question:
Do we require the driver questions to be mandatory?:
`/other-household-member-group/0/other-household-member-anyone-else-block`
and
`/student-household-member-group/0/student-household-member-anyone-else-block`
are both mandatory in schema but not in the spec. Is this a requirement for `repeat_until` cause we don't treat no answer the same way?

MebinAbraham (changes requested):
Questions for slide 10 and 11 are mandatory in the schema but not according to spec.

data/en/lms_2.json
```diff
@@ -556,15 +716,15 @@
                         "questions": [{
                             "id": "time-away-question",
                             "titles": [{
-                                    "value": "Is <em>{{[answers['first-name'][group_instance], answers['last-name'][group_instance]] | format_household_name }}</em> currently away from {{ metadata['display_address'] }} for a continuous period of 6 months or more?",
+                                    "value": "Is <em>{{[group_instances[group_instance_id]['primary-household-member-first-name'], group_instances[group_instance_id]['primary-household-member-last-name'], group_instances[group_instance_id]['other-household-member-first-name'], group_instances[group_instance_id]['other-household-member-last-name'], group_instances[group_instance_id]['student-household-member-first-name'], group_instances[group_instance_id]['student-household-member-last-name']] | format_household_name}}</em> currently away from {{ metadata['ru_name'] }} for a continuous period of 6 months or more?",
```
dcdarrell9: Same here with `ru_name`

app/questionnaire/navigation.py
```diff
@@ -146,18 +145,17 @@ def _generate_item(self, name, completed, location, highlight, repeating=False):
             'repeating': repeating,
         }
 
-    def _generate_link_names(self, label_answer_ids):
-        link_names = defaultdict(list)
+    def _generate_link_name(self, label_answer_ids, location):
+        link_names = []
+
+        group_instance_id = get_group_instance_id(self.schema, self.answer_store, location)
```
bitdivision, line 151: 👍

jonnyshaw89:
@MebinAbraham can you re-review?

dcdarrell9:
Sooo, when you navigate back to `About Household` If you change the name then it works fine, but then if you try to 'add' any more members to the household, it will overwrite the next in the list. 

For example: 
- If you had added -> 'darrell cox' / 'jon shaw' / 'mebin abraham'
- Return to About Household and be asked about 'darrell cox' name -> change to 'Darrell Cox' -> save
- On the next screen it will ask if you want to add another member -> add another member 'Andrew MadDog' -> save -> 'jon shaw' will have been overwritten

Not sure what the expected functionality is, or what we should be doing with this

bitdivision (approved):

MebinAbraham (commented):
Some routing on slide 5 (Hholdchk) seems to be missing.
```If no and answer to ineligbletype = ‘student accommodation rented from a private landlord’ then go to sharecooking```

Routing goes to `Addppl` instead of `sharecooking`

So:
- Person 1 `(/primary-household-member-group/0/primary-household-member-block)`
- No, the above address is not my main residence `(/primary-household-member-group/0/address-check-block)`
- Student accommodation rented through a private landlord `(/address-type-group/0/address-type-block)`
- No, this is not the main residence of any other adults, children or babies `(/other-household-member-group/0/other-household-member-anyone-else-block)`

Goes to: `/student-household-member-group/0/student-household-member-anyone-else-block`
but should go to: `/share-group/0/share-cooking-block`?

jonnyshaw89:
@MebinAbraham can you check again?

MebinAbraham:
@jonnyshaw89 Not sure if it was the recent change that @boxadesign or because of the groups/block structure, but some blocks don't have `Previous` links such as the sharecooking.

Also, the mandatoriness doesn't match the spec. Apart from that, the flow seems to be as expected.

jonnyshaw89:
@MebinAbraham any block at the beginning of a group doesnt gave a Previous link.
can you be more specific about missing mandatory?

MebinAbraham:
@jonnyshaw89 sharecooking

MebinAbraham (approved):

PR #1813, opened 2018-10-05
Fix translation for units

bitdivision:
[trello](https://trello.com/c/HWYJdleY/2431-fix-translation-for-units-s)

Remove the hardcoded 'en_GB' locale from calls to babel in jinja filters. Since the flask_babel locale is set in `setup.py` we can use `get_locale` getter to retrieve it.

I've added a functional test for the unit_patterns schema. This required some changes to generate pages to support multiple answers in summaries and to support input unit labels.

### How to review 
You can test this by doing the same thing as the functional tests (running test_unit_patterns.json and changing language_code to `cy`).

Please could someone who speaks welsh ensure that the different versions of 'years' that are shown are actually correct?

This does raise an issue with the pattern library I think in that the `type` shown in the input unit label does not scale with the length of the unit. This becomes apparent with the Welsh word for 'years'.

### Checklist

### DIFF
--- a/data/en/test_unit_patterns.json
+++ b/data/en/test_unit_patterns.json
@@ -15,6 +15,9 @@
     }, {
         "name": "ru_name",
         "validator": "string"
+    }, {
+        "name": "language_code",
+        "validator": "string"
     }],
     "sections": [{
         "id": "default-section",
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,7 @@ before_install:
     - pyenv install --skip-existing
     - pyenv rehash
     - pyenv global $(cat .python-version)
-    - pip install -U pip pipenv wheel
+    - pip install -U pip pipenv==2018.5.18 wheel
     - pipenv install --dev --deploy
     - pipenv check -i '36333 '
 cache:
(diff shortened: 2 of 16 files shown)


bitdivision:
Feel free to squash if everyone's happy, or I can sort any issues on Tuesday.

jonnyshaw89 (approved):

data/en/test_unit_patterns.json
```diff
@@ -15,6 +15,9 @@
     }, {
         "name": "ru_name",
         "validator": "string"
+    }, {
+        "name": "language_code",
```
insacuri, line 19: Does this need to be added to all of them? Seems odd to only add it to the one schema

jonnyshaw89: Not really, only ones that need to support multiple languages

insacuri: Oki dokes, just checking

insacuri (approved):

MebinAbraham:
@jonnyshaw89 @bitdivision @insacuri I have pinned pipenv temporarily to make it pass travis.
Open issue: https://github.com/pypa/pip/issues/5854

PR #1816, opened 2018-10-09
Fix translations for duration

bitdivision:
[trello](https://trello.com/c/lVONifUJ/2454-ensure-units-on-text-inputs-can-scale-for-translations)

This came up because test_unit_patterns uses a Unit type for hours etc. This should really be a duration, or be removed from that schema.

While investigating that, we noticed that the duration type doesn't support translations. That is required for LMS. The duration input does support longer labels, so we can use it for the Welsh translations.

This hardcodes a duration-month / duration-year type in the duration_unit template to ensure that they can be translated. Obviously this isn't ideal, but the core functionality of durations probably needs to be rethought (see [trello](https://trello.com/c/RwZLqaRi/2460-discuss-duration-type-implementation))

### How to review 
Check that test_durations.json works with Welsh labels.

### Checklist

### DIFF
--- a/app/assets/favicons/browserconfig.xml
+++ b/app/assets/favicons/browserconfig.xml
@@ -2,7 +2,7 @@
 <browserconfig>
     <msapplication>
         <tile>
-            <square150x150logo src="https://cdn.ons.gov.uk/sdc/v1.6.2/favicons/mstile-150x150.png"/>
+            <square150x150logo src="https://cdn.ons.gov.uk/sdc/v1.8.1/favicons/mstile-150x150.png"/>
             <TileColor>#da532c</TileColor>
         </tile>
     </msapplication>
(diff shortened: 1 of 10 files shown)


bitdivision:
Closing until the other issues around durations / input labels are resolved.

ajmaddaford (approved):

MebinAbraham (approved):
Duration labels working as expected. Assuming this is going this, we disabling the failing test temporarily until PL is updated?

bitdivision:
I've skipped the test in this one.

bitdivision:
Test unskipped and updated PL to 1.8.1

bitdivision:
I haven't gotten to the bottom of the failing test yet. For the moment I've disabled it.

@MebinAbraham The behaviour seems to be that it selects the checkbox and then selects an answer from the dropdown, but the checkbox is never cleared. If I then select something from the dropdown it is cleared. I've tried waiting, selecting two different items, and clicking somewhere else on the page. No luck yet.

PR #1819, opened 2018-10-11
Fixes to LMS Who Lives Here

jonnyshaw89:
Fix missing repeating navigation if 1st instance of repeat is skipped
Duplicate who lives here for the no primary flow
Corrected routing rules

### How to review 
Check that it behaves as per slides

### DIFF
--- a/app/questionnaire/rules.py
+++ b/app/questionnaire/rules.py
@@ -282,18 +282,18 @@ def evaluate_when_rules(when_rules, schema, metadata, answer_store, group_instan
     """
 
     for when_rule in when_rules:
-        group_instance_something = group_instance
+        group_instance_to_use = group_instance
         if 'id' in when_rule:
-            if group_instance > 0 and not schema.answer_is_in_repeating_group(when_rule['id']):
-                group_instance_something = 0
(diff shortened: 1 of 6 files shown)


app/questionnaire/rules.py
```diff
@@ -282,17 +282,18 @@ def evaluate_when_rules(when_rules, schema, metadata, answer_store, group_instan
     """
 
     for when_rule in when_rules:
+        group_instance_something = group_instance
```
insacuri: group_instance_something?  The name isn't very expressive I think.

insacuri:
I managed to break it.
- add yourself + 2 others.
- No extra students 
- yes to both sharing questions
- It then goes to an 'Who do you need to add to' page instead of relationships

insacuri:
Same thing happens if you do just a 1 person journey (No extra people or students and any answer to the sharing questions)

app/questionnaire/rules.py
jonnyshaw89: haha, yeh I forgot about that. I was trying it to see if it fixed my issue (which it did) then forgot to call it something sensible

PR #1828, opened 2018-10-15
Mutually Exclusive with Single Checkbox override

MebinAbraham:
Added the ability to have mutually exclusive behaviour with different answer types, all with a single checkbox override.

***Currently supports:***
- Checkbox
- Currency
- Day Month Year Date
- Month Year Date
- Year Date
- Dropdowns
- Number
- Percentage
- TextField
- Unit
- Duration
- TextArea (non JS only, needs PL changes)

### How to review 
- Ensure the behaviour of mutually exclusive with different answer types works as expected.
- Review how the server side validation is handled + how to determine which answer is displayed etc.

To Do:
* [x] More integration tests
* [x] Remove the existing mutually exclusive implementation

Sorry about the large amount of changes, there are a lot of tests.

### Checklist


app/forms/questionnaire_form.py
```diff
@@ -91,6 +88,21 @@ def validate_calculated_question(self, question):
...
+    def validate_mutually_exclusive_question(self, question):
+        is_mandatory = question.get('mandatory')
+        messages = question['validation'].get('messages') if 'validation' in question else None
+        answers = (getattr(self, answer['id']).data for answer in question['answers'] if 'parent_answer_id' not in answer)
+
+        validator = MutuallyExclusiveCheck(messages=messages)
+
+        try:
+            validator(answers, is_mandatory)
+        except validators.ValidationError as e:
```
insacuri, line 102: Could a StopValidation be thrown with this?  If so, it needs to be handled

app/templating/summary/question.py
```diff
@@ -27,24 +27,37 @@ def _get_answers(answer_store, answer_id, group_instance):
     def _get_answer(self, answer_store, answer_id, group_instance):
         return self._get_answers(answer_store, answer_id, group_instance)[0]
 
-    def _build_answers(self, answer_store, question_schema, answer_schemas, group_instance):
+    def _build_answers(self, answer_store, question_schema, group_instance):
```
insacuri, line 30: Nice to see function signatures getting shorter rather then longer ^^

app/templating/summary/question.py
```diff
@@ -74,16 +87,20 @@ def _find_other_value_in_answers(self, answer_store, answer_schema, answer, grou
...
-        return answer_value
+
+        answer_builder = {
+            'Checkbox': self._build_checkbox_answers,
+            'MutuallyExclusiveCheckbox': self._build_checkbox_answers,
+            'Radio': self._build_radio_answer,
+            'Dropdown': self._build_dropdown_answer,
+        }
+
+        return answer_builder[answer_schema['type']](answer_value, answer_schema) \
```
insacuri: I think this should be rewritten to 
```
if answer_schema['type'] in answer_builder.keys():
	return answer_builder[answer_schema['type']](answer_value, answer_schema)
else:
	return answer_value
```
I've never been a fan of writing long 'single' line if else's

app/templating/summary/question.py
```diff
@@ -98,7 +114,7 @@ def _build_checkbox_answers(answer, answer_schema):
                 else:
                     multiple_answers.append(CheckboxSummaryAnswer(label=option['label'],
                                                                   should_display_other=False))
-        return multiple_answers
+        return multiple_answers if multiple_answers != [] else None
```
insacuri: I'd be tempted to use `if not multiple_answers` as per https://stackoverflow.com/questions/53513/how-do-i-check-if-a-list-is-empty/53752#53752

But it's a stylistic opinion so I'm not fussed

app/validation/validators.py
```diff
@@ -1,19 +1,19 @@
...
 import re
+from datetime import datetime
+from decimal import Decimal, InvalidOperation
 
 import flask_babel
-from flask_babel import lazy_gettext as _, ngettext
+from flask_babel import ngettext
 from babel import numbers
 from dateutil.relativedelta import relativedelta
+
```
insacuri: Why's there a space here?

scripts/test_schemas.sh
```diff
@@ -6,8 +6,8 @@ if [ "$1" == "--local" ] || [ "$2" == "--local" ]; then
 fi
 
 if [ "$run_docker" == true ]; then
-    docker pull onsdigital/eq-schema-validator
-    validator="$(docker run -d -p 5001:5000 onsdigital/eq-schema-validator)"
+    docker pull onsdigital/eq-schema-validator:eq-2359-single-checkbox-override
```
insacuri: Usual remember to change this before merging comment ^^

tests/functional/spec/components/checkbox/mutually_exclusive/mutually_exclusive_currency.spec.js
```diff
@@ -0,0 +1,120 @@
...
+    return helpers.openQuestionnaire('test_mutually_exclusive_single_checkbox_override.json').then(() => {
+          return browser.click(helpers.navigationLink('Currency'));
+        });
+  });
+
+  describe('Given the user has entered a value for the non-exclusive currency answer', function() {
+    it('When then user clicks the mutually exclusive checkbox answer, Then only the mutually exclusive checkbox should be answered.', function() {
+
+      return browser
+        // Given]
```
insacuri: extra ]

app/templating/summary/question.py
bitdivision: Agreed, but I think this can just be `return multiple_answers or None`? Since empty list is Falsey

app/templating/summary/question.py
MebinAbraham: I find it fine but fair enough.
Also no need for `else` when the `if` returns

app/templating/summary/question.py
MebinAbraham: This is what I was meant to do 👀

app/forms/questionnaire_form.py
MebinAbraham: It only returns a question validation error if the fields itself are valid so should be fine as is unless it causes a problem?

app/validation/validators.py
MebinAbraham: Dodgy auto optimize. 👍

app/forms/questionnaire_form.py
insacuri: Oki dokes, I had no idea so I thought I'd ask ^^

app/templating/summary/question.py
insacuri: Agreed, I remembered at lunch that the else is superfluous :)

tests/integration/components/mutually_exclusive/test_percentage_single_checkbox_override.py
```diff
@@ -0,0 +1,47 @@
...
+        self.assertInUrl('section-summary')
+        self.assertInPage('99%')
+
+    def test_exclusive_answer(self):
+        # When
+        self.post({'percentage-answer': '',
+                   'percentage-exclusive-answer': ['I prefer not to say']})
+
+        # Then
+        # self.assertInUrl('section-summary')
```
insacuri: Can this be removed?

tests/integration/components/mutually_exclusive/test_month_year_date_single_checkbox_override.py
```diff
@@ -0,0 +1,52 @@
...
+        self.assertInPage('October 2018')
+
+    def test_exclusive_answer(self):
+        # When
+        self.post({'month-year-date-answer-month': '',
+                   'month-year-date-answer-year': '',
+                   'month-year-date-exclusive-answer': ['I prefer not to say']})
+
+        # Then
+        # self.assertInUrl('section-summary')
```
insacuri: Can this be removed? (and all the others in the same position)

tests/integration/components/mutually_exclusive/test_checkbox_single_checkbox_override.py
```diff
@@ -0,0 +1,45 @@
...
+        super().setUp()
+        self.launchSurvey('test', 'mutually_exclusive_single_checkbox_override')
+
+    def test_non_exclusive_answer(self):
+        # When
+        self.post({'checkbox-answer': ['British', 'Irish']})
+
+        # Then
+        self.assertInUrl('section-summary')
+        self.assertInPage('British')
```
insacuri: For this one (and all others in this PR), it should be assertInBody.  We were going to make this change as part of https://trello.com/c/7xsxQmNz/2407-change-assertinpage-to-assertinbody and doing it now would reduce the amount of work we'd have to do

tests/integration/components/mutually_exclusive/test_percentage_single_checkbox_override.py
MebinAbraham: Whoops, they are needed, commented out when I was debugging but forgot to change back after. Will sort. 👍

tests/integration/components/mutually_exclusive/test_month_year_date_single_checkbox_override.py
MebinAbraham: Same as above.

tests/integration/components/mutually_exclusive/test_checkbox_single_checkbox_override.py
MebinAbraham: Cool, will change.

MebinAbraham: We currently don't have `assertInBody` and don't really want to make this any muddier but will make the change since its a minor change.

MebinAbraham:
Changes to LMS routing where we used MutuallyExclusiveCheckbox previously. New MutuallyExclusive means 2 answer_ids so please ensure the routing for those options still work as expected. Docs on Trello card.

bitdivision:
@MebinAbraham I've double checked the routing on LMS, everything seems good :+1:

tests/integration/components/mutually_exclusive/test_checkbox_single_checkbox_override.py
insacuri: Just checked, I thought assertInPage/assertInBody were standard functions in unittest.
Yea, ignore that comment as I thought it was simple to do.

app/forms/questionnaire_form.py
```diff
@@ -32,24 +30,21 @@ def __init__(self, schema, block_json, answer_store, metadata, **kwargs):
...
-                if valid_date_range_form:
-                    valid_date_range_form = valid_date_range
+                valid_form = self.validate_date_range_question(question)
             elif question['type'] == 'Calculated':
-                valid_calculated = self.validate_calculated_question(question)
-                if valid_calculated_form:
-                    valid_calculated_form = valid_calculated
+                valid_form = self.validate_calculated_question(question)
+            elif question['type'] == 'MutuallyExclusive' and valid_fields:
+                valid_form = self.validate_mutually_exclusive_question(question)
```
dwyeradam: If we have multiple questions on a page then the last valid_form value will be the only one that is used. If one of the earlier valid_forms is a failure it will not be picked up.

dwyeradam (changes requested):

insacuri (changes requested):
Tested the TextArea part of the mutually exclusive test json.  All the other ones clear their fields down when you click the exclusive option then click somewhere else, however, TextArea doesn't for some reason.  I think this needs to be fixed

app/forms/questionnaire_form.py
MebinAbraham: Ah makes sense. Will change. Good spot 👍

MebinAbraham:
@insacuri TextArea mutually exclusive won't work until pattern lib changes are made/released.

insacuri (approved):
Code changes look reasonable and the testing I did on both lms and the specific test schema seem to work fine too.

dcdarrell9 (commented):
LGTM

ajmaddaford (approved):
Looks good. There's some work outside of the scope of this pull request that we need to do to simplify the templates, but these changes make sense right now.

dwyeradam (approved):
It will do I suppose.

scripts/test_schemas.sh
```diff
@@ -6,8 +6,8 @@ if [ "$1" == "--local" ] || [ "$2" == "--local" ]; then
...
+    validator="$(docker run -d -p 5001:5000 onsdigital/eq-schema-validator:eq-2359-single-checkbox-override)"
```
dwyeradam: @MebinAbraham reverty, squishy, squashy.

PR #1830, opened 2018-10-15
Filter answers based on routing path when evaluating routing

bitdivision:

[trello](https://trello.com/c/FD4aiOuW/2439-question-should-we-modify-answer-store-filter-to-only-return-answers-on-the-routing-path)

Currently, we use the full answer store when checking routing rules / skip conditions. This means that if you navigate backwards, change your path, and then encounter a routing rule which evaluates a (now invalid) answer value, it will still route you according to the invalid answer. 

The original proposal to fix this was to filter the answer store based on the routing path. This proved problematic since path_finder uses the answer store to build the path. The relationship between answer store and routing path probably needs to be discussed at a higher level, see [trello](https://trello.com/c/kanSjZJf/2479-discuss-interplay-between-routing-path-path-finder-and-answer-store). Given this, I haven't attempted to make any changes to the answer_store.

The scope of this card was then limited to fixing the evaluation of routing in `rules.py`. We discussed eventually using some representation of answer_id / group_instance stored with the routing_path to improve the performance of comparisons. I haven't implemented this yet, since it's likely better done as part of refactoring the routing path.

I ended up copying the answer_store in the `_get_answers_on_path` function. I think we could get away with mutating it, but that would be a potential source of issues down the line.

I had to refactor the path_finder a little to allow skip conditions to take the routing_path into account.

### How to review 
- Run `test_routing_not_affected_by_answers_not_on_path`. Follow the instructions to generate an invalid path and make sure you are routed correctly on the second path.

### Checklist

### DIFF
--- a/app/questionnaire/path_finder.py
+++ b/app/questionnaire/path_finder.py
@@ -30,6 +30,16 @@ class PathFinder:
                      b['group_instance'] == location.group_instance),
                     None)
 
+    @staticmethod
+    def _build_blocks_for_group(group, instance_idx):
+        for block in group['blocks']:
+
+            yield {
+                'group_id': group['id'],
+                'group_instance': instance_idx,
+                'block': block,
+            }
+
     def build_path(self):
         """
         Visits all the blocks from a location forwards and returns path
@@ -60,7 +70,8 @@ class PathFinder:
                     group_instance_id = get_group_instance_id(self.schema, self.answer_store, Location(group['id'], group_instance, first_block_in_group))
 
                     if evaluate_skip_conditions(group['skip_conditions'], self.schema, self.metadata,
-                                                self.answer_store, group_instance, group_instance_id):
+                                                self.answer_store, group_instance, group_instance_id,
+                                                routing_path=path):
                         continue
 
                 group_blocks = list(self._build_blocks_for_group(group, group_instance))
@@ -81,19 +92,6 @@ class PathFinder:
 
         return path
 
-    def _build_blocks_for_group(self, group, instance_idx):
-        for block in group['blocks']:
-            skip_conditions = block.get('skip_conditions')
-            if skip_conditions and evaluate_skip_conditions(
-                    skip_conditions, self.schema, self.metadata, self.answer_store, instance_idx):
-                continue
-
-            yield {
-                'group_id': group['id'],
-                'group_instance': instance_idx,
-                'block': block,
-            }
-
     def _get_first_group_in_section(self):
         return [
             section['groups'][0]['id']
@@ -109,10 +107,22 @@ class PathFinder:
             if block_index is None:
                 return path, prev_block_index
 
+            block = blocks[block_index]['block']
+
+            if block.get('skip_conditions') and \
+                    evaluate_skip_conditions(block['skip_conditions'], self.schema, self.metadata,
+                                             self.answer_store, this_location.group_instance, routing_path=path):
+
+                if block_index == len(blocks) - 1:
+                    return path, block_index
+
+                this_location = Location(blocks[block_index + 1]['group_id'],
+                                         blocks[block_index + 1]['group_instance'],
+                                         blocks[block_index + 1]['block']['id'])
+                continue
+
             if this_location not in path:
                 path.append(this_location)
-
-            block = blocks[block_index]['block']
 
             # If routing rules exist then a rule must match (i.e. default goto)
             if 'routing_rules' in block and block['routing_rules']:
@@ -140,7 +150,8 @@ class PathFinder:
                                         self.metadata,
                                         self.answer_store,
                                         this_location.group_instance,
-                                        group_instance_id=group_instance_id)
+                                        group_instance_id=group_instance_id,
+                                        routing_path=path)
 
             if should_goto:
                 return self._follow_routing_rule(this_location, rule, blocks, block_index, path)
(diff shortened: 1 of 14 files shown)


app/questionnaire/rules.py
```diff
@@ -302,13 +323,19 @@ def evaluate_when_rules(when_rules, schema, metadata, answer_store, group_instan
     return True
 
 
-def get_answer_store_value(answer_id, answer_store, schema, group_instance, group_instance_id=None):
+def get_answer_store_value(answer_id, answer_store, schema, group_instance, group_instance_id=None, routing_path=None):
+
     filtered = answer_store.filter(answer_ids=[answer_id])
 
-    if not filtered.count():
+    if routing_path:
```
jonnyshaw89, line 330: Do we ever want to not use the routing path?

app/questionnaire/rules.py
```diff
@@ -185,18 +197,26 @@ def _get_answers_on_routing_path(schema, routing_path, answer_store, repeat_rule
...
-    answers_to_remove = []
-    for answer in answers:
-        if not _is_answer_on_path(schema, answer, routing_path):
-            answers_to_remove.append(answer)
+    answers_on_path = _get_answers_on_path(answers, schema, routing_path)
+
+    return answers_on_path
+
+
+def _get_answers_on_path(answers, schema, routing_path) -> AnswerStore:
```
jonnyshaw89, line 208: Could this just return a list of the answers? then we wouldnt need to `.copy()` the answer store?

jonnyshaw89 (changes requested):

app/questionnaire/rules.py
bitdivision: Idea here was to avoid refactoring everything in runner to get the routing path

app/questionnaire/rules.py
bitdivision: Currently, yes, though that actually raises another issue. I hadn't changed `get_answer_store_value` to use the `answers_on_path` instead of `filtered`. It would have worked most of the time.

bitdivision: So, after that change, we need to call filter on the result of `_get_answers_on_path`

insacuri:
LGTM, code looks good and the test works.  I don't feel confident enough to approve this though as there might be something fundamental I'm missing

jonnyshaw89:
Does this handle where the question driving a repeating group is no longer on the path?

jonnyshaw89:
> Does this handle where the question driving a repeating group is no longer on the path?

It was already handled

jonnyshaw89 (approved):

bitdivision:
> > Does this handle where the question driving a repeating group is no longer on the path?
> 
> It was already handled

As discussed, handled for normal repeats but not handled for repeat until, that's a lot more complex to fix.

MebinAbraham (commented):
Functionality seems to be working as expected. Tested with LMS. 👍

app/questionnaire/path_finder.py
```diff
@@ -109,11 +107,22 @@ def _build_path_within_group(self, blocks, block_index, this_location, path):
...
+            if block.get('skip_conditions') and \
+                    evaluate_skip_conditions(block['skip_conditions'], self.schema, self.metadata,
+                                             self.answer_store, this_location.group_instance, routing_path=path):
+
+                if block_index < len(blocks) - 1:
+                    this_location = Location(blocks[block_index + 1]['group_id'],
+                                             blocks[block_index + 1]['group_instance'],
+                                             blocks[block_index + 1]['block']['id'])
+                    continue
+                return path, block_index
```
ajmaddaford: I think this would be easier to understand if we returned before changing `this_location`, something like:

```python
if block_index == len(blocks) - 1:
	return path, block_index

this_location = Location(blocks[block_index + 1]['group_id'],
                         blocks[block_index + 1]['group_instance'],
                         blocks[block_index + 1]['block']['id'])
continue
```

data/en/test_routing_not_affected_by_answers_not_on_path.json
```diff
@@ -0,0 +1,140 @@
...
+            }, {
+                "type": "Interstitial",
+                "id": "valid-final-interstitial",
+                "title": "You were routed correctly!",
+                "description": ""
+            }, {
+                "type": "Interstitial",
+                "id": "invalid-final-interstitial",
+                "title": "You were routed incorrectly.",
+                "description": ""
```
ajmaddaford: This doesn't work, it's always shown even if the `valid-final-interstitial` page is shown.

app/questionnaire/rules.py
```diff
@@ -185,18 +197,26 @@ def _get_answers_on_routing_path(schema, routing_path, answer_store, repeat_rule
...
 
     for answer in answers_to_remove:
-        answers.remove(answer)
+        answers_on_path.answers.remove(answer)
 
-    return answers
+    return answers_on_path
 
 
 def _is_answer_on_path(schema, answer, routing_path):
```
ajmaddaford, line 222: I think it would make more sense if the `_get_answers_on_path` and `_is_answer_on_path` methods were in path_finder.

ajmaddaford (changes requested):

app/questionnaire/rules.py
bitdivision: Agreed, however causes circular imports atm. Planning to leave this alone for now and fix it later.

ajmaddaford (approved):

PR #1833, opened 2018-10-16
Add static translations

dcdarrell9:
- Added static translations for runner.

### How to review 
- Check that translations work on different types of surveys for static content.
- Check switching between languages.
- Check error pages.

### Checklist


jonnyshaw89 (dismissed):
Do we want to avoid adding binary files to the repo and insead compile the translations as part of the build script?

jonnyshaw89:
we should probably update `.gitignore` so we don't all start commiting `.mo` files

scripts/translate_static.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
```
bitdivision: Would suggest adding `set -e` to ensure this doesn't fail and continue.

jonnyshaw89: The same could be said for `scripts/build.sh`
However i know that fails on `echo "Testing schemas"` in Jenkins as it doesnt have docker.
so maybe `echo "Testing schemas"` doesn't belong as part of `build.sh`?

jonnyshaw89: Sorry about the ramblingness of that comment 😄

jonnyshaw89:
How is `translate_static.sh` going to be run?
if we want to add it to the concourse build then the container should probably be updated to include the dependencies

jonnyshaw89:
Both concourse boxes would also need to be updated to include these dependencies

scripts/translate_static.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash -e
```
jonnyshaw89, line 1: What triggers this script?

MebinAbraham (commented):
Some of the translations are not being picked up?. Some missing on cookies and privacy page.

MebinAbraham (commented):
Problems with errors messages.
screen shot 2018-10-18 at 14 48 21

dcdarrell9:
Issue with some of the error messages not getting translated, most likely related to `ngettext()`

dcdarrell9:
This may need to be re-run if there are more static changes in the coming days.

When @MebinAbraham was doing some testing on this, there was a paragraph in the cookies & privacy page that wasn't translated, but translated for me. (sanity check that they're there)

There are some translations missing with the ngettext, not sure if this is a pybabel thing or something that we're doing wrong.

With regards to where this should be run, the build.sh script doesn't get run on a PR, but I think it does in eq-deploy? @jonnyshaw89 not sure where or even if we want this in a build long term?

bitdivision:
So build.sh will be called in eq-deploy. The only thing we need to check is that the python environment is loaded at that point (to ensure we have pybabel available).

MebinAbraham (commented):
Error page seems to be missing some?
screen shot 2018-10-23 at 13 55 14

ajmaddaford (approved):

bitdivision (approved):
I did find one issue, but I don't think it should block this:

please provide feedback if you have any suggestions on how we can improve this study - online household study

The characters remaining is not translated.

MebinAbraham:
@bitdivision That is a Pattern Lib issue as that string is set in the JS from CDN.

MebinAbraham (approved):
Comments addressed 👍

MebinAbraham (commented):
`Remove and answer to continue` error message still isn't translated. Thought we were adding this manually?

MebinAbraham (commented):
We forgot to mark `Select relationship` for translation also.

Issue #1838, opened 2018-10-18
Feedback not within grid column


bameyrick:
### Expected behaviour
The survey feedback should be the same width as the rest of the survey.

### Actual behaviour
It's full width and the line length of the text area is too long.

### Steps to reproduce the behaviour
Open feedback.

### Screenshot
![screenshot 2018-10-18 at 10 20 54](https://user-images.githubusercontent.com/8447446/47144557-97abb800-d2bf-11e8-9213-c6c66e36c770.png)

PR #1839, opened 2018-10-18
Wrap feedback in grid column

bameyrick:
Wrapped feedback in grid column to keep it same width as the main column. https://github.com/ONSdigital/eq-survey-runner/issues/1838

### How to review 
Check all affected templates are ok.
Check it's the same width as the rest of the content.

### DIFF
--- a/app/templates/introduction.html
+++ b/app/templates/introduction.html
@@ -80,9 +80,13 @@
 {% endblock %}
 {% block footer %}
 <div class="pre-footer">
-    <div class="container">
+  <div class="container u-mt-xl u-pt-l">
+    <div class="grid">
+      <div class="grid__col col-7@m">
         {% include theme('partials/feedback/expandable_inline.html') %}
+      </div>
     </div>
+  </div>
 </div>
 {{super()}}
 {% endblock %}
--- a/app/templates/layouts/_questionnaire.html
+++ b/app/templates/layouts/_questionnaire.html
@@ -40,9 +40,13 @@
 {% endblock main %}
 {% block footer %}
 <div class="pre-footer">
-    <div class="container">
-      {% include theme('partials/feedback/expandable_inline.html') %}
+  <div class="container u-mt-xl u-pt-l">
+    <div class="grid">
+      <div class="grid__col col-7@m">
+        {% include theme('partials/feedback/expandable_inline.html') %}
+      </div>
     </div>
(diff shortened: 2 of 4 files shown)


app/templates/introduction.html
```diff
@@ -80,7 +80,11 @@ <h2 class="neptune u-mb-xs" data-qa="legal-response">{{ _("Your response is lega
 {% endblock %}
 {% block footer %}
 <div class="container u-mt-xl u-pt-l">
-    {% include theme('partials/feedback/expandable_inline.html') %}
+  <div class="grid">
+    <div class="grid__col col-7@m">
+      {% include theme('partials/feedback/expandable_inline.html') %}
```
MebinAbraham: From what I can see, all usage of `explandable_inline.html` is wrapped in:
```
<div class="container u-mt-xl u-pt-l">
  	<div class="grid">
```
So is it worth just adding these to `expandable_inline.html` and reduce duplication?

bameyrick: I kept the grid layout outside of the partial incase the partial needed to be used elsewhere in a different layout, but as its only used in the same place here I could reduce duplication by moving the grid into the partial, although in my opinion layout elements should live outside partials

bitdivision (approved):
Changes look good, and it works in Chrome

app/templates/introduction.html
MebinAbraham: @bameyrick I don't really mind, but we already have layout elements inside the partial?

MebinAbraham:
@bameyrick Can you merge master?

bameyrick:
@MebinAbraham Merged and resolved conflicts 👍

MebinAbraham (approved):

MebinAbraham:
Welsh PR need to go first. Then this can be squashed and rebased.

bitdivision:
@bameyrick You'll need to set up signed commits if you haven't already:

> You may need to set up gpg keys, and add them to github. There are lots of guides online. 
The easiest way to ensure that all your commits are signed is by running `git config --global commit.gpgsign true`, which will sign all commits on all repos. Alternatively `git config commit.gpgsign true` can be run on a single repo to only sign that one.

Also, beware of using the github button to merge master. The merge commit is signed by github which means it can't be merged. Best to `git rebase` / `git merge` and push.

PR #1840, opened 2018-10-18
LMS - Answer summaries and no-primary routing

bitdivision:

This change includes:

- Duplicate the schema for LMS to allow for inclusion of the primary person.
  The new section has ids of `no-primary-...`.

- There is a fix for navigation to AnswerSummaries

- A fix for skip conditions to skip properly when the driving question has a group-instance-id and isn't in a repeating group.

- Add AnswerSummary to who lives here.

- Skip section in navigation if all groups are skipped or invalid (fixes bug with 'about-the-household')

- Content fix for LMS
### How to review 
Check that who lives here in LMS works as expected.

Known Issues:
- Navigation for who lives here will show 'About the household' as completed when it isn't

### Checklist

### DIFF
--- a/app/questionnaire/completeness.py
+++ b/app/questionnaire/completeness.py
@@ -49,6 +49,9 @@ class Completeness:
         def eval_state(state_to_compare):
             return (state == state_to_compare for state in group_states)
 
+        def eval_state_in(state_to_compare):
+            return (state in state_to_compare for state in group_states)
+
         section_state = self.NOT_STARTED
 
         if group_states:
@@ -60,6 +63,9 @@ class Completeness:
 
             elif any(eval_state(self.STARTED)):
                 section_state = self.STARTED
+
+            elif all(eval_state_in((self.SKIPPED, self.INVALID))):
+                section_state = self.SKIPPED
 
             elif all(state in self.COMPLETED_STATES for state in group_states):
                 section_state = self.COMPLETED
--- a/app/questionnaire/navigation.py
+++ b/app/questionnaire/navigation.py
@@ -43,9 +43,10 @@ class Navigation:
         navigation = []
 
         for section in self.schema.sections:
-            non_skipped_groups = self._get_non_skipped_groups(section)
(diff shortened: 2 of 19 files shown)


bitdivision:
@dwyeradam - I've reverted the change to use `navigable_groups` for the moment. It was breaking the tests because interstitials were previously being shown in navigation, but weren't after the change.

I don't know if we rely on that functionality anywhere, but it was specifically included in the tests, so I assume interstitials were intended to show up in navigation.

MebinAbraham (commented):
This needs changing to match the original path:
https://github.com/ONSdigital/eq-survey-runner/blob/281342aa6fabb9e4138cf4b94a83c031e9f64e57/data/en/lms_2.json#L735

https://github.com/ONSdigital/eq-survey-runner/blob/281342aa6fabb9e4138cf4b94a83c031e9f64e57/data/en/lms_2.json#L811

MebinAbraham (dismissed):
This is the no-primary actual hours, but it references `two-jobs-j1-usual-hours-answer` instead of `no-primary-two-jobs-j1-usual-hours-answer`, which results in a 500 error.

https://github.com/ONSdigital/eq-survey-runner/blob/281342aa6fabb9e4138cf4b94a83c031e9f64e57/data/en/lms_2.json#L13910-L13924

Same thing for `two-jobs-j2-usual-hours-answer`.
Think there are quite a few routing rules which reference the wrong version for the id.

MebinAbraham (commented):
Some routing seems to be invalid/missing.

The overtime questions are not being asked when:
- Non-primary
- 1 Job Working
- Some overtime

ajmaddaford (approved):
I'm happy that this works as expected for LMS. It's not very clear what the intentions are behind the checking of dependent groups/blocks/drivers and their relevance to getting answers. This is something we need to review in detail post LMS.

dwyeradam (approved):
Good enough for a Friday.

PR #1847, opened 2018-10-22
Add group instance to year field validation

bitdivision:
LMS was using 'minimum' validation on two questions, which validated against dates from other answers.

Since the date validation used a group_instance of 0, we were always getting the value from the first repeat not the current repeat.

The same logic needs to be added to all of the fields when we have time. ([trello](https://trello.com/c/3zgV8Aod/2503-make-form-validation-repeating-group-aware))

### How to review 
- Add two people, first person born in 2016, born outside uk.
- Check the `first-arrive-in-uk` question (When did you most recently arrive in the UK) only lets you say you arrived after 2015.
- Second born in e.g. 1980.
- born outside UK
- ensure `first-arrive-in-uk` allows you to enter a date after 1980 e.g. 1981 

- Check that `camyr2` question behaves the same way.

### Checklist

### DIFF
--- a/app/forms/date_form.py
+++ b/app/forms/date_form.py
@@ -36,8 +36,8 @@ class DateField(FormField):
 
 class MonthYearField(FormField):
 
-    def __init__(self, answer_store, metadata, answer, error_messages, **kwargs):
-        form_class = get_month_year_form(answer_store, metadata, answer, error_messages)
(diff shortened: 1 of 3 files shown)


MebinAbraham (approved):
Seems to work fine. Maybe a test?

insacuri (approved):
Tested it and it works 👍

bitdivision:
Rebased, Unit tested, functional tested, validated, linted.

@ajmaddaford

PR #1848, opened 2018-10-23
Skip shared facilities question if less than 2 members

MebinAbraham:
Doesn't make sense to ask if the facilities are shared when there is only 1 household member.
This PR adds a skip condition to ensure that these are only asked if there are at least 2 members in the household.

### How to review 
Shared facilities questions should only be asked when you add at least 2 members to the household.
[Slides](https://docs.google.com/presentation/d/1BWjHrAfjxgyR69Jwys4BPahm9JF6eUbz2lIkd1se-TQ/edit#slide=id.g4267003c02_2_29)

### Checklist

### DIFF
--- a/data/en/lms_2.json
+++ b/data/en/lms_2.json
@@ -589,6 +589,17 @@
                                 "id": "primary-household-member-address-check-answer",
                                 "condition": "not set"
                             }]
+                        }, {
+                            "when": [{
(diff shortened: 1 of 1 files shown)


bitdivision (approved):
Tested, looks good.

insacuri (approved):
Tested and it works!

PR #1861, opened 2018-10-25
Add RoutingPath object to optimize path lookups

nealedj:
During performance investigation it was found that a significant amount of time was spent in checking if a given `Location` object is present in the built routing path (a list of locations). On average each call to run this would be `O(n)`.

Given that the routing path list is static once generated it's possible to easily create a wrapper for the list to store a set alongside it which should average `O(1)`.

During testing this took about 10ms off of the average response time.

NB - it wasn't possible to simply return the routing path as a `set` because:
a) the path is ordered
b) locations may need to appear twice (for routing backwards)

### How to review 
- General application check functions correctly
- Does routing backwards work?
- Check the code for logical issues 

### Checklist

### DIFF
--- a/app/questionnaire/location.py
+++ b/app/questionnaire/location.py
@@ -20,7 +20,7 @@ class Location:
         return isinstance(other, Location) and self.__dict__ == other.__dict__
 
     def __hash__(self):
-        return hash(self.__dict__.values())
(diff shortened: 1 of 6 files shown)


nealedj:
Sorry, premature PR.

PR #1863, opened 2018-10-26
Ensure all schemas have default routing rules

bitdivision:

This is associated with https://github.com/ONSdigital/eq-schema-validator/pull/99

I've updated all of our schemas to include default routing rules in all cases.

Most of the changes here apply to questions that are mandatory and include exhaustive routing rules, but we should still include a default.

Two (live?) surveys are affected, `lms_1.json` and `1_0005.json`. These should be checked carefully.

For `1_0005.json` I've added defaults to be on the safe side (rather than removing when rules).

### Checklist

### DIFF
--- a/tests/app/questionnaire/test_path_finder.py
+++ b/tests/app/questionnaire/test_path_finder.py
@@ -366,21 +366,6 @@ class TestPathFinder(AppContextTestCase):  # pylint: disable=too-many-public-met
 
         self.assertEqual(next_location, expected_next_location)
 
-    def test_interstitial_post_blocks(self):
-        schema = load_schema_from_params('0', 'star_wars')
-
-        answer = Answer(
-            answer_id='choose-your-side-answer',
-            value='Light Side'
-        )
-
-        answers = AnswerStore()
-        answers.add(answer)
-
-        path_finder = PathFinder(schema, answer_store=answers, metadata={}, completed_blocks=[])
-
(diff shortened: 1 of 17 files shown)


tests/app/questionnaire/test_path_finder.py
```diff
@@ -366,21 +366,6 @@ def test_next_location_empty_routing_rules(self):
 
         self.assertEqual(next_location, expected_next_location)
 
-    def test_interstitial_post_blocks(self):
```
bitdivision, line 369: This test fails after adding default routes to star_wars

I can't see any reason that the summary shouldn't be on the path. It seems this has only been passing because of the missing default route. There are also no interstitials in star wars.

scripts/test_schemas.sh
```diff
@@ -6,8 +6,8 @@ if [ "$1" == "--local" ] || [ "$2" == "--local" ]; then
 fi
 
 if [ "$run_docker" == true ]; then
-    docker pull onsdigital/eq-schema-validator
-    validator="$(docker run -d -p 5001:5000 onsdigital/eq-schema-validator)"
+    docker pull onsdigital/eq-schema-validator:validate-default-routing-rules
```
jonnyshaw89: reminder here.

jonnyshaw89 (approved):

MebinAbraham (approved):

PR #1872, opened 2018-10-31
Playback unselected options

dwyeradam:
New filter that will take two lists and return a list of missing items

As a user
I want a playback/confirmation of the response options I didn't select from the previous question
So that I can confirm that I selected the correct options in the previous question

### How to review 
In 1_0005.json it currently plays back answered Pay Periods. Changing the description of `pay-pattern-frequency-confirmation-question` with the following will allow the opposite to be shown.
`"description": "{{ format_unordered_list_missing_items(['Weekly','Fortnightly','Calendar monthly','Four weekly','Five weekly'], answers['pay-pattern-frequency-answer']) }}",` 

### Checklist

### DIFF
--- a/app/jinja_filters.py
+++ b/app/jinja_filters.py
@@ -350,6 +350,17 @@ def format_unordered_list(context, list_items):
     return mark_safe(context, summary_list)
 
 
+@evalcontextfunction
+def format_unordered_list_missing_items(context, possible_items, list_items):
+
+    if list_items and list_items[0]:
+        missing_items = [item for item in possible_items if item not in list_items[0]]
+    else:
+        missing_items = possible_items
+
+    return format_unordered_list(context, [missing_items])
+
+
 @blueprint.app_template_filter()
 def concatenated_list(list_items, delimiter=', '):
     stripped_items = list(item.strip() for item in list_items if item)
--- a/tests/app/test_jinja_filters.py
+++ b/tests/app/test_jinja_filters.py
@@ -13,10 +13,10 @@ from app.jinja_filters import (
     format_multilined_string, format_percentage, format_date_range,
     format_household_member_name, format_datetime,
     format_number_to_alphabetic_letter, format_unit, format_currency_for_input,
-    format_number, format_unordered_list, format_unit_input_label,
-    format_household_member_name_possessive, concatenated_list,
-    calculate_years_difference, get_current_date, as_london_tz, max_value,
-    min_value, get_question_title, get_answer_label,
+    format_number, format_unordered_list, format_unordered_list_missing_items,
+    format_unit_input_label, format_household_member_name_possessive,
+    concatenated_list, calculate_years_difference, get_current_date, as_london_tz,
+    max_value, min_value, get_question_title, get_answer_label,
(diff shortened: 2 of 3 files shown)


app/jinja_filters.py
```diff
@@ -350,6 +350,24 @@ def format_unordered_list(context, list_items):
...
+@evalcontextfunction
+def format_unordered_list_missing_items(context, possible_items, list_items):
+    summary_list = ''
+
+    if list_items and list_items[0]:
+        not_answered = [item for item in possible_items if item not in list_items[0]]
+    else:
+        not_answered = possible_items
+
+    if not_answered:
```
bitdivision: Could we avoid some duplication here by having a filter to subtract a list from another list, and then pass that into `format_unordered_list`? Feel like it would make unit tests a little nicer too since we wouldn't be checking any HTML.

Obviously would make using the filter a little more complex though.

tests/app/test_jinja_filters.py
```diff
@@ -552,6 +552,44 @@ def test_format_unordered_list_with_empty_list(self):
 
         self.assertEqual('', formatted_value)
 
+    def test_format_unordered_list_missing_items(self):
+        possible_items = ['item 1', 'item 2', 'item 3', 'item 4']
+        list_items = [['item 1', 'item 3']]
```
MebinAbraham, line 557: Would `selected_items` make more sense in this context?

app/jinja_filters.py
```diff
@@ -350,6 +350,24 @@ def format_unordered_list(context, list_items):
     return mark_safe(context, summary_list)
 
 
+@evalcontextfunction
+def format_unordered_list_missing_items(context, possible_items, list_items):
+    summary_list = ''
+
+    if list_items and list_items[0]:
+        not_answered = [item for item in possible_items if item not in list_items[0]]
```
MebinAbraham: If this variable is `not_answered` then it makes sense to change `list_items` to `selected_items` or `selected_answered` or something like that?

MebinAbraham (commented):
What is the expected behaviour when a user clicks all the options as it would have nothing to display?.
Do we skip the negative playback question?

app/jinja_filters.py
dwyeradam: Actually I should probably call this `missing_items` to match filter name.

app/jinja_filters.py
dwyeradam: Yes, I prefer that and I don't feel that it really complicates the filter. However if I subtract one list from another the order of the list could get easily changed which I didn't want to do.

dwyeradam: I didn't do a subtraction though as you'd lose the list order while using set().

dwyeradam:
With regard to _"What is the expected behaviour when a user clicks all the options as it would have nothing to display?. Do we skip the negative playback question?"_. That's really a question around individual schema routing and not a consideration of this filter.

app/jinja_filters.py
bitdivision: Ah, I actually meant having two separate filters and using them both in the schema. But this looks good.

bitdivision (approved):

MebinAbraham (approved):

PR #1877, opened 2018-11-01
Remove Flask login which adds user_id to cookie

dwyeradam:
PR #1657 causes the user_id to get stored in the session cookie.

EQ's current encryption model for questionnaire state intends for the user_id to not be exposed to the user, only the user_ik.

This [line](https://github.com/ONSdigital/eq-survey-runner/commit/92f54df0160c2f15aa2afcc1024847de769a2fc8#diff-38d589711ca2d8ef50e47c9f2cf2ae0fR116) was only added as an optimisation to save an extra DB hit and therefore I have removed it.

### How to review 
The cookie can be verified by copying the value of a session cookie to https://www.kirsle.net/wizards/flask-session.cgi and decoding it).

### Checklist

### DIFF
--- a/tests/integration/integration_test_case.py
+++ b/tests/integration/integration_test_case.py
@@ -2,9 +2,12 @@ import os
 import re
 import unittest
 import json
+import zlib
 
 from bs4 import BeautifulSoup
 from moto import mock_dynamodb2
+
+from itsdangerous import base64_decode
 
 from sdc.crypto.key_store import KeyStore
 
@@ -193,6 +196,15 @@ class IntegrationTestCase(unittest.TestCase):  # pylint: disable=too-many-public
         """
         return self.last_response.get_data(True)
 
+    def getCookie(self):
+        """
+            Returns the last received response cookie session
+        """
+        cookie = self.last_response.headers['Set-Cookie']
+        cookie_session = cookie.split('session=.')[1].split(';')[0]
+        decoded_cookie_session = decode_flask_cookie(cookie_session)
+        return json.loads(decoded_cookie_session)
+
     def getHtmlSoup(self):
         """
         Returns the last received response data as a BeautifulSoup HTML object
@@ -268,3 +280,11 @@ class IntegrationTestCase(unittest.TestCase):  # pylint: disable=too-many-public
             self.assertRegex(text=self.last_url, expected_regex=regex)
         else:
             self.fail('last_url is invalid')
+
+
+def decode_flask_cookie(cookie):
+    """Decode a Flask cookie."""
+    data = cookie.split('.')[0]
+    data = base64_decode(data)
+    data = zlib.decompress(data)
+    return data.decode('utf-8')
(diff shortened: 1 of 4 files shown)


tests/integration/integration_test_case.py
```diff
@@ -268,3 +280,21 @@ def assertRegexUrl(self, regex):
             self.assertRegex(text=self.last_url, expected_regex=regex)
         else:
             self.fail('last_url is invalid')
+
+
+def decode_flask_cookie(cookie):
+    """Decode a Flask cookie."""
```
dwyeradam, line 286: This function seems like overkill to me. Can I do it more simply, ideally with existing EQ code or inbuilt Flask functions?

bitdivision:
No idea what's happened with codecov there...

tests/integration/integration_test_case.py
```diff
@@ -2,10 +2,13 @@
 import re
 import unittest
 import json
+import zlib
 
 from bs4 import BeautifulSoup
 from moto import mock_dynamodb2
 
+from itsdangerous import base64_decode
```
jonnyshaw89, line 10: What is `itsdangerous`? we could use..
```
import base64
base64.b64decode(data)
```

jonnyshaw89 (changes requested):

jonnyshaw89:
This is obviously not needed anymore then either.
https://github.com/ONSdigital/eq-survey-runner/pull/1657/files#diff-4d5ff7e65b71e6875dfead2645e34354R12

tests/integration/integration_test_case.py
```diff
@@ -268,3 +280,21 @@ def assertRegexUrl(self, regex):
...
+    payload = cookie
+
+    if payload.startswith('.'):
+        compressed = True
+        payload = payload[1:]
+
+    data = payload.split('.')[0]
+
+    data = base64_decode(data)
+    if compressed:
```
jonnyshaw89: If our cookie is always compressed then we should always try and decompress, therefore we can tell if we suddenly stop compressign it

jonnyshaw89 (changes requested):

tests/integration/integration_test_case.py
dwyeradam: My understanding is that Flask, by default, uses the URL-safe signed serializer "itsdangerous" to encode its client-side session cookies and therefore so did I.

tests/integration/integration_test_case.py
bitdivision: I don't think so, the flask code is all wrapped up. You have to override a bunch of stuff and I don't think it's really worth the effort.

dwyeradam: That's the conclusion I came to so happy if I didn't miss something.

bitdivision (approved):

jonnyshaw89 (approved):

PR #1887, opened 2018-11-05
Compress and remove base64 encoding on answer store

nealedj:
Tests showed that the latency of some datastores increased significantly with the payload of the data being saved to/retrieved from it. DynamoDB slowed down almost linearly with the size of the payload.

This PR seeks to reduce that payload using two changes:
- Compress the questionnaire store prior to encryption
- No longer base64-encode the data prior to encryption (this didn't appear to be beneficial and increased the size of the payload)

The changes have been implemented to be backwards compatible. Existing questionnaire stores (base64-encoded and not compressed) should be still be readable by the application but will be written back using the new format.

### How to review 
- Manually test the application
- Attempt to resume a survey that was started using the old format and ensure that it can still be completed

### Checklist

### DIFF
--- a/app/data_model/session_store.py
+++ b/app/data_model/session_store.py
@@ -1,4 +1,5 @@
 import simplejson as json
+from jwcrypto.common import base64url_decode
 from structlog import get_logger
 
 from app.data_model.app_models import EQSession
@@ -87,6 +88,14 @@ class SessionStore:
                 encrypted_session_data = self._eq_session.session_data
                 session_data = StorageEncryption(self.user_id, self.user_ik, self.pepper) \
                     .decrypt_data(encrypted_session_data)
+
+                session_data = session_data.decode()
+                # for backwards compatibility
+                # session data used to be base64 encoded before encryption
+                try:
+                    session_data = base64url_decode(session_data).decode()
+                except ValueError:
+                    pass
 
                 self.session_data = json.loads(session_data, object_hook=lambda d: SessionData(**d))
 
--- a/app/storage/encrypted_questionnaire_storage.py
+++ b/app/storage/encrypted_questionnaire_storage.py
@@ -1,6 +1,8 @@
-import simplejson as json
+import json
+import snappy
 
 from structlog import get_logger
+from jwcrypto.common import base64url_decode
 
 from app.data_model.app_models import QuestionnaireState
 from app.storage import data_access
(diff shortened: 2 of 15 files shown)


app/storage/encrypted_questionnaire_storage.py
```diff
@@ -18,28 +20,40 @@ def __init__(self, user_id, user_ik, pepper):
...
         if questionnaire_state:
-            data = json.loads(questionnaire_state.state_data)
             version = questionnaire_state.version or 0
 
-            if 'data' in data:
-                decrypted_data = self.encrypter.decrypt_data(data['data'])
-                return decrypted_data, version
+            try:
+                # legacy data was stored in a dict, base64-encoded, and not compressed
+                data = json.loads(questionnaire_state.state_data)['data']
```
jonnyshaw89: You can wrap this in a check if version is < 3 and bump the latest version https://github.com/ONSdigital/eq-survey-runner/blob/master/app/data_model/questionnaire_store.py#L9

jonnyshaw89 (changes requested):

jonnyshaw89:
Data size in DynamoDB looks good. LMS 5 person household creates a 10kb object. Compared to about 100kb before change

insacuri:
Tested 2 business surveys.  Both of them start to finish in one go; And another start (old version), save, restart (new version), finish.
EDIT:  Forgot to mention that everything worked as expected!

Not approving yet because I've not done more testing then that and I think it should be looked at a bit more throughly

insacuri (approved):
Tried lms submission, first with new compression all the way through, then another starting with the old version, then starting in the middle with the new version.
Everything worked fine.

bitdivision:
I'll fix the versioning on here

app/storage/encrypted_questionnaire_storage.py
```diff
@@ -18,28 +20,35 @@ def __init__(self, user_id, user_ik, pepper):
...
         questionnaire_state = self._find_questionnaire_state()
-        if questionnaire_state:
-            data = json.loads(questionnaire_state.state_data)
+        if questionnaire_state and questionnaire_state.state_data:
             version = questionnaire_state.version or 0
 
-            if 'data' in data:
-                decrypted_data = self.encrypter.decrypt_data(data['data'])
-                return decrypted_data, version
+            if questionnaire_state.version < 3:
```
bitdivision: Should we use version as above here? I guess that `or 0` was there from the original upgrade?

app/data_model/session_store.py
```diff
@@ -88,6 +89,14 @@ def _load(self):
                 session_data = StorageEncryption(self.user_id, self.user_ik, self.pepper) \
                     .decrypt_data(encrypted_session_data)
 
+                session_data = session_data.decode()
+                # for backwards compatibility
+                # session data used to be base64 encoded before encryption
+                try:
+                    session_data = base64url_decode(session_data).decode()
```
jonnyshaw89, line 96: This can be removed soon after deployment?

app/views/questionnaire.py
```diff
@@ -304,9 +306,18 @@ def get_view_submission(schema, eq_id, form_type):  # pylint: disable=unused-arg
...
 
             pepper = current_app.eq['secret_store'].get_secret_by_name('EQ_SERVER_SIDE_STORAGE_ENCRYPTION_USER_PEPPER')
+
             encrypter = StorageEncryption(current_user.user_id, current_user.user_ik, pepper)
+            submitted_data = encrypter.decrypt_data(submitted_data.data)
+
+            # for backwards compatibility
+            # submitted data used to be base64 encoded before encryption
+            try:
+                submitted_data = base64url_decode(submitted_data.decode()).decode()
```
jonnyshaw89, line 316: this can be removed after deployment too.

jonnyshaw89 (approved):

app/views/questionnaire.py
ajmaddaford: Issue raised to track the removal - https://github.com/ONSdigital/eq-survey-runner/issues/1898

bitdivision (approved):
Looks good :+1:

PR #1888, opened 2018-11-05
Add light caching to completeness module

nealedj:
The `Completeness` object handles some reasonably costly logic. It derives states for sections, groups and blocks; each type's state being composed from its children's states.

At some points during the request cycle the methods which calculate the states for these types were found to be getting called multiple times from different callers. This PR allows the `Completeness` object to persist its states and avoid this unnecessary processing.

### How to review 
Check that the application still functions correctly. Pay particular attention to schemas which use navigation, skip conditions and/or routing.

### Checklist

### DIFF
--- a/app/questionnaire/completeness.py
+++ b/app/questionnaire/completeness.py
@@ -22,6 +22,10 @@ class Completeness:
         self.metadata = metadata
         self.schema = schema
 
+        self._block_states = {}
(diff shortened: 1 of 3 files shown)


bitdivision:
Going to leave this for now. Likely do a review of completeness in general later on.

PR #1889, opened 2018-11-05
Cache template paths

nealedj:
The https://github.com/sysr-q/flask-themes2 library looks for a template given the current theme before falling back to a default. For each file required to render a response it walks the template directories searching for a matching template file.

This change caches the path to each template. On testing it represented an average 37ms drop in CPU time for GET requests (about 33% at that point). 

Note: this has nothing to do with the actual template cache that gets used once flask themes has identified the correct template.

### How to review 
General app testing. Ensure that schemas using different themes are tested without restarting the app.

### Checklist

### DIFF
--- a/app/flask_theme_cache.py
+++ b/app/flask_theme_cache.py
@@ -1,0 +1,19 @@
+import jinja2
+import flask_themes2
+
+
+def get_global_theme_template(cache):
+    @cache.memoize()
+    def _get_templatepath(theme, templatename, fallback):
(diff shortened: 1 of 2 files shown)


jonnyshaw89 (approved):

ajmaddaford (commented):
Is it worth considering adding this functionality to Flask-Themes?

bitdivision (approved):
I think we should get this in and then raise an issue / PR on flask_themes2.

ajmaddaford (approved):

PR #1890, opened 2018-11-05
Group answers by answer ID in answer store

nealedj:
This PR adds a dict mapping of answer ID to a list of answers to the `AnswerStore` object. This makes calls to `AnswerStore.filter()` more efficient, especially with a large number of answers.

This reduced the CPU time by around 10ms for both GETs and POSTs during profiling. At this point the answer store was relatively small and a greater improvement would be expected for requests later in a large survey or for ones with a large number of repeating groups. 

### How to review 
Check code makes sense and there is no potential for the new `answer_map` object to get out of sync with the main list of answers.

Manually test application.

### Checklist

### DIFF
--- a/app/data_model/answer_store.py
+++ b/app/data_model/answer_store.py
@@ -1,5 +1,7 @@
+import itertools
 import re
 
+from collections import defaultdict
 from datetime import datetime
 from jinja2 import escape
 import simplejson as json
@@ -59,6 +61,7 @@ class AnswerStore:
 
     def __init__(self, existing_answers=None):
         self.answers = existing_answers or []
+        self.answer_map = self._build_map(self.answers)
 
     def __iter__(self):
         return iter(self.answers)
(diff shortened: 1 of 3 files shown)


app/data_model/answer_store.py
```diff
@@ -59,16 +61,32 @@ class AnswerStore:
 
     def __init__(self, existing_answers=None):
         self.answers = existing_answers or []
+        self.answer_map = self._build_map(self.answers)
```
jonnyshaw89, line 64: I think we should consider removing the `self.answers` list completly and just ising the map.
We should also look at the gains of storing the map in the datastore rather than converting a map to and from a list

jonnyshaw89 (changes requested):

bitdivision:
Closing in favour of https://github.com/ONSdigital/eq-survey-runner/pull/1892 which removes the list entirely.

PR #1891, opened 2018-11-05
Add RoutingPath object to optimize path lookups

nealedj:
During performance investigation it was found that a significant amount of time was spent in checking if a given `Location` object is present in the built routing path (a list of locations). On average each call to run this would be `O(n)`.

Given that the routing path list is static once generated it's possible to easily create a wrapper for the list to store a set alongside it which should average `O(1)`.

During testing this took about 10ms off of the average response time.

NB - it wasn't possible to simply return the routing path as a `set` because:
a) the path is ordered
b) locations may need to appear twice (for routing backwards)

### How to review 
- General application check functions correctly
- Does routing backwards work?
- Check the code for logical issues 

### Checklist

### DIFF
--- a/app/questionnaire/location.py
+++ b/app/questionnaire/location.py
@@ -20,7 +20,7 @@ class Location:
         return isinstance(other, Location) and self.__dict__ == other.__dict__
 
     def __hash__(self):
-        return hash(self.__dict__.values())
+        return hash(frozenset(self.__dict__.values()))
 
     def __str__(self):
         """
--- a/app/questionnaire/path_finder.py
+++ b/app/questionnaire/path_finder.py
@@ -3,6 +3,7 @@ from structlog import get_logger
 
 from app.helpers.schema_helpers import get_group_instance_id
 from app.questionnaire.location import Location
+from app.questionnaire.routing_path import RoutingPath
 from app.questionnaire.rules import (
     evaluate_goto,
     evaluate_skip_conditions,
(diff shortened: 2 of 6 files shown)


jonnyshaw89:
To Solve the routing backwards issue we should remove the extra locations from the end of the list.
eg..
```
[1, 2, 3, 4, 3]
```
Would become 
```
[1, 2, 3]
```

This would both shorten the path and remove the needs for duplicates in a set

insacuri:
Tested it with a simple survey (102) and it works fine but I wasn't seeing the gains mentioned.  I'll test it with something more complicated with more pages and it should hopefully be more apparent

bitdivision:
Switching to an OrderedSet would currently cause some issues with confirmation questions, requiring path_finder to be refactored quite substantially. Will raise a ticket to look at that and for moving the logic around confirmation questions out of the routing path.

bitdivision (approved):
I've tested this a fair bit and not seen any issues. Performance doesn't seem to change much with small surveys but did seem to make a difference on LMS w/ 10 people. I haven't benchmarked the performance fully though.

MebinAbraham (approved):
Looks good. Have noticed a gain in performance when testing with large routing paths (large households).

ajmaddaford (approved):

bitdivision:
Don't think Codacy is going to work here...

PR #1892, opened 2018-11-07
Group answers by answer ID in answer store - mk2

jonnyshaw89:
My variation on https://github.com/ONSdigital/eq-survey-runner/pull/1890
Removed the answers list completly so we only hold a map.
This removed the complexity around maintaining two stores

### How to review 
Does it work? is it faster?

### DIFF
--- a/.ebextensions/enable_mod_deflate.conf
+++ b/.ebextensions/enable_mod_deflate.conf
@@ -23,10 +23,9 @@ BrowserMatch ^Mozilla/4\.0[678] no-gzip
 
 # MSIE masquerades as Netscape, but it is fine
 BrowserMatch \bMSI[E] !no-gzip !gzip-only-text/html
+</IfModule>
 
 <IfModule mod_headers.c>
 # Make sure proxies don't deliver the wrong content
 Header append Vary User-Agent env=!dont-vary
-</IfModule>
-
 </IfModule>
--- a/app/data_model/answer_store.py
+++ b/app/data_model/answer_store.py
@@ -1,52 +1,14 @@
-import re
+import itertools
 
+from collections import defaultdict
 from datetime import datetime
 from jinja2 import escape
-import simplejson as json
 from structlog import get_logger
+import simplejson as json
+
+from app.data_model.answer import Answer
 
 logger = get_logger()
-
-
-class Answer:
-    def __init__(self, answer_id, value, group_instance_id=None, group_instance=0, answer_instance=0):
-        if answer_id is None or value is None:
-            raise ValueError("Both 'answer_id' and 'value' must be set for Answer")
-
-        self.answer_id = answer_id
-        self.group_instance_id = group_instance_id
-        self.group_instance = group_instance
-        self.answer_instance = answer_instance
-        self.value = value
-
-    def matches(self, answer):
-        """
-        Check to see if two answers match.
-        Two answers are considered to match if they share the same answer_id, answer_instance and group_instance_id.
-
-        :param answer: An answer to compare
-        :return: True if both answers match, otherwise False.
-        """
-        return self.answer_id == answer.answer_id and \
-            self.group_instance_id == answer.group_instance_id and \
-            self.group_instance == answer.group_instance and \
-            self.answer_instance == answer.answer_instance
-
-    def matches_dict(self, answer_dict):
-        """
-        Check to see if a dict describes an answer the same as this object.
-
-        :param answer_dict: A dictionary representation of the answer.
-        :return: True if both answers match, otherwise False.
-        """
-
-        return self.matches(Answer(
-            answer_dict.get('answer_id'),
-            answer_dict.get('value'),
-            answer_dict.get('group_instance_id'),
-            answer_dict.get('group_instance', 0),
-            answer_dict.get('answer_instance', 0),
-        ))
 
 
 class AnswerStore:
@@ -58,55 +20,39 @@ class AnswerStore:
     EQ_MAX_NUM_REPEATS = 25
 
     def __init__(self, existing_answers=None):
-        self.answers = existing_answers or []
+        if isinstance(existing_answers, list):
+            self.answer_map = self._build_map(existing_answers or [])
+        else:
+            self.answer_map = existing_answers or defaultdict(list)
 
     def __iter__(self):
-        return iter(self.answers)
-
-    def __getitem__(self, key):
-        return self.answers[key]
+        return iter((answer for answers in self.answer_map.values() for answer in answers))
 
     def __len__(self):
-        return len(self.answers)
+        return sum(len(answers) for answers in self.answer_map.values())
+
+    def __eq__(self, other):
+        return self.answer_map == other.answer_map
+
+    @staticmethod
+    def _build_map(answers):
+        answer_map = defaultdict(list)
+
+        for answer in answers:
+            answer_map[answer['answer_id']].append(answer)
+
+        return answer_map
 
     @staticmethod
     def _validate(answer):
         if not isinstance(answer, Answer):
             raise TypeError('Method only supports Answer argument type')
 
-    def add(self, answer):
-        """
-        Add a new answer into the answer store.
-
-        :param answer: A dict of flattened answer details.
-        """
-        self._validate(answer)
-
-        answer_to_add = answer.__dict__.copy()
-
-        if self.find(answer) is not None:
-            raise ValueError('Answer instance already exists in store')
-        else:
-            self.answers.append(answer_to_add)
-
     def copy(self):
         """
         Create a new instance of answer_store with the same values.
         """
-        return self.__class__(existing_answers=self.answers.copy())
-
-    def update(self, answer):
-        """
-        Update the value of an answer already in the answer store.
-
-        :param answer: A dict of flattened answer details.
-        """
-        position = self.find(answer)
-
-        if position is None:
-            raise ValueError('Answer instance does not exist in store')
-        else:
-            self.answers[position]['value'] = answer.value
+        return self.__class__(self.answer_map.copy())
 
     def add_or_update(self, answer):
         """
@@ -114,10 +60,14 @@ class AnswerStore:
 
         :param answer: An answer object.
         """
-        try:
-            self.update(answer)
-        except ValueError:
-            self.add(answer)
+        self._validate(answer)
+        position = self.find(answer)
+
+        if position is None:
+            answer_to_add = vars(answer).copy()
+            self.answer_map[answer_to_add['answer_id']].append(answer_to_add)
+        else:
+            self.answer_map[answer.answer_id][position]['value'] = answer.value
 
     def find(self, answer):
         """
@@ -128,9 +78,10 @@ class AnswerStore:
         """
         self._validate(answer)
 
-        for index, existing in enumerate(self.answers):
-            if answer.matches_dict(existing):
-                return index
+        if answer.answer_id in self.answer_map:
+            for index, existing in enumerate(self.answer_map[answer.answer_id]):
+                if answer.matches_dict(existing):
+                    return index
 
         return None
 
@@ -151,7 +102,7 @@ class AnswerStore:
 
         :return: Return a list of answer values
         """
-        return [answer['value'] for answer in self.answers]
+        return [answer['value'] for answer in self]
 
     def escaped(self):
         """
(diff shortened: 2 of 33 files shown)


.ebextensions/enable_mod_deflate.conf
```diff
@@ -23,10 +23,14 @@ BrowserMatch ^Mozilla/4\.0[678] no-gzip
...
 BrowserMatch \bMSI[E] !no-gzip !gzip-only-text/html
+</IfModule>
 
 <IfModule mod_headers.c>
 # Make sure proxies don't deliver the wrong content
 Header append Vary User-Agent env=!dont-vary
 </IfModule>
 
-</IfModule>
+# One day for most static assets
```
bitdivision: Do we ever need to invalidate any of these assets?

app/data_model/answer_store.py
```diff
@@ -57,17 +19,26 @@ class AnswerStore:
...
-        self.answers = existing_answers or []
+    def __init__(self, existing_answers=None, answer_map=None):
+        self.answer_map = answer_map or self._build_map(existing_answers or [])
 
     def __iter__(self):
-        return iter(self.answers)
-
-    def __getitem__(self, key):
-        return self.answers[key]
+        return iter([answer for answers in self.answer_map.values() for answer in answers])
```
bitdivision: Not sure if it's of any real benefit, but this could be a generator I think

app/data_model/questionnaire_store.py
```diff
@@ -52,14 +52,17 @@ def _deserialise(self, data):
...
+        else:
+            self.answer_store = AnswerStore(existing_answers=json_data.get('ANSWERS', []))
         self.completed_blocks = completed_blocks
         self.collection_metadata = json_data.get('COLLECTION_METADATA', {})
 
     def _serialise(self):
         data = {
             'METADATA': self._metadata,
-            'ANSWERS': self.answer_store.answers,
+            'ANSWERS': list(iter(self.answer_store)),
```
bitdivision: Should be able to get rid of the `iter` I believe.

```suggestion
            'ANSWERS': list(self.answer_store),
```

app/forms/fields.py
```diff
@@ -262,10 +262,10 @@ def get_schema_defined_limit(answer_id, definition, answer_store):
         else:
             source_answer_id = definition.get('answer_id')
             answer_list = answer_store.filter(answer_ids=[source_answer_id])
-            value = answer_list[0].get('value')
+            value = next(iter(answer_list)).get('value')
             if not isinstance(value, int) and not isinstance(value, Decimal):
                 raise Exception('answer: {} value: {} for answer id: {} is not a valid number'
-                                .format(answer_list[0].get('answer_id'), value, answer_id))
+                                .format(source_answer_id, value, answer_id))
```
bitdivision, line 268: :+1:

app/helpers/form_helper.py
```diff
@@ -171,3 +170,22 @@ def get_mapped_answers(schema, answer_store, block_id, group_instance, group_ins
...
+    Converts number to an integer if possible.
+    :param text: Text to be converted
+    :return: Integer representation of text if a number.  Otherwise, it returns the text as is
+    """
+    return int(text) if text.isdigit() else text
+
+
+def natural_order(key):
+    """
+    Orders a set of items according to
```
bitdivision: This docstring and function name are wrong I think? it shouldn't be changing the order?

app/setup.py
```diff
@@ -153,8 +153,11 @@ def before_request():  # pylint: disable=unused-variable
 
     @application.after_request
     def apply_caching(response):  # pylint: disable=unused-variable
-        for k, v in CACHE_HEADERS.items():
-            response.headers[k] = v
+        if 'text/html' in response.content_type:
```
bitdivision: :+1:

app/templating/schema_context.py
```diff
@@ -42,7 +42,8 @@ def _build_answers(answer_store, schema, answer_ids_on_path):
             value = _create_answers_list(matching_answers, 'group_instance')
 
         elif matching_answers:
-            value = json_and_html_safe(matching_answers[0]['value'])
+            for answer in matching_answers:
+                value = json_and_html_safe(answer['value'])
```
bitdivision: Is it worth adding a method for getting the first and only item? It seems as though there are a few situations where we are assuming that there is only one item in the store. Not sure whether we want to actually assert that at runtime though.

bitdivision (changes requested):
Minor things. Looks good. Need to test performance against the original branch and master I think.

app/data_model/answer_store.py
```diff
@@ -87,13 +58,16 @@ def add(self, answer):
         if self.find(answer) is not None:
             raise ValueError('Answer instance already exists in store')
         else:
-            self.answers.append(answer_to_add)
+            if answer_to_add['answer_id'] not in self.answer_map:
+                self.answer_map[answer_to_add['answer_id']] = []
+
+            self.answer_map[answer_to_add['answer_id']].append(answer_to_add)
```
ajmaddaford: I don't think we need to have an add method - it's only used in the tests and we're duplicating the check for an answer existing in both the update and add methods.

.ebextensions/enable_mod_deflate.conf
jonnyshaw89: The app version is appended so every deploy is unique urls

app/data_model/answer_store.py
jonnyshaw89: i cant get that to work 🤷‍♂️

app/data_model/answer_store.py
```diff
@@ -87,13 +58,16 @@ def add(self, answer):
         if self.find(answer) is not None:
             raise ValueError('Answer instance already exists in store')
         else:
-            self.answers.append(answer_to_add)
+            if answer_to_add['answer_id'] not in self.answer_map:
+                self.answer_map[answer_to_add['answer_id']] = []
```
ajmaddaford: This isn't needed as the answer_map is initialised with `defaultdict(list)`

app/data_model/answer_store.py
```diff
@@ -198,13 +178,14 @@ def filter(self, answer_ids=None, group_instance=None, group_instance_id=None, a
...
                     break
+
         return self.__class__(existing_answers=filtered)
 
     def clear(self):
         """
         Clears answers *in place*
         """
-        self.answers.clear()
+        self.answer_map = defaultdict(list)
```
ajmaddaford: Why not `self.answer_map.clear()`?

app/questionnaire/rules.py
```diff
@@ -211,7 +211,7 @@ def _get_answers_on_path(answers, schema, routing_path) -> AnswerStore:
     answers_on_path = answers.copy()
 
     for answer in answers_to_remove:
-        answers_on_path.answers.remove(answer)
+        answers_on_path.remove(answer_ids=[answer.get('answer_id')], group_instance=answer.get('group_instance'), answer_instance=answer.get('answer_instance'))
```
ajmaddaford: It might be worth adding a `remove_answer` method to answer_store that takes an `Answer` object so we don't have to filter the store for each individual answer.

app/views/questionnaire.py
```diff
@@ -115,7 +114,7 @@ def before_post_submission_request():
 
 @questionnaire_blueprint.after_request
 def add_cache_control(response):
-    response.cache_control.no_cache = True
+    # response.cache_control.no_cache = True
```
ajmaddaford: Does this need to be removed?

ajmaddaford (changes requested):

.ebextensions/enable_mod_deflate.conf
```diff
@@ -23,10 +23,14 @@ BrowserMatch ^Mozilla/4\.0[678] no-gzip
...
 # MSIE masquerades as Netscape, but it is fine
 BrowserMatch \bMSI[E] !no-gzip !gzip-only-text/html
+</IfModule>
 
 <IfModule mod_headers.c>
 # Make sure proxies don't deliver the wrong content
 Header append Vary User-Agent env=!dont-vary
 </IfModule>
 
-</IfModule>
```
bitdivision, line 32: Ideally this should be in a seperate PR as discussed.

app/data_model/answer_store.py
jonnyshaw89: This also removes duplicated checks when adding a new answer

tests/app/submitter/test_convert_payload_0_0_2.py
```diff
@@ -344,13 +344,16 @@ def test_date_answer(self):
...
-            self.assertEqual(answer_object['data'][1]['answer_instance'], 0)
+            sorted_data = sorted(answer_object['data'], key=lambda t: t['value'])
+
+            self.assertEqual(sorted_data[1]['value'], '01-1990')
+            self.assertEqual(sorted_data[1]['group_instance'], 0)
+            self.assertEqual(sorted_data[1]['answer_instance'], 0)
+
+            self.assertEqual(sorted_data[0]['value'], '01-01-1990')
+            self.assertEqual(sorted_data[0]['group_instance'], 0)
+            self.assertEqual(sorted_data[0]['answer_instance'], 0)
```
ajmaddaford: I think it would be better to split this test into two methods - `test_date_answer` and `test_month_year_date_answer` - then the ordering won't be a concern. This is how it's done for the other answer types.

app/data_model/answer_store.py
```diff
@@ -57,17 +19,26 @@ class AnswerStore:
...
     def __iter__(self):
-        return iter(self.answers)
-
-    def __getitem__(self, key):
-        return self.answers[key]
+        return iter((answer for answers in self.answer_map.values() for answer in answers))
 
     def __len__(self):
-        return len(self.answers)
+        return len([answer for answers in self.answer_map.values() for answer in answers])
```
nealedj: Could be slightly more efficient as:

```
sum(len(answers) for answers in self.answer_map.values())
```

.ebextensions/enable_mod_deflate.conf
jonnyshaw89: just the fix for incorrect nesting left

jonnyshaw89:
@ajmaddaford @bitdivision @nealedj I think i've addressed all the comments

bitdivision:
Looks good. Just going to do some testing.

app/views/questionnaire.py
```diff
@@ -659,16 +651,8 @@ def _evaluate_skip_conditions(block_json, location, schema, answer_store, metada
...
-
-    if matches:
-        index, answer_id = matches.groups()
-    else:
-        answer_id = answer_instance_id
-        index = 0
-
-    return answer_id, int(index)
+def get_answer_instance_id(answer_id, answer_index):
+    return 'household-{}-{}'.format(answer_index, answer_id)
```
ajmaddaford, line 655: Should answer_index and answer_id be the other way around in the format string?

app/data_model/questionnaire_store.py
```diff
@@ -52,14 +52,17 @@ def _deserialise(self, data):
         completed_blocks = [Location.from_dict(location_dict=completed_block) for completed_block in
                             json_data.get('COMPLETED_BLOCKS', [])]
         self.set_metadata(json_data.get('METADATA', {}))
-        self.answer_store.answers = json_data.get('ANSWERS', [])
+        if isinstance(json_data.get('ANSWERS'), dict):
+            self.answer_store = AnswerStore(answer_map=json_data.get('ANSWERS'))
+        else:
+            self.answer_store = AnswerStore(existing_answers=json_data.get('ANSWERS', []))
```
ajmaddaford: Should this be done via the questionnaire store version?

ajmaddaford (changes requested):

app/views/questionnaire.py
jonnyshaw89: No, it generates something like `household-1-firstname`
screenshot 2018-11-09 at 13 44 46

bitdivision (approved):
:tada:

ajmaddaford (approved):
👍

app/views/questionnaire.py
```diff
@@ -308,7 +307,7 @@ def get_view_submission(schema, eq_id, form_type):  # pylint: disable=unused-arg
             encrypter = StorageEncryption(current_user.user_id, current_user.user_ik, pepper)
 
             submitted_data = json.loads(encrypter.decrypt_data(submitted_data.data))
-            answer_store = AnswerStore(existing_answers=submitted_data.get('answers'))
+            answer_store = AnswerStore(answer_map=submitted_data.get('answers'))
```
ajmaddaford: We'll need this to deal with the old answer store format for a short period of time after deploying these changes.

ajmaddaford (changes requested):

ajmaddaford (approved):

PR #1896, opened 2018-11-08
Remove functional test pages and generate each time

bitdivision:
The functional test pages of selectors were meant to be auto generated but were often manually modified. Changes to generate_pages.py often broke previous pages, but the pages weren't regenerated so this wasn't noticed until later.

To fix this, generate_pages.py has been modified to generate the pages before each test run. The run_tests_functional.sh script now deletes and recreates the generated_pages directory.

Originally I had planned to refactor generate_pages.py. I think that should probably be done as a seperate PR it likely needs to be rethought quite a bit.

### How to review 
Run all the scripts locally and make sure they make sense.
Check that I haven't broken any functionality in the tests.

### Checklist


dwyeradam (approved):
Tedious to check but good positive change. I also like the splitting of some of the larger spec files into their component parts. 
My favourite change by far is `.click(RadioButtonsPage.newMethodsOfOrganisingExternalRelationshipsWithOtherFirmsOrPublicInstitutions())`
to `.click(RadioButtonsPage.maybe())`

README.md
```diff
@@ -142,17 +142,23 @@ Upgrade usage of the pattern library
...
+This will delete the `tests/functional/generated_pages` directory and regenerate all the files in it from the schemas.
 
-The functional tests can be executed with:
+To generate the pages manually you can run the `generate_pages` scripts with the schema directory. Run it from the `tests/functional` directory as follows:
+
+`./generate_pages.py ../../data/en/ ./generated_pages -r "../../base_pages"`
+
+To generate a spec file with the imports included, you can use the `generate_pages.py` script on a single schema with the `-s` argument.
+
+`./generate_pages.py ../../data/en/test_multiple_piping.json ./temp_directory -r "../../base_pages" -s specs/test_multiple_piping.spec.js
```
dwyeradam: Not a blocker by any measure but we're missing the ` from the end of line. Also this command doesn't work unless the folder locations already exist.

MebinAbraham (changes requested):
We should update this in the readme:
`Runs the functional tests through ChimpJS (requires app running on localhost:5000)`

Need to also generate the pages. This is only mentioned further down the readme so we should either explain it here as well or only have 1 reference for running tests etc.

README.md
bitdivision: Sorry about that! Both fixed now.

README.md
```diff
@@ -142,17 +142,23 @@ Upgrade usage of the pattern library
...
+This will delete the `tests/functional/generated_pages` directory and regenerate all the files in it from the schemas.
 
-The functional tests can be executed with:
+To generate the pages manually you can run the `generate_pages` scripts with the schema directory. Run it from the `tests/functional` directory as follows:
+
+`./generate_pages.py ../../data/en/ ./generated_pages -r "../../base_pages"`
+
+To generate a spec file with the imports included, you can use the `generate_pages.py` script on a single schema with the `-s` argument.
+
+`./generate_pages.py ../../data/en/test_multiple_piping.json ./temp_directory -r "../../base_pages" -s specs/test_multiple_piping.spec.js`
```
MebinAbraham: This should be `spec/` not `specs/`.

MebinAbraham (changes requested):
☝️

MebinAbraham (commented):
Ignore - Wasn't worth it

MebinAbraham (approved):

bitdivision:
@dwyeradam I've changed the scripts a little, but it's not a significant change so I haven't dismissed your review

PR #1897, opened 2018-11-09
Add browser caching for static assets

jonnyshaw89:
Allow browser to cache static assets.
A unique value is appended to the end of the url for static assets so browser cache shouldn't be an issue between application versions

### How to review 
Run functional tests and see that static assets are not being requested on every request
Run the app and check only static assets are cached.

### Checklist

### DIFF
--- a/.ebextensions/enable_mod_deflate.conf
+++ b/.ebextensions/enable_mod_deflate.conf
@@ -29,3 +29,8 @@ BrowserMatch \bMSI[E] !no-gzip !gzip-only-text/html
 # Make sure proxies don't deliver the wrong content
 Header append Vary User-Agent env=!dont-vary
 </IfModule>
+
+# One day for most static assets
+<filesMatch ".(css|jpg|jpeg|png|gif|js|ico)$">
+Header set Cache-Control "max-age=86400, public"
+</filesMatch>
--- a/app/setup.py
+++ b/app/setup.py
@@ -157,8 +157,11 @@ def create_app(setting_overrides=None):  # noqa: C901  pylint: disable=too-compl
 
     @application.after_request
     def apply_caching(response):  # pylint: disable=unused-variable
-        for k, v in CACHE_HEADERS.items():
-            response.headers[k] = v
+        if 'text/html' in response.content_type:
+            for k, v in CACHE_HEADERS.items():
+                response.headers[k] = v
+        else:
+            response.headers['Cache-Control'] = 'max-age=2628000, public'
 
         return response
 
--- a/app/views/questionnaire.py
+++ b/app/views/questionnaire.py
@@ -114,13 +114,6 @@ def before_post_submission_request():
                        session_collection_id='')
 
 
-@questionnaire_blueprint.after_request
-def add_cache_control(response):
-    response.cache_control.no_cache = True
-
-    return response
-
-
 def save_questionnaire_store(func):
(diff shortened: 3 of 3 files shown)


bitdivision (approved):

insacuri (approved):
LGTM, server logs look a lot quieter whilst running functional tests

.ebextensions/enable_mod_deflate.conf
```diff
@@ -23,10 +23,14 @@ BrowserMatch ^Mozilla/4\.0[678] no-gzip
...
 <IfModule mod_headers.c>
 # Make sure proxies don't deliver the wrong content
 Header append Vary User-Agent env=!dont-vary
 </IfModule>
 
-</IfModule>
+# One day for most static assets
+<filesMatch ".(css|jpg|jpeg|png|gif|js|ico)$">
+Header set Cache-Control "max-age=86400, public"
+</filesMatch>
```
MebinAbraham, line 36: Not a blocker as it works but is this still needed if the app is setting it?. And why is the expiry of static assets by app and eb different?. Is the one for external assets?.

MebinAbraham (approved):

ajmaddaford (changes requested):
Doesn't this mean that if you're developing on the static files that your changes won't be displayed in the browser?

.ebextensions/enable_mod_deflate.conf
ajmaddaford: The application doesn't serve the static assets in Elastic Beanstalk, Apache does. This is the Apache config to enable the caching.

MebinAbraham:
@ajmaddaford It has to be recompiled anyway?. We have browsersync for development, which can auto refresh on change.

ajmaddaford:
@MebinAbraham As it uses the git hash for versioning it doesn't change once built unless something has been committed. My understanding is that browsersync will cause the browser to reload, but the browser will then pull the static resources from it's cache.

bitdivision:
@ajmaddaford @MebinAbraham I looked into this a bit as I was slightly confused. We're using `setup.py:versioned_url_for` for this. As Andrew says, it does just include the version.

To make this work with caching I think we would need to add the cache buster at build time (using timestamp or file hash) which is easy enough (e.g. [gulp-cache-bust](https://www.npmjs.com/package/gulp-cache-bust)). The issue is then getting the cache busted filename. I've done this before using a manifest file, but it's not an ideal solution.

MebinAbraham:
@ajmaddaford @bitdivision 
When running browser sync it appends a new query string value each time the file has changes, meaning that caching shouldn't be an issue.

`http://localhost:5000/s/css/census/theme.min.css?q=b85dd4f15945fc10f5d5a17cfd608f126d2de1e8&browsersync=1542099343953`


You can test this by running `yarn start` then make a change to css. It doesn't seems to be working currently for js/images. It does detect them but doesn't append a query string, can investigate why. From the looks of it, it is meant to be working.

ajmaddaford:
So on looking into this further browsersync is working for Javascript as well as css (it doesn't append anything in a query string, but the Javascript does reload). Images aren't reloading on changes, but if any work is done on these it's easy enough to turn off caching in the browser dev tools.

ajmaddaford (approved):

jonnyshaw89:
🎉

PR #1905, opened 2018-11-16
Don't save state in new format as it won't work during deployment

ajmaddaford:

As there is a chance of requests going to a new version of the app
and then back to an old version during deployment, we can't persist
any format changes to state in the next release. This release should
however have all the code required to understand the new state
formats, which are:
  - Questionnaire store compressed via snappy, and base64
    encoding and json data wrapper removed
  - Removal of base64 encoding for session data
  - View submission data in answer map format

So, to be clear, anything written (questionnaire state, session or
view submission data) in this version of the app should be readable
by the currently released version (v2.73.0).

### DIFF
--- a/app/data_model/questionnaire_store.py
+++ b/app/data_model/questionnaire_store.py
@@ -10,10 +10,8 @@ class QuestionnaireStore:
     The QuestionnaireStore is versioned to provide a way to migrate existing state. Versions are:
     1 - Reformat date answers from d/m/y to y-m-d
     2 - Add group_instance_id to all answers
-    3 - Compress state using snappy before encryption. Also removes base64 encoding and
-        unnecessary json wrapper around encrypted data.
     """
-    LATEST_VERSION = 3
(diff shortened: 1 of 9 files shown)


jonnyshaw89:
👍

bitdivision:
Code looks good, apart from the unforgivable import order :)

I'll do some testing upgrading / downgrading.

bitdivision (approved):
Tested mid survey upgrade from v2.73.0 -> this branch
mid survey downgrade from this branch -> v2.73.0

Tested save / complete later with both versions
Tested submission on each version.

No issues.

dwyeradam (approved):
Managed to create issues going between master and v2.73.0.
Could not recreate the same issue going between this branch and v2.73.0.
👍

bitdivision (approved):
I'm happy.

dwyeradam (approved):
New tests look fine to me.

PR #1910, opened 2018-11-21
Flask Babel issue with dates < 1902.

bitdivision:
The original issue stems from: [trello](https://trello.com/c/Fy8nWytY/2575-issue-piping-incorrect-date). If a date was passed to format_date before December 14 1901 then format_date would change the hour by -1, meaning that 1902 would become 1901 when piped into another question.

I originally thought we needed to set `rebase=False` on `flask_babel.format_date`, this did fix the issue, but we were not converting dates to BST anymore, breaking a test. 

Investigating further, I found that 1901 coincides with the earliest possible unix timestamp, e.g. minimum value of a signed 32 bit integer. If you run `zdump -v Europe/London` you can actually see a couple of BST transitions in 1901, which shouldn't really be there, since it was introduced in 1916. I found a few vague references to similar problems on stack overflow which suggested issues with the Olsen database (used by `pytz`).

One fix for the issue was to fork flask-babel (fork [here](https://github.com/ONSdigital/flask-babel/tree/replace-pytz-with-pendulum) can be removed if this is merged.) and replace their [use of pytz](https://github.com/python-babel/flask-babel/blob/master/flask_babel/__init__.py#L23) with [pendulum](https://github.com/sdispater/pendulum). Pendulum fixes the issue found in `pytz` with dates before 1902 (and potentially a few others). This solution is problematic because flask-babel isn't well maintained and we are unlikely to be able to merge any changes upstream. The issue would probably need to be fixed upstream in `pytz`, which lives on sourceforge and has a mailing list. 

An alternate fix is to skirt the issue in flask babel by avoiding the [`to_user_timezone`](https://github.com/python-babel/flask-babel/blob/master/flask_babel/__init__.py#L360) function in flask_babel, which is where the issue comes into play. I did this by calling `flask_babel.format_datetime` with a `date` object rather than a `datetime` object. This fixes our use of format_date and doesn't rely on maintaining a fork of flask_babel. I think it's the better solution in the short term.

As a part of this, I did have a look at replacing our use of datetime with pendulum, which on the face of it looks like it would fix some common datetime issues we have. I found that the `now()` method is a local time object and there is no `utcnow()` method, additionally, the datetime.strptime() function is not replicated exactly (different format strings etc.). Having investigated a little, I don't think there are any major advantages *for us* switching at the moment. Though it could become useful in future if we start doing complex manipulations of timezones etc.

### How to review 
Enter a date of e.g. January 1 1901 on the test_dates schema. The playback on the next page should show 1901, not December 1900. This is replicated in the functional test.

### Checklist

### DIFF
--- a/tests/app/test_jinja_filters.py
+++ b/tests/app/test_jinja_filters.py
@@ -144,6 +144,19 @@ class TestJinjaFilters(AppContextTestCase):  # pylint: disable=too-many-public-m
         # Then
         self.assertEqual(format_value, "<span class='date'>1 January 2017</span>")
 
+    def test_format_old_date_does_not_change_timezone(self):
+        """ Flask Babel shows some strange behaviour with dates prior to 1902.
+        This appears to be because of a bug in pytz relating to 32 bit time_t types
+        To avoid it, `datetimes` are converted to `date` objects before being passed
+        to `flask_babel.format_date()`.
+        """
+        date = '1901-01-01'
+
+        with self.app_request_context('/'):
+            format_value = format_date(self.autoescape_context, date)
+
+        assert format_value.striptags() == '1 January 1901'
+
     def test_format_date_month_year(self):
         # Given
         date = '2017-01'
@@ -187,7 +200,7 @@ class TestJinjaFilters(AppContextTestCase):  # pylint: disable=too-many-public-m
         self.assertIsNone(format_value)
 
     def test_format_date_time_in_bst(self):
-        # Given
+        # Given a date after DST started
         date_time = '2018-03-29T11:59:13.528680'
 
         # When
--- a/.bandit
+++ b/.bandit
@@ -1,0 +1,2 @@
+[bandit]
+exclude: /tests,/node_modules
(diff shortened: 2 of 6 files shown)


tests/app/test_jinja_filters.py
```diff
@@ -144,6 +144,19 @@ def test_format_date(self):
         # Then
         self.assertEqual(format_value, "<span class='date'>1 January 2017</span>")
 
+    def test_format_old_date_does_not_change_timezone(self):
+        """ Flask Babel shows some strange behaviour with dates prior to 1902.
+        This appears to be because of a bug in pytz relating to 32 bit time_t types
+        To avoid it, the flask babel library was forked and pytz was replaced with
+        pendulum.
```
ajmaddaford: This docstring is now out of date

ajmaddaford (changes requested):
LGTM, just a small docstring issue.

tests/app/test_jinja_filters.py
bitdivision: Good spot. Fixed.

ajmaddaford (approved):

dwyeradam (approved):

bitdivision:
Should probably ensure that Codacy works before this is merged. Not sure how to trigger it though...

bitdivision:
Codacy finally seems to have run @ajmaddaford. Should be ready to merge.

PR #1917, opened 2018-11-28
Remove hardcoding of trad_as_or_ru_name

MebinAbraham:
Banished `trad_as_or_ru_name`!
The same behaviour is achieved through the use of a jinja filter `display_first_valid_item`, feel free to suggest a better name.

This filter can also be reused to return `one of` any number of arguments. The filter returns the first valid value. I didn't want to make this `trad_as` specifc, so for the purposes of `trad_as_or_ru_name`, `trad_as` must be passed before `ru_name`.

[Trello](https://trello.com/c/DNQvLV4f/2345-removing-hardcoding-of-tradasorruname-logic-from-runner-s)

***Author:***
*Spoke to @samiwel and he is happy for this to go in as is since it doesn't break author. A card also on thier backlog to allow this new style.
For the time being, authors will use `ru_name` instead. This ofcouse will mean, for the time being, a manual change is required.*

### How to review 
Ensure:
- New filter allows same behaviours as before without the hardcoded logic.
- All references to the old style has been replaced with the new style. ***(Affects live surveys)***
- All `trad_as_or_ru_name` logic is removed from runner.

### Checklist

### DIFF
--- a/app/jinja_filters.py
+++ b/app/jinja_filters.py
@@ -5,11 +5,10 @@ from datetime import datetime
 
 import flask
 import flask_babel
+from babel import units, numbers
 from dateutil import relativedelta, tz
 from jinja2 import Markup, contextfunction, escape, evalcontextfilter, evalcontextfunction, Undefined
 from jinja2.exceptions import UndefinedError
-
-from babel import units, numbers
 
 from app.questionnaire.rules import convert_to_datetime
 
@@ -182,23 +181,14 @@ def format_datetime(context, value):
 
 @evalcontextfunction
 @blueprint.app_template_filter()
-def format_conditional_date(context, date1=None, date2=None):
+def format_conditional_date(context, *dates):
     """
-    This function format_conditional_date accepts two dates, a user submitted date and a metadata date
-
-    :param date1 user entered date
-    :param date2 is metadata date
-    :return: the value of the date to be piped
+    This is to be used when passing multiple dates as multiple arguments.
+    :param dates: tuple of dates
+    :return: a formatted date used for piping
     """
-    if date1:
-        date = date1
-    else:
-        date = date2
-
-    if date is None:
-        raise Exception('No valid dates passed to format_conditional_dates filter')
-
-    return format_date(context, date)
+    first_valid_date = _get_first_non_empty_item(dates)
+    return format_date(context, first_valid_date)
 
 
 @blueprint.app_template_filter()
@@ -309,6 +299,43 @@ def format_date_range_no_repeated_month_year(context, start_date, end_date, date
                                  to_date=format_date_custom(context, end_date, date_format))
 
     return mark_safe(context, output)
+
+
+def _get_first_non_empty_item(items):
+    """
+    :param items: anything that is iterable
+    :return: first non empty value
+
+    In this filter the following values are considered non empty:
+    - None
+    - any empty sequence, for example, '', (), [].
+    - any empty mapping, for example, {}.
+
+     Note: to guarantee the returned element is actually the first non empty element in the iterable,
+    'items' must be a data structure that preserves order, ie tuple, list etc.
+    If order is not important, this can be reused to return `one of` the elements which is non empty.
+
+    This filter will treat zero of any numeric type for example, 0, 0.0, 0j and boolean 'False'
+    as a valid item since they are naturally 'falsy' in Python but not empty.
+
+    Note: Booleans are a subtype of integers. Zero of any numeric type 'is not False' but 'equals False'.
+    Reference: https://docs.python.org/release/3.4.2/library/stdtypes.html?highlight=boolean#boolean-values
+    """
+    for item in items:
+        if item or item is False or item == 0:
+            return item
+
+    raise Exception('No valid items provided.')
+
+
+@evalcontextfunction
+def first_non_empty_item(context, *items):
+    """
+    :param items: tuple of items
+    :return: mark safe version of the first non empty item
+    """
+    non_empty_item = _get_first_non_empty_item(items)
+    return mark_safe(context, non_empty_item)
 
 
 @blueprint.app_template_filter()
@@ -535,6 +562,11 @@ def format_number_processor():
 @blueprint.app_context_processor
 def language_urls_processor():
     return dict(language_urls=language_urls)
+
+
+@blueprint.app_context_processor
+def format_conditional_item_processor():
+    return dict(first_non_empty_item=first_non_empty_item)
 
 
 def mark_safe(context, value):
--- a/app/templating/template_renderer.py
+++ b/app/templating/template_renderer.py
@@ -1,8 +1,8 @@
 # coding: utf-8
 
 import re
-import simplejson as json
 
+import simplejson as json
 from jinja2 import Environment
 
 import app.jinja_filters as filters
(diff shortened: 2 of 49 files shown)


app/jinja_filters.py
```diff
@@ -311,6 +310,15 @@ def format_date_range_no_repeated_month_year(context, start_date, end_date, date
     return mark_safe(context, output)
 
 
+@evalcontextfunction
+def one_of(context, *args):
+    one_of_exists = [x for x in args if x]
```
jonnyshaw89: Does this guarantee order?
What if both are supplied?

app/jinja_filters.py
```diff
@@ -311,6 +310,15 @@ def format_date_range_no_repeated_month_year(context, start_date, end_date, date
     return mark_safe(context, output)
 
 
+@evalcontextfunction
+def one_of(context, *args):
```
jonnyshaw89: Could you also refactor `format_conditional_date` to reduce duplication?

app/jinja_filters.py
MebinAbraham: As long as the data structure preserves order. In our case yes, it should guarantee that it is the first valid element.

app/jinja_filters.py
MebinAbraham: Made some changes.

app/jinja_filters.py
```diff
@@ -182,23 +181,15 @@ def format_datetime(context, value):
...
 @blueprint.app_template_filter()
-def format_conditional_date(context, date1=None, date2=None):
+def format_conditional_date(context, *dates):
     """
-    This function format_conditional_date accepts two dates, a user submitted date and a metadata date
-
-    :param date1 user entered date
-    :param date2 is metadata date
-    :return: the value of the date to be piped
+    This is to be used when passing multiple dates as multiple arguments.
```
insacuri, line 186: Could you re-add the param information for the dates?

insacuri (approved):
Just the 1 documentation change and it looks good to go!

insacuri (changes requested):
Just the 1 documentation change and it looks good to go!

app/jinja_filters.py
```diff
@@ -311,6 +302,30 @@ def format_date_range_no_repeated_month_year(context, start_date, end_date, date
...
+
+    Note: to guarantee the returned element is actually the first truthy element in the iterable,
+    'items' must be a data structure that preserves order, ie tuple, list etc.
+    If order is not important, this can be reused to return `one of` the elements which is truthy.
+    """
+    return next((item for item in items if item), None)
+
+
+@evalcontextfunction
+def display_first_valid_item(context, *items):
```
bitdivision: This could be `display_first_truthy_item`. IMO it conveys more information.

MebinAbraham: Can change, just thought it was too specific to be an acceptable name. Maybe it just needs a doc string here aswell?. Check doc string on `_get_first_valid_item`.

bitdivision: A doc string does work, but still means you'd have to look up the filter to ensure you knew what it does.

I don't think there's any reason not to be specific in the name of a filter? It will only ever be used in a jinja2 template, which was designed for Python, so `truthy` isn't out of place I don't think.

bitdivision: Maybe someone might use this on an array of numbers for instance, in which case:

input of `[0, 0.1, 0.2]`

would display 0.1, even though `0` is 'valid' to me.

MebinAbraham: Good point. ^

app/templating/template_renderer.py
```diff
@@ -11,27 +11,31 @@
...
         env.filters['format_household_name'] = filters.format_household_member_name
         env.filters['format_household_name_possessive'] = filters.format_household_member_name_possessive
         env.filters['format_household_summary'] = filters.format_household_summary
+        env.filters['format_number'] = filters.format_number
+        env.filters['format_repeating_summary'] = filters.format_repeating_summary
         env.filters['format_unordered_list'] = filters.format_unordered_list
-        env.globals['format_unordered_list_missing_items'] = filters.format_unordered_list_missing_items
+        env.filters['get_currency_symbol'] = filters.get_currency_symbol
+
+        env.globals['calculate_offset_from_weekday_in_last_whole_week'] = filters.calculate_offset_from_weekday_in_last_whole_week
```
bitdivision, line 26: :+1:

bitdivision (dismissed):

MebinAbraham:
@insacuri Can you re-review?. Thanks

insacuri (dismissed):
LGTM. Just needs a squash.

MebinAbraham:
@bitdivision @insacuri 
Hey guys, could you have another look at this, changes to the logic has changed. Zeros are now treated as a valid item since this filter can potentially be used for scenarios where 0 would be a valid value.

app/jinja_filters.py
```diff
@@ -311,6 +301,37 @@ def format_date_range_no_repeated_month_year(context, start_date, end_date, date
...
+    'items' must be a data structure that preserves order, ie tuple, list etc.
+    If order is not important, this can be reused to return `one of` the elements which is non empty.
+
+    This filter treats zeros as a valid item.
+    """
+    for item in items:
+        if item:
+            return item
+
+        if item is not False and item == 0:
```
bitdivision: This line probably deserves a comment. Esp. given that `if item is not False and item == False:` would work too.

tests/app/test_jinja_filters.py
```diff
@@ -1008,3 +1005,52 @@ def test_format_address_list_concatenated_list_no_values(self):
             format_address_list(answer_address, metadata_address)
 
         self.assertEqual('No valid address passed to format_address_list filter', error.exception.args[0])
+
+    def test_first_non_empty_item_filter_returns_first_non_empty_item(self):
+        # Given
+        expected_actual_scenarios = [
+            ['only value', (Markup('only value'),)],
```
bitdivision, line 1012: You could use [pytest parameterize](https://docs.pytest.org/en/latest/parametrize.html) here. Not worth changing, but worth bearing in mind in future.

app/jinja_filters.py
MebinAbraham: @bitdivision Have added some more context around it.

tests/app/test_jinja_filters.py
MebinAbraham: Ah this is exactly what I needed.

bitdivision: Not sure if it will actually work until we remove the unittest.TestCase inheritance.

bitdivision (approved):

app/jinja_filters.py
```diff
@@ -311,6 +301,45 @@ def format_date_range_no_repeated_month_year(context, start_date, end_date, date
...
+    If order is not important, this can be reused to return `one of` the elements which is non empty.
+    """
+    for item in items:
+        if item:
+            return item
+
+        # The below statement will treat zero of any numeric type for example, 0, 0.0, 0j as a valid item since they are naturally 'falsy' in Python.
+        # Note: Booleans are a subtype of integers. Zero of any numeric type 'is not False' but 'equals False'.
+        # Reference: https://docs.python.org/release/3.4.2/library/stdtypes.html?highlight=boolean#boolean-values
+        if item is not False and item == 0:
```
dwyeradam: Splitting this into a separate if seems excessive. Also this wouldn't deal with a list of booleans (not that I can imagine it would ever need to). Wouldn't it be simpler to change this to:
```
    for item in items:
        if item or item is False or item == 0:
            return item

    raise Exception('No valid items provided.')
```

dwyeradam (changes requested):

bitdivision (approved):

app/jinja_filters.py
```diff
@@ -306,25 +306,23 @@ def _get_first_non_empty_item(items):
...
-            return item
-
-        # The below statement will treat zero of any numeric type for example, 0, 0.0, 0j as a valid item since they are naturally 'falsy' in Python.
-        # Note: Booleans are a subtype of integers. Zero of any numeric type 'is not False' but 'equals False'.
-        # Reference: https://docs.python.org/release/3.4.2/library/stdtypes.html?highlight=boolean#boolean-values
-        if item is not False and item == 0:
+        if item or item is False or item == 0:
             return item
 
     raise Exception('No valid items provided.')
```
dwyeradam, line 328: I don't know the answer of this but starting the conversation.
Do we actually want to raise an exception here or just return None?

bitdivision: Discussed a bit in person, but I think this should still raise an exception. The behaviour of the filter isn't really defined for that situation, and I don't think it can be in a reasonable way, since the use cases are varied (Do you want an empty string / "Unknown" / None etc).

I would suggest passing a list of things with no 'valid' values is an improper use of the filter, indicating a schema error. I don't think we can reasonably expect to catch an error like this in validator, and would therefore expect an exception to be raised.

dwyeradam (approved):

dwyeradam (approved):

PR #1920, opened 2018-11-29
Any Option to allow additional detail input

dwyeradam:
Allow runner to provide respondents with more than 1 write in field for a response option so that they can give more detail about the response option they have chosen.

### How to review 
Existing "other" options still work as expected.
Additional features in:
`test_checkbox_multiple_detail_answers.json` and
`test_radio_mutlitple_detail_answers.json`

### Checklist

### DIFF
--- a/app/forms/fields.py
+++ b/app/forms/fields.py
@@ -1,10 +1,11 @@
 from decimal import Decimal
 from flask_babel import gettext as _
 from structlog import get_logger
-from wtforms import FormField, SelectField, SelectMultipleField, StringField
+from wtforms import FormField, SelectField, StringField
 from wtforms import validators
 
-from app.forms.custom_fields import MaxTextAreaField, CustomIntegerField, CustomDecimalField
+from app.forms.custom_fields import MaxTextAreaField, CustomIntegerField, CustomDecimalField, \
+    CustomSelectMultipleField, CustomSelectField
 from app.forms.date_form import MonthYearField, YearField, \
     DateField
 from app.forms.duration_form import get_duration_form
@@ -17,27 +18,27 @@ MAX_DECIMAL_PLACES = 6
 logger = get_logger()
 
 
-def get_field(answer, label, error_messages, answer_store, metadata, group_instance=0):
+def get_field(answer, label, error_messages, answer_store, metadata, group_instance=0, disable_validation=False):
     guidance = answer.get('guidance', '')
 
     if answer['type'] in ['Number', 'Currency', 'Percentage', 'Unit']:
-        field = get_number_field(answer, label, guidance, error_messages, answer_store)
+        field = get_number_field(answer, label, guidance, error_messages, answer_store, disable_validation)
     elif answer['type'] == 'Date':
         field = get_date_field(answer, label, guidance, error_messages, answer_store, metadata)
     elif answer['type'] == 'MonthYearDate':
         field = get_month_year_field(answer, label, guidance, error_messages, answer_store, metadata, group_instance=group_instance)
     elif answer['type'] == 'YearDate':
         field = get_year_field(answer, label, guidance, error_messages, answer_store, metadata, group_instance=group_instance)
-    elif answer['type'] == 'Checkbox':
-        field = get_select_multiple_field(answer, label, guidance, error_messages)
+    elif answer['type'] == 'Duration':
+        field = get_duration_field(answer, label, guidance, error_messages)
     else:
         field = {
+            'Checkbox': get_select_multiple_field,
             'Radio': get_select_field,
             'TextArea': get_text_area_field,
             'TextField': get_string_field,
             'Dropdown': get_dropdown_field,
-            'Duration': get_duration_field,
-        }[answer['type']](answer, label, guidance, error_messages)
+        }[answer['type']](answer, label, guidance, error_messages, disable_validation)
 
     return field
 
@@ -46,6 +47,14 @@ def build_choices(options):
     choices = []
     for option in options:
         choices.append((option['value'], option['label']))
+    return choices
+
+
+def build_choices_with_detail_answer_ids(options):
+    choices = []
+    for option in options:
+        detail_answer_id = option['detail_answer']['id'] if option.get('detail_answer') else None
+        choices.append((option['value'], option['label'], detail_answer_id))
     return choices
 
 
@@ -69,7 +78,7 @@ def get_length_validator(answer, error_messages):
 
 
 def get_mandatory_validator(answer, error_messages, mandatory_message_key):
-    validate_with = [validators.Optional()]
+    validate_with = validators.Optional()
 
     if answer['mandatory'] is True:
         mandatory_message = error_messages[mandatory_message_key]
@@ -78,16 +87,15 @@ def get_mandatory_validator(answer, error_messages, mandatory_message_key):
                 mandatory_message_key in answer['validation']['messages']:
             mandatory_message = answer['validation']['messages'][mandatory_message_key]
 
-        validate_with = [
-            ResponseRequired(
-                message=mandatory_message,
-            ),
-        ]
-    return validate_with
+        validate_with = ResponseRequired(message=mandatory_message)
+
+    return [validate_with]
 
 
-def get_string_field(answer, label, guidance, error_messages):
-    validate_with = get_mandatory_validator(answer, error_messages, 'MANDATORY_TEXTFIELD')
+def get_string_field(answer, label, guidance, error_messages, disable_validation=False):
+    validate_with = []
+    if disable_validation is False:
+        validate_with = get_mandatory_validator(answer, error_messages, 'MANDATORY_TEXTFIELD')
 
     return StringField(
         label=label,
@@ -96,9 +104,11 @@ def get_string_field(answer, label, guidance, error_messages):
     )
 
 
-def get_text_area_field(answer, label, guidance, error_messages):
-    validate_with = get_mandatory_validator(answer, error_messages, 'MANDATORY_TEXTAREA')
-    validate_with.extend(get_length_validator(answer, error_messages))
+def get_text_area_field(answer, label, guidance, error_messages, disable_validation=False):
+    validate_with = []
+    if disable_validation is False:
+        validate_with = get_mandatory_validator(answer, error_messages, 'MANDATORY_TEXTAREA')
+        validate_with.extend(get_length_validator(answer, error_messages))
 
     return MaxTextAreaField(
         label=label,
@@ -151,13 +161,15 @@ def get_duration_field(answer, label, guidance, error_messages):
     )
 
 
-def get_select_multiple_field(answer, label, guidance, error_messages):
-    validate_with = get_mandatory_validator(answer, error_messages, 'MANDATORY_CHECKBOX')
+def get_select_multiple_field(answer, label, guidance, error_messages, disable_validation=False):
+    validate_with = []
+    if disable_validation is False:
+        validate_with = get_mandatory_validator(answer, error_messages, 'MANDATORY_CHECKBOX')
 
-    return SelectMultipleField(
+    return CustomSelectMultipleField(
         label=label,
         description=guidance,
-        choices=build_choices(answer['options']),
+        choices=build_choices_with_detail_answer_ids(answer['options']),
         validators=validate_with,
     )
 
@@ -171,8 +183,10 @@ def _coerce_str_unless_none(value):
     return str(value) if value is not None else None
 
 
-def get_dropdown_field(answer, label, guidance, error_messages):
-    validate_with = get_mandatory_validator(answer, error_messages, 'MANDATORY_DROPDOWN')
+def get_dropdown_field(answer, label, guidance, error_messages, disable_validation=False):
+    validate_with = []
+    if disable_validation is False:
+        validate_with = get_mandatory_validator(answer, error_messages, 'MANDATORY_DROPDOWN')
 
     return SelectField(
         label=label,
@@ -183,8 +197,10 @@ def get_dropdown_field(answer, label, guidance, error_messages):
     )
 
 
-def get_select_field(answer, label, guidance, error_messages):
-    validate_with = get_mandatory_validator(answer, error_messages, 'MANDATORY_RADIO')
+def get_select_field(answer, label, guidance, error_messages, disable_validation=False):
+    validate_with = []
+    if disable_validation is False:
+        validate_with = get_mandatory_validator(answer, error_messages, 'MANDATORY_RADIO')
 
     # We use a custom coerce function to avoid a defect where Python NoneType
     # is coerced to the string 'None' which clashes with legitimate Radio field
@@ -193,23 +209,35 @@ def get_select_field(answer, label, guidance, error_messages):
     # https://github.com/ONSdigital/eq-survey-runner/issues/1013
     # See related WTForms PR: https://github.com/wtforms/wtforms/pull/288
 
-    return SelectField(
+    return CustomSelectField(
         label=label,
         description=guidance,
-        choices=build_choices(answer['options']),
+        choices=build_choices_with_detail_answer_ids(answer['options']),
         validators=validate_with,
         coerce=_coerce_str_unless_none,
     )
 
 
-def get_number_field(answer, label, guidance, error_messages, answer_store):
-    answer_errors = error_messages.copy()
+def get_number_field(answer, label, guidance, error_messages, answer_store, disable_validation=False):
+    validate_with = []
 
-    if 'validation' in answer and 'messages' in answer['validation']:
-        for error_key, error_message in answer['validation']['messages'].items():
-            answer_errors[error_key] = error_message
+    if disable_validation is False:
+        validate_with = _get_number_field_validators(answer, error_messages, answer_store)
 
-    mandatory_or_optional = get_mandatory_validator(answer, answer_errors, 'MANDATORY_NUMBER')
+    if answer.get('decimal_places', 0) > 0:
+        return CustomDecimalField(
+            label=label,
+            validators=validate_with,
+            description=guidance,
+        )
+    return CustomIntegerField(
+        label=label,
+        validators=validate_with,
+        description=guidance,
+    )
+
+
+def _get_number_field_validators(answer, error_messages, answer_store):
 
     max_decimals = answer.get('decimal_places', 0)
     if max_decimals > MAX_DECIMAL_PLACES:
@@ -232,27 +260,23 @@ def get_number_field(answer, label, guidance, error_messages, answer_store):
                         .format(max_value, MAX_NUMBER, answer['id']))
 
     if min_value > max_value:
-        raise Exception('min_value: {} > max_value: {} for answer id: {}'.format(min_value, max_value, answer['id']))
+        raise Exception(
+            'min_value: {} > max_value: {} for answer id: {}'.format(min_value, max_value, answer['id']))
 
-    validate_with = mandatory_or_optional + [
+    answer_errors = error_messages.copy()
+    if 'validation' in answer and 'messages' in answer['validation']:
+        for error_key, error_message in answer['validation']['messages'].items():
+            answer_errors[error_key] = error_message
+
+    mandatory_or_optional = get_mandatory_validator(answer, answer_errors, 'MANDATORY_NUMBER')
+
+    return mandatory_or_optional + [
         NumberCheck(answer_errors['INVALID_NUMBER']),
         NumberRange(minimum=min_value, minimum_exclusive=minimum_exclusive,
                     maximum=max_value, maximum_exclusive=maximum_exclusive,
                     messages=answer_errors, currency=answer.get('currency')),
         DecimalPlaces(max_decimals=max_decimals, messages=answer_errors),
     ]
-
-    if max_decimals > 0:
-        return CustomDecimalField(
-            label=label,
-            validators=validate_with,
-            description=guidance,
-        )
-    return CustomIntegerField(
-        label=label,
-        validators=validate_with,
-        description=guidance,
-    )
 
 
 def get_schema_defined_limit(answer_id, definition, answer_store):
--- a/app/forms/questionnaire_form.py
+++ b/app/forms/questionnaire_form.py
@@ -24,7 +24,7 @@ class QuestionnaireForm(FlaskForm):
         self.answer_store = answer_store
         self.metadata = metadata
         self.question_errors = {}
-        self.options_with_children = {}
+        self.options_with_detail_answer = {}
 
         super().__init__(**kwargs)
 
@@ -93,7 +93,7 @@ class QuestionnaireForm(FlaskForm):
     def validate_mutually_exclusive_question(self, question):
         is_mandatory = question.get('mandatory')
         messages = question['validation'].get('messages') if 'validation' in question else None
-        answers = (getattr(self, answer['id']).data for answer in question['answers'] if 'parent_answer_id' not in answer)
+        answers = (getattr(self, answer['id']).data for answer in question['answers'])
 
         validator = MutuallyExclusiveCheck(messages=messages)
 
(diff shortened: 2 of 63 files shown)


dwyeradam:
I've removed two tests from `test_question.py` as both seem to test scenarios that I can't see should ever happen but previously were accounted for.

`test_radio_button_other_option_empty` tries to default to "other" when the answer passed in is blank. How can you pass a blank answer into a radio? This is different to just not answering.

`test_radio_button_other_option_text` is trying to pass in an answer that isn't in the options. Again how could that happen.

I removed the code that would handle these situations so please check my understanding.

ajmaddaford (changes requested):
I think we can simplify this further. We're still creating a secondary list of all the detail answers and then remapping those back to the choice they're attached to. We shouldn't need to do this as the relationship is now explicit. 

For example, in `multiple_choice_widget.html` rather than:
```python
{% if answer.id in form.detail_answer and form.detail_answer[answer.id][loop.index0] != None %}
  {% set detail_answer = form.detail_answer[answer.id][loop.index0] %}
```
we should be able to do something like:
```python
{% if detail_answer in option %}
  {% set detail_answer = option.detail_answer %}
```

dwyeradam:
@ajmaddaford I can't see an easy way to join the `detail_answer` to an option without either writing our own `SelectMultipleField`/`SelectField` classes rather than using the wtforms standards or completely rewriting the templates to use the schema rather than the `form.fields` to build up the options. I may be wrong and i'll keep looking but this may turn into a larger piece of work than originally planned.

MebinAbraham (dismissed):
This may depend on the future of `q_code` etc but at the moment, the value of the parent is being written from the child answer (previously it was always just 'Other' hence it worked?). I used string q_codes to represent. 

```
data": {
      "cheese-details": "Cheese detail",
      "cheese-option": "Cheese detail",        //Should say `Cheese`?
      "pineapple-option": "Pineapple",
      "your-choice-option": "Your choice details"      //Should say `Your choice`?
    },
```

We should also enforce q_codes on `detail_answer`?

app/forms/questionnaire_form.py
```diff
@@ -253,33 +253,34 @@ def get_data(self, answer_id):
...
-                data[answer['parent_answer_id']] == 'Other':
-            answer['mandatory'] = \
-                next(a['mandatory'] for a in question['answers'] if a['id'] == answer['parent_answer_id'])
+
+        for option in answer.get('options', []):
+            if 'detail_answer' in option:
+                detail_answer = option['detail_answer']
+                detail_answer_error_messages = detail_answer['validation']['messages'] if detail_answer.get('validation') else error_messages
+                if option['value'] not in data.values():
+                    detail_answer['mandatory'] = False
```
nealedj: Minor thought here:

It looks like this mutates the `QuestionnaireSchema` object. This probably isn't an issue in survey runner right now because that object gets pickled and cached at the point it's loaded in `load_schema_from_params` (before any mutation).

Having to pickle this object in order to appears to have a performance impact (suspect it's even more significant with the LMS schema?)

For testing we tried storing it directly on the module and it sped things up. But it then means that all requests share the same instance so any mutation of the object could cause all kinds of headaches. In case that approach is useful to survey runner in future it might be worth treating `QuestionnaireSchema` as immutable.

This would be easier if [PEP351](https://www.python.org/dev/peps/pep-0351/) or [PEP416](https://www.python.org/dev/peps/pep-0416/) had gained traction...

dwyeradam: I'll try and find an alternative as I don't like this bit of code anyway. It's a simplification of what already existed. 

This code deals with the situation whereby if the additional text field associated with an option is mandatory it would check validation regardless of if that option was selected or not. This overrides that to say only validate its mandatory status if the associated option is chosen.

dwyeradam:
@MebinAbraham, I don't know what to do with your comment. At the moment I think my code follows the existing logic. I wouldn't want to change it without defined requirements. Since we don't currently have any live surveys that use the existing Child/Parent/Other and Q_Codes this won't affect anything either way. I also therefore doubt we'd get a definition of what is wanted downstream in these situations.

app/forms/fields.py
```diff
@@ -49,6 +50,14 @@ def build_choices(options):
     return choices
 
 
+def build_choices_with_detail_answer_ids(options):
```
ajmaddaford, line 53: I think this should just be `build_choices` as there won't always be detail answer ids

app/forms/questionnaire_form.py
```diff
@@ -253,33 +253,29 @@ def get_data(self, answer_id):
...
+
+                answer_fields[detail_answer['id']] = get_field(detail_answer, detail_answer.get('label', 'Please specify'),
```
ajmaddaford: Any reason not to make label mandatory so we can remove the default text?

app/forms/questionnaire_form.py
```diff
@@ -297,16 +293,16 @@ def map_subfield_errors(errors, answer_id):
...
+    detail_answer_errors = []
+    options_with_detail_answers = [o for o in answer_json['options']
+                                   if 'detail_answer' in o and o['detail_answer']['id'] in errors]
 
-    for option_with_child in options_with_children:
-        for error in errors[option_with_child['child_answer_id']]:
-            child_errors.append((answer_json['id'], error))
+    for option_with_detail_answer in options_with_detail_answers:
+        for error in errors[option_with_detail_answer['detail_answer']['id']]:
+            detail_answer_errors.append((answer_json['id'], error))
```
ajmaddaford: I think this would be easier to read with two loops rather than three:
```python
for option in answer_json['options']:
    if 'detail_answer' in option and option['detail_answer']['id'] in errors:
        for error in errors[option['detail_answer']['id']]:
            detail_answer_errors.append((answer_json['id'], error))
```

app/templating/summary/question.py
```diff
@@ -124,13 +110,22 @@ def _build_date_range_answer(self, answer_store, answer, group_instance):
...
+    def _build_radio_answer(self, answer, answer_schema, answer_store, group_instance):
+        label = None
         for option in answer_schema['options']:
-            if option['value'].lower() == 'other' and answer != option['value']:
-                return answer if answer else option['label']
             if answer == option['value']:
-                return option['label']
+                label = option['label']
+
+            if label:
```
ajmaddaford: `label` is required in the schema so we could remove the `label` variable and access `option['label']` directly below.

app/templating/view_context.py
```diff
@@ -64,7 +64,7 @@ def build_view_context_for_question(metadata, schema, answer_store, current_loca
             'mapped_errors': form.map_errors(),
             'answer_errors': {},
             'data': {},
-            'other_answer': {},
+            'detail_answer': {},
```
ajmaddaford: Is this needed now?

tests/functional/spec/components/checkbox/checkbox_multiple_detail_answers.spec.js
```diff
@@ -0,0 +1,98 @@
...
+      // When
+        .click(MandatoryCheckboxPage.yourChoice())
+        .click(MandatoryCheckboxPage.submit())
+      // Then
+        .isVisible(MandatoryCheckboxPage.error()).should.eventually.be.true
+        .getText(MandatoryCheckboxPage.errorNumber(1)).should.eventually.contain('Enter your topping choice to continue');
+    });
+  });
+
+  it('Given a mandatory checkbox answer, when there is an error on the page for other field and I enter valid value and submit page, then the error is cleared and I navigate to next page.s', function() {
```
ajmaddaford: Extra 's' at the end of the sentence.

tests/functional/spec/components/checkbox/checkbox_multiple_detail_answers.spec.js
```diff
@@ -0,0 +1,98 @@
...
+
+      // When
+        .setValue(MandatoryCheckboxPage.yourChoiceText(), 'Bacon')
+        .click(MandatoryCheckboxPage.submit())
+        .getUrl().should.eventually.contain(SummaryPage.pageName);
+    });
+  });
+
+
+  it('Given a non-mandatory detail answer, when the user does not provide any text, then "No answer provided" should be displayed on the summary screen', function() {
```
ajmaddaford: I realise this is existing behaviour, but this is odd, it shouldn't display anything for an optional detail field that is empty. Maybe now would be a good time to fix this?

tests/functional/spec/components/checkbox/checkbox_multiple_detail_answers.spec.js
```diff
@@ -0,0 +1,98 @@
...
+        .click(MandatoryCheckboxPage.yourChoice())
+        .setValue(MandatoryCheckboxPage.yourChoiceText(), 'Bacon')
+        .click(MandatoryCheckboxPage.submit())
+      // Then
+        .getText(SummaryPage.mandatoryCheckboxAnswer()).should.eventually.equal('Cheese\nMozzarella\nYour choice\nBacon');
+    });
+  });
+
+
+  it('Given I have previously added text in other textfield and saved, when I uncheck other options and select a different checkbox as answer, then the text entered in other field must be wiped.', function() {
```
ajmaddaford: This refers to 'other' textfield rather than detail answer. Also, 'must be wiped' sounds odd, something like 'Then the detail answer should be empty' would be better.

tests/functional/spec/components/checkbox/checkbox_multiple_detail_answers.spec.js
```diff
@@ -0,0 +1,98 @@
...
+      // When
+        .click(MandatoryCheckboxPage.cheese())
+        .isVisible(MandatoryCheckboxPage.cheeseText()).should.eventually.be.true
+        .click(MandatoryCheckboxPage.submit())
+      // Then
+        .getText(SummaryPage.mandatoryCheckboxAnswer()).should.eventually.equal('Cheese\nNo answer provided');
+    });
+  });
+
+  it('Given a detail answer, when the user provides text, then that text should be displayed on the summary screen', function() {
```
ajmaddaford: This tests two detail answer options, should we also test for one?

tests/functional/spec/components/checkbox/checkbox_multiple_detail_answers.spec.js
```diff
@@ -0,0 +1,98 @@
+const helpers = require('../../../helpers');
+
+const MandatoryCheckboxPage = require('../../../generated_pages/checkbox_multiple_detail_answers/mandatory-checkbox.page');
+const SummaryPage = require('../../../generated_pages/checkbox_multiple_detail_answers/summary.page');
+
+describe('Checkbox with multiple "detail_answer" options', function() {
+
+  const checkbox_schema = 'test_checkbox_multiple_detail_answers.json';
+
+  it('Given "detail_answer" options are available, when the user clicks an option the detail answer input should be visible.', function() {
```
ajmaddaford: Why does "detail_answer" have quotes and an underscore?

tests/functional/spec/components/radio/radio_multiple_detail_answers.spec.js
```diff
@@ -0,0 +1,95 @@
+const helpers = require('../../../helpers');
+
+const MandatoryRadioPage = require('../../../generated_pages/radio_multiple_detail_answers/radio-mandatory.page');
+const SummaryPage = require('../../../generated_pages/radio_multiple_detail_answers/summary.page');
+
+describe('Radio with multiple "detail_answer" options', function() {
+
+  const checkbox_schema = 'test_radio_multiple_detail_answers.json';
+
+  it('Given "detail_answer" options are available, when the user clicks an option the detail answer input should be visible.', function() {
```
ajmaddaford: Same comment as above about "detail_answer"

tests/functional/spec/components/radio/radio_multiple_detail_answers.spec.js
```diff
@@ -0,0 +1,95 @@
...
+      // When
+        .click(MandatoryRadioPage.favouriteNotListed())
+        .click(MandatoryRadioPage.submit())
+      // Then
+        .isVisible(MandatoryRadioPage.error()).should.eventually.be.true
+        .getText(MandatoryRadioPage.errorNumber(1)).should.eventually.contain('Enter your favourite to continue');
+    });
+  });
+
+  it('Given a mandatory checkbox answer, when there is an error on the page for other field and I enter valid value and submit page, then the error is cleared and I navigate to next page.s', function() {
```
ajmaddaford: - Same as above, extra 's'.
- Refers to other field rather than detail answer

ajmaddaford (changes requested):
Looks much better with the custom fields 🙂

app/forms/fields.py
dwyeradam: So there is still the original `build_choices` used by drop downs and household relationships. So happy for a name change but build_choices is already taken.

tests/functional/spec/components/checkbox/checkbox_multiple_detail_answers.spec.js
dwyeradam: Already done in `checkbox.spec.js` so I could move that into here instead.

tests/functional/spec/components/checkbox/checkbox_multiple_detail_answers.spec.js
dwyeradam: I agree as I don't like it either as it offers nothing. Will change the underlying code and remove/change this test.

app/forms/questionnaire_form.py
dwyeradam: I'll add this into schema validator and ensure this doesn't break any existing live schemas. If not we should enforce this going forward. If it does i'll check why?

dwyeradam: I've added label as mandatory for text field and any numeric field and updated schemas accordingly. Small LMS changes but rest are just test schemas.

app/forms/questionnaire_form.py
dwyeradam: Solution is to pass in a boolean when creating the field to disable validation.

app/forms/fields.py
```diff
@@ -193,54 +209,56 @@ def get_select_field(answer, label, guidance, error_messages):
...
+
+        mandatory_or_optional = get_mandatory_validator(answer, answer_errors, 'MANDATORY_NUMBER')
+
+        validate_with = mandatory_or_optional + [
+            NumberCheck(answer_errors['INVALID_NUMBER']),
+            NumberRange(minimum=min_value, minimum_exclusive=minimum_exclusive,
+                        maximum=max_value, maximum_exclusive=maximum_exclusive,
+                        messages=answer_errors, currency=answer.get('currency')),
+            DecimalPlaces(max_decimals=max_decimals, messages=answer_errors),
+        ]
```
ajmaddaford: There is a lot of logic here, I think it would be clearer and more consistent with the other methods to create a `_get_number_field_validators` method and call it like the other fields:
```python
if disable_validation is False:
    validate_with = _get_number_field_validators(answer, error_messages, answer_store)

```

app/forms/household_composition_form.py
```diff
@@ -20,7 +20,7 @@ class NameForm(Form):
             guidance = answer.get('guidance', '')
             label = answer.get('label') or get_question_title(question, answer_store, schema, metadata, group_instance)
 
-            field = get_string_field(answer, label, guidance, schema.error_messages)
+            field = get_string_field(answer, label, guidance, schema.error_messages, False)
```
ajmaddaford: We should default `disable_validation` to `False` in the `get_string_field` method so we don't need to pass it here.

app/templating/view_context.py
```diff
@@ -77,10 +76,10 @@ def build_view_context_for_question(metadata, schema, answer_store, current_loca
...
 
             if answer['type'] in ('Checkbox', 'Radio'):
-                options = answer['options']
-                context['form']['other_answer'][answer['id']] = []
-                for index in range(len(options)):
-                    context['form']['other_answer'][answer['id']].append(form.get_other_answer(answer['id'], index))
+                for option in answer['options']:
+                    if 'detail_answer' in option:
+                        context['form']['fields'][option['detail_answer']['id']] = \
+                            form.get_form_answer(option['detail_answer']['id'])
```
ajmaddaford: Is this needed or can we append the detail answer id to `answer_ids` and it will be added to the context in the loop over answer ids below?

app/submitter/convert_payload_0_0_1.py
```diff
@@ -64,14 +64,14 @@ def _get_checkbox_answer_data(answer_store, answer_schema, value):
...
+            if 'detail_answer' in option:
+                filtered = answer_store.filter(answer_ids=[option['detail_answer']['id']])
 
                 if filtered.count() > 1:
-                    raise Exception('Multiple answers found for {}'.format(option['child_answer_id']))
+                    raise Exception('Multiple answers found for {}'.format(option['detail_answer']['id']))
 
-                # if the user has selected 'other' we need to find the child value it refers to.
-                # the child value can be empty, in this case we just use the main value (e.g. other)
+                # if the user has selected 'other' we need to find the detail answer value it refers to.
```
ajmaddaford: The docstring refers to 'other'

tests/functional/generate_pages.py
```diff
@@ -227,9 +227,9 @@ def process_options(answer_id, options, page_spec, base_prefix):
 
         # Add a selector for an option which exposes another input, e.g.
         # an 'Other' option which exposes a text input for the user to fill in
-        if 'child_answer_id' in option:
+        if 'detail_answer' in option:
             option_context = {
-                'answerId': option['child_answer_id'],
+                'answerId': option['detail_answer']['id'],
                 'answerName': option_name + 'Text'
```
ajmaddaford: Text might not make sense if it's a number or percentage. Maybe something generic like `Detail` would work.

tests/functional/spec/components/checkbox/checkbox_multiple_detail_answers.spec.js
```diff
@@ -0,0 +1,110 @@
+const helpers = require('../../../helpers');
+
+const MandatoryCheckboxPage = require('../../../generated_pages/checkbox_multiple_detail_answers/mandatory-checkbox.page');
+const SummaryPage = require('../../../generated_pages/checkbox_multiple_detail_answers/summary.page');
+
+describe('Checkbox with multiple "detail_answer" options', function() {
+
+  const checkbox_schema = 'test_checkbox_multiple_detail_answers.json';
+
+  it('Given detail answer options are available, when the user clicks an option the detail answer input should be visible.', function() {
```
ajmaddaford: No 'Then' in the sentence

app/templates/partials/summary/radio.html
```diff
@@ -1,12 +1,8 @@
-{{answer.value}}
-{%- if answer.child_answer_value is not none -%}
+{{answer.value.label}}
+{%- if answer.value.detail_answer_value is not none and answer.value.detail_answer_value != '' -%}
```
MebinAbraham: Why not?
``` Jinja
{%- if answer.value.detail_answer_value -%}
```

tests/functional/spec/components/checkbox/checkbox_multiple_detail_answers.spec.js
```diff
@@ -0,0 +1,110 @@
...
+      // When
+        .click(MandatoryCheckboxPage.yourChoice())
+        .click(MandatoryCheckboxPage.submit())
+      // Then
+        .isVisible(MandatoryCheckboxPage.error()).should.eventually.be.true
+        .getText(MandatoryCheckboxPage.errorNumber(1)).should.eventually.contain('Enter your topping choice to continue');
+    });
+  });
+
+  it('Given a mandatory checkbox answer, when there is an error on the page for the detail answer field I enter valid value and submit page, then the error is cleared and I navigate to next page.', function() {
```
ajmaddaford: I think this would read more clearly if some content was moved to the given - "Given a selected  checkbox answer with an error for a mandatory detail answer, When I enter valid value and submit the page, Then ..."

app/templates/partials/widgets/multiple_choice_widget.html
```diff
@@ -23,7 +23,7 @@
       "name": option.name,
       "id": option.id,
       "checked": "checked" if option['checked'],
-      "data-qa": "has-other-option" if answer.id in form.other_answer and form.other_answer[answer.id][loop.index0] != None
+      "data-qa": "has-other-option" if option.detail_answer_id
```
MebinAbraham: Depends how picky we are being. "has-other-option" or "has-detail-answer"?

app/templates/partials/widgets/multiple_choice_widget.html
```diff
@@ -44,25 +44,25 @@
...
         {% set other_input = {
           "class": "input js-focusable " ~ exclusive_class,
           "type": "text",
-          "name": other_answer.name,
-          "id": other_answer.id,
-          "value": other_answer.data,
+          "name": detail_answer.name,
+          "id": detail_answer.id,
+          "value": detail_answer.data,
           "data-qa": "other-option"
```
MebinAbraham: "details-answer-option"?

app/templates/partials/summary/radio.html
```diff
@@ -1,12 +1,8 @@
-{{answer.value}}
-{%- if answer.child_answer_value is not none -%}
+{{answer.value.label}}
```
MebinAbraham, line 1: 👍

app/templates/partials/summary/checkbox.html
```diff
@@ -1,13 +1,9 @@
-{% macro answer_with_child(selected_answer) %}
+{% macro answer_with_detail_answer(selected_answer) %}
     {{selected_answer.label}}
-    {%- if answer.child_answer_value is not none and selected_answer.should_display_other %}
-    <ul>
+    {%- if selected_answer.detail_answer_value is not none and selected_answer.detail_answer_value != '' %}
```
MebinAbraham: Same as?
``` Jinja
 {%- if selected_answer.detail_answer_value -%}
```

MebinAbraham (changes requested):

ajmaddaford (changes requested):

app/forms/fields.py
dwyeradam: I was scared you were going to ask me to refactor the whole method. This is a nice start towards that.

app/templates/partials/widgets/multiple_choice_widget.html
dwyeradam: 👍 Not used so deleted.

app/templates/partials/widgets/multiple_choice_widget.html
dwyeradam: 👍 again not used so deleted.

tests/functional/generate_pages.py
dwyeradam: You're quite right you pain in the *!$%@. I'll update all the specs accordingly.

app/forms/questionnaire_form.py
```diff
@@ -253,33 +253,30 @@ def get_data(self, answer_id):
...
-    def get_other_answer(self, answer_id, option_index):
-        if not self.options_with_children:
-            self.options_with_children = self.schema.get_parent_options_for_block(self.block_json['id'])
-
-        if answer_id in self.options_with_children and self.options_with_children[answer_id]['index'] == option_index:
-            return getattr(self, self.options_with_children[answer_id]['child_answer_id'])
-        return None
+    def get_form_answer(self, answer_id):
+        if hasattr(self, answer_id):
+            return getattr(self, answer_id)
```
ajmaddaford: This can be removed now it's no longer used (and the associated tests).

ajmaddaford (changes requested):
Looking good, just one small thing left.

ajmaddaford (approved):
Nice! 👍

app/helpers/form_helper.py
```diff
@@ -123,10 +123,10 @@ def disable_mandatory_answers(block_json):
     return block_json
 
 
-def clear_other_text_field(data, questions_for_block):
+def clear_detail_answer_field(data, questions_for_block):
     """
     Checks the submitted answers and in the case of both checkboxes and radios,
-    removes the text entered into the other text field if the Other option is not
+    removes the text entered into the detail answer field if the Other option is not
```
MebinAbraham: References `Other`

MebinAbraham (approved):
👌

dwyeradam:
Needs a co-ordinated merge with Author.

PR #1922, opened 2018-12-04
Font sizing and input widths for small to large viewports

bameyrick:
* Updated font size classes to match renamed font size classes.
https://github.com/ONSdigital/sdc-global-design-patterns/pull/210

* Updated date, duration, currency, unit, inputs to use new CSS classes. Remove unnecessary js focus styling classes. Update date input to use number input for month field.
https://github.com/ONSdigital/sdc-global-design-patterns/pull/204

### How to review 
Font classes have just been renamed. Most of the changes are regarding the date input, and the duration, currency, and unit inputs (prefixed/suffixed inputs). Need to check that having the month names passed down to the date component doesn't break them (and ideally remove this but I didn't know where to do that), and that form data from the date input and other number type inputs is still correct.

### Checklist
* [x] Number type inputs still work.
* [x] Date input still works.
* [x] Update short hashes to pattern library version once a new version has been created.


app/templates/partials/widgets/percentage_widget.html
```diff
@@ -26,6 +26,6 @@
 
     {% include 'partials/forms/input.html' %}
 
-    <span class="input-type__type" id="{{answer.id}}-type">%</span>
+    <abbr title="%" class="input-type__type" id="{{answer.id}}-type">%</abbr>
```
bitdivision, line 29: You'll need to adjust the test in `tests/integration/components/test_render_percentage_widget.py` to use abbr instead of span to get travis working. There may be other instances where span has been used in the tests.

bitdivision (commented):
The changes look good. I've gone through a lot of the test schemas and I can't see any major visual changes.

app/templates/layouts/_base.html
```diff
@@ -1,5 +1,5 @@
 <!doctype html>
-{% set cdn_hash = "v1.8.4" %}
+{% set cdn_hash = "eae2f7d" %}
```
MebinAbraham: This is while we are waiting on the release on PL right?. We moved away from using the hash to release tags for visibility.

app/assets/styles/partials/vars/_vars.scss
```diff
@@ -1,3 +1,3 @@
 $s: "/s";
-$cdn-version: "v1.8.4";
+$cdn-version: "eae2f7d";
```
bitdivision: We probably shouldn't merge until there's a release.

bitdivision (dismissed):
I'm fairly sure that some of the functional tests will break too, since they reference the month by name.

app/templates/partials/answers/date.html
```diff
@@ -1,27 +1,30 @@
...
-    <div class="field field--select field--month">
-      <label class="label mercury" for="{{answer['id']}}-month" id="label-{{answer['id']}}-month" data-qa="label-month">{{ _("Month") }}</label>
-      {% set select = form.fields[answer['id']]['month'] %}
-
-      {% include 'partials/forms/select.html' %}
+    <div class="field field--select">
+      <label class="label u-fs-r" for="{{answer['id']}}-month" id="label-{{answer['id']}}-month" data-qa="label-month">{{ _("Month") }}</label>
+      {% set input = form.fields[answer['id']]['month'] %}
+      {% set exclusive_class = "input--w-2" %}
+      {% set type = "number" %}
```
bitdivision: This needs some cleanup in the code. I'll have a look at that now.

app/templates/partials/widgets/percentage_widget.html
bameyrick: Will do, I'll dig through to find them

app/templates/layouts/_base.html
bameyrick: Yeah once we create PL v1.8.5 I'll change this

app/templates/partials/answers/date.html
bameyrick: I've put this back to a select for now with the temporary fix for mobile 👍

app/templates/partials/answers/unit.html
```diff
@@ -22,6 +22,6 @@
         'id': answer.id
         } %}
         {% include 'partials/forms/input.html' %}
-        <span class="input-type__type" id="{{answer.id}}-type">{{ format_unit_input_label(answer.unit, unit_length=answer.unit_length) }}</span>
+        <abbr title="{{ _(label.text) }}" class="input-type__type" id="{{answer.id}}-type">{{ format_unit_input_label(answer.unit, unit_length=answer.unit_length) }}</abbr>
```
ajmaddaford: It would be clearer if `label.text` was `answer.label`. The label variable is constructed to pass to the label template.

app/templates/partials/answers/date.html
```diff
@@ -1,27 +1,29 @@
...
   <div class="fieldgroup__fields">
     {% if form.fields[answer['id']]['day'] %}
-    <div class="field field--input field--day">
-      <label class="label mercury" for="{{answer['id']}}-day" id="label-{{answer['id']}}-day" data-qa="label-day">{{ _("Day") }}</label>
+    <div class="field field--input">
+      <label class="label u-fs-r" for="{{answer['id']}}-day" id="label-{{answer['id']}}-day" data-qa="label-day">{{ _("Day") }}</label>
       {% set input = form.fields[answer['id']]['day'] %}
-      {% set input_placeholder = "DD" %}
+      {% set width_class = "input--w-2" %}
+      {% set type = "number" %}
```
ajmaddaford: This should be changed to create an input dict as we do with other input types. This would tidy up the changes to `input.html` (`width_class` and `input.type` are not needed). Something like:
```python
{% set input = {
    'type': 'number',
    'classes': 'input--w-2',
    'data': input._value(),
    'name': answer.id,
    'id': answer.id,
    'placeholder': 'DD'
} %}
```

Also, are the placeholders supposed to be removed?

app/templates/partials/answers/date.html
```diff
@@ -1,27 +1,29 @@
...
     {% endif %}
     {% if form.fields[answer['id']]['month'] %}
-    <div class="field field--select field--month">
-      <label class="label mercury" for="{{answer['id']}}-month" id="label-{{answer['id']}}-month" data-qa="label-month">{{ _("Month") }}</label>
+    <div class="field field--select">
+      <label class="label u-fs-r" for="{{answer['id']}}-month" id="label-{{answer['id']}}-month" data-qa="label-month">{{ _("Month") }}</label>
       {% set select = form.fields[answer['id']]['month'] %}
-
+      {% set width_class = "input--w-auto" %}
       {% include 'partials/forms/select.html' %}
```
ajmaddaford, line 32: Similar as comment for above, other uses of `select.html` create a select dict.

app/templates/partials/widgets/date_widget.html
```diff
@@ -1,39 +1,44 @@
 {% set subfields %}
```
ajmaddaford: I don't think this template is used, I think we should remove it in this pr.

ajmaddaford (changes requested):

app/templates/partials/forms/input.html
```diff
@@ -1,14 +1,14 @@
 {% set attr = {
-  "class": "input input--" ~ input.type ~ " " ~ input.classes ~ " " ~ exclusive_class,
-  "type": type,
+  "class": "input input--" ~ input.type ~ " " ~ input.classes ~ " " ~ exclusive_class ~ " " ~ width_class,
+  "type": type if type is not none else input.type,
```
MebinAbraham: Probably want: `"type": type or input.type`
I don't get why we have 2 variables that could hold the 'type', we shoul donly have one?.

Line 2 references `input.type` but this can sometimes be invalid?

MebinAbraham: Probably want: `"type": type or input.type`
I don't get why we have 2 variables that could hold the 'type', we should only have one?.

Line 2 references `input.type` but this can sometimes be invalid?

MebinAbraham (changes requested):

app/templates/partials/answers/date.html
bameyrick: Placeholders should be avoided as they are bad for accessibility so I've removed them to match the pattern library.

How would I get the correct name/id/value for each field in for the date answer?

app/templates/partials/forms/input.html
bameyrick: I've seen a few places where type was used, but I'll have a look through the templates and change these instances to use an input dict like Andrew mentioned and remove just type all together

bitdivision:
Is this ready for rereview now @bameyrick 	?

ajmaddaford (approved):

app/templates/partials/answers/checkbox.html
```diff
@@ -5,14 +5,14 @@
...
     {% else %}
-      <div class="field__label mars">{{ _("Select all that apply") }}:</div>
+      <div class="field__label u-fs-r">{{ _("Select all that apply") }}:</div>
     {% endif %}
 
   {% endif %}
 {% endset %}
 
-<div class="field field--checkbox field--multiplechoice {{'field--exclusive' if is_mutually_exclusive }} {{'field--cols' if use_grid}}">
+<div class="field field--checkbox field--multiplechoice {{'field--cols' if use_grid}}">
```
MebinAbraham, line 16: 👍

app/templates/partials/answers/date.html
```diff
@@ -1,27 +1,48 @@
...
     </div>
     {% endif %}
     {% if form.fields[answer['id']]['month'] %}
-    <div class="field field--select field--month">
-      <label class="label mercury" for="{{answer['id']}}-month" id="label-{{answer['id']}}-month" data-qa="label-month">{{ _("Month") }}</label>
-      {% set select = form.fields[answer['id']]['month'] %}
-
+    <div class="field field--select">
+      <label class="label u-fs-r" for="{{answer['id']}}-month" id="label-{{answer['id']}}-month" data-qa="label-month">{{ _("Month") }}</label>
+      {% set field = form.fields[answer['id']]['month'] %}
```
MebinAbraham: Existing but while we're here.
I think it would be better to set the field before the `if` then reuse the variable?
So:
``` Jinja
{% set month_field = form.fields[answer['id']]['month'] %}
{% if month_field %}
	{% set select = {
		'id': month_field.id,
		'name': month_field.id,
		'classes': 'input--w-auto',
		'choices': month_field.choices,
		'data': month_field.data,
		'flags': {
			'required': answer.mandatory
		}
	} %}

	<div class="field field--select">
		<label class="label u-fs-r" for="{{answer['id']}}-month" id="label-{{answer['id']}}-month" data-qa="label-month">{{ _("Month") }}</label>
		{% include 'partials/forms/select.html' %}
	</div>
```
Same concept for day and year also.

app/templates/partials/forms/select.html
```diff
@@ -1,4 +1,4 @@
-<select class='input input--select {{ exclusive_class }}' id='{{select.name}}' name='{{select.name}}'>
+<select class='input input--select {{ select.classes }} {{ exclusive_class }}' id='{{select.name}}' name='{{select.name}}'>
```
MebinAbraham, line 1: Can't tell which is better but just putting it out there, feel free to ignore.
``` Jinja
<select class='input input--select {{ select.classes ~' '~ exclusive_class }}' id='{{select.name}}' name='{{select.name}}'>
```

MebinAbraham:
It will also be worth updating the translation `messages.pot` file.

https://github.com/ONSdigital/eq-survey-runner#internationalisation

app/templates/partials/answers/date.html
bameyrick: Done 👍

app/templates/partials/forms/select.html
bameyrick: What do the tildes do? I can't find anything about it in the nunjucks docs

ajmaddaford: I think the original reads better

ajmaddaford: @bameyrick The tilde is documented at http://jinja.pocoo.org/docs/2.10/templates/#other-operators

bameyrick: Ah thank you, I thought it might have done something with whitespace management, but as it only concatenates strings I agree the original is more legible

MebinAbraham (approved):

MebinAbraham:
Can be squashed @bameyrick

PR #1923, opened 2018-12-05
Remove unnecessary padding from question title in summaries

bameyrick:
Removed unnecessary padding on summaries to fix visual defect (see total row). 

lms2-playback-total-incorrect-height

### DIFF
--- a/app/templates/partials/summary/display-question.html
+++ b/app/templates/partials/summary/display-question.html
@@ -1,1 +1,1 @@
-<div class="summary__question u-pb-s" id="{{ question.id }}" data-qa="summary-question-title">{{ question.number or ''}} {{question.title|striptags }}</div>
+<div class="summary__question" id="{{ question.id }}" data-qa="summary-question-title">{{ question.number or ''}} {{question.title|striptags }}</div>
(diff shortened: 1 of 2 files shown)


app/templates/partials/summary/display-question.html
```diff
@@ -1 +1 @@
-<div class="summary__question u-pb-s" id="{{ question.id }}" data-qa="summary-question-title">{{ question.number or ''}} {{question.title|striptags }}</div>
+<div class="summary__question" id="{{ question.id }}" data-qa="summary-question-title">{{ question.number or ''}} {{question.title|striptags }}</div>
```
MebinAbraham, line 1: This template is used for questions in general. So is it intended that there should be no `padding-bottom` in all cases or just the calculated summary grand total title?

Expected?
Before:
screenshot 2018-12-05 at 11 09 04
After:
screenshot 2018-12-05 at 11 09 16

bameyrick: It should be in all cases, I think the total was the first place it was noticed.

ajmaddaford (approved):

bitdivision (approved):

PR #1924, opened 2018-12-06
Remove unused schemas

ajmaddaford:
Removal of schemas that are no longer used.

### How to review 
Verify the schemas don't appear in the launcher drop down.


MebinAbraham (approved):

dwyeradam (approved):
The card doesn't include these but we should also look at getting rid of test_0102, test_0112, test_0203, test_0205 and test_0213 which are only there for integration tests. There should be an equivalent test schemas with the tested features already in runner somewhere. If not they should be made.

dwyeradam:
`test_0213.json` can be deleted already as it's not used.

ajmaddaford:
@dwyeradam I've removed `test_0213.json`

PR #1921, opened 2018-12-03
Always save the latest available version of QuestionnaireStore.

bitdivision:
### What is the context of this PR?
Currently, we are able to read version 3 in production, but can only write version 2.

While making the changes to write version 3, it became clear that there is an issue around what version we are writing for the questionnaire store.

In a deployment as follows (I've represented the version of the data with `data_version`):

1) Instance running `version 2` saves QuestionnaireStore. `{version=2, data_version=2}`
2) Instance running `version 3` loads QuestionnaireStore with `version 2`, upgrades to `version 3`, and saves `{version=3, data_version=3}`
3) Instance running `version 2` loads QuestionnaireStore with `version 3`, and saves with `{version=3, data_version=2}`

At this point we have a mismatched version number, which will cause errors when an application tries to load it.

To avoid this, we can ensure that we always write the latest version which the current code can write, rather than the highest possible version.

### How to review 
I have a branch which adds the writing functionality of version 3. This can be used to test it.

1) Checkout master, start a questionnaire
2) Checkout [`2647-update-questionnaire-store-version-for-compression-and-new-answerstore-storage-format`](https://github.com/ONSdigital/eq-survey-runner/tree/2647-update-questionnaire-store-version-for-compression-and-new-answerstore-storage-format), restart server, continue questionnaire.
3) Checkout master, restart server, continue questionnaire. 
4) You should see errors
5) Checkout this PR's branch, start a questionnaire
6) Checkout [`2647-update-questionnaire-store-version-for-compression-and-new-answerstore-storage-format`](https://github.com/ONSdigital/eq-survey-runner/tree/2647-update-questionnaire-store-version-for-compression-and-new-answerstore-storage-format), merge this PR into it, continue your questionnaire.
7) Checkout this PR, continue your questionnaire
8) Checkout [`2647-update-questionnaire-store-version-for-compression-and-new-answerstore-storage-format`](https://github.com/ONSdigital/eq-survey-runner/tree/2647-update-questionnaire-store-version-for-compression-and-new-answerstore-storage-format), continue questionnaire, and ensure it's all working

### Checklist

* [ ] New static content marked up for translation
* [ ] Newly defined schema content included in eq-translations repo

PR #1925, opened 2018-12-06
Save state in new format

bitdivision:
DO NOT MERGE until https://github.com/ONSdigital/eq-survey-runner/pull/1921 is in production!

Now that https://github.com/ONSdigital/eq-survey-runner/pull/1905 has gone into production, we need to enable writing of new formats.

This PR essentially reverts the changes from https://github.com/ONSdigital/eq-survey-runner/pull/1905. It adds this as version 3 of the questionnaire store.

- Data is written compressed / without base64 and without a JSON wrapper.

### How to review 

Test that nothing goes wrong when:
- starting a questionnaire on version 2 (with #1905 and #1921 )
- upgrade to version 3
- continue the questionnaire
- downgrade to version 2
- continue the questionnaire

At each stage of the above, we should probably test submission.

### Checklist

### DIFF
--- a/tests/integration/session/test_questionnaire_store_upgrade.py
+++ b/tests/integration/session/test_questionnaire_store_upgrade.py
@@ -28,15 +28,15 @@ class TestLogin(IntegrationTestCase):
 
     def test_questionnaire_store_is_upgraded(self):
         # Given
-        previous_version = QuestionnaireStore.LATEST_VERSION - 1
 
-        # Creates a QuestionnaireStore with previous version
-        with patch('app.data_model.questionnaire_store.QuestionnaireStore.get_latest_version_number', return_value=previous_version):
-            self.launchSurvey('test', '0205')
+        self.launchSurvey('test', 'numbers')
 
-        # On a subsequent request the `upgrade` method of answer_store should be called
-        with patch('app.data_model.questionnaire_store.AnswerStore.upgrade') as upgrade:
-            self.post(action='start_questionnaire')
+        # Increment the LATEST_VERSION so the `upgrade` method of answer_store is called when fetching
+        # the questionnaire store.
+        next_version = QuestionnaireStore.LATEST_VERSION + 1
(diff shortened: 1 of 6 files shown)


ajmaddaford (approved):

app/views/questionnaire.py
```diff
@@ -436,7 +436,7 @@ def submit_answers(routing_path, eq_id, form_type, schema):
     _store_submitted_time_in_session(submitted_time)
 
     if is_view_submitted_response_enabled(schema.json):
-        _store_viewable_submission(list(answer_store), metadata, submitted_time)
+        _store_viewable_submission(answer_store.answer_map, metadata, submitted_time)
```
ajmaddaford: This should probably be in the another pull request as it's not related to saving the new compressed questionnaire state.

ajmaddaford (changes requested):

ajmaddaford (approved):

MebinAbraham (approved):

tests/integration/session/test_questionnaire_store_upgrade.py
```diff
@@ -28,15 +28,15 @@ class TestLogin(IntegrationTestCase):
 
     def test_questionnaire_store_is_upgraded(self):
         # Given
-        previous_version = QuestionnaireStore.LATEST_VERSION - 1
 
-        # Creates a QuestionnaireStore with previous version
-        with patch('app.data_model.questionnaire_store.QuestionnaireStore.get_latest_version_number', return_value=previous_version):
-            self.launchSurvey('test', '0205')
+        self.launchSurvey('test', '0205')
```
dwyeradam: Does this need to use test_0205.json? Ideally we should be getting rid of these and using actual test schemas. Not a blocker as it's not related to this PR.

bitdivision: Fixed now.

PR #1926, opened 2018-12-06
Date ranges should not be wrapped in a fieldset as each date input is a fieldset in its own right

bameyrick:
Updated the question template for date range to not extend "generic" as each date in the range is an answer, and the generic would wrap them in another fieldset.

### How to review
Check this is a suitable fix for the issue.


app/templates/partials/questions/daterange.html
```diff
@@ -1 +1,42 @@
...
+    {{question_definition}}
+    {{question_guidance}}
+    {% include theme('partials/questions/errors.html') %}
+
+  {% else %}
+
+    {{question_markup}}
+    {{question_actions}}
+
+  {% endif %}
```
ajmaddaford, line 29: I realise this is something that is the same in `general.html` but this could be simplified to:
```
{{question_titles}}
{{question_description}}
{{question_definition}}
{{question_guidance}}

{% if invalid %}
  {% include theme('partials/questions/errors.html') %}
{% else %}
  {{question_answers}} 
  {{question_actions}}
{% endif %}
```

app/templates/partials/questions/daterange.html
```diff
@@ -1 +1,42 @@
-{% include theme('partials/questions/general.html') %}
+{% if not question_answers %}
+  {% set question_answers %}
+    <div class="question__answers">
+      {%- for answer in question.answers -%}
+        {% if not answer.parent_answer_id %}
```
ajmaddaford: `parent_answer_id` was removed from `general.html` in https://github.com/ONSdigital/eq-survey-runner/pull/1920. See https://github.com/ONSdigital/eq-survey-runner/blob/master/app/templates/partials/questions/general.html for the new version.

ajmaddaford (changes requested):

app/templates/partials/questions/daterange.html
bameyrick: I've updated this to match general.html from #1920 now

app/templates/partials/questions/daterange.html
bameyrick: Done 👍

bitdivision:
This is based on top of #1922 I think. Should we have a blocked label on it?

bameyrick:
Yeah I think I screwed up rebasing from master and rebased from another branch

bameyrick:
Did rebase incorrectly and cannot recover. New PR created.

PR #1933, opened 2018-12-07
Do not store unanswered questions in the answer store

bitdivision:
Merge only after #1925 has gone in. (I have merged that branch into this one).

[Trello](https://trello.com/c/O7ZDvcSl/2437-do-not-store-un-answered-questions-in-the-answer-store-m).

Currently, the data we store in the answer store for unanswered questions is quite varied (see Andrew's investigation in trello). This PR normalises that by ensuring we never store anything in the answer store for unanswered questions.

Since version 3 of the questionnaire store didn't involve an answer store upgrade, I had to modify the upgrade method to allow for skipped versions. The difficulties around this system of versioning / upgrades have been noted in a [trello](https://trello.com/c/NB7p7r33/2660-discuss-options-around-versioning-of-state) card, we likely need to refactor this. I have not gone very far in this PR.

### How to review 
The `/dump/answers` endpoint should prove useful. Run through some live surveys.

Ensure unanswered questions do not appear in the store. Test each of the question types.
### Checklist


app/data_model/answer_store.py
```diff
@@ -245,7 +249,24 @@ def upgrade_1_to_2_add_group_instance_id(answer_store, schema):
...
+            answer_store_copy.remove_answer(answer)
+
+    answer_store = answer_store_copy
+
+
+# Dictionary specifying upgrade methods. Key should be the version to upgrade from. It is assumed that each transform only upgrades the version by 1.
+UPGRADE_TRANSFORMS = {
+    0: upgrade_0_to_1_update_date_formats,
+    1: upgrade_1_to_2_add_group_instance_id,
+    3: upgrade_3_to_4_remove_empty_answers,
```
ajmaddaford: I think it would be clearer if the key was the version the upgrade is for, so would be 1,2,4 rather than 0,1,3. 

We should also drop the from version numbers from the method names as they don't make sense now version numbers can be skipped. For example `upgrade_3_to_4_remove_empty_answers` also upgrades version 2 answer stores.

ajmaddaford (changes requested):

app/data_model/answer_store.py
bitdivision: Fixed.

ajmaddaford (changes requested):
Empty answers are still being stored for household composition and household relationships.

tests/app/data_model/test_answer_store.py
```diff
@@ -559,6 +565,54 @@ def test_upgrade_multiple_versions_skipping_already_run(self):  # pylint: disabl
...
+    def test_upgrade_to_4(self):
+        """ The upgrade to version 4 should ensure that we don't store unanswered answers inc. empty answers
+        """
+        invalid_answer_values = ('', [])
+        valid_answer_values = ('None', 'VALID', 0)
+        answer_values = invalid_answer_values + valid_answer_values
+
+        answer_store = AnswerStore()
+        for i, answer_value in enumerate(answer_values):
+            print(i, answer_value)
```
MebinAbraham: Debugging...?

MebinAbraham (changes requested):

app/data_model/answer_store.py
```diff
@@ -245,7 +248,24 @@ def upgrade_1_to_2_add_group_instance_id(answer_store, schema):
...
+def upgrade_to_4_remove_empty_answers(answer_store, schema):  # pylint: disable=unused-argument
+    """ Previously, we stored [] and '' as answer values for unanswered, but completed answers.
+
+    This was changed to avoid storing unanswered values entirely, so this upgrade will
+    remove any answers in the store that have 'empty' answer values
+    """
+    answer_store_copy = answer_store.copy()
+
+    for answer in answer_store:
+        if answer['value'] in ('', [], None):
```
MebinAbraham: Why does doc string say `we stored [] and '' as answer values for unanswered,` but this has `None` as well?

app/jinja_filters.py
```diff
@@ -423,12 +424,15 @@ def format_repeating_summary(context, items, delimiter=' '):
...
+
+    If any of the lists are shorter than others, lists will be padded with `None`
     """
     if items:
         intermediates = []
         for item in items:
             if item and isinstance(item[0], list):
-                intermediates.extend(list(map(list, zip(*item))))
+
+                intermediates.extend(list(map(list, zip_longest(*item))))
```
MebinAbraham: Super important. Why the extra line? -___-

MebinAbraham (commented):
Need to ensure SDX still works as expected with this change.
Will have a dig tomorrow to see if there are any assumptions made in there.

MebinAbraham:
Haven't tested everything in SDX but it seems to work as expected.

MebinAbraham (commented):
Apart from the census related stuff Andrew raised, other types seem to be working as expected. 👍

bitdivision:
Adding household summary / relationships has raised some issues:

- Household summary relies on empty strings being used as answer values.
	- This only applies to optional fields, e.g. if middle name is optional and one instance includes a middle name but the other doesn't, then we don't know which to associate with which.
- The same type of issue comes up with `format_repeating_summary` which also zips lists together to create the names. While this is only used with LMS, which has mandatory fields, it currently could be used on future schemas.

Since fixing the above to work with optional answers requires a significant refactor and may be fixed when work is done on repeating groups, I'm closing this PR for the moment. It is likely worth revisiting once 2017 census test functionality has been removed and repeating groups / jinja filters have been rethought.

PR #1937, opened 2018-12-11
Add answer store upgrade path refactor

bitdivision:
The PR to handle empty answer store values was closed, but I wanted to pull this out since it's a valuable change to answer store upgrades.

This change means that upgrades no longer rely on the length of the upgrades list and allow for skipped versions.

I've also added a fix for the routing test which was always broken, but worked intermittently AFAICT.

### Checklist


ajmaddaford (approved):
Tested going from version 1 to 2 with an in flight survey - all good :)

dwyeradam (approved):
Tested upgrades and seemed to work.

PR #1945, opened 2018-12-17
Introduce AnswerStoreUpdater

MebinAbraham:

Since David's initial additions, I have also moved action relating to the household/relationship from the `questionnaire` view into the `AnswerStoreUpdater` also. I've kept the original commits so you can compare.

### How to review 
- All tests should pass
- Code coverage should not reduce
- No changes to functionality


tests/app/questionnaire/test_answer_store_updater.py
```diff
@@ -0,0 +1,249 @@
+# import unittest
```
MebinAbraham: I was debating whether I should keep this as well?. When modifying these to work with the current version of master, I seemed to end up mocking a lot on top of what's already mocked and seemed almost pointless.

The scenarios are covered via integration tests for the questionnaire.py

nealedj: My 2p...yeah I know what you mean. If high-level integration tests are present it's tempting to just add lots of them because they all end up exercising their dependencies anyway.

I guess a few arguments in favour of keeping these unit tests:

- they will run more quickly than integration tests
- they continue a useful convention where most python modules in survey runner have a corresponding set of unit tests
- it should be quicker/easier to work on the answer store updater module with a dedicated set of tests to verify changes. I.e. new scenarios

app/questionnaire/answer_store_updater.py
```diff
@@ -0,0 +1,163 @@
...
+        self._questionnaire_store = questionnaire_store
+        self._answer_store = self._questionnaire_store.answer_store
+
+    def save_form(self, form):
+        if self._should_save_serialised_answers():
+            self._update_questionnaire_store_with_answer_data(form.serialise())
+        else:
+            self._update_questionnaire_store_with_form_data(form.data)
+
+        self._questionnaire_store.add_or_update()
```
ajmaddaford, line 28: We could use the type of form here to determine which method to call here, something like:
```python
if isinstance(form, QuestionnaireForm):
else:
    self._update_questionnaire_store_with_answer_data(form.serialise())
```

app/questionnaire/answer_store_updater.py
```diff
@@ -0,0 +1,163 @@
...
+            if answer.answer_id in survey_answer_ids
+        )
+
+        for answer in valid_answers:
+            answer.group_instance_id = get_group_instance_id(self._schema, self._answer_store, self._current_location,
+                                                             answer.answer_instance)
+            self._answer_store.add_or_update(answer)
+
+        if self._current_location not in self._questionnaire_store.completed_blocks:
+            self._questionnaire_store.completed_blocks.append(self._current_location)
```
ajmaddaford: These two lines are duplicated in both `_update_questionnaire_store_with_` methods and could be moved to the `save_form` method.

app/questionnaire/answer_store_updater.py
```diff
@@ -0,0 +1,163 @@
...
+    """Component responsible for any actions that need to happen as a result of updating the answer store
+    """
+
+    def __init__(self, current_location, schema, questionnaire_store):
+        self._current_location = current_location
+        self._schema = schema
+        self._questionnaire_store = questionnaire_store
+        self._answer_store = self._questionnaire_store.answer_store
+
+    def save_form(self, form):
```
ajmaddaford: I think this method would be clearer if it was called `save_answers`

ajmaddaford (changes requested):

app/views/questionnaire.py
```diff
@@ -200,15 +188,17 @@ def post_block(routing_path, schema, metadata, collection_metadata, answer_store
...
+    current_location = Location(group_id, 0, 'household-composition')
 
-    if _household_answers_changed(answer_store, schema):
-        _remove_repeating_on_household_answers(answer_store, schema)
+    answer_store_updater = AnswerStoreUpdater(current_location, schema, get_questionnaire_store(current_user.user_id, current_user.user_ik))
+    household_answer_changed = answer_store_updater.household_answers_changed(request.form.copy())
 
-    disable_mandatory = any(x in request.form for x in ['action[add_answer]', 'action[remove_answer]', 'action[save_sign_out]'])
+    if household_answer_changed:
+        answer_store_updater.remove_repeating_on_household_answers()
```
dwyeradam: Can the logic around household answers changing (3 lines) all be dealt with with answer_store_updater? At the moment all the work is done there but the logic to call it sits outside.

dwyeradam (changes requested):

MebinAbraham:
@ajmaddaford @dwyeradam Ready for re-review.

tests/app/questionnaire/test_answer_store_updater.py
```diff
@@ -0,0 +1,266 @@
...
+
+        assert self.questionnaire_store.completed_blocks == [self.location]
+
+        assert len(answers) == self.answer_store.add_or_update.call_count
+
+        # answers should be passed straight through as Answer objects
+        answer_calls = map(mock.call, answers)
+        self.answer_store.add_or_update.assert_has_calls(answer_calls, any_order=True)
+
+    def test_save_form_with_form_data(self):
```
ajmaddaford: All `test_save_form_*` methods should be renamed `test_save_answers_*` to match the code change.

tests/app/questionnaire/test_answer_store_updater.py
```diff
@@ -0,0 +1,266 @@
...
+        answer_id = 'answer'
+        answer_value = '1000'
+        self.location.group_instance = 1
+        self.schema.get_answer_ids_for_block.return_value = [answer_id]
+        self.schema.block_has_question_type.return_value = False
+        self.schema.get_group_dependencies.return_value = None
+
+        form = MagicMock(spec=QuestionnaireForm, data={answer_id: answer_value})
+
+        # @patch('')
```
ajmaddaford: This should be removed.

tests/integration/views/test_questionnaire.py
```diff
@@ -34,17 +36,16 @@ def tearDown(self):
...
-        form_data = {
+        form_data = MagicMock(spec=QuestionnaireForm, data={
             'total-retail-turnover-answer': '1000',
-        }
+        })
 
         with self._application.test_request_context():
-            update_questionnaire_store_with_form_data(self.question_store, location, form_data, schema)
+            answer_store_updater = AnswerStoreUpdater(location, schema, self.question_store)
+            answer_store_updater.save_answers(form_data)
```
ajmaddaford: Is this test relevant now that we have `AnswerStoreUpdater` and it's tests. This looks like it is duplicating tests.

tests/integration/views/test_questionnaire.py
```diff
@@ -57,18 +58,17 @@ def test_update_questionnaire_store_with_form_data(self):
...
-        form_data = {
+        form_data = MagicMock(spec=QuestionnaireForm, data={
             'answer': None
-        }
+        })
 
         with self._application.test_request_context():
-            update_questionnaire_store_with_form_data(self.question_store, location, form_data, schema)
+            answer_store_updater = AnswerStoreUpdater(location, schema, self.question_store)
+            answer_store_updater.save_answers(form_data)
```
ajmaddaford: Same comment as above about duplication of tests.

tests/integration/views/test_questionnaire.py
```diff
@@ -81,54 +81,64 @@ def test_update_questionnaire_store_with_default_value(self):
...
             )
         ]
 
+        form = MagicMock()
+        form.serialise.return_value = answers
+
         with self._application.test_request_context():
-            update_questionnaire_store_with_answer_data(self.question_store, location, answers, schema)
+            answer_store_updater = AnswerStoreUpdater(location, schema, self.question_store)
+            answer_store_updater.save_answers(form)
```
ajmaddaford: Same comment as above about duplication of tests. Looking at the rest of these tests the same applies. If the functionality is performed by `AnswerStoreUpdater`, that's where the tests are/should be.

ajmaddaford (changes requested):

MebinAbraham:
Just converting some of the existing integration tests relating to `AnswerStoreUpdater` from `questionnaire.py` to unit tests.

MebinAbraham:
@ajmaddaford @dwyeradam Have removed tests from `questionnaire.py` which now fits into `AnswerStoreUpdater`.

tests/app/questionnaire/test_answer_store_updater.py
```diff
@@ -0,0 +1,506 @@
...
+        answers = []
+        answers.extend(answered)
+        answers.extend(unanswered)
+
+        form = MagicMock()
+        form.serialise.return_value = answers
+
+        self.answer_store_updater.save_answers(form)
+
+        self.answer_store.filter.return_value = iter(i[0][0].__dict__ for i in self.answer_store.add_or_update.call_args_list)
```
bitdivision: 🙀

MebinAbraham: 👀

ajmaddaford: We shouldn't need to do it this way, we  just need a mock for the filter method to return the answers i.e. we can remove the `save_answers` and `add_or_update stuff`.

MebinAbraham: Have updated the unit test to match this.

tests/app/questionnaire/test_answer_store_updater.py
```diff
@@ -161,54 +161,53 @@ def test_save_answers_data_with_default_value(self):
...
-                answer_id='middle-names',
-                answer_instance=1,
-                value=''
-            ), Answer(
-                group_instance=0,
-                answer_id='last-name',
-                answer_instance=1,
-                value=''
-            )
+        empty_household_answer = [
```
ajmaddaford: minor, this should be _answer**s**

ajmaddaford (changes requested):

dwyeradam (approved):

ajmaddaford (approved):

PR #1946, opened 2018-12-17
Document Jinja filter schema usage

dwyeradam:
Document Jinja filters used in schemas
Also:
Removes unused `format_currency_for_input` filter
Renames `format_household_member_name` methods to match names used to call them
Adds a unit test for `test_format_household_summary` which was missing

### How to review 
Read the new `jinja-filters.md` file and help me make it less boring/more understandable.
Also check the couple of minor out of scope changes haven't had a negative effect.

### Checklist

### DIFF
--- a/app/jinja_filters.py
+++ b/app/jinja_filters.py
@@ -43,34 +43,25 @@ def get_currency_symbol(currency='GBP'):
 
 
 @blueprint.app_template_filter()
-def format_currency_for_input(value, decimal_places=0):
-    if value is None or value == '':
-        return ''
-    if decimal_places is None or decimal_places == 0:
-        return format_number(value)
-    return get_formatted_currency(value).replace(get_currency_symbol(), '')
-
-
-@blueprint.app_template_filter()
 def format_percentage(value):
     return '{}%'.format(value)
 
 
 @blueprint.app_template_filter()
-def format_address_list(user_entered_address=None, metadata_address=None):
+def format_address_list(first_address=None, second_address=None):
     """
-        This function format_address_list accepts two addresses, a user submitted address and a metadata address
-        If all the items in address list are 'Undefined' (nothing in answer_store) then we retrieve the metadata.
+        The function format_address_list accepts two lists of address values.
+        If all the items in the first address list are empty or 'Undefined' then we use the second address.
 
-        :param user_entered_address user entered address which is a list of answer fields
-        :param metadata_address metadata address which is a list of answer fields passed to us
-        :return: the value of the address to be piped
+        :param (list) first_address
+        :param (list) second_address
+        :return: first address if values present else second address
     """
-    if all(isinstance(field, Undefined) for field in user_entered_address) or \
-       all(field == '' for field in user_entered_address):
-        address = metadata_address
+    if all(isinstance(field, Undefined) for field in first_address) or \
+       all(field == '' for field in first_address):
+        address = second_address
     else:
-        address = user_entered_address
+        address = first_address
 
     address_list = concatenated_list(list_items=address, delimiter='<br />')
 
@@ -339,13 +330,13 @@ def first_non_empty_item(context, *items):
 
 
 @blueprint.app_template_filter()
-def format_household_member_name(names):
+def format_household_name(names):
     return concatenated_list(list_items=names, delimiter=' ')
 
 
 @blueprint.app_template_filter()
-def format_household_member_name_possessive(names):
-    name = format_household_member_name(names)
+def format_household_name_possessive(names):
+    name = format_household_name(names)
     if name:
         last_char = name[-1:]
         if last_char.lower() == 's':
@@ -404,7 +395,7 @@ def format_household_summary(context, names):
     if names:
         person_list = []
         for first_name, middle_name, last_name in zip(names[0], names[1], names[2]):
-            person_list.append(format_household_member_name([first_name, middle_name, last_name]))
+            person_list.append(format_household_name([first_name, middle_name, last_name]))
 
         return format_unordered_list(context, [person_list])
     return ''
@@ -547,11 +538,6 @@ def format_currency_processor():
 @blueprint.app_context_processor
 def get_currency_symbol_processor():
     return dict(get_currency_symbol=get_currency_symbol)
-
-
-@blueprint.app_context_processor
-def format_currency_for_input_processor():
-    return dict(format_currency_for_input=format_currency_for_input)
 
 
 @blueprint.app_context_processor
--- a/app/templating/template_renderer.py
+++ b/app/templating/template_renderer.py
@@ -15,8 +15,8 @@ class TemplateRenderer:
         env.filters['concatenated_list'] = filters.concatenated_list
         env.filters['format_date'] = filters.format_date
         env.filters['format_date_custom'] = filters.format_date_custom
-        env.filters['format_household_name'] = filters.format_household_member_name
-        env.filters['format_household_name_possessive'] = filters.format_household_member_name_possessive
+        env.filters['format_household_name'] = filters.format_household_name
+        env.filters['format_household_name_possessive'] = filters.format_household_name_possessive
         env.filters['format_household_summary'] = filters.format_household_summary
         env.filters['format_number'] = filters.format_number
         env.filters['format_repeating_summary'] = filters.format_repeating_summary
--- a/doc/jinja-filters.md
+++ b/doc/jinja-filters.md
@@ -1,0 +1,446 @@
+# Jinja Filters Used by Schemas
+
+
+## format_number
+Returns the given number with the appropriate thousands grouping and decimal separator as determined by the locale.
+Uses Babel's [format_decimal](http://babel.pocoo.org/en/latest/api/numbers.html#babel.numbers.format_decimal) from `number.py`.
+
+##### Parameters: 
+- value: A single numeric value.
+
+##### Context:
+Most commonly used to display a previous answer's numeric value in a following question.
+```
+Of the <em>{{answers['total-number-employees']|format_number}}</em> total employees employed on 14 December 2018, how many male and female employees worked the following hours?
+```
+
+##### Examples:
+```
+Input:          Output:
+"100000000"     "100,000,000"
+"123.40"        "123.4"
+```
+
+## format_currency
+Returns a formatted currency value. 
+Uses Babel's [format_currency](http://babel.pocoo.org/en/latest/api/numbers.html#babel.numbers.format_currency) from `number.py`.
+
+##### Parameters: 
+- value:    A single numeric value.
+- currency: An optional currency in format 'GBP' (default), 'EUR' etc.
+
+##### Context:
+Most commonly used to display a previous answer's currency value in a following question.
+```
+Of the <em>{{format_currency(answers['total-retail-turnover-answer'])}}</em> total retail turnover, what was the value of internet sales?
+```
+
+##### Examples:
+```
+Input:              Output:
+("11.99", "GBP")    "<span class='date'>£11.99</span>"
+("11000", "USD")    "<span class='date'>US$11,000.00</span>"
+```
+
+## format_address_list
+This accepts two lists of address values.
+If all the items in the first address list are empty or 'undefined' then we use the second address.
+
+##### Parameters: 
+- first_address:    List of address values
+- second_address:   List of address values
+
+##### Context:
+Used to display a full address for use in a description. 
+Used where there is potential that a respondent may provide a new address. 
+Therefore it will use that respondent entered address if present otherwise the initial address passed as metadata.
+```
+{{ format_address_list ([answers['address-line1'], answers['address-line2'], answers['locality'], answers['town-name'], answers['postcode']], [metadata['address_line1'], metadata['address_line2'], metadata['locality'], metadata['town_name'], metadata['postcode']]) }}
+```
+
+##### Examples:
+```
+first_address = ["44", "High Stret", "", "Big City", ""]
+second_address = ["123", "Testy", "Place", "Newport", "NP5 7AR"]
+
+Input:                              Output:
+(first_address, second_address)     "44<br />High Street<br />Big City"
+
+first_address = [Undefined(), Undefined(), Undefined()]
+second_address = ["123", "Testy", "Place", "Newport", "NP5 7AR"]
+
+Input:                              Output:
+(first_address, second_address)     "123<br />Testy<br />Place<br />Newport<br />NP5 7AR",
+```
+
+
+## get_current_date
+Returns the current date.
+
+##### Parameters: 
+- N/A
+
+##### Context:
+No real usage but a single question in Census currently only for test purposes.
+```
+The number of visitors staying in the household on {{ get_current_date() }}
+```
+
+##### Example:
+```
+Output:
+<span class='date'>27 December 2018</span>
+```
+
+## format_date
+Formats a date string, which can be in the format "YYYY-MM-DD", "YYYY-MM" or "YYYY".
+The output will be of the form "d MMMM YYYY", "MMMM YYYY" or "YYYY" depending on value passed.
+If the value is not a string it will just return the initial value.
+Uses Babel's [format_date](http://babel.pocoo.org/en/latest/api/dates.html#babel.dates.format_date).
+
+##### Parameters: 
+- value:    String value representing a datetime. Allowable formats "YYYY-MM-DD", "YYYY-MM" or "YYYY".
+
+##### Context:
+Used when a date is used in a question.
+```
+Are you able to report for the period from {{metadata['ref_p_start_date']|format_date}} to {{metadata['ref_p_end_date']|format_date}}?
+```
+
+##### Examples:
+```
+Input:          Output:
+"2019-01-01"    "<span class='date'>1 January 2019</span>"
+"2019-01"       "<span class='date'>January 2019</span>"
+"2019"          "<span class='date'>2019</span>"
+```
+
+
+## format_date_custom
+Returns a date in a defined format. In most cases in order to include the day of the week.
+Uses Babel's [format_datetime](http://babel.pocoo.org/en/latest/api/dates.html#babel.dates.format_datetime).
+
+##### Parameters: 
+- value:        String value representing a datetime. Allowable formats "YYYY-MM-DD".
+- date_format:  Format of the date to return. Default 'EEEE d MMMM YYYY'
+
+##### Context:
+Used in LMS for both Questions and Options in order to add day of week to a date. 
+Often combined with `calculate_offset_from_weekday_in_last_whole_week` filter.
+```
+Did you have a paid job, either as an employee or self-employed, in the week {{ calculate_offset_from_weekday_in_last_whole_week(collection_metadata['started_at'], {}) | format_date_custom( 'EEEE d MMMM YYYY' ) }} to {{ calculate_offset_from_weekday_in_last_whole_week(collection_metadata['started_at'], {}, 'SU') | format_date_custom( 'EEEE d MMMM YYYY' ) }}
+ 
+As an option:
+{{calculate_offset_from_weekday_in_last_whole_week(collection_metadata['started_at'], {}, 'MO') | format_date_custom('EEEE dd MMMM') }}
+```
+
+##### Examples:
+```
+Input:                              Output:
+("2018-08-14", "EEEE d MMMM YYYY")  "<span class='date'>Tuesday 14 August 2018</span>"
+("2018-08-14", "EEEE d MMMM")       "<span class='date'>Tuesday 14 August</span>"
+("2018-08-14", "EEEE d")            "<span class='date'>Tuesday 14</span>"
+("2018-08-14", "d MMMM YYYY")       "<span class='date'>14 August 2018</span>"
+```
+
+## format_conditional_date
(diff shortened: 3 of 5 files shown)


doc/jinja-filters.md
```diff
@@ -0,0 +1,232 @@
...
+<br></br>
+## get_current_date
+Returns the current Date.
+
+##### Parameters: 
+- N/A
+
+##### Context:
+No real current usage but a single question in Census I believe only for test purposes.
+`The number of visitors staying in the household on {{ get_current_date() }}`
```
MebinAbraham: Just for consistency, can we add break tags before these so they are not on the same line?

MebinAbraham (changes requested):

app/templating/template_renderer.py
```diff
@@ -15,8 +15,8 @@ def __init__(self):
         env.filters['concatenated_list'] = filters.concatenated_list
         env.filters['format_date'] = filters.format_date
         env.filters['format_date_custom'] = filters.format_date_custom
-        env.filters['format_household_name'] = filters.format_household_member_name
-        env.filters['format_household_name_possessive'] = filters.format_household_member_name_possessive
+        env.filters['format_household_name'] = filters.format_household_name
+        env.filters['format_household_name_possessive'] = filters.format_household_name_possessive
```
ajmaddaford, line 19: 👍

doc/jinja-filters.md
```diff
@@ -0,0 +1,248 @@
+# Jinja Filters Used by Schemas
+
+###### Date 14/12/2018
```
ajmaddaford: I don't think we need the date here as we'll have the date it merges to master in source control.

doc/jinja-filters.md
```diff
@@ -0,0 +1,248 @@
...
+Returns the given number formatted for a specific locale. 
+Uses Babel's number.py and format_decimal.
+
+##### Parameters: 
+- value: A single numeric value.
+
+##### Context:
+Most commonly used to display a previous answer's numeric value in a following question.
+<br />
+`Of the <em>{{answers['total-number-employees']|format_number}}</em> total employees employed on 14 December 2018, how many male and female employees worked the following hours?`
```
ajmaddaford: It's not clear what does `format_number` does in this example. It was only after looking into `format_decimal` that I realised it will add comma separators to larger numbers.

doc/jinja-filters.md
```diff
@@ -0,0 +1,248 @@
...
+The address list is returned with break points between each address line.
+
+##### Parameters: 
+- user_entered_address: List of address values from answers
+- metadata_address: List of address values from metadata
+
+##### Context:
+Used to display a full address for use in a description.
+<br />
+`"description": "{{ format_address_list ([answers['address-line1'], answers['address-line2'], answers['locality'], answers['town-name'], answers['postcode']], [metadata['address_line1'], metadata['address_line2'], metadata['locality'], metadata['town_name'], metadata['postcode']]) }}"`
```
ajmaddaford: Some strings contain the json element e.g. description, and others do not. As it doesn't seem to add anything I think we should remove them from all the examples.

doc/jinja-filters.md
ajmaddaford: It's cleaner to put the fixed width content in a ``` block, something like:
```
The number of visitors staying in the household on {{ get_current_date() }}
```

doc/jinja-filters.md
```diff
@@ -0,0 +1,248 @@
...
+
+##### Context:
+No real current usage but a single question in Census I believe only for test purposes.
+<br />
+`The number of visitors staying in the household on {{ get_current_date() }}`
+<br />
+
+
+## format_date
+Formats a date string, returning "d MMMM YYYY", "MMMM YYYY" or "YYYY" depending on value passed.
```
ajmaddaford: I think this needs to be clearer. What values can be passed, and what will the output be for those values?

doc/jinja-filters.md
```diff
@@ -0,0 +1,248 @@
...
+Often combined with calculate_offset_from_weekday_in_last_whole_week filter.
+<br />
+`Did you have a paid job, either as an employee or self-employed, in the week {{ calculate_offset_from_weekday_in_last_whole_week(collection_metadata['started_at'], {}) | format_date_custom( 'EEEE d MMMM YYYY' ) }} to {{ calculate_offset_from_weekday_in_last_whole_week(collection_metadata['started_at'], {}, 'SU') | format_date_custom( 'EEEE d MMMM YYYY' ) }}`
+ 
+`"label": "{{calculate_offset_from_weekday_in_last_whole_week(collection_metadata['started_at'], {}, 'MO') | format_date_custom('EEEE dd MMMM') }}"`
+<br />
+
+
+## format_conditional_date
+This is to be used when passing multiple dates as multiple arguments. It will return the first valid date formatted.
```
ajmaddaford, line 147: Possibly not for this pull request, but I wonder now that we have `first_non_empty_item` whether we could combine it with `format_date` so we don't need this filter, something like:
```
{{ first_non_empty_item(answers['period-from'], metadata['ref_p_start_date']) | format_date }}
```

doc/jinja-filters.md
```diff
@@ -0,0 +1,248 @@
...
+If a user decides to report for different dates then following questions should use the user entered date. Otherwise continue asking for the metadata period.
+<br />
+`For the period {{ format_conditional_date (answers['period-from'], metadata['ref_p_start_date'])}} to {{ format_conditional_date (answers['period-to'], metadata['ref_p_end_date'])}}, what was the value of {{ first_non_empty_item(metadata['trad_as'], metadata['ru_name']) }}’s <em>total retail turnover</em>?`
+<br />
+
+
+## calculate_offset_from_weekday_in_last_whole_week
+Offsets a date from a particular day of the week in the previous week. 
+Intended to be used to generate reference date ranges around a particular date
+The offsets will always be based on one day of the week in the previous Mon-Sun
```
ajmaddaford, line 171: It's still not clear what this does, more examples with output could help.

ajmaddaford (changes requested):
This definitely helps highlight what filters we currently have. For each example I think it would be helpful to also document the output. Also, for some of the filters that can be passed different values, it would be useful to have more than one example.

doc/jinja-filters.md
dwyeradam: I'll add some additional text but I also think the Output examples will really help with understanding all the filters especially these smaller ones.

doc/jinja-filters.md
dwyeradam: I agree and thought exactly the same. I'd like to check with Mebin again as there was a reason he didn't do it at the time of writing this new filter.

MebinAbraham: So `format_conditional_date` reuses the same private method that `first_non_empty_item` uses `_get_first_non_empty_item`.

Having it as two filter just mean't I didn't have to essentially duplicate
``` python
	date_format = 'd MMMM YYYY'
    if value and re.match(r'\d{4}-\d{2}$', value):
        date_format = 'MMMM YYYY'
    if value and re.match(r'\d{4}$', value):
        date_format = 'YYYY'
```
to check whether what it got passed in regex matched any one of the different formats we support before passing it into the `format_date` filter and then just returning the value if it wasn't a date. I guess you could also try: except to see if it would parse as a date. 🤷‍♂️ 

Just wasn't sure if this would have be very nice.

doc/jinja-filters.md
```diff
@@ -0,0 +1,438 @@
+# Jinja Filters Used by Schemas
+
+
+## format_number
+Returns the given number with the appropriate thousands grouping and decimal separator as determined by the locale.
+Uses Babel's number.py and format_decimal.
```
ajmaddaford: We could link to the Babel documentation for this method, something like:

Uses Babel's [format_decimal](http://babel.pocoo.org/en/latest/api/numbers.html#babel.numbers.format_decimal) method.

doc/jinja-filters.md
```diff
@@ -0,0 +1,438 @@
...
+
+##### Examples:
+```
+Input:          Output:
+"100000000"     "100,000,000"
+"123.40"        "123.4"
+```
+
+## format_currency
+Returns a formatted currency value. Uses Babel's number.py and format_currency.
```
ajmaddaford: Same as comment above about linking to the Babel documentation.

doc/jinja-filters.md
```diff
@@ -0,0 +1,438 @@
...
+```
+
+## format_address_list
+This function format_address_list accepts two addresses, a user submitted address and a metadata address. 
+If all the items in the user entered address list are 'Undefined' (nothing in answer_store) then we retrieve the metadata.
+The address list is returned with break points between each address line.
+
+##### Parameters: 
+- user_entered_address: List of address values from answers
+- metadata_address: List of address values from metadata
```
ajmaddaford: This method could do with being updated to remove the specifics about user vs metadata address. It looks like it just takes two addresses and outputs the first if it's not empty, else the second.

doc/jinja-filters.md
```diff
@@ -0,0 +1,438 @@
...
+first_address = [Undefined(), Undefined(), Undefined()]
+second_address = ["123", "Testy", "Place", "Newport", "NP5 7AR"]
+
+Input:                              Output:
+(first_address, second_address)     "123<br />Testy<br />Place<br />Newport<br />NP5 7AR",
+```
+
+
+## get_current_date
+Returns the current Date.
```
ajmaddaford: Date -> date

doc/jinja-filters.md
```diff
@@ -0,0 +1,438 @@
...
+
+##### Parameters: 
+- N/A
+
+##### Context:
+No real current usage but a single question in Census I believe only for test purposes.
```
ajmaddaford: We should be more definitive here and remove "I believe only ".

doc/jinja-filters.md
```diff
@@ -0,0 +1,438 @@
...
+Output:
+<span class='date'>27 December 2018</span>
+```
+
+## format_date
+Formats a date string, which can be in the format "YYYY-MM-DD", "YYYY-MM" or "YYYY".
+The output will be of the form "d MMMM YYYY", "MMMM YYYY" or "YYYY" depending on value passed.
+
+##### Parameters: 
+- value: Value representing a datetime.
```
ajmaddaford: This method looks like it will only do something if it's passed a string (if it's passed anything else it just returns the value). It might be clearer to move the acceptable string formats from the description above to here.

doc/jinja-filters.md
```diff
@@ -0,0 +1,438 @@
...
+
+##### Examples:
+```
+Input:                              Output:
+("2018-08-10", {}, "SU")            "2018-08-05"  # Friday outputs previous Sunday
+("2018-08-10", {}, "FR")            "2018-08-03"  # Friday outputs previous Friday
+("2018-08-10", {"years": 1}, "FR")  "2019-08-03"  # Friday outputs previous Friday + 1 year
+
+("2018-08-05", {}, "SU")            "2018-07-29"  # Sunday outputs previous Sunday (Must be a full Sunday)
+("2018-08-05", {"weeks": 1}, "SU")  "2018-08-05"  # Previous sunday with +1 month offset, back to input
```
bitdivision: Should this be `+1 week`?

dwyeradam: Yes but since I copied this from your test_jinja_filters test I blame you. I'll update both now.

doc/jinja-filters.md
```diff
@@ -0,0 +1,438 @@
...
+"2019-01"       "<span class='date'>January 2019</span>"
+"2019"          "<span class='date'>2019</span>"
+```
+
+
+## format_date_custom
+Returns a date in a defined format. In most cases in order to include the day of the week.
+
+##### Parameters: 
+- value: Value representing a datetime.
```
ajmaddaford: Should mention the string formats mentioned in the comment about `format_date` above.

doc/jinja-filters.md
```diff
@@ -0,0 +1,438 @@
...
+```
+Input:          Output:
+"2019-01-01"    "<span class='date'>1 January 2019</span>"
+"2019-01"       "<span class='date'>January 2019</span>"
+"2019"          "<span class='date'>2019</span>"
+```
+
+
+## format_date_custom
+Returns a date in a defined format. In most cases in order to include the day of the week.
```
ajmaddaford, line 120: Should mention the Babel [format_date](http://babel.pocoo.org/en/latest/api/dates.html#babel.dates.format_date) method that's used here.

doc/jinja-filters.md
```diff
@@ -0,0 +1,438 @@
...
+```
+For the period {{ format_conditional_date (answers['period-from'], metadata['ref_p_start_date'])}} to {{ format_conditional_date (answers['period-to'], metadata['ref_p_end_date'])}}, what was the value of {{ first_non_empty_item(metadata['trad_as'], metadata['ru_name']) }}’s <em>total retail turnover</em>?
+```
+
+##### Examples:
+```
+Input:                              Output:
+("2016-01-12", "2016-02-12")        "<span class='date'>12 January 2016</span>"
+("2017-12-23", None)                "<span class='date'>23 December 2017</span>"
+(None, "2017-12-24")                "<span class='date'>24 December 2017</span>"
```
ajmaddaford: The examples would be clearer if the first and second dates were consistent.

doc/jinja-filters.md
```diff
@@ -0,0 +1,438 @@
...
+```
+
+## format_conditional_date
+This is to be used when passing multiple dates as multiple arguments. It will return the first valid date formatted.
+
+##### Parameters: 
+- *dates: Multiple date values
+
+##### Context:
+If a user decides to report for different dates then following questions should use the user entered date. Otherwise continue asking for the metadata period.
```
ajmaddaford: This is very specific to how we use this filter rather than what the filter does.

doc/jinja-filters.md
ajmaddaford: This is a +1 **week** offset

doc/jinja-filters.md
```diff
@@ -0,0 +1,438 @@
...
+```
+
+##### Examples:
+```
+Input:                              Output:
+("2018-08-10", {}, "SU")            "2018-08-05"  # Friday outputs previous Sunday
+("2018-08-10", {}, "FR")            "2018-08-03"  # Friday outputs previous Friday
+("2018-08-10", {"years": 1}, "FR")  "2019-08-03"  # Friday outputs previous Friday + 1 year
+
+("2018-08-05", {}, "SU")            "2018-07-29"  # Sunday outputs previous Sunday (Must be a full Sunday)
```
ajmaddaford: What does must be a full Sunday mean?

doc/jinja-filters.md
```diff
@@ -0,0 +1,438 @@
...
+##### Context:
+Lists selected pay patterns for MWSS. Used as a description: 
+```
+{{ answers['pay-pattern-frequency-answer']|format_unordered_list }}
+```
+
+##### Examples:
+```
+Input:                  Output:
+[['item 1', 'item 2']]  "<ul><li>item 1</li><li>item 2</li></ul>"
```
ajmaddaford, line 336: Is the list within a list correct?

doc/jinja-filters.md
```diff
@@ -0,0 +1,438 @@
...
+            [
+                ['Jane', 'Sarah'], 
+                ['Smith', 'Smythe']
+            ]
+        ])
+
+Ouput: "<ul>
+            <li>John Smith</li>
+            <li>Jane Smith</li>
+            <li>first3 last3</li>
```
ajmaddaford: I think this line needs to go

doc/jinja-filters.md
```diff
@@ -0,0 +1,438 @@
...
+             '', 
+             'Bristol', 
+             'BS9 1AW'
+            ]
+        ], 
+        delimiter=', ')
+
+Ouput: "<ul>
+            <li>51 Testing Gardens, Bristol, BS9 1AW</li>
+       </ul>"
```
ajmaddaford, line 445: Where do we use this filter for an address?

ajmaddaford (changes requested):

doc/jinja-filters.md
dwyeradam: I really wanted to do this so I will. I was worried about moving the scope too much but I really hate it so we've got two votes to change it as part of this.

doc/jinja-filters.md
```diff
@@ -0,0 +1,438 @@
...
+("2017-07-31", "2018-08-16", "YYYY")                "<span class='date'>2017</span> to <span class='date'>2018</span>"
+("2018-08-14", "2018-08-16", "EEEE d")              "<span class='date'>Tuesday 14</span> to <span class='date'>Thursday 16</span>"
+```
+
+
+## first_non_empty_item
+Returns the first non empty value in a list of values.
+
+##### Parameters: 
+- *items: List of items where some of them could have no value.
```
MebinAbraham: Not sure but `List of items` to me sounds like 
``` python 
first_non_empty_item([1,'',3])
# returns [1, '', 3]
```
instead of
``` python
first_non_empty_item(1, '', 3)
# returns 1
```
Maybe `1 or more arguments`?
🤷‍♂️

doc/jinja-filters.md
```diff
@@ -0,0 +1,438 @@
...
+
+##### Context:
+Implemented to replace trad_as_or_runame hard coding.
```
MebinAbraham: `trad_as_or_ru_name`

doc/jinja-filters.md
bitdivision: I believe it was intended to show that the current Sunday (e.g. today) would never be output.

doc/jinja-filters.md
dwyeradam: That's what context means? `"the situation within which something exists or happens"`. You're right though I should be more generic with how to use it.

doc/jinja-filters.md
dwyeradam: As Richard says. I've changed it to say must be a full week which is clearer.

doc/jinja-filters.md
dwyeradam: Yes. It's one of the tests. Will try and clarify it's use if I actually understand it.

doc/jinja-filters.md
dwyeradam: We don't. I just wanted to show an example where the delimiter is changed and address seemed the best fit for that. I could remove it or have a much simpler example of Surname, First Name. Is this PR just about how it's currently used or how it could be used? Have I just opened myself up for even more changes?

dwyeradam:
@ajmaddaford, @MebinAbraham I should have addressed the comments now so ready for re-review. @ajmaddaford `format_unordered_list` is an odd one. That's the format that is passed into it. We may have to look at how we store the answer here. For now I've tried to be clearer on what format the filter is expecting.

doc/jinja-filters.md
ajmaddaford: I'd say it's about how it's currently used, as we want it to drive the schema changes for piping.

bitdivision (approved):

app/jinja_filters.py
```diff
@@ -42,35 +42,26 @@ def get_currency_symbol(currency='GBP'):
...
-        If all the items in address list are 'Undefined' (nothing in answer_store) then we retrieve the metadata.
+        The function format_address_list accepts two lists of address values.
+        If all the items in the first address list are empty or 'Undefined' then we use the second address.
 
-        :param user_entered_address user entered address which is a list of answer fields
-        :param metadata_address metadata address which is a list of answer fields passed to us
-        :return: the value of the address to be piped
+        :param (list) first_address
+        :param (list) second_address
+        :return: first address is values present else second address
```
MebinAbraham: `if values`

doc/jinja-filters.md
```diff
@@ -0,0 +1,445 @@
...
+possible_items = ['item 1', 'item 2', 'item 3', 'item 4']
+list_items = [['item 1', 'item 3']]
+
+Input:                          Output:
+(possible_items, list_items)    "<ul><li>item 2</li><li>item 4</li></ul>"
+```
+
+
+## format_household_summary
+A merge of format_household_name and format_unordered_list in order to return a list of full names in "ul" and "li' tags
```
MebinAbraham: If you got some free time on your hands, feel free to highlight any variable/function names in sentences. 

A merge of `format_household_name` and `format_unordered_list` in order to return a list of full names in "ul" and "li' tags

dwyeradam: No I need you to point them all out.

dwyeradam: I've probably got them all. In the end not that many.

doc/jinja-filters.md
ajmaddaford: I've had a look at why this is and it's because the list of checkbox answers are being appended  to another list at https://github.com/ONSdigital/eq-survey-runner/blob/master/app/templating/schema_context.py#L116. It does this as it also deals with multiple answers in repeating groups where the answer value could not be a list. We should only create a list if the answer being returned is not already a list.

MebinAbraham (approved):

ajmaddaford (approved):

doc/jinja-filters.md
dwyeradam: Added card https://trello.com/c/DrZyF3sP to address this.

PR #1949, opened 2018-12-20
Change ukis to data version 0.0.1.  Added qcodes

insacuri:
https://trello.com/c/kUnLvqmA/772-implement-ukis-transform

UKIS was originally a 0.0.2 schema, but SDX downstream only knows how to deal with 0.0.1.  We started building the capability to handle this, but it would be faster and less risky if we just converted this schema.

### How to review 
Fill in the survey and make sure that the submission data is a dictionary full of qcodes and values rather then it being an array of dictionaries that contain the data.

### Checklist

### DIFF
--- a/data/en/ukis_0001.json
+++ b/data/en/ukis_0001.json
@@ -3,7 +3,7 @@
     "form_type": "0001",
     "mime_type": "application/json/ons/eq",
     "schema_version": "0.0.1",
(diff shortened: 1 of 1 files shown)


tgandrews (approved):
LGTM

dwyeradam (approved):

PR #1951, opened 2018-12-28
Evaluate MultipleSurveyError when processing CSRFError

### LINKED ISSUE #1899: Session Expired superseding MultipleSurveyError
### Expected behaviour
When launching an additional survey within the same browser when you go back to the initial survey and try to continue you should see an error message "Unfortunately you can only complete one survey at a time".

### Actual behaviour
Initially you see the "Your session has expired" message. Only after trying to go back do you see the "Unfortunately you can only complete one survey at a time." error message.

Start two surveys in a browser go back to the initial survey and try and continue it. 

### Technical information
Hits CSRFError first.
Check why `test_multiple_surveys_open.py` isn't catching this issue.

dwyeradam:
Fixes #1899

### Checklist


app/views/errors.py
```diff
@@ -23,13 +23,24 @@ class MultipleSurveyError(Exception):
...
 def unauthorized(error=None):
     log_exception(error, 401)
     return render_template('session-expired.html'), 401
 
 
+@errors_blueprint.app_errorhandler(CSRFError)
+def csrf_error(error=None):
+    metadata = get_metadata(current_user)
+    if metadata:
+        if check_multiple_survey(metadata, request.view_args):
```
bitdivision: Might be a little cleaner to combine these `if`s with an `and`

bitdivision (dismissed):
Good change :+1:

Very minor style point, not a blocking issue though.

app/views/errors.py
dwyeradam: I noticed that this morning and thought i'd been stupid not to do it at the time.

dwyeradam: Made the change but was naughty and squashed at the same time since it was pretty minor.

MebinAbraham (dismissed):
Looks good 👍

app/views/questionnaire.py
```diff
@@ -67,17 +67,14 @@ def before_questionnaire_request():
     logger.bind(tx_id=metadata['tx_id'])
 
     values = request.view_args
+
+    if check_multiple_survey(metadata, values):
```
MebinAbraham, line 71: As discussed, this can be removed.

Edit: Needed for GET requests.

dwyeradam:
@bitdivision can you have a quick peak at the later commit. Decent change since you reviewed it but all removals rather than additional code.

app/views/questionnaire.py
dwyeradam: Keeping as we've now found we need this for page refresh.

app/views/questionnaire.py
```diff
@@ -104,13 +101,6 @@ def before_post_submission_request():
     logger.bind(eq_id=values['eq_id'], form_type=values['form_type'])
     logger.info('questionnaire request', method=request.method, url_path=request.full_path)
 
-    _check_same_survey(url_eq_id=values['eq_id'],
```
MebinAbraham, line 107: As discussed, still needed.

app/views/questionnaire.py
```diff
@@ -104,12 +101,13 @@ def before_post_submission_request():
...
-                       url_collection_id='',
-                       session_eq_id=session_data.eq_id,
-                       session_form_type=session_data.form_type,
-                       session_collection_id='')
+    metadata_from_session_data = {
+        'tx_id': session_data.tx_id,
+        'eq_id': session_data.eq_id,
+        'form_type': session_data.form_type,
+    }
+    if check_multiple_survey(metadata_from_session_data, values):
```
MebinAbraham, line 109: We probably should also add this check in feedback?.
https://github.com/ONSdigital/eq-survey-runner/blob/master/app/views/feedback.py#L29-L36

At the moment, if i try POST a feedback on a previous survey, it says 'Thanks for your feedback', even though it was rejected by CSRF.

dwyeradam: So the code wouldn't get hit here as it handles the CSRFError first. So it is already hitting the new `csrf_error` handler in `errors.py`. Since it doesn't have `request.view_args` values it returns false from `check_multiple_survey`. It is therefore returning the `session-expired.html` but that doesn't filter to the page as a whole. If i'm honest I don't know how to easily fix this.

MebinAbraham (approved):

bitdivision (approved):
:+1:

PR #1952, opened 2019-01-04
Cypress Tests

bitdivision:
Not for review just yet.

Cypress provides a better developer experience when running functional tests.

Unfortunately it has some performance issues in comparison to webdriver / chimp.

To work around these, I've combined all the tests into one file (cypress/aggregated/index.spec.js) at runtime. This reduces test time by about half, but does mean you have to search for the description when looking for a failure since the filename isn't provided.

It is still slightly slower than webdriver / chimp on my machine.

### How to review 
Do we want to use cypress?

I would suggest looking at the spec files, they've changed a fair bit from selenium. The most major difference is that everything is a jQuery object which gives you quite a lot of flexibility.

Check what happens on test failure etc.

### Checklist


bitdivision:
Even with the combined spec files it looks as though travis is taking ~20 minutes instead of ~10 minutes with selenium.

bitdivision:
Closing since there's no obvious way to further speed this up and doubling the functional test time isn't really acceptable at the moment.

It might be that cypress improves it's performance drastically in future, in which case this could be revisited.

PR #1953, opened 2019-01-08
Revert all input fields from type number to text

bitdivision:
The input type of 'number' was causing browsers except
chrome to clear the value entirely rather than passing
it on to server side validation.

This fix just reverts the type of these inputs to 'text'.

### Checklist


pricem14pc (approved):

bameyrick (approved):

bitdivision:
I've updated this to include two more instances which @dwyeradam  caught and I missed.

PR #1958, opened 2019-01-09
Check for action before redirect to signed-out

dwyeradam:
To stop sign-out on POSTS on the `thank-you`, `view-submission` and `feedback` pages unless there is a `sign_out` action. 

### How to review 
Initial issue was caught by using the print button on view-submission page. Pressing cancel/print on that would result in a POST to the page that subsequently signed you out.

### Checklist

### DIFF
--- a/app/views/feedback.py
+++ b/app/views/feedback.py
@@ -91,7 +91,10 @@ def thank_you():
 @feedback_blueprint.route('/thank-you', methods=['POST'])
 @login_required
 def post_thank_you():
-    return redirect(url_for('session.get_sign_out'))
+    if 'action[sign_out]' in request.form:
+        return redirect(url_for('session.get_sign_out'))
+
+    return redirect(url_for('feedback.thank_you'))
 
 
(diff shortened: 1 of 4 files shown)


dwyeradam:
Raised PR so feel free to have a look at my implementation. I'd like to add some new tests though so not Prod ready. Will probably fail CodeCov anyway.

MebinAbraham (approved):

dwyeradam:
Didn't fail code coverage but add an extra view-submission test for peace of mind anyway. Squashed as easily seen in `test_view_submission.py`

bitdivision (approved):
I would have preferred if there was a way to do this without redirecting at all, but I can't see an obvious way to do this.

e.g. with this change:

I run through the survey, view my submission and look through the answers. 1 hour later I come back and click print, the page refreshes, and I'm signed out. If I wanted to print it again, I wouldn't be able to.

I would suggest that this is not what a user would expect to happen.

PR #1961, opened 2019-01-11
Fix routing bug in UKIS questionnaire.

samiwel:
The requirement is to skip to question 10 as per the slides, but it was previously skipping to 9.11.

Refactor to use skip condition on the group as it's cleaner than having routing rules on the interstitial page.

### How to review 
UKIS routing should be tested to ensure it works as expected.

### DIFF
--- a/data/en/ukis_0001.json
+++ b/data/en/ukis_0001.json
@@ -3210,6 +3210,67 @@
             "groups": [{
                 "id": "group613",
                 "title": "Co-operation on Innovation",
+                "skip_conditions": [{
+                    "when": [{
+                            "id": "answer3206",
+                            "condition": "not set"
+                        },
+                        {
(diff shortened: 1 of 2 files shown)


jonnyshaw89 (approved):

dwyeradam (approved):
Can we get this applied to the NI version as well.

dwyeradam (changes requested):
Can we get this applied to the NI version as well.

dwyeradam (changes requested):
Can we get this applied to the NI version as well.

dwyeradam (dismissed):
Can we get this applied to the NI version as well.

dwyeradam (approved):

dwyeradam:
@samiwel just needs a squash and rebase.

PR #1962, opened 2019-01-14
Add the language to google analytics

bitdivision:
Send the language to google analytics. [trello](https://trello.com/c/FeKTx8Zi/2548-google-analytics-language-value)

Field for language documented here: https://developers.google.com/analytics/devguides/collection/analyticsjs/field-reference#language

### How to review 
Ensure javascript is enabled 😆 	

Set the EQ_UA_ID environment variable, and change the analytics URL in _base.html to end in `analytics_debug.js` rather than `analytics.js`.

Open LMS 2 and change language to welsh. Ensure that the console shows that google analytics is sending the language.
### Checklist

### DIFF
--- a/app/templates/layouts/_base.html
+++ b/app/templates/layouts/_base.html
@@ -57,6 +57,8 @@
 
     ga('create', '{{analytics_ua_id}}', 'auto');
     ga('set', 'anonymizeIp', true);
+    // Language is setup as a custom dimension in google analytics
+    ga('set', 'dimension1', '{{language_code}}' || 'en');
     ga('send', 'pageview');
     </script>
     {% endif %}


bitdivision:
Mebin noted that setting 'language' may override the browser language, which would also be useful information.

GA has the concept of custom dimensions, you can have up to 20 of them and they allow you to set a custom attribute per hit / session / user etc.

https://support.google.com/analytics/answer/2709828

I've checked with Alison, and we currently don't have any custom dimensions, so this will be `dimension1`. 

Each custom dimension needs to be added in the google analytics admin before it will show up.

MebinAbraham (approved):
Was able to see the initial implementation in GA but not the custom dimension which I'm assuming is due to the update delay. But able to view actions set for GA locally so 👍 

Just need to remember to create a custom dimension for the main GA account.

MebinAbraham (commented):
Tested in GA and it's showing up now. 👍

bitdivision:
Best way to view this is to go to 'customization' -> 'Custom report' and add a new report with the custom dimension as a dimension.

app/templates/layouts/_base.html
```diff
@@ -57,6 +57,8 @@
 
     ga('create', '{{analytics_ua_id}}', 'auto');
     ga('set', 'anonymizeIp', true);
+    // Language is setup as a custom dimension in google analytics
+    ga('set', 'dimension1', '{{language_code}}');
```
MebinAbraham: Just a thought...I know we have the user's language from the browser as well but it might be worth defaulting the language for pages that we don't have the `language_code`.

dwyeradam (approved):

app/templates/layouts/_base.html
bitdivision: True enough. I'll do that.

MebinAbraham (commented):
👍 As expected.

dwyeradam:
@bitdivision. Happy for you to squash and rebase.

PR #1963, opened 2019-01-16
Update README to make it clearer that DynamoDB container is required in setup

adammahmood:
Minor change to `README` to make it clearer that DynamoDB container must be started during setup

### How to review 
- Ensure tests pass
- Ensure README changes make sense

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ Note, you will also need to run an upstream tool (eg, https://github.com/ONSDigi
 docker run -e SURVEY_RUNNER_SCHEMA_URL=http://docker.for.mac.host.internal:5000 -it -p 8000:8000 onsdigital/go-launch-a-survey:latest
 ```
 
-If you wish to view submitted data you will also need to run an additional upstream tool (eg, https://github.com/ONSDigital/eq-docker-dynamodb) to launch a dynamoDB container.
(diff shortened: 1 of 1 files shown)


dwyeradam (approved):
This is correct but raises the point that this shouldn't cause an issue. It used to be the case that DynamoDB was only used for submitted responses and therefore exceptions were handled when it wasn't available. I'm assuming that when we added DynamoDB use for Session etc this protection was removed to use a generic method for Dynamo read/writes. We would only want it to 500 if any of `EQ_QUESTIONNAIRE_STATE_DYNAMO_READ/WRITE` or `EQ_USED_JTI_CLAIM_DYNAMO_READ/WRITE` or `EQ_SESSION_DYNAMO_READ/WRITE` were set to True. 

With this failure it now means that we're reliant on DynamoDB regardless of what it's used for. We should probably have a separate PR to avoid this reliance.

Since these are all now True in Prod these settings should also probably be defaulted to True in `settings.py`. Again a different PR from this.

samiwel (approved):

adammahmood:
@dwyeradam How shall we capture / handle these bits of work? Do `eq-survey-runner` have a `Technical Improvements` board we can add and prioritise?

dwyeradam:
@adammahmood the 500 error will be solved as part of the work that will be done for v3. No real benefit for doing it in Runner as it stands as it's only a Dev issue. I'll raise a card to change default settings to match those in Prod though which will mitigate that as DynamoDB will become the default storage.

PR #1964, opened 2019-01-17
Remove javascript session timeout modal

adammahmood:
Removes the javascript session timeout modal so users no longer have the ability to continue a session.  This change eliminates a potential cause of the unexpected log out issues in production.

### How to review 
- Ensure tests pass
- Ensure timeout modal no longer appears when running `test_timeout.json` schema

### Checklist


adammahmood:
[Remove session timeout javascript from runner](https://trello.com/c/dm6AG2Mu/808-remove-session-timeout-javascript-from-runner)

tgandrews (approved):
Changes make sense to me and the tests pass on my local with this so LGTM

dwyeradam (changes requested):
There are a couple of things that we should also do as part of this. They do complicate things some what and without Python knowledge will be difficult so ideally @samiwel or @jonnyshaw89 could help out with these changes. Otherwise I will.

So the first is the removal of any code around a defined session timeout grace period. System setting `EQ_SESSION_TIMEOUT_GRACE_PERIOD_SECONDS`. This was used to give some extra time to continue with the survey before the server timeout kicked in. We'll have to also remove this from any build scripts that use this setting but I wouldn't be surprised if the default (30 secs) isn't just used everywhere.

There is also a `session_prompt_in_seconds` defined in the schema. We don't need this now as there is no prompt. Any reference to this in code should also be removed (including eq-schema-validator). 

Currently we've just removed the Functional test. It will probably be a better idea to have one that checks for the server side session expiry. I'm assuming there is a Integration test equivalent

adammahmood:
@dwyeradam I'll give it a go. There are some existing tests around the server side session timeout.  Are these sufficient?

https://github.com/ONSdigital/eq-survey-runner/blob/a013fbe04d052916c3f383ef3087d0c3f4f36874/tests/integration/session/test_timeout.py

https://github.com/ONSdigital/eq-survey-runner/blob/6c6e6d27bbfd6ca5fd0bf81d4b2f6a4f522a1eef/tests/integration/questionnaire/test_questionnaire_session_expired.py

dwyeradam:
> @dwyeradam I'll give it a go. There are some existing tests around the server side session timeout. Are these sufficient?
> 
> [/tests/integration/session/test_timeout.py@`a013fbe`](https://github.com/ONSdigital/eq-survey-runner/blob/a013fbe04d052916c3f383ef3087d0c3f4f36874/tests/integration/session/test_timeout.py)
> 
> [/tests/integration/questionnaire/test_questionnaire_session_expired.py@`6c6e6d2`](https://github.com/ONSdigital/eq-survey-runner/blob/6c6e6d27bbfd6ca5fd0bf81d4b2f6a4f522a1eef/tests/integration/questionnaire/test_questionnaire_session_expired.py)

So I don't think we need to worry about integration tests. As you've pointed out they should and seem to already exists for server side session expiry. I've gone back and fore as to wether integration is enough in this instance. I settled on a single functional test that tries to submit after a very short session timeout would be worthwhile to confirm the user experience. Others may disagree but it would be equivalent the integration tests except confirming the correct Session Timeout page is shown.

app/globals.py
```diff
@@ -56,10 +56,6 @@ def get_session_timeout_in_seconds(schema, with_grace_period=True):
         schema_session_timeout < default_session_timeout else \
```
tgandrews: Can we remove this `schema_session_timeout` logic?

tests/integration/star_wars/test_light_side_path.py
```diff
@@ -36,7 +36,6 @@ def test_light_side_path(self):
         self.assertInUrl(star_wars_test_urls.STAR_WARS_SUMMARY)
 
         # We are on the review answers page
-        self.assertInBody('>Star Wars</')
```
tgandrews, line 39: Why are these checks for the questionnaire title being removed?

tgandrews (approved):
Tested locally and it works 🎉

samiwel (approved):
Tested locally seems to work as expected

tests/integration/session/test_timeout.py
```diff
@@ -32,12 +30,3 @@ def test_when_session_times_out_server_side_401_is_returned(self):
         time.sleep(3)
         self.get(self.last_url)
         self.assertStatusUnauthorised()
-
-    def test_schema_defined_timeout_is_used(self):
-        self._application.config['EQ_SESSION_TIMEOUT_SECONDS'] = 10
-        self.launchSurvey('test', 'timeout')
-        self.assertInHead('window.__EQ_SESSION_TIMEOUT__ = 6')
-
-    def test_schema_defined_timeout_cant_be_higher_than_server(self):
```
MebinAbraham: Quick point, unless I missed somehting obvious, these would still be valid tests.
*Ensuring schema defined timeout cannot be longer than app default is useful.*

MebinAbraham: Quick point, I think this is still worth keeping.
*Ensuring schema defined timeout cannot be longer than the app default is useful.*

adammahmood: As far as I understand, this test only ensures that `window.__EQ_SESSION_TIMEOUT__ = 2` is correctly set which has now been removed in `/app/templates/layouts/_base.html`.

MebinAbraham: I guess a way of still achieving the same result would be to set `'EQ_SESSION_TIMEOUT_SECONDS'` lower than the schema timeout. Then you can `self.assertStatusUnauthorised()` after the sleeping for app specified timeout. This should then ensure that it used the app value and not the higher schema value.

dwyeradam: So at the moment during the test setUp the system timeout is being set to two seconds. Previously the schema had a timeout of 6 seconds which is greater than the system setting of two. Therefore the test was checking that  it used the system (2) rather than schema defined (6) timeout. This should therefore stay but it might be worth either decreasing the system defined timeout to 1 second in the setUp() or increasing the schema defined timeout to at least 3.

dwyeradam (changes requested):
Two really small things but otherwise good thanks.

tests/functional/spec/timeout.spec.js
```diff
@@ -1,77 +1,12 @@
...
-      return browser
-        .setValue(TimeoutBlockPage.timeout(), 'foo')
-        .click(TimeoutBlockPage.submit())
-        .getUrl().should.eventually.contain('summary')
-        .waitForVisible(dialog,5000)
-        .click('.js-timeout-save')
-        .getUrl().should.eventually.contain('localhost');
+      return browser.pause(testTimeoutJSON.session_timeout_in_seconds * 1000).then(() => {
+        return helpers.pressSubmit(1).getSource().should.eventually.contain('Your session has expired');
+      });
```
dwyeradam: To keep it in line with the other tests it should be:
```
const helpers = require('../helpers');
const TimeoutBlockPage = require('../generated_pages/timeout/timeout-block.page.js');

describe('Timeout', function() {
  it('Given I am completing an electronic questionnaire, when I have been inactive for the timeout period and attempt to submit data, then I will be redirected to a page confirming my session has timed out', function() {
    return helpers.openQuestionnaire('test_timeout.json').then(() => {
      return browser
          .pause(3000)
          .click(TimeoutBlockPage.submit())
          .getSource().should.eventually.contain('Your session has expired');
    });
  });
});

```

tests/functional/spec/timeout.spec.js
```diff
@@ -1,12 +1,13 @@
...
 
 describe('Timeout', function() {
-  it('Given I am completing an electronic questionnaire, when I have been inactive for X minutes and attempt to submit data, then I will be redirected to a page confirming my session has timed out', function() {
+  it('Given I am completing an electronic questionnaire, when I have been inactive for the timeout period and attempt to submit data, then I will be redirected to a page confirming my session has timed out', function() {
     return helpers.openQuestionnaire('test_timeout.json').then(() => {
-      return browser.pause(testTimeoutJSON.session_timeout_in_seconds * 1000).then(() => {
-        return helpers.pressSubmit(1).getSource().should.eventually.contain('Your session has expired');
-      });
+      return browser
+          .pause(3000)
```
dwyeradam: This is right on the limit now that you've increased the schema timeout so it might be worth increasing this slightly but happy to approve and you can just do that with the squash.

dwyeradam (approved):

PR #1965, opened 2019-01-17
Updating address-line claim names in the household schema to address_…

pricem14pc:
…line

Changed all address-line-[1|2] references to address_line[1|2] to align with metadata claim attribute names. This was previously presenting a comma ( , ) as an answer option in various questions rather than the actual address line.

### How to review 
launch eQ census_household and address shown in various answer options are a comma separated concatenation of the address lines 1 and 2 (and not simply a comma).

These are...
term-time-location-answer
term-time-location-answer-with-second-address
term-time-location-answer-with-second-address-non-uk
past-usual-address-answer
visitor-begin-section

### DIFF
--- a/data/en/census_household.json
+++ b/data/en/census_household.json
@@ -29,14 +29,6 @@
             "validator": "string"
         },
         {
-            "name": "address_line1",
-            "validator": "optional_string"
-        },
-        {
-            "name": "address_line2",
-            "validator": "optional_string"
-        },
-        {
             "name": "locality",
             "validator": "optional_string"
         },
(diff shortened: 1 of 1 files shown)


bitdivision (changes requested):
If the metadata for these address lines is now required, we should change the validator from "optional_string" to "string".

pricem14pc:
> If the metadata for these address lines is now required, we should change the validator from "optional_string" to "string".

I've just double checked with RM re: the Hannah 0.1 sample and there are null values for some address_lines. However, there will always be a non null display_address generated. I've updated the schema to use display_address.

bitdivision (approved):

bitdivision:
@pricem14pc Just a heads up, you'll need to setup signed commits before you squash this. https://help.github.com/articles/signing-commits/

MebinAbraham (approved):

MebinAbraham (commented):
There is no need for us to still define metadata that isn't referenced in the schema?. 

So things like `address-line1`, if it is no longer needed in the schema because we now use `display_address` and RH takes care of changing address then this can be removed.

PR #1966, opened 2019-01-17
Add Google Datastore backend

kierenbeckett:
Add a Google Datastore backend.  A new env variable then lets you choose between using Datastore, DynamoDB or SQL.  This PR removes the DynamoDB -> Postgres fallback functionality, you can only use one backend at a time.


bitdivision:
@kierenbeckett 	It would be nice if we could default to a datastore backend. It looks like this current points at dynamodb by default. 

I wonder if we could use something like https://hub.docker.com/r/knarz/datastore-emulator/ to get everything running locally?

app/setup.py
```diff
@@ -91,6 +115,8 @@ def create_app(setting_overrides=None):  # noqa: C901  pylint: disable=too-compl
 
     setup_dynamodb(application)
 
+    setup_datastore(application)
```
ajmaddaford: As we're going to have multiple backends and they will be mutually exclusive, I think it would be clearer here to have a `setup_storage` method that does the work of determining which `setup_` method to call. Something like: 

```python
def setup_storage(application):
  if application.config['EQ_STORAGE_BACKEND'] == 'sql':
	setup_database(application)
  elif application.config['EQ_STORAGE_BACKEND'] == 'dynamodb':
    setup_dynamodb(application):
  elif application.config['EQ_STORAGE_BACKEND'] == 'datastore':
    setup_datastore(application)
  else:
    raise Exception('EQ_STORAGE_BACKEND is not valid')
```

docker-compose.yml
```diff
@@ -40,30 +34,21 @@ services:
...
-      EQ_SESSION_DYNAMO_READ: "True"
-      EQ_SESSION_DYNAMO_WRITE: "True"
       EQ_USED_JTI_CLAIM_TABLE_NAME: dev-used-jti-claim
-      EQ_USED_JTI_CLAIM_DYNAMO_READ: "True"
-      EQ_USED_JTI_CLAIM_DYNAM_WRITE: "True"
-      AWS_ACCESS_KEY_ID: "${AWS_ACCESS_KEY_ID:-dummy}"
-      AWS_SECRET_ACCESS_KEY: "${AWS_SECRET_ACCESS_KEY:-dummy}"
+      EQ_DATASTORE_PROJECT_ID: dummy
+      EQ_DATASTORE_EMULATOR_CREDS: "True"
+      DATASTORE_EMULATOR_HOST: datastore:8432
```
ajmaddaford, line 47: I can't see `DATASTORE_EMULATOR_HOST` being used in runner

bitdivision:
About to review this properly, initial thought is that it would be nice if we could add something to the readme indicating how to run the emulator?

app/storage/data_access.py
```diff
@@ -92,78 +29,16 @@ def put(model, overwrite=True):
...
-        return False
+    _get_backend().delete(model)
 
-    # If we do not have a config key, assume dynamo is enabled
-    if 'dynamo_write_key' not in config:
-        return True
 
-    return app.config[config['dynamo_write_key']]
+def _get_backend():
+    return STORAGE_BACKENDS[current_app.config['EQ_STORAGE_BACKEND']]
```
ajmaddaford: I'm not sure this class is needed. The decision on which backend to use is made when the app starts up so could we instead set a global `storage` property that references the appropriate storage class? That storage class could have a client property set on it to store the specific storage client. Something like:

```python
client = datastore.Client(...)
application.eq['storage'] = new DataStore(client)
```

Usages would then be changed to:
```python
application.eq['storage'].get_by_key(...)
```

tests/app/storage/test_data_access.py
```diff
@@ -17,121 +13,31 @@
 class TestDataAccess(AppContextTestCase):
 
     def test_get_by_key(self):
-        dynamo_item = {'user_id': USER_ID, 'state_data': STATE_DATA, 'version': VERSION}
+        dynamo_item = QuestionnaireState(USER_ID, STATE_DATA, VERSION)
```
bitdivision: Could be renamed?

app/setup.py
```diff
@@ -49,6 +51,28 @@
 logger = get_logger()
 
 
+class EmulatorCreds(google.auth.credentials.Credentials):
```
bitdivision: I'm nitpicking but generally abbreviations should be avoided.

app/settings.py
```diff
@@ -64,19 +64,16 @@ def get_env_or_fail(key):
 EQ_SERVER_SIDE_STORAGE_USER_ID_ITERATIONS = ensure_min(int(os.getenv('EQ_SERVER_SIDE_STORAGE_USER_ID_ITERATIONS', '10000')),
                                                        1000)
 
+EQ_STORAGE_BACKEND = os.getenv('EQ_STORAGE_BACKEND', 'datastore')
+EQ_DATASTORE_PROJECT_ID = os.getenv('EQ_DATASTORE_PROJECT_ID')
+EQ_DATASTORE_EMULATOR_CREDS = parse_mode(os.getenv('EQ_DATASTORE_EMULATOR_CREDS', 'False'))
```
bitdivision:
```suggestion
EQ_DATASTORE_EMULATOR_CREDENTIALS = parse_mode(os.getenv('EQ_DATASTORE_EMULATOR_CREDENTIALS', 'False'))
```

app/setup.py
```diff
@@ -49,6 +51,28 @@
...
+    """A mock credential object.
+    Used to avoid unnecessary token refreshing or reliance on the network
+    while an emulator is running.
```
bitdivision: The docstring could be updated to show that it is used for google cloud authentication.

app/storage/datastore.py
```diff
@@ -0,0 +1,83 @@
...
+        'key_field': 'eq_session_id',
+        'table_name_key': 'EQ_SESSION_TABLE_NAME',
+        'schema': app_models.EQSessionSchema,
+    },
+    app_models.UsedJtiClaim: {
+        'key_field': 'jti_claim',
+        'table_name_key': 'EQ_USED_JTI_CLAIM_TABLE_NAME',
+        'schema': app_models.UsedJtiClaimSchema,
+    },
+}
```
ajmaddaford, line 31: I'm not sure what the solution is, but it feels wrong to have the properties specific to each model - `key_field` and `schema` in the storage class and not in the app_models.

ajmaddaford (dismissed):

data/en/ukis_0001.json
```diff
@@ -141,7 +141,7 @@
                                     "options": [{
                                             "q_code": "0410",
                                             "label": "{{ first_non_empty_item(metadata['trad_as'], metadata['ru_name']) }} was established",
-                                            "value": "{{ first_non_empty_item(metadata['trad_as'], metadata['ru_name']) }} was established"
+                                            "value": "The business was established"
```
bitdivision: Not sure how this got in here?

app/storage/data_access.py
```diff
@@ -1,88 +1,25 @@
...
-    },
-    app_models.UsedJtiClaim: {
-        'table_name_key': 'EQ_USED_JTI_CLAIM_TABLE_NAME',
-        'dynamo_read_key': 'EQ_USED_JTI_CLAIM_DYNAMO_READ',
-        'dynamo_write_key': 'EQ_USED_JTI_CLAIM_DYNAMO_WRITE',
-        'key_field': 'jti_claim',
-        'schema': app_models.UsedJtiClaimSchema,
-        'sql_model': models.UsedJtiClaim,
-    },
+STORAGE_BACKENDS = {
```
bitdivision: :+1:

bitdivision (changes requested):
Looks good in general, some relatively minor points.

data/en/ukis_0001.json
kierenbeckett: My PR wasn't building, pulled this from #1972

app/setup.py
kierenbeckett: Yeh, I wondered about this after copying it over. Will update

docker-compose.yml
kierenbeckett: It's read directly by the Datastore library

app/storage/data_access.py
kierenbeckett: Agreed this might be slightly nicer, I'll try it out

app/storage/datastore.py
kierenbeckett: It's not ideal.  I think much of the schema stuff is fairly redundant where the backends essentially take JSON as input already.  Potentially some of it could be taken out entirely, but didn't want to make too many changes in one go.

ajmaddaford: Yeah, I'm happy for a separate card to be created to review the use of the schemas.

README.md
```diff
@@ -104,7 +104,7 @@ docker run -e SURVEY_RUNNER_SCHEMA_URL=http://docker.for.mac.host.internal:5000
 To submit data you will also need to run an additional upstream tool (eg, https://github.com/ONSDigital/eq-docker-dynamodb) to launch a dynamoDB container.
```
bitdivision:
```suggestion
The data storage backend is configurable with the `EQ_STORAGE_BACKEND` environment variable. You will need to run an emulator if you are setting this to DynamoDB or Google Datastore:

DynamoDB: https://github.com/ONSDigital/eq-docker-dynamodb
Google Datastore: https://hub.docker.com/r/knarz/datastore-emulator/
```

README.md
```diff
@@ -104,7 +104,7 @@ docker run -e SURVEY_RUNNER_SCHEMA_URL=http://docker.for.mac.host.internal:5000
...
 
 ```
-docker run -it -p 6060:8000 onsdigital/eq-docker-dynamodb:latest
+docker run -it -p 8432:8432 knarz/datastore-emulator:latest
```
bitdivision, line 115:
```suggestion
docker run -it -p 6060:8000 onsdigital/eq-docker-dynamodb:latest
OR
docker run -it -p 8432:8432 knarz/datastore-emulator:latest
```

bitdivision (approved):
Looks good. I've suggested some changes to the readme, just to enumerate the options on the backend.

bitdivision (changes requested):
If the backend is set to SQL, store submitted response is causing a 500.

README.md
pricem14pc: There is a list of env variables in the readme which appears to be out of date (even prior to this PR). I'll raise a separate ticket to cover that.. and to introduce the new env vars delivered by this PR

bitdivision (approved):
Readme has been updated to say that SQL doesn't support view submitted responses. I think in the long run we may want to remove SQL entirely or properly support it for all runner features.

scripts/dev_settings.sh
```diff
@@ -28,12 +28,16 @@ if [ -z "$EQ_RABBITMQ_HOST_SECONDARY" ]; then
   export EQ_RABBITMQ_HOST_SECONDARY="localhost"
 fi
 
-if [ -z "$SQLALCHEMY_DATABASE_URI" ]; then
-  export SQLALCHEMY_DATABASE_URI="sqlite:////tmp/questionnaire.db"
+if [ -z "DATASTORE_EMULATOR_HOST" ]; then
```
MebinAbraham: `DATASTORE_EMULATOR_HOST` > `$DATASTORE_EMULATOR_HOST`

MebinAbraham (changes requested):

MebinAbraham (commented):
Should we be enforcing the group of ENV VARS?
So if `EQ_STORAGE_BACKEND` is `datastore` then the `EQ_DATASTORE_PROJECT_ID` must be defined?

kierenbeckett:
Probably, there are plenty of other sets of variables already there with dependencies between them that aren't enforced e.g. rabbit enabled + rabbit host.  Perhaps a ticket could be created to sort it out more generally?

scripts/dev_settings.sh
```diff
@@ -32,50 +32,34 @@ if [ -z "$SQLALCHEMY_DATABASE_URI" ]; then
   export SQLALCHEMY_DATABASE_URI="sqlite:////tmp/questionnaire.db"
 fi
 
-if [ -z "$EQ_DYNAMODB_ENDPOINT" ]; then
```
MebinAbraham: The same way the `$SQLALCHEMY_DATABASE_URI` was added back in, this should be kept to allow easy switching of storage when developing.

Happy with everything else.

MebinAbraham (approved):
Have tested locally with all backends and tested with GC datastore, working as expected.
👍

kierenbeckett:
:confetti_ball:

PR #1967, opened 2019-01-18
Document the use of group instances

MebinAbraham:
Highlight the current use of group instances for repeating groups.

### How to review 
- Read the new `.md` files and make sure it makes sense. 

Let me know of anything that will make it more understandable.

I have not covered every possible scenario but more of an overview of when and why they are used.

### Checklist

### DIFF
--- a/doc/group_instances.md
+++ b/doc/group_instances.md
@@ -1,0 +1,281 @@
+### Group Instance & Group Instance IDs
+
+---
+
+It is recommended that you familiarise yourself with [*repeating routing rules*](repeating_routing_rules.md) before proceeding with this.
+
+Group Instance / Instance IDs are used to identify the instance of an answer acquired from a repeating group.
+
+#### Group Instance (Index)
+An answer can have multiple instances, the *group_instance* tracks this instance and is part of the [*Location*](../app/questionnaire/location.py) object which consists of *group_id*, *group_instance*, and the *block_id*.
+Currently, `group_instance` is always used as it is part of the Location. Any non-repeating answers will have a `group_instance` of `0` by default.
+
+##### Group Instance - Repeats
+
+The routing [*path*](../app/questionnaire/path_finder.py) is a list of *Location* objects which is created by following routing rules and evaluating skip conditions for all groups and blocks in the schema. 
+*The generation of the routing path is carried out on every request.*
+
+When generating the Path, for each group in the schema, we increment the *group_instance* based on the number of times the group repeats. The *no. of repeats* is determined based on the evaluation of the `repeat` rule on each group. 
+
+
+##### Example
+
+```json
+"routing_rules": [{
+    "repeat": {
+        "type": "until",
+        "when": [{
+            "id": "repeating-anyone-else",
+            "condition": "equals",
+            "value": "No"
+        }]
+    }
+}]
+```
+The *no. of repeats* we will have for the above routing rule is based on the number of answers we have with the answer id *repeating-anyone-else* to which the answer is not `"No"`. The *no. of repeats* is incremented for each answer we have with matching id until the routing rule evaluates to *True*.
+If there are no answers yet, it will repeat once.
+
+---
+
+#### Group Instance ID
+
+Unlike `group_instance`, the `group_instance_id` is used only when a *Location* has group dependencies. A `group_instance_id` is a *UUID4* string which represents the instance of an answer. 
+Locations without dependencies are given the default value of `None` for the `group_instance_id`. 
+
+*The instance of an answer is also stored in `answer_instance` as part of the [Answer](../app/data_model/answer.py) object. However, `answer_instance` is now deprecated in favour of `group_instance_id`.*
+
+
+##### What defines a group dependency?
+A group dependency is a mapping of a `group_id` to the block ids that drive that group. For example if *"group-x"* repeats based on an answer in *"block-a"*, then *"group-x"* is dependent on the driving *"block-a"* - `{ 'group-x': ['block-a'] }`.
+
+Using repeating groups as an example. If `block A` collects a list of people's names and `block B` collects details about each person in the list that `block A` collected, we can say that `block B` is dependent on `block A` being answered. In this scenario, we refer to `block B` as a `dependent` and `block A` as a `driver`. 
+A group can also have multiple dependencies. 
+
+When parsing the schema, a [*GroupDependencies*](../app/questionnaire/group_dependencies.py) object is created which stores all the dependencies for a schema by walking through the repeating routing rules.
+
+
+*GroupDependencies* holds a list of dependents, group drivers and block drivers.
+
+
+> ***dependents*** (key, value) - the group that is dependent and its dependencies.
+
+> ***group drivers*** - list of blocks driving a group dependency (used with answer_ids)
+
+> ***block drivers*** - list of blocks driving a group dependency (used with answer_id)
+
+We only store unique group/block drivers.
+
+Dependencies for a group is evaluated if the group has any of the following types of repeats:
+- `answer_count`
+- `answer_count_minus_one`
+
+For groups with either of these repeat types, if the group also has a `skip_condition` then the `id` used in the `when` clause will also be added to the list of dependencies.
+
+
+An answer is given a `group_instance_id` for the following scenarios:
+ - the location's `group_id` is listed as a dependent or a group driver.
+ - the location's `block_id` is listed as a group or block driver.
+
+If either of the above conditions are true then the answer is allocated a new `group_instance_id` if there are no existing answers for the given Location.
+If the current Locations `group_id` is listed as a ***dependent*** (relies on another group/block) then the current answer is given the same `group_instance_id` as the driver.
+To do this, we get a list of all `group_instance_id` for each group/block driver then use the `group_instance` from the current Location to return the matching `group_instance_id`.
+
+##### Example
+Schema used: [test_routing_repeat_until.json](../data/en/test_routing_repeat_until.json)
+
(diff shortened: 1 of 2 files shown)


doc/group_instances.md
```diff
@@ -0,0 +1,269 @@
...
+
+#### Group Instance (Index)
+An answer can have multiple instances, the *group_instance* tracks this instance and is part of the [*Location*](../app/questionnaire/location.py) object which consists of *group_id*, *group_instance*, and the *block_id*.
+
+Currently, `group_instance` is always used as it is part of the Location. Any non-repeating answers will have a `group_instance` of `0` by default.
+
+##### Group Instance - Repeats
+
+On every request to `/questionnaire/<eq_id>/<form_type>/<collection_id>/`, we generate the full routing [*Path*](../app/questionnaire/path_finder.py) which holds a list of *Location* objects.
+In short, every block in every group is added to the path. Any block or group with skip conditions or routing rules which evaluates *True* is not added to the path.
```
bitdivision: This doesn't quite make sense to me. I think something like this might be better?

`Starting from the first block in the questionnaire, the routing path is created by following routing rules and evaluating skip conditions until the end of the survey is reached.`

doc/group_instances.md
```diff
@@ -0,0 +1,269 @@
...
+}]
+```
+The *no. of repeats* we will have for the above routing rule is based on the number of answers we have with the answer id *repeating-anyone-else* to which the answer is not `"No"`. The *no. of repeats* is incremented for each answer we have with matching id until the routing rule evaluates to *True*.
+If there are no answers yet, it will repeat once.
+
+---
+
+#### Group Instance ID
+
+Unlike `group_instance`, the `group_instance_id` is used only when a *Location* has group dependencies. A `group_instance_id` is a *UUID4* string which represents the instance of an answer. Locations without dependencies are given the default value of `None` for the `group_instance_id`.
```
bitdivision: A high level definition of a group dependency would be useful here.

bitdivision (changes requested):
Initial comments

doc/group_instances.md
MebinAbraham: The next paragraph highlights `GroupDependencies`?.

bitdivision: As discussed IRL, most of this section relies on the idea of a dependency, but I don't think it's ever explained what makes one group dependent on another group / block

doc/repeating_routing_rules.md
```diff
@@ -0,0 +1,121 @@
...
+
+---
+
+### Repeat Type: answer_value
+
+Used to repeat a group `n` times, where `n` is an integer provided by the user.
+
+##### Context
+
+Commonly used to obtain household visitor's name for the number of visitors specified by the user.
```
ajmaddaford: "name" should probably be "details" as we collect more than just their name?

doc/group_instances.md
```diff
@@ -0,0 +1,269 @@
+### Group Instance & Group Instance IDs
+
+---
+
+It is recommended that you familiarise yourself with [*repeating routing rules*](repeating_routing_rules.md) before proceeding with this.
+
+Group Instance / Instance IDs are used to identify the instance of an answer acquired from a repeating group.
+
+#### Group Instance (Index)
+An answer can have multiple instances, the *group_instance* tracks this instance and is part of the [*Location*](../app/questionnaire/location.py) object which consists of *group_id*, *group_instance*, and the *block_id*.
```
ajmaddaford, line 10: We also have answer instances, although I'm not sure they're used now we don't collect the household member names all on one page.

doc/group_instances.md
```diff
@@ -0,0 +1,269 @@
...
+
+Currently, `group_instance` is always used as it is part of the Location. Any non-repeating answers will have a `group_instance` of `0` by default.
+
+##### Group Instance - Repeats
+
+Starting from the first block in the schema, the routing [*path*](../app/questionnaire/path_finder.py) which holds a list of *Location* object is created by following routing rules and evaluating skip conditions until the end of the schema is reached.
```
ajmaddaford: "until the end of the schema" isn't quite true, it will generate a routing path as far as it can in a schema. For example if a route is dependent on an answer that hasn't been answered (that doesn't have a default) then it will stop.

doc/group_instances.md
```diff
@@ -0,0 +1,269 @@
...
+- `answer_count_minus_one`
+
+For groups with either of these repeat types, if the group also has a `skip_condition` then the `id` used in the `when` clause will also be added to the list of dependencies.
+
+
+An answer is given a `group_instance_id` for the following scenarios:
+ - the location's `group_id` is listed as a dependant or a group driver.
+ - the location's `block_id` is listed as a group or block driver.
+
+An answer is allocated a new `group_instance_id` if there are no existing answers for a given Location.
```
bitdivision: And it is inside a repeat?

MebinAbraham: Referring to if either of the above condition is true. I'll make it clearer.

doc/group_instances.md
MebinAbraham: Added some brief context, let me know if you think it needs more.

doc/group_instances.md
```diff
@@ -0,0 +1,278 @@
...
+
+When parsing the schema, a [*GroupDependencies*](../app/questionnaire/group_dependencies.py) object is created which stores all the dependencies for a schema by walking through the repeating routing rules.
+
+
+*GroupDependencies* holds a list of dependents, group drivers and block drivers.
+
+```
+dependents (key, value) - Key: the group that depends on another group/block. Value: group/block that it depends on (drivers).
+group drivers - list of groups/blocks driving a group dependency (used with group_ids/answer_ids)
+block drivers - list of blocks driving a group dependency (used with answer_id)
```
ajmaddaford: This isn't very clear (and I'm not sure why it's in a pre-formatted block). I think the dependents is covered by the extra content that's been added above. 

For the group drivers definition, I think we should remove the reference to groups/group_ids as it's not used:
> group drivers - list of blocks driving a group dependency (used with answer_ids)

doc/group_instances.md
```diff
@@ -0,0 +1,278 @@
...
+
+---
+
+#### Group Instance ID
+
+Unlike `group_instance`, the `group_instance_id` is used only when a *Location* has group dependencies. A `group_instance_id` is a *UUID4* string which represents the instance of an answer. Locations without dependencies are given the default value of `None` for the `group_instance_id`. 
+
+
+##### What defines a group dependency?
+A group is said to have a dependency if it cannot be self sufficient without knowledge of another group/block.
```
ajmaddaford: "self sufficient" is too vague. Something like this would be clearer:

> A group dependency is a mapping of a dependent group name to the block ids that drive that dependency. For example if "group-x" repeats based on an answer in "block-a", then "group-x" is  dependent on the driving "block-a" - `{ 'group-x': ['block-a'] }`.

doc/group_instances.md
```diff
@@ -0,0 +1,278 @@
...
+
+
+Using repeating groups as an example. If `group A` collects a list of people's names and `group B` collects details about each person in the list that `group A` collected, we can say that `group B` is dependent on `group A` being answered. In this scenario, we refer to `group B` as a `dependent` and `group A` as a `driver`.
```
ajmaddaford: The drivers are always block ids rather than group ids.

ajmaddaford (changes requested):

doc/group_instances.md
MebinAbraham: I didn't talk about answer_instance as we no longer use them. `RepeatingAnswer` is only in the test schemas and currently census but I assume it's just not updated.

Do you think it would be good to define answer_instance and how they work?

doc/group_instances.md
MebinAbraham: Changed to highlight its for every `group/block` in the schema, and not necessarily the `survey end`.

doc/group_instances.md
MebinAbraham: Previously kept `group_id` since the code was there to support it even though it was not in any schema.`

Updated the definitions.

doc/group_instances.md
MebinAbraham: Same reason as before with support for group_ids. Removed now.

doc/group_instances.md
```diff
@@ -0,0 +1,279 @@
...
+
+---
+
+#### Group Instance ID
+
+Unlike `group_instance`, the `group_instance_id` is used only when a *Location* has group dependencies. A `group_instance_id` is a *UUID4* string which represents the instance of an answer. Locations without dependencies are given the default value of `None` for the `group_instance_id`. 
+
+
+##### What defines a group dependency?
+A group dependency is a mapping of a dependent's `group_id` to the block ids that drive that dependency. For example if *"group-x"* repeats based on an answer in *"block-a"*, then *"group-x"* is dependent on the driving *"block-a"* - `{ 'group-x': ['block-a'] }`.
```
bitdivision: @ajmaddaford I think removing the references to dependency would help here?
```
A group dependency is a mapping of a `group_id` to the block ids that drive that group. For example if *"group-x"* repeats based on an answer in *"block-a"*, then *"group-x"* is dependent on the driving *"block-a"* - `{ 'group-x': ['block-a'] }`.
```

doc/group_instances.md
ajmaddaford: I think it would be worth acknowledging the existence of answer instances and just say that they are deprecated.

ajmaddaford (approved):

bitdivision (approved):
:+1:

bitdivision:
Squishy squashy

PR #1972, opened 2019-01-25
Bug - Downstream processing of UKIS failing due to answer piping 

tgandrews:
For the second question "Did any of the following significant events or changes occur to ESSENTIAL ENTERPRISE LTD?" the first answer option has piping for both the "label" and "value". Having piping in the value appears to block the answer being passed downstream. We therefore need to replace the piping in the value with a hard coded string. 

### How to review 
1. Ensure that the new hardcoded value is being posted to the server when you select that your business was established.
2. Ensure this is happening for both UKIS variants.

### DIFF
--- a/data/en/ukis_0001.json
+++ b/data/en/ukis_0001.json
@@ -141,7 +141,7 @@
                                     "options": [{
                                             "q_code": "0410",
                                             "label": "{{ first_non_empty_item(metadata['trad_as'], metadata['ru_name']) }} was established",
-                                            "value": "{{ first_non_empty_item(metadata['trad_as'], metadata['ru_name']) }} was established"
+                                            "value": "The business was established"
                                         },
                                         {
                                             "q_code": "0420",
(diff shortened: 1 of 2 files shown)


tgandrews:
[UKIS Issue with piping going into an answer value](https://trello.com/c/Wqo5RMKh/845-ukis-issue-with-piping-going-into-an-answer-value)

dwyeradam (approved):

dwyeradam:
When you test locally if you open a tab and use url http://localhost:5000/dump/submission you can see the output that is sent downstream. Within the data object we would expect to see `"0410": "The business was established"` amongst any other options for that and the previous question.
```
"data": {
   "0210": "UK regional within approximately 100 miles of this business",
   "0410": "The business was established"
},
```

jonnyshaw89 (approved):

tgandrews:
@dwyeradam I didn't know that. I can see it in there too.

PR #1975, opened 2019-01-29
Remove non-census and non-test schemas

MebinAbraham:
Removed any non-census and non-test schemas.
Removed the session timeout endpoint as it used a live survey to test and  was missed from #1964 
Have also renamed the star wars and rogue one test schemas to keep it consistent.

*I have also removed the use of `variables` which was defined in the schema. It caused coverage issues when test using live surveys which defined `variables` were removed but turns out we don't actually use it.*

*`variables` was used to pass context to the templates, but we no longer use this and have not been for a while I believe.*

### How to review 
- Ensure non-census and non-test schemas have been removed.

There are tests that specifically test business surveys such as MCI, these probably should be removed at some point?.


iwootten:
+50 lines extra? Crumbs.

ajmaddaford:
Looks good to me. I've raised a card for removing the survey specific integration tests (mci, rsi and star wars).

ajmaddaford (approved):

bitdivision:
I think we should remove test_star_wars.json. The schema is no longer representative of runner's functionality. 

Probably best to do that under a separate PR since it's a fairly large change. I know we'll be losing the majority of path_finder tests, and some navigation tests. I would suggest that path finder tests will need to be rethought, rather than modified to use a different schema. 

I would imagine most other integration tests could be switched to test schemas without a huge amount of trouble.

bitdivision (approved):
:+1:

PR #1977, opened 2019-01-29
Add missing qcodes to UKIS

tgandrews:
To resolve missing `qcodes` in UKIS

### How to review 
1. Ensure that the qcodes are used when selecting the answers.

### DIFF
--- a/data/en/ukis_0001.json
+++ b/data/en/ukis_0001.json
@@ -162,7 +162,7 @@
                                     "label": "",
                                     "description": "",
                                     "options": [{
-                                        "q_code": "d1",
+                                        "q_code": "0440",
                                         "label": "No significant changes occurred",
                                         "value": "No significant changes occurred"
                                     }]
@@ -271,7 +271,7 @@
                                     "label": "",
                                     "description": "",
                                     "options": [{
-                                        "q_code": "d2",
+                                        "q_code": "d1",
                                         "label": "No significant changes",
                                         "value": "No significant changes"
                                     }]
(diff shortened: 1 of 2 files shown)


tgandrews:
[UKIS - 0440 & 2671 Qcodes to be added to Runner](https://trello.com/c/53uPoWiU/855-ukis-0440-2671-qcodes-to-be-added-to-runner)

jonnyshaw89 (approved):

samiwel (approved):
LGTM. Need to just confirm that this doesn't have any knock-on effect with SDX. @insacuri is this likely to cause you any problems?

insacuri (approved):
The changes look fine.  I do have a question though.  Are you pressed for time at the minute?  If so, then I'll approve it and you can merge it in as is, but if not, can you sort out the d codes so they start from 1?  It's a hassle but keeping this tidy for future people will be helpful as I'm sure the question of 'oh, why didn't they start from 1, how hard could it be?' will be asked

insacuri:
And @samiwel this will 100% have a knock on effect with SDX as they're expecting those d codes and are doing something with it.  SDX will need some work

insacuri (changes requested):
Changed my mind.  This should be done properly.  These chanages are fine but the other d codes need to be reordered

samiwel:
Hey @insacuri can you explain what the d codes are used for? We can ensure that they appear in incremental order, but we're lacking knowledge about what they are used for and what is their purpose?

tgandrews:
@insacuri Thanks for the feedback. I didn't know anything of dummy codes before. I've updated them so they increment correctly now but this changes it in a number of places is this what you expected?

insacuri (approved):

Issue #1976, opened 2019-01-29
Optional String validator treats attribute as Mandatory


iwootten:
### Expected behaviour

Optional strings may be missed or supplied in metadata when defined in a survey schema.

### Actual behaviour

Optional strings are required in metadata when specified in a survey schema.

### Steps to reproduce the behaviour

Create a session missing optional string metadata through a custom functional test and the the survey won't start correctly.

#### Browser
71.0.3578.98
#### Operating System

OSX 10.12

### Screenshot
Using the census household schema without a locality (optional) yields the following error:

<img width="870" alt="screen shot 2019-01-29 at 14 27 07" src="https://user-images.githubusercontent.com/3598161/51915006-fd850100-23d1-11e9-95df-8aad8846b70e.png">

PR #1980, opened 2019-02-01
Add Census Household Functional Test

iwootten:

Added a "happy path" functional test for census household. Includes both cypress and selenium flavours, for discussion over pros and cons of each. The cypress version resides under functional/cypress and includes its own helpers. The selenium helpers also needed to be modified due to the issue I've raised here: https://github.com/ONSdigital/eq-survey-runner/issues/1976

Each one takes a significant amount of time to complete (100s) with cypress taking about 20s longer overall (120s total). To my mind, the extra context cypress provides is a lot of help during long tests like this, though it's superior parallel performance can't be seen.

We can remove the approach we decide against once we're all happy.


bitdivision:
Some of the main issues I had with cypress previously:

- No way to use keyboard navigation (e.g. tab between forms)
- No ability to open new tabs (useful when checking multiple surveys log you out etc.)
- Slower performance (particularly when using different spec files)

I agree that the developer experience using cypress is far better than webdriver etc. I would suggest that the performance difference doesn't matter much at the moment (120 vs 100s), even if we end up testing 20 different census surveys, that's 6 minutes, which I don't think is a huge deal.

I would suggest that Cypress is the way to go for these census functional tests.

bitdivision:
Jon pointed out that Cypress also only supports chrome / electron. 

This could become an issue if we have to do cross-device / browser testing for census rehearsal.

MebinAbraham (changes requested):
Can we update the README with the new changes.

tests/functional/cypress/helpers/jwt_helper.js
```diff
@@ -0,0 +1,133 @@
+const KJUR = require('jsrsasign');
+const uuid = require('uuid/v1');
+const JSONWebKey = require('json-web-key');
+const jose = require('node-jose');
+const JWK = jose.JWK;
+const JWE = jose.JWE;
+
+const crypto = require('crypto');
+
+const signingKeyString = '-----BEGIN RSA PRIVATE KEY-----\n' +
```
MebinAbraham: Can we not read these from files like we do with the current `jwt_helper`?
Also I don't get why we actually need a `jwt_helper` specifically for cypress?. We can just reuse the one we already have?

iwootten: Cypress doesn't have access to the fs module - so that's why these have been pulled in as strings. If you attempt to use the original jwt_helper, it will drop out failing to find the module. Creating a cypress version allows the two sets of tests to run alongside one another whilst we decide which approach is best.

package.json
```diff
@@ -18,6 +18,7 @@
     "test_unit": "gulp test:scripts:unit:watch",
     "test_unit_no_watch": "gulp test:scripts:unit",
     "test_functional": "chimp tests/functional/chimp.js",
+    "test_census": "./node_modules/cypress/bin/cypress open",
```
MebinAbraham: Can we point this at the spec file by default because at the moment it open a UI and requires another click to run the spec. We need to able to use this in CI potentially?

Can we also configure it so cypress runs in headless mode when `EQ_RUN_FUNCTIONAL_TESTS_HEADLESS` is set, same way the current functional test works?

tests/functional/cypress/support/commands.js
```diff
@@ -0,0 +1,25 @@
+// ***********************************************
```
MebinAbraham: Do we need this file?

MebinAbraham:
Apart from my minor comments, we just need to decide between Selenium or Cypress 🤷‍♂️

package.json
MebinAbraham: Also, we don't need to point to node_modules within `package.json`. 
```suggestion
    "test_census": "cypress run --spec 'tests/functional/cypress/spec/census-household.spec.js'",
```

And if we just wanted to run all the cy tests, we can just do run.
```suggestion
    "test_census": "cypress run",
```

jonnyshaw89: Can we also have an option to not run it headless?

tests/functional/cypress/support/commands.js
iwootten: Nope, part of cypress template, removed it.

package.json
iwootten: Yes, we can. Resolved.

iwootten: Used the current env var to do this as per @MebinAbraham's suggestion

tests/functional/cypress/plugins/index.js
```diff
@@ -0,0 +1,33 @@
...
+  config.viewportWidth = 1280
+  config.viewportHeight = 1080
+  config.video = false
+  config.integrationFolder = "tests/functional/cypress/spec/"
+  config.fixturesFolder = false
+  config.supportFile = false
+
+  config.headed = process.env.EQ_RUN_FUNCTIONAL_TESTS_HEADLESS ? false : true
+
+  console.log(config) // see what all is in here!
```
bitdivision: Still needed?

iwootten: Nope. Removed.

tests/functional/cypress.js
```diff
@@ -0,0 +1,12 @@
+const cypress = require('cypress');
+
+cypress.run({
+  headed: process.env.EQ_RUN_FUNCTIONAL_TESTS_HEADLESS ? false : true,
```
MebinAbraham, line 4: The double negative is confusing. But life isn't fair so 🤷‍♂️

iwootten: Yeah, I can't change the cypress interface unfortunately.

tests/functional/cypress.js
```diff
@@ -0,0 +1,12 @@
...
+  config: {
```
MebinAbraham, line 5: Instead of duplicating, can we read in?
https://github.com/ONSdigital/eq-survey-runner/blob/4b57796c5051d3dabcb18d8838666813e878a632/tests/functional/cypress/plugins/index.js#L14

iwootten: That plugin file isn't actually used, I've removed it.

tests/functional/jwt_helper.js
```diff
@@ -8,15 +7,47 @@ const JWE = jose.JWE;
 
 const crypto = require('crypto');
 
-const signingKey = './tests/functional/sdc-user-authentication-signing-rrm-private-key.pem';
-const encryptionKey = './tests/functional/sdc-user-authentication-encryption-sr-public-key.pem';
+const signingKeyString = '-----BEGIN RSA PRIVATE KEY-----\n' +
+  'MIIEogIBAAKCAQEAvZzMraB96Wd1zfHS3vW3z//Nkqz+9HfwViNje2Y5L6m3K/7r\n' +
+  'aA0kUsWD1f6X7/LIJfkCEctCEj9q19+cX30h0pi6IOu92MlIwdH/L6CTuzYnG4PA\n' +
+  'CKT8FZonLw0NYBqh8p4vWS8xtNHNjTWua/FFTlxdtYnEb9HbUZkg7dXAtnikozlE\n' +
```
MebinAbraham, line 13: Not a blocker but I think it would be nicer if we used the ES6 backticks `` ` ` `` syntax for multiline strings instead of concatenating each string.

iwootten: Tried it at @ajmaddaford's request, doesn't work with our current setup.

tests/functional/cypress.js
```diff
@@ -0,0 +1,12 @@
+const cypress = require('cypress');
+
+cypress.run({
```
MebinAbraham, line 3: The only other thing that I am concerned about is that fact that we should have a config for a dev env?.

So this `.run()` is used just to run the test but when developing we probably want `cypress.open({})` configured which would allow us actively and easily write the tests with visibility of changes. I am thinking `.open()` would have config such as not exiting when a test fails, auto picking up code changes and re running etc. Thoughts?

iwootten: Do our current selenium tests work like this?

The api for cypress.open differs from that of the run call - it doesn't have a headed option for instance. There's nothing stopping a developer running cypress directly if they want such flexibility. 

Personally, I find the way that open automatically re-runs tests after any change is kind of annoying, especially with something as long-running as the census.

tests/functional/generate_pages.py
```diff
@@ -515,7 +517,7 @@ def process_schema(in_schema, out_dir, spec_file, require_path='..'):
     else:
         if os.path.isdir(args.SCHEMA):
             for root, dirs, files in os.walk(args.SCHEMA):
-                for file in [os.path.join(root, file) for file in files if file.startswith(args.test_schema_prefix)]:
+                for file in [os.path.join(root, file) for file in files if file.startswith(args.test_schema_prefix) or file.startswith(args.census_schema_prefix)]:
```
bitdivision: We could probably just make this generate pages for all schemas? It's fast and doesn't get uploaded into the repo anymore. 

Not a big deal though.

tests/functional/cypress.js
bitdivision: I think running the cypress UI seperately is fine.

bitdivision (dismissed):
Should we update the run_functional_tests.sh script to include this?

tests/functional/generate_pages.py
iwootten: Sure, I think all the schemas match these filters anyway. Updated.

tests/functional/cypress.js
MebinAbraham: Our current selenium tests don't work like this..thats the point?. We said we are using cypress because it is easier to develop with but at the moment its just straight up replacing `chimp` as a runner?. Also if you did `cypress open` manually it wouldn't pick up the spec folder automatically etc.

With our current chimp set up, we can do things like `@watch` to re-run a test once you made changes.

I would suggest at least adding how to run cypress interactively `(cypress open)` and run a spec file etc to the README.

iwootten: The reason it has to be run in this way is its the only way to accommodate the headless behaviour desired. I can add a minimum cypress.json file to allow either open/run, but it'll mean some duplication between the two configs.

cypress.json
```diff
@@ -1,5 +1,7 @@
 {
     "fixturesFolder": false,
     "pluginsFile": false,
-    "supportFile": false
+    "supportFile": false,
+    "integrationFolder": "tests/functional/cypress/spec/",
```
bitdivision, line 5: :+1:

README.md
```diff
@@ -204,6 +205,10 @@ To run the tests against a remote deployment you will need to specify the enviro
 
 `EQ_FUNCTIONAL_TEST_ENV=https://staging-new-surveys.dev.eq.ons.digital/ yarn test_functional`
 
+To run the census functional tests within the cypress UI:
+
+./node_modules/cypress/bin/cypress open
```
MebinAbraham, line 217: I think it would be nicer to add it to yarn and add a yarn command here to keep it consistent the rest of the app.

MebinAbraham (approved):
👍

bitdivision (approved):

PR #1981, opened 2019-02-05
Onboard QPSES

SamGodwin2:
Onboarding of the 3 different form types of QPSES detailed in https://trello.com/c/poTAm72n/817-onboard-qpses

### How to review 
New schemas all work correctly.

### DIFF
--- a/data/en/qpses_0001.json
+++ b/data/en/qpses_0001.json
@@ -1,0 +1,816 @@
+{
+    "eq_id": "qpses",
+    "form_type": "0001",
+    "mime_type": "application/json/ons/eq",
+    "schema_version": "0.0.1",
+    "data_version": "0.0.1",
+    "survey_id": "160",
+    "title": "Quarterly Public Sector Employment Survey",
+    "sections": [{
+        "id": "qpses-section",
+        "title": "QPSES",
+        "groups": [{
+                "id": "qpses-group",
+                "title": "QPSES",
+                "blocks": [{
+                        "type": "Introduction",
+                        "id": "introduction-block",
+                        "primary_content": [{
+                            "type": "Basic",
+                            "id": "use-of-information",
+                            "content": [{
+                                "description": "The survey collects information on the number of employees in the public sector. Information is published in the quarterly Public Sector Employment Statistical Bulletin and is used to understand the performance of the labour market and UK economy, to provide accountability to Parliament and support policy making.",
+                                "list": [
+                                    "Data should relate to all sites in England, Scotland and Wales unless otherwise stated.",
+                                    "You can provide informed estimates if actual figures aren’t available.",
+                                    "We will treat your data securely and confidentially."
+                                ]
+                            }]
+                        }],
+                        "preview_content": {
+                            "id": "preview",
+                            "title": "Information you need",
+                            "content": [{
+                                "description": "You can select the dates of the period you are reporting for, if the given dates are not appropriate."
+                            }],
+                            "questions": [{
+                                    "question": "Permanent employees",
+                                    "content": [{
+                                            "title": "Include",
+
+                                            "list": [
+                                                "full-time and part-time employees",
+                                                "male  and female employees",
+                                                "agency workers paid directly from any of the organisation’s payrolls",
+                                                "those temporarily absent but still on the payrolls, for example on maternity leave"
+                                            ]
+                                        },
+                                        {
+                                            "title": "Exclude",
+                                            "list": [
+                                                "agency workers paid directly from the agency payroll",
+                                                "self-employed workers",
+                                                "voluntary workers",
+                                                "former employees only receiving a pension",
+                                                "directors who do not receive a salary"
+                                            ]
+                                        }
+                                    ]
+                                },
+                                {
+                                    "question": "Permanent full-time equivalent employees",
+                                    "content": [{
+                                            "title": "Include",
+                                            "list": [
+                                                "male and female employees",
+                                                "full-time employees, counted as 1 full-time equivalent",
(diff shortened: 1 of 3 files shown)


SamGodwin2:
[Onboard QPSES](https://trello.com/c/poTAm72n/817-onboard-qpses)

data/en/qpses_160.json
```diff
@@ -0,0 +1,820 @@
+{
+    "eq_id": "qpses",
+    "form_type": "160",
+    "mime_type": "application/json/ons/eq",
+    "schema_version": "0.0.1",
+    "data_version": "0.0.2",
```
insacuri: The data version needs to be 0.0.1 for sdx to be able to transform the data without a ton of work.  This applies for all 3 schemas.

data/en/qpses_160.json
```diff
@@ -0,0 +1,820 @@
...
+                                    },
+                                    "decimal_places": 0
+                                },
+                                {
+                                    "id": "PT-male-perm-employees",
+                                    "mandatory": true,
+                                    "type": "Number",
+                                    "label": "Part-time male permanent employees",
+                                    "description": "Working less than your organisation's normal weekly hours",
+                                    "q_code": "Q552",
```
insacuri: The q_code needs to be 552, not Q552.  This applies for every qcode in this PR

insacuri (changes requested):

insacuri:
Also, when I try and fire this up using the go launcher I get:
<code>raise InvalidTokenException('incorrect data in token for {}'.format(name)) from error sdc.crypto.exceptions.InvalidTokenException: incorrect data in token for date_ref</code>

I assume it's just a date like '2016-05-01' but maybe the go launcher should be updated to add a sensible default for this

EDIT:  Same with the tradas_or_runame.  These things should be precompleted in the go launcher

data/en/qpses_165.json
```diff
@@ -0,0 +1,824 @@
+{
+    "eq_id": "qpses",
+    "form_type": "165",
+    "mime_type": "application/json/ons/eq",
+    "schema_version": "0.0.1",
+    "data_version": "0.0.2",
+    "survey_id": "406",
```
insacuri: I think the survey_id and form_type are wrong.  I've been working from 
```
Survey ID - Form Type
160 - 0001
165 - 0002
169 - 0003
```
I'm going to get that confirmed later today, but I believe these are the correct ones.  They need to change for all 3 surveys

insacuri (changes requested):

data/en/qpses_165.json
SamGodwin2: Which one of those 2 are which and where would you find these details, doesn't seem to be anything about them on the slides?

insacuri: Mark K put it on this card that I've been working on: https://trello.com/c/MZRXgJCu/711-qpses-qcode-work-for-sdx-transformation 

Not sure if you can see it, but that's where.

SamGodwin2: Nah don't seem to have access, I'll just use what you've posted above, just to double check which of those numbers are the form_type and which are the survey_id?

insacuri: I edited the previous comment to show which is which.  I had it included already but it was hidden because I messed up the markdown >_<

SamGodwin2:
@insacuri Made the changes in the comments, as for the date_ref metadata not sure there's much point pre-loading it into go-launcher.Seems a bit pointless to preload these one time use(?) metadata keys as it would just lead to data getting added to go-launcher that doesn't get used again.

data/en/qpses_0001.json
```diff
@@ -0,0 +1,816 @@
...
+                            }]
+                        }]
+                    },
+                    {
+                        "id": "total-fte-block",
+                        "type": "Question",
+                        "questions": [{
+                            "id": "total-fte-question",
+                            "title": "On {{ metadata['date_ref'] | format_date }}, what was {{ first_non_empty_item(metadata['trad_as'], metadata['ru_name']) }}&apos;s total number of full-time equivalent employees?",
+                            "description": "<p> Please convert part-time employees hours into those worked by full-time employees. For example, if a part-time employee works 10 hours per week and the full-time working week in your organisation is 37 hours, the part-time employee would equate to 0.27 full-time equivalents (10 divided by 37). </p><p>Please use weekly contracted hours as the basis to calculate FTE.</p>",
```
tgandrews: There is a leading space here in the `<p>` and there is also a description here that does not exist on the slides.

I think it should be here given that it explains the calculation but it isn't on the slides.

tgandrews (changes requested):

insacuri:
@SamGodwin2 The reasoning for wanting to set a default for is to have the same simple experience launching this survey in the go launcher as every other survey.
For this one, you've got to select the survey, then manually type something before getting to launch it.  While it doesn't sound like much, it's not obvious what the date should look like, where a default would act as a hint, and if it's unimportant, then can just be left as is.

SamGodwin2:
@insacuri discussed it with both Sam T and Jon and the consensus seems to be hold off on adding the date_ref metadata to go-launcher until we're sure it is the correct key and it's proved to be launchable in RAS.

tgandrews (approved):

MebinAbraham:
https://github.com/ONSdigital/eq-survey-runner/blob/94a186612214f134e0c02cf49b26e59702c9ab9c/data/en/qpses_0001.json#L44
FYI, there are some non smart quotes in these schemas, they should be addressed. So `organisation's` should be `organisation’s`. 

Shortcut closing single: `ALT + SHIFT + ]`
Shortcut opening single: `ALT + ]`

insacuri (approved):

PR #1982, opened 2019-02-06
Don't query sql for data we know isn't there

jonnyshaw89:
UsedJtiClaim and EQSession don't exist for new sessions so we are unnecessarily querying sql for data we know wont exist.
This is having a performance hit as we are making queries we don't need to.
This also impacts scaling as we can exhaust the QueuePool and connections to the DB causing requests to hang waiting for a DB connection.

### How to review 
I pushed Runner to 2000 concurrent connections and get an error about an exhausted pool
`"sqlalchemy.exc.TimeoutError: QueuePool limit of size 5 overflow 10 reached, connection timed out, timeout 30 (Background on this error at: http://sqlalche.me/e/3o7r)"`

### Checklist


bitdivision (approved):

bitdivision:
Only thing to add is that this should be noted in the readme?

samiwel (approved):

PR #1987, opened 2019-02-08
Introduce Google Could Storage submitter

MebinAbraham:
Card: https://trello.com/c/hT1VML3Y/2752-census-submissions-to-use-google-cloud-storage
Some additions/modification to Davids’s initial work.
Submits completed questionnaires to GCS for downstream.

### How to review 
1. Create a GCP project.
2. Create a storage bucket. https://console.cloud.google.com/storage/

**or set up a project using https://github.com/ONSdigital/census-eq-terraform**
then your bucket id would be: `census-eq-<your-env>-survey-runner-submission`

Set the following ENV variables.
```
EQ_SUBMITTER = 'gcs'
EQ_GCS_SUBMITTER_PROJECT_ID = 'your-project-id'
EQ_GCS_SUBMITTER_BUCKET_ID = 'your-bucket-id'
```

https://github.com/ONSdigital/eq-survey-runner/blob/46990cf653cae2179691e24474df91013f9d7119/app/settings.py#L48-L51

#### Ensure:
- Submissions are successfully stored in your bucket.
- [Object/Blob](https://googleapis.github.io/google-cloud-python/latest/storage/blobs.html) name inside the bucket is same as submitted `case_id`
- [Object/Blob](https://googleapis.github.io/google-cloud-python/latest/storage/blobs.html) metadata is created using the `tx_id`. You can view it in the GCP console:
```
1. Open the Cloud Storage browser in the Google Cloud Platform Console.
OPEN THE CLOUD STORAGE BROWSER

2. In the list of buckets, click on the name of the bucket that contains the desired object, and navigate to the object.
Certain pieces of object metadata, such as the object's size and storage class, are displayed along with the object's name.

3. Click the more options button (More actions icon.) associated with the object.

4. Click Edit metadata.

The overlay window that appears shows the current values for the object's editable metadata.
```

***or you can use the `gsutil ls -L` command, replacing [VALUES_IN_BRACKETS] with the appropriate values:***
``` sh
gsutil ls -L  gs://[BUCKET_NAME]/[OBJECT_NAME]
```


tests/app/submitter/test_submitter.py
```diff
@@ -135,10 +135,29 @@ def test_when_message_sent_with_none_tx_id_then_tx_id_is_not_sent_in_header(self
...
+    def test_send_message(self):
+        with patch('app.submitter.submitter.storage.Client') as client:
+            submitter = GCSSubmitter(project_id='test_project', bucket_name='test_bucket')
+            published = submitter.send_message(message={'test_data'}, case_id='12345')
+
+            bucket = client.return_value.get_bucket.return_value
+            blob = bucket.blob.return_value
+
+            blob_name = bucket.blob.call_args[0][0]
+            assert blob_name == '12345'
```
iwootten: Minor nitpick but can we use assertEqual instead of assert? The value of the comparison object doesn't get shown in errors if we use assert.

iwootten (commented):
This looks good to me - I've been able to see the response get delivered to gcs.

iwootten (dismissed):

app/setup.py
```diff
@@ -288,6 +278,25 @@ def setup_datastore(application):
     application.eq['storage'] = DatastoreStorage(client)
 
 
+def setup_submitter(application):
+    if application.config['EQ_SUBMITTER'] == 'gcs':
+        application.eq['submitter'] = GCSSubmitter(
+            project_id=application.config['EQ_GCS_SUBMITTER_PROJECT_ID'],
+            bucket_name=application.config['EQ_GCS_SUBMITTER_BUCKET_ID'],
+        )
+    elif application.config['EQ_SUBMITTER'] == 'rabbitmq':
```
jonnyshaw89: Since we have removed `get_env_or_fail` from the `settings.py` we should add a check here to kill the app if required values are missing.
Currently the app fails to start if required values are missing causing deployment to fail and app to not be deployed.
With this current setup, the deployment will succeed with an app that will not be able to submit.

app/submitter/submitter.py
```diff
@@ -7,16 +8,33 @@
...
+
+    def __init__(self, project_id, bucket_name):
+        client = storage.Client(project_id)
+        self.bucket = client.get_bucket(bucket_name)
+
+    def send_message(self, message, **kwargs):
         logger.info('sending message')
-        logger.info('message payload', message=message, queue=queue)
+
+        blob = self.bucket.blob(kwargs['case_id'])
```
jonnyshaw89: I think we should consider putting some metadata on the object so we can identify what it is without having to decrypt

app/submitter/submitter.py
```diff
@@ -66,15 +83,17 @@ def send_message(self, message, queue, tx_id):
         try:
             connection = self._connect()
             channel = connection.channel()
-            channel.queue_declare(queue=queue, durable=True)
+
+            channel.queue_declare(queue=self.queue, durable=True)
             properties = BasicProperties(headers={},
                                          delivery_mode=2)
 
+            tx_id = kwargs['tx_id']
```
jonnyshaw89: if this is a `kwarg` do we need to check its existence?

jonnyshaw89: or use `kwargs.get('tx_id')`

scripts/dev_settings.sh
```diff
@@ -16,10 +16,6 @@ if [ -z "$EQ_ENABLE_SECURE_SESSION_COOKIE" ]; then
     export EQ_ENABLE_SECURE_SESSION_COOKIE=False
 fi
 
-if [ -z "$EQ_RABBITMQ_ENABLED" ]; then
-  export EQ_RABBITMQ_ENABLED=False
-fi
-
 if [ -z "$EQ_RABBITMQ_HOST" ]; then
```
jonnyshaw89, line 23: The convention to date has been to default to "production" settings and then override in dev.
With this setup if we don't set `EQ_SUBMITTER` then nothing will be submitted and we wont get any errors

jonnyshaw89 (dismissed):

app/submitter/submitter.py
MebinAbraham: Anything specific you had in mind?

app/submitter/submitter.py
MebinAbraham: Can do, only didn't because it's passed in using `.get()`

tests/app/submitter/test_submitter.py
MebinAbraham: Could do, but I think we previously said that we were trying to move towards using asserts where possible. Can't quite remember. @bitdivision do you remember?

app/submitter/submitter.py
jonnyshaw89: tx_id, which questionnaire maybe?

scripts/dev_settings.sh
MebinAbraham: Addressed.

app/setup.py
MebinAbraham: Have addressed this but we do have cross dependencies, so if `EQ_SUBMITTER` is 'gcs' but then the PROJECT ID or BUCKET_ID might not be set. Ideally, we should enforce groups of dependent env variables. When speaking to Mark, he said we can address that in a follow up card.

MebinAbraham: Added code to check the dependant env variable in this case.

app/submitter/submitter.py
MebinAbraham: Addressed, wasn't sure about key names. For now the format is:
```
{
  "metadata": {
    "tx_id": "744c0afc-6153-4047-b51b-28926990feca"
  },
  "submission": "encrypted_message"
}
```

tests/app/submitter/test_submitter.py
bitdivision: @iwootten Indeed, one of the main advantages of pytest is that you don't need to use those methods anymore: https://docs.pytest.org/en/latest/assert.html

Not using the assert methods makes the code more readable and avoids nastiness around inheriting from testcase etc.

MebinAbraham:
@jonnyshaw89 As discussed, have added metadata directly to the Object. Updated the PR description also.

app/submitter/submitter.py
```diff
@@ -7,16 +8,41 @@
 logger = get_logger()
 
 
-class LogSubmitter():  # pylint: disable=no-self-use
+class LogSubmitter:
+    @staticmethod
+    def send_message(message, **kwargs):
```
ajmaddaford: I think `case_id` and `tx_id` should be named args rather than kwargs. They are always passed to `send_message`.

ajmaddaford (changes requested):

app/submitter/submitter.py
MebinAbraham: Addressed.

ajmaddaford (approved):

app/settings.py
```diff
@@ -45,11 +45,16 @@ def get_env_or_fail(key):
 EQ_SECRETS_FILE = os.getenv('EQ_SECRETS_FILE', 'secrets.yml')
 EQ_KEYS_FILE = os.getenv('EQ_KEYS_FILE', 'keys.yml')
 
-EQ_RABBITMQ_HOST = get_env_or_fail('EQ_RABBITMQ_HOST')
-EQ_RABBITMQ_HOST_SECONDARY = get_env_or_fail('EQ_RABBITMQ_HOST_SECONDARY')
+EQ_SUBMITTER = os.getenv('EQ_SUBMITTER')
```
bitdivision: EQ_SUBMITTER_BACKEND perhaps?

bitdivision: or EQ_SUBMISSION_BACKEND

app/setup.py
```diff
@@ -288,6 +278,43 @@ def setup_datastore(application):
     application.eq['storage'] = DatastoreStorage(client)
 
 
+def setup_submitter(application):
```
bitdivision, line 281: We should update the readme to enumerate the options on EQ_SUBMITTER I think

app/submitter/submitter.py
```diff
@@ -52,11 +77,12 @@ def _disconnect(connection):
         except AMQPError as e:
             logger.error('unable to close connection', exc_info=e, category='rabbitmq')
 
-    def send_message(self, message, queue, tx_id):
+    def send_message(self, message, case_id=None, tx_id=None):
         """
         Sends a message to rabbit mq and returns a true or false depending on if it was successful
         :param message: The message to send to the rabbit mq queue
-        :param queue: the name of the queue
+        :param case_id: ID used to identify q single instance of a survey collection for a respondent
```
bitdivision:
```suggestion
        :param case_id: ID used to identify a single instance of a survey collection for a respondent
```

bitdivision (approved):
Nothing major. 

I've tested in GCP and it's working as expected.

bitdivision:
Should we be using tx_id rather than case_id for the object name now that we can put case_id in the metadata?

@ajmaddaford @pricem14pc

ajmaddaford:
@bitdivision we need to have a conversation with the RM team about how they will receipt, for now I'd say we shouldn't block this as we can easily swap the case_id and tx_id any time.

MebinAbraham:
@bitdivision @jonnyshaw89 @ajmaddaford 
Have renamed the env variables and as per Jon's suggestion, both case_id and tx_id have been added to the metadata.

bitdivision (approved):

app/setup.py
MebinAbraham: I can update once Jon's PR is in or he can sneak it in 😁

PR #1989, opened 2019-02-08
Use a multistage dockerfile to build

bitdivision:
This was prompted by simplifying the concourse pipeline, but I think it's a valuable change. https://github.com/ONSdigital/census-eq-concourse/pull/1#discussion_r255055493

I've removed some of the functional test requirements but the first stage is essentially the same as the one in eq-deploy which runner is currently being built from: https://github.com/ONSdigital/eq-deploy/blob/master/build_containers/survey-runner-build.dockerfile

I had to remove the .git directory from .dockerignore because it's needed for the application version used by the build script

### How to review 
You can test in docker-compose by running

```
docker build -t test-runner .
```

Change your docker-compose file to use `test-runner` and run docker-compose up.

### Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,21 +1,49 @@
-FROM python:3.4
+FROM ubuntu:18.04 as builder
 
-RUN pip install pipenv==8.2.7 \
-  && pip install awscli==1.11.174
-RUN apt update && apt install -y libsnappy-dev
+ARG APP_VERSION=unknown
+
+RUN apt-get update \
+    && apt-get install -y curl git locales make build-essential
+
+RUN locale-gen en_US.UTF-8
+
+ENV LANG=en_US.UTF-8
+RUN curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - \
+    && echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list \
+    && curl -sL https://deb.nodesource.com/setup_8.x | bash -
+
+RUN apt-get update \
+    && apt-get install -y yarn nodejs
+
+WORKDIR /usr/src/app
+
+COPY . /usr/src/app
+RUN yarn compile
+RUN ./scripts/translate_schemas.sh
+RUN printf $APP_VERSION > .application-version
+
+# We don't want node_modules to be copied to the runtime image
+RUN rm -rf node_modules
+###############################################################################
+# Second Stage
+###############################################################################
+
+FROM python:3.4-slim-stretch
+
+EXPOSE 5000
+
+RUN apt update && apt install -y libsnappy-dev build-essential
 
 RUN mkdir -p /usr/src/app
 WORKDIR /usr/src/app
 
-ENV AWS_DEFAULT_REGION eu-west-1
-
-COPY Pipfile Pipfile
 COPY Pipfile.lock Pipfile.lock
+COPY .python-version .python-version
 
-RUN pipenv install --deploy --system
+RUN pip install pipenv==8.2.7
 
-EXPOSE 5000
+RUN pipenv install --system --ignore-pipfile
 
-ENTRYPOINT ["sh", "docker-entrypoint.sh"]
+COPY --from=builder /usr/src/app/ /usr/src/app
 
-COPY . /usr/src/app
+CMD ["sh", "docker-entrypoint.sh"]
--- a/Pipfile
+++ b/Pipfile
@@ -40,6 +40,7 @@ flask-sqlalchemy = "*"
 flask-wtf = "*"
 gevent = {version = "*", platform_python_implementation="=='CPython'"}
 google-cloud-datastore = "*"
+grpcio = "*"
 gunicorn = "*"
 newrelic = "*"
 pika = "*"
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "7a466cf857f1206605ef37c365293416459488068e2623cef6f77d867b88bcfd"
+            "sha256": "abec5e75cb49d188747ea267e80904728d891e648a33ffa16e0a81b3a1068429"
         },
         "pipfile-spec": 6,
         "requires": {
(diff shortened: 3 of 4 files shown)


Dockerfile
```diff
@@ -1,5 +1,49 @@
...
+ENV LANG=en_US.UTF-8
+
+RUN git clone git://github.com/yyuu/pyenv.git .pyenv
+RUN git clone https://github.com/yyuu/pyenv-virtualenv.git ~/.pyenv/plugins/pyenv-virtualenv
+
+ENV HOME  /
+ENV PYENV_ROOT $HOME/.pyenv
+ENV PATH $PYENV_ROOT/shims:$PYENV_ROOT/bin:$PATH
+
+RUN pyenv install 3.4.2
```
jonnyshaw89: We should lean on the `.python_version` file

jonnyshaw89 (changes requested):

Dockerfile
```diff
@@ -1,5 +1,49 @@
...
+
+RUN mkdir -p /usr/src/app
+COPY . /usr/src/app
+RUN ls -al /usr/src/app
+
+WORKDIR /usr/src/app
+RUN ./scripts/build.sh
+
+# The .git folder would be in the .dockerignore if we didn't need it for the build script
+RUN rm -rf .git
```
jonnyshaw89: If we could find another way to set the version, maybe pass it in as a build arg? then we can remove git and just call `yarn compile` instead of `build.sh`

jonnyshaw89 (changes requested):

Dockerfile
```diff
@@ -1,5 +1,49 @@
...
+RUN curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - \
+    && echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list \
+    && curl -sL https://deb.nodesource.com/setup_6.x | bash -
+
+RUN apt-get update \
+    && apt-get install -y yarn nodejs
+
+RUN mkdir -p /usr/src/app
+COPY . /usr/src/app
+RUN ls -al /usr/src/app
```
jonnyshaw89: This can go?

jonnyshaw89 (changes requested):
Should `ENV AWS_DEFAULT_REGION eu-west-1` be in the dockerfile anymore?

jonnyshaw89:
This also creates a `1.26GB` image

Dockerfile
bitdivision: We still need to run the translations, but I'm happy to move those commands into here.

Git will still be required to clone the translations repo, but again, we should look at sorting this out properly.

Passing a build arg would certainly make concourse simpler but things like docker-compose would be difficult. How do we get the commit hash within the compose file?

jonnyshaw89: Do we need the commit hash when building through compose?

Dockerfile
```diff
@@ -1,5 +1,49 @@
...
+ENV PATH $PYENV_ROOT/shims:$PYENV_ROOT/bin:$PATH
+
+RUN pyenv install 3.4.2
+RUN pyenv global 3.4.2
+
+RUN pip install --upgrade pip setuptools pipenv
+
+RUN curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - \
+    && echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list \
+    && curl -sL https://deb.nodesource.com/setup_6.x | bash -
```
jonnyshaw89: We use node 8 don't we?

jonnyshaw89 (changes requested):

Dockerfile
bitdivision: I've spent quite a while playing with this. `pyenv` seems to cause a lot of problems in docker, so I'm going to stick with the python base container for now.

Dockerfile
```diff
@@ -1,5 +1,49 @@
...
+RUN ./scripts/build.sh
+
+# The .git folder would be in the .dockerignore if we didn't need it for the build script
+RUN rm -rf .git
+
+###############################################################################
+# Second Stage
+###############################################################################
+
 FROM python:3.4
```
bitdivision: Was going to change this to a pyenv version to allow us to be really flexible on the .python-version but it caused a lot of issues. Sticking with python for the moment.

bitdivision: > This also creates a `1.26GB` image

That was largely because of node_modules and any untracked files in your local directory. Should be sorted now.

Pipfile
```diff
@@ -40,6 +40,7 @@ flask-sqlalchemy = "*"
 flask-wtf = "*"
 gevent = {version = "*", platform_python_implementation="=='CPython'"}
 google-cloud-datastore = "*"
+grpcio = "*"
```
bitdivision, line 43: I was hitting errors from the google package if this wasn't installed.

Dockerfile
```diff
@@ -1,21 +1,49 @@
...
-COPY Pipfile Pipfile
-COPY Pipfile.lock Pipfile.lock
+# We don't want node_modules to be copied to the runtime image
+RUN rm -rf node_modules
+###############################################################################
+# Second Stage
+###############################################################################
 
-RUN pipenv install --deploy --system
+FROM python:3.4-slim-stretch
```
bitdivision, line 31: Unfortunately we can't use alpine at the moment because of postgres dependencies causing problems. This can be switched once postgres is removed.

bitdivision:
Ready for re-review @MebinAbraham

jonnyshaw89 (approved):
Codacy has a few suggestions

Dockerfile
```diff
@@ -1,21 +1,49 @@
...
+RUN apt update && apt install -y libsnappy-dev build-essential
 
-COPY . /usr/src/app
+RUN mkdir -p /usr/src/app
+WORKDIR /usr/src/app
+
+COPY Pipfile.lock Pipfile.lock
+COPY .python-version .python-version
+
+RUN pip install pipenv==8.2.7
```
MebinAbraham, line 43: About time we changed this?
```suggestion
RUN pip install pipenv==2018.11.26
```

MebinAbraham (approved):

Dockerfile
bitdivision: Nope

PR #1992, opened 2019-02-13
Remove survey test schemas and associated integration tests

jonnyshaw89:
Remove real surveys that are being used in tests. Refactor tests to use test schemas to maintain coverage

### How to review 
All real schemas and starwars should be removed.
Where tests have been removed, other scenarios should be present

### Checklist


tests/integration/rsi/test_rsi_submission_data.py
```diff
@@ -1,178 +0,0 @@
-from tests.integration.integration_test_case import IntegrationTestCase
-
-
-class TestRsiSubmissionData(IntegrationTestCase):
-
-    def test_submission_data_1_0102(self):
```
iwootten, line 6: Do we need a replacement test to cover submission data? Looks like it was only tested as part of rsi/mci.

app/submitter/convert_payload_0_0_1.py
```diff
@@ -80,8 +80,6 @@ def _get_checkbox_answer_data(answer_store, answer_schema, value):
 
     if all(q_code is not None for (q_code, _) in qcodes_and_values):
         checkbox_answer_data.update(qcodes_and_values)
-    else:
```
iwootten: How come this has been changed?

tests/integration/rsi/test_rsi_submission_data.py
jonnyshaw89: It should be tested as part of `test_convert_payload_0_0_1.py`

app/submitter/convert_payload_0_0_1.py
jonnyshaw89: I have added this back with a test

bitdivision (approved):
:+1:

jonnyshaw89:
@iwootten can you re-review

iwootten (approved):

PR #1998, opened 2019-02-21
Use tx_id for submission object name

MebinAbraham:
To ensure potential submission for the same case doesn't result in overwrites in GCS we should use tx_id instead of case_id.
[Trello](https://trello.com/c/t24PDe92/2811-change-gcs-submission-object-key-from-caseid-to-txid)
### How to review 
Set up app using 'gcs' as a submission backend.
Ensure:
- The object name is made from the submission tx_id and not the case_id

### DIFF
--- a/app/submitter/submitter.py
+++ b/app/submitter/submitter.py
@@ -10,7 +10,7 @@ logger = get_logger()
 
 class LogSubmitter:
     @staticmethod
-    def send_message(message, case_id=None, tx_id=None):
+    def send_message(message, case_id, tx_id):
         logger.info('sending message')
         logger.info('message payload', message=message, case_id=case_id, tx_id=tx_id)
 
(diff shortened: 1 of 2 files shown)


tests/app/submitter/test_submitter.py
```diff
@@ -181,15 +181,15 @@ def test_send_message():
...
 
     @staticmethod
-    def test_send_message_adds_tx_id_metadata():
+    def test_send_message_adds_metadata():
         # Given
         with patch('app.submitter.submitter.storage.Client') as client:
             # When
             submitter = GCSSubmitter(project_id='test_project', bucket_name='test_bucket')
-            submitter.send_message(message={'test_data'}, case_id='123', tx_id='456')
+            submitter.send_message(message={'test_data'}, tx_id='123', case_id='456')
```
bitdivision, line 165: You've swapped the values here, but not in the assert.

MebinAbraham: Sorted

bitdivision (approved):

app/submitter/submitter.py
```diff
@@ -23,18 +23,18 @@ def __init__(self, project_id, bucket_name):
         client = storage.Client(project_id)
         self.bucket = client.get_bucket(bucket_name)
 
-    def send_message(self, message, case_id=None, tx_id=None):
+    def send_message(self, message, tx_id=None, case_id=None):
```
jonnyshaw89: Are these ever `None`?

jonnyshaw89: Both are defined as required
https://ons-schema-definitions.readthedocs.io/en/latest/respondent_management_to_electronic_questionnaire.html#required-fields

jcox-dev: RM would be unable to receipt with a message containing no `case_id` in the meta, so would appreciate this as required.

MebinAbraham: I've removed code to check this as the app already make it mandatory hence would fail early if the values are not there.

MebinAbraham:
@jonnyshaw89 Addressed

jonnyshaw89 (approved):

bitdivision (approved):

PR #1999, opened 2019-02-21
Propose new URL structure

ajmaddaford:
The current URL structure we use in runner is too long and overloaded. The URL should be as clean and readable as possible. This pull request proposes a new simpler URL structure.

### DIFF
--- a/doc/architecture/decisions/0005-simplify-urls-and-extend-to-support-repeating-effectively.md
+++ b/doc/architecture/decisions/0005-simplify-urls-and-extend-to-support-repeating-effectively.md
@@ -1,0 +1,74 @@
+# 5. Simplify URLs and extend to support repeating effectively
+
+## Context
+
+The current URL structure used in runner is too long and overloaded. The URL should be as clean and readable as possible.
+
+The current URL structure of runner is:
+```
+/questionnaire/<eq_id>/<survey_id>/<collection_exercise_id>/<group_id>/<group_instance>/<block_id>
+```
+
+Which generates URLs that look like:
+```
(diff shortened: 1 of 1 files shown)


doc/architecture/decisions/0005-simplify-urls-and-extend-to-support-repeating-effectively.md
```diff
@@ -0,0 +1,74 @@
...
+```
+/<list_identifier>/<repeat_identifier>/<block_id>
+/householders/<repeat_identifier>/proxy
+/householders/<repeat_identifier>/sex
+/householders/<repeat_identifier>/date-of-birth
+```
+
+- `list_identifier` is a schema defined name for a list that can be created
+- `repeat_identifier` identifies an item in that list
+- List ids cannot clash with block ids
```
jonnyshaw89, line 67: How will we handle this in the code? would we have to check for a list and if we cant find one then assume its a block?

ajmaddaford: Yes, when we parse a schema we will be able to identify all the list ids. Then, as we do for checking block ids exist, we can check a list id exists.

jonnyshaw89 (approved):

MebinAbraham:
@ajmaddaford What about detection of multiple surveys?. We currently use the collection exercise id in the URL to compare against the one in the session data?

MebinAbraham (approved):

bitdivision (approved):
Looks good.

ajmaddaford:
@MebinAbraham 
>What about detection of multiple surveys?. We currently use the collection exercise id in the URL to compare against the one in the session data?

We'll need to look at what the requirement is for this and if needed how we can do it without the url params, potentially storing something in the cookie.

PR #2002, opened 2019-02-22
Upgrade Python to 3.7

MebinAbraham:
Upgrade runner's Python version to 3.7. 🙏 

### How to review 
Ensure:
- App behaviour is unchanged.


Dockerfile
```diff
@@ -17,15 +17,14 @@ WORKDIR /usr/src/app
...
-RUN ./scripts/translate_schemas.sh
 
 # We don't want node_modules to be copied to the runtime image
 RUN rm -rf node_modules
 ###############################################################################
 # Second Stage
 ###############################################################################
 
-FROM python:3.4-slim-stretch
+FROM python:3.7.2-slim-stretch
```
jonnyshaw89: How about `python:3.7-slim-stretch` so we get patches for free?

app/forms/questionnaire_form.py
```diff
@@ -254,7 +254,8 @@ def get_data(self, answer_id):
         return attr.raw_data[0] if attr.raw_data else ''
 
 
-def get_answer_fields(question, data, error_messages, schema, answer_store, metadata, group_instance, group_instance_id):
+def get_answer_fields(question, data, error_messages, schema, answer_store, metadata,  # pylint: disable=too-many-locals
```
jonnyshaw89: this `pylint` should probably go on the next line as it breaks the parameter list

jonnyshaw89 (changes requested):

app/forms/questionnaire_form.py
MebinAbraham: Doesn't work on the next line, moved it to top.

Dockerfile
```diff
@@ -17,15 +17,14 @@ WORKDIR /usr/src/app
 
 COPY . /usr/src/app
 RUN yarn compile
-RUN ./scripts/translate_schemas.sh
```
MebinAbraham, line 20: Has been removed until the translation repo is ready.

Dockerfile
```diff
@@ -36,12 +35,13 @@ WORKDIR /usr/src/app
 
 ENV GUNICORN_WORKERS 3
 
+COPY Pipfile Pipfile
 COPY Pipfile.lock Pipfile.lock
 COPY .python-version .python-version
```
bitdivision: Can get rid of this?

MebinAbraham:
I will try to sort the issue with Cypress as a separate PR as it's not related to this.

Dockerfile
MebinAbraham: Done

bitdivision (approved):

jonnyshaw89 (approved):

PR #2007, opened 2019-02-25
Fix Cypress In Travis

MebinAbraham:
So for some odd reason, Cypress cannot seem to connect to Docker reliably hence tests are failing due to `Socket Hang Up`.
```
We attempted to make an http request to this URL but the request failed without a response.
We received this error at the network level:
  > Error: socket hang up
Common situations why this would fail:
  - you don't have internet access
  - you forgot to run / boot your web server
  - your web server isn't accessible
  - you have weird network configuration settings on your computer
```
I was able to access the app fine using cURL and Selenium tests run fine against them also. I've tried multiple different configurations and used `wait-on` module which is recommended by Cypress to wait for the endpoint to be accessible but with no luck.

For now, the immediate solution is that we run the app locally in Travis and test against that.
Cypress tests are run against the app in Travis and the Selenium tests use the Docker images to ensure our docker images are always tested.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,20 +1,11 @@
 dist: xenial
 sudo: true
-addons:
-  apt:
-    sources:
-      - google-chrome
-    packages:
-      - google-chrome-stable
-      - libsnappy-dev
-      - libgconf-2-4
 language: node_js
 node_js: 6
 env:
   global:
     - EQ_RUN_LOCAL_LINT=True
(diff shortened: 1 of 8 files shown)


MebinAbraham:
Closing until we plan how we will deploy breaking changes etc. Possibly run the census test in a staging env and make it only deployable if it passes etc.

PR #2010, opened 2019-02-28
Implement new piping

iwootten:

Added support for placeholders as defined here: https://github.com/ONSdigital/eq-schema-validator/blob/v3/doc/decisions/0001-define-piping-in-a-structured-way.md. 

Includes the following:

- PlaceHolder Parser: Extracts placeholder detail from a survey
- Placeholder Transforms:  Defines which transforms can be used on placeholder source data
- Placeholder Renderer: Replaces keys which contain placeholders with the final interpreted text
- A number of test schemas which test the features added

Supports both new an old ways of piping, along with behaviour to support a list as a source so it can easily integrated at a later date.

Depends on https://github.com/ONSdigital/eq-schema-validator/pull/115

### How to review 

Check the unit tests included cover all scenarios. Walk through the test schemas and determine if any others need to be added.

### DIFF
--- a/app/questionnaire/placeholder_parser.py
+++ b/app/questionnaire/placeholder_parser.py
@@ -1,0 +1,61 @@
+from app.data_model.answer_store import AnswerStore
+from app.questionnaire.placeholder_transforms import PlaceholderTransforms
+
+
+class PlaceholderParser:
+    """
+    Parses placeholder statements from a schema dict and returns a map of their
+    final values
+    """
+    def __init__(self, answer_store=None, metadata=None):
+        self._answer_store = answer_store or AnswerStore()
+        self._metadata = metadata
+        self._transformer = PlaceholderTransforms()
+
+    def _lookup_answer(self, answer_id):
+        found_answers = self._answer_store.filter(answer_ids=[answer_id])
+
+        if found_answers.count() > 0:
+            return found_answers.values()[0]
+
+    def parse(self, placeholder_list):
+        placeholder_map = {}
+
+        for placeholder in placeholder_list:
+            placeholder_id = placeholder['placeholder']
+
+            if 'value' in placeholder:
+                source_id = placeholder['value']['identifier']
+
+                if placeholder['value']['source'] == 'answers':
+                    placeholder_map[placeholder_id] = self._lookup_answer(source_id)
+                elif placeholder['value']['source'] == 'metadata':
+                    placeholder_map[placeholder_id] = self._metadata[source_id]
+            elif 'transforms' in placeholder:
+                placeholder_map[placeholder_id] = self.parse_transforms(placeholder['transforms'])
+
+        return placeholder_map
+
+    def parse_transforms(self, transform_list):
+        transformed_value = None
+
+        for transform in transform_list:
+            transform_args = {}
+            for arg_key, arg_value in transform['arguments'].items():
+                if 'value' in arg_value:
+                    transform_args[arg_key] = arg_value['value']
+                elif 'source' not in arg_value:
+                    transform_args[arg_key] = arg_value
+                elif arg_value['source'] == 'answers':
+                    if isinstance(arg_value['identifier'], list):
+                        transform_args[arg_key] = [self._lookup_answer(identifier) for identifier in arg_value['identifier']]
+                    else:
+                        transform_args[arg_key] = self._lookup_answer(arg_value['identifier'])
+                elif arg_value['source'] == 'metadata':
+                    transform_args[arg_key] = self._metadata[arg_value['identifier']]
+                elif arg_value['source'] == 'previous_transform':
+                    transform_args[arg_key] = transformed_value
+
+            transformed_value = getattr(self._transformer, transform['transform'])(**transform_args)
+
+        return transformed_value
--- a/app/questionnaire/placeholder_transforms.py
+++ b/app/questionnaire/placeholder_transforms.py
@@ -1,0 +1,61 @@
+from datetime import datetime
+from dateutil.relativedelta import relativedelta
+
+from babel.numbers import format_currency
+from babel.dates import format_datetime
+from babel import numbers
+from app.settings import DEFAULT_LOCALE
+
+
+class PlaceholderTransforms:
+    """
+    A class to group the transforms that can be used within placeholders
+    """
+    locale = DEFAULT_LOCALE
+    input_date_format = '%Y-%m-%d'
+
+    def format_currency(self, number=None, currency='GBP'):
+        return format_currency(number, currency, locale=self.locale)
+
+    def format_date(self, date_to_format, date_format):
+        date_to_format = datetime.strptime(date_to_format, self.input_date_format)
+        return format_datetime(date_to_format, date_format, locale=self.locale)
+
+    @staticmethod
+    def format_list(list_to_format):
+        formatted_list = '<ul>'
+        for item in list_to_format:
+            formatted_list += '<li>{}</li>'.format(item)
+        formatted_list += '</ul>'
+
+        return formatted_list
+
+    @staticmethod
+    def concatenate_list(list_to_concatenate, delimiter):
+        return delimiter.join(list_to_concatenate)
+
+    @staticmethod
+    def format_possessive(string_to_format):
+        if string_to_format:
+            lowered_string = string_to_format.lower()
+
+            if lowered_string.endswith("'s") or lowered_string.endswith('’s'):
+                return string_to_format[:-2] + '’s'
+
+            if lowered_string[-1:] == 's':
+                return string_to_format + '’'
+
+            return string_to_format + '’s'
+
+    @staticmethod
+    def format_number(number):
+        if number or number == 0:
+            return numbers.format_decimal(number, locale=PlaceholderTransforms.locale)
+
+        return ''
+
+    def calculate_years_difference(self, first_date, second_date):
+        first_date = datetime.now() if first_date == 'now' else datetime.strptime(first_date, self.input_date_format)
+        second_date = datetime.now() if second_date == 'now' else datetime.strptime(second_date, self.input_date_format)
+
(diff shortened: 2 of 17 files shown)


app/questionnaire/placeholder_transforms.py
```diff
@@ -0,0 +1,62 @@
...
+
+from babel.numbers import format_currency
+from babel.dates import format_datetime
+
+
+class PlaceholderTransforms:
+    """
+    A class to group the transforms that can be used within placeholders
+    """
+    locale = 'en_GB'
```
MebinAbraham: Shouldn't we using `flask_babel.get_locale()`?. We also have `DEFAULT_LOCALE` in settings.py

iwootten: The intention here was that we don't have to hang everything from the application context that flask/flask_babel creates, where it's not really used for much. I've updated this to use default locale for now.

app/questionnaire/placeholder_parser.py
```diff
@@ -0,0 +1,65 @@
...
+    def __init__(self, answer_store=None, metadata=None, lists=None):
+        self._lists = lists
+        self._answer_store = AnswerStore({}) if answer_store is None else answer_store
+        self._metadata = metadata
+        self._transformer = PlaceholderTransforms()
+
+    def _lookup_answer(self, answer_id):
+        answer_index = self._answer_store.find(Answer(answer_id, ''))
+        answer = self._answer_store.answer_map[answer_id][answer_index]
+        return answer['value']
```
ajmaddaford: Using the answer_map is exposing functionality internal to the answer store. There is a filter method on the answer store that will return an answer store instance with all answers that match e.g. `answer_store.filter(answer_ids=[source_answer_id])`. The downside to this is that it will return multiple answers, but that should be dealt with here. The answer store search should be matching by both answer id and group instance, so that functionality like piping a persons name into their repeating section works.

app/questionnaire/placeholder_parser.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def parse(self, placeholder_json):
+        placeholder_list = placeholder_json if isinstance(placeholder_json, list) else [placeholder_json]
```
ajmaddaford: Is this just so the tests can just pass a single placeholder json rather than a list? If so, I think that should be done in the tests rather than here.

app/questionnaire/placeholder_parser.py
```diff
@@ -0,0 +1,65 @@
...
+            elif 'transforms' in placeholder:
+                placeholder_map[placeholder_id] = self.parse_transforms(placeholder['transforms'])
+
+        return placeholder_map
+
+    def parse_transforms(self, transform_list):
+        transformed_value = None
+
+        for transform in transform_list:
+            kwargs = {}
```
ajmaddaford: I think the code would be easier to read if this was called something like `transform_args`.

app/questionnaire/placeholder_parser.py
```diff
@@ -0,0 +1,65 @@
...
+        self._answer_store = AnswerStore({}) if answer_store is None else answer_store
+        self._metadata = metadata
+        self._transformer = PlaceholderTransforms()
+
+    def _lookup_answer(self, answer_id):
+        answer_index = self._answer_store.find(Answer(answer_id, ''))
+        answer = self._answer_store.answer_map[answer_id][answer_index]
+        return answer['value']
+
+    def parse(self, placeholder_json):
```
bitdivision: Can we avoid references to `json` in variable names please? To me that suggests that I need to run `json.loads` on them.

app/questionnaire/placeholder_parser.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+class PlaceholderParser:
+    """
+    Parses placeholder statements from a schema dict and returns a map of their
+    final values
+    """
+    def __init__(self, answer_store=None, metadata=None, lists=None):
+        self._lists = lists
+        self._answer_store = AnswerStore({}) if answer_store is None else answer_store
```
jonnyshaw89: Do we need to pass anything into `AnswerStore` can it not just be `AnswerStore()` as it gets set to a `defaultdict(list)` in the `__init__` if none is provided which differs from what is being set here.

app/questionnaire/placeholder_transforms.py
```diff
@@ -0,0 +1,63 @@
...
+        formatted_list += '</ul>'
+
+        return formatted_list
+
+    @staticmethod
+    def concatenate_list(list_to_concatenate, delimiter):
+        return delimiter.join(list_to_concatenate)
+
+    @staticmethod
+    def format_possessive(string_to_format):
```
jonnyshaw89, line 38: This should handle `'s` too shouldn't it?

jonnyshaw89 (changes requested):

app/views/questionnaire.py
```diff
@@ -124,7 +125,12 @@ def get_block(routing_path, schema, metadata, answer_store, eq_id, form_type, co
         return _redirect_to_location(collection_id, eq_id, form_type, next_location)
 
     block = _get_block_json(current_location, schema, answer_store, metadata)
-    context = _get_context(routing_path, block, current_location, schema)
+
+    placeholder_pointers = find_pointers_containing(block, 'placeholders')
+    placeholder_renderer = PlaceholderRenderer(block, answer_store=answer_store, metadata=metadata)
+    replaced_block = placeholder_renderer.render(placeholder_pointers)
```
ajmaddaford: I don't think we should be exposing the pointers functionality here, it seems like something that should be internal to the Renderer. I'd expect to see something like `replaced_block =  PlaceholderRenderer(block, answer_store=answer_store, metadata=metadata).render() ` here.

app/views/questionnaire.py
```diff
@@ -149,9 +155,13 @@ def post_block(routing_path, schema, metadata, collection_metadata, answer_store
 
     block = _get_block_json(current_location, schema, answer_store, metadata)
 
+    placeholder_pointers = find_pointers_containing(block, 'placeholders')
+    placeholder_renderer = PlaceholderRenderer(block, answer_store=answer_store, metadata=metadata)
+    replaced_block = placeholder_renderer.render(placeholder_pointers)
```
ajmaddaford: Same as the comment above.

ajmaddaford (changes requested):

app/questionnaire/placeholder_parser.py
jonnyshaw89: should `.answer_map` become `._answer_map` then?

app/questionnaire/placeholder_parser.py
iwootten: OK

app/questionnaire/placeholder_parser.py
iwootten: Sure thing, updated.

app/questionnaire/placeholder_parser.py
iwootten: OK

MebinAbraham:
@iwootten Was this missed from the implemented transforms on purpose?
https://github.com/ONSdigital/eq-schema-validator/blob/v3/schemas/string_interpolation/transforms/format_number.json

app/questionnaire/placeholder_parser.py
iwootten: Updated this to use filter. As discussed offline, leaving the group_instance for now given it's being completely reworked anyway.

iwootten:
> @iwootten Was this missed from the implemented transforms on purpose?
> https://github.com/ONSdigital/eq-schema-validator/blob/v3/schemas/string_interpolation/transforms/format_number.json

Nope, just missed as I was mainly using the examples from the docs to go by.

app/questionnaire/placeholder_transforms.py
iwootten: Yeah, seem to be missing tests for the transforms class too.

app/questionnaire/placeholder_parser.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+class PlaceholderParser:
+    """
+    Parses placeholder statements from a schema dict and returns a map of their
+    final values
+    """
+    def __init__(self, answer_store=None, metadata=None, lists=None):
+        self._lists = lists
+        self._answer_store = AnswerStore() if answer_store is None else answer_store
```
MebinAbraham: Maybe?
```python
self._answer_store = answer_store or AnswerStore()
```

iwootten: Maybe what?

bitdivision: There was a code quote in Mebin's comment

iwootten: Yeah, I saw it. I was being slightly facetious.

tests/app/questionnaire/test_placeholder_parser.py
```diff
@@ -0,0 +1,464 @@
...
+                }
+            ]
+        })
+
+        placeholders = parser.parse(placeholder_list)
+
+        assert placeholders['full_name_list'] == ['Joe Bloggs', 'Jane Bloggs']
+
+    @staticmethod
+    def test_list_transform_placeholder_no_concatenate():
```
ajmaddaford: I think we should remove the list functionality from this pull request. The structure of the list used in the tests doesn't match the proposed structure, and that could also change when we implement lists.

ajmaddaford (changes requested):

jonnyshaw89 (approved):

tests/app/questionnaire/test_placeholder_parser.py
iwootten: Sure, I've removed this behaviour now.

ajmaddaford (approved):
Nice 😃

PR #2011, opened 2019-03-01
Propose making lists a thing

ajmaddaford:
The current approach to collecting a list of things e.g. people that live in a household, and asking questions for each thing in that list, is more complex than it needs to be. This pull request proposes making lists a first class construct in runner.


doc/architecture/decisions/0006-make-lists-a-thing.md
```diff
@@ -0,0 +1,106 @@
...
+            ]
+        }
+    ]
+}
+```
+
+- Removal of `group_instance`, `group_instance_id` and `answer_instance` from answers.
+- Introduction of a `repeat_identifier` that identifies the list item that an answer is related to.
+- The order of the items in the list definiton will determine display order.
+- The repeat identifier should be a short randomised string (it only has to be unique within the questionnaire).
```
bitdivision: I worked this one out just for fun. `62^6` combinations so using `K^2 / 2N`, for 100 repeat identifiers per survey, the chance of collision = `8.8 * 10^-8`.

I guess we should still check it in any case

doc/architecture/decisions/0006-make-lists-a-thing.md
```diff
@@ -0,0 +1,106 @@
...
+    "question": {},
+    "add_question": {},
+    "edit_question": {},
+    "remove_question": {}
+}
+```
+
+- The routing path is kept simple as we can ignore the journeys taken adding/editing/removing when evaluating the path.
+- Each question is a standard question type.
+- Each question can have variants. This is important as the add question can change depending on whether the list is empty or not.
```
jonnyshaw89: How would the above example support variants? 
https://github.com/ONSdigital/eq-schema-validator/pull/113/files#diff-87a1851b9e8803d68cb8195b642ee9a6R29
If it was like the linked example would we need `add_variants`?

ajmaddaford: Yes, something like `add_variants` would fit with how question variants work.

ajmaddaford: Given that we could have question variants as well here, we could look at changing that from `variants` to `question_variants`, so that if it was used in combination with add question variants it would be clear, something like:

```json
{
  "question_variants": {},
  "add_question_variants": [],
  "edit_question": {}
}
```

Another option is we could consider making the top level a block rather than a question, something like:

```json
{
  "question": {},
  "add_block": {
     "id": "add-something",
	 "question_variants": []
  },
  "edit_block": {
     "id": "edit-something",
	 "question": {}
  }
}
```

ajmaddaford: Thinking about this a bit more, I also don't think we can enforce the mutual exclusivity of `add_question` / `add_question_variants` etc at the block level in JSON schema.

bitdivision: I think I prefer the block approach if only because it keeps everything more consistent and avoids another set of keys. 

I made a comment about the `question_variants` change on the validator PR, but I think it's probably worth doing.

bitdivision (approved):
:+1:

jonnyshaw89 (approved):

jonnyshaw89:
@ajmaddaford rebase?

PR #2013, opened 2019-03-05
Update census household schema

bitdivision:
We need to update the census household schema in runner to include the latest changes.

Since the schema is not being maintained, there is no need for the cypress functional test anymore. Given that we are very likely to reuse cypress in future, I have left the infrastructure around it in there for the moment.

The schema that is being copied is available here: https://github.com/ONSdigital/eq-census-schemas/blob/census_household_hannah_2/census_household.json as per the trello card: https://trello.com/c/iqEpCo7q/2813-update-census-household-in-runner

### How to review 
Make sure the census household schema works

### Checklist


scripts/run_tests_functional.sh
```diff
@@ -34,7 +34,3 @@ yarn test_functional
 
 display_result $? 5 "Front end functional tests"
 
-echo "Running census functional tests"
-yarn test_census
```
MebinAbraham, line 38: May as well remove it from the README also.

bitdivision: I've just changed it to test_cypress for the moment, since the functionality is still there.

MebinAbraham (approved):
Matches the other schema so I guess 👍

iwootten:
I get a fair amount of what looks like formatting differences between what's linked to in the trello card after it's been formatted (maybe worth linking that gist here too) and the schema in the PR here.

bitdivision:
@iwootten Yep, I did have to reformat it as we lint the schemas. I probably should have done that as two commits sorry.

bitdivision:
If you wanted to double check it, you could copy the schema, run `yarn format` and diff the two.

iwootten:
@bitdivision that's after me having run through the gulp format - should there be differences after that? I wonder if we could automate linting of the schemas in the census household repo in the future.

bitdivision:
@iwootten Ah, right. I'll have a look. There was only one material change for a bad ID I believe.

I'd rather just get rid of that repo. I don't think CI linting will work well unless we can run `yarn format` and auto commit the result (since it's committed through the web interface)

bitdivision:
@iwootten You're right. There are some formatting differences. I can only assume that was down to pycharm and the formatter doesn't fully manage it. 

I've updated the schema and avoided using pycharm.

iwootten (approved):

PR #2018, opened 2019-03-15
Welsh Census Test Updates

iwootten:

In preparation for the welsh test for census household, we need to make a update the census theme to allow switching and make a version available in AWS.

### How to review 

We've added census_household_cy so we don't have to update v2 of runner to use pot translations right now. It's been necessary to update the theme header for census to allow language switching.

Walk through the census_household_cy and identify if anything looks broken.

### DIFF
--- a/app/themes/census/templates/partials/header.html
+++ b/app/themes/census/templates/partials/header.html
@@ -1,12 +1,46 @@
-<header class="header header--census u-mb-s">
-  <div class="u-pt-s u-pb-s" role="banner">
-    <div class="container">
-      <div class="grid grid--gutterless">
-        <div class="grid__col col-6@s">
-          <div class="logo"><img src="{{ url_for('static', filename='img/theme/census/census-logo.svg') }}" alt="2017 Census" /></div>
+{% if language_code  %}
+  {% set languages = language_urls([['cy', 'Cymraeg'], ['en', 'English']], language_code) %}
+{% endif %}
+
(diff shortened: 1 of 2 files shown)


jonnyshaw89 (approved):

bitdivision (approved):

app/themes/census/templates/partials/header.html
```diff
@@ -1,12 +1,46 @@
...
+                          {% for language in languages %}
+                              <li class="nav__item" role="menuitem"><a data-qa="switch-language-{{language[0]}}" class="nav__link" href="?language_code={{language[0]}}">{{language[1]}}</a></li>
+                          {% endfor %}
+                        </ul>
+                    </nav>
+                </div>
+            </div>
+        </div>
+    </div>
+    <div class="header__main" style="background:#6e2585;">
```
hamishtaplin: is this inline style intentional?

hamishtaplin: is this inline style intentional?

hamishtaplin (changes requested):

app/themes/census/templates/partials/header.html
iwootten: Nope. We were looking for a way to quickly apply the census colour to the background but decided it doesn't matter for this test.

hamishtaplin (approved):

PR #2019, opened 2019-03-15
Remove support for old version of piping

jonnyshaw89:
Remove rendering of schema for piping

### How to review 
Check that rendering of schemas does not still occur


data/en/test_conditional_dates.json
```diff
@@ -62,7 +62,7 @@
                         "type": "Number"
                     }],
                     "id": "total-retail-turnover-question",
-                    "title": "For the period {{ format_conditional_date (answers['date-start-from'], metadata['ref_p_start_date'])}} to {{ format_conditional_date (answers['date-end-to'], metadata['ref_p_end_date'])}}, what was the value of the business\u2019s total retail turnover?",
+                    "title": "For the period {{ format_conditional_date (answe_\ntests/app/templating/test_schema_context.py:27:rs['date-start-from'], metadata['ref_p_start_date'])}} to {{ format_conditional_date (answers['date-end-to'], metadata['ref_p_end_date'])}}, what was the value of the business\u2019s total retail turnover?",
```
bitdivision: Accidental paste or something?

jonnyshaw89: looks like it

data/en/test_multiple_piping.json
```diff
@@ -104,14 +104,57 @@
...
+                                                "identifier": ["building", "address-line-1", "address-line-2", "address-line-3", "town-city", "county", "postcode", "country"]
+                                            }
+                                        }
+                                    }, {
+                                        "transform": "concatenate_list",
+                                        "arguments": {
+                                            "list_to_concatenate": {
+                                                "source": "previous_transform"
+                                            },
+                                            "delimiter": ", "
```
ajmaddaford, line 146: We could make `concatenate_list` remove empty items before concatenation. I can't think of a scenario when we'd want empty items in a concatenated list.

app/views/questionnaire.py
```diff
@@ -436,32 +429,19 @@ def _redirect_to_location(location):
...
 
 
-def _get_context(full_routing_path, block, current_location, schema, form=None):
+def _get_context(block, current_location, schema, form=None):
     metadata = get_metadata(current_user)
-    collection_metadata = get_collection_metadata(current_user)
 
     answer_store = get_answer_store(current_user)
-    schema_context = _get_schema_context(full_routing_path, metadata, collection_metadata, answer_store, schema)
-    rendered_block = renderer.render(block, **schema_context)
```
ajmaddaford, line 431: 🎉

data/en/test_confirmation_question.json
```diff
@@ -44,7 +44,21 @@
...
+                                "transform": "first_non_empty_item",
+                                "arguments": {
+                                    "items": {
+                                        "source": "metadata",
+                                        "identifier": ["trad_as", "ru_name"]
+                                    }
+                                }
+                            }]
+                        }]
+                    },
```
ajmaddaford, line 122: We should consider whether we should have piping in test surveys that don't need it, we should always be creating the minimal schema to test a feature.

data/en/test_schema_context.json
```diff
@@ -48,21 +48,93 @@
                     "id": "preview",
                     "title": "",
                     "content": [{
-                        "title": "period_str: {{ metadata['period_str'] }}"
+                        "description": {
```
ajmaddaford: I don't think the name of this file makes sense, this looks like it's just testing piping of metadata.

scripts/test_schemas.sh
```diff
@@ -6,8 +6,8 @@ if [ "$1" == "--local" ] || [ "$2" == "--local" ]; then
 fi
 
 if [ "$run_docker" == true ]; then
-    docker pull onsdigital/eq-schema-validator:use-non-global-named-args
-    validator="$(docker run -d -p 5001:5000 onsdigital/eq-schema-validator:use-non-global-named-args)"
+#    docker pull onsdigital/eq-schema-validator:use-non-global-named-args
```
ajmaddaford: Just commenting on the commented line as a reminder it needs to be uncommented 😃

app/jinja_filters.py
```diff
@@ -7,7 +7,7 @@
 import flask_babel
```
ajmaddaford, line 6: There is some duplication between this file and `placeholder_transforms.py`. This file should just contain filters that are called from templates. Longer term we will remove most of the filters and apply the transforms before the values get rendered in a template.

ajmaddaford (changes requested):

data/en/test_multiple_piping.json
jonnyshaw89: I did start doing that originally but then decided that all of the other functions were doing 1 thing and didn't want to get into the state we had before with the jinja filters

app/jinja_filters.py
jonnyshaw89: Yeh I forgot to go through and remove the old ones

app/questionnaire/placeholder_transforms.py
```diff
@@ -30,6 +30,10 @@ def format_list(list_to_format):
 
         return formatted_list
 
+    @staticmethod
+    def remove_empty(list_to_filter):
+        return [item for item in list_to_filter if item]
```
MebinAbraham: In the `first_non_empty_item` we decided that `falsy` values like `False` and `0` are not empty, hence we should be consistent if we are calling this remove_**empty** and ensure we don't remove `False`/`0`.

MebinAbraham (dismissed):

tests/functional/spec/conditional_dates.js
```diff
@@ -3,7 +3,7 @@ const DatesPage = require('../generated_pages/conditional_dates/date-block.page'
 const DatesConfirmationPage = require('../generated_pages/conditional_dates/date-value-test.page');
 const SummaryPage = require('../generated_pages/conditional_dates/summary.page');
 
-describe('Piped Dates', function () {
+describe.skip('Piped Dates', function () {
```
MebinAbraham: 👀

jonnyshaw89: Is a test we want to keep but won’t currently work

app/questionnaire/placeholder_transforms.py
bitdivision: Also makes it consistent with first_non_empty_item

tests/functional/generate_pages.py
```diff
@@ -396,7 +396,10 @@ def process_block(block, dir_out, schema_data, spec_file, relative_require='..')
 
 def process_schema(in_schema, out_dir, spec_file, require_path='..'):
 
-    data = json.loads(open(in_schema).read())
+    try:
+        data = json.loads(open(in_schema).read())
+    except Exception as ex:
+        logger.error('error reading %s', in_schema)
```
bitdivision, line 413: :+1:

bitdivision (commented):
Looks like the merge on this shouldn't be too horrendous. I've looked through everything and all looks good apart from existing comments. 

Will re-review once variants are merged in.

app/jinja_filters.py
```diff
@@ -170,57 +123,6 @@ def format_datetime(context, value):
...
-    day_of_week_offset = relativedelta.relativedelta(days=(-parsed_datetime.weekday() - (7 - day_of_week_index)))
-    day_of_last_week = parsed_datetime + day_of_week_offset
-
-    offset_output = day_of_last_week + offset
-
-    return datetime.strftime(offset_output, '%Y-%m-%d')
-
-
 @blueprint.app_template_filter()
 def calculate_years_difference(from_str, to_str):
```
bitdivision: Why do we still need this?

app/jinja_filters.py
```diff
@@ -246,105 +148,6 @@ def format_date_range(context, start_date, end_date=None):
...
-    if name:
-        last_char = name[-1:]
-        if last_char.lower() == 's':
-            return name + '’'
-
-        return name + '’s'
-
-
 @blueprint.app_template_filter()
 def answer_is_type(answer, target_answer_type):
```
bitdivision, line 137: This can go?

bitdivision (changes requested):

app/jinja_filters.py
jonnyshaw89: This is still used in some of the summary templates

data/en/test_multiple_piping.json
ajmaddaford: I understand that, but concatenate_list should never include empty items. I'd assume that it would filter them out without needing to remember to include an extra transform every time.

jonnyshaw89: Already fixed

app/jinja_filters.py
```diff
@@ -1,13 +1,12 @@
 # coding: utf-8
```
ajmaddaford, line 1: `min_value` and `max_value` don't seem to be used anywhere now, except for test_dates.json that still has old piping in it.

app/templating/metadata_context.py
```diff
@@ -41,3 +40,17 @@ def build_metadata_context_for_survey_completed(session_data):
...
 
     return metadata_context
+
+
+def build_schema_metadata(metadata, schema):
+    schema_metadata = schema.json['metadata']
+    parsed = {metadata_field['name']: json_and_html_safe(metadata[metadata_field['name']])
+              for metadata_field in schema_metadata if metadata_field['name'] in metadata}
+
+    return parsed
```
ajmaddaford: We may not need to do this now as the survey specific metadata was probably only used for piping.

app/views/questionnaire.py
```diff
@@ -137,27 +134,23 @@ def post_block(routing_path, schema, metadata, collection_metadata, answer_store
 
     block = schema.get_block(current_location.block_id)
 
-    placeholder_renderer = PlaceholderRenderer(block, answer_store=answer_store, metadata=metadata)
-    replaced_block = placeholder_renderer.render()
+    transformed_block = transform_variants(block, schema, metadata, answer_store)
 
-    transformed_block = transform_variants(replaced_block, schema, metadata, answer_store)
+    placeholder_renderer = PlaceholderRenderer(answer_store=answer_store, metadata=metadata)
+    replaced_block = placeholder_renderer.render(transformed_block)
```
ajmaddaford: I understand that we don't want to confuse this with the old way of rendering a block, but given that we're calling a render method with a block, `rendered_block` would be clearer than `replaced_block`.

app/views/questionnaire.py
```diff
@@ -414,33 +403,37 @@ def _save_sign_out(routing_path, current_location, current_question, form, schem
...
+def _evaluate_skip_conditions(block_json, schema, answer_store, metadata):
+    for question in schema.get_questions_for_block(block_json):
+        if 'skip_conditions' in question:
+            skip_question = evaluate_skip_conditions(question['skip_conditions'], schema, metadata, answer_store)
+            question['skipped'] = skip_question
+            for answer in question['answers']:
+                if answer['mandatory'] and skip_question:
+                    answer['mandatory'] = False
+
+    return block_json
```
ajmaddaford: Now we have variants, skipping questions should be removed.

ajmaddaford (changes requested):

app/views/questionnaire.py
jonnyshaw89: Wasn't even called by anything

data/en/test_confirmation_question.json
```diff
@@ -27,109 +27,152 @@
...
+                                "text": "Of the <em>{number_of_employees_total}</em> total employees employed, how many male and female employees worked the following hours?",
+                                "placeholders": [{
+                                    "placeholder": "number_of_employees_total",
+                                    "value": {
+                                        "source": "answers",
+                                        "identifier": "number-of-employees-total"
+                                    }
+                                }]
+                            },
+
```
bitdivision: 🤷‍♂️

scripts/test_schemas.sh
```diff
@@ -6,8 +6,8 @@ if [ "$1" == "--local" ] || [ "$2" == "--local" ]; then
 fi
 
 if [ "$run_docker" == true ]; then
-    docker pull onsdigital/eq-schema-validator:enforce-consistent-variant-ids
-    validator="$(docker run -d -p 5001:5000 onsdigital/eq-schema-validator:enforce-consistent-variant-ids)"
+#    docker pull onsdigital/eq-schema-validator:v3
```
bitdivision: Change before merge

bitdivision (approved):

data/en/test_confirmation_question.json
jonnyshaw89: 🤷‍♂️

ajmaddaford (approved):

app/questionnaire/placeholder_transforms.py
```diff
@@ -31,8 +31,27 @@ def format_list(list_to_format):
         return formatted_list
 
     @staticmethod
-    def concatenate_list(list_to_concatenate, delimiter):
-        return delimiter.join(list_to_concatenate)
+    def remove_empty(list_to_filter):
```
MebinAbraham: I think we should make this method's name a bit more descriptive, maybe `remove_all_empty` or `remove_all_empty_items`?.

MebinAbraham (changes requested):
Have tested the app and seems to be working as expected.
Just need to add tests for the newly added transform.

app/views/questionnaire.py
```diff
@@ -109,12 +105,12 @@ def get_block(routing_path, schema, metadata, answer_store, block_id):
 
     block = schema.get_block(current_location.block_id)
 
-    placeholder_renderer = PlaceholderRenderer(block, answer_store=answer_store, metadata=metadata)
-    replaced_block = placeholder_renderer.render()
+    transformed_block = transform_variants(block, schema, metadata, answer_store)
 
-    transformed_block = transform_variants(replaced_block, schema, metadata, answer_store)
+    placeholder_renderer = PlaceholderRenderer(answer_store=answer_store, metadata=metadata)
+    replaced_block = placeholder_renderer.render(transformed_block)
```
MebinAbraham: As per Andrew's comment, `rendered_block`?

MebinAbraham (approved):
👍

PR #2021, opened 2019-03-19
Add variant functionality and remove titles / multiple questions

bitdivision:
The test_variant_questions.json should show the main functionality of
question_variants, but needs to be expanded to cover more scenarios.

This PR deletes quite a lot of code which was there to handle multiple
questions and titles. Currently all of the unit tests / remaining functional tests 
are passing and the schemas should all pass validation.

This PR is going to be a pain to merge with the PR for removal of piping, due to the modification
of most of the schemas. I don't think there's any way around that.

@bameyrick This PR does change a few of the templates, they're not major changes, but wanted to let you know in case it affects your work.

### How to review 
- Check the variant test schemas make sense.
- Ensure I haven't been too over-zealous in deleting code / tests
- Probably a good idea to check I haven't made any visual changes to any of the templating. I think this is one thing the functional tests wouldn't be good at catching.


app/questionnaire/questionnaire_schema.py
```diff
@@ -96,9 +104,17 @@ def get_summary_and_confirmation_blocks(self):
         ]
 
     @staticmethod
-    def get_questions_for_block(block_json):
-        for question_json in block_json.get('questions', []):
-            yield question_json
+    def get_questions_for_block(block):
```
ajmaddaford: The naming between `get_all_questions_for_block` and `get_questions_for_block` is inconsistent, they both return all questions. It would be clearer if they were named something like `get_all_questions_for_block_id` and `get_all_questions_in_block`.

app/questionnaire/questionnaire_schema.py
```diff
@@ -69,11 +69,19 @@ def get_group_by_block_id(self, block_id):
...
+
     def get_answers_by_id_for_block(self, block_id):
         block = self.get_block(block_id)
         if block:
             answers_by_id = {}
-            for question in block.get('questions', []):
+
+            questions = self.get_all_questions_for_block(block_id)
+
+            for question in questions:
```
ajmaddaford: I don't think the answer objects this returns are ever used, we only seem to use the answer ids 🤔

app/questionnaire/questionnaire_schema.py
```diff
@@ -136,10 +152,20 @@ def _parse_schema(self):
...
 
+    def _get_questions_by_id(self):
+        questions_by_id = OrderedDict()
+
+        for block in self._blocks_by_id.values():
+            for question in self.get_questions_for_block(block):
+                questions_by_id[question['id']] = question
+                questions_by_id[question['id']]['parent_id'] = block['id']
+
+        return questions_by_id
```
ajmaddaford, line 172: I don't think this is very clear. As all question variants will have the same id, this will only ever capture the last variant. I don't think this would be clear to someone that calls the `questions` or `get_question` methods.

app/templates/partials/questions/relationship.html
```diff
@@ -1,9 +1,6 @@
 <div class="question question--repeatinganswer" id="{{question.id}}">
 
   <h1 class="question__title u-fs-l">
-    {%- if question_number -%}
-      <span class="question__number">{{question_number}}{{current_location.group_instance|format_number_to_alphabetic_letter}}. </span>
```
ajmaddaford, line 5: 👍

app/templating/utils.py
```diff
@@ -1,21 +1,11 @@
...
-            return value
-    return titles[-1]['value']
+def get_question_to_display(block, metadata, answer_store, schema):
+    if block.get('question'):
+        return block['question']
+
+    for question_variant in block.get('question_variants', []):
+        when_rules = question_variant.get('when', [])
+        if evaluate_when_rules(when_rules, schema, metadata, answer_store):
+            return question_variant['question']
```
ajmaddaford: This looks very similar to `choose_question_to_display` in `schema_helpers.py`..

app/templating/view_context.py
```diff
@@ -90,10 +91,14 @@ def build_view_context_for_non_question(rendered_block):
 
 
 def build_view_context_for_question(metadata, schema, answer_store, rendered_block, form):  # noqa: C901, E501  pylint: disable=too-complex,line-too-long,too-many-locals,too-many-branches
+    question = get_question_to_display(rendered_block, metadata, answer_store, schema)
+    rendered_block.pop('question_variants', None)
+    rendered_block.pop('question', None)
+
+    rendered_block['question'] = question
```
ajmaddaford: We should really be determining which variant will be displayed before evaluating piping (rendering the block), but that may be easier once the rendering is replaced by the placeholder changes.

app/views/questionnaire.py
```diff
@@ -447,9 +434,9 @@ def _get_context(full_routing_path, block, current_location, schema, form=None):
     return build_view_context(block['type'], metadata, schema, answer_store, schema_context, rendered_block, current_location, form=form)
 
 
-def _get_block_json(current_location, schema, answer_store, metadata):
+def _get_block_json(current_location, schema):
     block_json = schema.get_block(current_location.block_id)
-    return _evaluate_skip_conditions(block_json, schema, answer_store, metadata)
+    return block_json
```
ajmaddaford: It's probably worth removing this method now.

app/views/questionnaire.py
```diff
@@ -468,8 +455,8 @@ def get_page_title_for_location(schema, current_location, metadata, answer_store
         group = schema.get_group(schema.get_group_by_block_id(block['id'])['id'])
         page_title = '{group_title} - {survey_title}'.format(group_title=group['title'], survey_title=schema.json['title'])
     elif block['type'] == 'Question':
-        first_question = next(schema.get_questions_for_block(block))
-        question_title = get_question_title(first_question, answer_store, schema, metadata)
+        question = choose_question_to_display(block, schema, metadata, answer_store)
+        question_title = question.get('title')
```
ajmaddaford: Similar to the comment about choosing a variant after rendering the block, we should really be determining which question will be displayed in one place. Having said that, this might be easier once the new version of piping is implemented.

ajmaddaford (changes requested):

app/questionnaire/questionnaire_schema.py
bitdivision: Good spot, looks like it's only really used in tests. Will sort that out.

app/templates/partials/questions/relationship.html
bitdivision: I just noticed the filename on here. Think this whole template can go...

app/templating/view_context.py
bitdivision: Agreed, I'll have a think about the best way to do this.

app/questionnaire/questionnaire_schema.py
jonnyshaw89: if anything this should just return the ids as they should be consistent across the variants

app/templating/view_context.py
```diff
@@ -21,7 +21,8 @@ def build_view_context(block_type, metadata, schema, answer_store, schema_contex
         return build_view_context_for_question(metadata, schema, answer_store, rendered_block, form)
 
     if block_type in ('Introduction', 'Interstitial', 'Confirmation'):
-        return build_view_context_for_non_question(rendered_block)
+        ctx = build_view_context_for_non_question(rendered_block)
+        return ctx
```
jonnyshaw89: 🤔 ?

package.json
```diff
@@ -138,7 +138,9 @@
     "eslint-plugin-node": "^5.1.1",
     "events": "^1.1.1",
     "ie9-oninput-polyfill": "^1.1.1",
+    "jq": "^1.7.2",
     "lodash": "^4.17.5",
+    "rimraf": "^2.6.3",
```
jonnyshaw89: where are these used?

jonnyshaw89 (changes requested):
The removal of titles and questions (and skips) looks good!

app/questionnaire/questionnaire_schema.py
bitdivision: Consistent ids are not enforced in validator at the moment. Maybe we should?

jonnyshaw89: I would make the code easier if we could trust that assumption

app/templating/view_context.py
```diff
@@ -152,40 +159,46 @@ def _build_calculated_summary_section_list(schema, rendered_block, current_locat
...
-            question['answers'] = answers_to_keep
-            questions.append(question)
+
+    if block_question['id'] in questions_to_keep:
+        answers_to_keep = [answer for answer in block_question['answers'] if answer['id'] in answer_ids_to_keep]
+
+        block_question['answers'] = answers_to_keep
+        questions.append(block_question)
 
     reduced_block['questions'] = questions
```
MebinAbraham: These are all redundant?
`questions = []`
`questions.append(block_question)`
`reduced_block['questions'] = questions`

app/templating/view_context.py
```diff
@@ -152,40 +159,46 @@ def _build_calculated_summary_section_list(schema, rendered_block, current_locat
...
-def _remove_unwanted_questions_answers(block, answers_in_block, answer_ids_to_keep):
+def _remove_unwanted_questions_answers(block, answers_in_block, answer_ids_to_keep, answer_store, metadata, schema):
+    """
+    Evaluates questions in a block and removes any questions not containing a relevant answer
+    """
+    block_question = choose_question_to_display(block, schema, answer_store, metadata)
+
     reduced_block = block.copy()
     questions_to_keep = [
         answer['parent_id'] for answer in answers_in_block.values() if answer['id'] in answer_ids_to_keep
```
MebinAbraham: Can this be more than 1?

MebinAbraham (changes requested):

app/templating/view_context.py
jonnyshaw89: why?

app/templating/view_context.py
bitdivision: What do you mean?

app/templating/view_context.py
bitdivision: The code is creating a subset of questions and setting `reduced_block['questions']` to that set of questions. I don't see what's redundant here?

MebinAbraham: Unless I've misread something, there is only ever 1 question?
`questions.append(block_question)` only runs once?.

`reduced_block['questions'] = questions` -> We no longer use `questions`?
So would it not be `reduced_block['question'] = block_question or {}`?

app/templating/view_context.py
MebinAbraham: Was referring to the fact that there would only be 1 question. But seems fine to be a list of one.

bitdivision:
I think I've addressed all the comments now. I have also added some tests.

The code for calculated summary is less than ideal at the moment. Might be worth doing more work on it.

data/en/test_variants_question.json
```diff
@@ -0,0 +1,482 @@
...
+                                                        "identifier": ["first-name-answer", "last-name-answer"]
+                                                    },
+                                                    "delimiter": " "
+                                                }
+                                            }]
+                                        }]
+                                    },
+                                    "answers": [{
+                                        "id": "age-answer",
+                                        "label": "Your age?",
```
jonnyshaw89: This label is incorrect which is confusing

jonnyshaw89 (changes requested):

jonnyshaw89:
In this example, when I change the currency I don't get asked the subsequent questions again. 
With the context of this question i would expect the answers to be different if the currency changes.
I know this isn't something we want to do, so maybe we can change the context of the question?
Screenshot 2019-03-25 at 09 01 35

app/questionnaire/questionnaire_schema.py
```diff
@@ -96,9 +116,17 @@ def get_summary_and_confirmation_blocks(self):
         ]
 
     @staticmethod
-    def get_questions_for_block(block_json):
-        for question_json in block_json.get('questions', []):
-            yield question_json
+    def get_all_questions_for_block(block):
```
jonnyshaw89, line 111: The naming of this confused me a bit.
From what i can find, this never returns "all" of the questions?
it returns all that are left on the block after the variants have been transformed?

jonnyshaw89: or maybe not, i thin it uses it when getting answers on routing path as well as for generating the form

app/views/questionnaire.py
```diff
@@ -133,27 +135,29 @@ def post_block(routing_path, schema, metadata, collection_metadata, answer_store
         next_location = router.get_next_location()
         return _redirect_to_location(next_location)
 
-    block = _get_block_json(current_location, schema, answer_store, metadata)
+    block = schema.get_block(current_location.block_id)
 
     placeholder_renderer = PlaceholderRenderer(block, answer_store=answer_store, metadata=metadata)
     replaced_block = placeholder_renderer.render()
 
+    new_block = transform_variants(replaced_block, schema, metadata, answer_store)
```
jonnyshaw89: In `get_block` we call this `transformed_block` and here we use `new_block`
Should we be more consistent with this naming?

app/views/questionnaire.py
```diff
@@ -462,35 +443,33 @@ def _get_schema_context(full_routing_path, metadata, collection_metadata, answer
                                 answer_ids_on_path=answer_ids_on_path)
 
 
-def get_page_title_for_location(schema, current_location, metadata, answer_store):
+def get_page_title_for_location(schema, current_location, context):
     block = schema.get_block(current_location.block_id)
     if block['type'] == 'Interstitial':
         group = schema.get_group(schema.get_group_by_block_id(block['id'])['id'])
         page_title = '{group_title} - {survey_title}'.format(group_title=group['title'], survey_title=schema.json['title'])
```
jonnyshaw89, line 450: I know this is slightly out of the context of this PR, but using the `group_title` in the title for an interstitial page seems odd. I think it's the only place we use a groups title where we should probably use the block title.
This also looks odd as the page title does not really have anything to do with the heading on the page

app/questionnaire/questionnaire_schema.py
bitdivision: No, it will retrieve all of the questions on the block, including each of the variants.

app/views/questionnaire.py
bitdivision: I can remove it if there are no objections?

bitdivision:
Any suggestions on a better question for currency @jonnyshaw89 ?

tests/functional/spec/skip_conditions.spec.js
```diff
@@ -1,4 +1,4 @@
-const helpers = require('../helpers');
+/*const helpers = require('../helpers');
```
jonnyshaw89: Intentional?

bitdivision: Ah, I had forgotten about this. Commented out to get all the tests running, I'll sort this out now.

jonnyshaw89:
Have tested this feature and all seems to work well.

app/assets/styles/themes-common/_ui.scss
```diff
@@ -4,4 +4,3 @@
 @import "../partials/objects/group.scss";
 @import "../partials/objects/section.scss";
 @import "../partials/components/notice.scss";
-@import "../partials/components/relationship.scss";
```
ajmaddaford, line 7: 👍

jonnyshaw89 (approved):

app/helpers/form_helper.py
```diff
@@ -48,16 +48,25 @@ def post_form_for_block(schema, block_json, answer_store, metadata, request_form
...
+        # Here Be Dragons: This loop modifies the input in place.
+        for answer in question.get('answers', []):
+            if answer.get('mandatory', True) is True:
+                answer['mandatory'] = False
+
+    if block.get('question'):
+        set_mandatory_to_false(block.get('question'))
+
+    for question_variant in block.get('question_variants', []):
+        set_mandatory_to_false(question_variant['question'])
```
ajmaddaford: I don't think this is needed now that `transform_variants` is called on a block.

app/questionnaire/answer_store_updater.py
```diff
@@ -41,7 +42,7 @@ def _update_questionnaire_store_with_form_data(self, answers):
 
             # If answer is not answered then check for a schema specified default
             if answer_value is None:
-                answer_value = self._schema.get_answer(answer_id).get('default')
+                answer_value = self._schema.get_answer_with_context(answer_id, self._metadata, self._answer_store).get('default')
```
ajmaddaford: Needing to pass metadata here highlights this is the wrong place to be working out the default values. We should consider setting the default values before this method is called, at a point we have access to the relevant answer schema.

app/questionnaire/questionnaire_schema.py
```diff
@@ -69,25 +89,25 @@ def get_group_by_block_id(self, block_id):
         block = self.get_block(block_id)
         return self.get_group(block['parent_id'])
 
-    def get_answers_by_id_for_block(self, block_id):
+    def get_all_questions_for_block_id(self, block_id):
+        block = self.get_block(block_id)
+        return QuestionnaireSchema.get_all_questions_for_block(block)
```
ajmaddaford: I think we can remove this method - `get_all_questions_for_block_id` is only called from `get_answer_ids_for_block`, which already has the block, so can call `get_all_questions_for_block` directly.

app/forms/questionnaire_form.py
```diff
@@ -38,7 +37,7 @@ def validate(self):
         valid_calculated_form = True
         valid_mutually_exclusive_form = True
 
-        for question in self.schema.get_questions_for_block(self.block_json):
+        for question in self.schema.get_all_questions_for_block(self.block_json):
```
ajmaddaford: I think we should consider initialising the form with the question json rather than the block json so we wouldn't have to consider all the possible questions for the block here.

app/forms/questionnaire_form.py
```diff
@@ -232,17 +231,17 @@ def answers_all_valid(self, answer_id_list):
     def map_errors(self):
         ordered_errors = []
 
-        question_json_list = self.schema.get_questions_for_block(self.block_json)
+        question_list = self.schema.get_all_questions_for_block(self.block_json)
```
ajmaddaford: Same as the comment above, if the form was initialised with the question json rather than the block json, we wouldn't need to consider the irrelevant questions here.

ajmaddaford (changes requested):

app/forms/questionnaire_form.py
```diff
@@ -109,7 +108,7 @@ def _get_target_total_and_currency(self, calculation, question):
         if 'value' in calculation:
             return calculation['value'], question.get('currency')
 
-        target_answer = self.schema.get_answer(calculation['answer_id'])
+        target_answer = self.schema.get_answer_with_context(calculation['answer_id'], self.metadata, self.answer_store)
```
ajmaddaford: I think we can remove the need for the `get_answer_with_context` method here. The target answer id is `calculation['answer_id']`, and we can assume the target currency to be the same as the currencies for the inputs on the `Calculated` question. 

We should create a card to tighten up the validation of `Calculated` questions to ensure if a currencies are used they are consistent between the total question and all breakdown answers.

app/submitter/convert_payload_0_0_1.py
```diff
@@ -18,10 +19,9 @@ def convert_answers_to_payload_0_0_1(answer_store, schema, routing_path):
     for location in routing_path:
         answer_ids = schema.get_answer_ids_for_block(location.block_id)
         answers_in_block = answer_store.filter(answer_ids)
-        answer_schema_list = schema.get_answers_by_id_for_block(location.block_id)
 
         for answer in answers_in_block:
-            answer_schema = answer_schema_list[answer['answer_id']]
+            answer_schema = schema.get_answer_with_context(answer['answer_id'], metadata, answer_store)
```
ajmaddaford: I think we should look to remove the `get_answer_with_context` method from questionnaire schema as it's introducing a dependency on metadata and answer_store. The `get_answer_with_context` method uses the answer id to get the block id, then gets the block, and evaluates that to determine which question was displayed. Instead we could get the block to get the block and call `choose_question_to_display` here to determine the question, and from that the answers.

MebinAbraham (approved):
👍

bitdivision:
Sorry @ajmaddaford, forgot that rebasing breaks the commit timeline on github.

The last 4 commits were made after your latest review.

- Address most of the comments
- Address remaining comments
- Fix lint
- Fix bad argument

app/forms/questionnaire_form.py
```diff
@@ -255,30 +251,31 @@ def get_data(self, answer_id):
 
 
 # pylint: disable=too-many-locals
-def get_answer_fields(question, data, error_messages, schema, answer_store, metadata):
+def get_answer_fields(question, data, error_messages, answer_store, metadata):
     answer_fields = {}
-    for answer in question.get('answers', []):
+    if question:
```
ajmaddaford: We should return early if there is no question, it would be clearer and avoid the need for indenting everything another level.

bitdivision:
Just remembered the discussion yesterday about calculated summary looping over all blocks, so have changed that now @ajmaddaford

app/questionnaire/answer_store_updater.py
```diff
@@ -24,7 +25,7 @@ def save_answers(self, form):
         self._questionnaire_store.add_or_update()
 
     def _update_questionnaire_store_with_answer_data(self, answers):
-        survey_answer_ids = self._schema.get_answer_ids_for_block(self._current_location.block_id)
+        survey_answer_ids = self._schema.get_answer_ids_for_question(self._current_question)
```
ajmaddaford: The name `survey_answer_ids` is misleading, `question_answer_ids` or `answer_ids` would be better.

Also, the `_update_questionnaire_store_with_form_data` method is now inconsistent with this one as it calls `get_answer_ids_for_block`.

tests/integration/questionnaire/test_questionnaire_question_variants.py
```diff
@@ -0,0 +1,89 @@
...
+            'proxy-answer': proxy_answer
+        })
+
+        expected_answer = 'What age is <em>Linus Torvalds</em>?' if proxy else 'What is your age?'
+
+        self.assertInBody(expected_answer)
+
+        self.post({'age-answer': '49'})
+
+        expected_answer = '<em>Linus Torvalds</em> is over 16?' if proxy else 'You are over 16?'
```
ajmaddaford: expected_answer -> expected_question?

data/en/test_variants_question.json
```diff
@@ -0,0 +1,482 @@
...
+                        "type": "SectionSummary"
+                    }
+                ]
+            }]
+        },
+        {
+            "id": "currency-section",
+            "groups": [{
+                "id": "currency-group",
+                "title": "Calculated Summary With Variants",
```
ajmaddaford: This uses a Section summary rather than Calculated summary.

data/en/test_variants_question.json
```diff
@@ -0,0 +1,482 @@
...
+        },
+        {
+            "id": "basic-question-variant-section",
+            "groups": [{
+                "id": "basic-question-variant-group",
+                "title": "Variants",
+                "blocks": [{
+                        "type": "Question",
+                        "id": "age-block",
+                        "title": "questions only",
```
ajmaddaford: The block titles should be removed. We don't use them in real questionnaires and they aren't adding anything.

ajmaddaford (changes requested):

ajmaddaford (approved):
🎉

PR #2025, opened 2019-03-21
Remove transaction from datastore

jonnyshaw89:
The transaction around all of the datastore puts was causing issues under load with datastore erroring with contention issues

`google.api_core.exceptions.Conflict: 409 too much contention on these datastore entities. please try again.`

### How to review 
Check that datastore still behaves correctly


app/storage/datastore.py
```diff
@@ -46,15 +46,14 @@ def put(self, model, overwrite=True):
         table_name = current_app.config[config['table_name_key']]
         key = self.client.key(table_name, key_value)
 
-        with self.client.transaction():
```
bitdivision, line 49: Removing this transaction technically introduces a race condition, since the read doesn't provide any isolation without the transaction. https://cloud.google.com/datastore/docs/concepts/transactions

e.g.  
- Container 2 tries to read a key - Doesn't exist
- Container 1 tries to read a key - Doesn't exist
- Container 2 writes to that key
- Container 1 writes to that key

It's going to be a very rare event, but would be possible. Not sure what the solution is if performance with transactions is poor though.

bitdivision (changes requested):

app/storage/datastore.py
jonnyshaw89: I have tested keeping this transaction in place and using Firestore in Datastore mode. But the contention issue is still present

bitdivision: @kierenbeckett @nealedj Any ideas on how to improve performance here?

bitdivision: This seems to be a limitation of datastore: https://cloud.google.com/datastore/docs/concepts/limits

Writes are limited to 1 write per second 😮

jonnyshaw89: So i'm guessing that creating a transaction counts as a write?

kierenbeckett: Can you see if it makes any difference if you only open a transaction if `overwrite=False`

jonnyshaw89: I have just run a test with only opening a transaction if `overwrite=False` and the issue persists

jonnyshaw89: we are also now seeing `BadRequest: 400 The referenced transaction has expired or is no longer valid.`

jonnyshaw89:
Closing this in favour of looking into using redis as a kv store for ephemeral data

PR #2031, opened 2019-04-01
Use Redis as ephemeral storage for JTI

jonnyshaw89:
This PR introduces Redis as a storage backend for JTI claims.
This also disables the unique checking within the datastore implementation as it has its limitations.

### How to review 
Test that you are still able to launch a survey.
Check that the expiry is working and that records are being removed from the Redis instance

Related to https://github.com/ONSdigital/census-eq-terraform/pull/17

### DIFF
--- a/app/storage/datastore.py
+++ b/app/storage/datastore.py
@@ -3,7 +3,6 @@ from google.cloud import datastore
 from structlog import get_logger
 
 from app.data_model import app_models
-from app.storage.errors import ItemAlreadyExistsError
 
 logger = get_logger()
 
@@ -37,6 +36,9 @@ class DatastoreStorage:
         self.client = client
 
     def put(self, model, overwrite=True):
+        if not overwrite:
+            raise NotImplementedError('Unique key checking not supported')
+
         config = TABLE_CONFIG[type(model)]
 
         schema = config['schema'](strict=True)
@@ -46,15 +48,9 @@ class DatastoreStorage:
         table_name = current_app.config[config['table_name_key']]
         key = self.client.key(table_name, key_value)
 
-        with self.client.transaction():
-            if not overwrite:
-                existing = self.client.get(key)
-                if existing:
-                    raise ItemAlreadyExistsError()
-
-            entity = datastore.Entity(key=key, exclude_from_indexes=list(item.keys()))
-            entity.update(item)
-            self.client.put(entity)
+        entity = datastore.Entity(key=key, exclude_from_indexes=list(item.keys()))
+        entity.update(item)
+        self.client.put(entity)
 
     def get_by_key(self, model_type, key_value):
         config = TABLE_CONFIG[model_type]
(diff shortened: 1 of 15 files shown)


iwootten:
So, I've been able to run this successfully with the expiry reduced and it correctly expires as expected. Not sure how I'd check that redis is cleaning up expired tokens though.

jonnyshaw89:
@iwootten in GCP it shows your the key count in the DB. you can see it reducing over time

MebinAbraham (approved):
Tested and seems to be working as expected. Just waiting on the keys to expire.

Not sure if the `configmap` should live in the terraform repo or this repo. I guess it would be harder to manage if it lived in runner.

jonnyshaw89:
@MebinAbraham Yeh i was unsure about where to put that and ended up leaving it in terraform as then it has access to the values

app/storage/datastore.py
```diff
@@ -37,6 +36,9 @@ def __init__(self, client):
         self.client = client
 
     def put(self, model, overwrite=True):
+        if not overwrite:
+            raise Exception('Unique key checking not supported')
```
JamesGardiner: Should a more specific error be raised?

JamesGardiner (changes requested):

app/storage/datastore.py
jonnyshaw89: In what way?

JamesGardiner: I'm not a fan of raising a bare `Exception` when a more specific subclass of it can be raised. Even if it is a custom exception class. In this instance `NotImplementedError` might be more applicable. Or don't give the option to pass `overwrite` at all.

jonnyshaw89: Not passing it would cause more issues as we have 3 storage providers matching same interface

JamesGardiner: Then it sounds like NotImplementedError suits

docker-compose.yml
```diff
@@ -25,10 +30,16 @@ services:
       EQ_USED_JTI_CLAIM_TABLE_NAME: dev-used-jti-claim
       EQ_DATASTORE_EMULATOR_CREDENTIALS: "True"
       DATASTORE_EMULATOR_HOST: datastore:8432
+      GOOGLE_CLOUD_PROJECT: "local"
       EQ_SUBMISSION_BACKEND: "log"
+      EQ_REDIS_HOST: "redis"
+      EQ_REDIS_PORT: "6379"
+    volumes:
+      - ~/.config/:/root/.config
```
bitdivision: This should either be removed or have its scope narrowed.

bitdivision (changes requested):

docker-compose.yml
jonnyshaw89: Fixed

bitdivision: 👍

bitdivision (approved):

JamesGardiner (approved):

MebinAbraham:
https://github.com/ONSdigital/census-eq-terraform/pull/17

PR #2032, opened 2019-04-02
Propose relationships design

ajmaddaford:
This pull request proposes a design for relationships using the new [url structure](0005-simplify-urls-and-extend-to-support-repeating-effectively.md) and [lists](0006-make-named-lists-a-first-class-construct.md).

## How to review
- Is there enough detail, or too much?
- Is there a better approach?
- Is anything missing?

### DIFF
--- a/doc/architecture/decisions/0007-relationships.md
+++ b/doc/architecture/decisions/0007-relationships.md
@@ -1,0 +1,81 @@
+# 7. Relationships
+
+## Context
+
+We need to be able to collect the (one way) relationships between list items. For example, given a list of items A, B and C, we need to be able to capture the relationships between A and B, A and C, and B and C. The previous implementation of relationships relied on group instance, functionality that has now been removed. Now that we have a new [url structure](0005-simplify-urls-and-extend-to-support-repeating-effectively.md) and [lists](0006-make-named-lists-a-first-class-construct.md), we can design how relationships will work.
+
+## Decision
+
+### URLs
+
+The existing url design for lists provides an approach to asking questions related to a given list item:
+
+```
+/<list_identifier>/<list_item_id>/<block_id>
+```
+
+Using a list named "householders" and a block id of "relationships", the url for all relationships for a given list item would be:
+
+```
+/householders/<list_item_id>/relationships
+```
+
+Extending this to support a relationship to something, the url for an individual relation would be:
+
+```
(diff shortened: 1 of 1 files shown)


doc/architecture/decisions/0007-relationships.md
```diff
@@ -0,0 +1,79 @@
...
+  "identifiers": ["first-name", "last-name"],
+}
+```
+
+The second person name is more difficult, as we will need to resolve the answers for a list item outside of the current repeat. We could introduce a `list_item_id` property to the placeholder source definition. This could be a fixed value that can be used to resolve a list item id. In this example runner would need to resolve `related_list_item_id` to the url parameter:
+
+```
+{
+  "source": "answers",
+  "list_item_id": "related_list_item_id"
```
jonnyshaw89: I think we need a more explicit way to say that this is a value from a url parameter?

jonnyshaw89 (changes requested):
Mostly happy with this proposal

doc/architecture/decisions/0007-relationships.md
ajmaddaford: It's not as far as the schema is concerned. Getting the value from a url parameter is how we plan to do it in the runner implementation.

jonnyshaw89: I think if we make the value less like a string value and maybe prefix with to make it look like a pointer?

ajmaddaford: Thinking about this, `list_item_id` will never be an actual list item id, as the list items will only exist at runtime. We could either use a prefix (e.g. @):

```
{
  "source": "answers",
  "list_item_id": "@related_list_item_id"
  "identifiers": ["first-name", "last-name"],
}
```

Or change the name of the property, to something like `list_item_selector`:
```
{
  "source": "answers",
  "list_item_selector": "related_list_item_id"
  "identifiers": ["first-name", "last-name"],
}
```

jonnyshaw89: i think the latter is more appropriate

bitdivision: I agree, selector makes sense immediately, whereas you would need to think about the `@` if you didn't know anything about it.

MebinAbraham (approved):

bitdivision (approved):

jonnyshaw89 (approved):

PR #2036, opened 2019-04-02
Add Opencensus Tracing

iwootten:

In order to better understand where most time is spent within runner add support for opencensus tracing to capture traces to stackdriver. This uses the opencensus flask middleware to initially setup a overall trace and provides a decorator capture_trace which can be added to methods of interest.

### How to review 

Manually create a stackdriver project, add the environment vars:

EQ_STACKDRIVER_PROJECT_ID=$YOUR_PROJECT_ID

Then step through a survey and look through the trace list captured to stackdriver.

Screen Shot 2019-04-03 at 09 29 55


app/decorators/opencensus_decorators.py
```diff
@@ -0,0 +1,12 @@
+from functools import wraps
+from opencensus.trace import execution_context
+
+
+def capture_trace(f):
+    @wraps(f)
+    def decorated_function(*args, **kwargs):
+        tracer = execution_context.get_opencensus_tracer()
+        with tracer.span(name=f.__name__) as span:  # noqa: F841
```
jonnyshaw89:
```suggestion
        with tracer.span(name=f.__name__):
```

app/settings.py
```diff
@@ -48,6 +48,9 @@ def get_env_or_fail(key):
 EQ_SUBMISSION_BACKEND = os.getenv('EQ_SUBMISSION_BACKEND')
 EQ_GCS_SUBMISSION_BUCKET_ID = os.getenv('EQ_GCS_SUBMISSION_BUCKET_ID')
 
+EQ_OPENCENSUS_ENABLED = os.getenv('EQ_OPENCENSUS_ENABLED')
```
jonnyshaw89: Is this used?

jonnyshaw89 (changes requested):

app/settings.py
iwootten: Nope, removed

app/decorators/opencensus_decorators.py
```diff
@@ -0,0 +1,12 @@
+from functools import wraps
+from opencensus.trace import execution_context
+
+
+def capture_trace(f):
+    @wraps(f)
+    def decorated_function(*args, **kwargs):
+        tracer = execution_context.get_opencensus_tracer()
+        with tracer.span(name=f.__name__):
+            return f(*args, **kwargs)
```
jonnyshaw89: this is really petty but can we make `f` `function` or something like that? I have a real aversion to abbreviations in code

jonnyshaw89:
This doesn't seem to work when deployed to GCP. The requests take ages and the logs show...
Screenshot 2019-04-03 at 10 53 03

jonnyshaw89:
I am able to replicate the above issue using docker-compose and setting `EQ_STACKDRIVER_PROJECT_ID`

jonnyshaw89 (changes requested):
`/session` isn't being traced

jonnyshaw89:
You need to add permission for the cluster to push traces.
add `"https://www.googleapis.com/auth/trace.append",` to the `oauth_scopes` in
https://github.com/ONSdigital/census-eq-terraform/blob/master/main.tf#L201

tests/integration/test_app_create.py
```diff
@@ -249,6 +251,18 @@ def test_rabbit_submitter_secondary_host_not_set_raises_exception(self):
...
         assert 'Setting EQ_RABBITMQ_HOST_SECONDARY Missing' in str(ex.exception)
 
+    def test_adds_stackdriver_tracing_to_the_application(self):
+        self._setting_overrides['EQ_STACKDRIVER_PROJECT_ID'] = 'stackdriver-project'
+
+        application = create_app(self._setting_overrides)
+
+        context = application.test_request_context(path='/')
+
+        with context:
```
jonnyshaw89: Why not just `with application.test_request_context(path='/'):

jonnyshaw89:
@iwootten These changes look good but the issue with not being able to run it in the container is a bit of a show stopper at the moment.

k8s/helm/values.yaml
```diff
@@ -10,6 +10,7 @@ image:
   pullPolicy: Always
 
 submissionBucket: ""
+stackdriverProject: ""
```
MebinAbraham, line 13: You should add this to
https://github.com/ONSdigital/eq-survey-runner/blob/v3/k8s/deploy_app.sh
and update the README.

iwootten:
@jonnyshaw89 Has done a fair amount of performance investigation here and it looks like that currently runner can cope with around 475 requests/sec in our default terraform setup. This tracing branch takes that down to around 200 due to the use of a custom gevent worker which doesn't patch threading. Gevent replaces the behaviour of a lot of libraries to make use of libev (In particular threading which opencensus also uses).

Without tracing enabled, but still using the custom tracing worker class in this branch drops performance to 320 requests/sec.

As decided by the team in planning, this work therefore is going to be archived until such a point as we've investigated whether it is possible to improve the performance of runner by looking at making some of it run asynchronously.

PR #2040, opened 2019-04-07
Add the concept of a list collector

bitdivision:
See the design decisions:
- https://github.com/ONSdigital/eq-survey-runner/blob/v3/doc/architecture/decisions/0005-simplify-urls-and-extend-to-support-repeating-effectively.md
- https://github.com/ONSdigital/eq-survey-runner/blob/v3/doc/architecture/decisions/0006-make-named-lists-a-first-class-construct.md

This PR adds the functionality for a ListCollector block. There are some minor changes to the decision document, mostly just to use `list_item_id`.

Styling of the list summary is currently not included, it is just a basic list. The pattern for this has not been finalised and can be added once it has moved into the pattern library.

The answer store has been modified significantly. The underlying data-structure has been changed to make lookups more efficient and simplify the API. The serialisation format is currently still the same. Rather than using hashing to check for a dirty answer_store, this version uses a dirty flag on the store.

Associated eq-schema-validator PR: https://github.com/ONSdigital/eq-schema-validator/pull/126

### How to review 
- Can you break it?
- Is the code clear?
- Have I missed any edge cases?

### Checklist


app/views/questionnaire.py
```diff
@@ -464,6 +480,172 @@ def _render_template(context, current_location, template, previous_url, schema,
...
+    return list_collector_get_handler(schema, metadata, answer_store, block_id, 'edit', list_item_id)
+
+
+@questionnaire_blueprint.route('<block_id>/add', methods=['GET'])
+@login_required
+@with_answer_store
+@with_metadata
+@with_schema
+def get_add_list_item(schema, metadata, answer_store, block_id):
+    return list_collector_get_handler(schema, metadata, answer_store, block_id, 'add', '')
```
bitdivision: The `list_item_id` of `''` here needs to be addressed. Issue being that currently, the None value represents 'do not filter on list_item_id'.

app/data_model/answer_store.py
```diff
@@ -51,13 +56,31 @@ def copy(self):
...
         self._validate(answer)
+        position = self.find(answer)
 
-        answer_to_add = vars(answer).copy()
-        self.answer_map[answer.answer_id] = answer_to_add
+        if position is None:
+            answer_to_add = vars(answer).copy()
+            self.answer_map[answer_to_add['answer_id']].append(answer_to_add)
+        else:
+            self.answer_map[answer.answer_id][position]['value'] = answer.value
```
ajmaddaford: Now we only have one identifier for something that repeats, we should consider using it as a key rather than position in a list as it would be more efficient for creating and looking up answers.

MebinAbraham (commented):
Since the answer ids are the same, once you add a person, your previous answer state is saved for each iteration.
Screenshot 2019-04-09 at 09 18 56

app/data_model/answer.py
```diff
@@ -4,12 +4,13 @@
...
 
 class Answer:
-    def __init__(self, answer_id, value):
+    def __init__(self, answer_id, value, list_item_id=None):
         if answer_id is None or value is None:
             raise ValueError("Both 'answer_id' and 'value' must be set for Answer")
 
         self.answer_id = answer_id
         self.value = value
+        self.list_item_id = list_item_id
```
MebinAbraham: When we had `group_instance_id` and it was giving a default of `None` to answers that were not 'repeats'.
I thought for the lists we wanted to only show list item id if it is a `ListCollector` and not default it to None?.

bitdivision: What do you mean by 'show' list item id? I don't see the problem with having an optional `list_item_id` on the Answer, I don't think there's any other obvious way to do it?

Were you thinking of `group_instance` defaulting to 0 in the location?

MebinAbraham: Show as in, assign a `list_item_id` key with a value only if it needs one. So I was referring to this:
``` json
{
      "answer_id": "repeating-anyone-else",
      "answer_instance": 0,
      "group_instance": 0,
      "group_instance_id": null,
      "value": "Yes"
},
```
Where `group_instance_id` was always assigning a value even if it didn't need one, I get that it's None. I just thought we said we wouldn't want to assign the key value if it wasn't needed. I don't really see a problem either way, it was just I thought that what we said.

bitdivision: So, we can probably change the serialisation to ensure that list_item_id is not serialised if it is None. This would mean a custom json serialiser on here though I think?

ajmaddaford:
I think we should nest the `add`, `change` and `remove` urls under the list name rather than the block id of the list collector. This will simplify the routing and logic when dealing with locations. It's also more consistent with the `<list_name>/<list_item_id>/<block_id>` urls for questions relating to a list item, and the urls for relationships.

So:
```
<block_id>
<block_id>/add
<block_id>/<list_item_id>/change
<block_id>/<list_item_id>/remove
```
becomes:
```
<block_id>
<list_name>/<add_block_id>
<list_name>/<list_item_id>/<change_block_id>
<list_name>/<list_item_id>/<remove_block_id>
```

Where `add`, `change` and `remove` is defined in the block ids of the `add_block`, `change_block` and `remove_block`..

app/data_model/answer_store.py
bitdivision: I've implemented the naive approach for this here: https://github.com/ONSdigital/eq-survey-runner/tree/lists-new-answerstore

Discussed with @ajmaddaford, but not really happy with the implementation having to use two different types under each answer_id.

Also discussed some alternative structures for the answer_store, but the best approach seems to be to use a tuple of (`answer_id`, `list_item_id`) as a key for the answer store, and store a single answer under each key.

For non-repeating answers, this means we can query using (<answer_id>, None).

For repeating answers, if you have both answer_id and list_item_id, querying is efficient: (<answer_id>, <list_item_id>).

Where this solution is not so efficient is if you are querying a repeating answer with only a `list_item_id` or `answer_id` since you would need to iterate through all the keys. I can't see any use cases where we would need to do that.

Currently, this will only be implemented for the in-memory structure for the answer_store. The serialisation format should be decided based on the most efficient serialisation / deserialisation routine.

This will mean implementing a custom serialise / deserialise method on the answer_store.

bitdivision: Also discussed removing the filter method from the answer_store and replacing with methods to retrieve one answer and multiple answers.

bitdivision:
@ajmaddaford Not sure how the new URL scheme can work with multiple list collectors?

If we have two list collectors populating the 'people' list and the user navigates to:

`/people/<add_block_id>`  

I think the only way we could resolve this would be if we enforce that the sub_blocks have different ids for each list collector. Which doesn't seem ideal.

app/questionnaire/answer_store_updater.py
```diff
@@ -21,6 +21,7 @@ def save_answers(self, form):
             self._questionnaire_store.completed_blocks.append(self._current_location)
 
         self._questionnaire_store.add_or_update()
+        self._answer_store.dirty = True
```
ajmaddaford: This will always mark the answer store as dirty, even if the user didn't change an answer e.g. when going back through previously answered questions.

app/questionnaire/location.py
```diff
@@ -36,17 +46,22 @@ def __repr__(self):
...
+    def to_dict(self) -> dict:
+        attributes = vars(self)
+        return {k: v for k, v in attributes.items() if v is not None}
 
-    def url(self):
+    def url(self) -> str:
         """
         Return the survey runner url that this location represents
 
+        The structure
```
ajmaddaford: 🤔

app/questionnaire/questionnaire_schema.py
```diff
@@ -152,14 +156,48 @@ def is_confirmation_group(group):
...
+        return self.get_block(block['parent_id'])
+
+    def is_block_list_collector_child(self, block_id):
+        block = self.get_block(block_id)
+        if not block:
+            return False
+
+        parent = self.get_block(block['parent_id'])
+
+        return parent and parent['type'] == 'ListCollector'
```
ajmaddaford: Could this be replaced with `if block['type'] in LIST_COLLECTOR_CHILDREN`?

app/views/questionnaire.py
```diff
@@ -183,14 +183,15 @@ def post_block_handler(routing_path, schema, metadata, collection_metadata, list
                 answer_store_updater.save_answers(form)
                 add_url = url_for('questionnaire.get_add_list_item', list_name=rendered_block['populates_list'], add_block_id=rendered_block['add_block']['id'])
                 return redirect(add_url)
+            else:
+                answer_store_updater.save_answers(form)
```
MebinAbraham: Move this to above the `if` since it's in both `if` and `else`?

app/data_model/answer_store.py
```diff
@@ -42,97 +65,108 @@ def _validate(answer):
...
-        Removes answer *in place* from the answer store.
+        if self.answer_map.get((answer_id, list_item_id)):
+            del self.answer_map[(answer_id, list_item_id)]
+
+    def remove_all_answers_for_list_item_id(self, list_item_id: str):
+        """Remove all answers associated with a particular list_item_id
+        This method iterates through the entire list of answers.
 
-        :param answer: The answer to remove
+        *Not efficient.*
```
ajmaddaford, line 121: Would it be more efficient to iterate the list first to identify the items to delete, then loop over those and delete in place?

ajmaddaford (changes requested):
This is looking good. As discussed we should look at:
- Separating the validation of a route and the list actions in the view into their own methods.
- Use Answer objects internally in the answer store.
- Bring the list item creation and deletion into the answer store updater so they are kept as close as possible to the creation/removal of associated answers.
- Completed blocks shouldn't be modified when adding/changing/removing list items.

bitdivision:
I think I've addressed everything you've raised now @ajmaddaford .

I'm going to sort out the functional tests and then hopefully this should be ready to go.

app/data_model/answer.py
```diff
@@ -1,24 +1,41 @@
...
-        self.answer_id = answer_id
-        self.value = value
-
-    def matches(self, answer):
+    answer_id: str
+    value: str
+    list_item_id: str = field(default=None)
+
+    @classmethod
+    def from_dict(cls, answer_dict: Dict) -> Answer:
```
jonnyshaw89, line 18: oooo type annotations 🤔

app/data_model/answer.py
```diff
@@ -1,24 +1,41 @@
...
+        if not self.list_item_id:
+            del output['list_item_id']
+        return output
+
+    def to_dict(self) -> Dict:
+        return asdict(self)
+
+    def matches(self, answer: Answer):
         """
         Check to see if two answers match.
```
jonnyshaw89: Should probably clarify that this does not compare the value?

app/data_model/answer.py
```diff
@@ -1,24 +1,41 @@
...
+    def for_json(self) -> Dict:
+        output = self.to_dict()
+        if not self.list_item_id:
+            del output['list_item_id']
+        return output
+
+    def to_dict(self) -> Dict:
+        return asdict(self)
+
+    def matches(self, answer: Answer):
```
jonnyshaw89: I can't find any valid usages of this.

jonnyshaw89 (changes requested):

app/data_model/answer_store.py
bitdivision: Sorted I think

app/data_model/answer_store.py
```diff
@@ -1,6 +1,7 @@
-from jinja2 import escape
+from __future__ import annotations
```
jonnyshaw89, line 1: The current code is not good enough for Rich. He needs the code of the future!!!

app/forms/fields.py
```diff
@@ -283,8 +282,8 @@ def get_schema_defined_limit(answer_id, definition, answer_store):
             value = definition['value']
         else:
             source_answer_id = definition.get('answer_id')
-            answer_list = answer_store.filter(answer_ids=[source_answer_id])
-            value = next(iter(answer_list)).get('value')
+            answer = answer_store.get_answer(source_answer_id)
```
jonnyshaw89, line 285: 👍 🎉

jonnyshaw89:
I think it needs some functional tests

app/views/questionnaire.py
```diff
@@ -151,22 +169,102 @@ def post_block(routing_path, schema, metadata, collection_metadata, answer_store
...
+def perform_list_action(schema, metadata, answer_store, current_location, form, rendered_block, answer_store_updater, list_item_id):
+    list_collector_child = schema.is_block_list_collector_child(current_location.block_id)
+
+    block = schema.get_block(current_location.block_id)
+
+    parent_block = schema.get_list_collector_for_block_id(current_location.block_id)
+
+    if block['type'] == 'ListCollector':
         answer_store_updater.save_answers(form)
+        if list(form.data.values())[0] == block['add_answer_value']:
```
ajmaddaford: We need to find a better way to identify the add answer in the form data. We could expand the schema from:

```
add_answer_value: '<value>'
```

to:
```
add_answer: {
  id: <answer-id>,
  value: <value>
}
```

app/views/questionnaire.py
```diff
@@ -151,22 +169,102 @@ def post_block(routing_path, schema, metadata, collection_metadata, answer_store
...
+            add_url = url_for('questionnaire.get_add_list_item', list_name=rendered_block['populates_list'], add_block_id=rendered_block['add_block']['id'])
+            return add_url
+    elif block['type'] == 'ListRemoveQuestion':
+        if list(form.data.values())[0] == parent_block['remove_answer_value']:
```
ajmaddaford: Same as comment above about the add answer value

app/views/questionnaire.py
```diff
@@ -151,22 +169,102 @@ def post_block(routing_path, schema, metadata, collection_metadata, answer_store
...
+            list_name = parent_block['populates_list']
+            answer_store_updater.remove_all_answers_with_list_item_id(list_name, list_item_id)
+        else:
+            return url_for('questionnaire.get_block', block_id=parent_block['id'])
+    elif block['type'] == 'ListAddQuestion':
+        answer_store_updater.add_new_list_item_answers(form, parent_block['populates_list'])
+    elif block['type'] == 'ListEditQuestion':
+        answer_store_updater.save_answers(form, save_completed_blocks=False)
+    else:
+        raise NotImplementedError(f'List collection block type: {block["type"]} not implemented')
```
ajmaddaford: We should probably remove this as it can never be reached - line 182 checks for list block types before calling this method.

app/views/questionnaire.py
```diff
@@ -151,22 +169,102 @@ def post_block(routing_path, schema, metadata, collection_metadata, answer_store
...
+
+    block = schema.get_block(current_location.block_id)
+
+    parent_block = schema.get_list_collector_for_block_id(current_location.block_id)
+
+    if block['type'] == 'ListCollector':
         answer_store_updater.save_answers(form)
+        if list(form.data.values())[0] == block['add_answer_value']:
+            add_url = url_for('questionnaire.get_add_list_item', list_name=rendered_block['populates_list'], add_block_id=rendered_block['add_block']['id'])
+            return add_url
```
ajmaddaford, line 227: We should return `None` here if we're not redirecting to the add screen, as the rest of the logic in this method is irrelevant to ListCollector blocks. Also means we can remove the `if list_collector_child` conditional below.

app/data_model/answer_store.py
bitdivision: For anyone wondering, this is to allow using the current class as a type annotation.

app/helpers/session_helpers.py
```diff
@@ -17,6 +17,18 @@ def wrapped_function(*args, **kwargs):
...
 
+def with_list_store(function):
+    """Adds the `list_store` as an argument, where the `current_user` is defined.
+    Use on flask request handlers or methods called by flask request handlers.
+
+    May error unless there is a `current_user`."""
+    @wraps(function)
+    def wrapped_function(*args, **kwargs):
+        answer_store = get_list_store(current_user)
+        return function(answer_store, *args, **kwargs)
```
MebinAbraham: `answer_store` > `list_store`
```python
def wrapped_function(*args, **kwargs):
        list_store = get_list_store(current_user)
        return function(list_store, *args, **kwargs)
```

app/views/questionnaire.py
```diff
@@ -469,6 +571,37 @@ def _render_template(context, current_location, template, previous_url, schema,
...
+    return _render_page(rendered_block['type'], context, current_location, schema)
+
+
+def validate_list_block(schema, block_id, list_name, list_item_id=None):
+    parent_block = schema.get_list_collector_for_block_id(block_id)
+
+    try:
+        validate_list_collector_route(list_name, list_item_id)
+    except InvalidListItemId:
+        return redirect(url_for('questionnaire.get_block', block_id=parent_block['id']))
```
ajmaddaford: Now that this method and `validate_list_collector_route` have been simplified it might be clearer to remove the methods and do the logic in `validate_location`.

scripts/test_schemas.sh
```diff
@@ -6,8 +6,8 @@ if [ "$1" == "--local" ] || [ "$2" == "--local" ]; then
 fi
 
 if [ "$run_docker" == true ]; then
-    docker pull onsdigital/eq-schema-validator:v3
-    validator="$(docker run -d -p 5001:5000 onsdigital/eq-schema-validator:v3)"
+    docker pull onsdigital/eq-schema-validator:list-collector-changes
+    validator="$(docker run -d -p 5001:5000 onsdigital/eq-schema-validator:list-collector-changes)"
```
MebinAbraham: Reminder to revert before merge.

tests/functional/spec/list_collector.spec.js
```diff
@@ -0,0 +1,174 @@
...
+        .setValue(ListCollectorAddPage.lastName(), 'Clemens')
+        .click(ListCollectorAddPage.submit())
+        .click(ListCollectorPage.yes())
+        .click(ListCollectorPage.submit())
+        .setValue(ListCollectorAddPage.firstName(), 'Suzy')
+        .setValue(ListCollectorAddPage.lastName(), 'Clemens')
+        .click(ListCollectorAddPage.submit());
+    });
+
+    it('Shows all of the household members in the summary', function() {
```
MebinAbraham: I get why these are labelled like these but we normally have the `Given`, `When`, `Then` but I guess this entire spec is one big test.

MebinAbraham (changes requested):
Looks good, tested some journeys and seems to behave as expected.

app/views/questionnaire.py
```diff
@@ -216,8 +216,6 @@ def perform_list_action(schema, metadata, answer_store, current_location, form,
         answer_store_updater.add_new_list_item_answers(form, parent_block['populates_list'])
     elif block['type'] == 'ListEditQuestion':
         answer_store_updater.save_answers(form, save_completed_blocks=False)
-    else:
-        raise NotImplementedError(f'List collection block type: {block["type"]} not implemented')
```
MebinAbraham: 👍 Validator shouldn't allow arbitrary types anyway.

app/views/questionnaire.py
```diff
@@ -151,22 +168,99 @@ def post_block(routing_path, schema, metadata, collection_metadata, answer_store
...
+@login_required
+@with_answer_store
+@with_list_store
+@with_collection_metadata
+@with_metadata
+@with_schema
+@full_routing_path_required
+# pylint: disable=too-many-locals, too-many-return-statements
+def post_block(routing_path, schema, metadata, collection_metadata, list_store, answer_store, block_id):
+    return post_block_handler(routing_path, schema, metadata, collection_metadata, list_store, answer_store, block_id)
```
ajmaddaford: I think we should consider a `@with_questionnaire store` decorator at the top level of routes, as currently the routes have individual `with_answer_store`, `with_list_store`, `with_collection_metadata`, `with_metadata` - all of which call `get_questionnaire_store`. `questionnaire_store` could then be passed to post_block_handler instead of `metadata, collection_metadata, list_store, answer_store`.

We can then only pass what's required to methods called from `post_block_handler` etc by accessing them from `questionnaire_store`, e.g. `questionnaire_store.answer_store`.

ajmaddaford (changes requested):

tests/functional/spec/list_collector.spec.js
bitdivision: Yeah, it's a little different. I'm not sure the given, when, then pattern really helps here (Or ever IMO). 

I'll reword them a little to fit better.

app/views/questionnaire.py
bitdivision: Forgot about this, sounds good. Will work through it now.

tests/functional/spec/list_collector.spec.js
```diff
@@ -23,14 +23,14 @@ function checkPeopleInList(peopleExpected) {
   return chain;
 }
 
-describe('List Collector', function() {
+describe('@watch List Collector', function() {
```
MebinAbraham: FYI

app/views/questionnaire.py
```diff
@@ -91,50 +91,72 @@ def before_post_submission_request():
...
 @with_schema
-def get_add_list_item(schema, metadata, answer_store, list_name, add_block_id):
-    return get_block_handler(None, schema, metadata, answer_store, add_block_id, list_name)
+def get_add_list_item(schema, questionnaire_store, list_name, add_block_id):
+    return get_block_handler(
+        None,
+        schema,
+        questionnaire_store.metadata,
+        questionnaire_store.answer_store,
+        questionnaire_store.list_store,
```
ajmaddaford:
```
questionnaire_store.metadata,
questionnaire_store.answer_store,
questionnaire_store.list_store,
```
could be passed as just `questionnaire_store`

app/views/questionnaire.py
```diff
@@ -91,50 +91,72 @@ def before_post_submission_request():
...
     router = Router(schema, routing_path, current_location, completed_locations)
 
     if schema.is_block_list_collector_child(current_location.block_id):
-        to_redirect = validate_list_block(schema, current_location.block_id, current_location.list_name, current_location.list_item_id)
-        if to_redirect:
-            return to_redirect
+        list_item_id = current_location.list_item_id
+        if list_item_id and list_item_id not in list_store[current_location.list_name]:
+            parent_block = schema.get_list_collector_for_block_id(current_location.block_id)
+            return redirect(url_for('questionnaire.get_block', block_id=parent_block['id']))
```
ajmaddaford: I think we need to expand on the validation - you can visit `/questionnaire/another-remove-person` and click remove person, giving a 500 error.

app/data_model/answer_store.py
```diff
@@ -42,97 +55,80 @@ def _validate(answer):
...
-
-    def add_or_update(self, answer):
+    def add_or_update(self, answer: Answer):
         """
         Add a new answer into the answer store, or update if it exists.
-
-        :param answer: An answer object.
         """
         self._validate(answer)
+        store_key = (answer.answer_id, answer.list_item_id)
```
ajmaddaford: Consistency, this is called `key` in another method.

app/data_model/list_store.py
```diff
@@ -0,0 +1,80 @@
...
+    def __getitem__(self, item):
+        try:
+            return self._lists[item]
+        except KeyError:
+            return list()
+
+    def _generate_identifier(self):
+        """ Generate an unused random 6 character string"""
+        while True:
+            candidate = random_string(6)
```
ajmaddaford: We should use a constant for this.

app/helpers/session_helpers.py
```diff
@@ -17,6 +27,18 @@ def wrapped_function(*args, **kwargs):
     return wrapped_function
 
 
+def with_list_store(function):
```
ajmaddaford: Are any of the `with_*` methods other than `with_questionnaire_store` used now?

app/questionnaire/answer_store_updater.py
```diff
@@ -1,3 +1,4 @@
+from typing import List
```
ajmaddaford: Now that this is updating the list store, in addition to completed blocks, we should rename it `questionnaire_store_updater`.

MebinAbraham (changes requested):
List answers are not submitted.
```json
// http://localhost:5000/dump/answers

{
  "answers": [
    {
      "answer_id": "first-name",
      "value": "A",
      "list_item_id": "BtfFBi"
    },
    {
      "answer_id": "last-name",
      "value": "B",
      "list_item_id": "BtfFBi"
    },
    {
      "answer_id": "anyone-else",
      "value": "No"
    },
    {
      "answer_id": "another-anyone-else",
      "value": "No"
    }
  ]
}
```

```json
// http://localhost:5000/dump/submission

{
  "submission": {
    "tx_id": "fb26c9f6-2bf2-468e-9855-ab21bad86a25",
    "type": "uk.gov.ons.edc.eq:surveyresponse",
    "version": "0.0.2",
    "origin": "uk.gov.ons.edc.eq",
    "survey_id": "0",
    "flushed": false,
    "submitted_at": "2019-04-18T13:17:19.201441",
    "collection": {
      "exercise_sid": "72ac6aac-eb37-4f22-8132-8ea4276086f0",
      "instrument_id": "list_collector",
      "period": "201605"
    },
    "metadata": {
      "user_id": "UNKNOWN",
      "ru_ref": "94632870674y"
    },
    "case_id": "32206e56-08ee-4fc9-b4f0-8475ce4ec851",
    "started_at": "2019-04-18T13:12:35.562838",
    "data": [
      {
        "answer_id": "anyone-else",
        "value": "No"
      },
      {
        "answer_id": "another-anyone-else",
        "value": "No"
      }
    ]
  }
}
```

app/questionnaire/answer_store_updater.py
```diff
@@ -13,15 +14,26 @@ def __init__(self, current_location, schema, questionnaire_store, current_questi
...
 
-        if self._current_location not in self._questionnaire_store.completed_blocks:
-            self._questionnaire_store.completed_blocks.append(self._current_location)
+        if save_completed_blocks:
+            if self._current_location not in self._questionnaire_store.completed_blocks:
+                self._questionnaire_store.completed_blocks.append(self._current_location)
 
         self._questionnaire_store.add_or_update()
 
+    def add_new_list_item_answers(self, form, list_name):
```
ajmaddaford: To be consistent with the `save_answers` method this should be named `save_new_list_item_answers`.

ajmaddaford (changes requested):

app/views/questionnaire.py
```diff
@@ -89,52 +89,69 @@ def before_post_submission_request():
...
+def validate_location(schema, routing_path, current_location):
     completed_locations = get_completed_blocks(current_user)
     router = Router(schema, routing_path, current_location, completed_locations)
 
-    if not router.can_access_location():
-        next_location = router.get_next_location()
-        return _redirect_to_location(next_location)
+    if schema.is_block_list_collector_child(current_location.block_id):
+        to_redirect = validate_list_block(schema, current_location.block_id, current_location.list_name, current_location.list_item_id)
+        if to_redirect:
```
MebinAbraham: Would this check be better suited to live in the `Router` object?

app/views/questionnaire.py
MebinAbraham: Visiting `/questionnaire/<list_name>/<list_item_id>/invalid_block_id` returns a 500 also.

app/data_model/answer_store.py
```diff
@@ -11,13 +12,24 @@ class AnswerStore:
...
-            self.answer_map = self._build_map(existing_answers or [])
-        else:
-            self.answer_map = existing_answers or {}
+    def __init__(self, existing_answers: List[Answer] = None):
+        """ Instantiate an answer_store.
+
+        Args:
+            existing_answers: If a list of answer objects is provided, this will be used to initialise the store.
+        """
+        self.answer_map = self._build_map(existing_answers or [])
```
MebinAbraham, line 31: 👍

app/data_model/answer.py
JamesGardiner: ooooh dataclasses :smile:

app/data_model/answer_store.py
```diff
@@ -29,11 +41,12 @@ def __eq__(self, other):
...
 
     @staticmethod
-    def _build_map(answers):
+    def _build_map(answers: List[Union[Answer, Dict]]):
+        """ Builds the answer_store's data structure from a list of Answer objects"""
         answer_map = {}
 
         for answer in answers:
-            answer_map[answer['answer_id']] = answer
+            answer_map[answer['answer_id'], answer.get('list_item_id')] = Answer.from_dict(answer)
```
JamesGardiner:
```suggestion
		answer_map = {(answer['answer_id'], answer.get('list_item_id')]): Answer.from_dict(answer) for answer in answers}
```

JamesGardiner: Dict comp would be more efficient?

app/data_model/list_store.py
```diff
@@ -0,0 +1,82 @@
...
+
+class ListStore:
+    """ Storage for lists.
+    """
+
+    def __init__(self, existing_items=None):
+        if not existing_items:
+            existing_items = []
+
+        self._lists = OrderedDict(existing_items)
```
JamesGardiner: We aren't using any of the features that OrderedDict offers over a standard dict in 3.7, e.g. ordered equality comparison or `move_to_end`. This should be a standard Dict.

app/data_model/list_store.py
```diff
@@ -0,0 +1,82 @@
...
+            ids.extend(named_list)
+
+        return ids
+
+    def delete_list_item_id(self, list_name, item_id):
+        self._lists[list_name].remove(item_id)
+        if not self._lists[list_name]:
+            del self._lists[list_name]
+
+    def add_list_item(self, list_name):
```
JamesGardiner, line 50: Could this not be implemented with `__setitem__`?

app/data_model/list_store.py
```diff
@@ -0,0 +1,82 @@
...
+        new_list_identifier = self._generate_identifier()
+
+        named_list.append(new_list_identifier)
+
+        self._lists[list_name] = named_list
+
+        return new_list_identifier
+
+    def serialise(self):
+        serialised = [{'name': name, 'items': list_items} for name, list_items in self._lists.items()]
```
JamesGardiner:
```suggestion
        return [{'name': name, 'items': list_items} for name, list_items in self._lists.items()]
```

app/data_model/answer_store.py
```diff
@@ -42,97 +55,80 @@ def _validate(answer):
...
+        if existing_answer != answer:
+            self.dirty = True
 
-            `answer_store.filter(answer_id=example_id).count()`
+        self.answer_map[key] = answer
 
-        :return: Number of answers in this store.
-        """
-        return len(self)
+    def get_answer(self, answer_id: str, list_item_id: str = None) -> Optional[Answer]:
```
JamesGardiner, line 71:
```suggestion
    def __getitem__(self, (answer_id: str, list_item_id: str = None)) -> Optional[Answer]:
```

app/globals.py
```diff
@@ -83,6 +83,11 @@ def get_answer_store(user):
     return questionnaire_store.answer_store
 
 
+def get_list_store(user):
+    questionnaire_store = get_questionnaire_store(user.user_id, user.user_ik)
```
JamesGardiner:
```suggestion
    return get_questionnaire_store(user.user_id, user.user_ik).list_store
```

app/helpers/form_helper.py
```diff
@@ -88,20 +88,19 @@ def clear_detail_answer_field(data, question):
...
+    block_id = location.block_id
+
     answer_ids = schema.get_answer_ids_for_block(block_id)
 
     result = {}
-    for answer in answer_store.filter(answer_ids=answer_ids):
-        answer_id = answer['answer_id']
-        result[answer_id] = answer['value']
+    answers = answer_store.get_answers_by_answer_id(answer_ids=answer_ids, list_item_id=location.list_item_id)
+    for answer in answers:
```
JamesGardiner:
```suggestion
    result = {answer_id: answer.value for answer in answers}
```

JamesGardiner (dismissed):

app/data_model/answer_store.py
bitdivision: Not sure how much more efficient it is, but certainly cleaner.

app/data_model/list_store.py
bitdivision: We do need to preserve the order of these elements, and I was under the impression that python dictionaries shouldn't be used as ordered dicts (even though they work) since the ordering is not a language feature, just an implementation detail?

app/data_model/list_store.py
bitdivision: I'm not sure whether `__setitem__` makes sense here. 

This method is generating a new identifier and adding it to the named list.
```
new_list_item_id = store.add_list_item('people')
```
So if we used `__setitem__` you'd need to split the generation out. That might be a good idea though. I'll have a think

app/data_model/answer_store.py
```diff
@@ -29,110 +41,90 @@ def __eq__(self, other):
         return self.answer_map == other.answer_map
 
     @staticmethod
-    def _build_map(answers):
-        answer_map = {}
-
-        for answer in answers:
-            answer_map[answer['answer_id']] = answer
+    def _build_map(answers: List[Union[Answer, Dict]]):
```
jonnyshaw89: We should remove the `Dict` and consistently pass a list of `Answer`. Currently this method looks overloaded with multiple input types

bitdivision:
The only major remaining issue is the fact that submission currently only submits answers which are on the routing path. Currently, the list answers (e.g. names) are not on the routing path. This means that they will not be submitted.

Having thought about this a bit, I think the best way to resolve this is to add a special case for ListCollector blocks which lie on the routing path. For any of these blocks, get the list which they populate and use any list answers for that list. I'll start working on that now.

app/data_model/answer_store.py
```diff
@@ -29,110 +41,90 @@ def __eq__(self, other):
...
-        answer_to_add = vars(answer).copy()
-        self.answer_map[answer.answer_id] = answer_to_add
+        existing_answer = self.answer_map.get(key)
 
-    def count(self):
-        """
-        Count of the number of answers in the answer store.
-        NB: can be combined with `filter` method to find count of an answer, e.g.:
+        if existing_answer != answer:
+            self.dirty = True
```
jonnyshaw89: 👍

app/data_model/answer_store.py
```diff
@@ -29,110 +41,90 @@ def __eq__(self, other):
...
+                keys_to_delete.append((answer.answer_id, answer.list_item_id))
 
-        :return: Return a unique hash value
-        """
-        return hash(json.dumps(self.answer_map, sort_keys=True))
+        for key in keys_to_delete:
+            del self.answer_map[key]
+
+    def serialise(self):
+        return list(self.answer_map.values())
```
jonnyshaw89, line 135: Doesn't `.values()` already return a list?

app/data_model/list_store.py
```diff
@@ -0,0 +1,82 @@
...
+    def _generate_identifier(self):
+        """ Generate an unused random 6 character string"""
+        while True:
+            candidate = random_string(EQ_LIST_ITEM_ID_LENGTH)
+            if candidate not in self.list_item_ids():
+                return candidate
+
+    def list_item_ids(self):
+        ids = []
+        for named_list in self._lists.values():
```
jonnyshaw89, line 40: can we not just `return self._lists.values()`

app/globals.py
```diff
@@ -83,6 +83,11 @@ def get_answer_store(user):
     return questionnaire_store.answer_store
 
 
+def get_list_store(user):
```
jonnyshaw89: not used

app/helpers/session_helpers.py
```diff
@@ -2,40 +2,15 @@
 
 from flask_login import current_user
 
-from app.globals import get_answer_store, get_metadata, get_collection_metadata
+from app.globals import get_answer_store, get_metadata, get_collection_metadata, get_list_store, get_questionnaire_store
 
-
-def with_answer_store(function):
-    """Adds the `answer_store` as an argument, where the `current_user` is defined.
+def with_questionnaire_store(function):
```
jonnyshaw89, line 8: 👍

app/questionnaire/location.py
```diff
@@ -1,53 +1,44 @@
...
+        return cls(block_id, list_name, list_item_id)
 
-    def to_dict(self):
-        return vars(self)
+    def for_json(self) -> dict:
+        """
+        Used to serialise a location to json.
+        """
+        attributes = vars(self)
+        return {k: v for k, v in attributes.items() if v is not None}
```
jonnyshaw89, line 37: `is not None` 👍

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -13,15 +14,26 @@ def __init__(self, current_location, schema, questionnaire_store, current_questi
...
         self._answer_store = self._questionnaire_store.answer_store
+        self._list_store = self._questionnaire_store.list_store
 
-    def save_answers(self, form):
+    def save_answers(self, form, save_completed_blocks=True):
         self._update_questionnaire_store_with_form_data(form.data)
 
-        if self._current_location not in self._questionnaire_store.completed_blocks:
-            self._questionnaire_store.completed_blocks.append(self._current_location)
+        if save_completed_blocks:
```
jonnyshaw89, line 22: I don't get why?

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -35,8 +47,25 @@ def _update_questionnaire_store_with_form_data(self, form_data):
...
+
+    def remove_all_answers_with_list_item_id(self, list_name: str, list_item_id: str):
+        """ Remove answers from the answer store and update the list store to remove it
+        """
+        self._list_store.delete_list_item_id(list_name, list_item_id)
+
+        self._answer_store.remove_all_answers_for_list_item_id(list_item_id=list_item_id)
+
+        self._questionnaire_store.add_or_update()
+        self._answer_store.dirty = True
```
jonnyshaw89: can we not set `dirty` in `_answer_store.remove_all_answers_for_list_item_id` as that knows if anything was removed?

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -35,8 +47,25 @@ def _update_questionnaire_store_with_form_data(self, form_data):
...
+
+    def remove_answer_ids(self, answer_ids: List):
+        for answer_id in answer_ids:
+            self._answer_store.remove_answer(answer_id)
+        self._questionnaire_store.add_or_update()
+        self._answer_store.dirty = True
```
jonnyshaw89: Same here `_answer_store.remove_answer` knows if anything was removed to the `dirty` should be set there?
Also should we make `dirty` `_dirty` so its only modified from within the store?

app/questionnaire/rules.py
```diff
@@ -231,20 +216,20 @@ def evaluate_when_rules(when_rules, schema, metadata, answer_store, routing_path
 
 
 def get_answer_store_value(answer_id, answer_store, schema, routing_path=None):
+    """ Return answer value assuming it is on the routing path.
+    If answer is not on the routing path, return None
+    If routing_path empty, return answer value
```
jonnyshaw89, line 224: Why do we still return if `routing_path=None`?

jonnyshaw89 (changes requested):

app/data_model/list_store.py
bitdivision: I'm out of date. Looks like this is now a language feature in 3.7. I'll switch it over.

app/data_model/answer_store.py
bitdivision: Not quite, it returns a view as a dict_values instance. Unfortunately that's not JSON serialisable.

app/data_model/list_store.py
bitdivision: We could. I don't think I would expect this to return multiple lists though.

Maybe we'd be better served by a `test_membership` method or similar instead.

app/questionnaire/questionnaire_store_updater.py
bitdivision: We don't want to add blocks which are not on the routing path to completed blocks.

app/data_model/list_store.py
jonnyshaw89: Ah yes, because its extending

app/questionnaire/questionnaire_store_updater.py
bitdivision: Agree it should only be mutated within the answer store, but does need to be read outside.

I'll add a property to the class to read it.

app/questionnaire/rules.py
bitdivision: I think so? Caught by the else I think?

jonnyshaw89: I don't think we ever want to return a value without checking if it's on the path. It could lead to bugs with returning or routing on values not on the path

bitdivision: Ah, sorry, I misread. Yeah, makes sense.

app/globals.py
bitdivision: Still used by _get_context()

bitdivision: But that can be switched to the questionnaire_store

app/data_model/answer_store.py
```diff
@@ -121,10 +126,11 @@ def remove_all_answers_for_list_item_id(self, list_item_id: str):
 
         for answer in self:
             if answer.list_item_id == list_item_id:
-                keys_to_delete.append((answer.answer_id, answer.list_item_id))
+                self.remove_answer(answer.answer_id, answer.list_item_id)
 
         for key in keys_to_delete:
```
jonnyshaw89: now redundant

app/data_model/answer_store.py
bitdivision: I have changed this to only accept a list of Dicts. This method is only really needed for deserialisation, at which point we will always have dicts, not Answers (assuming we serialise to JSON)

app/submitter/convert_payload_0_0_3.py
```diff
@@ -8,16 +14,56 @@ def convert_answers_to_payload_0_0_3(answer_store, schema, routing_path):
...
+
+    For list answers, this method will query the list store and get all answers from the
+    add list item block. If there are multiple list collectors for one list, they will have
+    the same answer_ids, and will not be duplicated.
+
+    Returns:
+        A list of answer dictionaries.
     """
-    data = []
+    temp_answers = AnswerStore()
```
jonnyshaw89: what is temp about these answers?

app/submitter/converter.py
```diff
@@ -14,16 +14,10 @@ def __str__(self):
         return 'Data version {} not supported'.format(self.version)
 
 
-def convert_answers(metadata, collection_metadata, schema, answer_store, routing_path, flushed=False):
+def convert_answers(schema, questionnaire_store, routing_path, flushed=False):
```
jonnyshaw89, line 18: 👍

app/submitter/convert_payload_0_0_3.py
bitdivision: I'll change the name.

I'm also not entirely happy with this solution. I wanted to use a set here, which removes a load of code from this loop. This doesn't work because we sometimes store lists as values in answers, which means the Answer object is not hashable. I suppose one solution here is to override __hash__ and do something hacky, but the list would still be mutable, so it's not ideal.

tests/app/submitter/test_convert_payload_0_0_3.py
```diff
@@ -0,0 +1,341 @@
+from app.data_model.answer_store import AnswerStore
+from app.data_model.answer import Answer
+from app.questionnaire.questionnaire_schema import QuestionnaireSchema
+from app.questionnaire.location import Location
+from app.submitter.converter import convert_answers
+from tests.app.submitter.schema import make_schema
+
+
+def test_convert_answers_to_payload_0_0_3(fake_questionnaire_store):
```
jonnyshaw89, line 12: Missing a test with `list_item_id`

jonnyshaw89 (changes requested):

tests/app/submitter/test_convert_payload_0_0_3.py
bitdivision: fixed

tests/functional/spec/section_summary.spec.js
```diff
@@ -5,7 +5,7 @@ const AddressDurationPage = require('../generated_pages/section_summary/address-
 const PropertyDetailsSummaryPage = require('../generated_pages/section_summary/property-details-summary.page.js');
 const FinalSummaryPage = require('../generated_pages/section_summary/summary.page.js');
 
-describe('Section Summary', function() {
+describe('@watch Section Summary', function() {
```
jonnyshaw89: ding ding

tests/app/submitter/test_convert_payload_0_0_3.py
```diff
@@ -339,3 +340,63 @@ def test_unit_answer(fake_questionnaire_store):
...
+    })
+
+    fake_questionnaire_store.answer_store = answers
+    fake_questionnaire_store.list_store = list_store
+
+    schema = load_schema('test', 'list_collector')
+
+    answer_object = convert_answers(schema, fake_questionnaire_store, routing_path)
+
+    assert len(answer_object['data']) == len(answers)
```
jonnyshaw89: Should probably check that the `list_item_id` is in the output?

jonnyshaw89 (changes requested):

jonnyshaw89 (dismissed):
🎉 💥

data/en/test_list_collector.json
```diff
@@ -0,0 +1,252 @@
...
+    "sections": [{
+        "id": "section",
+        "groups": [{
+                "id": "group",
+                "title": "List",
+                "blocks": [{
+                        "type": "Introduction",
+                        "id": "introduction",
+                        "title": "Introduction",
+                        "description": ""
```
ajmaddaford: Do the test schemas need an introduction page?

app/data_model/list_store.py
```diff
@@ -0,0 +1,79 @@
...
+    return ''.join(random.choice(ascii_letters) for _ in range(length))
+
+
+class ListStore:
+    """ Storage for lists.
+    """
+
+    def __init__(self, existing_items=None):
+        if not existing_items:
+            existing_items = []
```
MebinAbraham, line 21: Was there a reason this is not `def __init__(self, existing_items=[]):` as the `if` is essentially redundant.?

Pipfile
```diff
@@ -60,6 +60,7 @@ python-snappy = "*"
 google-cloud-storage = "*"
 jsonpointer = "*"
 redis = "*"
+isort = "==4.3.10"
```
ajmaddaford: Should this be in dev-packages?

app/data_model/answer_store.py
```diff
@@ -29,110 +41,89 @@ def __eq__(self, other):
...
+            raise TypeError(f'Method only supports Answer argument type, found type: {type(answer)}')
 
-    def copy(self):
-        """
-        Create a new instance of answer_store with the same values.
-        """
-        return self.__class__(self.answer_map.copy())
+    @property
+    def dirty(self):
+        return self._dirty
```
ajmaddaford: Minor, in other places we use the prefix `is_` to make it clear it's a boolean e.g. `is_goto_rule` or `is_survey_completed`, this could be `is_dirty`

app/data_model/list_store.py
```diff
@@ -0,0 +1,79 @@
...
+            candidate = random_string(EQ_LIST_ITEM_ID_LENGTH)
+            if candidate not in self.list_item_ids():
+                return candidate
+
+    def list_item_ids(self):
+        ids = []
+        for named_list in self._lists.values():
+            ids.extend(named_list)
+
+        return ids
```
MebinAbraham, line 43: Maybe just a list comprehension?

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -13,15 +14,26 @@ def __init__(self, current_location, schema, questionnaire_store, current_questi
...
         self._questionnaire_store.add_or_update()
 
+    def save_new_list_item_answers(self, form, list_name):
+        new_list_item_id = self._list_store.add_list_item(list_name)
+
+        self._current_location.list_item_id = new_list_item_id
+
+        self.save_answers(form, False)
+
+        return self._current_location
```
MebinAbraham: I don't get why the current location is returned in this method?

app/questionnaire/location.py
```diff
@@ -1,53 +1,44 @@
...
-    def url(self):
+    def url(self) -> str:
         """
         Return the survey runner url that this location represents
 
         :return:
         """
-        return url_for('questionnaire.get_block',
-                       block_id=self.block_id)
+        return url_for('questionnaire.get_block', block_id=self.block_id)
```
ajmaddaford: Won't this return an invalid url for a location with list name and list item id?

app/submitter/convert_payload_0_0_1.py
```diff
@@ -1,88 +0,0 @@
-from collections import OrderedDict
```
ajmaddaford: As discussed, this should be re-instated as the this will be needed to migrate business surveys to v3

app/submitter/convert_payload_0_0_3.py
```diff
@@ -0,0 +1,67 @@
...
+    Convert answers into the data format below
+    'data': [
+        {
+            'value': 'Joe Bloggs',
+            'answer_id': 'household-full-name',
+        },
+        {
+            'value': 'Husband or wife',
+            'list_item_id': 'abc123',
+            'answer_id': 'who-is-related',
```
ajmaddaford: Probably not the best example as relationships will also need to capture the list item id of the person they're related to.

app/submitter/convert_payload_0_0_3.py
```diff
@@ -0,0 +1,67 @@
...
+        answers_in_block = answer_store.get_answers_by_answer_id(answer_ids, list_item_id=location.list_item_id)
+        for answer_in_block in answers_in_block:
+            if answer_in_block:
+                answers.add_or_update(answer_in_block)
+
+    return list(answers.answer_map.values())
+
+
+def convert_list_collector_answers(answer_store, list_store, schema, location) -> List[Answer]:
+    """For the given location, check for list collector and generate submission payload
```
ajmaddaford: This method doesn't generate a submission payload, it returns a list of matching answers from the answer store.

app/submitter/convert_payload_0_0_3.py
```diff
@@ -0,0 +1,67 @@
...
+        A list of answer objects (or None)
+    """
+    answer_output = []
+    block = schema.get_block(location.block_id)
+    if block['type'] == 'ListCollector':
+        add_block_answer_ids = schema.get_answer_ids_for_block(block['add_block']['id'])
+        list_name = block['populates_list']
+        try:
+            list_item_ids = list_store[list_name]
+        except KeyError:
```
MebinAbraham: Would this ever be a `KeyError` since this is only done for answers on the routing path hence the list would exist?

app/submitter/convert_payload_0_0_3.py
```diff
@@ -0,0 +1,67 @@
...
+
+def convert_list_collector_answers(answer_store, list_store, schema, location) -> List[Answer]:
+    """For the given location, check for list collector and generate submission payload
+
+    Returns:
+        A list of answer objects (or None)
+    """
+    answer_output = []
+    block = schema.get_block(location.block_id)
+    if block['type'] == 'ListCollector':
```
ajmaddaford: I think it would be clearer if this conditional was moved outside of this method

app/views/questionnaire.py
```diff
@@ -150,22 +200,110 @@ def post_block(routing_path, schema, metadata, collection_metadata, answer_store
...
-        _set_started_at_metadata_if_required(form, collection_metadata)
-        questionnaire_store = get_questionnaire_store(current_user.user_id, current_user.user_ik)
-        answer_store_updater = AnswerStoreUpdater(current_location, schema, questionnaire_store, rendered_block.get('question'))
+    if not form.validate():
+        context = build_view_context(block['type'], metadata, schema, list_store, answer_store, rendered_block,
+                                     current_location, form)
+        return _render_page(block['type'], context, current_location, schema)
+
+    _set_started_at_metadata_if_required(form, collection_metadata)
+    questionnaire_store = get_questionnaire_store(current_user.user_id, current_user.user_ik)
```
ajmaddaford: Is this different to the `questionnaire_store` passed into this method?

app/views/questionnaire.py
```diff
@@ -150,22 +200,110 @@ def post_block(routing_path, schema, metadata, collection_metadata, answer_store
...
+    answer_store_updater = QuestionnaireStoreUpdater(current_location, schema, questionnaire_store,
```
ajmaddaford: `answer_store_updater` -> `questionnaire_store_updater`

app/views/questionnaire.py
```diff
@@ -150,22 +200,110 @@ def post_block(routing_path, schema, metadata, collection_metadata, answer_store
...
+    answer_store_updater.save_answers(form)
+    next_location = path_finder.get_next_location(current_location=current_location)
+
+    if _is_end_of_questionnaire(block, next_location):
+        return submit_answers(routing_path, schema)
+
+    return redirect(next_location.url())
+
+
+def perform_list_action(schema, metadata, answer_store, current_location, form, rendered_block, answer_store_updater, list_item_id):
```
ajmaddaford: `answer_store_updater` -> `questionnaire_store_updater`

app/views/questionnaire.py
```diff
@@ -150,22 +200,110 @@ def post_block(routing_path, schema, metadata, collection_metadata, answer_store
...
+    )
+
+
+@questionnaire_blueprint.route('<list_name>/<list_item_id>/<block_id>', methods=['POST'])
+@login_required
+@with_questionnaire_store
+@with_schema
+def post_list_item_block(schema, questionnaire_store, list_name, list_item_id, block_id):
+    return post_block_handler(
+        None,
```
ajmaddaford: I think the routing path is required for these methods as we'll need to validate that they can get to the list collector for the add/change/remove block they're posting.

app/views/questionnaire.py
```diff
@@ -150,22 +200,110 @@ def post_block(routing_path, schema, metadata, collection_metadata, answer_store
...
+@with_questionnaire_store
+@with_schema
+def post_add_list_item(schema, questionnaire_store, list_name, block_id):
+    return post_block_handler(
+        None,
+        schema,
+        questionnaire_store,
+        block_id,
+        list_name,
+        None,
```
ajmaddaford: `None` isn't needed here as `list_item_id` defaults to `None`

MebinAbraham (changes requested):
For `test_list_collection.json`, the name collection block is mandatory but you are able to bypass this by visiting an invalid block. I assume it's because those blocks are not on the path.

Ie. When you are on: `http://localhost:5000/questionnaire/people/add-person`, if you visit: `http://localhost:5000/questionnaire/people/invalid-block`, it lets you skip that block even though it is mandatory.

ajmaddaford (changes requested):
There is some inconsistency in the redirects, if you: 
- go to a list item id with a trailing slash and no block id e.g. `/people/<list_item_id>/` you get a 404 page
- go to a list item id with no trailing slash e.g. `/people/<list_item_id> you get redirected to the summary

app/submitter/convert_payload_0_0_3.py
bitdivision: Ah, this is from the v2 tests that were refactored a bit to use fixtures.

I'll update the example.

app/data_model/list_store.py
bitdivision: Yep. https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html

app/data_model/list_store.py
bitdivision: Since we're using `extend` not `append` list comprehension doesn't quite work.

I guess I could use some sort of itertools chain thing, but this looks okay to me.

app/submitter/convert_payload_0_0_3.py
bitdivision: I think I was covering the situation where the list collector is optional. i.e. no answer is ever added to that list so the list_name doesn't exist

I think this line is missing coverage, so I do need to add a test for it anwyay which should reveal the use case if there is one.

app/questionnaire/questionnaire_store_updater.py
bitdivision: Fixed

app/submitter/convert_payload_0_0_1.py
bitdivision: Done

app/data_model/list_store.py
MebinAbraham: Whoops, misread as `append`.

data/en/test_list_collector.json
bitdivision: Fixed

Pipfile
bitdivision: I'm not sure where this came from. Removed for now. I can put it back in dev dependencies if people want it.

app/questionnaire/location.py
bitdivision: Yup, fixed.

app/data_model/answer.py
```diff
@@ -1,34 +1,30 @@
...
-        :param answer_dict: A dictionary representation of the answer.
-        :return: True if both answers match, otherwise False.
-        """
-
-        return self.matches(Answer(
-            answer_dict.get('answer_id'),
-            answer_dict.get('value'),
-        ))
+    answer_id: str
+    value: Union[str, int, float, list]
```
benhowes:
```suggestion
    value: Union[str, int, float, List]
```

app/submitter/convert_payload_0_0_3.py
```diff
@@ -0,0 +1,72 @@
...
+    the same answer_ids, and will not be duplicated.
+
+    Returns:
+        A list of answer dictionaries.
+    """
+    answers = AnswerStore()
+
+    for location in routing_path:
+        if schema.get_block(location.block_id)['type'] == 'ListCollector':
+            for answer in get_answers_for_add_block(answer_store, list_store, schema, location):
```
MebinAbraham: Just me being picky but can we make it consistent between calling a method directly here vs calling a reference like on line 44.

app/submitter/convert_payload_0_0_3.py
```diff
@@ -0,0 +1,72 @@
...
+                if answer:
```
MebinAbraham: This `if` isn't needed?

app/submitter/convert_payload_0_0_3.py
```diff
@@ -0,0 +1,72 @@
...
+                    answers.add_or_update(answer)
+
+        answer_ids = schema.get_answer_ids_for_block(location.block_id)
+        answers_in_block = answer_store.get_answers_by_answer_id(answer_ids, list_item_id=location.list_item_id)
+        for answer_in_block in answers_in_block:
+            if answer_in_block:
```
MebinAbraham: This `if` is also not needed?.

MebinAbraham (changes requested):

MebinAbraham:
My [comment](https://github.com/ONSdigital/eq-survey-runner/pull/2040#pullrequestreview-231603727) about the mandatory answers being skipped is still an issue.

app/submitter/convert_payload_0_0_3.py
bitdivision: The function should just return empty list rather than None if there's a keyerror

bitdivision:
@MebinAbraham I'm working on routing & validation at the moment. Will let you know when that's done.

app/questionnaire/location.py
```diff
@@ -1,53 +1,44 @@
...
-        """
-        String representation of the location, handy for debug messages
-
-        :return:
-        """
-        return str(self)
-
     @classmethod
-    def from_dict(cls, location_dict):
+    def from_dict(cls, location_dict: dict):
```
benhowes: These should use `Dict` from `typing`  (and again in the return type for `from_json`). These will fail if you run it through mypy at the moment

jonnyshaw89: I think we should use mypy so @benhowes can finally be free.

bitdivision:
@andrewmaddaford: Regarding the inconsistency with trailing slashes I have set strict_slashes to False.

Flask's behaviour is documented here: http://flask.pocoo.org/docs/1.0/api/#url-route-registrations

According to that, if the route is defined with a trailing slash (and strict_slashes is True as it is by default), and the user visits the route without a slash, they will be redirected. I was hitting werkzeug errors when trying this, and couldn't easily get to the bottom of it. 

I think strict_slashes=False should work fine for us, but let me know if you disagree.

ajmaddaford (approved):

jonnyshaw89 (approved):

bitdivision:
@MebinAbraham @jonnyshaw89 @ajmaddaford Can you double check my last commit? 

@MebinAbraham  found an issue where using the wrong list_name on e.g. the add page doesn't cause an error, just adds the person to the original list.

MebinAbraham (commented):
As discussed:
`http://localhost:5000/questionnaire/people/add-person/` should be valid and is valid
`http://localhost:5000/questionnaire/people/blah-blah-blah/add-person/` should not be vaild but currently is.

bitdivision:
Have changed this to redirect to the next location if the list name isn't valid (since the list_name in a schema could change in flight).

MebinAbraham (approved):
👍

PR #2041, opened 2019-04-08
Combined routing rules

MebinAbraham:
Added the ability to define multiple response options to route to the same location, in a single routing rule.

Reduced routing rule from:
```
{
    "goto": {
      "block": "current-marriage-status",
      "when": [
        {
          "id": "confirm-marriage-type-answer",
          "condition": "equals",
          "value": "Married"
        }
      ]
    }
  },
  {
    "goto": {
      "block": "current-marriage-status",
      "when": [
        {
          "id": "confirm-marriage-type-answer",
          "condition": "equals",
          "value": "Separated, but still legally married"
        }
      ]
    }
  }
```
To:
```
{
    "goto": {
      "block": "current-marriage-status",
      "when": [
        {
          "id": "confirm-marriage-type-answer",
          "condition": "contains any",
          "value": [
               "Married",
               "Separated, but still legally married"
          ],
        }
      ]
    }
  }
```

### How to review 
Ensure:
- You can route to same location based on multiple response options in a single routing rule.


app/questionnaire/rules.py
```diff
@@ -55,26 +54,43 @@ def evaluate_date_rule(when, answer_store, schema, metadata, answer_value):
...
-        'not equals': lambda answer_value, match_value: answer_value != match_value,
-        'contains': lambda answer_value, match_value: isinstance(answer_value, list) and match_value in answer_value,
-        'not contains': lambda answer_value, match_value: isinstance(answer_value, list) and match_value not in answer_value,
-        'set': lambda answer_value, _: answer_value is not None and answer_value != [],
-        'not set': lambda answer_value, _: answer_value is None or answer_value == [],
-        'greater than': lambda answer_value, match_value: answer_and_match and answer_value > match_value,
-        'greater than or equal to': lambda answer_value, match_value: answer_and_match and answer_value >= match_value,
-        'less than': lambda answer_value, match_value: answer_and_match and answer_value < match_value,
-        'less than or equal to': lambda answer_value, match_value: answer_and_match and answer_value <= match_value,
+        'equals': lambda lhs, rhs: lhs == rhs,
```
jonnyshaw89: 🤢 abbreviation hell

data/en/test_conditional_combined_routing.json
```diff
@@ -51,19 +58,28 @@
                         "block": "response-yes",
                         "when": [{
                             "id": "conditional-routing-answer",
-                            "condition": "equals",
-                            "value": "yes"
+                            "condition": "contains",
+                            "value": ["yes", "sometimes"]
```
jonnyshaw89: I wonder if we can make this more explicit and make this `"values": ["yes", "sometimes"]`

jonnyshaw89 (dismissed):

app/questionnaire/rules.py
MebinAbraham: 👀 hell is worse. Changed.

data/en/test_conditional_combined_routing.json
MebinAbraham: Possibly?. I thought about it but wasn't sure if to go ahead with it. Just an extra property in JSON schema and an extra `if` in runner. Also wasn't sure if additional checks would be appropriate, such as values to be only used with this kind of comparison.

jonnyshaw89: Yeh, I can see why, it makes the code simpler

ajmaddaford:
- I don't think using contains on a singular answer - "does this singular answer value contain this list of values" - reads very well . What we're trying to express is "does this singular answer value equal one of these values".
- The current approach only allows for OR logic, we should take this opportunity to add AND logic (or at least leave it open for this in the future). This would allow us to route multiple checkbox options that must be selected together in a single routing rule.

After discussing this with @bitdivision and @MebinAbraham we've come up with `equals any`, `contains any` and `contains all` - which reads more clearly and allows for AND logic. The example above becomes:

```
{
  "goto": {
    "block": "current-marriage-status",
    "when": [
      {
        "id": "confirm-marriage-type-answer"
        "condition": "equals any",
        "values": [
          "Married",
          "Separated, but still legally married"
        ],
        
      }
    ]
  }
}
```

If we implement this we could remove the contains rule as you should be able to use either `contains any` or `contains all`. We should also add validation in eq-schema-validator that checkbox answers only can use contains comparisons, and all other answer types can't.

MebinAbraham:
@ajmaddaford @jonnyshaw89
Introduced these new conditions:
1. `equals any`
2. `not equals any`
3. `contains any`
4. `contains all`

data/en/test_conditional_combined_routing.json
```diff
@@ -32,12 +32,19 @@
...
                             "label": "No, I prefer tea",
-                            "value": "no",
-                            "description": ""
+                            "value": "no"
+                        }, {
+                            "label": "I don't like coffee",
+                            "value": "I dont like coffee"
+                        }, {
+                            "label": "No, I don't drink any hot drinks",
+                            "value": "no hot drinks"
```
ajmaddaford: The values for these should match the labels, as we do for any real schemas. It's confusing to see the different variations of text in the routing rules.

ajmaddaford (changes requested):
Looks good 👍 

We need to review how we write rules in schema as it's not very clear. I've raised a card in the backlog to discuss this.

data/en/test_conditional_combined_routing.json
MebinAbraham: Done

ajmaddaford (approved):

jonnyshaw89 (approved):

scripts/test_schemas.sh
```diff
@@ -6,8 +6,8 @@ if [ "$1" == "--local" ] || [ "$2" == "--local" ]; then
 fi
 
 if [ "$run_docker" == true ]; then
-    docker pull onsdigital/eq-schema-validator:v3
-    validator="$(docker run -d -p 5001:5000 onsdigital/eq-schema-validator:v3)"
+    docker pull onsdigital/eq-schema-validator:eq-2917-combined-routing-rules
```
jonnyshaw89: This needs changing once merged

PR #2044, opened 2019-04-11
Format possessive placeholder transform only for English surveys

jonnyshaw89:
Only modify the name if we are using the english locale.

### How to review 
Run a test using the welsh language and test that the name is not modified.

### DIFF
--- a/app/templating/summary/group.py
+++ b/app/templating/summary/group.py
@@ -10,7 +10,9 @@ class Group:
         self.title = group_schema.get('title')
 
         self.blocks = self._build_blocks(group_schema, path, answer_store, metadata, schema)
-        self.placeholder_renderer = PlaceholderRenderer(answer_store=answer_store, metadata=metadata)
+        self.placeholder_renderer = PlaceholderRenderer(language='en',
+                                                        answer_store=answer_store,
+                                                        metadata=metadata)
 
     @staticmethod
     def _build_blocks(group_schema, path, answer_store, metadata, schema):
--- a/app/views/questionnaire.py
+++ b/app/views/questionnaire.py
@@ -1,5 +1,6 @@
 from datetime import datetime, timedelta
 
+import flask_babel
 import humanize
 import simplejson as json
 from dateutil.tz import tzutc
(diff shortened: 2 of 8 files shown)


app/views/questionnaire.py
```diff
@@ -107,7 +108,9 @@ def get_block(routing_path, schema, metadata, answer_store, block_id):
 
     transformed_block = transform_variants(block, schema, metadata, answer_store)
 
-    placeholder_renderer = PlaceholderRenderer(answer_store=answer_store, metadata=metadata)
+    placeholder_renderer = PlaceholderRenderer(locale=flask_babel.get_locale(),
```
ajmaddaford: Locale could be passed as a string e.g. `en` or `en_GB` that Babel will understand. This would mean the external interface of the Placeholder renderer wouldn't expose the use of Babel for translations.

app/templating/summary/group.py
```diff
@@ -10,7 +12,9 @@ def __init__(self, group_schema, path, answer_store, metadata, schema):
         self.title = group_schema.get('title')
 
         self.blocks = self._build_blocks(group_schema, path, answer_store, metadata, schema)
-        self.placeholder_renderer = PlaceholderRenderer(answer_store=answer_store, metadata=metadata)
+        self.placeholder_renderer = PlaceholderRenderer(locale=Locale('en', 'GB'),
```
MebinAbraham: This locale would also need to be dynamic.

MebinAbraham (changes requested):
Mostly working as expected, only spotted 1 issue.

app/templating/summary/group.py
jonnyshaw89: Summaries are not currently used in Census, so we can leave this defaulted to `en` for now and address this when we re-write summaries

MebinAbraham: Flask context is available here anyway?. But ya fair.

MebinAbraham (approved):

app/templating/summary/group.py
jonnyshaw89: But re comments yesterday in "planning". We want to limit how much we use Flask outside of the views so we can switch away easier.

ajmaddaford (approved):

PR #2045, opened 2019-04-11
V3 templating upgrade

bameyrick:
Updated runner to use the new design system macros to render templates. I had to create quite a lot of (poorly written) jinja filters to format the configs for some of the macros / add extra attributes for functional/intergration tests to use so I don't know if it would be a separate task to refactor these and make them not so terrible.

### How to review 
1. boot local runner

### Skipped tests
I have skipped some of the functional tests for now as I'm unsure how to fix them:
1. `save_sign_out.spec.js`: I'm not sure if this is a schema error, but the save and sign out button doesn't seem to be appearing when it should. This also seems to be the issue in the `v3` branch.
2. `static-feedback.spec.js` and `test_feedback.py`: I'm not sure if we're actually going to be using the standalone feedback page rather than the footer one yet so didn't dig into why this was failing too much.

### Checklist
* [ ] Check templates are rendering correctly (there may be some minor spacing issues to address so if you spot anything funky looking let me know)
* [ ] Check performance isn't adversely affected.

### Still to do
* [x] Deploy built CSS/JS/Images/fonts etc to CDN
* [x] Figure out a way to import macros to the project as a dependency


saedwards:
Good effort Ben, looks like a hefty piece of work and can't have been easy. This is definitely a step in the right direction. Removing more presentational concerns from EQ will decrease coupling of this problem from the application and promote code reuse across the service, potentially further.

This is a first visit of integrating externalised templates into EQ, but catching problems early on will save a lot of time and money, some observations;

- A lot of conditional config mappings in templates/base-templates to meet macro preconditions, this will be unavoidable to a certain extent but could we organise/arrange this better? _base.html has a bit too much logic in one file. It’ll probably be the case that the application would have to adjust to supply the correct view-model mappings for the templates.

- There’s a lot of great reuse with low level components but it would be great to see more structural level components used where you can inject arbitrary content inside a structural node. Good examples of these would be: header, body, footer (the onsPanel component using caller()).

- Not too sure about using the parameters “classes” and “attributes” for some of the macros and probably should be a last resort. It means the calling context knows too much about the inner workings of exact markup/styles/behaviours that are imported from the design system for each component causing tight coupling. You wouldn’t be able to change the name of classes/attributes in the design system without changing all applications that integrate with it.

- Would be good to see less hard-coded HTML removed to completely abstract presentation markup in future, kind of a given.

bameyrick:
> Good effort Ben, looks like a hefty piece of work and can't have been easy. This is definitely a step in the right direction. Removing more presentational concerns from EQ will decrease coupling of this problem from the application and promote code reuse across the service, potentially further.
> 
> This is a first visit of integrating externalised templates into EQ, but catching problems early on will save a lot of time and money, some observations;
> 
> * A lot of conditional config mappings in templates/base-templates to meet macro preconditions, this will be unavoidable to a certain extent but could we organise/arrange this better? _base.html has a bit too much logic in one file. It’ll probably be the case that the application would have to adjust to supply the correct view-model mappings for the templates.
> * There’s a lot of great reuse with low level components but it would be great to see more structural level components used where you can inject arbitrary content inside a structural node. Good examples of these would be: header, body, footer (the onsPanel component using caller()).
> * Not too sure about using the parameters “classes” and “attributes” for some of the macros and probably should be a last resort. It means the calling context knows too much about the inner workings of exact markup/styles/behaviours that are imported from the design system for each component causing tight coupling. You wouldn’t be able to change the name of classes/attributes in the design system without changing all applications that integrate with it.
> * Would be good to see less hard-coded HTML removed to completely abstract presentation markup in future, kind of a given.

I think you've raised valid points, definitely worth us discussing in a tech session or design meeting.

 * I agree about the conditional mappings, I think the amount of mapping in templates should be minimal, and most of the mapping be done in python to create a view model for the templates. There are a lot of jinja filters I created to map things for components, but these should be run before the view is hit, rather that from the view. This is quite a big thing to fix and will likely need to be a separate piece of work.

* Structural components are something we're quite weak on in the design system, as really we've only really been working with components as units, and not actually any page structure (the anatomy of a question for example), so I think we can definitely improve this by creating more macros that are purely for structure. It's something we should discuss in a design meeting.

* You're right about classes and attribute only being a last resort. I think a lot of components rely on you manually applying a class such as `btn--secondary` to change it's presentation rather than that being abstracted to some other parameter that sets the correct classes to change the presentation of that component. Those parameters should only be used to add stuff like `data-ga` attributes etc, and utility classes. If you spot examples of those attributes being misused though, if you could let me know or place a comment against the relevant component in the [design system forum](https://github.com/ONSdigital/design-system/issues?q=is%3Aissue+is%3Aopen+label%3A%22Design+System+forum%22) we can get on top of those issues.

* I think we can reduce the amount of `HTML` hand coded in templates by creating more macros dedicated to structure, but considering a lot of content in runner is hard coded rather than populated from a CMS WYSIWYG, there will always be a small amount knocking around.

scripts/load_templates.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
```
bitdivision, line 1: We need to add a call to this in all our scripts and also in the readme as a setup step if you're running anything manually

bitdivision:
I'm hitting a few errors when running the integration tests:
`jinja2.exceptions.TemplateNotFound: _themes/default/sectionsummary.html`

Looks like there's still some setup code referencing themes. Do we need to remove all of this?

bameyrick:
> I'm hitting a few errors when running the integration tests:
> `jinja2.exceptions.TemplateNotFound: _themes/default/sectionsummary.html`
> 
> Looks like there's still some setup code referencing themes. Do we need to remove all of this?

I think so, removing all theming for now.

tests/functional/spec/components/checkbox/mutually_exclusive/mutually_exclusive_dropdown.spec.js
```diff
@@ -45,6 +45,7 @@ describe('Component: Mutually Exclusive Dropdown With Single Checkbox Override',
 
         // When
         .selectByValue(DropdownPage.dropdown(), 'Liverpool')
+        .click(DropdownPage.dropdownExclusiveIPreferNotToSayLabel())
```
bameyrick: By clicking the label you may be effectively clicking the checkbox so this might not be the correct fix

bameyrick: By clicking the label you may be effectively clicking the checkbox so this isn't a safe way of fixing this. I think `selectByValue` may need to be modified to trigger a change event.

scripts/load_templates.sh
```diff
@@ -0,0 +1,15 @@
...
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+cd "${DIR}"/.. || exit
+
+DESIGN_SYSTEM_VERSION="4.0.0"
+
+rm -rf /tmp/templates*
+curl -L --url "https://github.com/ONSdigital/design-system/releases/download/$DESIGN_SYSTEM_VERSION/templates.zip" --output /tmp/templates.zip
+unzip /tmp/templates.zip -d /tmp/templates
```
ajmaddaford: This should unzip to a randomly generated directory name as we're forcefully removing it and there is always a chance that another app/service is using a directory of the same name. It looks like the `mktemp` command would be a good option - https://stackoverflow.com/questions/4632028/how-to-create-a-temporary-directory

ajmaddaford (changes requested):
- We should move the templates directory to the top level i.e. at the same level as the app directory. They're currently the only non-Python content in the app folder.
- The design system dependencies should be brought in to the same folder to easily isolate and identify them from the other templates - something like /templates/design-system.

I think we should get this merged as is. We can look at generating the required view data before rendering the templates (rather than via Jinja filters called from the templates) after it's merged.

bitdivision:
When clicking the Contact Us link while viewing in Welsh, I'm running into a 500, looks like it's related to the brackets around ru_ref?

```  File "/Users/forter/Code/eq-survey-runner/app/../templates/static/contact-us.html", line 16, in block "main"
    <p>{{ _("Please quote the reference <b>%(ru_ref)s</b> when contacting ONS about this survey.", ru_ref=session.ru_ref) }}</p>
  File "/Users/forter/virtualenvs/eq-survey-runner-mBP_POI1/lib/python3.7/site-packages/jinja2/ext.py", line 134, in _gettext_alias
    return __context.call(__context.resolve('gettext'), *args, **kwargs)
  File "/Users/forter/virtualenvs/eq-survey-runner-mBP_POI1/lib/python3.7/site-packages/jinja2/ext.py", line 143, in gettext
    return rv % variables
ValueError: unsupported format character '(' (0x28) at index 27```

File: README.md [hunk 4]
```diff
@@ -70,6 +70,14 @@ If you want to run the app locally, but all other dependencies in docker, you ca
 docker-compose -f docker-compose-dev.yml up -d
 ```
 
+You will also need these environment variables to your bash profile:
+
+```
+export EQ_DATASTORE_EMULATOR_CREDENTIALS="True"
```
MebinAbraham, line 85: I know it was just an addition but these should go in env var section?
https://github.com/ONSdigital/eq-survey-runner/blob/347ec4f0ca2907a60682a59fca125ffce2249844/README.md#environment-variables

MebinAbraham (changes requested):

bitdivision:
padding_guidance

The padding around guidance doesn't look quite right to me.

Edit: In firefox

bitdivision:
padding
Another padding issue.

MebinAbraham (changes requested):
A lot of text is in bold for me in Chrome/Firefox on this page only. This is on the first page of the census individual schema.
Screenshot 2019-05-10 at 11 01 54
And footer on same page:
Screenshot 2019-05-10 at 11 07 36

bitdivision:
twisty

The main text on the twisty here isn't showing up:

twisty-open

MebinAbraham (commented):
Since all the static file compiling has gone, the readme needs to be updated. It still outlines things like yarn and compile etc.

bitdivision:
I'm not sure about this, but the thank you page looks a little more condensed than it used to.

thanks

I could be making this up though

EDIT: I think I'm making this up. Ignore this one

bitdivision:
On test_final_confirmation.json, the submit button is showing inline:
submit-answers

MebinAbraham (dismissed):
Answer option description alignment issue?
#### Before:

Screenshot 2019-05-10 at 11 19 34

#### After:
Screenshot 2019-05-10 at 11 20 40

MebinAbraham (commented):
The summary seems very padded. Also the Change link is missing as well as the Previous link?

#### Before:
Screenshot 2019-05-10 at 11 22 45

#### After:
Screenshot 2019-05-10 at 11 24 07

MebinAbraham (commented):
The print button not looking so good today :P
Screenshot 2019-05-10 at 11 27 51

MebinAbraham:
Not sure how to explain it but every time you load the definitions, you can see/notice the transition (flicker) from non-js to js version when the js scripts are loaded. You can see the page changing after most of it has already rendered in. Give me a shout if it doesn't make sense and I'll explain in person. Doesn't happen with the current version of v3.
Screenshot 2019-05-10 at 11 37 20

MebinAbraham:
For the summary and pages after summary ie, submission etc, the page content width has been increased, is that intentional?

bameyrick:
> Answer option description alignment issue?
> 
> #### Before:
> Screenshot 2019-05-10 at 11 19 34
> 
> #### After:
> Screenshot 2019-05-10 at 11 20 40

@will-madebymoon asked for this to be changed here: https://github.com/ONSdigital/design-system/issues/56#issuecomment-456444482

bameyrick:
> The summary seems very padded. Also the Change link is missing as well as the Previous link?
> 
> #### Before:
> Screenshot 2019-05-10 at 11 22 45
> 
> #### After:
> Screenshot 2019-05-10 at 11 24 07

I've fixed the missing change links, but I'm not sure what the padding on that summary should be... will have to ask @will-madebymoon

bameyrick:
> The print button not looking so good today :P
> Screenshot 2019-05-10 at 11 27 51

This will be fixed once this pr is merged and released: https://github.com/ONSdigital/design-system/pull/362

bameyrick:
> eady rendered in. Give me a shout if it doesn't make sense and I'll explain in person. Doesn't happen with the current version of v3.

Will need to take a look when we're both in, it shouldn't be slow enough to see the flicker... we can raise a bug against that component in the design system.

bameyrick:
> For the summary and pages after summary ie, submission etc, the page content width has been increased, is that intentional?

All pages should be using 8 columns now rather than 7

bameyrick:
> On test_final_confirmation.json, the submit button is showing inline:
> submit-answers

I'll fix this

templates/feedback.html
```diff
@@ -0,0 +1,58 @@
+{% extends 'layouts/_base.html' %}
+{% from "components/field/_macro.njk" import onsField %}
+{% from "components/textarea/_macro.njk" import onsTextarea %}
+{% from "components/input/_macro.njk" import onsInput %}
+{% from "components/button/_macro.njk" import onsButton %}
+
+{% set page_title  = _("Survey feedback") %}
+
+{% block main %}
+  <form id="feedback-form" method="POST">
```
MebinAbraham: There is no closing tag for `form` in this file.

MebinAbraham, line 10: There is no closing tag for `<form>` in this file.

ajmaddaford (approved):

bitdivision (approved):

bitdivision:
🎉 👏

bameyrick:
WholeNarrowAmurratsnake-size_restricted

PR #2046, opened 2019-04-12
Dont update session expiry too much

jonnyshaw89:
We are making a lot of writes to the session table that we potentially don't need to make.
When we create a new session from the `/session` login page, we create the session and then we make a request to update the expiry milliseconds later.
Also when we do a POST and redirect to a GET, we update the expiry on the POST and then again on the GET immediately after.

This PR introduces a change to not update the session expiry if it has been updated in the last 60 seconds

### How to review 
Does Session expiry continue to work as expected?

### DIFF
--- a/tests/app/authentication/test_authenticator.py
+++ b/tests/app/authentication/test_authenticator.py
@@ -94,8 +94,7 @@ class TestAuthenticator(AppContextTestCase): # pylint: disable=too-many-public-m
                 self.assertIsNone(user)
                 self.assertIsNone(cookie_session.get(USER_IK))
 
-
-    def test_valid_user_extends_session_expiry(self):
+    def test_valid_user_does_not_extend_session_expiry_when_expiry_less_than_60_seconds_different(self):
         with self.app_request_context('/status'):
             with patch('app.authentication.authenticator.get_session_store', return_value=self.session_store):
                 # Given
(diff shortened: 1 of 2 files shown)


tests/app/authentication/test_authenticator.py
```diff
@@ -94,8 +94,7 @@ def test_no_user_when_session_has_expired(self):
                 self.assertIsNone(user)
                 self.assertIsNone(cookie_session.get(USER_IK))
 
-
-    def test_valid_user_extends_session_expiry(self):
+    def test_valid_user_does_not_extend_session_expiry_when_time_close(self):
```
MebinAbraham:
```suggestion
    def test_valid_user_does_not_extend_session_expiry_when_expiry_less_than_60_seconds_different(self):
```
I don't think this really makes sense. Maybe like your other one?. I guess its less than or equal to though.

Or maybe
``` python
def test_valid_user_does_not_extend_session_expiry_when_expiry_not_greater_than_60_seconds_different(self):
```
Or even change the code to `>= 60`. 🤷‍♂️ 
Up to you.

MebinAbraham (changes requested):
Apart from the minor comments, have tested and seems to be working as expected. Should save a bunch of writes to db.  👍

tests/app/authentication/test_authenticator.py
```diff
@@ -110,8 +109,24 @@ def test_valid_user_extends_session_expiry(self):
                 self.assertEqual(user.user_id, 'user_id')
                 self.assertEqual(user.user_ik, 'user_ik')
                 self.assertEqual(user.is_authenticated, True)
-                self.assertGreater(self.session_store.expiration_time, self.expires_at)
+                self.assertEqual(self.session_store.expiration_time, self.expires_at)
+
+    def test_valid_user_extends_session_expiry_when_expiry_greater_then_60_seconds_different(self):
```
MebinAbraham:
```suggestion
    def test_valid_user_extends_session_expiry_when_expiry_greater_than_60_seconds_different(self):
```

MebinAbraham (approved):

bitdivision (approved):
:+1:

PR #2047, opened 2019-04-12
Remove support for SQL database

jonnyshaw89:
We are currently making a query on questionnaire state before writing changes.
This is done to support SQL database where we need to know if we are performing an INSERT or an UPDATE.
We no longer need to do this if we don't want to support SQL databases any more as the NoSQL databases that we are using don't need us to specify if we are inserting or updating.

This PR removes that check.

### How to review 
Does the saving of questionnaire state work as expected?


bitdivision:
If we're breaking support for SQL databases here, shouldn't this PR be removing all SQL database references?

jonnyshaw89:
@bitdivision If we decide that this is an approach we want to take then yeh I can do that.

nealedj:
Or do it on a feature flag? I implemented the same thing before - I thought I'd discussed it with you but it must have got lost in the other changes: https://github.com/ONSdigital/census-survey-runner/pull/11

Rationale for closing it in a feature flag was because for business there was a greater chance of two users interacting with the same questionnaire at the same time.

jonnyshaw89:
@nealedj Im not sure that this would have had much impact saving data when two users were interacting as it blatted over the data after it read anyway.

https://github.com/ONSdigital/eq-survey-runner/pull/2047/files#diff-91332e4f9b1ffb424d8ceb2168bc05aeL27

bitdivision:
I think this is the right approach. Continuing to support SQL databases is very prone to error if we're not actively using it in any projects. It shouldn't be hard to add back in if someone really needs it.

bitdivision (approved):
🎉

ajmaddaford (changes requested):
- The removal of elastic beanstalk related files should be in another pull request
- There are still some references to the database:
  - app/secrets.py
  - docker-secrets.yml
  - scripts/generate_secrets.py

ajmaddaford (commented):
Minor quibble, the commit message should be capitalised - "remove" -> "Remove"

ajmaddaford (approved):

bitdivision:
You forgot to setup GPG @jonnyshaw89

ajmaddaford (changes requested):
The `database-scripts` folder needs to be deleted

MebinAbraham (approved):

ajmaddaford (approved):

PR #2049, opened 2019-04-15
Clone jsonnet if Makefile not present in Travis cache

MebinAbraham:
Travis cache was cleared and jsonnet is not being installed because an empty directory is created by travis to store cache and the jsonnet is not cloned if the directory exists.
This ensures it is cloned if the Makefile in the dir is not present.

Ensure:
- Travis builds successfully?


MebinAbraham:
Fixed

jonnyshaw89 (approved):

bitdivision (approved):

PR #2050, opened 2019-04-16
Fix dates without day_of_month piping transform bug

bambrp:
This fixes the issue where a transform of `calculate_years_difference()` would fail on dates where only a year and month were supplied, such as 2016-02, whilst working for full dates of 2016-02-23.

Issue was a single date format string. This has been supplemented with a fallback string which is tried if the `strptime()` method throws a ValueError exception.

Converted method to static, as it does not rely on any instance state.

### How to review 
Run any test schema that passes just year and month (e.g. _test_difference_in_years_month_year.json_.
Observe the lack of error.

### DIFF
--- a/app/questionnaire/placeholder_transforms.py
+++ b/app/questionnaire/placeholder_transforms.py
@@ -1,6 +1,7 @@
 from datetime import datetime
 
 from dateutil.relativedelta import relativedelta
+from dateutil.tz import tzutc
 
 from babel.numbers import format_currency
 from babel.dates import format_datetime
@@ -19,6 +20,7 @@ class PlaceholderTransforms:
 
     locale = DEFAULT_LOCALE
     input_date_format = '%Y-%m-%d'
+    input_date_format_month_year_only = '%Y-%m'
 
     def format_currency(self, number=None, currency='GBP'):
         return format_currency(number, currency, locale=self.locale)
@@ -80,11 +82,29 @@ class PlaceholderTransforms:
 
         return ''
 
-    def calculate_years_difference(self, first_date, second_date):
-        first_date = datetime.now() if first_date == 'now' else datetime.strptime(first_date, self.input_date_format)
-        second_date = datetime.now() if second_date == 'now' else datetime.strptime(second_date, self.input_date_format)
+    @staticmethod
+    def calculate_years_difference(first_date, second_date):
+        return str(relativedelta(PlaceholderTransforms.parse_date(second_date),
+                                 PlaceholderTransforms.parse_date(first_date)).years)
 
-        return str(relativedelta(second_date, first_date).years)
+    @staticmethod
+    def parse_date(date):
+        """
+        :param date: string representing a date
+        :return: datetime of that date
+
+        Convert `date` from string into `datetime` object. `date` can be 'YYYY-MM-DD', 'YYYY-MM'
+        or 'now'. Note that in the shorthand YYYY-MM format, day_of_month is assumed to be 1.
+        """
+        if date == 'now':
+            return datetime.now(tz=tzutc())
+
+        try:
+            return datetime.strptime(date, PlaceholderTransforms.input_date_format)\
+                .replace(tzinfo=tzutc())
+        except ValueError:
+            return datetime.strptime(date, PlaceholderTransforms.input_date_format_month_year_only)\
+                .replace(tzinfo=tzutc())
 
     def first_non_empty_item(self, items):
         """
--- a/tests/app/questionnaire/test_placeholder_transforms.py
+++ b/tests/app/questionnaire/test_placeholder_transforms.py
@@ -56,9 +56,12 @@ class TestPlaceholderParser(unittest.TestCase):
         assert welsh_transforms.format_possessive('Alice Aardvark’s') == 'Alice Aardvark’s'
 
     def test_calculate_years_difference(self):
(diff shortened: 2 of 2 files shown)


app/questionnaire/placeholder_transforms.py
```diff
@@ -19,6 +19,7 @@ def __init__(self, language):
 
     locale = DEFAULT_LOCALE
     input_date_format = '%Y-%m-%d'
+    input_date_format_md_only = '%Y-%m'
```
iwootten: Can we use a variable name here that isn't abbreviated?

app/questionnaire/placeholder_transforms.py
```diff
@@ -80,11 +81,27 @@ def format_number(number):
...
+        """
+        :param date: string representing a date
+        :return: datetime of that date
+
+        Convert `date` from string into `datetime` object. `date` can be 'YYYY-MM-DD', 'YYYY-MM'
+        or 'now'. Note that in the shorthand YYYY-MM format, day_of_month is assumed to be 1.
+        """
+        if date == 'now':
+            return datetime.now()
+        else:
```
iwootten: Travis is complaining here that the else is unnecessary.

app/questionnaire/placeholder_transforms.py
```diff
@@ -80,11 +81,27 @@ def format_number(number):
...
+        or 'now'. Note that in the shorthand YYYY-MM format, day_of_month is assumed to be 1.
+        """
+        if date == 'now':
+            return datetime.now()
 
-        return str(relativedelta(second_date, first_date).years)
+        try:
+            return datetime.strptime(date, PlaceholderTransforms.input_date_format)
+        except ValueError:
+            return datetime.strptime(date, PlaceholderTransforms.input_date_format_md_only)
```
iwootten: I wonder if inspecting the input date string and checking if it matches a known format would be better than falling back here?

iwootten (changes requested):

app/questionnaire/placeholder_transforms.py
```diff
@@ -80,11 +81,27 @@ def format_number(number):
...
+    def parse_date(date):
+        """
+        :param date: string representing a date
+        :return: datetime of that date
+
+        Convert `date` from string into `datetime` object. `date` can be 'YYYY-MM-DD', 'YYYY-MM'
+        or 'now'. Note that in the shorthand YYYY-MM format, day_of_month is assumed to be 1.
+        """
+        if date == 'now':
+            return datetime.now()
```
bitdivision: Should this be UTC?

bitdivision (changes requested):

app/questionnaire/placeholder_transforms.py
bambrp: I didn't want to change the behaviour any more than necessary to fix the bug, but if you chaps would like to do that, I can, certainly.

app/questionnaire/placeholder_transforms.py
bambrp: Oh you python types and your love of widescreen code. If you prefer it, I'll do that.

app/questionnaire/placeholder_transforms.py
bambrp: As before/later (depending on the order in which the comments are read), I wasn't wanting to tread on anyone's toes and change the way it worked more than I had to. I had considered a regexp, or a count of dashes. I was surprised there wasn't an equivalent to a tryParse() method. You have any preference?

app/questionnaire/placeholder_transforms.py
bitdivision: Generally the style guide in runner is to avoid abbreviations if at all possible.

app/questionnaire/placeholder_transforms.py
bitdivision: This should be UTC AFAICS. I can't see any reason to work with local time on the server side.

It's likely never going to cause a major issue since we're not working with times, but probably worth cleaning up while here.

bambrp: Hmm, given that it is date only, how does TZ come into play? And given that we are not aware of the TZ of the other date provided, it wouldn't be enough to change the outcome even if we were asking for the time.

bitdivision: `datetime.now()` returns a local datetime, so if it's around midnight, the timezone can cause a change to the date.

Agree that the timezone of the other date is not defined but given that this is a generic method, I think it should assume UTC to keep things consistent rather than using local time which can vary depending on the server this runs on.

app/questionnaire/placeholder_transforms.py
iwootten: I should think whether it matches a regex for either date format would be enough.

bambrp: Consensus seems to be that it isn't necessary, but I won't be offended if you add that in.

bitdivision (approved):
:+1:

iwootten (approved):
Happy to go along with the consensus if others don't think checking format is necessary.

bitdivision:
@iwootten @bambrp I think that consensus is just me? I quite like the exception catching, and I don't really see much advantage to the regex, apart from a possible slight performance gain.

bitdivision:
@bambrp You'll need to setup gpg for signed commits.

We tend to squash before merge too, which needs to be done locally for signed commits.

`git fetch; git rebase -i origin/v3`

PR #2052, opened 2019-04-23
Add a tracing decorator

jonnyshaw89:
This PR adds a tracing decorator that we can link to multiple tracing backends.
Currently only AWS X-Ray is configured but more can be added for each cloud provider

### How to review 
Run in AWS with `AWS_XRAY_SDK_ENABLED` enabled and deployed with https://github.com/ONSdigital/eq-ecs-deploy/pull/24
You should see traces in X-Ray


bitdivision:
Needs a readme change for the new env var.

tests/integration/test_tracing.py
```diff
@@ -0,0 +1,42 @@
...
+    def setUp(self, setting_overrides=None):
+
+        self.begin_subsegment = patch('aws_xray_sdk.core.xray_recorder.begin_subsegment')
+        self.begin_subsegment_mock = self.begin_subsegment.start()
+        self.end_subsegment = patch('aws_xray_sdk.core.xray_recorder.end_subsegment')
+        self.end_subsegment_mock = self.end_subsegment.start()
+        self.XRayMiddleware = patch('aws_xray_sdk.ext.flask.middleware.XRayMiddleware', Mock())
+        self.XRayMiddleware_mock = self.XRayMiddleware.start()
+        self.patch_all = patch('aws_xray_sdk.core.patch_all', Mock())
+        self.patch_all_mock = self.patch_all.start()
```
bitdivision, line 20: I wonder if we could use the patch decorator to clean these up?

app/tracing.py
```diff
@@ -0,0 +1,19 @@
+import wrapt
+from aws_xray_sdk.core import xray_recorder
+from flask import current_app
+
+
+def trace():
+    @wrapt.decorator
```
bitdivision, line 7: What is the advantage of this library over `functools.wraps` etc

tests/integration/components/mutually_exclusive/test_checkbox_single_checkbox_override.py
```diff
@@ -7,7 +7,7 @@ class TestCheckboxSingleCheckboxOverride(IntegrationTestCase):
     with single checkbox override function as expected. These tests emulate the non-JS version.
     """
 
-    def setUp(self):
+    def setUp(self, setting_overrides=None):
```
bitdivision, line 10: I can't see why the new argument is needed?

bitdivision (changes requested):

tests/integration/components/mutually_exclusive/test_checkbox_single_checkbox_override.py
jonnyshaw89: pylint moans

bitdivision: about what?

tests/integration/test_tracing.py
jonnyshaw89: yes probably, would look cleaner

app/tracing.py
jonnyshaw89: its what was used within the collector for `xray_recorder`, i can switch it to `functools.wraps` i dont really know which is better

tests/integration/components/mutually_exclusive/test_checkbox_single_checkbox_override.py
bitdivision: ```tests/integration/components/mutually_exclusive/test_currency_single_checkbox_override.py:11:4: W0221: Parameters differ from overridden 'setUp' method (arguments-differ)```

app/tracing.py
bitdivision: It looks like wrapt does the same stuff but better, I just wasn't sure whether something broke if you used the standard one. 

I don't mind leaving it in.

tests/integration/components/mutually_exclusive/test_checkbox_single_checkbox_override.py
bitdivision: If we're adding an argument for it, then it should be passed to the super().

iwootten:
I thought much of the reason for attempting to use opencensus was to prevent being bound to a single tracing service? Have we made a decision to implement tracing without it now?

If that's the case, shouldn't Stackdriver tracing be implemented first given it's the provider of choice for the census version of runner?

I'd be interested in knowing what the performance overhead of adding x-ray without opencensus is.

jonnyshaw89:
@iwootten I guess we should have started with direct to stackdriver, the reason I raised this is because i had it working locally and thought it might add some value if we had an abstraction around the tracing to try multiple things.

If people don't think it's valuable then we can bin it?

iwootten:
@jonnyshaw89 Having something to compare the overhead that tracing adds, be it opencensus or otherwise seems valuable to me - I guess we need to decide whether that should be a one off branch or something that exists in v3.

bitdivision:
@iwootten I feel like having some kind of tracing available is still useful even if we decide not to use it in production at the moment.

jonnyshaw89:
@ajmaddaford Thoughts?

jonnyshaw89:
Closing in favour of looking at OpenCensus Agent https://opencensus.io/service/components/agent/

jonnyshaw89:
Closing in favour of looking at OpenCensus Agent https://opencensus.io/service/components/agent/

Issue #1809, opened 2018-10-03
Submitting feedback after submission errors


jonnyshaw89:
### Expected behaviour
Open question...
Should this same form allow feedback to be given after submission?

### Actual behaviour
Errors are generated
```
"exception": [
        "Traceback (most recent call last):",
        "  File '/opt/python/run/venv/local/lib64/python3.4/site-packages/flask/app.py', line 1813, in full_dispatch_request",
        "    rv = self.dispatch_request()",
        "  File '/opt/python/run/venv/local/lib64/python3.4/site-packages/flask/app.py', line 1799, in dispatch_request",
        "    return self.view_functions[rule.endpoint](**req.view_args)",
        "  File '/opt/python/run/venv/local/lib64/python3.4/site-packages/flask_login/utils.py', line 261, in decorated_view",
        "    return func(*args, **kwargs)",
        "  File '/opt/python/current/app/app/views/feedback.py', line 64, in send_feedback",
        "    g.schema.json['survey_id'],",
        "  File '/opt/python/current/app/app/submitter/converter.py', line 131, in convert_feedback",
        "    'tx_id': metadata['tx_id'],",
        "KeyError: 'tx_id'"
    ],
```
```
"exception": [
        "Traceback (most recent call last):",
        "  File '/opt/python/run/venv/local/lib64/python3.4/site-packages/flask/app.py', line 1813, in full_dispatch_request",
        "    rv = self.dispatch_request()",
        "  File '/opt/python/run/venv/local/lib64/python3.4/site-packages/flask/app.py', line 1799, in dispatch_request",
        "    return self.view_functions[rule.endpoint](**req.view_args)",
        "  File '/opt/python/run/venv/local/lib64/python3.4/site-packages/flask_login/utils.py', line 261, in decorated_view",
        "    return func(*args, **kwargs)",
        "  File '/opt/python/current/app/app/views/feedback.py', line 46, in get_form",
        "    return _render_template('feedback', content=content)",
        "  File '/opt/python/current/app/app/helpers/template_helper.py', line 31, in session_wrapper",
        "    **kwargs",
        "  File '/opt/python/current/app/app/helpers/template_helper.py', line 51, in analytics_wrapper",
        "    return func(*args, analytics_ua_id=current_app.config['EQ_UA_ID'], **kwargs)",
        "  File '/opt/python/current/app/app/helpers/template_helper.py', line 41, in metadata_context_wrapper",
[...]

PR #2056, opened 2019-04-29
Fix error when feedback form submitted after questionnaire submitted.

pricea1:
…mitted and session ended

Submission of feedback form gives error when submitted after questionnaire form has been submitted

### How to review 
To reproduce on Master:
1. Open survey in runner
2. Open Feedback form in new tab (right click feedback link > open in new tab)
3. Complete & submit questionnaire
4. Submit feedback form and observe error being reported

On branch, repeat above steps without error.

### Checklist

### DIFF
--- a/app/views/feedback.py
+++ b/app/views/feedback.py
@@ -12,6 +12,7 @@ from app.submitter.submission_failed import SubmissionFailedException
 from app.globals import get_metadata, get_session_store
 from app.submitter.converter import convert_feedback
 from app.utilities.schema import load_schema_from_session_data
+from app.authentication.no_token_exception import NoTokenException
 
 logger = get_logger()
 
@@ -53,12 +54,13 @@ def send_feedback():
     if 'action[sign_out]' in request.form:
         return redirect(url_for('session.get_sign_out'))
 
+    metadata = get_metadata(current_user)
+    if not metadata:
+        raise NoTokenException(401)
+
     form = FeedbackForm()
 
     if form.validate():
-
-        metadata = get_metadata(current_user)
-
         message = convert_feedback(
             escape(request.form.get('message')),
             escape(request.form.get('name')),
--- a/app/helpers/template_helper.py
+++ b/app/helpers/template_helper.py
@@ -8,6 +8,7 @@ from structlog import get_logger
 from app.globals import get_metadata, get_session_timeout_in_seconds
 from app.templating.metadata_context import build_metadata_context
 from app.templating.template_renderer import TemplateRenderer
+from app.authentication.no_token_exception import NoTokenException
 
 logger = get_logger()
 
@@ -30,6 +31,9 @@ def with_metadata_context(func):
     @wraps(wraps)
     def metadata_context_wrapper(*args, **kwargs):
         metadata = get_metadata(current_user)
+        if not metadata:
+            raise NoTokenException(401)
+
         metadata_context = build_metadata_context(metadata)
 
         return func(*args, metadata_context=metadata_context, **kwargs)
--- a/tests/integration/views/test_feedback.py
+++ b/tests/integration/views/test_feedback.py
@@ -14,6 +14,8 @@ FEEDBACK_FORM_URL = '/feedback'
 FEEDBACK_THANKYOU_URL = '/feedback/thank-you'
 SIGNED_OUT_URL = '/signed-out'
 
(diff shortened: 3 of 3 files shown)


pricea1:
[500 error on feedback form](https://trello.com/c/Ep2bekXf/997-500-error-on-feedback-form)

SamGodwin2 (approved):
Tested and works for me.

MebinAbraham:
@pricea1 Has a decision been made regarding if a user should be able to submit the feedback form without a valid session?. Because without a valid session, someone could essentially submit thousands of submissions. 

I think the user should be redirected to the session expiry page when they don't have a valid reason. Also without the session data the business would not be able to relate the submission to a specific collex/transaction.

Thoughts?

MebinAbraham:
Fixes https://github.com/ONSdigital/eq-survey-runner/issues/1809

pricea1:
> @pricea1 Has a decision been made regarding if a user should be able to submit the feedback form without a valid session?. Because without a valid session, someone could essentially submit thousands of submissions.
> 
> I think the user should be redirected to the session expiry page when they don't have a valid reason. Also without the session data the business would not be able to relate the submission to a specific collex/transaction.
> 
> Thoughts?

No decision as far as I know. @samiwel - who would decide the expected behaviour? It would be straightforward to redirect instead of submitting the form if session has expired.

app/views/feedback.py
```diff
@@ -59,22 +59,23 @@ def send_feedback():
...
-        )
-
-        encrypted_message = encrypt(message, current_app.eq['key_store'], key_purpose=KEY_PURPOSE_SUBMISSION)
-        sent = current_app.eq['submitter'].send_message(encrypted_message,
-                                                        current_app.config['EQ_RABBITMQ_QUEUE_NAME'],
-                                                        metadata.get('tx_id'))
-
-        if not sent:
-            raise SubmissionFailedException()
+        if metadata:
```
samiwel: I wonder if the safest thing to do here would be to throw a `NoTokenException` when there is no metadata in the same way as I did in my last commit?

There's potential with this change for the user to be redirected to the thank you page even when the feedback is not submitted, which might be a bit confusing.

samiwel (changes requested):

app/views/feedback.py
MebinAbraham: Personally I don't think this is necessary. So with this change, if you post a feedback without filling any fields and without a valid session, this would evaluate line 58:
https://github.com/ONSdigital/eq-survey-runner/blob/8906c8219c69ad85cc347f7908891b7a4563fd56/app/views/feedback.py#L58

Then line 80:
https://github.com/ONSdigital/eq-survey-runner/blob/8906c8219c69ad85cc347f7908891b7a4563fd56/app/views/feedback.py#L80-L81

This would then redirect you to thank you page, from which the decorator `with_metadata_context` is called on the GET request, then on line 34, 35 of template helper it would evaluate false and raise a no token exception.
https://github.com/ONSdigital/eq-survey-runner/blob/8906c8219c69ad85cc347f7908891b7a4563fd56/app/helpers/template_helper.py#L34-L35

---

However, we could move line 60:
https://github.com/ONSdigital/eq-survey-runner/blob/8906c8219c69ad85cc347f7908891b7a4563fd56/app/views/feedback.py#L60 above the`if form.validate()` call and fail early to avoid the extra evaluations. So above form.validate we could do:

```python
if not metadata:
    raise NoTokenException(401)
```

Hope that is clear, let me know if I've made no sense.

pricea1: It depends on what we want to display to the user. The current method redirects to the thank you page which then displays the session expired message. Raising the 401 in feedback.py displays the generic "Error 500 - something has gone wrong" message.

Personally I think the session expired message is clearer.

MebinAbraham: Raising a 401 will return a session expired page as it has a custom handler.
https://github.com/ONSdigital/eq-survey-runner/blob/8906c8219c69ad85cc347f7908891b7a4563fd56/app/views/errors.py#L25-L29

pricea1: Sorry, you're right. I hadn't imported NoTokenException when I initially tested. I've updated so the exception is raised before form validation.

samiwel (approved):

MebinAbraham (approved):

PR #2057, opened 2019-05-01
Reduce size of html payload

jonnyshaw89:
The PR compresses and minifies the html returned from runner.
There is a significant reduction in request size.

### How to review 
Check that the html response from runner is compresses and minified.


nealedj:
I'm sure it's negligible but it might be worth considering if this has much of an impact on the time it takes the application to process each request.

iwootten:
Are you able to give an indication on the size of the reduction?

jonnyshaw89:
@iwootten IIRC one page went from about 20kb to 4kb

app/setup.py
```diff
@@ -178,6 +183,19 @@ def apply_caching(response):  # pylint: disable=unused-variable
 
         return response
 
+    @application.after_request
+    def response_minify(response):  # pylint: disable=unused-variable
```
MebinAbraham, line 186: Worth adding a test to ensure this is working.
Have tested manually and looked good.

app/setup.py
```diff
@@ -178,6 +183,19 @@ def apply_caching(response):  # pylint: disable=unused-variable
...
+        """
+        minify html response to decrease site traffic
```
MebinAbraham, line 188:
```suggestion
        Minify HTML response to reduce the request size
```

app/setup.py
jonnyshaw89: What would the test include? checking that there are no spaces? is there much value there?

MebinAbraham: We could check the `content-length` in the header request is smaller when the minify is enabled?. And `Content-Encoding: gzip` if that is available?.

bitdivision: We would essentially be testing a configuration flag here?

MebinAbraham: Which we already do?.

jonnyshaw89: Also the current testing harness wont support binary responses like we would get with gzip enabled. to support this we would need to re-write the testing harness, which seems like quite an overhead

MebinAbraham: Had a quick fiddle and in the testing, the content-length header is available which is smaller when the minify is set to true. But to explicitly check `gzip` maybe not.

MebinAbraham (approved):

bitdivision (approved):
Tested this with profiling turned on and got 0.003136 total time in compress over 42 requests. 0.00007466666 seconds per request.

Haven't tested minification with profiling

PR #2058, opened 2019-05-01
Ensure answer and match value is valid before evaluating rules

MebinAbraham:
Was causing error on `census_individual.json` when performing an `in` operation against a `None` value because the question has not been answered yet.

### How to review 
1. Ensure you can get past the dob question on the `census_individual.json`


pricem14pc (approved):
Retest this and I'm now able to progress past the dob question and complete/submit census_individual

ajmaddaford (approved):

PR #2059, opened 2019-05-01
Build schemas based on region

MebinAbraham:
Removed routing based on `region_code` instead, the schema now build only the relevant blocks and rules for the specified regions in `build_schemas.sh`

### How to review 
1. Ensure a schema per region is built
2. The survey still functions as expected for each region.

Some of the additions I have done could potentially be made into a method in a separate file for reuse, as at the moment, the region-specific code is just implemented inline in the block files. Would be good to hear thought on making it any simpler/clearer.
Also, the `if` conditions currently only takes into account Wales/England, but will need altering when we have NI.

### DIFF
--- a/data-source/blocks/individual/identity-and-health/arrive_in_uk.jsonnet
+++ b/data-source/blocks/individual/identity-and-health/arrive_in_uk.jsonnet
@@ -23,7 +23,7 @@ local proxyTitle = {
   ],
 };
 
-{
+function(region_code, census_date) {
   type: 'Question',
   id: 'arrive-in-uk',
   question_variants: [
@@ -45,7 +45,7 @@ local proxyTitle = {
             id: 'arrive-in-uk-answer',
             condition: 'greater than',
             date_comparison: {
-              value: std.extVar('census_date'),
+              value: census_date,
               offset_by: {
                 years: -1,
               },
@@ -62,7 +62,7 @@ local proxyTitle = {
             id: 'arrive-in-uk-answer',
             condition: 'equals',
             date_comparison: {
-              value: std.extVar('census_date'),
+              value: census_date,
               offset_by: {
                 years: -1,
               },
@@ -73,13 +73,7 @@ local proxyTitle = {
     },
     {
       goto: {
-        block: 'understand-welsh',
-        when: [rules.regionWales],
-      },
-    },
-    {
-      goto: {
-        block: 'language',
+        block: if region_code == 'GB-WLS' then 'understand-welsh' else 'language',
       },
     },
   ],
--- a/data-source/census_individual.jsonnet
+++ b/data-source/census_individual.jsonnet
@@ -79,7 +79,9 @@ local comments = import 'blocks/comments.json';
 
 local confirmation = import 'blocks/confirmation.json';
 
-{
+local understandWelshBlock(region_code) = if region_code == 'GB-WLS' then [understand_welsh] else [];
+
+function(region_code, census_date) {
   mime_type: 'application/json/ons/eq',
   schema_version: '0.0.1',
   data_version: '0.0.3',
@@ -89,7 +91,7 @@ local confirmation = import 'blocks/confirmation.json';
   theme: 'census',
   legal_basis: 'Voluntary',
   eq_id: 'census',
-  form_type: 'household',
+  form_type: if region_code == 'GB-WLS' then 'individual_gb_wls' else 'individual_gb_eng',
   navigation: {
     visible: false,
   },
@@ -100,10 +102,6 @@ local confirmation = import 'blocks/confirmation.json';
     },
     {
       name: 'period_id',
-      validator: 'string',
-    },
-    {
-      name: 'region_code',
       validator: 'string',
     },
     {
@@ -136,7 +134,7 @@ local confirmation = import 'blocks/confirmation.json';
             date_of_birth,
             confirm_dob,
             sex,
-            marriage_type,
+            marriage_type(census_date),
             current_marriage_status,
             previous_marriage_status,
             current_partnership_status,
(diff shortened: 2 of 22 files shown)


data-source/census_individual.jsonnet
```diff
@@ -154,21 +152,21 @@ local confirmation = import 'blocks/confirmation.json';
...
-            arrive_in_uk,
-            when_arrive_in_uk,
-            length_of_stay,
-            understand_welsh,
-            language,
+            country_of_birth(region_code),
+            arrive_in_uk(region_code),
+            when_arrive_in_uk(region_code),
+            length_of_stay(region_code),
+          ] + understandWelshBlock(region_code) + [
```
MebinAbraham, line 159: Can't tell if I am happy with this or not.

data-source/blocks/individual/identity-and-health/arrive_in_uk.jsonnet
```diff
@@ -23,7 +23,49 @@ local proxyTitle = {
...
+    },
+  },
+  {
+    goto: {
+      block: if region_code == 'GB-WLS' then 'understand-welsh' else 'language',
+    },
+  },
+];
+
+function(region_code) {
```
ajmaddaford: If region code was used in the same way as census_date this wouldn't need to be a function. What are the advantages of converting it to a function?

data-source/blocks/individual/identity-and-health/arrive_in_uk.jsonnet
```diff
@@ -36,51 +78,5 @@ local proxyTitle = {
...
-        when: [rules.regionWales],
-      },
-    },
-    {
-      goto: {
-        block: 'language',
-      },
-    },
-  ],
+  routing_rules: routingRules(region_code),
```
ajmaddaford: Why has this been changed to a method call? Separating the routing rules out makes it harder to read.

data-source/blocks/individual/identity-and-health/arrive_in_uk.jsonnet
MebinAbraham: Thought it would suit better to have a top level argument that is passed to only the places we need rather than having global var, but ya could also be done using `std.extVar`.

According to Jsonnet:
>Generally, top-level arguments are the safer and easier way to parameterize an entire config, because the variables are not global and it's clear what parts of the config are dependent on their environment.

data-source/blocks/individual/identity-and-health/arrive_in_uk.jsonnet
MebinAbraham: Think I initially changed it cause it thought it would have been more than just an `if` but you're right, I have changed these to be inline.

data-source/blocks/individual/identity-and-health/arrive_in_uk.jsonnet
ajmaddaford: Yeah, it is nice to make it clear what blocks are dependent on region code. I think we should change the census date to use the same thing so that they're consistent.

data-source/blocks/individual/identity-and-health/arrive_in_uk.jsonnet
ajmaddaford: The same thing applies to a lot of other places where question variants have been split out into a method rather than kept inline.

data-source/census_individual.jsonnet
ajmaddaford: It's not pretty, but I can't think of anything better.

data-source/blocks/individual/identity-and-health/arrive_in_uk.jsonnet
MebinAbraham: So have changed census date to use the same thing but requires passing census_date into proxy titles etc, have a look and see.

data-source/blocks/individual/identity-and-health/arrive_in_uk.jsonnet
MebinAbraham: Has been changed now.

data-source/blocks/individual/identity-and-health/arrive_in_uk.jsonnet
ajmaddaford: I think that's fine, we should look at moving the proxy text choice into the question in another pr.

ajmaddaford (dismissed):
Nice work. The top level arguments work much better than the external variables.

MebinAbraham:
Since Phil's changes, there are a bunch of new ones added that I need to port over. Working through them now.

bitdivision:
Tested this with profiling turned on and got 0.003136 total time in compress over 42 requests. 0.00007466666.

Haven't tested minification with profiling

ajmaddaford (approved):

scripts/build_schemas.sh
```diff
@@ -1,9 +1,23 @@
...
+
+if [[ -x "$(command -v jsonnet)" ]]; then
+    # Build schemas for each region
+    for region_code in GB-WLS GB-ENG; do
+        SOURCE_FILE="data-source/census_individual.jsonnet"
+        # Lowercase the region code and replace '-' with '_'
+        FORMATTED_REGION_CODE=$(echo "${region_code}" | tr '[:upper:]' '[:lower:]' | tr - _)
+        DESTINATION_FILE="data-source/census_individual_${FORMATTED_REGION_CODE}.json"
+
+        jsonnet --tla-str region_code=${region_code} --tla-str census_date="2019-09-01" "${SOURCE_FILE}" > "${DESTINATION_FILE}"
```
bitdivision, line 11: Currently this will not generate unique eq_id / form_type per region.

We're getting rid of eq_id / form_type, so until then quick_launch won't work.

bitdivision (approved):
I've looked through it all and can't see any issues.

I don't think the form_type is a problem at the moment, since it's about to be replaced.

iwootten:
This seems to be out of sync with what the readme says - pipenv run ./scripts/build_schemas.sh fails for me due to having a dependency on a system version of gulp.

scripts/build_schemas.sh
MebinAbraham: Have based the form type on region for now.

MebinAbraham:
@iwootten Gulp was introduced before this PR. I assume you are just missing it. Try https://github.com/ONSdigital/eq-survey-runner/tree/v3#front-end-toolkit

iwootten:
> @iwootten Gulp was not introduced before this PR. I assume you are just missing it. Try https://github.com/ONSdigital/eq-survey-runner/tree/v3#front-end-toolkit

Following the docs yields this behaviour, do we require gulp systemwide for local dev?

bitdivision (approved):

MebinAbraham:
@iwootten As discussed, have updated the script to use `yarn` installed `gulp` and not look for system-wide `gulp`.

bitdivision (changes requested):
The form_type should be `individual` not `household`

MebinAbraham:
@bitdivision Sorted, good spot.

bitdivision (approved):

PR #2067, opened 2019-05-08
Fix issue with census individual main job block page title

MebinAbraham:
Requested by Charlotte to fix the block title not being the H1 on the page.
Since the block title is not part of variants, I have duplicated the block.

### How to review 
- Ensure title for `main-employment` and `last main-employment` blocks are now be wrapped in H1 tags.
- Routing works as expected.

### DIFF
--- a/data-source/blocks/individual/employment/ever_worked.jsonnet
+++ b/data-source/blocks/individual/employment/ever_worked.jsonnet
@@ -89,7 +89,7 @@ local proxyLabel = 'No, has never worked';
     },
     {
       goto: {
-        block: 'main-employment-block',
+        block: 'last-main-employment-block',
       },
     },
   ],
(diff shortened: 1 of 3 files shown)


ajmaddaford (changes requested):
Why the rush to fix this? We should fix this properly and move the title into the variants.

MebinAbraham:
Well the block title wouldn't be in variants?. Do you mean to use a title key in variants and use that in the html file instead of the block title?. Not urgent so doesn't need to go in now.

ajmaddaford:
> Do you mean to use a title key in variants and use that in the html file instead of the block title?

Yes.

Discussed with @MebinAbraham, closing this so we can implement it properly.

MebinAbraham:
https://trello.com/c/GAN8QJzl/2992-page-title-variants

PR #2069, opened 2019-05-08
Generate user_id and user_ik using the response_id

bitdivision:
See [trello](https://trello.com/c/RPDhFERZ/2981-jwt-claims-generate-userid-userik-from-responseid).

Previously we used a combination of `collection_exercise_sid, eq_id, form_type, ru_ref` to create the `user_id` and `user_ik`. Upstream will now be sending a `response_id` which is to replace the above fields during generation of `user_id` / `user_ik`.

I have made the `response_id` field mandatory on the assumption that upstream are currently sending it.

### How to review 

This PR can be tested in combination with the [go-launcher PR](https://github.com/ONSdigital/go-launch-a-survey/pull/134) and [benchmark PR](https://github.com/ONSdigital/eq-survey-runner-benchmark/pull/4/files).

### Checklist


MebinAbraham (approved):
Since this is mandatory, we should add it to?
https://github.com/ONSdigital/eq-survey-runner/blob/v3/app/storage/metadata_parser.py#L52-L77

We should also add a validator for it?.

MebinAbraham (changes requested):
Wrong button ☝️

bitdivision:
I have added a validator for numeric strings, with a length parameter.

app/authentication/user_id_generator.py
```diff
@@ -28,38 +28,35 @@ def generate_id(self, metadata):
         if metadata is None:
             raise ValueError('metadata is required')
 
-        collection_exercise_sid, eq_id, form_type, ru_ref = UserIDGenerator._get_token_data(metadata)
-        logger.debug('generating user id', ru_ref=ru_ref, ce_id=collection_exercise_sid, eq_id=eq_id, form_type=form_type)
+        token_data = UserIDGenerator._get_token_data(metadata)
```
ajmaddaford: Now that `response_id` is the only field we're using from the metadata, we should change the `generate_id` and `generate_ik` methods to accept that as a parameter rather than all of the metadata.

app/authentication/user_id_generator.py
```diff
@@ -28,38 +28,35 @@ def generate_id(self, metadata):
...
 
-        if ru_ref and collection_exercise_sid and eq_id and form_type:
-            return collection_exercise_sid, eq_id, form_type, ru_ref
+        response_id = metadata.get('response_id')
 
-        logger.error('missing values for ru_ref, collection_exercise_sid, form_type or eq_id', metadata=metadata)
-        raise InvalidTokenException('Missing values in JWT token')
+        if not response_id:
+            logger.error('Missing value for response_id in metadata', metadata=metadata)
+            raise InvalidTokenException('Missing value for response_id in JWT token')
```
ajmaddaford: Does this ever happen given the claims are validated in `metadata_parser.py`?

app/storage/metadata_parser.py
```diff
@@ -26,6 +26,24 @@ def string_parser(string_value):
         raise TypeError('Claim not a valid string value')
     return string_value
 
+def numeric_string_parser(to_parse, length=None):
```
MebinAbraham: To keep it consistent with the other method?
`to_parse` > `string_value`?, or `string_to_parse`.

app/utilities/strings.py
```diff
@@ -18,3 +22,7 @@ def to_str(bytes_or_str):
     if isinstance(bytes_or_str, bytes):
         return bytes_or_str.decode()
     return bytes_or_str
+
+
+def generate_numeric_string(length=16):
```
ajmaddaford: I don't think this needs to be factored out into it's own method in another file as this is only used in one place.

tests/app/authentication/test_user_id_generator.py
```diff
@@ -12,99 +12,57 @@ def setUp(self):
...
-        user_id_2 = id_generator.generate_id(self.create_token('1', '2', '3', '4'))
-        user_id_3 = id_generator.generate_id(self.create_token('1', '2', '4', '4'))
-        user_id_4 = id_generator.generate_id(self.create_token('2', '2', '3', '4'))
-        user_id_5 = id_generator.generate_id(self.create_token('1', '1', '3', '4'))
-        user_id_6 = id_generator.generate_id(self.create_token('2', '2', '4', '4'))
-        user_id_7 = id_generator.generate_id(self.create_token('2', '2', '4', '5'))
-        user_id_8 = id_generator.generate_id(self.create_token('1', '2', '3', '5'))
+        user_id_1 = id_generator.generate_id(self.create_token('1234567890123456'))
+        user_id_2 = id_generator.generate_id(self.create_token('1234567890123456'))
+        user_id_3 = id_generator.generate_id(self.create_token('0000000000000000'))
```
ajmaddaford: The `create_token` method could be removed if `generate_id` was refactored as previously mentioned.

ajmaddaford (changes requested):
I think we should be more flexible in what we allow in response id. The current requirements are for a numeric string, but other clients may want to use something like a uuid. The main concern is ensuring it is long enough as we use it to generate the user id and ik. For this, the length validation is appropriate, but I'd consider deferring that work into another card for replacing the current implementation with something like Marshmallow.

MebinAbraham (commented):
The `response_id` should be in the submission payload?

ajmaddaford (approved):

MebinAbraham (approved):

PR #2070, opened 2019-05-08
Add proposal for lookups

iwootten:

This proposes how lookups might be represented in a survey schema. 

### How to review 

- See if you think the proposal makes sense.
- Is there anything I've missed?

### DIFF
--- a/doc/architecture/decisions/0008-lookups.md
+++ b/doc/architecture/decisions/0008-lookups.md
@@ -1,0 +1,44 @@
+# 8. Lookups
+
+## Context
+
+We need to be able to specify a lookup component within a runner schema. These are to be used to provide a dropdown list of suggestions and simple entry before a user enters the full answer text.
+
+## Decision
+
+We will define a new answer type 'Lookup' within the schema which identifies the answer should be rendered as a lookup component.
+
+    {
+        id: 'religion-question',
+        title: 'What is your religion?',
+        description: 'This question is voluntary',
+        type: 'General',
+        answers: [
+          {
+            id: 'religion-answer',
+            mandatory: false,
+            label: 'Religion',
+            type: 'Lookup',
+            lookup_type: 'Religion'
+          }
+        ]
+    }
+
+The answer behaves as a Textfield and specifies a `lookup_type` reference to a configuration value in runner used to determine a api url.
+
+The api url will then be passed to the frontend component to retrieve suggestions from when JavaScript is enabled.
+
+### REST Service
+
+The api to query is a service that takes an input string and returns a single flat array of suggestion strings. 
+
+Initially there will be no authentication.
+
+One service is expected to host all single tier lookups.
+
+The region code is supplied as a query parameter and understood by the service in order to provide language variations when available.
+
+## Consequences
+
+- Lookups are able to be represented within a schema.
+- Lookup urls are not hardcoded and able to be specified per environment


iwootten:
I had pondered whether including a limit for the number of results to display might be appropriate?

bitdivision:
I think when this was initially discussed, we talked about having the ability to avoid using a lookup api for simple lookups, i.e. where there are not many options. Obviously there are some disadvantages to this, so I assume this was ruled out?

iwootten:
@bitdivision - don't remember that conversation, but I guess that's still an option if we set a flag to indicate the query_path should be used against the runner host itself?

ajmaddaford:
@bitdivision I think the intent of the lookup has changed as it's gone through user testing. The current design/prototype is for the suggestions from the lookup to behave more like a search than an autocomplete. If we weren't using the api we'd end up having to have two implementations of that, one client side and another in the api.

doc/architecture/decisions/0008-lookups.md
```diff
@@ -0,0 +1,46 @@
...
+            id: 'religion-answer',
+            mandatory: false,
+            label: 'Religion',
+            type: 'Lookup',
+            query_path: 'path/to/query'
+          }
+        ]
+    }
+
+The answer behaves as a Textfield when JavaScript is disabled and specifies a query_path for the frontend component to retrieve suggestions from when enabled.
```
ajmaddaford: I think it behaves as a Textfield regardless of Javascript. The Javascript will add the suggestions list below the textfield.

image

pricem14pc:
> I think when this was initially discussed, we talked about having the ability to avoid using a lookup api for simple lookups, i.e. where there are not many options. Obviously there are some disadvantages to this, so I assume this was ruled out?

Presumably all the look ups need to behave the same regardless of the number of entries in the datasets? Unless a simple lookup could replicate the behaviour of an elasticsearch api (e.g. partial tokenized matches etc) then it might impact the user experience.

ajmaddaford (approved):

bitdivision (approved):

MebinAbraham (changes requested):
>I had pondered whether including a limit for the number of results to display might be appropriate?

Do we not want to do add an option for this, instead of leaving it to the front end to determine a hard limit on all look ups?.

Is the only reason we may not want it to be configurable in the schema is that it's researched and the current hard limit just works and wouldn't need changing?

@iwootten @ajmaddaford @bitdivision

doc/architecture/decisions/0008-lookups.md
```diff
@@ -0,0 +1,46 @@
...
+            id: 'religion-answer',
+            mandatory: false,
+            label: 'Religion',
+            type: 'Lookup',
+            query_path: 'path/to/query'
+          }
+        ]
+    }
+
+The answer behaves as a Textfield and specifies a query_path for the frontend component to retrieve suggestions from when JavaScript is enabled.
```
MebinAbraham: 🤷‍♂ Personally prefer if variables in sentences are highlighted in some way.
```suggestion
The answer behaves as a Textfield and specifies a `query_path` for the frontend component to retrieve suggestions from when JavaScript is enabled.
```

doc/architecture/decisions/0008-lookups.md
```diff
@@ -0,0 +1,46 @@
...
+
+The api to query is a service that takes an input string and returns a single flat array of suggestion strings. 
+
+Initially there will be no authentication.
+
+One service is expected to host all single tier lookups.
+
+The hostname of the service is specified as an environment variable.
+
+The hostname would be combined with the query_path attribute at runtime to determine the full api url.
```
MebinAbraham:
```suggestion
The hostname would be combined with the `query_path` attribute at runtime to determine the full api url.
```

MebinAbraham:
Curious, does `query_path` need to live in the schema?. Have we considered a `lookup_type` option instead?. Means it would only need to live in runner and validator wouldn't need knowledge of the path itself and that's just another place to update when/if the path changes.

bitdivision:
@MebinAbraham Regarding the number of items shown, we discussed this briefly yesterday and I think the conclusion was that there should be one number which makes sense for all devices and that number can be chosen by the product owner. There shouldn't ever be a need to modify it on a per schema basis.

MebinAbraham (approved):

bitdivision (approved):

PR #2073, opened 2019-05-13
Fail Travis / Docker build if Jsonnet does not run

jonnyshaw89:
Jsonnet currently fails silently

### How to review 
Are census schemas built in the container?


Dockerfile
```diff
@@ -1,7 +1,11 @@
 FROM ubuntu:18.04 as builder
 
 RUN apt-get update \
-    && apt-get install -y curl unzip
+    && apt-get install -y curl unzip git make build-essential
```
ajmaddaford, line 4: Are curl and unzip needed?

ajmaddaford (changes requested):

Dockerfile
jonnyshaw89: yeh for `load_templates.sh`

.gitignore
```diff
@@ -92,7 +90,7 @@ requirements.txt
 .screenshots
 
 # Ignore built schemas
-data/en/census_individual.json
+/data
```
bitdivision: Should this have a leading slash?

jonnyshaw89: yes

jonnyshaw89: no

bitdivision (approved):

ajmaddaford (approved):

PR #2075, opened 2019-05-13
Add form type 70 and remove metadata

rmccar:
Added form type 70 and removed trad_as and employment_date metadata from being mandatory 

### How to review 
Check that the forms launch without needing trad_as or employment_date metadata and that form type 70 looks ok and launches without error

### DIFF
--- a/data/en/stocks_0001.json
+++ b/data/en/stocks_0001.json
@@ -123,7 +123,7 @@
                         "type": "Question",
                         "questions": [{
                             "id": "question4717",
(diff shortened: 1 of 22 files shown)


samiwel:
[Stocks: Error 403 when eQ is launched from RAS-RM](https://trello.com/c/3dUO82hS/1115-stocks-error-403-when-eq-is-launched-from-ras-rm)

heofs (approved):

samiwel:
@rmccar I've formatted all of the date piping values and pushed up commit. Do you want to just sanity check my change?

rmccar:
They look good to me

samiwel (approved):

PR #2076, opened 2019-05-14
Marshmallow flavoured metadata

bitdivision:
Currently, runner uses a custom set of validation and parsing code to manage the claims we receive from upstream. This code could be cleaner and it would be nice to use a standard serialisation / validation framework.

Marshmallow seems to be a good choice. It allows us to define required claims within a schema, and dynamically generate the schema for any other fields which might be specified for a particular questionnaire.

I ended up upgrading marshmallow to a prerelease version (v3). The project recommends v3 for new products and it seems to be fairly stable. The documentation for v2 is wrong in a few places, and the API for v3 is much clearer. This did mean upgrading the two or three places we use marshmallow v2, but the only real change is that we don't need `strict=True` and return values are not tuples any more.

The schema for the questionnaire has also changed slightly. Using 'type' rather than 'validator' is clearer. I have added a `length` key and `optional` key which provide extra validation.

Metadata parsing now consists of generating a marshmallow schema for the questionnaire and then running `.load` on that schema. This will deserialise the claims into a python object, and validate claims against the schema.

I had a few niggles with marshmallow. There is currently no way to treat empty strings as None values, which would have made things much easier. For now, I've set a min length of 1 on all required runner claims. Fields are also not required by default, hence a bit of hackery with `functools.partial()`

There will be a corresponding go-launcher and eq-schema-validator PR:
https://github.com/ONSdigital/go-launch-a-survey/pull/135
https://github.com/ONSdigital/eq-schema-validator/pull/130

### How to review 
- Ensure the tests are sufficient for all types of validation.
- Do we need more validation on the runner claims schema?

### Checklist


app/storage/metadata_parser.py
```diff
@@ -1,103 +1,90 @@
...
-    {'name': 'tx_id', 'validator': 'uuid'},
-    {'name': 'case_id', 'validator': 'uuid'},
-    {'name': 'response_id', 'validator': 'string'},
-]
-
 
-def parse_runner_claims(claims):
-    cleaned_claims = clean_leading_trailing_spaces(claims.copy())
-    validate_metadata(cleaned_claims, MANDATORY_METADATA)
+class MetadataSchema(Schema):
```
ajmaddaford: It might be clearer if this was called `RunnerMetadataSchema` to differentiate it from `QuestionnaireMetadataSchema`

app/storage/metadata_parser.py
```diff
@@ -1,103 +1,90 @@
...
+    """ The metadata which is required for runner functionality.
+    This is used to generate a dynamic schema including questionnaire specific claims
```
ajmaddaford: I don't think this comment is relevant here, it should be at the point the two claims are merged

app/storage/metadata_parser.py
```diff
@@ -1,103 +1,90 @@
...
+        return items
 
-def validate_metadata(claims, required_metadata):
-    _validate_metadata_values_are_valid(claims, required_metadata)
+    def __repr__(self):
+        return '<MetadataSchema({!r})>'.format(self.__dict__)
 
 
-def _validate_metadata_values_are_valid(claims, required_metadata):
+def validate_marshmallows(claims, questionnaire_specific_metadata):
```
ajmaddaford: I think the creation of the `QuestionnaireMetadataSchema` should be its own method, rather than combining it with the validation.

app/storage/metadata_parser.py
```diff
@@ -1,103 +1,90 @@
...
+        'QuestionnaireMetadataSchema', (MetadataSchema,), dynamic_fields
+    )(unknown=INCLUDE)
+
+    # The load method performs validation.
+    return questionnaire_metadata_schema.load(claims)
+
+
+def validate_runner_claims(claims: Dict):
+    """ Validate claims required for runner to function"""
+    return validate_marshmallows(claims, [])
```
ajmaddaford: It's really confusing to pass an empty array to `validate_marshmallows`, to create a `QuestionnaireMetadataSchema` object with no dynamic fields, to do the validation. This would be clearer just using `MetadataSchema`.

app/storage/metadata_parser.py
```diff
@@ -1,103 +1,90 @@
...
+
+
+def validate_questionnaire_claims(claims: Dict, questionnaire_metadata: List):
+    """ Validate any survey specific claims required for a questionnaire"""
```
ajmaddaford: Doesn't this validate all claims, i.e. including the runner claims, as `QuestionnaireMetadataSchema` is based on `MetadataSchema`?

app/views/session.py
```diff
@@ -53,13 +57,18 @@ def login():
...
+    try:
+        claims = validate_runner_claims(decrypted_token)
+    except ValidationError as e:
+        logger.exception('Invalid claims on login')
+        raise InvalidTokenException('Invalid claims') from e
 
     g.schema = load_schema_from_metadata(claims)
     schema_metadata = g.schema.json['metadata']
-    validate_metadata(claims, schema_metadata)
+    claims = validate_questionnaire_claims(claims, schema_metadata)
```
ajmaddaford: What happens if there are missing claims? Previously this raised a `InvalidTokenException`

ajmaddaford (changes requested):

app/storage/metadata_parser.py
```diff
@@ -1,103 +1,78 @@
...
+    @pre_load
+    def strip_whitespace(self, items):  # pylint: disable=no-self-use
+        for key, value in items.items():
+            if isinstance(value, str):
+                items[key] = value.strip()
+        return items
+
+
+class RunnerMetadataSchema(Schema, StripWhitespaceMixin):
+    jti = VALIDATORS['uuid'](validate=validate.Length(min=1))
```
jonnyshaw89: uuid validator with a min length of 1?

jonnyshaw89: What happens with a uuid with an empty string?

application.py
```diff
@@ -28,7 +28,9 @@ def configure_logging():
         'DEBUG': logging.DEBUG,
     }
     handler = logging.StreamHandler()
-    logging.basicConfig(level=levels[EQ_LOG_LEVEL], format=log_format, handlers=[handler])
+    logging.basicConfig(
```
jonnyshaw89, line 31: If you have `black`'d this then you should probably update the format stop to include this.

scripts/generate_schemas.py
```diff
@@ -8,9 +8,7 @@
 
 
 def generate_schema(manifest, survey):
-    schema = {
-        'groups': [],
-    }
+    schema = {'groups': []}
```
jonnyshaw89, line 11: Same comment re `black`ing of `scripts/`

jonnyshaw89 (changes requested):

app/storage/metadata_parser.py
bitdivision: Tested that empty strings are not valid UUIDs

app/storage/metadata_parser.py
```diff
@@ -1,103 +1,108 @@
...
+    jti = VALIDATORS['uuid']()
+    eq_id = VALIDATORS['string'](validate=validate.Length(min=1))
+    form_type = VALIDATORS['string'](validate=validate.Length(min=1))
+    ru_ref = VALIDATORS['string'](validate=validate.Length(min=1))
+    collection_exercise_sid = VALIDATORS['string'](validate=validate.Length(min=1))
+    tx_id = VALIDATORS['uuid']()
+    case_id = VALIDATORS['uuid']()
+    response_id = VALIDATORS['string']()
+
+    # Account Service URL should be mandatory
```
MebinAbraham: Shouldn't?

bitdivision: No, it should be required, but I had to make it optional.

bitdivision: Apparently I hadn't thought about this much. It should have been mandatory and could have been mandatory. 

It's now mandatory.

app/storage/metadata_parser.py
```diff
@@ -1,103 +1,106 @@
...
+        return items
 
-def parse_runner_claims(claims):
-    cleaned_claims = clean_leading_trailing_spaces(claims.copy())
-    validate_metadata(cleaned_claims, MANDATORY_METADATA)
 
-    return cleaned_claims
+class RunnerMetadataSchema(Schema, StripWhitespaceMixin):
+    """ Metadata which is required for the operation of runner itself
+    """
```
MebinAbraham, line 50: **Critical**, PEP recommends closing quotes for single line docstrings to be on the same line. 🤷‍♀  😑  👀 🤷‍♂

app/storage/metadata_parser.py
```diff
@@ -1,103 +1,106 @@
...
+        return super()._deserialize(*args, **kwargs).strftime('%Y-%m-%d')
 
 
 VALIDATORS = {
-    'date': iso_8601_date_parser,
-    'uuid': uuid_4_parser,
-    'boolean': boolean_parser,
-    'string': string_parser,
-    'optional_string': optional_string_parser,
+    'date': functools.partial(DateString, format='%Y-%m-%d', required=True),
```
jonnyshaw89, line 31: 👍

scripts/run_lint.sh
```diff
@@ -28,7 +28,7 @@ pylint --reports=n --output-format=colorized --rcfile=.pylintrc -j 0 ./app ./tes
 # http://stackoverflow.com/questions/6626351/how-to-extract-bits-from-return-code-number-in-bash
 display_result $? 2 "Pylint linting check"
 
-black --check app tests
+black --check . --exclude node_modules
```
jonnyshaw89, line 31: 👍

jonnyshaw89 (approved):
Looks good!

app/views/session.py
```diff
@@ -53,13 +57,25 @@ def login():
...
-    validate_metadata(claims, schema_metadata)
+
+    try:
+        questionnaire_claims = validate_questionnaire_claims(
+            decrypted_token, schema_metadata
+        )
+    except ValidationError as e:
+        raise InvalidTokenException('Invalid questionnaire claims') from e
+
+    claims = {**runner_claims, **questionnaire_claims}
```
ajmaddaford, line 78: 👍 This is much clearer 🙂

scripts/test_schemas.sh
```diff
@@ -6,8 +6,8 @@ if [ "$1" == "--local" ] || [ "$2" == "--local" ]; then
 fi
 
 if [ "$run_docker" == true ]; then
-    docker pull onsdigital/eq-schema-validator:v3
-    validator="$(docker run -d -p 5001:5000 onsdigital/eq-schema-validator:v3)"
+    docker pull onsdigital/eq-schema-validator:marshmallow-flavoured-metadata
+    validator="$(docker run -d -p 5001:5000 onsdigital/eq-schema-validator:marshmallow-flavoured-metadata)"
```
ajmaddaford: This needs to be changed back to v3 when the validator changes merge.

ajmaddaford (approved):

PR #2081, opened 2019-05-15
Applied question page title style to calculated summary page title

pricea1:
Calculated Summary Title is not applying highlight styles.

### How to review 
Add a few pages with number answers.
Add a calculated summary page.
In the title of the calculated summary page, use the rich text editor to highlight some of the content.
Preview/launch the questionnaire.
Observe that the highlighting is visible on the calculated summary page.

### DIFF
--- a/app/templates/summary.html
+++ b/app/templates/summary.html
@@ -37,8 +37,8 @@
     </div>
 </div>
 {% elif content.summary.summary_type == 'CalculatedSummary' %}
-<div>
-  <h1 class="u-fs-l" data-qa="calculated-summary-title">{{content.summary.title}}</h1>
+<div class="question">
+  <h1 class="u-fs-l question__title" data-qa="calculated-summary-title">{{content.summary.title}}</h1>
   <div class="print__hidden panel panel--simple panel--success u-mb-l">
     <h2 class="u-fs-r--b">{{ _("Please review your answers and confirm these are correct") }}.</h2>
   </div>


pricea1:
https://github.com/ONSdigital/eq-author-app/issues/343

pricea1:
[Highlighting in title not shown on calculated summary page on runner](https://trello.com/c/O3HHKy2o/1120-highlighting-in-title-not-shown-on-calculated-summary-page-on-runner)

app/templates/summary.html
```diff
@@ -30,15 +30,15 @@ <h2 class="u-fs-l">{{ _("Please remember to submit these answers") }}.</h2>
 {% elif content.summary.summary_type in ('SectionSummary', 'AnswerSummary') %}
 <div>
     {% set group = content.summary.groups %}
-    <h1 class="u-fs-l">{{content.summary.title}}</h1>
+    <h1 class="u-fs-l question__title">{{content.summary.title}}</h1>
```
samiwel: @hamishtaplin is it okay to use this style on the summary page? Seems to fix the problem, but semantically this is a summary title not a question title? 🤷‍♂

hamishtaplin: Not really. Survey Runner follows [The BEM convention](http://getbem.com/naming/) of writing CSS/components and so `question__title` should only be used within the scope of a `question` block.

Either we don't support it here, or we should make a CSS change.

pricea1: Now changed so its within scope of `question` block

samiwel:
IMO I think we're better off reverting this change. They are trying to highlight in a summary title which survey runner doesn't support. Rather we should probably make a change to Author to disable the highlighting button on the toolbar for summary titles.

samiwel (approved):
Tested locally and this change styles the title as expected.

hamishtaplin (approved):

PR #2084, opened 2019-05-16
Add Questionnaire ID

iwootten:

Introduce the questionnaire id as part of the required claims in the metadata passed to EQ and pass it on to submission payloads. If case id is present in the list of claims, pass this on too.

https://trello.com/c/0fkw3X5F/2984-jwt-claims-questionnaireid-and-caseid

### How to review 

You need to use the following launcher PR to launch a survey: https://github.com/ONSdigital/eq-questionnaire-launcher/pull/2

### DIFF
--- a/app/storage/metadata_parser.py
+++ b/app/storage/metadata_parser.py
@@ -55,10 +55,11 @@ class RunnerMetadataSchema(Schema, StripWhitespaceMixin):
     ru_ref = VALIDATORS['string'](validate=validate.Length(min=1))
     collection_exercise_sid = VALIDATORS['string'](validate=validate.Length(min=1))
     tx_id = VALIDATORS['uuid']()
-    case_id = VALIDATORS['uuid']()
-    response_id = VALIDATORS['string']()
+    questionnaire_id = VALIDATORS['string'](validate=validate.Length(min=1))
+    response_id = VALIDATORS['string'](validate=validate.Length(min=1))
     account_service_url = VALIDATORS['url']()
 
+    case_id = VALIDATORS['uuid'](required=False)
     account_service_log_out_url = VALIDATORS['url'](required=False)
     roles = fields.List(fields.String(), required=False)
     survey_url = VALIDATORS['url'](required=False)
--- a/app/views/feedback.py
+++ b/app/views/feedback.py
@@ -71,8 +71,9 @@ def send_feedback():
         )
         sent = current_app.eq['submitter'].send_message(
             encrypted_message,
-            case_id=metadata.get('case_id'),
             tx_id=metadata.get('tx_id'),
+            questionnaire_id=metadata.get('questionnaire_id'),
+            case_id=metadata.get('case_id'),
         )
 
         if not sent:
--- a/app/views/flush.py
+++ b/app/views/flush.py
@@ -66,8 +66,9 @@ def _submit_data(user):
 
         sent = current_app.eq['submitter'].send_message(
             encrypted_message,
-            case_id=metadata.get('case_id'),
             tx_id=metadata.get('tx_id'),
+            questionnaire_id=metadata.get('questionnaire_id'),
+            case_id=metadata.get('case_id'),
         )
 
         if not sent:
--- a/app/views/questionnaire.py
+++ b/app/views/questionnaire.py
@@ -71,6 +71,7 @@ def before_questionnaire_request():
         eq_id=metadata['eq_id'],
         metadata=metadata['form_type'],
         ce_id=metadata['collection_exercise_sid'],
+        questionnaire_id=metadata['questionnaire_id'],
     )
 
     logger.info(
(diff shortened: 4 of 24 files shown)


app/storage/metadata_parser.py
```diff
@@ -47,22 +52,24 @@ def clean_leading_trailing_spaces(metadata):
     'boolean': boolean_parser,
     'string': string_parser,
     'optional_string': optional_string_parser,
+    'optional_uuid': optional_uuid_4_parser,
 }
 
-MANDATORY_METADATA = [
+METADATA_VALIDATION_RULES = [
```
MebinAbraham: We should really rebase against Rich's changes as this is overlapping and has changed.

app/views/feedback.py
```diff
@@ -71,8 +71,9 @@ def send_feedback():
         )
         sent = current_app.eq['submitter'].send_message(
             encrypted_message,
-            case_id=metadata.get('case_id'),
             tx_id=metadata.get('tx_id'),
+            questionnaire_id=metadata.get('questionnaire_id'),
+            case_id=metadata.get('case_id', None),
```
MebinAbraham: Redundant `None`.

app/views/flush.py
```diff
@@ -66,8 +66,9 @@ def _submit_data(user):
 
         sent = current_app.eq['submitter'].send_message(
             encrypted_message,
-            case_id=metadata.get('case_id'),
             tx_id=metadata.get('tx_id'),
+            questionnaire_id=metadata.get('questionnaire_id'),
+            case_id=metadata.get('case_id', None),
```
MebinAbraham: Same here, don't need to specify a default.

app/views/questionnaire.py
```diff
@@ -550,7 +551,10 @@ def submit_answers(routing_path, schema):
         message, current_app.eq['key_store'], KEY_PURPOSE_SUBMISSION
     )
     sent = current_app.eq['submitter'].send_message(
-        encrypted_message, case_id=metadata.get('case_id'), tx_id=metadata.get('tx_id')
+        encrypted_message,
+        questionnaire_id=metadata.get('questionnaire_id'),
+        case_id=metadata.get('case_id', None),
```
MebinAbraham: And here

app/storage/metadata_parser.py
iwootten: Those changes aren't yet approved right?

MebinAbraham: Nope, I meant more when it is in?.

iwootten: Sure, but I can't see how that's different to our usual workflow.

MebinAbraham: I mean from the point that this card has overlaps with Rich's and maybe should have been on hold, as reviewing metadata parser is essentially redundant if/when Rich's goes in?.

iwootten: The card was originally on hold, and subsequently marked as unblocked - I consulted with Rich on whether he felt it ready to start. Yes they overlap, but then as I understand it, this card was originally marked the higher priority.

app/storage/metadata_parser.py
```diff
@@ -55,10 +55,11 @@ class RunnerMetadataSchema(Schema, StripWhitespaceMixin):
     ru_ref = VALIDATORS['string'](validate=validate.Length(min=1))
     collection_exercise_sid = VALIDATORS['string'](validate=validate.Length(min=1))
     tx_id = VALIDATORS['uuid']()
-    case_id = VALIDATORS['uuid']()
+    questionnaire_id = VALIDATORS['string']()
```
MebinAbraham: We should add length to these as currently, empty strings would be valid.

MebinAbraham (changes requested):

app/storage/metadata_parser.py
iwootten: I've changed these for now, but for case_id/questionnaire_id - I wonder if we should be validating them at all. Given that we don't use them for anything - we probably should just be passing them on?

MebinAbraham: Possibly, guess it depends on how mandatory it actually is. Do we return a 403 if they don't provide us a questionnaire_id vs rm to preventing launching without questionnaire_id.

iwootten: questionnaire_id will cause a failure, case_id will not

MebinAbraham: Ya I think your change is fine to validate questionnaire_id/response_id since they are needed but not case_id

ajmaddaford (approved):

bitdivision (commented):
Looks great :+1:

bitdivision (approved):

MebinAbraham (approved):

jonnyshaw89 (approved):

PR #2086, opened 2019-05-17
Refactor date forms

jonnyshaw89:
Removed duplication from validators
Dates functional test was commented out, so this had been uncommented and fixed

### How to review 
Do dates still work as expected.
Is the refactor of the date form good enough?


app/forms/date_form.py
```diff
@@ -16,191 +16,125 @@
...
-        super(DateField, self).__init__(form_class, **kwargs)
-
-    def process(self, formdata, data=None):
-        if data is not None:
-            substrings = data.split('-')
-            data = {'year': substrings[0], 'month': substrings[1], 'day': substrings[2]}
-
-        super().process(formdata, data)
-
+class DateFormTypes(Enum):
```
bitdivision: :+1:

Though, I think this should be called `DateFormType`. It represents a single type.

app/forms/date_form.py
```diff
@@ -16,191 +16,125 @@
...
+        metadata,
+        answer,
+        error_messages,
+        **kwargs,
+    ):
+        form_class = get_form(
+            date_form_type, answer, answer_store, metadata, error_messages
         )
-        super(YearField, self).__init__(form_class, **kwargs)
+        super(DateField, self).__init__(form_class, **kwargs)
```
bitdivision: You should be able to use `super().__init__(form_class, **kwargs)`. Using the name of the parent class is not needed in python 3.

app/forms/date_form.py
```diff
@@ -16,191 +16,125 @@
...
-    """
-    Returns a month year form metaclass with appropriate validators. Used in both date and
-    date range form creation.
+class DateForm(Form):
+    @CachedProperty
+    def data(self):
+        data = super().data
+        try:
+            if all(k in data for k in ('day', 'month', 'year')):
+                return '{year:04d}-{month:02d}-{day:02d}'.format(
```
bitdivision, line 89: Might be a bit cleaner:
                `return f'{data['year']:04d}-{data['month']:02d}-{data['day']:02d}'`

app/forms/date_form.py
```diff
@@ -16,191 +16,125 @@
...
+                    month=int(data['month']),
+                    day=int(data['day']),
+                )
 
-    :param answer: The answer on which to base this form
-    :param error_messages: The messages to use upon this form during validation
-    :return: The generated MonthYearDateForm metaclass
-    """
+            if all(k in data for k in ('month', 'year')):
+                return '{year:04d}-{month:02d}'.format(
```
bitdivision, line 96: Same again: `return f'{data['year']:04d}-{data['month']:02d}'`

app/forms/date_form.py
```diff
@@ -16,191 +16,125 @@
...
+                    year=int(data['year']), month=int(data['month'])
+                )
 
-    class MonthYearDateForm(Form):
-        year = StringField()
-        month = StringField()
+            if 'year' in data and data['year']:
+                return '{year:04d}'.format(year=int(data['year']))
```
bitdivision, line 101: `return f'{data['year']:04d}'`

app/forms/date_form.py
```diff
@@ -16,191 +16,125 @@
...
-        def data(self):
-            data = super().data
-            if not data['year']:
-                return None
+    class CustomDateForm(DateForm):
+        pass
 
-            return '{:04d}'.format(int(data['year']))
+    if form_type in DateFormTypes:
+        # Validation os only ever added to the 1 field that shows in all 3 variants
```
bitdivision:
```suggestion
        # Validation is only ever added to the 1 field that shows in all 3 variants
```

bitdivision (changes requested):
This is looking much much better.

And thanks for fixing my tests :)

app/forms/date_form.py
jonnyshaw89: would it not have to be 
`return f'{int(data['year']):04d}-{int(data['month']):02d}-{int(data['day']):02d}'`

bitdivision: Yeah, you're right. I guess we could cast those in one place, but doesn't make much difference overall.

bitdivision (approved):

ajmaddaford (approved):

PR #2087, opened 2019-05-17
Add missing routing rules for stocks 0013 and 0014.

samiwel:
The routing rules for the first question of the survey were removed incorrectly.
This change re-introduces the routing rules.

### How to review 
1. Launch stocks 0013 survey.
1. Answer yes for the first question.
1. Should be routed to the other questions.
1. Answer no for the first question.
1. Should be prompted for the reporting dates.

Repeat the above test for 0014.

### DIFF
--- a/data/en/stocks_0013.json
+++ b/data/en/stocks_0013.json
@@ -140,7 +140,23 @@
                                     }
                                 ]
                             }]
-                        }]
+                        }],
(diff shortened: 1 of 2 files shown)


tgandrews (approved):
LGTM. Works locally for 14 and 13

rmccar (approved):

PR #2090, opened 2019-05-20
Fix margins

bameyrick:
- Updated the design system version to ~~7.0.1~~ 8.0.1
- Added paragraph filter to jinja_filters to ensure question descriptions are marked up correctly
- Removed unnecessary font size utility classes as a result of updated design system

The paragraph filter I've added feels a bit janky and that this should maybe be done before we reach the template, so if anyone has any suggestions on how to fix this please let me know. Also I'm unsure if when a question description has two paragraphs whether it is split by a carriage return or a newline because I couldn't find any schemas with two paragraphs for a question description.


app/jinja_filters.py
```diff
@@ -546,3 +546,17 @@ def map_list_collector_config(
 @blueprint.app_context_processor
 def map_list_collector_config_processor():
     return dict(map_list_collector_config=map_list_collector_config)
+
+
+@blueprint.app_template_filter()
+def format_paragraphs(text):
```
iwootten, line 552:
```suggestion
def format_paragraphs(text):
        return "\n".join(f'<p>{paragraph}</p>' for paragraph in text.splitlines())
```

jonnyshaw89: Should we not be questioning the need for paragraphs in question descriptions? if this much content is needed for a question are there bigger issues?

iwootten: Can @bameyrick point to an example where this is causing formatting issues at the moment?

app/jinja_filters.py
```diff
@@ -546,3 +546,18 @@ def map_list_collector_config(
...
+    paragraphs = text.split('\n')
+
+    if len(paragraphs) > 1:
+        return '<p>' + paragraphs.join('</p><p>') + '</p>'
+    else:
+        return '<p>' + text + '</p>'
```
bambrp:
```suggestion
 -   else
 -       return '<p>' + text + '</p>'
 +    return '<p>' + text + '</p>'
```

app/jinja_filters.py
```diff
@@ -546,3 +546,18 @@ def map_list_collector_config(
...
     return dict(map_list_collector_config=map_list_collector_config)
+
+
+@blueprint.app_template_filter()
+def format_paragraphs(text):
+    paragraphs = text.split('\n')
+
+    if len(paragraphs) > 1:
+        return '<p>' + paragraphs.join('</p><p>') + '</p>'
+    else:
```
bambrp: Need to lose this else and de-indent the following line one level.

app/jinja_filters.py
bameyrick: I couldn't find any examples where there were multiple paragraphs of question description, but I checked with @hamishtaplin whether it was possible to do multiple paragraphs in Author and you can. I thought it would be safest to wrap the description text in paragraph tags here than in the template just incase somebody authors a survey with multiple paragraphs.

app/jinja_filters.py
bameyrick: This has been superseded by @iwootten's suggestion

README.md
```diff
@@ -84,6 +84,12 @@ pip install --upgrade pip setuptools pipenv
 pipenv install --dev
 ```
 
+To update the design system templates run:
+
+```
+pipenv run ./scripts/load_templates.sh
```
MebinAbraham: Can you add this to the Makefile, maybe `make load-templates`?

scripts/load_templates.sh
```diff
@@ -4,7 +4,7 @@ DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
 cd "${DIR}"/.. || exit
 
-DESIGN_SYSTEM_VERSION="6.1.0"
+DESIGN_SYSTEM_VERSION="8.0.1"
```
MebinAbraham: Not really related but was there a reason we did not use an env var for this?. Means we would be able to upgrade the non-breaking changes without app changes.

jonnyshaw89: Because we would always want this to be fixed and not change without tests

templates/interstitial.html
```diff
@@ -6,17 +6,17 @@
 
 {% block form_content %}
   {% if block['section_number'] %}
-    <h1 class="u-fs-l">Section {{ block['section_number'] }}: {{ block['title'] }}</h1>
+    <h1>Section {{ block['section_number'] }}: {{ block['title'] }}</h1>
```
MebinAbraham, line 9: These are removed in another open PR, need to ensure we merge properly to make sure your changes are accounted for.

README.md
bameyrick: Maybe I just add it as part of the `make run` task and remove this from the readme? Seems to download and unzip the templates pretty fast

bitdivision: Should still be in the readme but I would agree, it can go in make run

bameyrick: I've updated the README and makefile now

templates/partials/question.html
```diff
@@ -76,27 +82,4 @@
   {% else %}
     {{ question_answers }}
   {% endif %}
-{% endset %}
-
-{% if wrap_question_in_fieldset %}
```
iwootten, line 81: I'm guessing this entire section has now been superseded by the design system, where in the previous version 6, it was still necessary for some reason?

iwootten (approved):
Seems legit. I can't see any obvious padding nightmares from stepping through census and aside from my single comment it seems like it's just simple class changes.

bitdivision (approved):

templates/partials/question.html
bameyrick: Yeah we have a component that takes care of this markup now

PR #2091, opened 2019-05-21
Remove block title and description

jonnyshaw89:
Currently, it is not possible to vary block title (page title) without duplicating the block and adjusting any routing rules.

This PR removes this content that is outside of `content_variants` and `question_variants` 

### How to review 
Do current schemas work as expected?

### DIFF
--- a/data-source/json/test_routing_not_affected_by_answers_not_on_path.json
+++ b/data-source/json/test_routing_not_affected_by_answers_not_on_path.json
@@ -25,7 +25,7 @@
                 "question": {
                     "type": "General",
                     "id": "initial-choice-question",
-                    "title": "Answer First, then, after answering a quesiton, go back to this question and answer Second",
+                    "title": "Answer First, then, after answering a question, go back to this question and answer Second",
                     "answers": [{
                         "type": "Radio",
                         "id": "initial-choice-answer",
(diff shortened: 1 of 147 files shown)


MebinAbraham:
@jonnyshaw89 Is there another PR planned?. Need templating changes to apply the page title from variants.

jonnyshaw89:
> @jonnyshaw89 Is there another PR planned?. Need templating changes to apply the page title from variants.

Which page title?

MebinAbraham:
> > @jonnyshaw89 Is there another PR planned?. Need templating changes to apply the page title from variants.
> 
> Which page title?

So block title is currently the page title. We need the title to be allowed in content variants. The current title within `content_variants` > `content` is not the main page title(wrapped in `h1`).

templates/interstitial.html
```diff
@@ -5,16 +5,14 @@
...
-  {% if block['section_number'] %}
-    <h1 class="u-fs-l">Section {{ block['section_number'] }}: {{ block['title'] }}</h1>
-  {% else %}
-    <h1 class="u-fs-l">{{ block['title'] }}</h1>
+  {% if block.content.title %}
+    <h1 class="u-fs-l">{{ block.content.title }}</h1>
   {% endif %}
   {% if block['description'] %}
     <p>{{ block['description'] | safe }}</p>
   {% endif %}
```
MebinAbraham: As per the validator changes, the block description can go?. Same for the multiple occurrences.

iwootten (approved):
Again, difficult to identify minor issues that may have crept in here, but builds and validates with sister PR and I can't see anything obvious, so happy to accept on that basis.

Have fun rebasing this.

templates/partials/contents.html
```diff
@@ -0,0 +1,15 @@
+{% for item in content_block.contents %}
```
ajmaddaford: This is a little confusing as contents aren't always in a block content element. It would be clearer if a `contents` variable was set in templates that include this template. So using `{% set contents = content_block.contents %}` in the parent template and this would just be `{% for item in contents %}`.

ajmaddaford (changes requested):

data-source/json/test_routing_not_affected_by_answers_not_on_path.json
```diff
@@ -115,10 +110,11 @@
             }, {
                 "type": "Interstitial",
                 "id": "valid-skipped-interstitial",
-                "title": "This page should have been skipped!",
-                "content": [{
-                    "description": ""
-                }],
+                "content": {
+                    "contents": [{
+                        "description": ""
```
MebinAbraham, line 117: Bad find/replace?. Context missing from v3. A couple of occurrence like this.

bitdivision: Hmm. Not sure what happened there. I'll look through and find any others, but if you spot any let me know.

ajmaddaford (approved):

PR #2095, opened 2019-05-23
Completeness of a section

JamesGardiner:
This PR alters the way routing is handled in the survey. Specifically routing paths are now only built a single section at a time instead of for all sections, and completed sections are tracked as an attribute of a Completed object.

Changing the scope of a routing path to be constrained to a section means that additional checks are required to understand when a user is completing the last question in a section (and therefore the section should be complete), as well as when a new routing path based on the next section is required.

### How to review 
Run the full test suite. Run the application and observe moving back and fore between sections, as well as the section objects being tracked in the Completed object correctly (including when a user goes back to a section therefore marking it incomplete).

https://trello.com/c/3g29VCp7/2943-completeness-of-a-section-l


app/questionnaire/questionnaire_schema.py
```diff
@@ -14,75 +14,96 @@ def __init__(self, questionnaire_json, language_code=DEFAULT_LANGUAGE_CODE):
...
+
+    def _section_via_block_id(self, block_id):
+        block = self.block(block_id)
+
+        if block:
+            try:
+                section_id = self.group(block['parent_id'])['parent_id']
+            except TypeError:
+                # Handle lists - they are nested one more level
+                section_id = self.group(self.block(block['parent_id'])['parent_id'])[
```
iwootten: This would be better handled under a named method, preventing the need for the comment.

app/questionnaire/completed.py
```diff
@@ -0,0 +1,41 @@
+from typing import List, Mapping
+
+from app.questionnaire.location import Location
+
+
+class Completed:
```
iwootten: Completed sounds like an event to me. Possibly better to name this class a 'thing', as per elsewhere. 'Completeness' or 'CompletionTracker' might be better.

app/questionnaire/questionnaire_schema.py
```diff
@@ -14,75 +14,96 @@ def __init__(self, questionnaire_json, language_code=DEFAULT_LANGUAGE_CODE):
...
+        section_ids = tuple(self._sections_by_id.keys())
+        next_section_id = section_ids[section_ids.index(section_id) + 1]
+        return self.section(next_section_id)
+
+    def _section_via_block_id(self, block_id):
+        block = self.block(block_id)
+
+        if block:
+            try:
+                section_id = self.group(block['parent_id'])['parent_id']
```
iwootten: Seems like a few of these accessors are repeated, might make things clearer to break them into named methods.

app/questionnaire/questionnaire_schema.py
```diff
@@ -14,75 +14,96 @@ def __init__(self, questionnaire_json, language_code=DEFAULT_LANGUAGE_CODE):
...
-    def get_group_by_block_id(self, block_id):
-        block = self.get_block(block_id)
-        if block['type'] in LIST_COLLECTOR_CHILDREN:
-            parent = self.get_block(block['parent_id'])
-            return self.get_group(parent['parent_id'])
-        return self.get_group(block['parent_id'])
-
     @classmethod
-    def get_answer_ids_for_question(cls, question):
+    def answer_ids_via_question(cls, question):
```
iwootten: I think the original method name was clearer here.

app/helpers/path_finder_helper.py
```diff
@@ -28,7 +33,12 @@ def get_path_finder():
 def full_routing_path_required(function):
     @wraps(function)
     def wrap_function(*args, **kwargs):
-        routing_path = path_finder.get_full_routing_path()
+        if 'block_id' in kwargs:
+            section = get_current_section(block_id=kwargs['block_id'])
+            if not section:
+                section = get_current_section()
+
+        routing_path = path_finder.routing_path(section)
```
jonnyshaw89, line 37: if `'block_id'` isn't in `kwargs`?

app/questionnaire/questionnaire_schema.py
```diff
@@ -14,75 +14,96 @@ def __init__(self, questionnaire_json, language_code=DEFAULT_LANGUAGE_CODE):
...
-        return self._blocks_by_id.get(block_id)
+    def answer_ids(self):
+        return self._answers_by_id.values()
 
-    def get_block_for_answer_id(self, answer_id):
-        answers = self.get_answers(answer_id)
-        # All matching questions / answers must be within the same block
-        questions = self.get_questions(answers[0]['parent_id'])
-        return self.get_block(questions[0]['parent_id'])
+    def answers(self, answer_id):
```
iwootten: The method name here doesn't imply that it's actually filtering the answers collection.

app/questionnaire/path_finder.py
```diff
@@ -164,36 +195,38 @@ def _remove_rule_answers(self, goto_rule, this_location):
...
 
-    def get_full_routing_path(self):
+    def routing_path(self, current_section: Mapping) -> RoutingPath:
         """
         Returns a list of the block ids visited based on answers provided
         :return: List of block location dicts
         """
-        if self._full_routing_path and not self.answer_store.is_dirty:
-            return self._full_routing_path
+        if self._routing_path and not self.answer_store.is_dirty:
```
jonnyshaw89: I wonder if `self._routing_path` should be a map? keyed by the section?
that way we can potentially cache the path for multiple sections.

app/data_model/questionnaire_store.py
```diff
@@ -43,22 +43,21 @@ def set_metadata(self, to_set):
...
-        self.completed_blocks = completed_blocks
         self.collection_metadata = json_data.get('COLLECTION_METADATA', {})
 
     def serialise(self):
         data = {
             'METADATA': self._metadata,
             'ANSWERS': list(self.answer_store),
             'LISTS': self.list_store.serialise(),
-            'COMPLETED_BLOCKS': self.completed_blocks,
+            **self.completed.serialise(),
```
jonnyshaw89: I think it would be better to nest this as 
```
{
...
'COMPLETED': {
    'BLOCKS': {},
    'SECTIONS': {}
},
...
```

iwootten: Think its probably better that the completed object has its own key at this level, rather than use the serialised objects attributes so we can make changes to the structure of it without affecting the top level of the questionnaire store.

    'COMPLETED': self.completed.serialise()

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -82,5 +82,9 @@ def remove_answer_ids(self, answer_ids: List):
         self._questionnaire_store.add_or_update()
 
     def remove_completed_blocks(self, location):
-        self._questionnaire_store.remove_completed_blocks(location)
+        self._questionnaire_store.completed.blocks.remove(location)
+        self._questionnaire_store.add_or_update()
```
jonnyshaw89: Why do we need to call this now when we previously didn't?
We should check that this wont same the state multiple times per request as this has a substantial impact on performance.

app/questionnaire/questionnaire_schema.py
```diff
@@ -93,22 +114,22 @@ def get_answer_ids_for_question(cls, question):
 
         return answer_ids
 
-    def get_answer_ids_for_block(self, block_id):
-        block = self.get_block(block_id)
+    def answer_ids_for_block(self, block_id):
```
iwootten: Again, I think the original method name is clearer.

docker-compose-dev-linux.yml
```diff
@@ -0,0 +1,23 @@
+version: '3'
```
iwootten: This probably should be in a separate PR.

jonnyshaw89: should probably rename the other file to `-mac`?

gulpfile.babel.js
```diff
@@ -65,7 +65,14 @@ gulp.task('lint:json', () => {
 
 gulp.task('format:json', () => {
   return gulp
-    .src(['./data-source/json/*.json'])
+    .src(['./data/*/*.json'])
```
jonnyshaw89: Dont care about `/data` as its `.gitignore`

iwootten: Formatting the data output directory, shouldn't be necessary - the source files should only need to be?

package.json
```diff
@@ -12,12 +12,12 @@
   },
   "scripts": {
     "start": "gulp watch",
-    "lint": "gulp lint",
+    "lint": "yarn format && gulp lint",
```
jonnyshaw89: this will stop travis failing on lint errors

jonnyshaw89 (changes requested):

app/questionnaire/questionnaire_schema.py
```diff
@@ -14,75 +14,96 @@ def __init__(self, questionnaire_json, language_code=DEFAULT_LANGUAGE_CODE):
...
+            return self._blocks_by_id.get(block_id)
 
-    def get_section(self, section_id):
-        return self._sections_by_id.get(section_id)
+        if answer_id:
+            return self._block_via_answer(answer_id)
 
-    def get_group(self, group_id):
-        return self._groups_by_id.get(group_id)
+    def _block_via_answer(self, answer_id):
```
iwootten: _get_block_for_answer_id would be a better name here, given it's not an answer being supplied

gulpfile.babel.js
JamesGardiner: It was failing format tests without this

package.json
JamesGardiner: I'll remove this. @bameyrick just a heads up in case you thought it would go in with this PR.

iwootten (dismissed):
There's quite a few naming changes here that are counter to how we're doing things throughout the rest of the project. Dropping the use of get_ for example on things that aren't properties seems confusing.

docker-compose-dev-linux.yml
bitdivision: And update the readme for that.

JamesGardiner:
@iwootten I'll revert the get_ name changes - my personal preference is for the less verbose form here, but consistency across the project is better.

app/questionnaire/questionnaire_schema.py
JamesGardiner: `get_block_via_answer_id`?

jonnyshaw89: i think its the `via` that is contentious.

JamesGardiner: Ooof ok

docker-compose-dev-linux.yml
JamesGardiner: Making a seperate PR

JamesGardiner: https://github.com/ONSdigital/eq-survey-runner/pull/2099

iwootten:
Erk, we shouldn't be force pushing during review as we can't follow changes introduced in new commits.

app/questionnaire/questionnaire_schema.py
```diff
@@ -36,13 +36,13 @@ def _section_via_block_id(self, block_id):
...
             except TypeError:
-                # Handle lists - they are nested one more level
-                section_id = self.group(self.block(block['parent_id'])['parent_id'])[
-                    'parent_id'
-                ]
+                section_id = self._section_id_via_list_block(block)
 
             return self.section(section_id)
 
+    def _section_id_via_list_block(self, block):
```
jonnyshaw89: 🤣

app/globals.py
```diff
@@ -96,6 +98,42 @@ def get_answer_store(user):
...
+    elif block_id:
+        section = schema.section(block_id=block_id)
+
+    else:
+        section = schema.section(schema.json['sections'][0]['id'])
+
+    if not section:
+        get_current_section(user=user)
+
+    return section
```
ajmaddaford: It's not clear what this method is for and how it will behave given the various ways section can be resolved. 
- `get_current_section` is only probably true if you pass in a block_id, the others are about calculating where to go next or defaulting to the first section. 
- It's too opaque to use the user parameter to determine that we want to go to the next section if the current one is complete. 
- I don't think it will work to just use the last completed block in the list of completed blocks to determine section completeness as completed blocks is not section aware. This could be done by working out the routing path through a section, working out how far we get on that path with completed blocks, and then determine if that is the end of the section.
- I don't get the `if not section` logic. Why call the method again? What has changed that will succeed on a subsequent call?

app/questionnaire/path_finder.py
```diff
@@ -31,27 +42,47 @@ def _block_index_for_location(blocks, location):
...
+            last_completed_block = self.completed.blocks[-1].block_id
+            last_block_in_section = section['groups'][-1]['blocks'][-1]['id']
+
+            if last_completed_block == last_block_in_section:
+                section = self.schema.next_section(section['id'])
+
+        except IndexError:
+            section = self.schema.section(list(self.schema.sections())[0]['id'])
+
+        return section
```
ajmaddaford: This looks very similar to `get_current_section`

app/questionnaire/path_finder.py
```diff
@@ -231,47 +267,66 @@ def _get_valid_section_summary(self, current_block_id, routing_path):
...
-                if location not in self.completed_blocks:
+                if location not in self.completed.blocks:
                     return None
 
-    def get_previous_location(self, current_location, schema):
+    def get_previous_location(
+        self,
+        current_location: Location,
+        current_section: Mapping,
+        schema: QuestionnaireSchema = None,
```
ajmaddaford: We probably don't need to have a `schema` parameter as it defaults to `self.schema`.

ajmaddaford: In fact, now that the schema is accessed via `self.schema` we don't need the `schema` variable either.

app/questionnaire/router.py
```diff
@@ -55,13 +58,35 @@ def _get_survey_redirect_location(self):
...
+        completed_blocks = [
+            block.block_id for block in self._completed_locations.blocks
+        ]
+
+        incomplete_blocks = [
+            block
+            for group in first_incomplete_section['groups']
+            for block in group['blocks']
+            if block['id'] not in completed_blocks
+        ]
```
ajmaddaford: This will loop over all blocks in a group regardless if they're complete or not. As we only care about the first incomplete block it would be better to return as soon as we get one.

app/views/dump.py
```diff
@@ -25,22 +24,32 @@ def dump_debug():
     return questionnaire_store.serialise()
 
 
+@dump_blueprint.route('/dump/routing', methods=['GET'])
+@login_required
+@role_required('dumper')
+def dump_routing():
+    response = {'routing': 'routing-goes-here'}
```
ajmaddaford: 🤔

app/views/dump.py
```diff
@@ -25,22 +24,32 @@ def dump_debug():
...
     session_data = get_session_store().session_data
     schema = load_schema_from_session_data(session_data)
-    completed_blocks = questionnaire_store.completed_blocks
-    routing_path = PathFinder(
-        schema, answer_store, metadata, completed_blocks
-    ).get_full_routing_path()
+    completed = questionnaire_store.completed
+    routing_path = PathFinder(schema, answer_store, metadata, completed).routing_path(
+        get_current_section(user=current_user)
+    )
```
ajmaddaford: Why `get_current_section`, surely this should be all sections?

app/views/questionnaire.py
```diff
@@ -191,6 +193,20 @@ def get_block_handler(
     answer_store = questionnaire_store.answer_store
 
     current_location = Location(block_id, list_name, list_item_id)
+    try:
+        current_section_id = schema.section(block_id=block_id)['id']
+    except (AttributeError, TypeError):
+        current_section_id = None
```
ajmaddaford: Why would this ever fail? The block_id is validated and must exist, and must exist within a section.

app/views/questionnaire.py
```diff
@@ -191,6 +193,20 @@ def get_block_handler(
...
+
+    if current_location in questionnaire_store.completed.blocks:
+        questionnaire_store.completed.blocks.remove(current_location)
+        questionnaire_store.add_or_update()
+
+    if current_section_id in questionnaire_store.completed.sections and schema.block(
+        current_location.block_id
+    )['type'] not in ['SectionSummary', 'Summary']:
+        questionnaire_store.completed.sections.remove(current_section_id)
```
ajmaddaford: Adding is done via the `add_completed_section` method on questionnaire store updater, remove should be consistent.

app/views/questionnaire.py
```diff
@@ -280,15 +301,24 @@ def post_block_handler(
             return redirect(next_url)
 
     questionnaire_store_updater.save_answers(form)
-    next_location = path_finder.get_next_location(current_location=current_location)
 
-    if _is_end_of_questionnaire(block, next_location):
+    next_location = path_finder.get_next_location(
+        current_location, get_current_section(user=current_user)
+    )
```
ajmaddaford: Why does `get_next_location` need to know the current section?

app/questionnaire/path_finder.py
```diff
@@ -212,12 +245,15 @@ def get_next_location(self, current_location):
 
             return routing_path[current_location_index + 1]
 
+        routing_path = self.build_path(current_section)
```
ajmaddaford: What's the significance of rebuilding the path by calling `build_path` here that's different to the path returned when calling `routing_path` on line 227?

ajmaddaford (changes requested):

app/views/dump.py
JamesGardiner: That was a placeholder... Am I right in saying that this is now all encapsulated by the `/dump/submission/` endpoint?

app/questionnaire/path_finder.py
JamesGardiner: routing_path will return an already built path if it exists and the answer store isn't dirty. If we make it to this point we need to re-evaluate the path, as it is no longer valid (for instance, when skipping a question)

app/views/questionnaire.py
JamesGardiner: It needs the section to pass to `routing_path` - it doesn't need it as an arg though, as the section can be determined from the `current_location`

app/data_model/questionnaire_store.py
```diff
@@ -15,12 +15,12 @@ def __init__(self, storage, version=None):
...
         self.version = version
         self._metadata = {}
-        # metadata is a read-only view over self._metadata
+        # self.metadata is a read-only view over self._metadata
         self.metadata = MappingProxyType(self._metadata)
         self.collection_metadata = {}
         self.list_store = ListStore()
         self.answer_store = AnswerStore()
-        self.completed_blocks = []
+        self.completed = CompletedItems()
```
bitdivision: Should this be `completed_items` too?

app/helpers/path_finder_helper.py
```diff
@@ -4,19 +4,24 @@
...
-
-    if finder is None:
+    try:
+        finder = g.path_finder
+    except AttributeError:
         metadata = get_metadata(current_user)
         answer_store = get_answer_store(current_user)
-        completed_blocks = get_completed_blocks(current_user)
-        finder = PathFinder(g.schema, answer_store, metadata, completed_blocks)
+        completed = get_completed_locations(current_user)
```
bitdivision: `completed_locations` ?

app/questionnaire/completed.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class CompletedItems:
+    """
+    An object that stores and updates references to sections and blocks
+    that have been completed, ready for serialisation via the Questionnaire Store.
+    """
+
+    def __init__(
+        self, blocks: List[Mapping] = None, sections: List[str] = None
```
bitdivision: `blocks` here holds a list of locations? I would rename this argument `block_locations`.

The docstring and self.blocks should also be renamed.

app/questionnaire/location.py
```diff
@@ -23,13 +23,13 @@ def __hash__(self):
         return hash(frozenset(self.__dict__.values()))
 
     @classmethod
-    def from_dict(cls, location_dict: Dict):
+    def from_dict(cls, location_dict: Mapping):
```
bitdivision, line 26: :+1:

app/views/dump.py
JamesGardiner: There is no functionality for getting all sections anymore, only for a single section. I can put this in if it's required.

app/questionnaire/path_finder.py
```diff
@@ -13,12 +18,18 @@
 
 
 class PathFinder:
-    def __init__(self, schema, answer_store, metadata, completed_blocks):
+    def __init__(
+        self,
+        schema: QuestionnaireSchema,
+        answer_store: AnswerStore,
+        metadata: Mapping,
+        completed: CompletedItems = CompletedItems(),
```
bitdivision: `completed` should default to `None` rather than an instance. Currently if we have two instances of PathFinder, they will both use the same instance of CompletedItems?

app/questionnaire/path_finder.py
```diff
@@ -212,12 +245,15 @@ def get_next_location(self, current_location):
 
             return routing_path[current_location_index + 1]
 
+        routing_path = self.build_path(current_section)
+        return routing_path[0]
+
     def _is_survey_completed(self, routing_path):
```
bitdivision: This doesn't necessarily represent survey completion anymore?

app/questionnaire/questionnaire_schema.py
```diff
@@ -14,75 +14,88 @@ def __init__(self, questionnaire_json, language_code=DEFAULT_LANGUAGE_CODE):
...
+        block = self.block(block_id)
+
+        try:
+            group_id = block['parent_id']
+        except TypeError:
+            return None
+
+        try:
+            section_id = self.group(group_id)['parent_id']
+        except TypeError:
```
bitdivision: I'm slightly confused by this, when is the group == `None`?

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -35,7 +35,7 @@ def save_new_list_item_answers(self, form, list_name):
         self.save_answers(form, False)
 
     def _update_questionnaire_store_with_form_data(self, form_data):
-        answer_ids_for_question = self._schema.get_answer_ids_for_question(
+        answer_ids_via_question = self._schema.answer_ids_via_question(
```
bitdivision: I'm unsure about the use of `via` in general, but the variable in this case should remain as `answer_ids_for_question` I think.

app/templating/summary_context.py
```diff
@@ -1,26 +1,31 @@
...
 from app.templating.summary.group import Group
 
 
-def build_summary_rendering_context(schema, sections, answer_store, metadata):
+def build_summary_rendering_context(
+    schema: QuestionnaireSchema,
+    answer_store: AnswerStore,
+    metadata: Mapping,
+    sections: List[Mapping] = None,
+) -> List:
```
bitdivision, line 14: Ooo, type annotations :grin:

app/views/questionnaire.py
```diff
@@ -280,15 +301,24 @@ def post_block_handler(
...
+    if _is_end_of_questionnaire(block):
         return submit_answers(routing_path, schema)
 
+    if not next_location:
+        path_finder.get_next_location(current_location, current_section)
     return redirect(next_location.url())
 
 
-def perform_list_action(
+def perform_list_action(  # pylint: disable=bad-continuation
```
bitdivision: I'm assuming this is because there are too many lines? Maybe disable this rule in the config, given that we have black?

bitdivision (changes requested):

app/data_model/questionnaire_store.py
JamesGardiner: Yes

app/views/dump.py
ajmaddaford: No, we have a `/dump/debug` endpoint, but that just dumps the contents of the questionnaire store. It would be useful to be able to view the routing path per section as an endpoint.

app/views/dump.py
ajmaddaford: Does this dump all answers, or only the answers for the current section?

app/views/questionnaire.py
ajmaddaford: Getting it from `current_location` would be better

app/questionnaire/path_finder.py
ajmaddaford: Sorry, I can't see what happens between calling `routing_path` and `build_path` that would make the path any different.

app/questionnaire/completed.py
JamesGardiner: Blocks in the init should remain blocks (it takes a list of blocks). I do agree though that the attribute blocks should be block_locations

app/views/dump.py
JamesGardiner: All answers

app/views/questionnaire.py
JamesGardiner: I was getting TypeErrors in the following tests:

TestQuestionnaireListCollector.test_invalid_block
TestQuestionnaireEndpointRedirects.test_get_invalid_questionnaire_location_redirects_to_latest

AttributeError is unnecessary

ajmaddaford: Those errors should be caught in the `validate_location` method rather than here.

app/questionnaire/completed.py
bitdivision: Ah, of course

app/views/dump.py
ajmaddaford: This doesn't work. I loaded the `test_section_summary` questionnaire and `/dump/submission` displays the first section answers when in the first section, but when you go to the second section it no longer displays answers. 

It also doesn't let me get past the first question in the second section - `house-type` - whichever answer I choose it just refreshes the page.

JamesGardiner: That is one of the schemas with the issues around skipping the last group in a section

app/views/questionnaire.py
JamesGardiner: This can go now

app/questionnaire/questionnaire_schema.py
JamesGardiner: When an incorrect block id is used.

app/questionnaire/path_finder.py
JamesGardiner: It probably should. It would be useful to have the full path followed by the user in a number of scenarios.

app/questionnaire/questionnaire_store_updater.py
JamesGardiner: If we don't update at this point, a user that uses 'save and come back' on a block won't be directed to that block after signing in

JamesGardiner: e.g. the test test_save_sign_out_complete_a_block_then_revisit_it fails

app/data_model/questionnaire_store.py
```diff
@@ -43,23 +46,28 @@ def set_metadata(self, to_set):
...
         json_data = json.loads(data, use_decimal=True)
-        completed_blocks = [
-            Location.from_dict(location_dict=completed_block)
-            for completed_block in json_data.get('COMPLETED_BLOCKS', [])
-        ]
+        completed = json_data.get('COMPLETED', {})
+        self.completed_items = CompletedItems(
+            completed.get('COMPLETED_BLOCKS', []),
+            completed.get('COMPLETED_SECTIONS', []),
+        )
```
ajmaddaford: `CompletedItems` internals are bleeding into `QuestionnaireStore` here. `CompletedItems` should be initialised with just `json_data.get('COMPLETED')`, which would also be more consistent with the other `QuestionnaireStore` properties:
```
self.completed_items = CompletedItems(json_data.get('COMPLETED'))
```

app/questionnaire/completed_items.py
```diff
@@ -0,0 +1,43 @@
...
+            self.sections = []
+
+    def serialise(self):
+        return {
+            'COMPLETED_BLOCKS': self.block_locations.copy(),
+            'COMPLETED_SECTIONS': self.sections.copy(),
+        }
+
+    def clear(self) -> None:
+        """Clears answers *in place*.
```
ajmaddaford: The comment is wrong (and probably isn't needed).

app/questionnaire/path_finder.py
```diff
@@ -13,45 +18,110 @@
...
-    @staticmethod
-    def _block_index_for_location(blocks, location):
-        return next(
-            (
-                index
-                for (index, block) in enumerate(blocks)
-                if block['id'] == location.block_id
-            ),
-            None,
+    def routing_path(self, current_section: Mapping) -> SectionRoutingPath:
```
ajmaddaford: We should rename `current_section` to `section` as this won't necessarily be the current section.

app/questionnaire/path_finder.py
```diff
@@ -164,62 +242,21 @@ def _remove_rule_answers(self, goto_rule, this_location):
...
         for location in routing_path[
             :-1
         ]:  # Don't evaluate end block (Summary or Confirmation)
-            if location not in self.completed_blocks:
+            if location not in self.completed_items.block_locations:
                 return False
-
         return True
 
     def _get_valid_section_summary(self, current_block_id, routing_path):
```
ajmaddaford: Is this method used?

app/questionnaire/path_finder.py
```diff
@@ -164,62 +242,21 @@ def _remove_rule_answers(self, goto_rule, this_location):
...
-            return routing_path[current_location_index + 1]
-
-    def _is_survey_completed(self, routing_path):
+    def _is_path_completed(self, routing_path):
         # Check all blocks on routing path are complete
         for location in routing_path[
             :-1
         ]:  # Don't evaluate end block (Summary or Confirmation)
-            if location not in self.completed_blocks:
+            if location not in self.completed_items.block_locations:
```
ajmaddaford: This won't work now that he routing path is per section, as a section does not have to finish on a Summary block.

app/questionnaire/path_finder.py
```diff
@@ -13,45 +18,110 @@
...
+        routing_path = self.routing_path(current_section)
+        blocks_remaining_on_path = [
+            block
+            for block in routing_path
+            if block not in completed_items.block_locations
+        ]
+
+        if not blocks_remaining_on_path:
+            try:
+                next_section = self.schema.get_next_section(current_section['id'])
```
ajmaddaford: I think path finder should be limited to a section. If there are no more blocks we could just return `None` and the `Router` could determine where to go next.

app/questionnaire/path_finder.py
```diff
@@ -13,45 +18,110 @@
...
+    ) -> Union[Location, None]:
+        """
+        Returns the previous 'location' to visit given a set of user answers
+        """
+
+        current_block = current_location.block_id
+        group = self.schema.group(block_id=current_block)
+
+        if self.is_first_block(group['id'], current_block):
+            return None
```
ajmaddaford: I think the desired behaviour is to have no previous link if it's the first block in a section, rather than the first block in a group.

app/questionnaire/path_finder.py
```diff
@@ -13,45 +18,110 @@
...
+        If Summary or SectionSummary is available then next location will be those.
+        :param current_block_location:
+        :return: The next location as a dict
+        """
+        routing_path = self.routing_path(current_section)
+        blocks_remaining_on_path = [
+            block
+            for block in routing_path
+            if block not in completed_items.block_locations
+        ]
```
ajmaddaford: This will loop over all blocks in the routing path, when all we really care about is the first incomplete location. It would be clearer as:

```
for block in routing_path:
	if block not in completed_items.block_locations:
		return block
```

Also, it would be clearer if `block` was renamed `location`.

app/questionnaire/path_finder.py
```diff
@@ -13,45 +18,110 @@
...
+        current_block = current_location.block_id
+        group = self.schema.group(block_id=current_block)
+
+        if self.is_first_block(group['id'], current_block):
+            return None
+
+        block = self.schema.block(current_block)
+        if self.schema.is_block_list_collector_child(current_block):
+            # If this is a list collector sub block, return the collector in the previous link
+            return Location(block_id=block['parent_id'])
```
ajmaddaford: This may be better placed somewhere else as locations of list collector children don't appear on the routing path.

app/questionnaire/path_finder.py
```diff
@@ -13,45 +18,110 @@
...
+        routing_path = self.routing_path(current_section)
+        current_location_index = PathFinder._current_location_index(
+            routing_path, current_location
         )
 
-    def build_path(self):
+        if current_location_index is not None and current_location_index != 0:
+            return routing_path[current_location_index - 1]
+
+        return None
```
ajmaddaford, line 47: Possibly clearer with a try:

```
try:
    current_location_index = path.index(current_location)
except ValueError:
	return None

if current_location_index != 0:
    return routing_path[current_location_index - 1]
```

app/questionnaire/path_finder.py
```diff
@@ -13,45 +18,110 @@
...
 
-        for group in self.schema.groups:
-            if group['id'] in first_groups:
-                first_block_in_group = self.schema.get_first_block_id_for_group(
-                    group['id']
-                )
+        try:
+            first_group_id = current_section['groups'][0]['id']
+        except TypeError:
+            return self._build_path(self.schema.json['sections'][0])
```
ajmaddaford: When can this happen? All sections should have at least one group.

app/questionnaire/path_finder.py
```diff
@@ -13,45 +18,110 @@
...
+        Use a block's id to check if it is the first block in a group.
         """
-        Visits all the blocks from a location forwards and returns path
-        taken given a list of answers.
+        return self.schema.first_block_id_for_group(group_id) == block_id
 
-        :param blocks: A list containing all block content in the survey
-        :param this_location: The location to visit, represented as a dict
-        :return: A list of locations followed through the survey
+    def _build_path(self, current_section: Mapping) -> SectionRoutingPath:
```
ajmaddaford: Same as earlier comment, we should rename `current_section` to `section` as this won't necessarily be the current section.

app/questionnaire/completed_items.py
```diff
@@ -0,0 +1,43 @@
...
+            self.block_locations = [
+                Location.from_dict(location_dict=block) for block in blocks
+            ]
+        else:
+            self.block_locations = []
+
+        if sections:
+            self.sections = sections
+        else:
+            self.sections = []
```
MebinAbraham: Maybe?
```python
self.sections = sections or []
```

Makefile
```diff
@@ -18,10 +18,16 @@ run:
 	pipenv run flask run
 
 dev-compose-up:
-	docker-compose -f docker-compose-dev.yml up -d
+	docker-compose -f docker-compose-dev-mac.yml up -d
+
+dev-compose-up-linux:
+	docker-compose -f docker-compose-dev-linux.yml up -d
```
iwootten, line 30: I think detecting the platform and switching here would be better, less to remember/doc.

app/helpers/path_finder_helper.py
```diff
@@ -4,31 +4,32 @@
...
 path_finder = LocalProxy(get_path_finder)
 
 
-def full_routing_path_required(function):
+def section_routing_path_required(function):
     @wraps(function)
     def wrap_function(*args, **kwargs):
-        routing_path = path_finder.get_full_routing_path()
+        routing_path = path_finder.routing_path(
+            g.schema.get_section_via_block_id(kwargs['block_id'])
```
iwootten: I was under the impression use of 'via' was to be removed.

app/forms/questionnaire_form.py
```diff
@@ -135,7 +133,7 @@ def _get_target_total_and_currency(self, calculation, question):
         if 'value' in calculation:
             return calculation['value'], question.get('currency')
 
-        target_answer = self.schema.get_answers(calculation['answer_id'])[0]
+        target_answer = self.schema.answers_by_answer_id(calculation['answer_id'])[0]
```
iwootten: Typically something without get in the codebase would indicate a property - where this is a function. I think the original method name is clearer here, or maybe the accessor for underlying property isn't needed and the property should be marked public.

app/helpers/form_helper.py
```diff
@@ -96,7 +96,7 @@ def get_mapped_answers(schema, answer_store, location):
 
     block_id = location.block_id
 
-    answer_ids = schema.get_answer_ids_for_block(block_id)
+    answer_ids = schema.answer_ids_for_block(block_id)
```
iwootten: Again with whether this is a property or a method, not obvious from the name.

app/questionnaire/questionnaire_schema.py
```diff
@@ -14,73 +15,97 @@ def __init__(self, questionnaire_json, language_code=DEFAULT_LANGUAGE_CODE):
...
-        questions = self.get_questions(answers[0]['parent_id'])
-        return self.get_block(questions[0]['parent_id'])
+    def answers_by_answer_id(self, answer_id):
+        """ Return answers matching answer id, including all matching answers inside
+        variants
+        """
+        return self._answers_by_id.get(answer_id)
 
-    def get_questions(self, question_id):
+    def questions(self, question_id):
```
iwootten: Again with the naming here, not sure why this has been changed.

app/questionnaire/questionnaire_schema.py
```diff
@@ -14,73 +15,97 @@ def __init__(self, questionnaire_json, language_code=DEFAULT_LANGUAGE_CODE):
...
+        if group:
+            return group['blocks'][0]['id']
+
     def blocks(self):
         return self._blocks_by_id.values()
 
-    @property
-    def answers(self):
-        return self._answers_by_id.values()
+    def block(self, block_id=None, answer_id=None):
```
iwootten: Noun type names - i.e blocks, groups, sections should really be properties of a class, though given the method behaviour hasn't changed (and is really doing some work as a look up) should probably be as they were in the original methods.

app/helpers/path_finder_helper.py
JamesGardiner: Done :+1:

app/questionnaire/questionnaire_schema.py
JamesGardiner: I'd prefer this to be two methods - 1. get_block_by_block_id and 2. get_block_by_answer_id - does that sound good?

app/questionnaire/path_finder.py
```diff
@@ -71,17 +141,25 @@ def build_path(self):
                     blocks, block_index, this_location, path
                 )
```
ajmaddaford: This might be out of scope for this pull request, but feels like it could be simplified. It's not clear why the full list of all blocks for all groups are passed to `_build_path_within_group`, rather than just the blocks for the current group.

app/questionnaire/questionnaire_schema.py
```diff
@@ -14,73 +14,90 @@ def __init__(self, questionnaire_json, language_code=DEFAULT_LANGUAGE_CODE):
...
+
+    def get_section_via_group_id(self, group_id: str):
+        group = self.group(group_id)
+
+        try:
+            return self.get_section_via_section_id(group['parent_id'])
+        except KeyError:
+            return None
+
+    def get_section_via_block_id(self, block_id: str):
```
ajmaddaford: This is inconsistent with the `group` and `block` methods, which take multiple mutually exclusive parameters rather than separate method names. Whichever approach we decide, we should be consistent.

app/questionnaire/questionnaire_schema.py
iwootten: I think that a method named get_block is going to expect block_id as an arg, and we generally use 'for' to indicate something different - so it probably should just be get_block, get_block_for_answer_id if they're being broken apart. That applies to other accessors too, so get_section_for_section_id could be shortened to just get_section, for instance.

We probably should agree a style guide on the team so we're all happy about the style we're using going forward so we can minimise some of this.

JamesGardiner: Ok, in that case, are you fine for me to leave it as is for now, and update following a style guide upddate?

Makefile
JamesGardiner: This has been merged as a separate PR and will disappear following rebase

app/questionnaire/router.py
```diff
@@ -55,13 +59,35 @@ def _get_survey_redirect_location(self):
...
+        }
+
+        section_routing_path = path_finder.routing_path(first_incomplete_section)
+
+        for group in first_incomplete_section['groups']:
+            for block in group['blocks']:
+                if (
+                    block['id'] not in completed_blocks
+                    and Location(block['id']) in section_routing_path
+                ):
```
ajmaddaford: Couldn't we just loop over `section_routing_path` here and return the first non complete location?

app/views/questionnaire.py
```diff
@@ -189,8 +191,21 @@ def get_block_handler(
...
+    if current_location in questionnaire_store.completed_items.block_locations:
+        questionnaire_store_updater.remove_completed_block(current_location)
+
+    try:
+        current_section_id = schema.get_section_via_block_id(block_id)['id']
+    except TypeError:
+        current_section_id = schema.json.get('sections')[0]['id']
+
+    if current_section_id in questionnaire_store.completed_items.sections:
+        questionnaire_store_updater.remove_completed_section(current_section_id)
```
ajmaddaford: State should not be changed on a GET request. Why does a GET need to remove completeness?

app/views/questionnaire.py
```diff
@@ -280,10 +301,21 @@ def post_block_handler(
             return redirect(next_url)
 
     questionnaire_store_updater.save_answers(form)
-    next_location = path_finder.get_next_location(current_location=current_location)
 
-    if _is_end_of_questionnaire(block, next_location):
-        return submit_answers(routing_path, schema)
+    if _user_posted_last_block_on_section_routing_path(
```
ajmaddaford: The `_user_posted_` bit is unnecessary here.

app/views/questionnaire.py
```diff
@@ -280,10 +301,21 @@ def post_block_handler(
...
+        block_id, questionnaire_store.completed_items
+    ):
+        current_section = schema.get_section_via_block_id(block_id)
+        questionnaire_store_updater.add_completed_section(current_section['id'])
```
ajmaddaford: This logic should be done in the questionnaire store updater when calling `save_answers`.

ajmaddaford (dismissed):

app/questionnaire/questionnaire_schema.py
iwootten: Given it's such a simple change highlighted here, I think it should probably be addressed as part of this PR.

app/questionnaire/path_finder.py
JamesGardiner: This isn't new, just moved to make public methods the first methods in this class (I find it clearer).

app/questionnaire/questionnaire_schema.py
JamesGardiner: @iwootten has requested changes _re_ method names previously. I'll make these consistent by not having mutually exclusive arguments, and instead having names such as `get_section` (when a section id is supplied) and `get_section_for_block_id` when a block id is supplied.

app/questionnaire/path_finder.py
JamesGardiner: No - I'll remove it

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -1,5 +1,17 @@
...
 from app.data_model.answer_store import Answer
+from app.questionnaire.location import Location
+from app.questionnaire.routing_path import SectionRoutingPath
+
+
+def _is_last_block_on_path(
+    block_id: str, routing_path: SectionRoutingPath
+) -> bool:
+    try:
+        return block_id == routing_path[-1].block_id
```
iwootten:
```suggestion
        return hasattr(routing_path[-1], 'block_id') and block_id == routing_path[-1].block_id
```

Using hasattr would save the need for the try here.

app/questionnaire/path_finder.py
JamesGardiner: Tech debt card?

app/questionnaire/path_finder.py
JamesGardiner: It isn't used anywhere - I've removed it.

app/questionnaire/router.py
JamesGardiner: Yes

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -81,6 +100,31 @@ def remove_answer_ids(self, answer_ids: List):
...
+
+    def add_completed_section(self, section_id):
+        self._questionnaire_store.completed_items.sections.append(section_id)
+
+    def remove_completed_section(self, section_id):
+        self._questionnaire_store.completed_items.sections.remove(section_id)
+
+        section_summary_id = self._schema.get_section_summary(section_id)
+        if section_summary_id:
+            self._questionnaire_store.completed_items.block_locations.remove(
```
MebinAbraham: This method makes a completed section incomplete on a GET request to any block within the completed section. This isn't ideal. Due to this behavior, this is causing some 500s when clicking back/using change links, etc. Happy to discuss the behaviors.

This would also cause weird behavior with hub and spoke as the user may visit a block to view details but then return to the hub. Currently, a section is only marked a complete on a POST request from the section summary (last block).

MebinAbraham (changes requested):

app/views/questionnaire.py
```diff
@@ -308,16 +297,22 @@ def post_block_handler(
...
-    questionnaire_store_updater.save_blocks(block_id, routing_path)
-    questionnaire_store.add_or_update()
-
     if _is_end_of_questionnaire(block):
         full_routing_path = _get_full_routing_path(schema)
         return submit_answers(full_routing_path, schema)
 
-    # recreate routing path now that we have processed the answers
-    routing_path = path_finder.routing_path(section)
+    questionnaire_store_updater.save_answers(form)
```
jonnyshaw89: 👍

jonnyshaw89:
there are a lot of calls to `add_or_update` added in this PR. i would double check that the state isn’t written to the db on each of those calls as they will kill the performance.

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -18,20 +18,50 @@ def __init__(self, current_location, schema, questionnaire_store, current_questi
         self._list_store = self._questionnaire_store.list_store
         self._completed_items = self._questionnaire_store.completed_items
 
-    def save_answers(self, form, save_completed_blocks=True):
+    def save(self):
+        self._questionnaire_store.add_or_update()
```
jonnyshaw89: do we need dirty checking? so we only action the save if something has actually changed?

app/questionnaire/path_finder.py
```diff
@@ -13,45 +18,54 @@
...
-    def build_path(self):
+    def routing_path(self, section: Mapping) -> SectionRoutingPath:
         """
-        Visits all the blocks from a location forwards and returns path
-        taken given a list of answers.
-
-        :param blocks: A list containing all block content in the survey
-        :param this_location: The location to visit, represented as a dict
-        :return: A list of locations followed through the survey
+        Returns block ids it is possible to visit in this section.
```
MebinAbraham: The comment doesn't make sense?. Also `_build_path` is only used in this one method, so do we need this method?

app/questionnaire/router.py
```diff
@@ -1,67 +1,124 @@
...
+            return last_block_location
 
-    def _get_default_location(self):
-        """
-        Calculates the next location if the current location is invalid
-        or if the current location is valid but not accessible.
-        :return: Location to direct to.
-        """
+        if self.is_survey_complete():
+            all_sections = [section['id'] for section in self._schema.get_sections()]
```
MebinAbraham: Think the var should be  exactly what the method needs. Also should be more efficient that list comp.
```python
last_section_id = list(self._schema.get_sections())[-1]['id']
```
`final_block_id` should be `last_block_id` for consistency.

tests/integration/views/test_questionnaire.py
```diff
@@ -18,7 +18,7 @@ def setUp(self):
         self._application_context.push()
 
         storage = Mock()
-        data = {'METADATA': 'test', 'ANSWERS': [], 'COMPLETED_BLOCKS': []}
+        data = {'METADATA': 'test', 'ANSWERS': [], 'blocks': []}
```
MebinAbraham: Is this meant to be `blocks`?

MebinAbraham (commented):
Thought I'd leave some petty comments before testing it functionally. 👀

app/views/questionnaire.py
```diff
@@ -109,6 +109,21 @@ def before_post_submission_request():
...
+
+
+@questionnaire_blueprint.route('/', methods=['GET'])
+@login_required
+@with_questionnaire_store
+@with_schema
+def get_questionnaire(schema, questionnaire_store):
+    router = Router(schema, completed_items=questionnaire_store.completed_items)
+    redirect_location = router.get_first_incomplete_location_in_survey()
+    return redirect(redirect_location.url())
```
MebinAbraham, line 119: This ends up in a 500 sometimes as `router.get_first_incomplete_location_in_survey()` can return `None` and you can't call `.url` on `NoneType`.

MebinAbraham (changes requested):

app/views/questionnaire.py
ajmaddaford: What scenario did you get the 500 for?

ajmaddaford: Fixed in latest commit

app/questionnaire/path_finder.py
ajmaddaford: Fair point. I've renamed `_build_path` to `routing_path`.

app/questionnaire/router.py
ajmaddaford: As it's done in a few places I've added a `get_section_ids` method to questionnaire_schema.py.

ajmaddaford: And renamed `final_block_id` to `last_block_id`

tests/integration/views/test_questionnaire.py
ajmaddaford: Renamed to 'COMPLETED'

bitdivision:
I've been trying to break this, and not having much luck. 

Did just find that the `test_multiple_answers` schema goes straight to the summary upon launch though

EDIT: Ignore this, looks like launcher must have reused the same details.

MebinAbraham (commented):
You previously mentioned some edge cases with the section summary schema but currently stepping through `test_section_summary.json` errors as it is infinitely redirecting.

- Launch schema and keep clicking continue until it tries to take you to the summary.

ajmaddaford:
@MebinAbraham Good spot. That's broken as there are skip conditions on both groups in the second section. This can generate an empty routing path, redirecting you to the summary. As the second section is not in completed sections, the summary redirects you back to the second section. And so on. 

For now I'm going to remove the skip conditions as I think we need to introduce skip conditions at a section level to solve this properly.

app/views/dump.py
```diff
@@ -25,22 +37,30 @@ def dump_debug():
...
     response = {
         'submission': convert_answers(schema, questionnaire_store, routing_path)
     }
-    return json.dumps(response, for_json=True)
+    return json.dumps(response, for_json=True), 200
+
+
+def _get_full_routing_path(schema):
+    sections = schema.get_sections()
+    return [path_finder.routing_path(section) for section in sections]
```
bitdivision: I think it would be clearer if this returned a single merged list. I don't see much value in per section paths if you've asked for the full path.

app/questionnaire/router.py
```diff
@@ -1,67 +1,141 @@
...
+    def _get_allowable_path(self, routing_path):
         """
-        Ensure that the user can't access the end of the survey if the survey is not complete.
-        :return: Location to direct to if current location is not accessible.
+        The allowable path is the completed path plus the next location
         """
-        if not self._is_current_location_valid():
-            return None
+        allowable_path = []
+        for location in routing_path:
```
iwootten, line 109:
```suggestion
        for location in routing_path:
            allowable_path.append(location)
            if location not in self._completed_items.block_locations:
                return allowable_path
        return allowable_path
```

bitdivision (dismissed):
I've done a fair bit of testing and not found any issues not already raised.

:+1:

Makefile
```diff
@@ -16,21 +16,21 @@ test-functional:
 test-schemas:
 	pipenv run ./scripts/test_schemas.sh
 
-load_templates:
+load-templates:
```
MebinAbraham, line 19: 👍

app/questionnaire/questionnaire_schema.py
```diff
@@ -14,73 +15,79 @@ def __init__(self, questionnaire_json, language_code=DEFAULT_LANGUAGE_CODE):
...
-    @property
-    def answers(self):
-        return self._answers_by_id.values()
+    def get_section_for_block_id(self, block_id):
+        block = self.get_block(block_id)
 
-    def get_section(self, section_id):
-        return self._sections_by_id.get(section_id)
+        try:
+            group_id = block['parent_id']
```
MebinAbraham: Personal preference I guess but just to be consistent?. Most places we don't use exception control flow. You could explicitly `return None` also but we don't in most places. 🤷‍♂ 
```python
    def get_section_for_block_id(self, block_id):
        block = self.get_block(block_id)
        group_id = block.get('parent_id')

        if group_id:
            section_id = self.get_group(group_id)['parent_id']
            if section_id:
                return self.get_section(section_id)
```

app/questionnaire/path_finder.py
```diff
@@ -13,45 +18,48 @@
...
-        return next(
-            (
-                index
-                for (index, block) in enumerate(blocks)
-                if block['id'] == location.block_id
-            ),
-            None,
-        )
+    def is_path_complete(self, path):
+        if self.get_first_incomplete_location_on_path(path):
```
bitdivision: `return not self.get_first_incomplete_location_on_path(path)` ?

app/questionnaire/router.py
```diff
@@ -1,67 +1,141 @@
...
-        expected_location = self._get_first_incomplete_location_in_survey()
-        block = self._schema.get_block(self._current_location.block_id)
-        cannot_access_survey_end = self._current_location != expected_location
+        return allowable_path
 
-        if block['type'] in ['Confirmation', 'Summary'] and cannot_access_survey_end:
-            return expected_location
+    def _get_incomplete_section_ids(self):
+        all_section_ids = self._schema.get_section_ids()
+        completed_sections = {section for section in self._completed_store.sections}
```
MebinAbraham: Can we get rid of this var and avoid the extra comprehension?
`self._completed_store.sections` are unique but that doesn't matter anyway?

Makefile
```diff
@@ -16,21 +16,21 @@ test-functional:
...
 	pipenv run ./scripts/load_templates.sh
 
-run: load_templates
+run: build
 	ln -sf .development.env .env
 	pipenv run flask run
 
 dev-compose-up:
-	docker-compose -f docker-compose-dev.yml up -d
+	docker-compose -f docker-compose-dev-mac.yml up -d
```
MebinAbraham, line 27: The file needs renaming, it is still called `docker-compose-dev.yml`

Makefile
```diff
@@ -16,21 +16,21 @@ test-functional:
...
 
 dev-compose-up-linux:
 	docker-compose -f docker-compose-dev-linux.yml up -d
 
 dev-compose-down:
-	docker-compose -f docker-compose-dev.yml down
+	docker-compose -f docker-compose-dev-mac.yml down
```
MebinAbraham, line 33: The file needs renaming, it is still called docker-compose-dev.yml

app/data_model/completed_store.py
```diff
@@ -0,0 +1,71 @@
...
+    """
+    An object that stores and updates references to sections and blocks
+    that have been completed, ready for serialisation via the Questionnaire Store.
+    """
+
+    def __init__(self, completed: Mapping = None) -> None:
+        """ Instantiate a CompletedStore object that tracks completed blocks and sections.
+
+        Args:
+            completed: A dict containing completed block  locations and section ids
```
MebinAbraham:
```suggestion
            completed: A dict containing completed block locations and section ids
```

app/questionnaire/questionnaire_schema.py
```diff
@@ -176,6 +180,11 @@ def _parse_schema(self):
         self._answers_by_id = self._get_answers_by_id()
         self.error_messages = self._get_error_messages()
 
+    def _get_section_id_for_list_block(self, block):
```
MebinAbraham, line 178: I can't see any usages for this?

app/questionnaire/router.py
```diff
@@ -1,67 +1,141 @@
...
-    def _is_current_location_valid(self):
-        """
-        Checks whether the current location is within the routing path.
-        :return: boolean
-        """
-        return self._current_location in self._routing_path
+        # If the section is complete and contains a SectionSummary, return the SectionSummary location
+        if (
+            last_block_type == 'SectionSummary'
+            and current_block_type != last_block_type
```
MebinAbraham, line 40: Sorry if I have misunderstood but `current_block_type == last_block_type` for `SectionSummary` can never happen on a POST unless you had a SectionSummary followed by another SectionSummary which wouldn't be the case?

package.json
```diff
@@ -17,7 +17,7 @@
     "test_cypress": "node tests/functional/cypress.js",
     "cypress": "cypress open",
     "generate_pages": "rm -rf ./tests/functional/generated_pages && cd tests/functional &&  pipenv run ./generate_pages.py ../../data/en/ ./generated_pages -r '../../base_pages' && cd ../..",
-    "format": "gulp format:json && pipenv run black app tests"
+    "format": "gulp format:json && gulp format:census && pipenv run black app tests"
```
MebinAbraham: `format:census` was removed previously.

app/questionnaire/questionnaire_store_updater.py
ajmaddaford: Good spot. I've converted CompletedItems into CompletedStore and added is_dirty checking to it and the ListStore. The updater now checks the is_dirty status of all three stores before deciding to save.

app/questionnaire/router.py
```diff
@@ -109,6 +109,10 @@ def _get_allowable_path(self, routing_path):
         """
         allowable_path = []
         for location in routing_path:
+            allowable_path.append(location)
+            if location not in self._completed_items.block_locations:
+                return allowable_path
+        return allowable_path
             if location in self._completed_store.block_locations:
                 allowable_path.append(location)
```
MebinAbraham: Multiline suggestions don't really work 😄

ajmaddaford: Already suggested by @iwootten 😛

MebinAbraham: I was 👉 at it inserting Ian's suggestion instead of replacing with the older commit. 
🙃 
https://github.com/ONSdigital/eq-survey-runner/blob/bd34ab2a000968d5fcc7906f1d4f1b55cca98145/app/questionnaire/router.py#L110-L122

app/questionnaire/router.py
ajmaddaford: It can because there is a submit button on a SectionSummary. Without this check you get stuck in a loop on a SectionSummary.

app/questionnaire/questionnaire_schema.py
ajmaddaford: It's used on line 38 - `section_id = self._get_section_id_for_list_block(block)`

app/views/dump.py
ajmaddaford: Done

app/questionnaire/router.py
MebinAbraham: Ah forgot about that button. 👍

app/questionnaire/questionnaire_schema.py
MebinAbraham: Weird, IDE must have been lazy.

app/questionnaire/router.py
ajmaddaford: Oh, yeah, that didn't work so well 😕

app/questionnaire/questionnaire_schema.py
ajmaddaford: I've refactored this method to be a lot clearer.

app/questionnaire/path_finder.py
ajmaddaford: Done

app/questionnaire/router.py
ajmaddaford: Heh, good spot. I've removed the `completed_sections` variable.

MebinAbraham (approved):
Looks good to me 👍 🎉

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -15,24 +16,58 @@ def __init__(self, current_location, schema, questionnaire_store, current_questi
         self._questionnaire_store = questionnaire_store
         self._answer_store = self._questionnaire_store.answer_store
         self._list_store = self._questionnaire_store.list_store
-
-    def save_answers(self, form, save_completed_blocks=True):
+        self._completed_store = self._questionnaire_store.completed_store
+
+    def save(self):
+        if self.is_dirty():
```
jonnyshaw89, line 22: What a great idea 😄, this will save lots of unnecessary db writes. 👍

tests/functional/spec/section_summary.spec.js
```diff
@@ -31,11 +32,11 @@ describe('Section Summary', function() {
     it('When I continue on the section summary page, Then I should be taken to the next section', function() {
       return browser
         .click(PropertyDetailsSummaryPage.submit())
-        .getUrl().should.eventually.contain(FinalSummaryPage.pageName);
+        .getUrl().should.eventually.contain(HouseType.pageName);
```
jonnyshaw89, line 35: Why has this route changed?

tests/functional/spec/section_summary.spec.js
```diff
@@ -31,11 +32,11 @@ describe('Section Summary', function() {
...
     });
   });
 
-  describe('Given I start a Test Section Summary survey and complete to Final Summary', function() {
+  describe.skip('Given I start a Test Section Summary survey and complete to Final Summary', function() {
```
jonnyshaw89: intentional skip? if so, why?

tests/functional/spec/section_summary.spec.js
ajmaddaford: The section summary schema had a section with two groups that could be skipped, resulting in an empty section, which doesn't work. The skip conditions were removed on the understanding this is not something we should be doing - skip conditions need to be added to sections to effectively fix this.

tests/functional/spec/section_summary.spec.js
ajmaddaford: Not intentional, I re-enabled it and it highlighted a change in behaviour. 

I've amended the tests to reflect the new behaviour - the difference to previous behaviour is when clicking edit from a final summary and then submit on the answer, the user is directed to the section summary (if the section is complete), rather than back to the final summary.

jonnyshaw89 (approved):

app/questionnaire/path_finder.py
```diff
@@ -13,45 +18,54 @@
...
-            None,
-        )
+    def is_path_complete(self, path):
+        return not self.get_first_incomplete_location_on_path(path)
 
-    def build_path(self):
-        """
-        Visits all the blocks from a location forwards and returns path
-        taken given a list of answers.
+    def get_first_incomplete_location_on_path(self, path):
```
iwootten: Could a type hint help this method name be less wordy? It's a collection of locations being passed in right?

ajmaddaford: I don't think so as `path` is a well understood concept in runner. We could drop the `_on_path` bit so it becomes `get_first_incomplete_location`?

iwootten: Sounds good.

pricem14pc (approved):

PR #2097, opened 2019-05-23
Add Wales Census Individual Schema in Welsh

iwootten:
This adds a translated version of the current wales census individual schema in welsh. This has been manually created for now until we determine a better build process for schemas generated from translations.

### How to review 

Check the schema appears to be correct. Check you're able to run the new schema when using the cy region code. Ensure that build process continues as normal with the new cy directory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -91,7 +91,7 @@ requirements.txt
 .screenshots
 
 # Ignore built schemas
-data/
+data/en/
 
 # Ignore any generated pages for the functional tests
 tests/functional/generated_pages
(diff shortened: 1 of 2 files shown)


berroar:
I can run the new schema when using the cy code, looks OK but I noticed the Guidance tab for the establishment-position page still appears in English.

iwootten:
> establishment-position page

Yes. This is one of two missing messages we're aware of at the moment - good spot though.

bambrp (approved):

MebinAbraham (approved):
Some of the guidance and definition text is not translated. But you mentioned you are aware of some of them. Not sure if all of them are captured. But the majority is as expected. 👍
@iwootten if you are happy that the known issues are expected than I'm happy to merge. There were about 20 strings I noticed for content that was in English.

ajmaddaford:
Merging this as it's blocking all other pull requests. We'll create a release once this merges and any fixes can be done as a branch off that tag.

PR #2100, opened 2019-05-24
Remove direction from voluntary questions

berroar:
Removing 'Select one option only' from voluntary questions due to accessibility problems. 

### How to review 
Check that the 'Select one option only' labels have been removed from the schema for the questions listed on the Trello card:
[Trello](https://trello.com/c/yvhpHMgC/2995-remove-direction-from-voluntary-questions)


bitdivision (approved):

bitdivision:
You'll need to squash these commits before this can go in @berroar 

If you're not familiar with the process, you can do `git rebase -i origin/v3` and then change the `pick`s to `squash`s as per https://www.devroom.io/2011/07/05/git-squash-your-latests-commits-into-one/ 
You'll then need to `git push -f`

bambrp (approved):

PR #2108, opened 2019-05-30
Allow managed google tls cert

bitdivision:
Allow the use of a managed SSL certificate which has been provisioned in GCP.

This required changing the template to include more of the stuff that was in values.yaml. I don't think that's particularly bad here, but happy to discuss.

### How to review 
- Deploy an environment to google using a managed SSL certificate (as per whitelodge) and ensure that the ingress service yaml does not include the tls secret.
- Deploy a dev environment (with no extra configuration) and ensure that the tls secret is used in the ingress service yaml

### DIFF
--- a/k8s/deploy_app.sh
+++ b/k8s/deploy_app.sh
@@ -7,10 +7,10 @@ if [[ -z "$1" ]]; then
   exit 1
 fi
 
-#PROJECT_ID=$1
 SUBMISSION_BUCKET_NAME=$1
 DOCKER_REGISTRY="${2:-eu.gcr.io/census-eq-ci}"
 IMAGE_TAG="${3:-latest}"
+MANAGED_CERTIFICATE_NAME="${4:-}"
 
 helm tiller run \
     helm upgrade --install \
@@ -18,4 +18,5 @@ helm tiller run \
     k8s/helm \
     --set submissionBucket=${SUBMISSION_BUCKET_NAME} \
     --set image.repository=${DOCKER_REGISTRY}/eq-survey-runner \
-    --set image.tag=${IMAGE_TAG}
+    --set image.tag=${IMAGE_TAG} \
+    --set ingress.managedCertificateName=${MANAGED_CERTIFICATE_NAME}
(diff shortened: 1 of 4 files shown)


k8s/helm/templates/ingress.yaml
```diff
@@ -8,18 +8,17 @@ metadata:
...
   backend:
     serviceName: {{ .Chart.Name }}
     servicePort: 80
-{{- if .Values.ingress.tls }}
+{{- if .Values.ingress.tlsSecretName }}
   tls:
-  {{- range .Values.ingress.tls }}
-    - secretName: {{ .secretName }}
-  {{- end }}
+    - secretName: {{ .Values.ingress.tlsSecretName }}
```
ajmaddaford: We could remove the `tlsSecretName` variable and just use `tls-cert` here:

```
{{- if not .Values.ingress.managedCertificateName }}
  tls:
    - secretName: tls-cert
{{- end }}
```

ajmaddaford (changes requested):

k8s/helm/templates/ingress.yaml
bitdivision: Yep, that makes things simpler

k8s/helm/values.yaml
```diff
@@ -17,10 +17,9 @@ service:
 
 ingress:
   enabled: true
-  annotations:
-    kubernetes.io/ingress.global-static-ip-name: survey-runner
-  tls:
-    - secretName: tls-cert
+  # Use a managedCertificateName when terraform is used to create a managed certificate
+  # This should be left undefined when using a kubernetes secret for the certificate
```
ajmaddaford: Minor, but this shouldn't mention Terraform as it doesn't matter what created the managed certificate.

k8s/deploy_app.sh
```diff
@@ -7,15 +7,16 @@ if [[ -z "$1" ]]; then
   exit 1
 fi
 
-#PROJECT_ID=$1
 SUBMISSION_BUCKET_NAME=$1
 DOCKER_REGISTRY="${2:-eu.gcr.io/census-eq-ci}"
 IMAGE_TAG="${3:-latest}"
+MANAGED_CERTIFICATE_NAME="${4:-}"
```
MebinAbraham, line 13: Update the README?

ajmaddaford (approved):

MebinAbraham (approved):

PR #2110, opened 2019-06-03
Add Year Check

iwootten:

YearCheck was removed as part of https://github.com/ONSdigital/eq-survey-runner/pull/2086 and is currently allowing invalid years to be entered with less than 4 digits.

### How to review 

Confirm with the date tests if dates with less than 4 digits can no longer be entered.

### DIFF
--- a/app/validation/validators.py
+++ b/app/validation/validators.py
@@ -1,6 +1,7 @@
 from datetime import datetime
 from decimal import Decimal, InvalidOperation
 
+import re
 import flask_babel
 from flask_babel import ngettext
 from babel import numbers
@@ -224,7 +225,7 @@ class DateCheck:
 
     def __call__(self, form, field):
 
-        if not form.data:
+        if not form.data or not re.match(r'\d{4}$', str(form.year.data)):
             raise validators.StopValidation(self.message)
 
         try:
--- a/tests/app/validation/test_date_check_validator.py
+++ b/tests/app/validation/test_date_check_validator.py
@@ -25,6 +25,7 @@ class TestDateCheckValidator(unittest.TestCase):
 
         mock_form = Mock()
         mock_form.data = '2016-12-'
+        mock_form.year.data = '2016'
 
         mock_field = Mock()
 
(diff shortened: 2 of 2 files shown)


app/validation/validators.py
```diff
@@ -218,6 +219,16 @@ def __call__(self, form, field):
             raise validators.StopValidation(self.message)
 
 
+class YearCheck:
```
MebinAbraham: Should the check be done as part of `DateCheck`, because I don't think a DateCheck vs YearCheck is very clear?.

MebinAbraham: Should the check be done as part of DateCheck, because I don't think a DateCheck vs YearCheck is very clear?.

app/validation/validators.py
```diff
@@ -218,6 +219,16 @@ def __call__(self, form, field):
...
+    def __init__(self, message=None):
+        self.message = message or error_messages['INVALID_DATE']
+
+    def __call__(self, form, field):
+        # Must be 4 digits long
+        if not re.match(r'\d{4}$', str(form.year.data)):
```
MebinAbraham: This would allow `0001` for the year, which isn't ideal?

MebinAbraham (changes requested):

app/validation/validators.py
jonnyshaw89: Agree with Mebo, we should do this check in `DateCheck` as the fix applies to all dates not just years

app/validation/validators.py
iwootten: 0001 is still a valid 4 digit year - unless the question specifies a minimum date, I can't see why this would be blocked as a date.

app/validation/validators.py
iwootten: Fair enough.

app/validation/validators.py
ajmaddaford: I've spoken with @charlotte-hirst about this. This is fine as the check is primarily so people don't accidentally enter a 2 digit year without realising. Another card will be raised to consider making this a format validation error rather than an invalid date error.

ajmaddaford (approved):

MebinAbraham (approved):

PR #2112, opened 2019-06-06
Fix descriptions on mutually exclusive checkbox and prepare for "checked" parameter

bameyrick:
Fixes descriptions not showing on the mutually exclusive checkbox, and prepares to accept a "checked" parameter. I can't seem to see if runner stores if a user selects the mutually exclusive option.

### DIFF
--- a/data-source/json/test_mutually_exclusive.json
+++ b/data-source/json/test_mutually_exclusive.json
@@ -64,6 +64,7 @@
                             "type": "Checkbox",
                             "options": [{
                                 "label": "I prefer not to say",
+                                "description": "Some description",
                                 "value": "I prefer not to say"
                             }]
                         }]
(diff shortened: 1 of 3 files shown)


bameyrick:
By upgrading from design system 6.0.1 to 8.0.4 this has caused tests to fail, however I've already fixed this in #2090, so ideally that needs merging first.

bitdivision (approved):
Seems to work :+1:

iwootten (changes requested):
I'm unable to get this description to show when adding it to test_mutually_exclusive.json. I wonder if it might be an idea to add an example in there as part of the PR ?

bitdivision:
Worked for me with: 
```                        
                        "id": "checkbox-exclusive-answer",
                        "mandatory": false,
                        "type": "Checkbox",
                        "options": [{
                            "label": "I prefer not to say",
                            "description": "Some description",
                            "value": "I prefer not to say"
                        }]

```

But I would agree, the description should be added to the test schema.

iwootten:
@bitdivision Looks like I hadn't rebuilt my data sources - is working now.

bambrp (approved):
Works for me

iwootten (approved):

PR #2113, opened 2019-06-07
Remove repeated bullet point from stocks form types.

samiwel:
There was a bullet point which was repeated several times throughout the survey. This change removes the bullet point in all include and exclude lists.

### How to review 
Ensure that the following bullet point appears only once in the following form types:

- 0001
- 0002	
- 0057
- 0058
- 0061
- 0070

> All stocks owned and either held by you or currently in transit within the United Kingdom.


tgandrews (approved):
LGTM and works locally

But I must admit I'm not sure where the duplication was

rmccar (approved):

PR #2114, opened 2019-06-07
Fix stocks form types content.

samiwel:
A number of stocks form types were missing whitespace between a closing parenthesis and the next word. This change fixes this in all known stocks form types.

### How to review 
Ensure content looks correct now.

### DIFF
--- a/data/en/stocks_0001.json
+++ b/data/en/stocks_0001.json
@@ -41,7 +41,7 @@
                                                 "Raw materials and components purchased for incorporation into products for sale",
                                                 "Consumable stores",
                                                 "Semi-processed goods",
-                                                "Office supplies (including stationery)and packaging materials",
+                                                "Office supplies (including stationery) and packaging materials",
                                                 "Duty for dutiable goods held out of bond",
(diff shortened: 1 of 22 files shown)


data/en/stocks_0001.json
```diff
@@ -504,9 +504,9 @@
                                             "value": "Change of business structure, merger or takeover"
                                         },
                                         {
-                                            "label": "One off increase in stocks",
+                                            "label": "One-off increase in stocks",
                                             "q_code": "146f",
-                                            "value": "One off increase in stocks"
+                                            "value": "One-off increase in stocks"
```
tgandrews, line 505: Do they want the value changing as well? As I can see this is different in the dump submission

tgandrews (approved):
LGTM one question but I presume its fine

data/en/stocks_0001.json
samiwel: 🤷‍♂

samiwel: I assume it's okay as long as the q-codes remain the same.

rmccar (approved):

PR #2104, opened 2019-05-29
Content tweaks for NISRA (typo, school address, quals route)

bambrp:
### What is the context of this PR?
* 1 year ago address - current response option "Student term time in the UK" change to Student term time address in the UK
* Routing wrong on quals- if any yes answers to nvq, gcse, a-levels, degree, skip other-quals and onto apprenticeship. Otherwise show other-quals then apprenticeship.
* Missing two questions https://overflow.io/s/30TRR3/?node=739ef6db & https://overflow.io/s/30TRR3/?node=21a93787 which is the place of study and how to travel to school for under 16s

I've also made a change to the Over16 lib test, changing it to "less than or equal", to capture people who have turned 16 that day- this does bring up a corner case of time zones, but this is hopefully a better, not perfect, option.

Had to de-share the carer block, due to it not longer having the same routing rules across all regions.

### How to review 
Build schema, run through and test.

PR #2118, opened 2019-06-11
Switch to summary-type census submission block

bambrp:
Switched from a fixed text based confirmation screen to an automatic summary as the survey submission screen.

### How to review 
Checkout branch, run through an individual census (eng/wal or ni), and check final summary page reflects the answers you provided.


iwootten (approved):

iwootten (changes requested):
This seems to be applied to wls/nir, where the card says it should be for eng?

MebinAbraham (commented):
If both surveys now don't use confirmation, `confirmation.json` file can be removed?

bambrp:
> If both surveys now don't use confirmation, `confirmation.json` file can be removed?

I don't see any other uses of it, so it could- I just wasn't sure if that was premature, and if we'd be happy to have the text content of it easily to hand, but I suppose that's what histories are for.

iwootten (approved):
Approving this based on @charlotte-hirst saying it can be applied to all versions.

MebinAbraham:
@bambrp I don't see a reason to keep it if it is not used. Git history is always here if it is ever needed I guess.

berroar (approved):
Tested the changes, all looks okay to me.

bitdivision:
@MebinAbraham removed as part of https://github.com/ONSdigital/eq-survey-runner/pull/2104/files

MebinAbraham (approved):

PR #2117, opened 2019-06-11
Updating routing rules for individual census.

berroar:
### What is the context of this PR?
Adding new routing rules for the individual census (ENG/WLS). 

Summary of the changes:
1. Routing age 15 years and under past the Marital status questions (currently 16 & under)
2. Routing age 4 years and under past the Education questions
3. Routing age 3 years and under past the Language questions
4. Routing a child under 1 year past the Address one year ago question
5. Routing age 4 and under past the Caring questions

### How to review 
Build the schema and test to ensure that the new routing rules are correct. A more detailed list of the routing changes can be found here: [Trello](https://trello.com/c/TEvX8a4Z/2974-routing-individual-census-eng-wls-m)

PR #2123, opened 2019-06-13
Fixing question variants for the sex question

berroar:
Need to revert a recent change [PR](https://github.com/ONSdigital/eq-survey-runner/pull/2117) to question variant rules for the sex question. 

### How to review 
Test the schema to ensure that the question guidance for the sex question is only being displayed when the individual is aged 16 years or older.

### DIFF
--- a/data-source/jsonnet/common/lib/rules.libsonnet
+++ b/data-source/jsonnet/common/lib/rules.libsonnet
@@ -21,7 +21,7 @@
   },
   over15: {
     id: 'date-of-birth-answer',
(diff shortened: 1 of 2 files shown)


bitdivision (approved):
Under 16 will not see guidance..

bambrp (approved):
Works for me, tested with 16th birthday today, and yesterday, and it behaves as expected.

MebinAbraham (approved):

MebinAbraham (approved):

PR #2126, opened 2019-06-13
Primary Person Design Proposal

iwootten:

Added a description of possible primary person design for runner.

### How to review 

See if the suggested approach makes sense. 
See if anything is missing.


doc/architecture/decisions/0009-primary-person.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Context
+
+We need to be able to identify/uniquely handle responses and the routing path of a primary person within a household.
+
+## Decision
+
+### Extend custom view behaviour for block types
+
+We will extend current block type handling so blocks route to custom view behaviour based on their type. This allows a block type of 'primary-person-selector' to custom handle forward and backward routing paths for instance.
```
jonnyshaw89:
```suggestion
We will extend current block type handling so blocks route to custom view behaviour based on their type. This allows a block type of 'primary-person-selector' to handle custom forward and backward routing paths for instance.
```

bambrp (approved):
Seems eminently sensible to me considering the MVP, basic HH census scenario, and the alternatives.

doc/architecture/decisions/0009-primary-person.md
```diff
@@ -31,21 +27,30 @@ The fact store would be used for storage of items that are not direct answers bu
...
+    ]
 }
 ```
 
-It is not necessary for a fact to be stored against anything other than a list item at the time of writing.
+### Create a PrimaryPersonCollector block type
+
+A new block type "PrimaryPersonCollector" will be defined to support the necessary behaviour. It will extend the existing ListCollector and support adding a primary person through adding context to the list store. The following diagram identifies the flow to be implemented in runner.  
+
+![Primary Flow](0009-primary-flow.png)
```
jonnyshaw89, line 56: i think the id in the url needs to be removed from `/do-you-live-at/<id>/add`?

iwootten: Does it? This is the suggested url for adding answers to an existing list item

jonnyshaw89: oh, is this the list id or the list item id?

iwootten: List item id - I can update the diagram to make things clearer though

jonnyshaw89: is this not inconsistent with the list collector add?

ajmaddaford: The root part of the urls should be the list name, not the list collector block name. This was changed during the implementation of lists - https://github.com/ONSdigital/eq-survey-runner/blob/v3/doc/architecture/decisions/0005-simplify-urls-and-extend-to-support-repeating-effectively.md was updated with the new urls, but the images in https://github.com/ONSdigital/eq-survey-runner/blob/v3/doc/architecture/decisions/0006-make-named-lists-a-first-class-construct.md weren't.

ajmaddaford: I think it would also be clearer if the url for the add primary person was `/<list_name>/<list_item_id>/primary-person` as it's both the add and edit screen.

doc/architecture/decisions/0009-primary-person.md
```diff
@@ -0,0 +1,58 @@
+# 9. Primary Person
+
+## Context
+
+We need to be able to identify/uniquely handle responses and the routing path of the primary person within a household. Currently people within a household would be represented as a list, which generates a number of pages distinct from the main routing path.
```
ajmaddaford: I think we need to make it clear there is a separation between the collection of the primary person and the places where we make use of that fact. We handle responses for the primary person in the same way as any other person, the same for the routing path.

doc/architecture/decisions/0009-primary-person.md
```diff
@@ -0,0 +1,58 @@
...
+
+### Create a PrimaryPersonCollector block type
+
+A new block type "PrimaryPersonCollector" will be defined to support the necessary behaviour. It will extend the existing ListCollector and support adding a primary person through adding context to the list store. The following diagram identifies the flow to be implemented in runner.  
+
+![Primary Flow](0009-primary-flow.png)
+
+Both the PrimaryPersonListCollector and ListCollector operate upon the same list in the store. The PrimaryPersonCollector would create an item in the list store, before defining any answers it refers to. This allows the identification of a primary person in one block and deferring their name entry to another.
+
+On routing backwards from the ListCollector, a custom route handler would be implemented to prevent the primary person reaching add behaviour. This is represented in the reverse flow below.
```
ajmaddaford: This doesn't make sense. The routing path wouldn't include the add screen as it would be defined as part of the `PrimaryPersonListCollector`, so I assume this is referring to the back button behaviour? If this is the case, if the url for the primary person add/edit screen has the list item id in it, we should be safe to display it?

ajmaddaford (changes requested):

doc/architecture/decisions/0009-primary-person.md
iwootten: I thought the point needed to be made here that navigating using previous/back button to the add screen should result in the driving question being shown?

doc/architecture/decisions/0009-primary-person.md
JamesGardiner:
```suggestion
The primary person in a household needs to be identified, so that they they can be identified where additional context is necessary. Currently people within a household would be represented as a list, which generates a number of pages distinct from the main routing path.
```

JamesGardiner (changes requested):

doc/architecture/decisions/0009-primary-person.md
ajmaddaford: I don't think so, we shouldn't break the default back button behaviour.

doc/architecture/decisions/0009-primary-person.md
```diff
@@ -0,0 +1,71 @@
+# 9. Primary Person
+
+## Context
+
+The census requires that the primary person within a household is able to be identified and routed differently than other household members. Currently all people within a household would be represented as a list, with no distinction between them.
```
JamesGardiner:
```suggestion
The census requires that the primary person within a household can be identified and routed differently than other household members. Currently all people within a household would be represented as a list, with no distinction between them.
```

JamesGardiner (changes requested):

doc/architecture/decisions/0009-primary-person.md
```diff
@@ -0,0 +1,69 @@
...
+
+### Create a PrimaryPersonCollector block type
+
+A new block type "PrimaryPersonCollector" will be defined to support the necessary behaviour. It will extend the existing ListCollector and allow addition of a primary person through adding context to the list store. The following diagram identifies the flow to be implemented in runner.  
+
+![Primary Flow](0009-primary-flow.png)
+
+Both the PrimaryPersonListCollector and ListCollector operate upon the same list in the store. The PrimaryPersonCollector would create an item in the list store, before defining any answers it refers to. This allows the identification of a primary person in one block and deferring their name entry to another. 
+
+On routing backwards from the ListCollector, a custom route handler would be implemented. This is represented in the reverse flow below, where the primary person would be unable to reach the primary name block when using the "Previous" link.
```
ajmaddaford: This doesn't require a custom route handler if the add primary person block is defined as part of the "Do you live at?" block as the add primary person block would never be included in the routing path.

doc/architecture/decisions/0009-primary-person.md
```diff
@@ -0,0 +1,69 @@
+# 9. Primary Person
+
+## Context
+
+The census requires that the primary person within a household can be identified and routed differently than other household members. Currently all people within a household would be represented as a list, with no distinction between them.
```
ajmaddaford: would be -> are

ajmaddaford (changes requested):

JamesGardiner (approved):

ajmaddaford (approved):

PR #2128, opened 2019-06-14
Content changes to stocks form types 0009 and 0010.

samiwel:
A change request was made for stocks form types 0009 and 0010.
This PR makes small changes to the content as requested.

### How to review 
Ensure that both schemas can be launched and content is as described on Trello cards.

### DIFF
--- a/data/en/stocks_0010.json
+++ b/data/en/stocks_0010.json
@@ -20,7 +20,7 @@
                             "content": [{
                                 "list": [
                                     "Data should relate to all sites in England, Scotland, Wales and Northern Ireland unless otherwise stated.",
-                                    "You can provide info estimates if actual figures aren&#x2019;t available.",
+                                    "You can provide informed estimates if actual figures aren&#x2019;t available.",
                                     "We will treat your data securely and confidentially."
                                 ]
                             }]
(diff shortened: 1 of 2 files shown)


data/en/stocks_0010.json
```diff
@@ -123,8 +123,8 @@
                                         {
                                             "list": [
                                                 "Products that you own in intermediate stages of completion (even if not held by you)",
-                                                "Long term business contract balances in line with UK Generally Accepted Accountancy Practice or International GAAP",
-                                                "All stocks owned; whether held by you or in transit in the UK or abroad",
+                                                "Long term business contract balances in line with UK Generally Accepted Accountancy Practice (GAAP) or International GAAP",
```
tgandrews, line 126: There is no definition of GAAP but it is used. I'm no writing expert but this looks weird

tgandrews (approved):
Haven't tested it fully but the surveys launch and the changes look fine.

data/en/stocks_0010.json
samiwel: Generally Accepted Accountancy Practice?

tgandrews: Yeah I know it is but this change removes the definition. That's what I think is weird.

I only know what GAAP means by either: guessing from the sentence context or, by knowing it up front.

tgandrews: You can ignore this as this is content. I just think its odd

samiwel: Doesn't this change add the definition?

rmccar (approved):

SamGodwin2 (approved):

PR #2129, opened 2019-06-14
Fix year in date format from YYYY to yyyy

ajmaddaford:
The date format is interpreted in runner using Babel. Using a format of YYYY
meant an ISO year-week calendar was used and so dates close to the start of
the year could get formatted with the wrong year. Using yyyy fixes this.

See http://babel.pocoo.org/en/latest/dates.html#date-fields
and https://github.com/python-babel/babel/issues/419


MebinAbraham (approved):

bambrp (approved):
Fixes for me

PR #2130, opened 2019-06-17
Fix untranslated dates in transform / schema

bambrp:
Date placeholders- i.e. the census date -are currently showing up with English month names ("June") instead of Welsh ("Mehefin") regardless of chosen UI language.

Changed date_formatter locale from current locale to language based. This is fed in from the URL and not currently validated? Should we be changing the locale instead?

### How to review 
* Get to at least the marriage-type question to see a piped date that includes the month by name, and switch languages using the link in the header.

* Try manually changing URL to a different language, such as "de" and observe the month change to Juli. Try "xx".

### DIFF
--- a/app/questionnaire/placeholder_transforms.py
+++ b/app/questionnaire/placeholder_transforms.py
@@ -3,9 +3,8 @@ from datetime import datetime
 from dateutil.relativedelta import relativedelta
 from dateutil.tz import tzutc
 
-from babel.numbers import format_currency
+from babel.numbers import format_currency, format_decimal
 from babel.dates import format_datetime
-from babel import numbers
 
 from app.settings import DEFAULT_LOCALE
 
@@ -17,8 +16,8 @@ class PlaceholderTransforms:
 
     def __init__(self, language):
         self.language = language
+        self.locale = DEFAULT_LOCALE if language == 'en' else language
 
-    locale = DEFAULT_LOCALE
     input_date_format = '%Y-%m-%d'
     input_date_format_month_year_only = '%Y-%m'
 
@@ -75,10 +74,9 @@ class PlaceholderTransforms:
 
         return string_to_format
 
-    @staticmethod
-    def format_number(number):
+    def format_number(self, number):
         if number or number == 0:
-            return numbers.format_decimal(number, locale=PlaceholderTransforms.locale)
+            return format_decimal(number, locale=self.locale)
 
         return ''
 


berroar (approved):
Tested the change - the month displayed is now being translated successfully when I update the URL.

MebinAbraham (commented):
We should change the schema in `data/cy/`?. The Welsh schema still references old dates.

app/questionnaire/placeholder_transforms.py
```diff
@@ -27,7 +27,7 @@ def format_currency(self, number=None, currency='GBP'):
 
     def format_date(self, date_to_format, date_format):
         date_to_format = datetime.strptime(date_to_format, self.input_date_format)
-        return format_datetime(date_to_format, date_format, locale=self.locale)
+        return format_datetime(date_to_format, date_format, locale=self.language)
```
iwootten: Seems strange to use language here, when there is a locale variable in the class already. Shouldn't it be initialised to the correct value when the class is initialised instead?

scripts/build_schemas.sh
```diff
@@ -8,13 +8,19 @@ for region_code in GB-WLS GB-ENG GB-NIR; do
     FORMATTED_REGION_CODE=$(echo "${region_code}" | tr '[:upper:]' '[:lower:]' | tr - _)
     DESTINATION_FILE="data-source/jsonnet/census_individual_${FORMATTED_REGION_CODE}.json"
 
+    if [[ "$region_code" = "GB-WLS" ]]; then
+        CENSUS_DATE="2019-06-20"
```
iwootten: As mentioned on slack, these dates don't need to be different

app/questionnaire/placeholder_transforms.py
iwootten: Additionally, we shouldn't be using locale on some methods and language on others, it should be the same throughout.

iwootten (changes requested):

scripts/build_schemas.sh
bambrp: The card only referenced Wales, but I'm told that's not the case, so I've removed the conditional code and updated the date to the 20th.
What I didn't know is that the Welsh schema wasn't also built automatically. Alas!

bambrp:
> We should change the schema in `data/cy/`?. The Welsh schema still references old dates.

Done!

scripts/build_schemas.sh
bambrp: Built the schema.

data/cy/census_individual_gb_wls.json
```diff
@@ -1815,37 +1860,37 @@
                                         "answers": [
                                             {
                                                 "id": "other-address-answer-building",
-                                                "label": "Enw neu rif y t\u0177",
+                                                "label": "House name or number",
```
iwootten: It does seem like we're dropping translations here that we have..has something changed with the content?

data/cy/census_individual_gb_wls.json
```diff
@@ -1815,37 +1860,37 @@
...
                                                 "mandatory": true,
                                                 "type": "TextField"
                                             },
                                             {
                                                 "id": "other-address-answer-street",
-                                                "label": "Stryd",
+                                                "label": "Street",
```
iwootten: As above..

data/cy/census_individual_gb_wls.json
```diff
@@ -1815,37 +1860,37 @@
...
                                                 "id": "other-address-answer-county",
-                                                "label": "Sir (dewisol)",
+                                                "label": "County (optional)",
                                                 "mandatory": false,
                                                 "type": "TextField"
                                             },
                                             {
                                                 "id": "other-address-answer-postcode",
-                                                "label": "Cod post",
+                                                "label": "Postcode",
```
iwootten: and again...

iwootten (dismissed):

data/cy/census_individual_gb_wls.json
bambrp: ~Not that I can see, and it's in the .po file with the correct matching context. I don't know why it didn't insert it into the schema.~

app/questionnaire/placeholder_transforms.py
bambrp: I think I've made this more consistent with this commit.

data/cy/census_individual_gb_wls.json
bambrp: Ah, got it. Yes, at least in some cases- e.g. a change of question from "One year ago, what was your usual address?" to "Enter details of their address one year ago.", which breaks the connection.
Not really a way around that other than updating the pot file and getting the missing translations done, which we need anyway, and isn't part of this PR (the generated schema was just to ease the testing).

app/questionnaire/placeholder_transforms.py
```diff
@@ -17,8 +17,12 @@ class PlaceholderTransforms:
 
     def __init__(self, language):
         self.language = language
+        # Expand language code into true locale
+        if language == 'en':
+            self.locale = DEFAULT_LOCALE
+        else:
```
MebinAbraham: 🙃 Personal preference I guess.
```python
self.locale = DEFAULT_LOCALE if language == 'en' else language
```

MebinAbraham (approved):

app/questionnaire/placeholder_transforms.py
```diff
@@ -17,8 +17,12 @@ class PlaceholderTransforms:
 
     def __init__(self, language):
         self.language = language
+        # Expand language code into true locale
```
JamesGardiner: This comment isn't necessary.

app/questionnaire/placeholder_transforms.py
```diff
@@ -75,10 +79,9 @@ def format_possessive(self, string_to_format):
 
         return string_to_format
 
-    @staticmethod
-    def format_number(number):
+    def format_number(self, number):
         if number or number == 0:
-            return numbers.format_decimal(number, locale=PlaceholderTransforms.locale)
+            return numbers.format_decimal(number, locale=self.locale)
```
JamesGardiner:
```suggestion
            return format_decimal(number, locale=self.locale)
```

This would be more consistent with `format_currency` and `format_datetime`, which are imported explicitly from the Babel numbers package.

JamesGardiner (changes requested):

JamesGardiner (approved):

PR #2135, opened 2019-06-21
Adding datalayer and extra variables for GTM

rmccar:
This PR is to parameterise the extra variables needed for GTM to work and to initialise the datalayer to allow it to be used.

The variables removed from the code (gtm_preview and gtm_cookies_win) are now set in the environment variable which previously just contained gtm_auth but will now contain all 3

### How to review 
Check to see that GTM is working

### DIFF
--- a/app/templates/layouts/_base.html
+++ b/app/templates/layouts/_base.html
@@ -48,10 +48,13 @@
     {% endblock head -%}
     <!-- Google Tag Manager -->
     {% if analytics_gtm_id and analytics_gtm_env_id %}
+      <script>
+        dataLayer = [];
+      </script>
       <script nonce="{{ csp_nonce() }}">(function(w,d,s,l,i,e){w['GoogleTagManagerObject']=l;w[l]=w[l]||[];w[l].push({'gtm.start':
       new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
       j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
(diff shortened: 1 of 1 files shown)


app/templates/layouts/_base.html
```diff
@@ -48,10 +48,13 @@
...
     <!-- Google Tag Manager -->
     {% if analytics_gtm_id and analytics_gtm_env_id %}
+      <script>
+        dataLayer = [];
+      </script>
       <script nonce="{{ csp_nonce() }}">(function(w,d,s,l,i,e){w['GoogleTagManagerObject']=l;w[l]=w[l]||[];w[l].push({'gtm.start':
       new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
       j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
-      "https://www.googletagmanager.com/gtm.js?id=" + i + dl + "&gtm_auth=" + e + "&gtm_preview=env-8&gtm_cookies_win=x";f.parentNode.insertBefore(j,f);
+      "https://www.googletagmanager.com/gtm.js?id=" + i + dl + "&gtm_auth=" + e;f.parentNode.insertBefore(j,f);
```
samiwel, line 57: Any reason for not passing the full URL in a single variable e.g. `analytics_gtm_url`?
Data layer is always a literal string here and the GTM Id and ENV are just per-environment configuration, so feels like you could construct this string without needing all this concatenation.

samiwel (changes requested):

app/templates/layouts/_base.html
rmccar: I see what you mean, and I suppose datalayer could be added to the variable too but the GTM id is separate because that variable is per project and not per environment

samiwel (approved):
LGTM

tgandrews (approved):

PR #2136, opened 2019-06-24
Add lists to when rules

ajmaddaford:
Adds lists to when rules to enable routing logic based on the number of items in a list.

### How to review 
The `test_skip_condition_list.json` can be used to test the new functionality.

### DIFF
--- a/app/questionnaire/rules.py
+++ b/app/questionnaire/rules.py
@@ -132,7 +132,9 @@ def convert_to_datetime(value):
     return datetime.strptime(value, date_format) if value else None
 
 
-def evaluate_goto(goto_rule, schema, metadata, answer_store, routing_path=None):
+def evaluate_goto(
+    goto_rule, schema, metadata, answer_store, list_store, routing_path=None
+):
     """
     Determine whether a goto rule will be satisfied based on a given answer
     :param goto_rule: goto rule to evaluate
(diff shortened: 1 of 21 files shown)


scripts/test_schemas.sh
```diff
@@ -6,8 +6,8 @@ if [ "$1" == "--local" ] || [ "$2" == "--local" ]; then
 fi
 
 if [ "$run_docker" == true ]; then
-    docker pull onsdigital/eq-schema-validator:v3
-    validator="$(docker run -d -p 5001:5000 onsdigital/eq-schema-validator:v3)"
+    docker pull onsdigital/eq-schema-validator:add-lists-to-when-rules
```
JamesGardiner: Reminder to remove before merge

JamesGardiner (approved):

app/questionnaire/rules.py
```diff
@@ -257,5 +275,10 @@ def get_metadata_value(metadata, key):
     return metadata.get(key)
 
 
+def get_list_count(list_store, list_name):
+    list_values = list_store[list_name]
+    return len(list_values) if list_values else 0
```
MebinAbraham: Since you always get an empty list back?
```suggestion
    return len(list_values)
```

MebinAbraham (changes requested):
Minor comment otherwise looks good. 👍

app/questionnaire/rules.py
ajmaddaford: I've also removed `list_values` so this is now `return len(list_store[list_name])`.

MebinAbraham (approved):

PR #2138, opened 2019-06-25
Update survey runner from Python 3.4 -> 3.7

samiwel:
There are some security alerts in Github that dependabot could resolve automatically for us but unfortunately python 3.4 is not supported. Survey runner was historically pinned to 3.4 because of runtime availability in Elastic Beanstalk. Now that Runner is built and run in containers,  the availability of a 3.7 runtime is no longer an issue other than any specific compatibility issues between project dependencies, there's no real reason why we shouldn't be able to upgrade to a newer version of Python.

The Census version of Runner has has already upgraded to Python 3.7  on the v3 branch. This PR is based on the same changes but applies to the business and social EQ runner.

### How to review 
All tests and checks should pass.
Try running this branch locally as a sanity check to make sure that the container image runs okay.


MebinAbraham:
Should be handy: https://github.com/ONSdigital/eq-survey-runner/pull/2002

tgandrews (approved):

heofs (approved):
Cool. No more relying on 3.4!

PR #2139, opened 2019-06-26
Use schema_name and allow census claims

bitdivision:
Context available in [trello](https://trello.com/c/ZhMmtMYB/2954-use-schemaname-instead-of-eqid-formtype)

Removes the use of `eq_id` and `form_type`. These are now equivalent to `schema_name`
Runner uses `schema_name` throughout. The census claims:
`survey` / `case_type` / `region_code` can now be used instead of `schema_name`

Additionally, the schema endpoint has been updated to only send filenames without
the file extension. This now matches the format of `schema_name`. The file extension seems like an internal implementation detail.

Basic validation for the census parameters has been added using marshmallow. Marshmallow will also insert a `schema_name` parameter into the metadata as soon as it has been validated.

`instrument_id` in the downstream payload has been changed to `schema_name`.

### How to review 
- This should be tested in combination with the changes to eq-questionnaire-launcher.

- Ensure that runner works when questionnaire-launcher passes just a schema_name (on a test schema)
- Ensure runner works when questionnaire-launcher passes the survey / case_type / region_code

### Checklist


app/views/schema.py
```diff
@@ -20,7 +20,7 @@ def get_schema_json(eq_id, form_type):
 def list_schemas():
     schema_path = 'data/en'
     onlyfiles = [
```
iwootten: This variable name seems a bit weird now.

app/storage/metadata_parser.py
```diff
@@ -65,6 +83,51 @@ class RunnerMetadataSchema(Schema, StripWhitespaceMixin):
...
+                raise ValidationError(
+                    "Either 'schema_name' or 'survey' and 'case_type' and 'region_code' must be defined"
+                )
+
+    @post_load
+    def convert_schema_name(self, data, **kwargs):
+        """ Temporary function for census to transform parameters into a census schema
+        This can be removed after census.
+        """
+        if not data.get('schema_name'):
```
iwootten: Could flip the cases here to remove the not.

app/utilities/schema.py
```diff
@@ -31,35 +32,68 @@ def load_schema_from_session_data(session_data):
...
+
+def transform_survey(survey_input):
+    return survey_input.lower()
+
+
+def get_schema_name_from_census_params(survey, case_type, region_code):
+    try:
+        case_type_transformed = transform_case_type(case_type)
+    except KeyError:
+        # TODO: Raise a reasonable error here
```
iwootten: Missing an error?

scripts/generate_schemas.py
```diff
@@ -1,102 +0,0 @@
-#!/usr/bin/env python
```
iwootten, line 1: I'm guessing this is no longer used?

iwootten (changes requested):

scripts/generate_schemas.py
bitdivision: I don't know why it's still here. So I deleted it. It was used for combining yaml files AFAICS, which we're no longer doing in favour of jsonnet.

bambrp (approved):
Tested with matching branch of questionnaire-launcher, ran through with eng,wls,nisra schemata successfully. All functional tests passing locally.

iwootten (approved):

PR #2140, opened 2019-06-26
Add GTM blacklist

rmccar:
This adds the blacklist to the DataLayer object which will lockdown the injecting of any JS into the app but will still allow the use of Google Tag Manager

### How to review 
Make sure changes make sense and GTM still works locally by inspecting the page, making sure the tags are rendered and checking preview window displays

### DIFF
--- a/app/templates/layouts/_base.html
+++ b/app/templates/layouts/_base.html
@@ -49,7 +49,7 @@
     <!-- Google Tag Manager -->
     {% if analytics_gtm_id and analytics_gtm_env_id %}
       <script>
-        dataLayer = [];
+        dataLayer = [{'gtm.blacklist': ['customScripts' , 'html']}];
       </script>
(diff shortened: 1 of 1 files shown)


app/templates/layouts/_base.html
```diff
@@ -49,7 +49,7 @@
     <!-- Google Tag Manager -->
     {% if analytics_gtm_id and analytics_gtm_env_id %}
       <script>
-        dataLayer = [];
+        dataLayer = [{'gtm.blacklist': ['customScripts' , 'html']}];
```
samiwel, line 52: What is the reason for blocking these specific tags? Is there any significance to the customScripts and html tags within tag manager?

samiwel (commented):
Looks fine, but I don't know that I fully understand the reason why these tags are blacklisted?

samiwel (approved):

AndyPitt180 (approved):
As discussed with @samiwel

PR #2141, opened 2019-06-28
Default survey claim to census

bitdivision:
The survey claim is meant to default to 'CENSUS' if not provided. The test I added for this in https://github.com/ONSdigital/eq-survey-runner/pull/2139 was broken, and survey wasn't actually being defaulted.

This change ensures it is actually defaulted and fixes the test.


pricem14pc (approved):
Looks ok an I've tested it locally and its launching ok. However, I guess we're not able to actually omit the claim itself without hacking launcher.

iwootten (approved):

PR #2142, opened 2019-06-28
Feature: Hub & Spoke

MebinAbraham:
Implements the basic functionality of the hub and spoke feature. This does not cater for repeating sections.

For now, the hub is defined as a top-level schema object but this is subject to change once https://trello.com/c/BJPC98fW/3086-move-final-summary-block-out-of-sections is implemented. We should potentially have a single selection of the submission type for a schema ie Summary/Confirmation/Hub.

`CompletedStore` is now `ProgressStore`, this new store will keep track of the 'progress' of sections in the survey and its completed location. Prior to this change, the Hub was calculating the routing path for each section so that it could work out the sections status and the URL the section is supposed to link to. Storing the state of each section now means that this is no longer necessary and should be much more efficient. 

Since each section also needs to link to its relevant location ie summary, first incomplete location etc, we have created an endpoint `/hub/<section_id>/` that will be used to redirect to the correct location for a given section. This eliminates having to calculate the path on the hub for each section.

Progress Store's structure is now hierarchical, previously, the completed location and completed sections were both flat but a location always belongs to a section, therefore to make certain aspects such as calculating whether a section is complete or not, the location has been nested by the section it belongs to.

The new structure is:
```python
{
    'section-1': {
        'status': 'COMPLETED',
        'locations': ['location-1', 'location-2'],
    }
}
```
Only complete and in-progress sections are stored. If the one and only location for a section is removed, then the section is also removed from the progress store, ie not storing sections with empty locations.

For repeating section, the idea was to potentially store it same/similarly to the answer store, where the key is a tuple of `section_id` and the `list_item_id`. The hub's section endpoint could become `/hub/<section_id>/<list_items_id/`

[Trello](https://trello.com/c/n0bp9POq/3064-hub-spoke-introduce-basic-hub-functionality-l)

### How to review 
Ensure:
1. The functionality is as expected using the `test_hub_and_spoke` schema.
2. The different scenarios highlighted by the Google Slides are as expected. *(Link in Trello)*
2. Non hub schemas behave the same.


app/questionnaire/hub.py
```diff
@@ -0,0 +1,136 @@
...
+            'title': _('Choose another section to complete'),
+            'description': _(
+                'You must complete all sections in order to submit this survey'
+            ),
+            'button': _('Continue'),
+        },
+    }
+
+    COMPLETION_STATES = {
+        'COMPLETED': 'COMPLETED',
```
bitdivision: It's probably worth looking at whether this is still needed. It might be better to change this to an `enum` (and use `auto()`).

app/questionnaire/hub.py
```diff
@@ -0,0 +1,136 @@
...
+        if state == self.COMPLETION_STATES['COMPLETED']:
+            context['rowItems'][0]['icon'] = 'check-green'
+
+        return context
+
+    def get_rows(self):
+        rows = []
+
+        for section in self.sections:
+            routing_path = path_finder.routing_path(section)
```
ajmaddaford: We don't want to be generating a routing path for every section on every visit to the hub. There is no need if the section is complete (the section id will be in completed sections). For in-progress sections we should consider storing that state when a a section is started.

app/questionnaire/hub.py
```diff
@@ -0,0 +1,136 @@
+from flask_babel import lazy_gettext as _
```
ajmaddaford: It kind of feels like this should be two classes. One that calculates and stores the overall state of the survey, and another that is a view context that takes that and does the view specific mapping. It would also mean the survey progress would be more unit testable without getting into view specifics.

app/questionnaire/router.py
```diff
@@ -54,7 +56,7 @@ def get_next_location(self, location, routing_path):
 
         return routing_path[location_index + 1]
 
-    def get_previous_location(self, location, routing_path):
+    def get_previous_location_url(self, location, routing_path):
```
ajmaddaford, line 67: This is now inconsistent with `get_next_location`.

ajmaddaford (changes requested):

app/questionnaire/hub.py
```diff
@@ -0,0 +1,138 @@
...
+        return {
+            'title': self.get_content('title'),
+            'description': self.get_content('description'),
+            'rows': self.get_rows(),
+            'submit_button': self.get_content('button'),
+        }
+
+    @staticmethod
+    def get_section_title(section):
+        # Need to handle repeating sections
```
JamesGardiner: Is this a TODO for this PR or future PR?

app/questionnaire/hub.py
JamesGardiner: Or when a section is revisited

JamesGardiner (changes requested):

MebinAbraham:
Will reopen once new changes are complete.

MebinAbraham:
@bitdivision @ajmaddaford @JamesGardiner 
PR description updated with recent changes.

app/questionnaire/hub.py
MebinAbraham: Now uses a class inheriting from `IntEnum`

app/questionnaire/hub.py
MebinAbraham: @ajmaddaford As per the latest changes, this should be addressed.

@JamesGardiner when a section is revised, the path would need to be evaluated as we can't guarantee something has not changed since the last request, hence we calculate the path on every request?

app/questionnaire/hub.py
MebinAbraham: `ProgressStore` now stores the state of the survey/section and the `Hub` is then initialized using `ProgressStore` to build the view context.

app/questionnaire/router.py
MebinAbraham: Both of them now return the url. `get_next_location ` -> `get_next_location_url`.

app/questionnaire/hub.py
MebinAbraham: Has been removed.

app/data_model/progress_store.py
```diff
@@ -0,0 +1,114 @@
...
+
+    def add_completed_location(self, section_id: str, location: Location) -> None:
+        locations = self._sections.get(section_id, {}).get('locations', [])
+
+        if location not in locations:
+            locations.append(location)
+
+            section_progress = {'locations': locations}
+
+            self._sections[section_id] = section_progress
```
ajmaddaford: Won't this clear out the section status?

app/data_model/progress_store.py
```diff
@@ -0,0 +1,114 @@
...
+            if not self._sections[section_id]['locations']:
+                del self._sections[section_id]
+
+            self._is_dirty = True
+
+    def serialise(self) -> Mapping:
+        serialised = {}
+
+        for section_id, section_status in self._sections.items():
+            locations = self.get_section_locations(section_id)
```
ajmaddaford: This could just be `section_status['locations']`

app/data_model/progress_store.py
```diff
@@ -0,0 +1,114 @@
...
+    """
+
+    def __init__(self, in_progress_sections: Mapping = None) -> None:
+        """
+        Instantiate a ProgressStore object that tracks the status of sections and its completed locations
+        Args:
+            in_progress_sections: A hierarchical dict containing the section status and completed locations
+        """
+        self._is_dirty = False  # type: bool
+        self._sections = {}  # type: MutableMapping
```
ajmaddaford: I think it might make some of the methods clearer if this was called `_progress`

app/data_model/questionnaire_store.py
```diff
@@ -43,7 +43,7 @@ def set_metadata(self, to_set):
 
     def _deserialise(self, data):
         json_data = json.loads(data, use_decimal=True)
-        self.completed_store = CompletedStore(json_data.get('COMPLETED'))
+        self.progress_store = ProgressStore(json_data.get('COMPLETED'))
```
ajmaddaford: Should `COMPLETED` now be `PROGRESS`?

app/questionnaire/path_finder.py
```diff
@@ -24,28 +24,32 @@ def __init__(
...
+        if path:
+            first_location = path[0]
+            section_id = self.schema.get_section_id_for_block_id(
+                first_location.block_id
+            )
+
+            for location in path:
+                block = self.schema.get_block(location.block_id)
+                block_type = block.get('type')
+                if location not in self.progress_store.get_section_locations(
```
ajmaddaford: I think this would be clearer if it was named `get_completed_locations`

app/views/questionnaire.py
```diff
@@ -300,30 +360,55 @@ def post_block_handler(
...
-    next_location = router.get_next_location(current_location, routing_path)
+    last_block_id_in_path = routing_path[-1].block_id
 
-    return redirect(next_location.url())
+    if (
+        schema.is_hub_enabled()
+        and current_location.block_id == last_block_id_in_path
+        and is_path_complete
+    ):
+        return redirect_to_hub()
```
ajmaddaford: This if logic looks like it would fit better in router as part of `get_next_location`

app/data_model/progress_store.py
```diff
@@ -0,0 +1,114 @@
+from enum import IntEnum
+from typing import List, Mapping, MutableMapping
+
+from app.questionnaire.location import Location
+
+
+class CompletionStatus(IntEnum):
+    COMPLETED = 1
+    IN_PROGRESS = 2
+    NOT_STARTED = 3
```
ajmaddaford: Using numbers means the debug output is less readable 😕

app/questionnaire/hub.py
```diff
@@ -0,0 +1,118 @@
...
+    def _get_content(self, content: str, survey_status: CompletionStatus) -> str:
+        return self.HUB_CONTENT_STATES[survey_status][content]
+
+    def get_contents(self) -> Mapping:
+        return {
+            'title': self._get_content('title', self._survey_status),
+            'description': self._get_content('description', self._survey_status),
+            'rows': self._get_rows(),
+            'submit_button': self._get_content('button', self._survey_status),
+        }
```
ajmaddaford: `_get_content` could be removed and this simplified to:
```python
content = self.HUB_CONTENT_STATES[self._survey_status]
content['rows'] = self._get_rows()
return content
```

app/questionnaire/hub.py
```diff
@@ -0,0 +1,118 @@
+from typing import List, Mapping, Union
```
ajmaddaford: I think we should consider moving this to the `templating` directory and renaming it `hub_context.py` to be consistent with the other view contexts we generate.

app/questionnaire/hub.py
```diff
@@ -0,0 +1,118 @@
...
+        survey_status: CompletionStatus,
+    ) -> None:
+        self._progress_store = progress_store
+        self._sections = sections
+        self._survey_status = survey_status
+
+    def _get_content(self, content: str, survey_status: CompletionStatus) -> str:
+        return self.HUB_CONTENT_STATES[survey_status][content]
+
+    def get_contents(self) -> Mapping:
```
ajmaddaford: This should be `get_context`. I have no idea why questionnaire.py passes this to the templates as `content`.

app/questionnaire/hub.py
```diff
@@ -0,0 +1,118 @@
...
+                    'valueList': [
+                        {'text': self.SECTION_CONTENT_STATES[section_status]['text']}
+                    ],
+                    'actions': [
+                        {
+                            'text': self.SECTION_CONTENT_STATES[section_status]['link'][
+                                'text'
+                            ],
+                            'ariaLabel': self.SECTION_CONTENT_STATES[section_status][
+                                'link'
```
ajmaddaford: Assigning `self.SECTION_CONTENT_STATES[section_status]` to a variable would make this reformatting less horrible.

app/questionnaire/router.py
```diff
@@ -89,6 +96,28 @@ def get_first_incomplete_location_in_survey(self):
...
+        if section_id in self._progress_store.sections:
+            for location in routing_path[::-1]:
+                if location in self._progress_store.sections[section_id]['locations']:
+                    return location
+
+    def get_survey_status(self):
+        if self.is_survey_complete():
+            return CompletionStatus.COMPLETED
+
+        return CompletionStatus.IN_PROGRESS
```
ajmaddaford: I don't see the value in this being a status rather than a boolean. A survey is either complete or it's not.

ajmaddaford (changes requested):

app/data_model/progress_store.py
```diff
@@ -0,0 +1,112 @@
...
+                del self._progress[section_id]
+
+            self._is_dirty = True
+
+    def serialise(self) -> Mapping:
+        serialised = {}
+
+        for section_id, section_status in self._progress.items():
+            serialised[section_id] = {
+                'status': section_status['status'],
```
ajmaddaford: Minor, but as we have a `status` property, it would be clearer if `section_status` was renamed `section_progress`.

app/questionnaire/router.py
```diff
@@ -89,6 +102,21 @@ def get_first_incomplete_location_in_survey(self):
...
+    def get_first_incomplete_location_for_section(self, section_id, routing_path):
+        if section_id in self._progress_store:
+            for location in routing_path:
+                if location not in self._progress_store.get_completed_locations(
+                    section_id
+                ):
+                    return location
+        return routing_path[0]
+
+    def get_last_complete_location(self, section_id, routing_path):
```
ajmaddaford: This should be named `get_last_complete_location_for_section` to be consistent with `get_first_incomplete_location_for_section`

app/views/questionnaire.py
```diff
@@ -300,30 +360,46 @@ def post_block_handler(
...
 
     questionnaire_store_updater.save()
 
-    next_location = router.get_next_location(current_location, routing_path)
+    next_location_url = router.get_next_location_url(current_location, routing_path)
+
+    return redirect(next_location_url)
+
+
+def _perform_form_action(
```
ajmaddaford: `form_action` is too generic here as it's only dealing with signing out, `_handle_sign_out` possibly?

ajmaddaford (changes requested):

ajmaddaford (approved):
Looks good! 👍

JamesGardiner (approved):

PR #2144, opened 2019-07-01
Update to design system 12.1.1 and fix breaking changes relating to question title and legends

bameyrick:
We've had to change how we handle `h1`s (The question title) and `legend`s as JAWS cannot find a `h1` when it's nested within a `legend`.


app/jinja_filters.py
```diff
@@ -190,19 +190,38 @@ def setAttributes(dictionary, attributes):
...
+
+
+@blueprint.app_template_filter()
+def answer_requires_legend(question):
+    if answers_require_legend(question):
+        return False
+
+    question_type = question['type']
+
+    if question_type == 'MutuallyExclusive':
```
iwootten: If it can, it would make sense to make this the first check in the filter, so it can return early without calling answers_require_legend.

app/jinja_filters.py
```diff
@@ -190,19 +190,38 @@ def setAttributes(dictionary, attributes):
...
+    return False
+
+
+@blueprint.app_context_processor
+def answers_require_legend_processor():
+    return dict(answers_require_legend=answers_require_legend)
+
+
+@blueprint.app_template_filter()
+def answer_requires_legend(question):
```
iwootten: This seems to be named for an answer argument, but actually operates on a question.

app/jinja_filters.py
```diff
@@ -190,19 +190,38 @@ def setAttributes(dictionary, attributes):
 
 
 @blueprint.app_template_filter()
-def question_as_legend(question):
+def answers_require_legend(question):
```
iwootten: This seems to be named for an answer argument, but actually operates on a question.

iwootten (changes requested):
Seems like these filters might be better split between methods that return whether a question/type requires a legend and question answers require a legend. May also make sense to have some unit tests around them so it's clear what they're expected to do.

app/jinja_filters.py
```diff
@@ -190,19 +190,38 @@ def setAttributes(dictionary, attributes):
...
 
     if question_type == 'MutuallyExclusive':
-        return True
+        return False
 
     answers = question['answers']
     more_than_one_question = len(answers) > 1
 
-    if more_than_one_question:
+    if more_than_one_question and question_type != 'DateRange':
```
JamesGardiner:
```Python
return len(answers) > 1 and question_type != 'DateRange'
```

app/jinja_filters.py
```diff
@@ -190,19 +190,38 @@ def setAttributes(dictionary, attributes):
...
         return True
 
-    answer = answers[0]
+    return False
+
+
+@blueprint.app_context_processor
+def answers_require_legend_processor():
+    return dict(answers_require_legend=answers_require_legend)
```
JamesGardiner: Using `{}` is significantly faster than dict for non empty dictionaries.

```Python
return {'answers_require_legend': answers_require_legend}
```

app/jinja_filters.py
```diff
@@ -190,19 +190,38 @@ def setAttributes(dictionary, attributes):
...
+    question_type = question['type']
+
+    if question_type == 'MutuallyExclusive':
+        return True
+
+    answer = question['answers'][0]
+
     if 'type' in answer and any(
         answer['type'] in answer_type
         for answer_type in ['Checkbox', 'Radio', 'Date', 'Duration']
```
JamesGardiner: Not part of your PR but this is ugly Python. Would be much clearer as:

```Python
try:
    return answer['type'] in ['Checkbox', 'Radio', 'Date', 'Duration']
except KeyError:
    return False
```

app/jinja_filters.py
```diff
@@ -213,8 +232,8 @@ def question_as_legend(question):
 
 
 @blueprint.app_context_processor
-def question_as_legend_processor():
-    return dict(question_as_legend=question_as_legend)
+def answer_requires_legend_processor():
+    return dict(answer_requires_legend=answer_requires_legend)
```
JamesGardiner:
```suggestion
    return {'answer_requires_legend': answer_requires_legend}
```

JamesGardiner (changes requested):

app/jinja_filters.py
JamesGardiner: Please change it, for me :smile:

iwootten:
```python     
return 'type' in answer and answer['type'] in ['Checkbox', 'Radio', 'Date', 'Duration']
```

app/jinja_filters.py
```diff
@@ -190,31 +190,18 @@ def setAttributes(dictionary, attributes):
...
         return True
 
     return False
 
 
 @blueprint.app_context_processor
-def question_as_legend_processor():
-    return dict(question_as_legend=question_as_legend)
+def should_wrap_with_fieldset_processor():
+    return dict(should_wrap_with_fieldset=should_wrap_with_fieldset)
```
JamesGardiner:
```suggestion
    return {'should_wrap_with_fieldset'=should_wrap_with_fieldset}
```

JamesGardiner (dismissed):

app/jinja_filters.py
```diff
@@ -190,31 +190,18 @@ def setAttributes(dictionary, attributes):
...
 
-    if more_than_one_question:
-        return True
-
-    answer = answers[0]
-    if 'type' in answer and any(
-        answer['type'] in answer_type
-        for answer_type in ['Checkbox', 'Radio', 'Date', 'Duration']
-    ):
+    if len(answers) > 1 and not any(answer['type'] == 'Date' for answer in answers):
```
iwootten, line 196: Do we still need the exception as per @JamesGardiner's earlier suggestion for when type is not present?

ajmaddaford: I'd say not, `type` is mandatory and validated in schema validator.

iwootten (approved):

bitdivision:
How come all the JSON formatting has changed here? I can't see any changes to the formatter config?

In future, it would be nice if we could keep formatting changes separate from feature changes. I'm guilty of this too though.

bameyrick:
I'm not sure, this reformat should have happened in my previous PR to update the node dependencies because thats when I changed the formatter to use prettier. Maybe I forgot to run it in that PR and then I've run it since on this branch.

But yeah I agree that we should avoid doing this.

bitdivision:
@bameyrick That does suggest that the linter is not failing anymore? 

I just added a tab to one of the json files and the linter seems to still pass..

bameyrick:
@bitdivision Turns out I had broken the `.editorconfig` file which prettier looks at and thats why it reformatted everything when I merged in the latest version of v3

templates/partials/answers/duration.html
```diff
@@ -37,7 +33,19 @@
...
+    "legend": question_title,
+    "legendClasses": "u-vh"
+  }) -%}
+
+  {%- call onsMutuallyExclusive(mutuallyExclusive) -%}
+    {{- fields | safe -}}
+  {%- endcall -%}
+{%- else -%}
+  {{- fields | safe -}}
+{%- endif -%}
```
ajmaddaford, line 51: Possibly not as part of this pr, but can we make duration an input type like `onsDateInput`? This would mean the mutually exclusive could be dealt with in the same way as other inputs.

bameyrick: Yeah probably out of scope for this PR as will require a new design system release and we're already 1 major version ahead of the version we're updating to in this PR, but it's definitely something we can do!

JamesGardiner (approved):

bameyrick:
🎉

PR #2153, opened 2019-07-03
Update runner node dependencies

bameyrick:
- Replace gulp eslint with eslint-cli, updated eslint configs
- Removed unused cypress config
- Removed gulp json formatter and replaced with prettier
- Removed browsersync and replaced with livereload
- Updated `make dev-compose-up` to pull latest version of launcher
- Updated `yarn start` to run `make dev-compose-up`, `make run`, and `livereload`
- Added .nvmrc
- Removed unused dependencies
- Removed unused variables and code that linter complained about


MebinAbraham:
[Update runner node dependencies](https://trello.com/c/9iB25pE4/3060-update-runner-node-dependencies)

JamesGardiner (approved):

app/setup.py
```diff
@@ -59,7 +59,7 @@
         'https://fonts.googleapis.com',
         "'unsafe-inline'",
     ],
-    'connect-src': ["'self'", 'https://cdn.ons.gov.uk'],
+    'connect-src': ["'self'", 'ws://localhost:35729', 'https://cdn.ons.gov.uk'],
```
iwootten: Is this for live reload? We probably should only be enabling it through a configuration as per setup_secure_headers does with the EQ_ENABLE_LIVE_RELOAD flag?

iwootten (dismissed):

app/setup.py
bameyrick: Yeah it's for live reload... Could you give me a hand with that?

iwootten: Think you should just be able to add it here, given the flag already exists: https://github.com/ONSdigital/eq-survey-runner/blob/0552d46048dcd77deff59b4f833a03d1cf39821f/app/setup.py#L236

bameyrick: Cheers, updated this now

iwootten (approved):

iwootten:
Just needs a squash after https://github.com/ONSdigital/eq-survey-runner/pull/2155 goes in.

PR #2154, opened 2019-07-03
Remove backported dependencies.

samiwel:
Noticed some warnings on the console relating to dependencies that required versions of python.
Some of the dependencies were backported versions of the libraries required for earlier versions of python.
Now that we're on 3.7 we can hopefully remove some of these.

### How to review 
Changes look sensible.
All checks and tests should pass.
Observe that no warnings reported when performing `pipenv install --dev`

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -596,13 +596,6 @@
             ],
             "version": "==0.95"
         },
-        "backports.ssl-match-hostname": {
-            "hashes": [
-                "sha256:bb82e60f9fbf4c080eabd957c39f0641f0fc247d9a16e31e26d594d8f42b9fd2"
-            ],
-            "markers": "python_version < '3.5'",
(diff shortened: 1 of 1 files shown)


samiwel:
@bitdivision any idea what the `typed-ast` and `typing` dependencies were being used for? Removed from Pipfile.lock as they were generating warnings following #2138. It doesn't seem to have broken anything (all tests passing) 🤷‍♂  but bit nervous about merging without understanding what it was being used for.

bitdivision:
@samiwel I don't think this will cause any problems.

I would imagine that `typing` was to allow type hinting while we were using py3.4 but obviously no longer needed. Not sure exactly where `typed-ast` came from, but I don't think removing it will cause any issues. 

:+1:

bitdivision (approved):

tgandrews (approved):

PR #2155, opened 2019-07-03
Add relationships feature

ajmaddaford:
Adds relationships feature.

### How to review 
- Use the `test_relationships` schema

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -92,6 +92,7 @@ requirements.txt
 
 # Ignore built schemas
 data/en/
+data/cy/census_household_gb_wls.json
 
 # Ignore any generated pages for the functional tests
 tests/functional/generated_pages
--- a/app/data_model/relationship_store.py
+++ b/app/data_model/relationship_store.py
@@ -1,0 +1,94 @@
+from dataclasses import dataclass, asdict
+from typing import List, Mapping, Optional
+
+
+@dataclass
+class Relationship:
+    """
+    Represents a relationship between two items.
+    """
+
+    from_list_item_id: str
+    to_list_item_id: str
+    relationship: str
+
+    def for_json(self) -> Mapping:
+        return asdict(self)
+
+
+class RelationshipStore:
+    """
+    Stores and updates relationships.
+    """
+
+    def __init__(self, relationships: Optional[List[Mapping]] = None) -> None:
+        self._is_dirty = False
+        self._relationships = self._build_map(relationships or [])
+
+    def __iter__(self):
+        return iter(self._relationships.values())
+
+    def __contains__(self, relationship):
+        return (
+            relationship.from_list_item_id,
+            relationship.to_list_item_id,
+        ) in self._relationships
+
+    def __len__(self):
+        return len(self._relationships)
+
+    @property
+    def is_dirty(self):
+        return self._is_dirty
+
+    def clear(self):
+        self._relationships.clear()
+        self._is_dirty = True
+
+    def serialise(self):
+        return [
+            relationship.for_json() for relationship in self._relationships.values()
+        ]
+
+    def get_relationship(self, from_list_item_id, to_list_item_id):
+        key = (from_list_item_id, to_list_item_id)
+        return self._relationships.get(key)
+
+    def add_or_update(self, relationship: Relationship):
+        key = (relationship.from_list_item_id, relationship.to_list_item_id)
+
+        existing_relationship = self._relationships.get(key)
+
+        if existing_relationship != relationship:
+            self._is_dirty = True
+            self._relationships[key] = relationship
+
+    def remove_all_relationships_for_list_item_id(self, list_item_id: str):
+        """Remove all relationships associated with a particular list_item_id
+        This method iterates through the entire list of relationships.
+        """
+
+        keys_to_delete = []
+
+        for relationship in self:
+            if list_item_id in (
+                relationship.to_list_item_id,
+                relationship.from_list_item_id,
+            ):
+                keys_to_delete.append(
+                    (relationship.from_list_item_id, relationship.to_list_item_id)
+                )
+
+        for key in keys_to_delete:
+            del self._relationships[key]
+            self._is_dirty = True
+
+    @staticmethod
+    def _build_map(relationships):
+        return {
+            (
+                relationship['from_list_item_id'],
+                relationship['to_list_item_id'],
+            ): Relationship(**relationship)
+            for relationship in relationships
+        }
--- a/app/forms/fields.py
+++ b/app/forms/fields.py
@@ -54,6 +54,7 @@ def get_field(
         field = {
             'Checkbox': get_select_multiple_field,
             'Radio': get_select_field,
+            'Relationship': get_select_field,
             'TextArea': get_text_area_field,
             'TextField': get_string_field,
             'Dropdown': get_dropdown_field,
--- a/app/helpers/form_helper.py
+++ b/app/helpers/form_helper.py
@@ -3,7 +3,9 @@ from collections import OrderedDict
 from structlog import get_logger
 from werkzeug.datastructures import MultiDict
 
+from app.data_model.relationship_store import RelationshipStore
 from app.forms.questionnaire_form import generate_form
+from app.questionnaire.relationship_location import RelationshipLocation
 
 logger = get_logger()
 
@@ -93,14 +95,23 @@ def get_mapped_answers(schema, answer_store, location):
     """
     Maps the answers in an answer store to a dictionary of key, value answers.
     """
+    result = {}
 
-    block_id = location.block_id
-
-    answer_ids = schema.get_answer_ids_for_block(block_id)
-
-    answers = answer_store.get_answers_by_answer_id(
-        answer_ids=answer_ids, list_item_id=location.list_item_id
-    )
-    result = {answer.answer_id: answer.value for answer in answers if answer}
+    if isinstance(location, RelationshipLocation):
+        answer_id = schema.get_relationship_answer_id_for_block(location.block_id)
+        answer = answer_store.get_answer(answer_id)
+        if answer:
+            relationship_store = RelationshipStore(answer.value)
+            relationship = relationship_store.get_relationship(
+                location.from_list_item_id, location.to_list_item_id
+            )
+            if relationship:
+                result = {answer.answer_id: relationship.relationship}
+    else:
+        answer_ids = schema.get_answer_ids_for_block(location.block_id)
+        answers = answer_store.get_answers_by_answer_id(
+            answer_ids=answer_ids, list_item_id=location.list_item_id
+        )
+        result = {answer.answer_id: answer.value for answer in answers if answer}
 
     return OrderedDict(sorted(result.items()))
--- a/app/jinja_filters.py
+++ b/app/jinja_filters.py
@@ -264,6 +264,25 @@ class RadioConfig:
             self.other = OtherConfig(detail_answer)
 
 
+class RelationshipRadioConfig:
+    def __init__(self, option, index, answer):
+        self.id = option.id
+        self.name = option.name
+        self.value = option.data
+        self.checked = option.checked
+
+        label_description = None
+        answer_option = answer['options'][index]
+
+        self.label = LabelConfig(option.id, option.label.text, label_description)
+
+        if answer_option:
+            self.attributes = {
+                'data-title': answer_option['title'],
+                'data-playback': answer_option['playback'],
+            }
+
+
 class OtherConfig:
     def __init__(self, detail_answer):
         self.id = detail_answer.id
@@ -294,6 +313,20 @@ def map_radio_config(form, answer):
 @blueprint.app_context_processor
 def map_radio_config_processor():
     return dict(map_radio_config=map_radio_config)
+
+
+@blueprint.app_template_filter()
+def map_relationships_config(form, answer):
+    options = form['fields'][answer['id']]
+
+    return [
+        RelationshipRadioConfig(option, i, answer) for i, option in enumerate(options)
+    ]
+
+
+@blueprint.app_context_processor
+def map_relationships_config_processor():
+    return dict(map_relationships_config=map_relationships_config)
 
 
 class SelectOptionConfig:
--- a/app/questionnaire/questionnaire_store_updater.py
+++ b/app/questionnaire/questionnaire_store_updater.py
@@ -1,6 +1,7 @@
 from typing import List, Tuple
 
 from app.data_model.answer_store import Answer
+from app.data_model.relationship_store import Relationship, RelationshipStore
 
 
 class QuestionnaireStoreUpdater:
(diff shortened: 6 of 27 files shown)


MebinAbraham (commented):
Just testing some functionality. I know this is not supposed to handle adding and removing of people etc.
When changing a relationship answer, we should remove the existing relationship and update with the new relationship?. At the moment it just adds another one which doesn't make sense?
```json
value": [
        {
          "from_list_item_id": "sRfHMx",
          "to_list_item_id": "QFgfUl",
          "relationship": "Husband or Wife"
        },
        {
          "from_list_item_id": "sRfHMx",
          "to_list_item_id": "QFgfUl",
          "relationship": "Legally registered civil partner"
        }
      ]
```

app/data_model/relationships_store.py
```diff
@@ -0,0 +1,84 @@
...
+
+    from_list_item_id: str
+    to_list_item_id: str
+    relationship: str
+
+    def for_json(self) -> Mapping:
+        return asdict(self)
+
+
+class RelationshipsStore:
```
iwootten: Relationship\*s\*Store doesn't seem like the best name here.

app/questionnaire/relationships_router.py
```diff
@@ -0,0 +1,38 @@
+from app.questionnaire.relationship_location import RelationshipLocation
+
+
+class RelationshipsRouter:
```
iwootten: Again, I think this would be better named in the singular.

.gitignore
```diff
@@ -92,6 +92,7 @@ requirements.txt
 
 # Ignore built schemas
 data/en/
+data/cy/census_household_gb_wls.json
```
iwootten, line 95: Is this needed?

iwootten (dismissed):

app/data_model/relationships_store.py
JamesGardiner: Why?

iwootten: Some of the classes introduced begin Relationship, others Relationship*s* - could lead to confusion during development.

JamesGardiner: I disagree. Singular class names represent a single object. The plural denotes that there are multiple relationships stored on the object.

app/questionnaire/relationships_router.py
JamesGardiner: This is routing between multiple relationships, so plural is the correct form

.gitignore
JamesGardiner: I don't think it is... I'll revert

app/data_model/relationships_store.py
iwootten: I'm not sure I follow - the AnswerStore stores answers, the ListStore stores lists so surely a RelationshipStore would store relationships?

.gitignore
ajmaddaford: It's needed so it doesn't get committed. It gets copied as part of the build schemas in https://github.com/ONSdigital/eq-survey-runner/blob/build-relationships-minus-primary-person/scripts/build_schemas.sh#L33.

app/data_model/relationships_store.py
```diff
@@ -0,0 +1,92 @@
...
+    def __init__(self, relationships: List[Mapping] = None) -> None:
+        self._is_dirty = False
+        self._relationships = (
+            [
+                Relationship(
+                    relationship['from_list_item_id'],
+                    relationship['to_list_item_id'],
+                    relationship['relationship'],
+                )
+                for relationship in relationships
```
MebinAbraham: Personal preference I guess. Instead of the `if else`. Thoughts? 
```python
        self._relationships = (
            [
                Relationship(
                    relationship['from_list_item_id'],
                    relationship['to_list_item_id'],
                    relationship['relationship'],
                )
                for relationship in relationships or []
            ]
        )
```

app/views/questionnaire.py
```diff
@@ -13,6 +13,7 @@
 from app.authentication.no_token_exception import NoTokenException
 from app.data_model.answer_store import AnswerStore
 from app.data_model.list_store import ListStore
+from app.data_model.relationship_store import RelationshipStore, Relationship
```
MebinAbraham: The filename needs updating, still using the plural form.

app/views/questionnaire.py
```diff
@@ -32,6 +33,8 @@
 from app.questionnaire.placeholder_renderer import PlaceholderRenderer
 from app.questionnaire.questionnaire_store_updater import QuestionnaireStoreUpdater
 from app.questionnaire.router import Router
+from app.questionnaire.relationship_location import RelationshipLocation
+from app.questionnaire.relationship_router import RelationshipRouter
```
MebinAbraham, line 37: The filename needs updating, still using the plural form.

app/questionnaire/relationships_router.py
```diff
@@ -0,0 +1,38 @@
...
+
+class RelationshipRouter:
+    def __init__(self, block_id, list_item_ids):
+        self.path = self._generate_relationships_routing_path(block_id, list_item_ids)
+
+    def can_access_location(self, location):
+        return location in self.path
+
+    def get_first_location(self):
+        return self.path[0]
```
MebinAbraham: Getting when adding people and then answering No. Was working previously.
```
get_first_location
    return self.path[0]
IndexError: list index out of range
```

app/data_model/relationships_store.py
```diff
@@ -0,0 +1,92 @@
...
+            ):
+                return relationship
+        return None
+
+    def add_or_update(self, relationship: Relationship):
+        relationship_to_update = self.get_relationship(
+            relationship.from_list_item_id, relationship.to_list_item_id
+        )
+        if relationship_to_update:
+            if relationship.relationship != relationship_to_update.relationship:
```
MebinAbraham: Can this `if` just be part of the first `if` since it has no `else` clause:
```python3
if relationship_to_update and relationship.relationship != relationship_to_update.relationship:
```

MebinAbraham (changes requested):
There is a redirect loop that just loops between the confirmation page when you say `No` to adding a person then change your answer to `Yes` and add some people and go to relationships.

app/forms/fields.py
```diff
@@ -54,6 +54,7 @@ def get_field(
         field = {
             'Checkbox': get_select_multiple_field,
             'Radio': get_select_field,
+            'Relationships': get_select_field,
```
iwootten: I think this field name should probably be updated to singular too.

iwootten (dismissed):
I'm actually getting a 500 error at the moment when I submit relationships. Looks like _build_map is trying to iterate over a NoneType object when the RelationshipStore is first created.

ajmaddaford:
@iwootten This is currently broken after the latest push.

ajmaddaford:
Closing while this is being worked on.

app/data_model/relationships_store.py
ajmaddaford: This is now a map, consistent with the answer store.

app/data_model/relationships_store.py
ajmaddaford: This has been changed to be consistent with the answer store remove.

app/questionnaire/relationships_router.py
ajmaddaford: This was possibly while the branch was broken. It's working now.

ajmaddaford:
@MebinAbraham @iwootten The relationship store now uses a map similar to the answer store, so we can lookup a relationship by the tuple `(from_id, to_id)`.

app/data_model/relationship_store.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def add_or_update(self, relationship: Relationship):
+        key = (relationship.from_list_item_id, relationship.to_list_item_id)
+
+        existing_relationship = self._relationships.get(key)
+
+        if existing_relationship != relationship:
+            self._is_dirty = True
+
+        self._relationships[key] = relationship
```
MebinAbraham: This should be nested under the `if` above?

app/helpers/form_helper.py
```diff
@@ -94,13 +96,22 @@ def get_mapped_answers(schema, answer_store, location):
...
-
-    answers = answer_store.get_answers_by_answer_id(
-        answer_ids=answer_ids, list_item_id=location.list_item_id
-    )
-    result = {answer.answer_id: answer.value for answer in answers if answer}
+    answer_ids = schema.get_answer_ids_for_block(location.block_id)
+    result = {}
+
+    if isinstance(location, RelationshipLocation):
+        answer = answer_store.get_answer(answer_ids[0])
```
bitdivision: Very minor, we could leave `answer_ids` as a set and use `next(iter(answer_ids))` rather than converting to a list and indexing.

app/data_model/relationship_store.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def remove_all_relationships_for_list_item_id(self, list_item_id: str):
```
MebinAbraham, line 66: The card mentioned that removing a person isn't handled but we have this method to do it, so we can use it here?
https://github.com/ONSdigital/eq-survey-runner/blob/c62316b70e3200192f6da5a99949f5d4ab304237/app/views/questionnaire.py#L477-L479
Were there other caveats around it hence we chose to not use it?

MebinAbraham (commented):
Looks good.

data-source/json/test_relationships.json
```diff
@@ -0,0 +1,174 @@
...
+                    "question": {
+                        "id": "relationship-question",
+                        "type": "General",
+                        "title": "Thinking of {first_person_name}, {second_person_name} is their <em>...</em>",
+                        "answers": [{
+                            "id": "relationship-answer",
+                            "mandatory": true,
+                            "type": "Relationship",
+                            "playback": "{second_person_name} is {first_person_name_possessive} <em>…</em>",
+                            "options": [{
```
iwootten, line 141: It would be good to get the full list of relationships in here, so when they're entered during a test we can represent a full household.

app/data_model/relationship_store.py
ajmaddaford: Yeah. That was copied from the answer store, so I've also updated it there.

app/data_model/relationship_store.py
ajmaddaford: It would add relationship specific code to the list collector views. I think we need to think about this and potentially have a solution where the list collector actions (add/edit/remove) can be monitored by observers (e.g. relationships) and acted on.

ajmaddaford: Also there would be conflicts with the primary person collector work currently in progress.

data-source/json/test_relationships.json
ajmaddaford: The idea was to have the smallest possible schema to demonstrate the functionality.

MebinAbraham (changes requested):
When you are asked to select a relationship, if you click a relationship ie `Husband or Wife` then click the `Save and sign out` button then it causes a 500 error.

```sh
File "eq-survey-runner/app/questionnaire/questionnaire_store_updater.py", line 98, in _update_questionnaire_store_with_form_data
    list_item_id=self._current_location.list_item_id,
AttributeError: 'RelationshipLocation' object has no attribute 'list_item_id'
```

scripts/test_schemas.sh
```diff
@@ -6,8 +6,9 @@ if [ "$1" == "--local" ] || [ "$2" == "--local" ]; then
 fi
 
 if [ "$run_docker" == true ]; then
-    docker pull onsdigital/eq-schema-validator:v3
-    validator="$(docker run -d -p 5001:5000 onsdigital/eq-schema-validator:v3)"
+    branch=Relationship-Collector
+    docker pull onsdigital/eq-schema-validator:$branch
```
bitdivision, line 10: Wow! 🎉

app/data_model/relationship_store.py
```diff
@@ -0,0 +1,94 @@
...
+    def for_json(self) -> Mapping:
+        return asdict(self)
+
+
+class RelationshipStore:
+    """
+    Stores and updates relationships.
+    """
+
+    def __init__(self, relationships: List[Mapping] = None) -> None:
```
bitdivision: I don't think it really matters, but this should be `Optional[List[Mapping]]

app/jinja_filters.py
```diff
@@ -264,6 +264,25 @@ def __init__(self, option, index, form, answer):
...
+
+        label_description = None
+        answer_option = answer['options'][index]
+
+        self.label = LabelConfig(option.id, option.label.text, label_description)
+
+        if answer_option:
+            self.attributes = {
+                'data-title': answer_option['title'],
+                'data-playback': answer_option['playback'],
```
bitdivision, line 282: We could clean some of these up by inheriting from a base class, maybe this should be done seperately though.

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -38,6 +39,29 @@ def is_dirty(self):
     def update_answers(self, form):
         self._update_questionnaire_store_with_form_data(form.data)
 
+    def update_relationship_answer(self, form_data, from_list_item_id, to_list_item_id):
+        answer_ids_for_question = self._schema.get_answer_ids_for_question(
+            self._current_question
+        )
+        relationship_answer_id = answer_ids_for_question[0]
```
bitdivision: Again, could be `next(...)`

app/questionnaire/relationship_location.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+@dataclass
+class RelationshipLocation:
+    block_id: str
+    from_list_item_id: str
+    to_list_item_id: str
+
+    @classmethod
+    def from_dict(cls, relationship_location_dict: Mapping):
```
bitdivision: Is there any reason to use this method? Could we just use `RelationShipLocation(**relationship_location_dict)`?

bitdivision (changes requested):
If you fill in a relationship then click previous, the radio is still selected but both the title and the description still show `Thinking of 1 1, 2 2 is their ...` rather than `Thinking of 1 1, 2 2 is their Husband or Wife`.

app/helpers/form_helper.py
bitdivision: As discussed IRL, the only operation we do on the result is to iterate over it. So this should be converted to a list.

app/jinja_filters.py
ajmaddaford: Yeah, I did think about this but decided it would be better to tackle this in the card about refactoring the way we use context.

app/helpers/form_helper.py
ajmaddaford: I hadn't quite realised the significance of the set being used for question variants will have duplicate answer ids. However, as the answer ids must be the same across all question variants, this seems unnecessary. I'll look into removing the set and getting the answer ids from the first variant.

app/questionnaire/questionnaire_store_updater.py
ajmaddaford: I think it's clearer accessing the first answer directly. I've changed this to a `get_relationship_answer_id_for_block` method in `QuestionnaireSchema` as it's a schema specific thing.

app/helpers/form_helper.py
ajmaddaford: I've added a `get_relationship_answer_id_for_block` method to `QuestionnaireSchema` so that we don't need access the list in this way here. I've also removed the use of of sets as they weren't necessary.

bitdivision (approved):
Looks great :+1:

app/helpers/form_helper.py
```diff
@@ -94,13 +96,24 @@ def get_mapped_answers(schema, answer_store, location):
...
 
-    block_id = location.block_id
-
-    answer_ids = schema.get_answer_ids_for_block(block_id)
-
-    answers = answer_store.get_answers_by_answer_id(
-        answer_ids=answer_ids, list_item_id=location.list_item_id
-    )
-    result = {answer.answer_id: answer.value for answer in answers if answer}
+    answer_ids = schema.get_answer_ids_for_block(location.block_id)
```
MebinAbraham: This is redundant. It is in both `if` and `else`

MebinAbraham (approved):
Just that 1 minor comment. Otherwise looks good. 👌

PR #2163, opened 2019-07-04
Add a Primary Person List Collector block.

bitdivision:
Allows collecting one person in a list who is stored as a primary person in the list store.

Designed as per https://github.com/ONSdigital/eq-survey-runner/pull/2126/

I have tried to keep functionality somewhat seperated from core list collector functionality since it is likely to be pulled out after census.

The functionality in perform_list_action is becoming unwieldy. I'm looking at how to refactor that now.

### How to review 
Compare the journey in test_list_collector_primary_person to the journey in v13 prototype of WLH.

### Checklist


app/data_model/list_store.py
```diff
@@ -12,24 +13,80 @@ def random_string(length):
...
 
+class ListRepresentation:
+    def __init__(self, list_items: List = None, context: Mapping = None):
+        if not list_items:
+            list_items = list()
+        if not context:
+            context = dict()
+
+        self.list_items = list_items
+        self.context = context
```
MebinAbraham: Instead of the ifs?
```python
class ListRepresentation:
    def __init__(self, list_items: List = None, context: Mapping = None):
        self.list_items = list_items or []
        self.context = context or {}

```

templates/primarypersonlistaddoreditquestion.html
```diff
@@ -0,0 +1 @@
+{% extends 'question.html' %}
```
bambrp: If not overriding anything, should this be an include too, in common with primarypersonlistcollector.html?

app/data_model/list_store.py
```diff
@@ -12,24 +13,80 @@ def random_string(length):
...
+
+    def __eq__(self, other):
+        if not isinstance(other, ListRepresentation):
+            return NotImplemented
+
+        return self.list_items == other.list_items and self.context == other.context
+
+    def remove_item_id(self, to_remove):
+        self.list_items.remove(to_remove)
```
MebinAbraham: Currently, if I pass the primary persons list item id to the remove URL, it removes them. And then results in a 500 if I change the answer to `No` since the id no longer exists.
```sh
File "eq-survey-runner/app/data_model/list_store.py", line 33, in remove_item_id
    self.list_items.remove(to_remove)
ValueError: list.remove(x): x not in list
```

app/data_model/list_store.py
```diff
@@ -12,24 +13,80 @@ def random_string(length):
...
+    def __init__(self, existing_items: Optional[List[Mapping]] = None):
         if not existing_items:
-            existing_items = []
+            existing_items = [{}]
+
+        self._lists: Mapping[str, ListRepresentation] = {
+            x.get('name'): ListRepresentation(
+                list_items=x.get('items'), context=x.get('context', {})
+            )
+            for x in existing_items
```
MebinAbraham: `for existing_item in existing_items`? or just refer to them as list_items?

tests/functional/generate_pages.py
```diff
@@ -194,11 +194,11 @@
 
 """
 
-LIST_SUMMARY_EDIT_LINK_GETTER = r"""  listEditLink(instance) { return `tbody:nth-child(${instance}) td:last-child a:first-child`; }
+LIST_SUMMARY_EDIT_LINK_GETTER = r"""  listEditLink(instance) { return `tbody:nth-child(${instance}) td:last-child a[aria-label^='Change'`; }
```
bambrp: This works great for now, though I would be happier still with something that wasn't subject to a potential content change (i.e. "could you make this say 'alter' instead?") Perhaps a template change to add a class we could select on, or a non user-visible attribute value of our own invention, such as eq-primary-person?

MebinAbraham (changes requested):
POST to `/questionnaire/people/ZlwafC/add-or-edit-primary-person/` is currently not idempotent.
Screenshot 2019-07-04 at 12 52 04

tests/functional/generate_pages.py
bitdivision: I think the only other reasonable way to do it would be to add a data tag. We do this elsewhere, but this would require a change to the pattern library in this case I think.

app/data_model/list_store.py
bitdivision: I'm using `item` now, keeps it a little shorter

bitdivision: Fixed.

app/data_model/list_store.py
bitdivision: Fixed.

tests/functional/generate_pages.py
bambrp: I hadn't known of the data-* attributes before, that would be a good way to do this, I like the side effects of populating the element.dataset in the DOM. Handy.

app/data_model/list_store.py
bitdivision: Fixed.

bitdivision:
Duplicates are now fixed @MebinAbraham

app/data_model/list_store.py
```diff
@@ -48,42 +100,64 @@ def is_dirty(self):
...
             del self._lists[list_name]
             self._is_dirty = True
 
-    def add_list_item(self, list_name):
+    def remove_primary_person(self, list_name):
+        if list_name in self._lists:
+            del self[list_name].context['primary_person']
+
+    def get_primary_person(self, list_name):
+        return self[list_name].context.get('primary_person')
```
ajmaddaford: This is specific to primary person. It would be cleaner if this was either `get_context(list_name)` and the caller extracts what it needs from the returned context, or `get_context_item(list_name, context_key)`.

app/data_model/list_store.py
bitdivision: Fixed.

app/data_model/list_store.py
```diff
@@ -48,42 +103,64 @@ def is_dirty(self):
...
+
+    def add_list_item(
+        self, list_name, list_item_id: str = None, context: Mapping = None
```
ajmaddaford: Why has this been changed to allow a predetermined list id?

app/data_model/list_store.py
```diff
@@ -48,42 +103,64 @@ def is_dirty(self):
...
+        if not list_item_id:
+            list_item_id = self.generate_identifier()
 
-        new_list_identifier = self._generate_identifier()
+        if list_item_id not in named_list.list_items:
+            named_list.list_items.append(list_item_id)
 
-        named_list.append(new_list_identifier)
+        if context:
+            named_list.context = context
```
ajmaddaford: I think the context needs to be separated out from the list items. Currently the full context is overwritten on adding a single list item. 

Also, we should consider making context a string for the caller as it's unnecessary to provide the list item id twice e.g. `add_context(list_item_id=123456, context='primary-person')`

app/data_model/list_store.py
```diff
@@ -12,24 +13,78 @@ def random_string(length):
     return ''.join(random.choice(ascii_letters) for _ in range(length))
 
 
+class ListRepresentation:
```
ajmaddaford: Possibly just `List`?

app/data_model/list_store.py
```diff
@@ -12,24 +13,78 @@ def random_string(length):
...
+
 class ListStore:
     """ Storage for lists.
+
+    Stores in the form:
+
+    ```
+    self._lists: {
+        list_name_1: ListRepresentation(
+            list_items: [1,2,3],
```
ajmaddaford: `list_items` could be shortened to `items`.

app/jinja_filters.py
```diff
@@ -514,32 +514,40 @@ def map_list_collector_config(
...
+                {
+                    'text': remove_link_text,
+                    'ariaLabel': remove_link_aria_label.format(item_name=item_name),
+                    'url': list_item.get('remove_link'),
+                    'attributes': {'data-qa': 'remove-item-link'},
+                }
+            )
+
+        if list_item.get('primary_person'):
+            rows.insert(0, new_row)
```
ajmaddaford: This shouldn't be needed as the list items should be in the correct order (they have to be so we can iterate over the sections in the right order).

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -38,13 +38,34 @@ def remove_answers(self, answer_ids: List):
...
 
+    def add_primary_person_and_answers(self, form, list_name, list_item_id):
+        """
+        If there is already an existing primary person, modify that item.
+        """
+        context = {'primary_person': list_item_id}
+
+        self._list_store.add_list_item(
+            list_name, list_item_id=list_item_id, context=context
+        )
```
ajmaddaford: I think it would be much clearer to have an `add_context` method.

app/templating/view_context.py
```diff
@@ -92,24 +95,32 @@ def build_view_context_for_list_collector(
...
 
     list_name = rendered_block['populates_list']
-    list_item_ids = list_store[list_name]
+    list_item_ids = list_store[list_name].list_items
 
     list_title_answer_ids = [
         answer['id'] for answer in rendered_block['add_block']['question']['answers']
     ]
 
+    primary_person = list_store.get_primary_person(list_name)
```
ajmaddaford: We need to review the use of context for the primary person. The list store abstracts it into a context, but here we're adding primary person specific logic to the view context.

MebinAbraham (commented):
As discussed in IRL @bitdivision, at the moment since the list item id is not stored in the list store before visiting the primary person page, you are able to POST arbitrary ids and essentially change primary person's id with an existing list item id etc.

So as planned, we should generate and store the list item id for the primary person before visiting the primary person add or edit block.

ajmaddaford (changes requested):

tests/functional/generate_pages.py
bitdivision: Fixed with data attribute

app/data_model/list_store.py
```diff
@@ -12,22 +13,92 @@ def random_string(length):
...
+        try:
+            self.items.remove(to_remove)
+        except ValueError:
+            pass
+
+        if self.primary_person == to_remove:
+            self.primary_person = None
+
+    @classmethod
+    def from_dict(cls, input_dict: dict):
```
JamesGardiner: Should probably use the `typing.Mapping` type here to be more generic in acceptance

app/data_model/list_store.py
```diff
@@ -12,22 +13,92 @@ def random_string(length):
...
+            items: [1,2,3]
+        }
+    ]
+    ```
     """
 
-    def __init__(self, existing_items=None):
-        if not existing_items:
-            existing_items = []
+    def __init__(self, existing_items: Mapping[str, ListRepresentation] = None):
```
JamesGardiner: Is this type hint accurate? I think it should be `Union[Mapping[str, ListRepresentation], None]`

app/data_model/list_store.py
```diff
@@ -48,42 +119,69 @@ def is_dirty(self):
...
+        return list_item_id
 
     def serialise(self):
-        return [
-            {'name': name, 'items': list_items}
-            for name, list_items in self._lists.items()
-        ]
+        output = []
+
+        for name, list_representation in self._lists.items():
```
JamesGardiner: Could keep this as a list comp:
```Python
return [list_representaion.to_dict(name=name) for name, list_representaion in self._lists.items()]
```

app/views/questionnaire.py
```diff
@@ -326,12 +333,14 @@ def post_block_handler(
     return redirect(next_location.url())
 
 
-def perform_list_action(
+def perform_list_action(  # noqa: C901 pylint: disable=too-complex
```
JamesGardiner: Can this not be refactored rather than disabling lint warnings?

scripts/test_schemas.sh
```diff
@@ -6,8 +6,8 @@ if [ "$1" == "--local" ] || [ "$2" == "--local" ]; then
 fi
 
 if [ "$run_docker" == true ]; then
-    docker pull onsdigital/eq-schema-validator:v3
-    validator="$(docker run -d -p 5001:5000 onsdigital/eq-schema-validator:v3)"
+    docker pull onsdigital/eq-schema-validator:3092-implement-primary-person-list
```
JamesGardiner: Reminder to revert these before merge

tests/app/data_model/test_list_store.py
```diff
@@ -1,18 +1,19 @@
 from unittest.mock import patch
-from app.data_model.list_store import ListStore
+import pytest
```
JamesGardiner, line 3: Although this is a test module, it should still follow PEP8 standards for import ordering. Do these not get parsed by the linter?

JamesGardiner (changes requested):

app/data_model/list_store.py
bitdivision: As discussed, changed this class to use a `primary_person` class property. 

Though this does pollute the list_store with a primary person, I don't think any further functionality will be built on top of this and it should be simple to rip out if we want to.

app/data_model/list_store.py
bitdivision: Fixed.

app/data_model/list_store.py
bitdivision: Changed.

app/data_model/list_store.py
bitdivision: This one is annoying because List conflicts with `typing.List`. I've been struggling to think of anything better.

app/data_model/list_store.py
bitdivision: Done

app/jinja_filters.py
bitdivision: Fixed.

app/questionnaire/questionnaire_store_updater.py
bitdivision: Probably not relevant anymore with `primary_person` property/.

tests/app/data_model/test_list_store.py
bitdivision: This one is still in-progress...

bitdivision: It does still pass linting though? I can't see anything wrong there, std library > external libraries > app imports?

app/views/questionnaire.py
bitdivision: Ideally yes. I'm still looking at how best to do that.

tests/app/data_model/test_list_store.py
JamesGardiner: I was under the impression that a blank line needed to be separating them to pass linting?

e.g.
```Python
from unittest.mock import patch

import pytest
```

JamesGardiner: As per PEP8:

```
Imports should be grouped in the following order:

    Standard library imports.
    Related third party imports.
    Local application/library specific imports.

You should put a blank line between each group of imports.
```

bitdivision: Ah, I don't think that's enforced. I'll add it in though.

bitdivision:
Going to wait until #2155 goes in before making any further changes to questionnaire.py on here.

app/data_model/list_store.py
iwootten: ListModel?

bitdivision: Sounds reasonable to me. Any objections @ajmaddaford ?

app/views/questionnaire.py
```diff
@@ -465,6 +483,68 @@ def perform_list_action(
     return list_collector_url
 
 
+def handle_primary_person_action(
```
iwootten, line 595: There are a lot of args to this method. I think a couple (router/routing_path) could be removed/replaced if rather than return the next url it was determined following the questionnaire store update?

app/data_model/list_store.py
ajmaddaford: 👍 It's better than anything I can think of!

app/views/questionnaire.py
bitdivision: The problem with this is that you end up having a load of if statements for the questionnaire store update, and then a load more ifs when you come to work out the next URL, since each of these block types have slightly different behaviour for both of those.

app/submitter/convert_payload_0_0_3.py
```diff
@@ -33,40 +33,57 @@ def convert_answers_to_payload_0_0_3(
...
+    """ Add answers from list_collector for a specific location.
+    Output is added to the `answers_payload` argument."""
+    collector = schema.get_block(location.block_id)['type']
+    sub_block_map = {
+        'ListCollector': 'add_block',
+        'PrimaryPersonListCollector': 'add_or_edit_block',
+    }
+    if collector in sub_block_map.keys():
+        sub_block_name = sub_block_map[collector]
+        answers_in_add_block = get_answers_for_list_collector_sub_block(
```
ajmaddaford: I think it would be more appropriate to add a method on `QuestionnaireSchema` to get the answer ids for list items, something like `get_answer_ids_for_list_items`. This would keep some block specific logic out of this method.

app/views/questionnaire.py
```diff
@@ -565,16 +566,24 @@ def perform_list_action(
         questionnaire_store_updater.add_list_item_and_answers(
             form, parent_block['populates_list']
         )
-    elif block['type'] == 'ListEditQuestion':
+    if block['type'] == 'ListEditQuestion':
         questionnaire_store_updater.update_answers(form)
 
+    if schema.is_primary_person_block_type(block['type']):
```
ajmaddaford: I can't see why primary person uses the `perform_list_action` method. It would be clearer if this was called from `post_block_handler`.

ajmaddaford (changes requested):

app/views/questionnaire.py
bitdivision: Done. Though annoyingly makes post_handler 'too complex'. I've ignored this for now on the basis that this is all going to be refactored fairly soon.

app/data_model/list_store.py
```diff
@@ -12,22 +13,91 @@ def random_string(length):
...
+    def serialise(self):
+        serialised = {'items': self.items, 'name': self.name}
+
+        if self.primary_person:
+            serialised['primary_person'] = self.primary_person
+
+        return serialised
+
+    def __repr__(self):
+        return f'<ListRepresentation name={self.name} items={self.items}, primary_person={self.primary_person}>'
```
MebinAbraham:
```suggestion
        return f'<ListModel name={self.name} items={self.items}, primary_person={self.primary_person}>'
```

app/data_model/list_store.py
```diff
@@ -12,22 +13,91 @@ def random_string(length):
...
+
+
 class ListStore:
     """ Storage for lists.
+
+    Stores in the form:
+
+    ```
+    self._lists: {
+        list_name_1: ListRepresentation(
```
MebinAbraham: `ListRepresentation` > `ListModel`

app/data_model/list_store.py
```diff
@@ -12,22 +13,91 @@ def random_string(length):
...
```
MebinAbraham:
```suggestion
        list_name_1: ListModel(
```

JamesGardiner (approved):

MebinAbraham (commented):
Looks good.

Is the previous link from a list collector supposed to take you to the driving question or should it go to the edit screen?. 

Or should there not be a previous link?. Non-primary person list collector doesn't have the previous link, not sure if that was intentional.

MebinAbraham (changes requested):
So there is no remove link for the primary person but if I manually POST to the remove person endpoint with the list item id of the primary person then I am able to remove the primary person. We should prevent this?

1. Add a primary person.
2. Add another person.
3. Click remove on 'another' person.
4. Change list item id in url to the primary person's list item id
5. Click Continue.

bitdivision:
The previous link is meant to take you back to the driving question, not the edit page.

https://github.com/ONSdigital/eq-survey-runner/blob/56932bc2302dc54755a1276881cb8bb507ed72ae/doc/architecture/decisions/0009-primary-person.md

You have just reminded me that I need to update that design doc though :)

bitdivision:
> So there is no remove link for the primary person but if I manually POST to the remove person endpoint with the list item id of the primary person then I am able to remove the primary person. We should prevent this?
> 
>     1. Add a primary person.
> 
>     2. Add another person.
> 
>     3. Click remove on 'another' person.
> 
>     4. Change list item id in url to the primary person's list item id
> 
>     5. Click Continue.

I think this came up before and the conclusion was that it was possible to remove a primary person by answering 'No', so this scenario wasn't a huge issue. 

Thinking about it more though, downstream would receive an inconsistent payload. The primary-person answer would be 'Yes' but that would not be reflected in the list store.

I'll see how much extra logic this would introduce, but would appreciate thoughts @ajmaddaford @JamesGardiner @MebinAbraham

bitdivision:
I am going to make the change to disallow removing a primary person from the list collector. The change is simple, and should avoid the possibility of some strange routing bugs.

app/data_model/list_store.py
JamesGardiner: ListyMcListFace

bitdivision:
@ajmaddaford @JamesGardiner @MebinAbraham I've made that change now.

JamesGardiner (approved):

MebinAbraham (approved):
👍Thailand awaits.

bitdivision:
Will wait for @ajmaddaford before squashing.

ajmaddaford (approved):

PR #2194, opened 2019-07-12
Fix default answers populating answer store

JamesGardiner:
Default answers were previously stored in the AnswerStore, even when an answer wasn't submitted by the user. This PR alters the behaviour so that default answers are never stored on the answer store. To continue routing correctly and displaying the right placeholders, a new check has been added to the get_answer_store_value function in rules.py, that populates an answer object based on it's default value. This only happens when a user supplied answer is not present on the answer store.

### How to review 
Complete surveys with default answers in their schema.
- Notice that for answers with default values, the default values aren't displayed in the summary and aren't passed downstream when an answer isn't provided.
- Check that placeholders are evaluated using the default answer when an answer isn't provided.
- Check that routing is evaluated using the default value when an answer isn't provided

### DIFF
--- a/app/questionnaire/questionnaire_schema.py
+++ b/app/questionnaire/questionnaire_schema.py
@@ -3,6 +3,7 @@ from typing import Union
 
 from flask_babel import force_locale
 
+from app.data_model.answer import Answer
 from app.validation.error_messages import error_messages
 
 DEFAULT_LANGUAGE_CODE = 'en'
@@ -82,6 +83,15 @@ class QuestionnaireSchema:  # pylint: disable=too-many-public-methods
         variants
         """
         return self._answers_by_id.get(answer_id)
+
+    def get_default_answer(self, answer_id):
+        try:
+            answer_schema = self.get_answers_by_answer_id(answer_id)[0]
+            answer = Answer(answer_schema['id'], answer_schema['default'])
+        except (KeyError, TypeError):
+            return None
+
+        return answer
 
     def get_add_block_for_list_collector(self, list_collector_id):
         add_block_map = {
--- a/app/questionnaire/rules.py
+++ b/app/questionnaire/rules.py
@@ -260,7 +260,8 @@ def get_answer_store_value(answer_id, answer_store, schema, routing_path=None):
     If answer is not on the routing path, return None
     If routing_path empty, return answer value
     """
-    answer = answer_store.get_answer(answer_id)
+    answer = answer_store.get_answer(answer_id) or schema.get_default_answer(answer_id)
+
     if not answer:
         return None
 
--- a/tests/app/questionnaire/test_placeholder_renderer.py
+++ b/tests/app/questionnaire/test_placeholder_renderer.py
@@ -9,6 +9,7 @@ from tests.app.app_context_test_case import AppContextTestCase
 class TestPlaceholderRenderer(AppContextTestCase):
     def setUp(self):
         super().setUp()
+
         self.question_json = {
             'id': 'confirm-date-of-birth-proxy',
             'title': 'Confirm date of birth',
--- a/tests/app/questionnaire/test_questionnaire_store_updater.py
+++ b/tests/app/questionnaire/test_questionnaire_store_updater.py
@@ -67,7 +67,7 @@ class TestQuestionnaireStoreUpdater(unittest.TestCase):
             'value': answer_value,
         }
 
-    def test_save_answers_data_with_default_value(self):
+    def test_default_answers_are_not_saved(self):
         answer_id = 'answer'
         default_value = 0
 
@@ -87,14 +87,7 @@ class TestQuestionnaireStoreUpdater(unittest.TestCase):
         )
         self.questionnaire_store_updater.update_answers(form)
 
-        assert self.answer_store.add_or_update.call_count == 1
-
-        created_answer = self.answer_store.add_or_update.call_args[0][0]
-        assert created_answer.__dict__ == {
-            'answer_id': answer_id,
-            'list_item_id': None,
-            'value': default_value,
-        }
+        assert self.answer_store.add_or_update.call_count == 0
 
     def test_empty_answers(self):
         string_answer_id = 'string-answer'
--- a/tests/integration/views/test_view_submission.py
+++ b/tests/integration/views/test_view_submission.py
@@ -268,6 +268,48 @@ class TestViewSubmissionTradingAs(IntegrationTestCase):
             self.assertInUrl('view-submission')
             self.assertNotInBody('(Integration Tests)')
 
+    def test_view_submission_default_shows_no_answer_provided(self):
+        no_trading_as_payload = {
+            'user_id': 'integration-test',
+            'period_str': 'April 2016',
+            'period_id': '201604',
+            'collection_exercise_sid': '789',
+            'questionnaire_id': '0123456789000000',
+            'ru_ref': '123456789012A',
+            'response_id': '1234567890123456',
+            'ru_name': 'Integration Testing',
+            'ref_p_start_date': '2016-04-01',
+            'ref_p_end_date': '2016-04-30',
+            'return_by': '2016-05-06',
+            'employment_date': '1983-06-02',
+            'region_code': 'GB-ENG',
+            'language_code': 'en',
+            'roles': [],
+            'trad_as': '',
+            'account_service_url': 'http://upstream.url',
+        }
+        with patch('tests.integration.create_token.PAYLOAD', no_trading_as_payload):
+            self.launchSurvey('test_view_submitted_response')
+            self.assertInBody('What is your favourite breakfast food')
+
+            form_data = {}
+
+            self.post(form_data)
+            self.assertInBody('Please enter test values (none mandatory)')
+
+            form_data = {
+                'test-currency': '12',
+                'square-kilometres': '345',
+                'test-decimal': '67.89',
+            }
+
+            self.post(form_data)
+            self.post(action=None)
+            self.get('submitted/view-submission')
+
+            self.assertInUrl('view-submission')
+            self.assertInBody('No answer provided')
+
 
 class TestViewSubmissionCompression(IntegrationTestCase):
     def test_compressed_submitted_response_data(self):
--- a/app/questionnaire/questionnaire_store_updater.py
+++ b/app/questionnaire/questionnaire_store_updater.py
@@ -118,18 +118,6 @@ class QuestionnaireStoreUpdater:
 
         for answer_id, answer_value in form_data.items():
 
-            # If answer is not answered then check for a schema specified default
-            if answer_value in self.EMPTY_ANSWER_VALUES:
-                answer = next(
-                    (
-                        answer
-                        for answer in self._current_question.get('answers', [])
-                        if answer['id'] == answer_id
-                    ),
-                    {},
-                )
-                answer_value = answer.get('default')
-
             if answer_id in answer_ids_for_question:
                 if answer_value not in self.EMPTY_ANSWER_VALUES:
                     answer = Answer(
--- a/data-source/json/test_default_with_skip.json
+++ b/data-source/json/test_default_with_skip.json
@@ -40,7 +40,7 @@
                                     }
                                 ],
                                 "id": "question",
-                                "title": "Don't enter an answer. A default value will be used",
+                                "title": "Question One",
                                 "type": "General"
                             }
                         },
@@ -53,13 +53,53 @@
                                         "id": "answer-two",
                                         "mandatory": false,
                                         "type": "Number",
-                                        "label": "Enter a Value"
+                                        "label": "Enter a Value",
+                                        "default": 1
                                     }
                                 ],
                                 "id": "question2",
-                                "title": "Enter an answer. It will be shown on the summary page",
+                                "title": "Question Two",
                                 "type": "General"
-                            }
+                            },
+                            "skip_conditions": [
+                                {
+                                    "when": [
+                                        {
+                                            "condition": "not equals",
+                                            "id": "answer-one",
+                                            "value": 1
+                                        }
+                                    ]
+                                }
+                            ]
+                        },
+                        {
+                            "type": "Question",
+                            "id": "number-question-three",
+                            "question": {
+                                "answers": [
+                                    {
+                                        "id": "answer-three",
+                                        "mandatory": false,
+                                        "type": "Number",
+                                        "label": "Leave blank"
+                                    }
+                                ],
+                                "id": "question3",
+                                "title": "Question Three",
+                                "type": "General"
+                            },
+                            "skip_conditions": [
+                                {
+                                    "when": [
+                                        {
+                                            "condition": "equals",
+                                            "id": "answer-two",
+                                            "value": 1
+                                        }
+                                    ]
+                                }
+                            ]
                         },
                         {
                             "type": "Summary",
--- a/data-source/json/test_view_submitted_response.json
+++ b/data-source/json/test_view_submitted_response.json
@@ -76,7 +76,8 @@
                                             }
                                         ],
                                         "q_code": "20",
-                                        "type": "Radio"
+                                        "type": "Radio",
+                                        "default": "Bacon"
                                     }
                                 ],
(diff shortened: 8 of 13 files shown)


app/questionnaire/rules.py
```diff
@@ -281,3 +290,34 @@ def get_list_count(list_store, list_name):
 
 def is_goto_rule(rule):
     return any(key in rule.get('goto', {}) for key in ('when', 'block', 'group'))
+
+
+def get_default_answer(answer_id, answer_store, schema):
+    try:
+        block = schema.get_block_for_answer_id(answer_id)
+    except TypeError:
+        raise NoDefaultAnswerError
```
bitdivision: What situation does this raise in? I'm not sure I like raising an error about default answers when there is a bigger problem if this happens. Isn't this more of a schema error?

app/questionnaire/rules.py
```diff
@@ -281,3 +290,34 @@ def get_list_count(list_store, list_name):
...
+
+    question = None
+    try:
+        question = block['question']
+    except KeyError:
+        variants = block['question_variants']
+        for variant in variants:
```
bitdivision: We have a method in schema_utils.py to evaluate variants. `choose_question_to_display` I believe, though that could be named better :)

bitdivision: Using that should eliminate all of this try/except

app/questionnaire/rules.py
```diff
@@ -281,3 +290,34 @@ def get_list_count(list_store, list_name):
...
+                question = variant['question']
+                break
+
+    if not question:
+        raise NoDefaultAnswerError
+
+    answers = [answer for answer in question['answers'] if answer['id'] == answer_id]
+
+    try:
+        answer = Answer(answers[0]['id'], answers[0]['default'])
```
bitdivision: This will break when there are multiple answers in a question?

tests/app/questionnaire/test_questionnaire_store_updater.py
```diff
@@ -67,7 +67,7 @@ def test_save_answers_with_form_data(self):
             'value': answer_value,
         }
 
-    def test_save_answers_data_with_default_value(self):
+    def test_answers_data_with_default_values_arent_saved(self):
```
bitdivision:
```suggestion
    def test_answer_data_with_default_values_arent_saved(self):
```

bitdivision (dismissed):

app/questionnaire/rules.py
bitdivision: You will need to pass in some more arguments, but those should be there anyway, as currently this won't work for rules based on metadata / lists.

app/questionnaire/rules.py
bitdivision: No it won't, answers is filtered on the answer_id.

app/questionnaire/rules.py
JamesGardiner: Just spoken to @ajmaddaford about this. I think we will make a change to the schema validations so that question variants in the same block require the same default answer. Won't need to check which of the variants was displayed then.

app/questionnaire/rules.py
```diff
@@ -281,3 +290,30 @@ def get_list_count(list_store, list_name):
 
 def is_goto_rule(rule):
     return any(key in rule.get('goto', {}) for key in ('when', 'block', 'group'))
+
+
+def get_default_answer(answer_id, answer_store, schema):
```
iwootten: From the name this method sounds like it should be on the questionnaire schema object and would always return the default answer. The schema logic below seems like it might be better placed there.

app/questionnaire/rules.py
```diff
@@ -281,3 +290,30 @@ def get_list_count(list_store, list_name):
...
+
+    question = None
+
+    try:
+        question = block['question']
+    except KeyError:
+        variants = block['question_variants']
+        question = variants[0]['question']
+
+    if not question:
```
iwootten: Isn't the problem that there's no default question here?

iwootten (changes requested):

app/questionnaire/rules.py
```diff
@@ -281,3 +290,30 @@ def get_list_count(list_store, list_name):
...
+        raise NoDefaultAnswerError
+
+    answers = [answer for answer in question['answers'] if answer['id'] == answer_id]
+
+    try:
+        answer = Answer(answers[0]['id'], answers[0]['default'])
+    except (KeyError, TypeError):
+        raise NoDefaultAnswerError
```
ajmaddaford: I don't think we should be using the `NoDefaultAnswerError` exception throughout this method. It's actually more likely for a default answer to not exist. It would be better to just return `None`.

ajmaddaford:
[Don't store default answers in the answer store (S)](https://trello.com/c/uQGZoGnS/3058-dont-store-default-answers-in-the-answer-store-s)

ajmaddaford (dismissed):
I think we should have tests for:
- Default value in placeholders
- The default answer is not in the submission data

app/questionnaire/questionnaire_schema.py
```diff
@@ -83,6 +84,28 @@ def get_answers_by_answer_id(self, answer_id):
...
+            return None
+
+        try:
+            question = block['question']
+        except KeyError:
+            question = block['question_variants'][0]['question']
+
+        answers = [
+            answer for answer in question['answers'] if answer['id'] == answer_id
+        ]
```
ajmaddaford: This should be a single answer rather than an array. Answer ids in a given question are unique, so this would only ever generate a list of 0 or 1 items.

app/questionnaire/questionnaire_schema.py
```diff
@@ -95,12 +95,10 @@ def get_default_answer(self, answer_id):
         except KeyError:
             question = block['question_variants'][0]['question']
 
-        answers = [
-            answer for answer in question['answers'] if answer['id'] == answer_id
-        ]
+        answer = question['answers'][0]
```
ajmaddaford: This will always choose the first answer. It should loop through all answers looking for one that matches the `answer_id` in the method call.

app/questionnaire/placeholder_parser.py
```diff
@@ -8,13 +8,16 @@ class PlaceholderParser:
     final values
     """
 
-    def __init__(self, language, answer_store=None, metadata=None):
+    def __init__(self, language, answer_store=None, metadata=None, schema=None):
```
iwootten: We should keep the order of the args in placeholder_parser/placeholder_renderer the same.

app/questionnaire/placeholder_renderer.py
```diff
@@ -38,8 +38,9 @@ class PlaceholderRenderer:
     strings
     """
 
-    def __init__(self, language, answer_store=None, metadata=None):
+    def __init__(self, language, schema, answer_store=None, metadata=None):
```
iwootten: Given placeholder parser doesn't require schema, it shouldn't be required here should it?

tests/app/questionnaire/test_placeholder_renderer.py
```diff
@@ -143,15 +241,36 @@ def test_renders_json(self):
 
     def test_errors_on_invalid_pointer(self):
 
-        renderer = PlaceholderRenderer(language='en')
+        renderer = PlaceholderRenderer(
+            language='en', schema=self.default_question_schema
```
iwootten: There shouldn't be a need to pass the schema into existing tests?

tests/integration/views/test_view_submission.py
```diff
@@ -268,6 +268,62 @@ def test_view_submission_does_not_show_parenthesis_trading_as_is_empty_string(se
...
+            'trad_as': '',
+            'account_service_url': 'http://upstream.url',
+        }
+        with patch('tests.integration.create_token.PAYLOAD', no_trading_as_payload):
+            self.launchSurvey('test_view_submitted_response')
+
+            # check we're on first page
+            self.assertInBody('What is your favourite breakfast food')
+
+            # We fill in our answer
```
iwootten: Not sure all these comments are necessary

app/questionnaire/placeholder_parser.py
```diff
@@ -8,13 +8,16 @@ class PlaceholderParser:
...
         self._answer_store = answer_store or AnswerStore()
         self._metadata = metadata
+        self._schema = schema
         self._transformer = PlaceholderTransforms(language)
 
     def _lookup_answer(self, answer_id):
-        answer = self._answer_store.get_answer(answer_id)
+        answer = self._answer_store.get_answer(
```
iwootten: Seems like the problem of getting a default answer is one that's going to regularly come up. Can we add a method to answer_store to return it given a schema object?

iwootten (changes requested):

app/questionnaire/questionnaire_schema.py
```diff
@@ -83,6 +84,29 @@ def get_answers_by_answer_id(self, answer_id):
...
+            return None
+
+        try:
+            question = block['question']
+        except KeyError:
+            question = block['question_variants'][0]['question']
+
+        for this_answer in question['answers']:
+            if this_answer['id'] == answer_id:
+                answer = this_answer
```
MebinAbraham: Can the variable `answer` be called something else because it is reassigned again below? Maybe `answer_schema`?

data-source/json/test_default.json
```diff
@@ -28,19 +28,36 @@
...
+                                        "id": "answer-one",
                                         "mandatory": false,
                                         "type": "Number",
                                         "label": "Leave blank",
                                         "default": 0
                                     }
                                 ],
                                 "id": "question",
-                                "title": "Don't enter an answer. Zero will be shown on summary",
+                                "title": "Don't enter an answer. A default value will be used",
```
MebinAbraham, line 43: Can we add/extend this schema to test `when` rules with a default answer?

MebinAbraham (changes requested):

tests/integration/views/test_view_submission.py
JamesGardiner: Yeah that's copypasta

app/questionnaire/placeholder_parser.py
JamesGardiner: Answer store shouldn't have knowledge of default answers. That should stay with the QuestionnaireSchema

data-source/json/test_default.json
berroar: Added a new test schema and a functional test.

MebinAbraham (approved):

tests/app/questionnaire/test_placeholder_renderer.py
```diff
@@ -3,12 +3,108 @@
...
     find_pointers_containing,
 )
+from app.questionnaire.questionnaire_schema import QuestionnaireSchema
 from tests.app.app_context_test_case import AppContextTestCase
 
 
 class TestPlaceholderRenderer(AppContextTestCase):
     def setUp(self):
         super().setUp()
+        self.default_question_schema = QuestionnaireSchema(
```
iwootten: This doesn't need to appear in the setup, given it's only used in a single test. Probably worth putting alongside the other placeholder schemas as test_placeholder_defaults or something, given theres value in being able to test it manually.

tests/app/questionnaire/test_placeholder_renderer.py
```diff
@@ -3,12 +3,108 @@
...
+                                                                                        'last-name',
+                                                                                    ],
+                                                                                },
+                                                                                'delimiter': ' ',
+                                                                            },
+                                                                        }
+                                                                    ],
+                                                                },
+                                                                {
+                                                                    'placeholder': 'age_in_years',
```
iwootten: This placeholder doesn't appear to be used in the test.

tests/app/questionnaire/test_placeholder_renderer.py
```diff
@@ -3,12 +3,108 @@
...
+                                        'title': 'Confirm date of birth',
+                                        'type': 'General',
+                                        'answers': [
+                                            {
+                                                'id': 'confirm-date-of-birth-answer-proxy',
+                                                'mandatory': True,
+                                                'options': [
+                                                    {
+                                                        'label': {
+                                                            'text': '{person_name} is 33 years old. Is this correct?',
```
iwootten: Should this value be fixed?

tests/app/questionnaire/test_placeholder_renderer.py
```diff
@@ -155,3 +251,20 @@ def test_errors_on_invalid_json(self):
...
+            answer_store=AnswerStore(
+                [
+                    {'answer_id': 'last-name', 'value': 'Aho'},
+                    {'answer_id': 'date-of-birth-answer', 'value': '1986-01-01'},
+                ]
+            ),
+            schema=self.default_question_schema,
+        )
+
+        rendered_schema = renderer.render(self.question_json.copy())
```
iwootten: Shouldn't this be rendering the dict within the default_question_schema?

iwootten (changes requested):

app/questionnaire/questionnaire_schema.py
```diff
@@ -83,6 +84,29 @@ def get_answers_by_answer_id(self, answer_id):
...
+        except KeyError:
+            question = block['question_variants'][0]['question']
+
+        for this_answer in question['answers']:
+            if this_answer['id'] == answer_id:
+                answer_schema = this_answer
+                break
+
+        try:
+            answer = Answer(answer_schema['id'], answer_schema['default'])
```
MebinAbraham, line 90: `answer_schema` may not exist by this point.

MebinAbraham (changes requested):

app/questionnaire/questionnaire_schema.py
```diff
@@ -95,11 +95,15 @@ def get_default_answer(self, answer_id):
...
             question = block['question_variants'][0]['question']
 
+        answer_schema = None
         for this_answer in question['answers']:
             if this_answer['id'] == answer_id:
                 answer_schema = this_answer
                 break
 
+        if not answer_schema:
+            return None
```
MebinAbraham: Since there are already multiple `try excepts` you could just add `NameError` into the except below instead of the `if`.
Thoughts?

JamesGardiner: I prefer the try/except :+1:

MebinAbraham: Ya, that's why I am saying we could remove the `if` and use the already defined try/except?

ajmaddaford: We could remove all of this and get the answer by id from `self._answers_by_id`

MebinAbraham: This is to get the schema default though? Ignore, got what you are saying.

JamesGardiner: garh that's what you get for copy pasting form a different module without thinking :man_facepalming:

tests/app/questionnaire/test_questionnaire_store_updater.py
```diff
@@ -87,14 +87,7 @@ def test_save_answers_data_with_default_value(self):
...
 
-        assert self.answer_store.add_or_update.call_count == 1
-
-        created_answer = self.answer_store.add_or_update.call_args[0][0]
-        assert created_answer.__dict__ == {
-            'answer_id': answer_id,
-            'list_item_id': None,
-            'value': default_value,
-        }
+        assert self.answer_store.add_or_update.call_count == 0
```
iwootten, line 90: This test doesn't assert what the method name suggests. Shouldn't we be checking the content of the answer store only?

iwootten (changes requested):

tests/app/questionnaire/test_questionnaire_store_updater.py
JamesGardiner: `test_default_answers_are_not_saved`?

iwootten (approved):

MebinAbraham (approved):

bitdivision:
Validator changes have caused some breakages here. Working on a fix now.

bitdivision:
https://github.com/ONSdigital/eq-schema-validator/pull/142

PR #2195, opened 2019-07-12
Check translation templates in travis

iwootten:

We want to fail builds where strings have been updated either within runner or within runners census individual schemas. This PR adds the following commands for building/testing a current version of the empty translation templates:

    make translations
    make test-translations

Both are written in python, as they share logic and an original bash version was becoming difficult to read. Test translations has been added to travis. Currently translations are compared ignoring creation date of files. Static messages also have context ignored, since this includes the source file/line no and may be updated frequently.

### How to review 

Checkout the branch, make a change to one of the schemas, or within runners strings. When you run test-translations, the script should fail and highlight which template has uncommitted updates.

### Checklist

* [x] Templates can be generated for both app and schema messages
* [x] Most recent versions of templates are checked in
* [x] PR's trigger Template creation and check


iwootten:
[Generate POT file on PR (M)](https://trello.com/c/g9OAcq7V/3138-generate-pot-file-on-pr-m)

scripts/test_translation_templates.py
```diff
@@ -0,0 +1,150 @@
...
+
+    return failure_count
+
+
+if __name__ == '__main__':
+    print("--- Testing Translations ---")  # noqa: T001
+
+    temp_dir = "app/translations/tmp"
+
+    os.makedirs(temp_dir, exist_ok=True)
```
ajmaddaford: You can use `tempfile.TemporaryDirectory` as a context manager to create a random directory outside of the runner directories and deal with the cleanup:

```python
with tempfile.TemporaryDirectory() as tmpdirname:
  # do stuff with tmpdirname
```

https://docs.python.org/3.7/library/tempfile.html#tempfile-examples

scripts/test_translation_templates.py
```diff
@@ -0,0 +1,150 @@
...
+import os
+import shutil
+import docker
+import sys
+import subprocess
+
+WELSH_REGIONS = ["wls", "eng"]
+NI_REGIONS = ["nir"]
+
+LANGUAGE_MAP = {"cy": WELSH_REGIONS, "gd": NI_REGIONS}
```
ajmaddaford: - The `eng` can be dropped as it won't be translated into Welsh. 
- There shouldn't be any reference to language in this script. This script should generate a pot file for each schema. pot files should be checked into the `app/translations` folder.

scripts/build_translation_templates.py
```diff
@@ -0,0 +1,10 @@
+from scripts.test_translation_templates import (
+    build_schema_templates,
+    build_static_template,
+)
+
+if __name__ == '__main__':
+    print("--- Build Translation Templates ---")  # noqa: T001
+
+    build_static_template('app/translations/messages.pot')
+    build_schema_templates("app/translations/")
```
ajmaddaford: Any reason not to have one script that does the extraction, that has something like a `--check`  command line argument for the testing?

scripts/test_translation_templates.py
```diff
@@ -0,0 +1,150 @@
...
+
+    build_static_template(f'{temp_dir}/messages.pot')
+    build_schema_templates(temp_dir)
+
+    static_failures = check_static_template("app/translations", temp_dir)
+    dynamic_failures = check_schema_templates("app/translations", temp_dir)
+
+    shutil.rmtree(temp_dir)
+
+    if dynamic_failures + static_failures == 0:
```
ajmaddaford: These two counts are never used as counts, they could be booleans.

scripts/test_translation_templates.py
```diff
@@ -0,0 +1,150 @@
...
+        print(  # noqa: T001
+            color(
+                "\nNo changes to translation source strings have been found.",
+                fg='green',
+            )
+        )
+    else:
```
ajmaddaford: If handled failures first the `else` wouldn't be needed.

scripts/test_translation_templates.py
```diff
@@ -0,0 +1,150 @@
...
+                    not l.startswith(param) for param in line_beginnings_to_ignore
+                ),
+                file.readlines(),
+            )
+        )
+
+
+def print_filename_results(filename, success=True):
+    if success:
+        print(color(f"{filename} - NO CHANGES", fg='green'))  # noqa: T001
```
ajmaddaford: Can we just use basic logging to avoid the `noqa` throughout the script?

ajmaddaford (changes requested):
Linting should be added to the scripts directory.

scripts/test_translation_templates.py
iwootten: This gives both the script and docker container access to the temp directory mounted within the docker volume. tempfile generates a directory under /var/folders which can't be mounted without special filesharing configuration under docker. https://docs.docker.com/docker-for-mac/osxfs/#namespaces

scripts/build_translation_templates.py
```diff
@@ -0,0 +1,10 @@
+from scripts.test_translation_templates import (
+    build_schema_templates,
+    build_static_template,
+)
+
+if __name__ == '__main__':
+    print("--- Build Translation Templates ---")  # noqa: T001
```
JamesGardiner: This should be a logger not a print statement.

scripts/test_translation_templates.py
```diff
@@ -0,0 +1,132 @@
...
+                lambda l: all(
+                    not l.startswith(param) for param in line_beginnings_to_ignore
+                ),
+                file.readlines(),
+            )
+        )
+
+
+def print_filename_results(filename, success=True):
+    if success:
```
JamesGardiner: Printing successes is just noise. I'd only be interested in those that fail.

JamesGardiner (changes requested):

scripts/test_translation_templates.py
iwootten: The main reason for this was to get coloured output in the script, given that's how we print other successful output in bash scripts.

scripts/test_translation_templates.py
iwootten: I disagree - we do this with eq-schema-validator tests, and it allows us to identify what's been processed. Not sure why we'd adopt a different approach here.

scripts/test_translation_templates.py
JamesGardiner: Tempfile should try the path /tmp first, which is available to mount as a volume in the container

JamesGardiner: https://docs.python.org/3/library/tempfile.html#tempfile.gettempdir

iwootten: Tempfile probably is limited by the virtualenv it is running under as I'm finding I have to explicitly pass /tmp as the dir to it to get it created under there on my machine.

JamesGardiner: Do you have any of the metnioned env vars set? That may be throwing off the resolution order. Virtualenv *shouldn't* cause any issues, but can't rule it out for sure without further info.

Passing `/tmp` explicitly should be good enough for now?

scripts/test_translation_templates.py
JamesGardiner: Do you ever parse the successful output line by line (either with human eyeball or some kind of scripted approach) to ensure it matches your expectations? If not, then I'm still pretty sure it's noise. It's far easier to grep the output if it is done on a per exception basis.

scripts/test_translation_templates.py
iwootten: Hmm, turns out I do - TMPDIR is getting set somewhere..

app/translations/census_individual_gb_eng.pot
```diff
@@ -0,0 +1,4062 @@
+# Translations template for PROJECT.
```
MebinAbraham: This file can now be removed?

ajmaddaford: This file shouldn't be checked in (as we're not translating the english region survey into any other languages).

MebinAbraham (commented):
Did we want to merge the translation `.pot` files?

scripts/test_translation_templates.py
```diff
@@ -0,0 +1,127 @@
...
+    )
+    target_contents = get_template_content(
+        f"{target_dir}/messages.pot", ignore_context=True
+    )
+
+    success = source_contents == target_contents
+
+    print_filename_results(f"{source_dir}/messages.pot", success)
+
+    return 0 if success else 1
```
MebinAbraham: Just return the boolean?
```suggestion
    return success
```

scripts/test_translation_templates.py
```diff
@@ -0,0 +1,127 @@
...
+                output_dir: {'bind': '/usr/src/templates', 'mode': 'rw'},
+            },
+            remove=True,
+        )
+        logger.info(f"Building {output_dir}/{template_file}")
+        logger.info(run_results.decode())
+
+
+def check_schema_templates(source_dir, target_dir):
+    failure_count = 0
```
MebinAbraham: Since this is not outputted anywhere, can we just use a boolean?
```python
def check_schema_templates(source_dir, target_dir):
    is_valid = True

    for schema_name in SCHEMAS_TO_EXTRACT:
        template_file = f"{schema_name}.pot"

        source_contents = get_template_content(f"{source_dir}/{template_file}")
        target_contents = get_template_content(f"{target_dir}/{template_file}")
        is_valid = source_contents == target_contents

        print_filename_results(f"app/translations/{template_file}", is_valid)

    return is_valid
```

scripts/test_translation_templates.py
```diff
@@ -0,0 +1,127 @@
...
+    logger.debug("--- Testing Translations ---")
+
+    with tempfile.TemporaryDirectory(dir='/tmp') as temp_dir:
+        build_static_template(f'{temp_dir}/messages.pot')
+        build_schema_templates(temp_dir)
+
+        static_failures = check_static_template("app/translations", temp_dir)
+        dynamic_failures = check_schema_templates("app/translations", temp_dir)
+
+        if dynamic_failures + static_failures != 0:
```
MebinAbraham: `static_failures` and `dynamic_failures` could be booleans since the count is not used?
```suggestion
        if not all((dynamic_failures, static_failures)):
```

scripts/test_translation_templates.py
iwootten: Yes. Every time we inspect the output of a test run, or lint run we're doing exactly that. This is how we know a process is doing something and not just hanging. I think changing how scripts work to only output errors is outside the scope of this PR.

scripts/run_translation_templates.py
```diff
@@ -0,0 +1,144 @@
+import os
```
ajmaddaford: I'm not sure `run` is the right word, would be clearer as `extract` or possibly `generate`.

scripts/test_translation_templates.py
JamesGardiner: Thought it might be something like that :smile: :+1:

JamesGardiner (approved):

scripts/run_translation_templates.py
```diff
@@ -0,0 +1,144 @@
...
+            static_success = check_static_template('app/translations', temp_dir)
+            dynamic_success = check_schema_templates('app/translations', temp_dir)
+
+            if not all((dynamic_success, static_success)):
+                logger.error(
+                    'Changes to translation source strings have been found. Please regenerate templates.'
+                )
+                sys.exit(1)
+
+            logger.debug('No changes to translation source strings have been found.')
```
ajmaddaford: I'm not sure that 'changes' is the right language here. I think it's more about checking that the template files are up to date. It would also be helpful to print the command to fix the problem. So we could say:

"Translation templates are not up to date. Run `make translation-templates` to fix this" 

or 

"Translation templates are up to date"

scripts/run_translation_templates.py
```diff
@@ -0,0 +1,144 @@
...
+        sys.exit(0)
+
+    logger.info('--- Build Translation Templates ---')
```
ajmaddaford: This line seems unnecessary in the output.

Makefile
```diff
@@ -16,6 +16,12 @@ test-functional:
 test-schemas:
 	pipenv run ./scripts/test_schemas.sh
 
+translation-templates:
+	pipenv run python -m scripts.run_translation_templates
+
+test-translation-templates:
+	pipenv run python -m scripts.run_translation_templates --check
```
ajmaddaford: Given the command line flag is `check` this should be `check-translation-templates`.

scripts/run_translation_templates.py
```diff
@@ -0,0 +1,144 @@
...
+        description='Extract translation templates from runner'
+    )
+    parser.add_argument(
+        '--check', help='Test the templates without making changes', action='store_true'
+    )
+
+    args = parser.parse_args()
+
+    if args.check:
+        logger.info('--- Testing Translations ---')
```
ajmaddaford: This line seems unnecessary in the output.

ajmaddaford (changes requested):

Makefile
iwootten: I actually think test is more appropriate here given the other script commands use it.

ajmaddaford (changes requested):
image

The error logging is not great. Success is in green, errors are white. The "Changes to .." error is also duplicated.

iwootten:
> The error logging is not great. Success is in green, errors are white. The "Changes to .." error is also duplicated.

I can't see the colour differences you're talking about here under terminal using bash or zsh.

scripts/generate_translation_templates.py
```diff
@@ -0,0 +1,145 @@
...
+                os.getcwd(): {'bind': '/usr/src/eq-survey-runner', 'mode': 'rw'},
+                output_dir: {'bind': '/usr/src/templates', 'mode': 'rw'},
+            },
+            remove=True,
+        )
+        logger.info('Building %s/%s', output_dir, template_file)
+        logger.info(run_results.decode())
+
+
+def check_schema_templates(source_dir, target_dir):
```
MebinAbraham, line 100: This is currently always True if the last schema is True.

```python
def check_schema_templates(source_dir, target_dir):
    return all(
        compare_files(source_dir, target_dir, f'{schema_name}.pot')
        for schema in SCHEMAS_TO_EXTRACT
    )
```

scripts/generate_translation_templates.py
```diff
@@ -0,0 +1,145 @@
...
+            static_success = compare_files('app/translations', temp_dir, 'messages.pot')
+            dynamic_success = check_schema_templates('app/translations', temp_dir)
+
+            if not all((dynamic_success, static_success)):
+                logger.error(
+                    'Translation templates are not up to date. Run make translation-templates to fix this'
+                )
+                sys.exit(1)
+
+            logger.debug('Translation templates are up to date.')
```
MebinAbraham, line 131: This is currently always printed, `check_schema_templates` needs changing.

MebinAbraham (changes requested):

MebinAbraham (approved):
👍

ajmaddaford (approved):

PR #2202, opened 2019-07-15
Fix error skip links

bameyrick:
Fix the skip links for errors which autofocus the related field.

Affected answer types:
- Date input (fixed by design system update in previous PR)
- Durations

### DIFF
--- a/app/jinja_filters.py
+++ b/app/jinja_filters.py
@@ -194,7 +194,7 @@ def should_wrap_with_fieldset(question):
     answers = question['answers']
 
     if len(answers) > 1 and not any(
-        answer['type'] in ['Date', 'MonthYearDate', 'Duration'] for answer in answers
+        answer['type'] in {'Date', 'MonthYearDate', 'Duration'} for answer in answers
     ):
         return True
 
--- a/app/jinja_filters.py
+++ b/app/jinja_filters.py
@@ -194,7 +194,7 @@ def should_wrap_with_fieldset(question):
     answers = question['answers']
 
     if len(answers) > 1 and not any(
-        answer['type'] == 'Date' or answer['type'] == 'Duration' for answer in answers
+        answer['type'] in ['Date', 'MonthYearDate', 'Duration'] for answer in answers
     ):
         return True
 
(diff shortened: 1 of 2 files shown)


bameyrick:
[Issue ID: DAC_Broken_Skip_Link_Issue1](https://trello.com/c/kvgKpncJ/3116-issue-id-dacbrokenskiplinkissue1)

app/jinja_filters.py
```diff
@@ -193,7 +193,9 @@ def setAttributes(dictionary, attributes):
 def should_wrap_with_fieldset(question):
     answers = question['answers']
 
-    if len(answers) > 1 and not any(answer['type'] == 'Date' for answer in answers):
+    if len(answers) > 1 and not any(
+        answer['type'] == 'Date' or answer['type'] == 'Duration' for answer in answers
```
MebinAbraham:
```suggestion
        answer['type'] in ['Date', Duration'] for answer in answers
```
Are there other types that need to be considered such as MonthYear date types etc?

MebinAbraham (commented):
How do I test this?. It seems to focus on the field on v3?

app/jinja_filters.py
```diff
@@ -193,7 +193,9 @@ def setAttributes(dictionary, attributes):
 def should_wrap_with_fieldset(question):
     answers = question['answers']
 
-    if len(answers) > 1 and not any(answer['type'] == 'Date' for answer in answers):
+    if len(answers) > 1 and not any(
+        answer['type'] in ['Date', 'MonthYearDate', 'Duration'] for answer in answers
```
MebinAbraham: We also have `YearDate` type. Also since Duration type is checked here, are `Unit` types special?

JamesGardiner:
```suggestion
        answer['type'] in {'Date', 'MonthYearDate', 'Duration'} for answer in answers
```

Set lookup is far quicker than list lookup

JamesGardiner (changes requested):

app/jinja_filters.py
bameyrick: Year only renders one field so it doesn't wrap itself in a fieldset, so if there were multiple year answers this would be valid. Units are also considered as an answer each so this should still cater for that

MebinAbraham (approved):

JamesGardiner (approved):

PR #2206, opened 2019-07-16
Update to to design system 13.2.1

bameyrick:
- Updated how field errors are rendered – Error object is now passed to input but only one error is accepted. This will require a refactor of how field errors are handled as they currently pass an array.
- Updated 500 error page to match what's documented in the design system: https://ons-design-system.netlify.com/patterns/error-status-pages/#500-server-errors
- Updated durations answer to use new duration component: https://ons-design-system.netlify.com/patterns/durations/


bameyrick:
[Update to design-system 13.1.0](https://trello.com/c/bMrbC4r6/3147-update-to-design-system-1310)

iwootten:
I've added a couple of changes here that move the title handling to the back end and gives the ability to set a default error page. It's likely that this will be handled better as part of the view refactor work coming in here https://github.com/ONSdigital/eq-survey-runner/pull/2228

templates/layouts/_error.html
```diff
@@ -1,17 +1,6 @@
...
-  {% set error_title = _("Page not found") %}
-{% elif status_code == 403 %}
-  {% set error_title = "Error 403" %}
-{% endif %}
-
-{% set survey_title = error_title + " – " + _("Office for National Statistics") | string %}
-
 {% block main %}
-  {% include 'partials/errors/' ~ status_code | lower ~ '.html' %}
+  {% include 'partials/errors/' ~ status_partial ~ '.html' %}
```
ajmaddaford: Why have the include? The relevant template is known by the handling code in `errors.py`.

ajmaddaford (changes requested):

templates/errors/401.html
```diff
@@ -0,0 +1,6 @@
+{% extends 'layouts/_base.html' %}
+
+{% block main %}
+<h1>{{ _("Sorry, there is a problem") }}</h1>
+<p>{{ _("To access this page you need to <a href='#'>enter your unique access code.</a>") }}</p>
+{% endblock %}
```
ajmaddaford: I don't think this template is used, as 401 errors seem to use the `session-expired` template?

app/routes/errors.py
```diff
@@ -8,14 +9,29 @@
...
+DEFAULT_ERROR = 500
+
+
+def get_error_title(status_code):
+    return (
+        ERROR_MAP.get(status_code, ERROR_MAP.get(DEFAULT_ERROR))
+        + ' - '
+        + _('Office for National Statistics')
+    )
+
```
ajmaddaford: We could remove the error map and move the strings to the templates, setting `page_title` as we do for example in `signed-out.html`.

app/routes/errors.py
```diff
@@ -69,21 +63,8 @@ def log_exception(error, status_code):
...
-    template = template or 'error'
-
-    return (
-        render_template(
-            template=template, status_code=status_code, ua=user_agent, tx_id=tx_id
-        ),
-        status_code,
-    )
+    status_template = status_code
+    template = template or status_template
```
ajmaddaford:
```suggestion
    template = template or status_code
```

ajmaddaford (approved):

app/routes/errors.py
```diff
@@ -39,7 +33,7 @@ def forbidden(error=None):
 @errors_blueprint.app_errorhandler(SubmissionFailedException)
 def service_unavailable(error=None):
     log_exception(error, 503)
-    return _render_error_page(503)
+    return _render_error_page(500)
```
bitdivision: Why has this been changed to 500? And should it log a 500 too?

bitdivision (approved):
Changes all seem reasonable. 

As discussed with @ajmaddaford and @iwootten we should remove all references to 503 (in logs) and check whether displaying a normal 500 is okay on a failed submission.

bitdivision (approved):

PR #2208, opened 2019-07-17
Add channel claim to downstream submission

bambrp:
We are required to pass some new claims downstream- channel, case_type and region_code. The latter two had been ingested from RH, but were not being passed on the the submission JSON. The channel claim wasn't being handled at all.
This adds validation for channel, enforcing an enumerated list of FIELD, RH, AD and CC, if channel is present (it should be for all census surveys, but making it mandatory would break all the other schemas).

### How to review 
Using the eq-2754-add-channel-to-submission branch of eq-questionnaire-launcher, you can inject values for the new claim types, and test they would be sent downstream using the /dump/submission API endpoint during a survey.

### DIFF
--- a/app/storage/metadata_parser.py
+++ b/app/storage/metadata_parser.py
@@ -82,6 +82,9 @@ class RunnerMetadataSchema(Schema, StripWhitespaceMixin):
     roles = fields.List(fields.String(), required=False)
     survey_url = VALIDATORS['url'](required=False)
     language_code = VALIDATORS['string'](required=False)
+    channel = VALIDATORS['string'](
+        required=False, validate=validate.OneOf(('RH', 'FIELD', 'CC', 'AD'))
+    )
 
     # Either schema_name OR the three census parameters are required. Should be required after census.
     schema_name = VALIDATORS['string'](required=False)
@@ -120,10 +123,8 @@ class RunnerMetadataSchema(Schema, StripWhitespaceMixin):
         """
         if data.get('schema_name'):
             logger.info(
-                f'Ignoring claims: survey: {data.get("survey")}, case_type: {data.get("case_type")} because schema_name was specified'
+                'Using schema_name claim to specify schema, overriding survey, case_type and region_code'
             )
-            data.pop('survey', None)
-            data.pop('case_type', None)
         else:
             data['schema_name'] = get_schema_name_from_census_params(
                 data.get('survey'), data.get('case_type'), data.get('region_code')
--- a/app/submitter/converter.py
+++ b/app/submitter/converter.py
@@ -35,19 +35,21 @@ def convert_answers(schema, questionnaire_store, routing_path, flushed=False):
         'submitted_at': '2016-03-07T15:28:05Z',
         'response_id': '1234567890123456',
         'questionnaire_id': '1234567890000000',
+        'channel': 'RH',
         'metadata': {
           'user_id': '789473423',
           'ru_ref': '432423423423'
         },
-        'data': {
+        'data': [
             ...
-        },
+        ],
       }
     ```
 
     Args:
         schema: QuestionnaireSchema instance with populated schema json
-        full_routing_path: The full routing path followed by the user when answering the questionnaire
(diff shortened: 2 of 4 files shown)


bambrp:
[Add channel, case_type & region_code to metadata for downstream (S)](https://trello.com/c/VeQ2MjkS/2754-add-channel-casetype-regioncode-to-metadata-for-downstream-s)

bambrp:
Build has shown some fixes needed to tests due to changing output. Will work on this today.

app/storage/metadata_parser.py
```diff
@@ -120,10 +123,8 @@ def convert_schema_name(self, data, **kwargs):
         """
         if data.get('schema_name'):
             logger.info(
-                f'Ignoring claims: survey: {data.get("survey")}, case_type: {data.get("case_type")} because schema_name was specified'
+                f'Using schema_name claim to specify schema, overriding survey, case_type and region_code'
```
JamesGardiner:
```suggestion
                'Using schema_name claim to specify schema, overriding survey, case_type and region_code'
```

app/submitter/converter.py
```diff
@@ -74,6 +76,12 @@ def convert_answers(schema, questionnaire_store, routing_path, flushed=False):
         'questionnaire_id': metadata['questionnaire_id'],
     }
 
+    if metadata.get('channel'):
```
JamesGardiner, line 79: Can the dict objects be set to None and then ignored downstream? If yes, the if statements could be removed here.

JamesGardiner (changes requested):

app/submitter/converter.py
MebinAbraham: I think to keep it consistent it should only be added to the payload if it exists.

app/submitter/converter.py
```diff
@@ -35,19 +35,21 @@ def convert_answers(schema, questionnaire_store, routing_path, flushed=False):
         'submitted_at': '2016-03-07T15:28:05Z',
         'response_id': '1234567890123456',
         'questionnaire_id': '1234567890000000',
+        'channel': 'rh',
```
MebinAbraham:
```suggestion
        'channel': 'RH',
```

app/submitter/converter.py
```diff
@@ -35,19 +35,21 @@ def convert_answers(schema, questionnaire_store, routing_path, flushed=False):
...
             ...
-        },
+        ],
       }
     ```
 
     Args:
         schema: QuestionnaireSchema instance with populated schema json
-        full_routing_path: The full routing path followed by the user when answering the questionnaire
+        questionnaire_store: EncryptedQuestionnaireStorage instance for accessing current questionnaire data
```
MebinAbraham, line 51:
```suggestion
        questionnaire_store: QuestionnaireStorage instance for accessing current questionnaire data
```

MebinAbraham (approved):

app/submitter/converter.py
pricem14pc: For Census, channel should technically be a mandatory claim. However, that's simply because we have a requirement to send it downstream as opposed to runner needing it for any processing. It has been discussed with Data Architecture that channel should actually be a mandatory claim post census (i.e. all ONS submitted data should include the channel of launch.. that may well require a None value, or a JWT validation error at launch). However, at this point in the proceedings I'm happy to leave it as it is (i.e. required=false at eQ launch and then checked for its presence during submission). Runner will not break if it is not included. It is up to the calling system (e.g. RH or CC) to include it.

pricem14pc (approved):
Census schema submitted response as agreed with DAP.

app/submitter/converter.py
bambrp: Yes, it is an interesting case of us not caring about channel ourselves, but gatekeeping for downstream, proxying the rules further up the chain of custody. Arguably, the census schema should be the contract specifying what is essential and what is optional, agreed by all parties. The rules should probably not be internal to runner.

JamesGardiner (approved):

PR #2211, opened 2019-07-17
Seperate NI schema from ENG/WLS

MebinAbraham:
Separates the NI schema from ENG/WLS. NI will be changing a lot more and diverging further and further from ENG/WLS so it makes it easier to manage and review by accepting some duplication for now.

### How to review 
Ensure:
- Built schema JSON on v3 matches built schemas in this branch. *Run: `make build`*


MebinAbraham:
[Split NI schema from England and Wales](https://trello.com/c/Yo96vYoL/3161-split-ni-schema-from-england-and-wales)

iwootten (approved):
Looks good to me! 👍

bambrp (approved):
As you note, this introduces lots of duplication the common blocks were designed to remove, but as they drift further apart, it doesn't seem worth the complication any more.

PR #2213, opened 2019-07-18
Make Translation Part of Build Process

iwootten:
Remove the committed welsh schema and add translations. Create a script to generate translated schemas on build. Additionally compile translations to generate mo files from committed static messages.

### How to review 

A new make translate command has been added, this should generate translated versions of census individual in the appropriate data/ directories.


iwootten:
[Remove committed welsh schema (S)](https://trello.com/c/doUno3jc/3140-remove-committed-welsh-schema-s)

scripts/translate_schemas.py
```diff
@@ -0,0 +1,42 @@
+import logging
+import os
+import subprocess
+import docker
+
+logger = logging.getLogger(__name__)
+
+TRANSLATION_MAP = {
+    'cy': ['census_individual_gb_wls'],
+    'gd': ['census_individual_gb_nir'],
```
MebinAbraham:
```suggestion
    'ga': ['census_individual_gb_nir'],
```

scripts/translate_schemas.py
```diff
@@ -0,0 +1,47 @@
+import logging
+import os
+import sys
+import subprocess
+import docker
```
JamesGardiner, line 5:
```suggestion

import docker
```

JamesGardiner (changes requested):

scripts/translate_schemas.py
iwootten: I'm confused as to why this should block the PR?

scripts/translate_schemas.py
```diff
@@ -0,0 +1,47 @@
...
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.DEBUG)
+logger.addHandler(logging.StreamHandler(sys.stdout))
+
+TRANSLATION_MAP = {
+    'cy': ['census_individual_gb_wls'],
+    'ga': ['census_individual_gb_nir'],
```
MebinAbraham: As discussed, we should remove this until we have NIR translations?

scripts/translate_schemas.py
```diff
@@ -0,0 +1,45 @@
...
+                f'pipenv run python -m cli.translate_survey ../eq-survey-runner/data/en/{schema_file} {language_dir}/{translation_file} {output_dir}',
+                volumes={
+                    os.getcwd(): {'bind': '/usr/src/eq-survey-runner', 'mode': 'rw'}
+                },
+                remove=True,
+                detach=True,
+            )
+            for log in container.logs(stdout=True, stream=True):
+                logger.info(log.decode().strip('\n'))
+            logger.info('Building %s/%s', output_dir, translation_file)
```
ajmaddaford: The log line should be before the docker run command. Currently it's printed at the end and there's no output while the docker command is running so it looks like it's hanging.

scripts/translate_schemas.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+if __name__ == '__main__':
+    subprocess.run(['pipenv', 'run', 'pybabel', 'compile', '-d', 'app/translations'])
```
ajmaddaford: I don't think this should be in this script as it's nothing to do with translating schemas. The command could just be added to the Makefile?

scripts/translate_schemas.py
ajmaddaford: Also, the log line is a bit odd. 
```
Building /usr/src/eq-survey-runner/data/cy/census_individual_gb_wls_cy.po
```
- It has an irrelevant path used within the docker container (`usr/src/eq-survey-runner`)
- The po filename doesn't exist in that directory, this should probably be `schema_file`

ajmaddaford (changes requested):

scripts/translate_schemas.py
iwootten: Maybe the name of the script needs to be changed instead?

iwootten: I kind of think we need one command to "do translations" rather than needing to know that there are different types of translations across multiple commands..

ajmaddaford: That's what the Makefile provides..

iwootten: Ah - I see, originally misread that.

Makefile
```diff
@@ -16,11 +16,15 @@ test-functional:
 test-schemas:
 	pipenv run ./scripts/test_schemas.sh
 
+translate:
```
MebinAbraham, line 19: I think we need better namings for these. 
Maybe `make translations` etc

MebinAbraham (approved):
Am able to translate a schema so 👍

Makefile
iwootten: I thought about this, but then the translations are really the .po files...

ajmaddaford (approved):

scripts/translate_schemas.py
JamesGardiner: Imports should be grouped in the following order:

    Standard library imports.
    Related third party imports.
    Local application/library specific imports.

You should put a blank line between each group of imports.

iwootten: Oh right, it's part of PEP8 suggestions, but this isn't something that any of our linters check.

iwootten: This has been updated in a previous commit.

JamesGardiner: I don't see any change to the file?

iwootten: Don't know why that would be - was changed in this commit: https://github.com/ONSdigital/eq-survey-runner/pull/2213/commits/d2a80077f3856a9475117b701c07bfcea2a91270

JamesGardiner (approved):

PR #2224, opened 2019-07-22
Change channel validation to not check for explicit values

bambrp:
Due to RH not at this time sending one of the required values for the channel claim, we will skip checking for any specific value, instead just checking it is a string of minimum length one.

### How to review 
Load up a census schema using questionnaire-launcher, and test with values for channel, such as "", "X", or "RH" (only the last one would have worked prior to this PR). All non-empty values should work, empty should give you a 500 error.

### DIFF
--- a/app/storage/metadata_parser.py
+++ b/app/storage/metadata_parser.py
@@ -82,9 +82,7 @@ class RunnerMetadataSchema(Schema, StripWhitespaceMixin):
     roles = fields.List(fields.String(), required=False)
     survey_url = VALIDATORS['url'](required=False)
     language_code = VALIDATORS['string'](required=False)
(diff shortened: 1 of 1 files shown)


bambrp:
[Change channel validation to not check for explicit values](https://trello.com/c/y1gV1Ua2/3172-change-channel-validation-to-not-check-for-explicit-values)

pricem14pc (approved):
Working as expected

iwootten (commented):
I'm a bit confused by this. The test examples given are a bit odd. Not sure why anyone would send a quoted string of anything? An empty string obviously fails at the moment.

iwootten (approved):
Forget what I said, misunderstood the final statement.

PR #2226, opened 2019-07-23
Move schema copy back in to build schemas

iwootten:

Our Dockerfile currently calls the build_schemas.sh script to generate schemas, but the copy of household schemas has been moved to build.sh since translation has been made part of the build process. Currently, this means that the current household schemas are not currently being created in any environment. 

Translation needs to happen post build of schemas. The Welsh language variation of household is not necessary for our current environments so has been removed for now given it depends on translation. 

### How to review 

Check that the copy command has been correctly moved back into build_schemas.sh and that make build/docker build creates schemas as expected.

A subsequent card to review making docker/travis use the same build script will be necessary at a later date.

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -14,12 +14,4 @@ echo "Building schemas"
 echo "Translating schemas"
 make translate
 
-echo "Creating temporary household schemas based on individual schema in 'data' dir"
-
-# Until household is delivered, temporarily create household schema as copies of the individual schema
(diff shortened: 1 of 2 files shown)


MebinAbraham (approved):

iwootten:
I think this may need further thought - given that currently we don't have any environment that would have translated schemas due to this dockerfile setup.

JamesGardiner (approved):

PR #2228, opened 2019-07-24
Refactor view logic

berroar:

The logic in questionnaire.py needed some refactoring.

### What has changed?
- The top level `views` folder has been renamed `routes`. The intention is logic should be moved out of the routes and into view handlers.
- The top level `views` folder now contains `contexts` and `handlers`.
  - `contexts` contains what was previously in `templating`. Code here is responsible for generating a context to be used when rendering a template. 
  - `handlers` contains the block handling logic previously in `questionnaire.py`

A block handler is now created on each request. Each block hander exposes a number of defined methods used in the route handling:
- `is_location_valid` - should the user be able to access the requested location
- `get_previous_location_url` - returns the previous url, this is overridden for certain block types e.g. list actions
- `get_previous_location_url` - returns the next url, this is overridden for certain block types e.g. list actions
- `handle_post` - deal with POST request, persist changes and update completeness

### Other related changes
- The hub is now displayed when visiting `/questionnaire` rather than `/questionnaire/sections`
- If the user tries to access any invalid locations they are now redirected to `/questionnaire`. Previously we redirected to the list collector if an invalid child list location was requested.

### How to review 
Check that nothing has been broken by this change.


MebinAbraham:
[Refactor view logic in questionnaire.py (L)](https://trello.com/c/b1tBaPXn/3131-refactor-view-logic-in-questionnairepy-l)

app/contexts/blocks/list_collector.py
```diff
@@ -0,0 +1,87 @@
+from flask import url_for
+from flask_babel import lazy_gettext as _
+
+
```
JamesGardiner: Delete one of these lines

app/contexts/blocks/list_collector.py
```diff
@@ -0,0 +1,87 @@
+from flask import url_for
+from flask_babel import lazy_gettext as _
```
JamesGardiner: Does importing as `_` not trigger any lint errors?? This is a terrible choice of import name for a package

app/handlers/block_factory.py
```diff
@@ -0,0 +1,48 @@
...
+    'Introduction': Content,
+    'Interstitial': Content,
+    'Confirmation': Content,
+    'Summary': Summary,
+    'SectionSummary': SectionSummary,
+    'CalculatedSummary': CalculatedSummary,
+}
+
+
+class BlockFactory:
```
JamesGardiner: I still don't see the benefit of this being a class - it would be simpler to have it as a module level function

app/handlers/blocks/block.py
```diff
@@ -0,0 +1,124 @@
...
+
+from app.data_model.progress_store import CompletionStatus
+from app.questionnaire.location import InvalidLocationException
+from app.questionnaire.path_finder import PathFinder
+from app.questionnaire.placeholder_renderer import PlaceholderRenderer
+from app.questionnaire.questionnaire_store_updater import QuestionnaireStoreUpdater
+from app.questionnaire.router import Router
+from app.questionnaire.schema_utils import transform_variants
+
+logger = get_logger()
```
JamesGardiner: It doesn't look like logger is used

app/views/questionnaire.py
```diff
@@ -182,483 +180,103 @@ def get_hub_section(schema, questionnaire_store, section_id):
...
-    list_item_id = current_location.list_item_id
-    list_name = current_location.list_name
-    block = schema.get_block(current_location.block_id)
-    parent_block = schema.get_list_collector_for_block_id(current_location.block_id)
-
-    if list_name != parent_block['for_list']:
-        logger.info(
-            f'Mismatched list_name: {list_name} and block_id: {current_location.block_id}'
+    try:
+        block_handler = BlockFactory.get_handler(
```
JamesGardiner:
```suggestion
        block_handler = get_handler(
```

app/views/questionnaire.py
```diff
@@ -182,483 +180,103 @@ def get_hub_section(schema, questionnaire_store, section_id):
...
+    schema, questionnaire_store, block_id, from_list_item_id, to_list_item_id
 ):
-    return post_block_handler(
-        routing_path, schema, questionnaire_store, block_id, list_name, list_item_id
+    current_location = RelationshipLocation(
+        block_id, from_list_item_id, to_list_item_id
     )
 
+    try:
+        block_handler = BlockFactory.get_handler(
```
JamesGardiner:
```suggestion
        block_handler = get_handler(
```

tests/app/handlers/test_block_factory.py
```diff
@@ -0,0 +1,22 @@
+from unittest import TestCase
+
+from mock import Mock
+
+from app.handlers.block_factory import BlockFactory
+from app.questionnaire.location import InvalidLocationException, Location
+
+
+class TestBlockFactory(TestCase):
```
JamesGardiner: This will need refactoring when the BlockFactory class becomes a function

tests/app/handlers/test_block_factory.py
```diff
@@ -0,0 +1,22 @@
...
+    def test_get_handler_invalid_block(self):
+        schema = Mock()
+        schema.get_block = Mock(return_value=None)
+        with self.assertRaises(InvalidLocationException):
+            location = Location(block_id='block-id')
+            BlockFactory.get_handler(schema, location, None, None)
```
JamesGardiner:
```suggestion
            get_handler(schema, location, None, None)
```

tests/app/handlers/test_block_factory.py
```diff
@@ -0,0 +1,22 @@
...
+
+    def test_get_handler_invalid_block_type(self):
+        schema = Mock()
+        schema.get_block = Mock(return_value={'type': 'MadeUpType'})
+        with self.assertRaises(InvalidLocationException):
+            location = Location(block_id='block-id')
+            BlockFactory.get_handler(schema, location, None, None)
```
JamesGardiner:
```suggestion
            get_handler(schema, location, None, None)
```

tests/integration/views/test_errors.py
```diff
@@ -57,7 +57,8 @@ def test_errors_500(self):
         # Patch out a class in post to raise an exception so that the application error handler
         # gets called
         with patch(
-            'app.views.questionnaire.Router', side_effect=Exception('You broked it')
+            'app.views.questionnaire.BlockFactory',
```
JamesGardiner: Again this will need updating when we aren't patching the class

tests/integration/views/test_errors.py
```diff
@@ -75,7 +76,8 @@ def test_errors_500_with_payload(self):
             # Patch out a class in post to raise an exception so that the application error handler
             # gets called
             with patch(
-                'app.views.questionnaire.Router', side_effect=Exception('You broked it')
+                'app.views.questionnaire.BlockFactory',
```
JamesGardiner: And here

JamesGardiner (changes requested):

app/views/questionnaire.py
```diff
@@ -109,11 +103,15 @@ def before_post_submission_request():
 @with_questionnaire_store
 @with_schema
 def get_questionnaire(schema, questionnaire_store):
+    if schema.is_hub_enabled():
+        return redirect(url_for('.get_hub'))
```
MebinAbraham: 👍 

Since this has been added to this endpoint, we can remove:

https://github.com/ONSdigital/eq-survey-runner/blob/66f6c019ec7735c31589f51d399de3a0c21f4f69/app/views/session.py#L97-L98

Could keep in order to reduce the no. of redirects.

app/views/questionnaire.py
```diff
@@ -182,483 +180,103 @@ def get_hub_section(schema, questionnaire_store, section_id):
...
-@section_routing_path_required
-def get_block(
-    routing_path,
-    schema,
-    questionnaire_store,
-    block_id,
-    list_name=None,
-    list_item_id=None,
-):
+def get_block(schema, questionnaire_store, block_id, list_name=None, list_item_id=None):
```
MebinAbraham: I think we should rename this from `get_block` since it now handles `POST` as well?

MebinAbraham (changes requested):
Looking good.

app/views/questionnaire.py
bitdivision: `handle_block` perhaps?

app/contexts/blocks/list_collector.py
ajmaddaford: The was copied from the original code in `view_context.py`, and is done in more places. It seems reasonable to change it in this pull request. I've now updated all references to gettext in the code with the full method names.

app/handlers/blocks/block.py
ajmaddaford: It's used in the `set_started_at_metadata` method.

app/views/questionnaire.py
ajmaddaford: Please don't comment on every line that would need to be changed, it makes for a very noisy pull request!

app/handlers/block_factory.py
ajmaddaford: Yeah, I'll take a look at this.

app/views/questionnaire.py
ajmaddaford: I think I will remove the hub endpoint and render the hub at `/` rather than redirect.

ajmaddaford: I've now refactored the endpoints so the hub is rendered at `/questionnaire`

app/views/questionnaire.py
ajmaddaford: After a quick search it seems the general convention is to name the route without the get or post. I will change this from `get_block` to `block` and `get_relationship` to `relationship`.

app/contexts/blocks/list_collector.py
MebinAbraham: `gettext` is commonly aliased as `_()` according to docs, so I think it was fine as is. But can see why some won't like it.

app/handlers/block_factory.py
ajmaddaford: I've renamed `get_handler` to `get_block_handler` and moved the functionality into `block.py`.

app/handlers/block_factory.py
```diff
@@ -0,0 +1,48 @@
...
+    def get_handler(schema, location, questionnaire_store, language):
+        block = schema.get_block(location.block_id)
+        if not block:
+            raise InvalidLocationException(
+                f'block id {location.block_id} is not valid for this schema'
+            )
+
+        block_type = block['type']
+        block_class = BLOCK_MAPPINGS.get(block_type)
+        if not block_class:
```
MebinAbraham: Do we need this?. If it is a valid block, then it should be a valid block type since it goes through validator?

bitdivision: It will be useful when developing and adding new block types, I would have thought it makes sense to keep it.

ajmaddaford: I think we should have this as a block type can be added without an implementation in runner. I will change it to an `ValueError` rather than `InvalidLocationException` as it shouldn't happen.

app/handlers/block_factory.py
ajmaddaford: This is now just a method.

app/views/contexts/summary_context.py
```diff
@@ -0,0 +1,228 @@
...
+    answer_store: AnswerStore,
+    list_store: ListStore,
+    metadata: Mapping,
+    sections: List[Mapping] = None,
+) -> List:
+    """
+    Build questionnaire summary context containing metadata and content from the answers of the questionnaire.
+    :param schema: schema of the current questionnaire
+    :param answer_store: all of the answers to the questionnaire
+    :param metadata: all of the metadata
```
bitdivision: These are out of date. I suggest we just remove this docstring.

app/views/handlers/primary_person_list_collector.py
```diff
@@ -0,0 +1,43 @@
+from flask import url_for
+
+from app.views.handlers.question import Question
+from app.views.contexts.question import build_question_context
+
+
+class PrimaryPersonListCollector(Question):
+    is_adding = False
```
bitdivision: Should these be instance attributes rather than class attributes? 

I realise it doesn't matter for the current use case, but if these ever get instantiated elsewhere it could presumably cause a problem.

bitdivision (approved):
I think this looks really good, questionnaire.py looks so much better. 

I've raised a couple of minor points, but no real objections

MebinAbraham (approved):
👍

bitdivision (approved):

app/contexts/blocks/list_collector.py
JamesGardiner: @MebinAbraham - this could cause issues in a number of cases, for instance if you wanted to ignore a return value, which is the most common pattern for using `_`

JamesGardiner (approved):

PR #2238, opened 2019-07-30
Adding list collector summary context

JamesGardiner:
List collector summary rows were being built in code, rather than from the survey schema. This change implements a new list collector key in a survey schema, 'summary', that contains information required for runner to identify what list items to display and whether any transforms are required (i.e. list concatenation).
Changes to eq-schema-validator can be found here:
https://github.com/ONSdigital/eq-schema-validator/pull/141

### How to review 
Run the test suite, ensure it passes.
Using a survey that contains a list collector with a summary, fill in the survey noting that the correct list items should be displayed during the list collection process. Primary persons should be correctly identified and displayed with a ` (You)` appended to the list item.


app/views/contexts/list_collector.py
```diff
@@ -2,57 +2,64 @@
...
-def build_list_collector_context(rendered_block, form, list_store, answer_store):
-    question_context = build_question_context(rendered_block, form)
-
+def build_list_items_summary_context(rendered_block, answer_store, list_store):
     list_name = rendered_block['for_list']
     list_item_ids = list_store[list_name].items
 
+    try:
+        item_titles = rendered_block.get('summary', {})['item_title']
+        item_titles = item_titles.strip('[]').replace("'", '').split(',')
```
bitdivision: Why does the `KeyError` cover this line?

bitdivision: I also don't understand what this line is for? Is it escaping the title?

JamesGardiner: Rendered blocks contain the titles as strings in the following format:
`"['a a', 'b b', 'c c']"`

ajmaddaford (changes requested):
I don't think the approach here is quite right.

- The logic for repeating over all list items shouldn't be in the placeholder transform code. The default behaviour of the placeholder answer lookups should be for a single list item (needed for repeating sections), rather than for all list items. We should repeat over the list and call the relevant placeholder transform for each list item id. 
 
- The list of item titles that are generated from the summary definition should also contain the associated list item id for each title. This would avoid the need to zip the titles with the list item ids. 

Currently the whole block is passed to the placeholder renderer. We could change this to call the placeholder renderer separately for the `summary` and `question` elements, allowing us to implement different behaviour for the summary.

app/views/handlers/list_collector.py
```diff
@@ -38,3 +41,29 @@ def handle_post(self, form):
...
+
+        try:
+            summary = transformed_block.pop('summary')
+        except KeyError:
+            return placeholder_renderer.render(transformed_block)
+
+        return {
+            **placeholder_renderer.render(transformed_block),
+            **{'summary': summary},
+        }
```
ajmaddaford: I think we can change the render block behaviour for all blocks to just render the question rather than the whole block. Then the list collector wouldn't need to override the render_block method.

app/questionnaire/placeholder_parser.py
```diff
@@ -8,13 +8,14 @@ class PlaceholderParser:
     final values
     """
 
-    def __init__(self, language, answer_store=None, metadata=None):
+    def __init__(self, language, list_item_id=None, answer_store=None, metadata=None):
```
ajmaddaford: Is there a reason to add the `list_item_id` param here rather than as the last param?

app/questionnaire/placeholder_parser.py
```diff
@@ -8,13 +8,14 @@ class PlaceholderParser:
...
+        self.list_item_id = list_item_id
```
ajmaddaford: `self._list_item_id` rather than `self.list_item_id`?

ajmaddaford (dismissed):
Looks good 🙂

app/questionnaire/placeholder_parser.py
JamesGardiner: Consistent with @MebinAbraham's upcoming pr for repeating sections

app/views/contexts/list_collector.py
```diff
@@ -1,58 +1,70 @@
...
-                        'questionnaire.block',
-                        list_name=list_name,
-                        block_id=rendered_block['remove_block']['id'],
-                        list_item_id=list_item_id,
-                    ),
-                    'primary_person': is_primary,
-                }
-            )
 
+def build_list_collector_context(
```
iwootten, line 62: This and the other methods created here and within parser should probably have a consistent order to their args. Given answer_store is the main store over lists, it probably should precede it and data models tend to be the last arguments we pass to methods elsewhere.

iwootten (changes requested):

app/questionnaire/placeholder_renderer.py
```diff
@@ -38,8 +38,9 @@ class PlaceholderRenderer:
     strings
     """
 
-    def __init__(self, language, answer_store=None, metadata=None):
+    def __init__(self, language, list_item_id=None, answer_store=None, metadata=None):
```
iwootten: I know you've said the argument order matches @MebinAbraham's upcoming PR, but that's a WIP.  I think the order of args here should match that of placeholder_parser.

tests/app/views/contexts/test_list_collector_context.py
```diff
@@ -1,45 +0,0 @@
-from app.views.contexts.list_collector import _generate_list_item_title
```
iwootten, line 1: I realise the views are difficult to test, but there doesn't appear to be existing unit tests that cover the list collector context class?

iwootten (changes requested):

app/views/handlers/question.py
```diff
@@ -18,3 +25,36 @@ def handle_post(self, form):
         self._update_section_completeness()
 
         self.questionnaire_store_updater.save()
+
+    @property
+    def questionnaire_store_updater(self):
+        if not self._questionnaire_store_updater:
```
bitdivision, line 33: :+1:

tests/app/views/contexts/conftest.py
```diff
@@ -0,0 +1,53 @@
...
+    with open(path, 'r') as fp:
+        return json.load(fp)
+
+
+@pytest.fixture
+def answer_store():
+    return AnswerStore(
+        [
+            {'answer_id': 'first-name', 'value': 'a', 'list_item_id': 'PlwgoG'},
+            {'answer_id': 'last-name', 'value': 'a', 'list_item_id': 'PlwgoG'},
```
iwootten: Can we make these actual first and last names?

JamesGardiner: Yep :smile:

tests/app/views/contexts/conftest.py
```diff
@@ -0,0 +1,53 @@
...
+@pytest.fixture
+def list_store():
+    return ListStore([{'items': ['PlwgoG', 'UHPLbX', 'FnoDHP'], 'name': 'people'}])
+
+
+@pytest.fixture
+def form():
+    mock_form = MagicMock(
+        spec=QuestionnaireForm,
+        data={'first-name': 'a', 'last-name': 'a'},
```
iwootten: Again, would be good if these were names.

tests/app/views/contexts/test_list_collector.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+def test_build_list_collector_context(
+    rendered_block, answer_store, list_store, form, app
+):
+    context = list_collector.build_list_collector_context(
+        rendered_block, answer_store, list_store, 'en', form
+    )
+
+    assert {'list_items', 'add_link'} <= {*context.keys()}
```
iwootten: As we spoke about, this is doing a set comparison where the keys() can be a superset. Probably need to be checking they're the same?

JamesGardiner: The comparison operator here checks that the left hand set is a subset of the righthand: https://docs.python.org/3.7/library/stdtypes.html#frozenset.issubset

I'm happy with that :+1:

tests/app/views/contexts/conftest.py
```diff
@@ -0,0 +1,53 @@
...
+from app.data_model.answer import Answer
+from app.data_model.answer_store import AnswerStore
+from app.data_model.list_store import ListStore
+from app.forms.questionnaire_form import QuestionnaireForm
+from app.setup import create_app
+
+
+@pytest.fixture
+def rendered_block():
+    path = Path(__file__).parent.joinpath('test_list_collector_rendered_block.json')
```
bitdivision: Loading a big block of JSON like this is something we've tried to avoid in the unit tests where possible. 

Is there a way we can reduce this and keep it in python as has been done elsewhere?

app/routes/questionnaire.py
```diff
@@ -211,36 +211,16 @@ def block(schema, questionnaire_store, block_id, list_name=None, list_item_id=No
     except InvalidLocationException:
         return redirect(url_for('.get_questionnaire'))
 
-    if block_handler.rendered_block['type'] == 'RelationshipCollector':
+    if schema.get_block(block_id)['type'] == 'RelationshipCollector':
```
bitdivision: Is there a reason this doesn't use `block_handler.block['type']`?

tests/app/views/contexts/test_list_collector.py
```diff
@@ -0,0 +1,58 @@
...
+def test_assert_primary_person_string_appended(
+    rendered_block, answer_store, list_store, app
+):
+    list_store['people'].primary_person = 'PlwgoG'
+    list_item_context = list_collector.build_list_items_summary_context(
+        rendered_block, answer_store, list_store, 'en'
+    )
+
+    assert list_item_context[0]['primary_person'] == True
+    assert 'Toni Morrison (You)' == list_item_context[0]['item_title']
```
bitdivision, line 255: Should this test also check that other list items do not have `(You)` appended to them?

tests/app/views/contexts/test_list_collector.py
iwootten: I think that a clearer approach would be to assert the full structure returned as context is as expected, as we're doing with test_hub_context.py.

tests/app/views/contexts/conftest.py
```diff
@@ -0,0 +1,53 @@
...
+    with open(path, 'r') as fp:
+        return json.load(fp)
+
+
+@pytest.fixture
+def answer_store():
+    return AnswerStore(
+        [
+            {'answer_id': 'first-name', 'value': 'Toni', 'list_item_id': 'PlwgoG'},
```
bitdivision: Is it worth adding another person here? We're testing a summary of only one person at the moment I think.

bitdivision (changes requested):

app/routes/questionnaire.py
JamesGardiner: No - which one do you prefer?

JamesGardiner: No, there isn't, and your suggestion is easier to read

tests/app/views/contexts/conftest.py
JamesGardiner: Can do - are you suggesting adding it as a dict to conftest?

bitdivision: Potentially, I guess the ideal would be if we could make a more minimal example that could be represented as a dict.

@iwootten Did make a good point in that when reading the test, you normally need to look at the fixtures as well. So unless there are multiple test files sharing the fixtures it might be worth moving them into the test file for the moment. That way we don't end up with a load of apparently common functionality which isn't actually common.

app/routes/questionnaire.py
bitdivision: :+1: 

I think it just makes sense to use the block_handler wherever possible since we already have it available.

iwootten (approved):

app/views/handlers/question.py
```diff
@@ -1,8 +1,15 @@
+from app.questionnaire.placeholder_renderer import PlaceholderRenderer
+from app.questionnaire.questionnaire_store_updater import QuestionnaireStoreUpdater
+from app.questionnaire.schema_utils import transform_variants
 from app.views.handlers.block import BlockHandler
 from app.views.contexts.question import build_question_context
 
 
 class Question(BlockHandler):
+    @property
```
bitdivision: Should this (and others) be a cached property?

bitdivision (approved):
Looks good. One minor thing, but I'm not going to block the PR on it.

app/views/handlers/question.py
JamesGardiner: Good point - it's an easy optimisation

bitdivision: I wasn't expecting this to be applied to questionnaire_store_updater too, since it's already cached through that if statement, but I guess it doesn't really make much difference.

PR #2240, opened 2019-07-31
Use python package for translations

bitdivision:
This removes the requirement for using docker for running translations. It updates the dockerfile to use the build.sh script and adds the python package for eq_translations.

It also removes the docker part of template extraction and replaces it with a call to the python package.

Should be reviewed in tandem with https://github.com/ONSdigital/eq-translations/pull/38

### How to review 
Test that translations work correctly, both outside and inside of docker.

Easiest way to do this is to call the scripts manually, and to run the docker-compose build process, which will call build.sh and build all the schemas.

### Checklist


JamesGardiner (dismissed):

JamesGardiner (commented):
Having issues running the translations from docker

iwootten:
Is it worth updating the template extraction script to use the package in the same way, so we don't use docker from scripts at all?

bitdivision:
@iwootten - Yes absolutely. I'll do that now.

@JamesGardiner Any details?

bitdivision:
Template extraction now handled through the package @iwootten .

I've also had to pin redis to an older version, it looks like there is some churn in that package, which is causing our tests to fail. Hopefully that will settle down soon.

JamesGardiner (approved):
Approving based on the translations working as specified here however, there are issues when trying to run the commands from inside a docker container on some systems (failed on Manjaro Linux). This won't be an issue as we don't intend to use the scripts this way.

bitdivision:
@JamesGardiner Cheers. 

As discussed, I'm not going to worry too much about the scripts not working when run through the runtime docker containers. The scripts should really only ever be run during the build of the container (first stage). If we want to provide a docker environment for running scripts then we could do that based off the first stage, but I don't think there's a real need for that at the moment.

ajmaddaford (approved):

MebinAbraham (changes requested):
Am I missing something obvious?

```
ImportError: cannot import name 'extract_template' from 'eq_translations.entrypoints' (/Users/abrahm/.local/share/virtualenvs/eq-survey-runner-7OUsly8u/src/eq-translations/eq_translations/entrypoints.py)
make: *** [translation-templates] Error 1
```

MebinAbraham (approved):

PR #2243, opened 2019-08-01
Updating the 'Thank You' page

berroar:
Updated the 'Thank You' page to match the prototype. See [sketch](https://sketch.cloud/s/jP21z/a/G0npyb). Note: The feedback link is not required.

### How to review 
Check that the new Thank You page matches the prototype.


berroar:
[Post submission page (S)](https://trello.com/c/ad8MFrRw/3019-post-submission-page-s)

bitdivision (commented):
Looks good :+1:

You will need to also update the static translations to include the new strings.

bitdivision:
I would also be surprised if a few functional tests didn't fail now.

berroar:
> I would also be surprised if a few functional tests didn't fail now.

Have now removed the only functional test that was failing.

tests/integration/routes/test_view_submission.py
```diff
@@ -1,363 +0,0 @@
-import simplejson as json
```
bitdivision: If we're removing the test for this, then we should probably remove the code that handles view submission as well.

bitdivision: Just spoke to @ajmaddaford about this. Suggestion is to leave the view submission feature in runner. So in that case, we should leave the tests here as well.

You will probably need to just remove the references to the thank you page and go directly to the view submission URL.

berroar: Have put this file back and removed references to the thank you page.

JamesGardiner (approved):

iwootten (approved):

tests/integration/routes/test_view_submission.py
```diff
@@ -29,10 +29,6 @@ def setUp(self):
 
     def test_view_submission(self):
```
iwootten: I've just noticed that this test is no longer asserting anything, which suggests it should just be completely removed.

tests/integration/routes/test_view_submission.py
```diff
@@ -88,16 +84,9 @@ def test_try_view_submission_when_not_available(self):
         # Submit answers
         self.post(action=None)
 
-        # check we're on the thank you page and view submission link is not available
```
iwootten, line 91: This test method name no longer matches what is actually being tested. It should end on an assertion. I think the only thing that needed to be removed was the "self.assertNotInBody"?

PR #2246, opened 2019-08-01
Runner now accepts a list of option Ids for routing

SamGodwin2:
Added the ability to pass in an array of answer values in order to better support checkbox answers.

### How to review 
Tests pass and schemas built using the new functionality work correctly.

### DIFF
--- a/app/questionnaire/rules.py
+++ b/app/questionnaire/rules.py
@@ -33,7 +33,7 @@ def evaluate_rule(when, answer_value):
     :return (bool): The result of the evaluation
     """
 
-    match_value = when['value'] if 'value' in when else None
+    match_value = when.get('value', when.get('values'))
 
     condition = when['condition']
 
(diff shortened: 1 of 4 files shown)


ajmaddaford:
@SamGodwin2 This was added to v3 in https://github.com/ONSdigital/eq-survey-runner/pull/2041

SamGodwin2:
@ajmaddaford Thanks for the heads up. After talking about it though it looks like what came in in that pr doesn't cover all our use cases. So for now we'd like to go with these rules in v2 then switch over when it comes time to move over to v3 later on.

SamGodwin2:
[Spike - Change Runner to accept list of option ID (1 day)](https://trello.com/c/Z7TnIevV/926-spike-change-runner-to-accept-list-of-option-id-1-day)

app/assets/js/app/main.js.tmp
```diff
@@ -0,0 +1,5 @@
+import './modules/inpagelink'
+import './modules/help'
+import './modules/household'
+import './modules/totaliser'
+import './modules/print.js'
```
pricea1: This file seems to have been added in error

app/questionnaire/rules.py
```diff
@@ -67,14 +67,42 @@ def evaluate_condition(condition, answer_value, match_value):
     comparison_operators = {
         'equals': lambda answer_value, match_value: answer_value == match_value,
         'not equals': lambda answer_value, match_value: answer_value != match_value,
-        'contains': lambda answer_value, match_value: isinstance(answer_value, list) and match_value in answer_value,
-        'not contains': lambda answer_value, match_value: isinstance(answer_value, list) and match_value not in answer_value,
+        'contains':
+            lambda answer_value, match_value: isinstance(answer_value, list)
+            and match_value in answer_value,
+        'all of':
```
pricea1: When launched "All of" fails if other items selected too.

e.g. 
	"condition": "all of",
	"values": ["Box1", "Box2", "Box3"]
passes if Box1, Box2, Box3 selected but
fails if Box1, Box2, Box3, Box4 selected

data/en/test_routing_values_array.json
```diff
@@ -0,0 +1,279 @@
...
+                                    "condition": "not all of",
+                                    "values": ["Box1", "Box2", "Box3"]
+                                }]
+                            }
+                        }, {
+                            "goto": {
+                                "block": "block-six",
+                                "when": [{
+                                    "id": "answer-four",
+                                    "condition": "not any of",
```
pricea1: Is this being tested? Is this rule only invoked if previous "not all of" rule is false i.e. if Box1, Box2, Box3 all selected.

pricea1 (changes requested):
A couple of logic queries otherwise looks good.

data/en/test_routing_values_array.json
SamGodwin2: Good spot, its being testing on block 6 I forgot to delete it from here when I moved it.

bitdivision:
I don't feel like I can request changes here, but it seems like this is going to be implementing features in v2 and author which we know will be removed when v3 is merged.

I think it would be better to do any translation required within publisher and use the v3 rules for now.

SamGodwin2:
@bitdivision You do make fair point but while I _think_ it is technically possible to do the translation in publisher the amount of work required to support the range of routing possibilities would be far, far larger. Isn't the routing on the shortlist of things to be changed in v3 anyway so even the ones in there now may not be valid when it comes to merging them both.

samiwel:
It will be a some time before we are able to fully support all v3 features so for now I suggest we strive for consistency in our naming of the comparators (i.e. "contains all", "contains any") and add the inverse comparators not available in v3 (i.e. "not contains all", "not contains any") and implement on the v2 branch.

The additional comparators may need to be implemented on v3 at a later date. But doing it in this order will allow us to deliver on requirements for complex routing conditions on upcoming business surveys and hopefully make the migration to get routing working on v3 less painful.

app/questionnaire/rules.py
```diff
@@ -67,14 +67,42 @@ def evaluate_condition(condition, answer_value, match_value):
...
         'equals': lambda answer_value, match_value: answer_value == match_value,
         'not equals': lambda answer_value, match_value: answer_value != match_value,
-        'contains': lambda answer_value, match_value: isinstance(answer_value, list) and match_value in answer_value,
-        'not contains': lambda answer_value, match_value: isinstance(answer_value, list) and match_value not in answer_value,
+        'contains':
+            lambda answer_value, match_value: isinstance(answer_value, list)
+            and match_value in answer_value,
+        'contains all':
+            lambda answer_value, match_value: isinstance(answer_value, list)
+            and isinstance(match_value, list)
```
bitdivision, line 75: Should types be checked in here? This seems like a check that  should be done within validator.

tests/app/questionnaire/test_rules.py
```diff
@@ -175,6 +176,40 @@ def test_evaluate_rule_less_than_with_number(self):
         self.assertFalse(evaluate_rule(when, 5))
         self.assertFalse(evaluate_rule(when, 7))
 
+    def test_evaluate_rule_checkbox_contains_all(self):
```
bitdivision, line 179: pytest parameterize is a useful tool on tests like these: https://docs.pytest.org/en/latest/parametrize.html

pricea1 (approved):
LGTM

tests/app/questionnaire/test_rules.py
SamGodwin2: Would be a nice simplification but unfortunately didn't have any luck in getting it set up and working so decided to leave as is for now.

app/questionnaire/rules.py
SamGodwin2: They were being checked in here in the existing implementation so I mostly just decided to extend that, also feels like checking them here is simpler that in validator?

samiwel (approved):
I think the build needs kicking off again, but otherwise LGTM.

PR #2251, opened 2019-08-02
Add Primary Person Variant to Relationships

iwootten:

Currently during relationship definition, we have no way of displaying question variants for a primary person. 

### How to review 
This adds a new test_relationship_primary schema which includes a question variant for the primary person. To support display of the variant a new "when" comparison needs to be added to allow comparisons against location.

This needs to be run with: https://github.com/ONSdigital/eq-schema-validator/pull/143

### DIFF
--- a/app/questionnaire/rules.py
+++ b/app/questionnaire/rules.py
@@ -161,12 +161,17 @@ def _is_answer_on_path(schema, answer, routing_path):
     return location in routing_path
 
 
-def _get_comparison_id_value(when_rule, answer_store, schema):
+def _get_comparison_id_value(when_rule, answer_store, schema, current_location):
     """
-        Gets the value of an answer specified as an operand in a comparator ( i.e right hand argument ,
-        or rhs in if lhs>rhs ) In a when clause
+        Gets the value of a comparison id specified as an operand in a comparator
     """
-    answer_id = when_rule['comparison_id']
+    if when_rule['comparison']['source'] == 'location':
+        try:
+            return getattr(current_location, when_rule['comparison']['id'])
+        except AttributeError:
+            return None
+
+    answer_id = when_rule['comparison']['id']
 
     return get_answer_store_value(answer_id, answer_store, schema)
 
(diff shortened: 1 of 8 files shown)


scripts/test_schemas.sh
```diff
@@ -6,7 +6,7 @@ if [ "$1" == "--local" ] || [ "$2" == "--local" ]; then
 fi
 
 if [ "$run_docker" == true ]; then
-    branch=v3
+    branch=eq-3067-add-primary-comparison
```
iwootten: Update the branch once validator PR is merged

iwootten:
[Add primary person to relationships (M)](https://trello.com/c/B20tHhR5/3067-add-primary-person-to-relationships-m)

app/questionnaire/rules.py
```diff
@@ -210,15 +215,24 @@ def _get_when_rule_value(
     elif 'meta' in when_rule:
         value = get_metadata_value(metadata, when_rule['meta'])
     elif 'list' in when_rule:
-        value = get_list_count(list_store, when_rule['list'])
+        if 'id_selector' in when_rule:
```
JamesGardiner:
```Python
try:
    value = getattr(list_store.get(when_rule['list']), when_rule['id_selector'])
except KeyError:
    value = get_list_count(list_store, when_rule['list'])
```

JamesGardiner (changes requested):
LGTM - one minor change request

app/questionnaire/rules.py
iwootten: Neither one of these is an exception, where this change implies one of them is. I think the original is more readable.

JamesGardiner: I disagree - the suggested change is an idiomatic approach to handling mapping access control flow.

iwootten: I'm sorry but I still maintain that my change is more readable and consistent with what is already there. The project hasn't ever prescribed the EAFP approach as far as I'm aware, so I'm not sure why this should block the PR?

bitdivision: I don't really see much of a difference in readability, but the if pattern does keep this consistent with the rest of runner. 

We should probably avoid having this argument on all PRs until we decide on an approach with the whole team.

bitdivision (approved):
Works well, and the changes look sensible.

JamesGardiner (approved):

app/questionnaire/rules.py
```diff
@@ -214,11 +214,10 @@ def _get_when_rule_value(
         )
     elif 'meta' in when_rule:
         value = get_metadata_value(metadata, when_rule['meta'])
+    elif 'id_selector' in when_rule:
```
bitdivision, line 217: :+1:

PR #2252, opened 2019-08-02
Revert change to application-version print

bitdivision:
I broke the version used in the status endpoint as part of this PR: https://github.com/ONSdigital/eq-survey-runner/pull/2240/files

There was an attempted fix in https://github.com/ONSdigital/eq-survey-runner/pull/2250 but it appears to behave differently locally.

This PR reverts the original change and goes back to using `printf`


MebinAbraham (dismissed):
`number_of_commits += 1` 😂

ajmaddaford (dismissed):

bitdivision:
The issue is of course that this line was never previously called in the dockerfile, and `.git` is in the dockerignore.

We can either unignore .git and then delete it after the build stage, or can move that line into the travis build.

I've gone for moving it into the travis build for the moment. Open to suggestions.

MebinAbraham (approved):

ajmaddaford (approved):

PR #2264, opened 2019-08-08
Redirect when required sections are incomplete

iwootten:

It needs to be possible to define sections which must be completed prior to the hub being accessible.

### How to review 

Check if you agree with the approach and the tests cover all new behaviour.


iwootten:
[Disable hub until sections are completed (M)](https://trello.com/c/7nzx65CH/3182-disable-hub-until-sections-are-completed-m)

app/routes/questionnaire.py
```diff
@@ -110,8 +110,13 @@ def get_questionnaire(schema, questionnaire_store):
         progress_store=questionnaire_store.progress_store,
         list_store=questionnaire_store.list_store,
     )
+    completed_section_ids = questionnaire_store.progress_store.completed_section_ids
+    hub_preceding_sections_complete = all(
```
bitdivision: Can we change to `hub_required_sections_complete`?

tests/integration/questionnaire/test_questionnaire_hub.py
```diff
@@ -120,3 +120,21 @@ def test_hub_section_url_when_hub_enabled_and_section_complete(self):
 
         # Then I should be redirected to the last completed question in the section
         self.assertEqualUrl('/questionnaire/accommodation-details-summary/')
+
+    def test_hub_inaccessible_if_sections_required_and_incomplete(self):
+        self.launchSurvey('test_hub_complete_sections')
```
bitdivision, line 125: In the spirit of reducing integration tests, can we make this a functional test? Or is this needed for coverage?

bitdivision (changes requested):

tests/integration/questionnaire/test_questionnaire_hub.py
iwootten: Coverage doesn't drop when I remove these, but the switch won't be tested. I'll add the alternative to the functional tests.

tests/functional/spec/features/hub_and_spoke/hub_and_spoke.spec.js
```diff
@@ -229,7 +229,27 @@ describe('Feature: Hub and Spoke', function () {
         .click(HubPage.submit())
         .getUrl().should.eventually.contain('thank-you');
     });
-
   });
 
+  describe('Given a user opens a schema with required sections', function () {
+    it('It should not show first of all', function () {
```
LJBabbage: Very Minor, but I didn't immediately know what 'it' referred to

app/questionnaire/questionnaire_schema.py
```diff
@@ -25,6 +25,9 @@ def __init__(self, questionnaire_json, language_code=DEFAULT_LANGUAGE_CODE):
     def is_hub_enabled(self):
         return self.json.get('hub', {}).get('enabled')
 
+    def get_hub_required_section_ids(self):
```
LJBabbage: This makes me think this is required for the hub and spoke part, rather than is needed before it gets there

LJBabbage (approved):
Looks good and works as expected. 2 very minor comments not worth blocking approval.

app/questionnaire/questionnaire_schema.py
```diff
@@ -25,6 +25,9 @@ def __init__(self, questionnaire_json, language_code=DEFAULT_LANGUAGE_CODE):
     def is_hub_enabled(self):
         return self.json.get('hub', {}).get('enabled')
 
+    def get_section_ids_required_for_hub(self):
+        return self.json.get('hub', {}).get('required_completed_sections', [])
```
JamesGardiner, line 29: I think that as the second key is within the hub we can be a bit self-referential and go with `requires_completed_sections` - it reads better to me given the tense change

app/routes/questionnaire.py
```diff
@@ -110,8 +110,13 @@ def get_questionnaire(schema, questionnaire_store):
         progress_store=questionnaire_store.progress_store,
         list_store=questionnaire_store.list_store,
     )
+    completed_section_ids = questionnaire_store.progress_store.completed_section_ids
+    hub_required_sections_complete = all(
```
JamesGardiner: Our other boolean vars usually begin with `is_` - in this case it could be an `are_` - `are_hub_required_sections_complete`

scripts/test_schemas.sh
```diff
@@ -6,7 +6,7 @@ if [ "$1" == "--local" ] || [ "$2" == "--local" ]; then
 fi
 
 if [ "$run_docker" == true ]; then
-    branch=v3
+    branch=eq-3182-disable-hub-until-sections-complete
```
JamesGardiner: Just a reminder to switch this back before merging

JamesGardiner (changes requested):

app/questionnaire/questionnaire_schema.py
iwootten: I think this the current name is more in line with the rest of the accessors in this file. 'requires' could potentially be construed as being a boolean indicating if it needs completed sections.

JamesGardiner (approved):

bitdivision (approved):

PR #2276, opened 2019-08-13
Revisit relationships when list changes

JamesGardiner:
When a list changes after relationships for the list items have been established, the relationships need to be evaluated again. There are three scenarios covered by this PR:

1) A list item is deleted:
    - All relationships for that list item should be deleted. Users should not be asked the relationship questions again.
2) A list item is added:
    - A user should be taken through all of the relationships questions again. Existing questions will have the previously selected answers already selected. New relationships should not have any answers selected.
3) A list item is added, but removed before submitting the list collector question:
    - No change to the relationship store. User shouldn't be asked relationship questions again.

### How to review 
Look at the code changes. Launch any relationships questionnaire schema. Using the scenarios above check the flow is correct for each. Check the /dump/debug endpoint to monitor updates to the answer store and list store.

### DIFF
--- a/app/data_model/answer_store.py
+++ b/app/data_model/answer_store.py
@@ -117,14 +117,12 @@ class AnswerStore:
             self._is_dirty = True
 
     def remove_all_answers_for_list_item_id(self, list_item_id: str):
-        """Remove all answers associated with a particular list_item_id
+        """Remove all answers associated with a particular list_item_id.
         This method iterates through the entire list of answers.
 
         *Not efficient.*
         """
-
         keys_to_delete = []
-
         for answer in self:
             if answer.list_item_id == list_item_id:
                 keys_to_delete.append((answer.answer_id, answer.list_item_id))
--- a/app/questionnaire/questionnaire_store_updater.py
+++ b/app/questionnaire/questionnaire_store_updater.py
@@ -1,4 +1,6 @@
-from typing import List, Tuple, Optional
+from itertools import combinations
+
+from typing import List, Optional, Tuple, Union
 
 from app.data_model.answer_store import Answer
 from app.data_model.relationship_store import Relationship, RelationshipStore
@@ -40,27 +42,83 @@ class QuestionnaireStoreUpdater:
         relationship_answer_id = self._schema.get_relationship_answer_id_for_block(
             self._current_location.block_id
         )
-
         answer = self._answer_store.get_answer(relationship_answer_id)
-        if answer:
+        self._create_relationship_store_and_update_answer(
+            relationship_answer_id, answer, form_data, list_item_id, to_list_item_id
+        )
+
+    def _create_relationship_store_and_update_answer(
+        self, relationship_answer_id, answer, form_data, list_item_id, to_list_item_id
+    ):
+        try:
             relationship_store = RelationshipStore(answer.value)
-        else:
+        except AttributeError:
             relationship_store = RelationshipStore()
 
         relationship_answer = form_data.get(relationship_answer_id)
-
         relationship = Relationship(list_item_id, to_list_item_id, relationship_answer)
         relationship_store.add_or_update(relationship)
-
         self._answer_store.add_or_update(
             Answer(relationship_answer_id, relationship_store.serialise())
         )
+
+    def remove_completed_relationship_locations_for_list_name(
+        self, list_name: str
+    ) -> None:
+        target_relationship_collectors = self._get_relationship_collectors_by_list_name(
+            list_name
+        )
+        if target_relationship_collectors:
+            for target in target_relationship_collectors:
+                block_id = target['id']
+                section_id = self._schema.get_section_for_block_id(block_id)['id']
+                self.remove_completed_location(Location(section_id, block_id))
+
+    def update_relationship_question_completeness(self, list_name: str) -> None:
+        relationship_collectors = self._get_relationship_collectors_by_list_name(
+            list_name
+        )
+
+        if not relationship_collectors:
+            return None
+
+        list_items = self._list_store.get(list_name).items
+
+        for collector in relationship_collectors:
+            expected_pairs = {pair for pair in combinations(list_items, 2)}
+            relationship_answer_id = self._schema.get_relationship_answer_id_for_block(
+                collector['id']
+            )
+            relationship_answers = self._get_relationships_in_answer_store(
+                relationship_answer_id
+            )
+
+            if relationship_answers:
+                pairs = {
+                    (answer['list_item_id'], answer['to_list_item_id'])
+                    for answer in relationship_answers
+                }
+
+                if expected_pairs == pairs:
+                    section_id = self._schema.get_section_for_block_id(collector['id'])[
+                        'id'
+                    ]
+                    location = Location(section_id, collector['id'])
+                    self.add_completed_location(location)
+
+    def _get_relationship_collectors_by_list_name(self, list_name: str):
+        return self._schema.get_relationship_collectors_by_list_name(list_name)
+
+    def _get_relationships_in_answer_store(self, relationship_answer_id: str):
+        return self._answer_store.get_answer(relationship_answer_id).value
 
     def remove_answers(self, answer_ids: List):
         for answer_id in answer_ids:
             self._answer_store.remove_answer(answer_id)
 
     def add_primary_person(self, list_name):
+        self.remove_completed_relationship_locations_for_list_name(list_name)
+
         if self._list_store[list_name].primary_person:
             return self._list_store[list_name].primary_person
 
@@ -68,10 +126,9 @@ class QuestionnaireStoreUpdater:
 
     def add_list_item_and_answers(self, form, list_name):
         new_list_item_id = self._list_store.add_list_item(list_name)
-
         self._current_location.list_item_id = new_list_item_id
-
         self.update_answers(form)
+        self.remove_completed_relationship_locations_for_list_name(list_name)
 
     def remove_primary_person(self, list_name: str):
         """ Remove the primary person and all of their answers.
@@ -82,7 +139,8 @@ class QuestionnaireStoreUpdater:
             self.remove_list_item_and_answers(list_name, list_item_id)
 
     def remove_list_item_and_answers(self, list_name: str, list_item_id: str):
-        """ Remove answers from the answer store and update the list store to remove it
+        """ Remove answers from the answer store and update the list store to remove it.
+        Any related relationship answers are re-evaluated for completeness.
         """
         self._list_store.delete_list_item(list_name, list_item_id)
 
@@ -90,7 +148,40 @@ class QuestionnaireStoreUpdater:
             list_item_id=list_item_id
         )
 
+        answers = self.get_relationship_answers_for_list_name(list_name)
+        if answers:
+            self.remove_relationship_answers_for_list_item_id(list_item_id, answers)
+            self.update_relationship_question_completeness(list_name)
+
         self._progress_store.remove_progress_for_list_item_id(list_item_id=list_item_id)
+
+    def get_relationship_answers_for_list_name(
+        self, list_name: str
+    ) -> Union[List[Answer], None]:
+        assosciated_relationship_collectors = self._get_relationship_collectors_by_list_name(
+            list_name
+        )
+        if not assosciated_relationship_collectors:
+            return None
+
+        relationship_answer_ids = [
+            self._schema.get_relationship_answer_id_for_block(block['id'])
+            for block in assosciated_relationship_collectors
+        ]
+
+        return self._answer_store.get_answers_by_answer_id(relationship_answer_ids)
+
+    def remove_relationship_answers_for_list_item_id(
+        self, list_item_id: str, answers: List
+    ) -> None:
+        for answer in answers:
+            answers_to_keep = [
+                value
+                for value in answer.value
+                if list_item_id not in {value['to_list_item_id'], value['list_item_id']}
+            ]
+            answer.value = answers_to_keep
+            self._answer_store.add_or_update(answer)
 
     def add_completed_location(self, location: Optional[Location] = None):
         location = location or self._current_location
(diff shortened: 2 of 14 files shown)


app/data_model/answer_store.py
```diff
@@ -117,21 +117,50 @@ def remove_answer(self, answer_id: str, list_item_id: str = None):
...
         """
-
         keys_to_delete = []
-
+        relationship_answers = []
         for answer in self:
             if answer.list_item_id == list_item_id:
                 keys_to_delete.append((answer.answer_id, answer.list_item_id))
 
+            if isinstance(answer.value, list):
```
LJBabbage: Land mine with a sand marked arrow pointing it out.

JamesGardiner:
[Revisit relationships when the list changes (M)](https://trello.com/c/XbFD1fyi/3094-revisit-relationships-when-the-list-changes-m)

app/data_model/answer_store.py
```diff
@@ -117,21 +117,50 @@ def remove_answer(self, answer_id: str, list_item_id: str = None):
...
+        """Remove all answers associated with a particular list_item_id,
+        including relationships.
         This method iterates through the entire list of answers.
 
         *Not efficient.*
         """
-
         keys_to_delete = []
-
+        relationship_answers = []
```
bitdivision: It seems like we should avoid referencing relationships within the answer_store.

bitdivision: As discussed, probably worth moving this out and referencing from questionnaire_store_updater.

app/data_model/answer_store.py
bitdivision: 👀

bitdivision (changes requested):

app/data_model/answer_store.py
JamesGardiner: :see_no_evil:

app/data_model/answer_store.py
JamesGardiner: I agree - will look at moving it today.

MebinAbraham:
Can you merge v3.

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -39,28 +42,81 @@ def is_dirty(self):
...
+        if target_relationship_collectors:
+            for target in target_relationship_collectors:
+                self.remove_completed_location(Location(target['id']))
+
+    def assess_relationship_question_completeness(self, list_name: str) -> None:
+        target_relationship_collectors = self.get_relationship_collectors_by_list_name(
+            list_name
+        )
+
+        if target_relationship_collectors:
```
bitdivision: Worth doing an early return here to avoid the nesting.

app/data_model/answer_store.py
```diff
@@ -117,21 +117,50 @@ def remove_answer(self, answer_id: str, list_item_id: str = None):
...
         for answer in self:
             if answer.list_item_id == list_item_id:
                 keys_to_delete.append((answer.answer_id, answer.list_item_id))
 
+            if isinstance(answer.value, list):
+                # Only relationship answers are stored as lists at present
+                # Should other objects start getting stored here there really should
+                # be some form of type lookup key.
+                relationship_answers.append(answer.answer_id)
```
LJBabbage: Not really necessary as this would mean 2 RelationshipCollector for a list, which isn't needed and might never be needed. However it does future proof the code and did work when I manipulated the JSON. Maybe needs a test if it stays like this

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -81,34 +72,44 @@ def mark_relationship_question_incomplete(self, list_name: str) -> None:
         )
         if target_relationship_collectors:
             for target in target_relationship_collectors:
-                self.remove_completed_location(Location(self._schema.get_section_for_block_id(target['id'])['id'], target['id']))
+                self.remove_completed_location(
```
MebinAbraham: Can we make this a bit clearer?

```python
            for target in target_relationship_collectors:
                block_id = target['id']
                section_id = self._schema.get_section_id_for_block_id(block_id)
                location = Location(section_id, block_id)
                
                self.remove_completed_location(location)
```

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -81,34 +72,44 @@ def mark_relationship_question_incomplete(self, list_name: str) -> None:
...
-                    pairs = [
-                        (answer['list_item_id'], answer['to_list_item_id'])
-                        for answer in relationship_answers
-                    ]
-
-                    if expected_pairs == pairs:
-                        self.add_completed_location(
-                                Location(self._schema.get_section_for_block_id(target['id'])['id'], target['id'])
+        if not target_relationship_collectors:
+            return None
```
MebinAbraham, line 83: Is this needed since the for loop exits on the first loop?

app/data_model/answer_store.py
```diff
@@ -117,21 +117,47 @@ def remove_answer(self, answer_id: str, list_item_id: str = None):
...
+        self, list_item_id, relationship_answers
+    ):
+        for key in relationship_answers:
+            answer = self.get_answer(key)
+            values = answer.value
+            values_to_delete = []
+
+            for value in values:
+                if list_item_id in {value['list_item_id'], value['to_list_item_id']}:
+                    values_to_delete.append(value)
```
MebinAbraham: Maybe a comprehension?

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -36,26 +39,81 @@ def is_dirty(self):
...
         relationship_answer = form_data.get(relationship_answer_id)
-
         relationship = Relationship(list_item_id, to_list_item_id, relationship_answer)
         relationship_store.add_or_update(relationship)
-
         self._answer_store.add_or_update(
             Answer(relationship_answer_id, relationship_store.serialise())
         )
 
+    def mark_relationship_question_incomplete(self, list_name: str) -> None:
```
MebinAbraham: Not sure about the naming. What about something like `incomplete_relationships_for_list`?

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -36,26 +39,81 @@ def is_dirty(self):
...
+        if target_relationship_collectors:
+            for target in target_relationship_collectors:
+                self.remove_completed_location(
+                    Location(
+                        self._schema.get_section_for_block_id(target['id'])['id'],
+                        target['id'],
+                    )
+                )
+
+    def assess_relationship_question_completeness(self, list_name: str) -> None:
```
MebinAbraham: Not sure about `assess`. Maybe `update` or `evaluate`?

MebinAbraham (changes requested):

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -36,26 +39,81 @@ def is_dirty(self):
...
+        target_relationship_collectors = self.get_relationship_collectors_by_list_name(
```
LJBabbage: Not a huge fan of the prefix target, do we need it?

app/questionnaire/questionnaire_store_updater.py
LJBabbage: For me its not marking the relationship question incomplete which makes me feel like a flag, its removing the location if it there are relationships. Possibly remove_relationship_list_completed

LJBabbage (changes requested):

app/questionnaire/questionnaire_store_updater.py
JamesGardiner: because there may be no relationship collectors for a given list name

app/questionnaire/questionnaire_store_updater.py
JamesGardiner: Good point both - I'll update the name

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -36,26 +38,83 @@ def is_dirty(self):
     def update_answers(self, form):
         self._update_questionnaire_store_with_form_data(form.data)
 
+    def get_relationship_collectors_by_list_name(self, list_name: str):
```
bitdivision: Does this need to be a method? And if it does should it be private?

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -36,26 +38,83 @@ def is_dirty(self):
...
+                relationship_answer_id
+            )
+
+            if relationship_answers:
+                pairs = [
+                    (answer['list_item_id'], answer['to_list_item_id'])
+                    for answer in relationship_answers
+                ]
+
+                if expected_pairs == pairs:
```
bitdivision, line 102: Should these be sets? Doesn't order matter in the current comparison?

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -36,26 +38,83 @@ def is_dirty(self):
...
+                section_id = self._schema.get_section_for_block_id(block_id)['id']
+                self.remove_completed_location(Location(section_id, block_id))
+
+    def update_relationship_question_completeness(self, list_name: str) -> None:
+        relationship_collectors = self.get_relationship_collectors_by_list_name(
+            list_name
+        )
+
+        if not relationship_collectors:
+            return None
```
bitdivision, line 83:
```suggestion
            return
```

tests/functional/spec/relationships.spec.js
```diff
@@ -191,6 +192,72 @@ describe('Relationships', function() {
     });
   });
 
+  it('Given I am completing the survey, When I have added one or more household members after answering the relationships question, Then I will be asked for the original household members relationships to all new members', function() {
```
bitdivision: Not specific to this PR, but the given should probably be in the initial describe call. e.g. `describe('Given I am completing the test_relationships survey')`

bitdivision: You could make this section a bit clearer by nesting these tests under another describe saying that they've added three people.

bitdivision (changes requested):
I don't see a functional test for the third scenario in your PR description? Probably worth adding.

app/questionnaire/questionnaire_store_updater.py
JamesGardiner: I spoke to @ajmaddaford about this. My personal preference is not to specify the None, but the decision was got go with the explicit version

bitdivision: That seems strange given that this function doesn't normally return anything? I can see the value in being explicit if this normally returned a list or something, but here it is just a standard early return.

Not a big issue though. I'm happy to approve as is.

app/questionnaire/questionnaire_store_updater.py
JamesGardiner: It doesn't need to be, but there are other examples of this in this module - see update_section_status, which simply passes along the args to the corresponding progress store method.

app/questionnaire/questionnaire_store_updater.py
bitdivision: Looking through the code, we do normally seem to do that. So :+1:

app/questionnaire/questionnaire_store_updater.py
JamesGardiner: Good catch

app/questionnaire/questionnaire_store_updater.py
bitdivision: Can we make it private then? This should always be reference through the schema, not questionnaire_store_updater.

bitdivision (approved):
:+1:

ajmaddaford:
@JamesGardiner I've just tried running this and can't test it as the list summary is missing from the list collector. It looks like it's not defined in the schema.

ajmaddaford (dismissed):
I think I broke it 😬 

I added a primary and two others, removed primary, re-added primary, next takes me to summary i.e. skipping required primary person relationships. Clicking previous from summary takes me back through the relationship screens with missing relationships for the primary person.

JamesGardiner:
Just tested this locally following the steps outlined and it worked... i.e. asked me for the relationships as expected

JamesGardiner:
Ok, I think the missing step is to add the relationships before removing and re-adding primary person. The additional step has allowed me to recreate the scenario outlined

app/views/handlers/list_add_question.py
```diff
@@ -12,5 +12,8 @@ def handle_post(self, form):
             form, self.parent_block['for_list']
         )
         self.questionnaire_store_updater.update_answers(form)
+        self._questionnaire_store_updater.remove_completed_relationship_locations_for_list_name(
```
ajmaddaford: This should be done as part of the `add_list_item_and_answers` method as:
- This should always be done when adding list items.
- We should keep relationship functionality in questionnaire store updater.

LJBabbage (approved):

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -40,38 +42,96 @@ def update_relationship_answer(self, form_data, list_item_id, to_list_item_id):
...
+            if relationship_answers:
+                pairs = {
+                    (answer['list_item_id'], answer['to_list_item_id'])
+                    for answer in relationship_answers
+                }
+
+                if expected_pairs == pairs:
+                    self.add_completed_location(
+                        Location(
+                            self._schema.get_section_for_block_id(collector['id'])[
```
MebinAbraham: Nit pick, not going to block but would be nice to clean this up a bit.

```py
                if expected_pairs == pairs:
                    section_id = self._schema.get_section_id_for_block_id(collector['id'])
                    location = Location(section_id, collector['id'])

                    self.add_completed_location(location)
```

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -40,38 +42,96 @@ def update_relationship_answer(self, form_data, list_item_id, to_list_item_id):
...
+                                'id'
+                            ],
+                            collector['id'],
+                        )
+                    )
+
+    def _get_relationship_collectors_by_list_name(self, list_name: str):
+        return self._schema.get_relationship_collectors_by_list_name(list_name)
+
+    def _check_relationships_in_answer_store(self, relationship_answer_id: str):
```
MebinAbraham: Should this not be `get` and not `check` since it is getting the value?

MebinAbraham (approved):
Minor comments, not going to block.

PR #2279, opened 2019-08-14
Adding validation on number detail answers

rmccar:
Added validation for the number of cars question. Making field a number field and making the max value able to be entered 20

This PR is dependant on the changes in this PR [#149](https://github.com/ONSdigital/eq-schema-validator/pull/149) and this PR [#586](https://github.com/ONSdigital/design-system/pull/586) (this will need to be merged first and then design system version upgraded before tests will pass)

### How to review 
- Enter non numeric values into field and check that it errors
- Enter a number of cars greater than 20 and check that it errors


rmccar:
[Number write in field on radio option (on 'Other') (S)](https://trello.com/c/QwU5rLcl/3183-number-write-in-field-on-radio-option-on-other-s)

bitdivision:
Detail answer is currently enforced as textfield only in eq-schema-validator. You'll need to make a corresponding PR there

JamesGardiner:
@bitdivision is there a reason it's enforced as textfield?

bitdivision:
I would think because we've never tested it with any other field type.

I suggest we either only allow fields that have test schemas in runner, or make test schemas for all the available types.

scripts/test_schemas.sh
```diff
@@ -6,7 +6,7 @@ if [ "$1" == "--local" ] || [ "$2" == "--local" ]; then
 fi
 
 if [ "$run_docker" == true ]; then
-    branch=v3
+    branch=add-validation-number-of-cars
```
bitdivision: This is the old branch, should be allow-numeric-detail-answer

bitdivision (changes requested):
We should really add a numeric field to a test schema since this is now allowed in validator, potentially also a functional test.

JamesGardiner (changes requested):
Change looks good - needs new tests (test schema and functional test) to verify numeric field works as expected.

data-source/json/test_checkbox.json
```diff
@@ -156,6 +156,95 @@
...
+                                                "q_code": "2"
+                                            },
+                                            {
+                                                "label": "Other",
+                                                "q_code": "3",
+                                                "description": "Choose any number of toppings",
+                                                "value": "Other",
+                                                "detail_answer": {
+                                                    "mandatory": true,
+                                                    "id": "other-answer-mandatory",
```
LJBabbage: Double up of id 'other-answer-mandatory' which will fail validation

tests/app/views/contexts/summary/test_question.py
```diff
@@ -268,6 +268,43 @@ def test_checkbox_answer_with_detail_answer_returns_the_value(self):
         self.assertEqual(len(question.answers[0]['value']), 2)
         self.assertEqual(question.answers[0]['value'][1].detail_answer_value, 'Test')
 
+    def test_checkbox_answer_with_numeric_detail_answer_returns_the_value(self):
```
LJBabbage: Maybe test_checkbox_answer_with_numeric_detail_answer_returns_number?

tests/app/views/contexts/summary/test_question.py
```diff
@@ -268,6 +268,43 @@ def test_checkbox_answer_with_detail_answer_returns_the_value(self):
...
+        # Given
+        self.answer_store.add_or_update(
+            Answer(answer_id='answer_1', value=['Light Side', 'Other'])
```
LJBabbage: This doesn't need to be changed, but the question doesn't make a great deal of sense, Light side or 1. If there is no better question then happy for it to stay as is

data-source/json/test_checkbox.json
```diff
@@ -156,6 +156,95 @@
...
+                                                "label": "2",
+                                                "value": "2"
+                                            },
+                                            {
+                                                "label": "Other",
+                                                "value": "Other",
+                                                "description": "Choose any number of toppings",
+                                                "detail_answer": {
+                                                    "mandatory": false,
+                                                    "id": "other-answer-non-mandatory",
```
LJBabbage: Another duplicate id 'other-answer-non-mandatory'

LJBabbage (changes requested):

tests/app/views/contexts/summary/test_question.py
rmccar: This question is just copied from the test before it. Don't think it matters too much what the question is for the test but happy to change if people think it needs to make more sense

bitdivision: Yeah, it would be better if this did make sense I think. Helps people who are looking at the test to see the use case. Not hugely important though.

JamesGardiner:
Needs formatting

ajmaddaford (changes requested):
The other number field doesn't behave exactly like a standard number field:
- Entering zero and going back the field isn't populated
- Entering letters and submitting the invalid field content is wiped

The test schemas should have some validation added so this can be easily tested without going through the census schema. We should also consider separating out the other detail answer variations into separate schemas as has been done for the radio variations.

ajmaddaford (changes requested):
The other number field doesn't behave exactly like a standard number field:
- Entering zero and going back the field isn't populated
- Entering letters and submitting the invalid field content is wiped

The test schemas should have some validation added so this can be easily tested without going through the census schema. We should also consider separating out the other detail answer variations into separate schemas as has been done for the radio variations.

tests/functional/spec/components/checkbox/checkbox_numeric_detail_answers.spec.js
```diff
@@ -0,0 +1,116 @@
...
+const CheckboxNumericDetailPage = require('../../../generated_pages/checkbox_numeric_detail_answers/checkbox-numeric-detail.page');
+const SummaryPage = require('../../../generated_pages/checkbox_numeric_detail_answers/summary.page');
+
+describe('Checkbox with a numeric "detail_answer" option', function() {
+
+  const checkbox_schema = 'test_checkbox_numeric_detail_answers.json';
+
+  it('Given a numeric detail answer options are available, When the user clicks an option, Then the detail answer input should be visible.', function() {
+    // Given
+    return helpers.openQuestionnaire(checkbox_schema).then(() => {
```
JamesGardiner: this helper can be made into a beforeEach on the describe method

tests/functional/spec/components/checkbox/checkbox_numeric_detail_answers.spec.js
```diff
@@ -0,0 +1,116 @@
...
+      return browser
+      // When
+        .click(CheckboxNumericDetailPage.other())
```
JamesGardiner: This is also executed for each test so can be moved into the beforeEach method available on describe

tests/functional/spec/components/radio/radio_numeric_detail_answers.spec.js
```diff
@@ -0,0 +1,116 @@
...
+const RadioNumericDetailPage = require('../../../generated_pages/radio_numeric_detail_answers/radio-numeric-detail.page');
+const SummaryPage = require('../../../generated_pages/radio_numeric_detail_answers/summary.page');
+
+describe('Radio with a numeric "detail_answer" option', function() {
+
+  const radio_schema = 'test_radio_numeric_detail_answers.json';
+
+  it('Given a numeric detail answer options are available, When the user clicks an option, Then the detail answer input should be visible.', function() {
+    // Given
+    return helpers.openQuestionnaire(radio_schema).then(() => {
```
JamesGardiner: Same as above, opening the questionnaire should be done in a beforeEach on the describe

JamesGardiner (changes requested):

app/jinja_filters.py
```diff
@@ -279,7 +279,16 @@ class OtherConfig:
...
+
+        if detail_answer.data or detail_answer.data == 0:
+            self.value = detail_answer.data
+        elif detail_answer.raw_data and (
+            detail_answer.raw_data[0] or detail_answer.raw_data[0] == 0
+        ):
+            self.value = detail_answer.raw_data[0]
+        else:
+            self.value = ''
+
```
ajmaddaford: This should be consistent with how it is done in the textfield and number templates. All of this can be replaced with `self.value = escape(detail_answer._value())`.

ajmaddaford (changes requested):

app/jinja_filters.py
```diff
@@ -279,7 +279,16 @@ class OtherConfig:
     def __init__(self, detail_answer):
         self.id = detail_answer.id
         self.name = detail_answer.name
-        self.value = detail_answer.data or ''
+
+        if detail_answer.data or detail_answer.data == 0:
+            self.value = detail_answer.data
+        elif detail_answer.raw_data and (
+            detail_answer.raw_data[0] or detail_answer.raw_data[0] == 0
```
LJBabbage: Don't believe you need the detail_answer.raw_data[0] == 0 check as '0' would be valid and therefore caught in the first if

LJBabbage (changes requested):

app/jinja_filters.py
rmccar: It was needed because even though 0 is valid, it would be falsy. But it doesn't matter now because I've removed this to address Andrews comment

ajmaddaford (approved):

JamesGardiner (approved):

LJBabbage (approved):

PR #2293, opened 2019-08-19
Add lists to section summary

iwootten:

Section summary now includes a summary of list collectors within a section. It is assumed traditional questions won't be mixed with list collector sections, so list collector summaries appear beneath a usual section summary to reduce the complexity of mixing different content types on templates. This also means all summaries can continue to share the same template.

Additionally, I've modified the tests to share common fixtures between list collector tests.

### How to review 

See if you agree with the keys that have been added, the way additional context has been introduced for views. Uses the new keys introduced in https://github.com/ONSdigital/eq-schema-validator/pull/151

### Checklist


iwootten:
[Add lists to section summary (L)](https://trello.com/c/7ld7bDvO/3134-add-lists-to-section-summary-l)

scripts/test_schemas.sh
```diff
@@ -6,7 +6,7 @@ if [ "$1" == "--local" ] || [ "$2" == "--local" ]; then
 fi
 
 if [ "$run_docker" == true ]; then
-    branch=v3
+    branch="eq-3134-add-lists-to-section-summary"
```
iwootten: Reminder to be changed back once the validator PR is merged

app/views/contexts/summary_context.py
```diff
@@ -111,7 +114,31 @@ def build_view_context_for_section_summary(
...
+                block_id=list_collector_block['add_block']['id'],
+            ),
+            'add_link_text': list_collector_block['summary']['add_link_text'],
+            'empty_list_text': list_collector_block['summary']['empty_list_text'],
+            'list_items': build_list_items_summary_context(
+                list_collector_block, schema, answer_store, list_store, language
+            ),
+            'list_name': list_collector_block['for_list'],
+        }
+        list_item_summaries.append(list_item_summary)
```
ajmaddaford: I think when referring to the summary of a list we should not include the word `item` i.e. `list_item_summary` should be shortened to `list_summary` and `list_item_summaries` to `list_summaries`.

LJBabbage (commented):
Not sure this is working for me, using test_list_collector_section_summary.json I get a grey button on the last screen that doesn't seem to do anything
Screen Shot 2019-08-21 at 09 49 30

tests/app/views/contexts/conftest.py
```diff
@@ -0,0 +1,166 @@
...
+            {'answer_id': 'last-name', 'value': 'Morrison', 'list_item_id': 'PlwgoG'},
+            {'answer_id': 'first-name', 'value': 'Barry', 'list_item_id': 'UHPLbX'},
+            {'answer_id': 'last-name', 'value': 'Pheloung', 'list_item_id': 'UHPLbX'},
+        ]
+    )
+
+
+@pytest.fixture
+def people_list_store():
+    return ListStore([{'items': ['PlwgoG', 'UHPLbX', 'FnoDHP'], 'name': 'people'}])
```
ajmaddaford: Where does `FnoDHP` come from?

iwootten:
> Not sure this is working for me, using test_list_collector_section_summary.json I get a grey button on the last screen that doesn't seem to do anything

Yes, that's the final summary as it stands - included here only because validator requires it.

ajmaddaford (changes requested):
Looks good. The assumptions made about when lists will appear on section summaries makes sense. 

When choosing add/edit/remove from the section summary and then submitting the add/edit/remove screen, the user should be returned to the section summary rather than the list collector.

tests/app/views/contexts/conftest.py
iwootten: This is an old item that is no longer in the tests

iwootten:
> When choosing add/edit/remove from the section summary and then submitting the add/edit/remove screen, the user should be returned to the section summary rather than the list collector.

We had a couple of discussions around this in your absence, we decided the collector confirmation step and then return to summary was good enough. As discussed offline, given as this behaviour has been added as part of @MebinAbraham's work, we should create a card to return to the summary immediately using it once merged.

iwootten:
Added a card to cover returning to summary here: https://trello.com/c/yq5GWXVc/3231-add-returnto-parameter-to-list-summaries

app/views/contexts/list_collector.py
```diff
@@ -6,12 +6,13 @@
...
-    rendered_block, schema, answer_store, list_store, language
+    list_collector_block, schema, answer_store, list_store, language
 ):
-    list_name = rendered_block['for_list']
+    list_name = list_collector_block['for_list']
     list_item_ids = list_store[list_name].items
     list_answer_ids = [
-        answer['id'] for answer in rendered_block['add_block']['question']['answers']
+        answer['id']
+        for answer in list_collector_block['add_block']['question']['answers']
```
ajmaddaford: This breaks if there are variants in the `add_block`. I've had a look and I don't think `list_answer_ids` is needed as it's only used to add an `answers` key to the list item, which is not used now that `item_title` exists.

ajmaddaford (approved):

app/views/contexts/summary_context.py
```diff
@@ -111,7 +114,29 @@ def build_view_context_for_section_summary(
...
     )
 
-    context['summary'].update({'title': title})
+    list_collector_blocks = schema.get_visible_list_blocks_for_section(section)
+
+    list_summaries = []
+
+    for list_collector_block in list_collector_blocks:
+        list_summary = {
+            'title': list_collector_block['summary']['title'],
```
MebinAbraham: We need to add support for placeholder rendering for `title` as on census it pipes in the census date.

MebinAbraham (changes requested):

MebinAbraham (approved):

PR #2294, opened 2019-08-19
Amending Census Downstream Submission Data

berroar:
Amending Census downstream submission data - [See Trello](https://trello.com/c/ed4j3mgZ/2963-sending-census-submission-data-downstream-m).

We currently send the data field in the following format:
```python
data: [{
  answer_id: 'test'
  list_item_id: '123'
  value: 'testing'
}]
```

This was amended to be:
```python
data: {
    answers: [{
    answer_id: 'test'
    list_item_id: '123'
    value: 'testing'
  }],
  lists: [{
    name: 'people',
    items: ['123', 'abc']
  }]
}
```

### How to review 
Use a test schema and check that the 'data' field in the submission data matches the new format shown above.

### DIFF
--- a/app/submitter/converter.py
+++ b/app/submitter/converter.py
@@ -90,9 +90,12 @@ def convert_answers(schema, questionnaire_store, routing_path, flushed=False):
         payload['case_ref'] = metadata['case_ref']
 
     if schema.json['data_version'] == '0.0.3':
-        payload['data'] = convert_answers_to_payload_0_0_3(
-            answer_store, list_store, schema, routing_path
-        )
+        payload['data'] = {
+            'answers': convert_answers_to_payload_0_0_3(
+                answer_store, list_store, schema, routing_path
+            ),
+            'lists': list_store.serialise(),
+        }
     elif schema.json['data_version'] == '0.0.1':
         payload['data'] = convert_answers_to_payload_0_0_1(
             metadata, answer_store, list_store, schema, routing_path
--- a/tests/app/submitter/test_convert_payload_0_0_3.py
+++ b/tests/app/submitter/test_convert_payload_0_0_3.py
@@ -66,9 +66,9 @@ def test_convert_answers_to_payload_0_0_3(fake_questionnaire_store):
     )
 
     # Then
-    assert len(answer_object['data']) == 2
-    assert answer_object['data'][0].value == 'Joe Bloggs'
-    assert answer_object['data'][1].value, '62 Somewhere'
+    assert len(answer_object['data']['answers']) == 2
+    assert answer_object['data']['answers'][0].value == 'Joe Bloggs'
+    assert answer_object['data']['answers'][1].value, '62 Somewhere'
 
 
 def test_convert_payload_0_0_3_multiple_answers(fake_questionnaire_store):
@@ -104,8 +104,8 @@ def test_convert_payload_0_0_3_multiple_answers(fake_questionnaire_store):
         QuestionnaireSchema(questionnaire), fake_questionnaire_store, routing_path
     )
     # Then
-    assert len(answer_object['data']) == 1
-    assert answer_object['data'][0].value == ['Ready salted', 'Sweet chilli']
+    assert len(answer_object['data']['answers']) == 1
+    assert answer_object['data']['answers'][0].value == ['Ready salted', 'Sweet chilli']
 
 
 def test_radio_answer(fake_questionnaire_store):
@@ -137,8 +137,8 @@ def test_radio_answer(fake_questionnaire_store):
         QuestionnaireSchema(questionnaire), fake_questionnaire_store, routing_path
     )
 
-    assert len(answer_object['data']) == 1
-    assert answer_object['data'][0].value == 'Coffee'
+    assert len(answer_object['data']['answers']) == 1
+    assert answer_object['data']['answers'][0].value == 'Coffee'
 
 
 def test_number_answer(fake_questionnaire_store):
@@ -161,8 +161,8 @@ def test_number_answer(fake_questionnaire_store):
         QuestionnaireSchema(questionnaire), fake_questionnaire_store, routing_path
     )
 
-    assert len(answer_object['data']) == 1
-    assert answer_object['data'][0].value == 1.755
+    assert len(answer_object['data']['answers']) == 1
+    assert answer_object['data']['answers'][0].value == 1.755
 
 
 def test_percentage_answer(fake_questionnaire_store):
@@ -185,8 +185,8 @@ def test_percentage_answer(fake_questionnaire_store):
         QuestionnaireSchema(questionnaire), fake_questionnaire_store, routing_path
     )
 
-    assert len(answer_object['data']) == 1
-    assert answer_object['data'][0].value == 99
+    assert len(answer_object['data']['answers']) == 1
+    assert answer_object['data']['answers'][0].value == 99
 
 
 def test_textarea_answer(fake_questionnaire_store):
@@ -211,8 +211,8 @@ def test_textarea_answer(fake_questionnaire_store):
         QuestionnaireSchema(questionnaire), fake_questionnaire_store, routing_path
     )
 
-    assert len(answer_object['data']) == 1
-    assert answer_object['data'][0].value == 'This is an example text!'
+    assert len(answer_object['data']['answers']) == 1
+    assert answer_object['data']['answers'][0].value == 'This is an example text!'
 
 
 def test_currency_answer(fake_questionnaire_store):
@@ -235,8 +235,8 @@ def test_currency_answer(fake_questionnaire_store):
         QuestionnaireSchema(questionnaire), fake_questionnaire_store, routing_path
     )
 
-    assert len(answer_object['data']) == 1
-    assert answer_object['data'][0].value == 100
+    assert len(answer_object['data']['answers']) == 1
+    assert answer_object['data']['answers'][0].value == 100
 
 
 def test_dropdown_answer(fake_questionnaire_store):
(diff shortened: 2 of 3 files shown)


berroar:
[Sending census submission data downstream (M)](https://trello.com/c/ed4j3mgZ/2963-sending-census-submission-data-downstream-m)

app/submitter/convert_payload_0_0_3.py
```diff
@@ -72,3 +72,14 @@ def add_list_collector_answers(
...
                 if answer:
                     answers_payload.add_or_update(answer)
+
+
+def convert_lists_to_payload_0_0_3(list_store) -> Dict[str, Sequence[str]]:
+
+    lists_payload = {'name': '', 'items': []}
+
+    list_items = list_store.serialise()
+    for item in list_items:
```
JamesGardiner:
```python
return {'name': '', 'items': list_store.serialise()}
```

tests/integration/routes/test_dump.py
```diff
@@ -136,7 +139,11 @@ def test_dump_submission_authenticated_with_role_with_answers(self):
                     'exercise_sid': '789',
                     'schema_name': 'test_radio_mandatory',
                 },
-                'data': [{'answer_id': 'radio-mandatory-answer', 'value': 'Coffee'}],
+                'data': {
+                    'answers': [{'answer_id': 'radio-mandatory-answer',
+                                 'value': 'Coffee'}],
+                    'lists': {}
```
JamesGardiner: This needs an additional test to check the list key is correctly populated

app/submitter/convert_payload_0_0_3.py
```diff
@@ -72,3 +72,14 @@ def add_list_collector_answers(
                 answer = answer_store.get_answer(answer_id, list_item_id)
                 if answer:
                     answers_payload.add_or_update(answer)
+
+
+def convert_lists_to_payload_0_0_3(list_store) -> Dict[str, Sequence[str]]:
+
+    lists_payload = {'name': '', 'items': []}
```
JamesGardiner: Where will name get populated?

JamesGardiner (changes requested):

app/submitter/converter.py
```diff
@@ -76,6 +77,13 @@ def convert_answers(schema, questionnaire_store, routing_path, flushed=False):
         'questionnaire_id': metadata['questionnaire_id'],
     }
 
+    data = {
```
LJBabbage: Any reason this is outside the 0.0.3 if statement, I can't see it being used anywhere else?

app/submitter/converter.py
```diff
@@ -76,6 +77,13 @@ def convert_answers(schema, questionnaire_store, routing_path, flushed=False):
...
+        'answers': convert_answers_to_payload_0_0_3(
+            answer_store, list_store, schema, routing_path
+        ),
+        'lists': convert_lists_to_payload_0_0_3(list_store),
```
LJBabbage: I might be missing something, but doesn't list_store.serialise() give you what you need here? Meaning you can get rid of all the code for convert_lists_to_payload_0_0_3(
```'lists': list_store.serialise()```

File: tests/integration/routes/test_dump.py [hunk 6]
```diff
@@ -136,7 +136,12 @@ def test_dump_submission_authenticated_with_role_with_answers(self):
                     'exercise_sid': '789',
                     'schema_name': 'test_radio_mandatory',
                 },
-                'data': [{'answer_id': 'radio-mandatory-answer', 'value': 'Coffee'}],
+                'data': {
+                    'answers': [
+                        {'answer_id': 'radio-mandatory-answer', 'value': 'Coffee'}
+                    ],
+                    'lists': {},
```
LJBabbage: I might have missed it but I didn't see any tests, testing the lists being in 0.0.3?

LJBabbage (changes requested):

app/submitter/converter.py
MebinAbraham: Since this is only used for `0.0.3` this could be moved under `if schema.json['data_version'] == '0.0.3':` so we only create it when needed.

app/submitter/convert_payload_0_0_3.py
```diff
@@ -72,3 +72,14 @@ def add_list_collector_answers(
                 answer = answer_store.get_answer(answer_id, list_item_id)
                 if answer:
                     answers_payload.add_or_update(answer)
+
+
+def convert_lists_to_payload_0_0_3(list_store) -> Dict[str, Sequence[str]]:
```
MebinAbraham: This should be returning a list of dicts not a single dict as you can have more the one list in a survey.

MebinAbraham (dismissed):

app/submitter/convert_payload_0_0_3.py
JamesGardiner:
```Python
def convert_lists_to_payload_0_0_3(list_store) -> Sequence[Mapping]:
    serialised = list_store.serialise()
    lists = [{'name': list_['name'], 'items': list_['items'], 'primary_person': list_.get('primary_person')} for list_ in ]
	
	for list_ in lists:
		if list_['primary_person'] is None:
			del list_['primary_person']
	return lists
```

will do the trick

petechd: James do you mean:
```Python
def convert_lists_to_payload_0_0_3(list_store) -> Sequence[Mapping]:
    serialised = list_store.serialise()
    lists = [{'name': list_['name'], 'items': list_['items'], 'primary_person': list_.get('primary_person')} for list_ in serialised]
	
	for list_ in lists:
		if list_['primary_person'] is None:
			del list_['primary_person']
	return lists
```
or should I use list_store?

JamesGardiner: Yes, yes I do

app/submitter/converter.py
JamesGardiner: Good point - I think there would still need to be some code to filter out `primary_person` key value pairs where the value is None

app/submitter/converter.py
petechd: It's done now

tests/integration/routes/test_dump.py
petechd: We've added extra test for it

app/submitter/converter.py
petechd: We've added it now, thanks

app/submitter/converter.py
```diff
@@ -90,9 +91,12 @@ def convert_answers(schema, questionnaire_store, routing_path, flushed=False):
...
 
     if schema.json['data_version'] == '0.0.3':
-        payload['data'] = convert_answers_to_payload_0_0_3(
-            answer_store, list_store, schema, routing_path
-        )
+        payload['data'] = {
+            'answers': convert_answers_to_payload_0_0_3(
+                answer_store, list_store, schema, routing_path
+            ),
+            'lists': convert_lists_to_payload_0_0_3(list_store),
```
JamesGardiner: I don't think it's worth having this as a function - it could just as easily be:

```
'lists': list_store.serialise()
```

JamesGardiner (changes requested):

JamesGardiner (approved):

iwootten (dismissed):
There only appears to be a single integration test covering the new list serialisation in submitter. It would make sense to introduce/extend current unit tests to cover the submitter behaviour - that the list structure/order is as expected and that primary person keys are included.

tests/app/submitter/test_convert_payload_0_0_3.py
```diff
@@ -390,7 +390,7 @@ def test_primary_person_list_item_conversion(fake_questionnaire_store):
 
     output = convert_answers(schema, fake_questionnaire_store, routing_path)
 
-    data_dict = json.loads(json.dumps(output['data'], for_json=True))
+    data_dict = json.loads(json.dumps(output['data']['answers'], for_json=True))
```
ajmaddaford, line 393: Now that the list items are in the payload we should add tests for them as well as answers.

ajmaddaford (changes requested):

tests/app/submitter/test_convert_payload_0_0_3.py
```diff
@@ -492,7 +492,7 @@ def test_default_answers_not_present_when_not_answered(fake_questionnaire_store)
     ]
 
     output = convert_answers(schema, fake_questionnaire_store, routing_path)
-    data = json.loads(json.dumps(output['data'], for_json=True))
+    data = json.loads(json.dumps(output['data']['answers'], for_json=True))
```
bitdivision, line 495: As far as I can see, the unit tests still never assert anything about the lists key?

petechd: We've added two tests to check lists creation: test_list_structure_in_payload_is_as_expected & test_primary_person_not_in_payload_when_not_answered

tests/app/submitter/test_convert_payload_0_0_3.py
```diff
@@ -496,3 +496,85 @@ def test_default_answers_not_present_when_not_answered(fake_questionnaire_store)
...
+
+    output = convert_answers(schema, fake_questionnaire_store, routing_path)
+
+    data_dict = json.loads(json.dumps(output['data']['lists'], for_json=True))
+
+    keys = []
+    for _, lists_key in enumerate(data_dict[0]):
+        keys.append(lists_key)
+
+    assert 'primary_person' not in keys
```
ajmaddaford: This could be simplified to:
```python
assert 'primary_person' not in data_dict[0]
```

petechd: @ajmaddaford we've simplified this test now, thanks

ajmaddaford (approved):

LJBabbage (approved):

PR #2302, opened 2019-08-21
Define relationship name piping in schema

JamesGardiner:
Allows the definition of relationship name piping in schema

### How to review 
Run the tests. Launch the relationship schemas and check that strings are correctly rendered and that the correct placeholders are used.

Test that the who lives here sections for NIR and ENG/WAL census are rendering correctly


JamesGardiner:
[Define relationship name piping in schema (M)](https://trello.com/c/6Lpfmh3R/3152-define-relationship-name-piping-in-schema-m)

app/views/handlers/relationship_collector.py
```diff
@@ -118,3 +111,27 @@ def _is_last_relationship(self):
...
+            self._current_location,
+        )
+
+        placeholder_renderer = PlaceholderRenderer(
+            language=self._language,
+            schema=self._schema,
+            answer_store=self._questionnaire_store.answer_store,
+            metadata=self._questionnaire_store.metadata,
+            location=self._current_location,
+        )
```
ajmaddaford: I think if you add location to the placeholder renderer initialisation in the base block handler you won't need to override the `_render_block` method.

ajmaddaford (changes requested):
I can't see any unit tests for location in placeholder parser. We also seem to be missing tests for list item id there as well 😩

app/views/handlers/relationship_collector.py
JamesGardiner: > I can't see any unit tests for location in placeholder parser. We also seem to be missing tests for list item id there as well weary

Will add these in now; just waiting to see if this is the preferred approach

iwootten: We already have this in v3 if you merge in the latest changes.

data-source/json/test_relationships.json
```diff
@@ -167,31 +167,351 @@
...
+                                        {
+                                            "placeholder": "first_person_name",
+                                            "transforms": [
+                                                {
+                                                    "transform": "concatenate_list",
+                                                    "arguments": {
+                                                        "list_to_concatenate": {
+                                                            "source": "answers",
+                                                            "identifier": ["first-name", "last-name"],
+                                                            "list_item_selector": "list_item_id"
```
iwootten: Are we decided on 'list_item_selector'? We use 'id_selector' or 'id' within an object within when rules to indicate which id to select.

ajmaddaford: We could use the same format as the when rules:
```
{
	"list_item_selector": { 
		"source": "location",
		"id": "list_item_id"
	}
}

Comment (iwootten) [reply data-source/json/test_relationships.json]: This was my thinking - currently it isn't immediately obvious from the arguments the source of the id.

Comment (JamesGardiner) [reply data-source/json/test_relationships.json]: I'll make the change in the validator pr as well

File: scripts/test_schemas.sh [hunk 3]
```diff
@@ -6,7 +6,7 @@ if [ "$1" == "--local" ] || [ "$2" == "--local" ]; then
 fi
 
 if [ "$run_docker" == true ]; then
-    branch=v3
+    branch=relationship-name-piping
```
JamesGardiner: Reminder to revert this before merging

app/questionnaire/placeholder_parser.py
```diff
@@ -24,35 +27,41 @@ def __init__(
...
 
-    def _lookup_answer(self, answer_id, list_item_id=None):
-        answer = self._answer_store.get_answer(answer_id, list_item_id)
-
-        if answer:
-            return answer.value
+    def __call__(self, placeholder_list: Sequence[Mapping]) -> Mapping:
+        for placeholder in placeholder_list:
+            try:
+                self._placeholder_map[placeholder['placeholder']]
```
ajmaddaford: Is this just to check if the placeholder is already in _placeholder_map? If so, it would be clearer replace the try/except with: 

```python
if placeholder['placeholder'] not in self._placeholder_map
```

ajmaddaford (changes requested):

data-source/jsonnet/england-wales/lib/placeholders.libsonnet
```diff
@@ -54,4 +54,64 @@
...
+      {
+        transform: 'format_possessive',
+        arguments: {
+          string_to_format: {
+            source: 'previous_transform'
+          }
+        }
+      }
+    ]
+  },
```
ajmaddaford: These placeholder shouldn't be in this file. 
- This file is intended for placeholders that are used in a number of places
- They don't make sense outside of the context of the relationship collector block

ajmaddaford (changes requested):

data-source/jsonnet/england-wales/lib/placeholders.libsonnet
JamesGardiner: Fair enough, I originally had them inside the relationships file. I'll move them back

ajmaddaford (approved):

app/questionnaire/placeholder_parser.py
```diff
@@ -9,52 +12,63 @@ class PlaceholderParser:
...
-
-        answer = self._answer_store.get_answer(answer_id, list_item_id)
-
-        if answer:
-            return answer.value
-
-    def parse(self, placeholder_list):
-        placeholder_map = {}
-
+    def __call__(self, placeholder_list: Sequence[Mapping]) -> Mapping:
```
bitdivision, line 32: :+1:

tests/app/questionnaire/conftest.py
```diff
@@ -1,5 +1,47 @@
+# pylint: disable=redefined-outer-name
 import pytest
 
+from app.data_model.answer_store import AnswerStore
+from app.questionnaire.location import Location
+from app.questionnaire.placeholder_parser import PlaceholderParser
+
+
+@pytest.fixture
+def placeholder_list():
```
bitdivision, line 10: These guys are only really used in one place, so not sure of the value of having them in the conftest, but it's not a blocking issue.

bitdivision (approved):
I can't break it. :+1:

PR #2305, opened 2019-08-22
Move placeholder renderer to handler

iwootten:

Reduce the use of different placeholder renderer instances and instead add a common one to block handler. I've kept this out of my current PR to keep things simple to review.

### How to review 

See if you agree with the approach


JamesGardiner (approved):

JamesGardiner:
Changes look good, but fails schema validation

bitdivision (approved):

PR #2310, opened 2019-08-23
Section summaries and titles for repeating sections

bitdivision:
New schema for placeholders on repeating section titles.
Section summaries should support repeating sections.
Hub and spoke should show titles for repeating sections.
Section summaries should show titles for repeating sections.

I also refactored summary_context a little. It still needs work, and calculated summary hasn't really been touched (just moved).

### How to review 
Check section summaries support repeating sections.
Check titles are rendered correctly on section summaries and on hubs.

### Checklist


tests/app/views/contexts/test_list_collector.py
```diff
@@ -1,11 +1,11 @@
-from app.data_model.answer import Answer
+import pytest
 from app.utilities.schema import load_schema_from_name
 from app.views.contexts import list_collector
 from app.questionnaire.questionnaire_schema import DEFAULT_LANGUAGE_CODE
 
-
+@pytest.mark.usefixtures('app')
```
bitdivision, line 7: I've been getting lint errors for these fixtures. Not sure what has changed, but this seems to resolve it,.

app/views/contexts/summary/block.py
```diff
@@ -15,22 +15,31 @@ def __init__(
...
+            )
+        else:
+            return url_for('questionnaire.block', block_id=block_id)
 
     @staticmethod
     def get_question(
         block_schema, answer_store, list_store, metadata, schema, current_location=None
     ):
         """ Taking question variants into account, return the question which was displayed to the user """
+        list_item_id = current_location.list_item_id if current_location else None
```
MebinAbraham: Shouldn't need this as `current_location.list_item_id` should always be set. (default=None).

app/routes/questionnaire.py
```diff
@@ -121,8 +121,11 @@ def get_questionnaire(schema, questionnaire_store):
     language_code = get_session_store().session_data.language_code
 
     hub = HubContext(
+        language_code,
         questionnaire_store.progress_store,
         questionnaire_store.list_store,
+        questionnaire_store.answer_store,
```
MebinAbraham, line 127: Prob not as part of this, feel like hub should just be passed questionnaire store.

MebinAbraham (changes requested):
Looks good.

Would be worth reviewing these test and changing them to assert the section name, since runner now supports it?

https://github.com/ONSdigital/eq-survey-runner/blob/v3/tests/integration/questionnaire/test_questionnaire_hub.py#L174-L182

app/views/contexts/hub_context.py
```diff
@@ -102,19 +113,37 @@ def get_section_url(section_id, list_item_id) -> str:
...
+    def _get_row_for_repeating_section(self, section_title, section_id, list_item_id):
+        repeating_title = self._schema.get_repeating_title_for_section(section_id)
+
+        placeholder_renderer = PlaceholderRenderer(
+            language=self._language,
+            schema=self._schema,
+            answer_store=self._answer_store,
+            metadata=self._metadata,
+            list_item_id=list_item_id,
+        )
```
ajmaddaford, line 133: We should consider initialising the placeholder renderer once, rather than for every repeating section title. 

Probably a separate card, but we should also consider supporting placeholders in section titles that don't repeat.

MebinAbraham (commented):
Placeholders in question title don't seem to be working?
Screenshot 2019-08-27 at 15 22 25

app/views/contexts/hub_context.py
```diff
@@ -102,19 +113,37 @@ def get_section_url(section_id, list_item_id) -> str:
 
         return url_for('questionnaire.get_section', section_id=section_id)
 
-    def _get_row_for_section(self, section_name, section_id, list_item_id=None):
+    def _get_row_for_repeating_section(self, section_title, section_id, list_item_id):
```
MebinAbraham: `section_title` is unused

app/views/contexts/summary/block.py
bitdivision: Not if the argument is not provided?

MebinAbraham: Ah ofc, misread.

app/routes/questionnaire.py
bitdivision: I actually started passing SummaryContext the questionnaire store for a while, but it became painful since it's used by view submission to generate the context (with no real questionnaire store).

app/views/contexts/hub_context.py
bitdivision: Don't quite understand how that didn't throw a lint error.. 🤔

ajmaddaford (changes requested):
As discussed - the use of current_location needs to determine the summary type should be removed in favour of explicit method calls.

app/views/contexts/hub_context.py
bitdivision: I've added a property for the renderer, and moved list_item_id into the render call to allow us to initialise once.

bitdivision:
@ajmaddaford I've tidied up the use of location a bit. I think it's much better now, but would be great if you could have a look and make sure it looks as you were expecting.

app/views/contexts/hub_context.py
```diff
@@ -47,12 +48,35 @@ class HubContext:
...
         self._progress_store = progress_store
         self._list_store = list_store
+        self._answer_store = answer_store
+        self._metadata = metadata
         self._schema = schema
         self._survey_complete = survey_complete
+        self._placeholder_renderer = None
+
+    @property
+    def placeholder_renderer(self):
```
MebinAbraham, line 72: Cached property?
```py
    @cached_property
    def placeholder_renderer(self):
        self._placeholder_renderer = PlaceholderRenderer(
            language=self._language,
            schema=self._schema,
            answer_store=self._answer_store,
            metadata=self._metadata,
        )

        return self._placeholder_renderer
```

MebinAbraham (approved):
Functionality seems to be all there. 👍

app/views/contexts/calculated_summary.py
```diff
@@ -0,0 +1,188 @@
...
+from app.questionnaire.location import Location
+from app.views.contexts.summary.group import Group
+
+
+def build_groups_for_section(answer_store, list_store, metadata, schema, section):
+    path_finder = PathFinder(schema, answer_store, metadata, list_store=list_store)
+
+    section_path = path_finder.routing_path(section['id'])
+
+    location = Location(section['id'])
```
MebinAbraham, line 21: Wouldn't work for repeats?.

bitdivision: Nope. I've moved calculated summary out of the summary, but I think it's out of scope for this PR.

app/views/contexts/hub_context.py
bitdivision: Serves me right! :joy:

MebinAbraham (approved):
Tested and working with household repeating section also. 👍

app/views/contexts/summary_context.py
```diff
@@ -1,316 +1,146 @@
...
+        section_id = current_location.section_id
+        section = self._schema.get_section(current_location.section_id)
+        list_item_id = current_location.list_item_id
+        list_name = current_location.list_name
+        block = self._schema.get_block(current_location.block_id)
 
-    questions_to_keep = [answer['parent_id'] for answer in matching_answers]
+        context = self.build_context_for_section(section_id, list_name, list_item_id)
+        context['summary']['collapsible'] = block.get('collapsible', False)
+        context['summary']['summary_type'] = 'SectionSummary'
```
ajmaddaford: `build_context_for_section` is only called from here. It would be clearer to just inline that method here:
```python
groups = self.build_groups_for_section(section_id, list_name, list_item_id)
context = {
    'summary': {
        'groups': groups, 
        'answers_are_editable': True,
        'collapsible': block.get('collapsible', False),
        'summary_type': 'SectionSummary',
    }
}
```

ajmaddaford (changes requested):

app/views/contexts/summary_context.py
bitdivision: Done

ajmaddaford (approved):
👍

PR #2313, opened 2019-08-27
Simplify Summary Context

iwootten:

As part of the work done in https://github.com/ONSdigital/eq-survey-runner/pull/2293 it was very difficult to understand how summaries were constructed due to the nesting of multiple classes to construct the final summary dictionary. 

The summary context class also has a large number of similarly named methods which operate on the same arguments, all called by different handlers. Most of these arguments are passed in order to construct other objects and evaluate when rules deeper down through multiple method calls. 

This change seeks to remove some of the classes used only to serialise a summary, whilst using the a child class structure within handlers to share behaviour amongst each of the different Summary types. The summary tests have also been broken apart and are testing against appropriate handlers.

### How to review 

See if you agree with the design/approach. Overall, it's up for debate whether handlers should be returning context or calling matching classes to return it. In the case of summaries, there is nothing to each handler, so context can easily be created there without confusion.


app/routes/questionnaire.py
```diff
@@ -379,11 +382,14 @@ def get_view_submission(schema):  # pylint: too-many-locals
...
-            summary_rendered_context = build_summary_rendering_context(
-                schema=schema,
-                answer_store=answer_store,
-                list_store=list_store,
-                metadata=metadata,
+            placeholder_renderer = PlaceholderRenderer(
+                'en', schema, answer_store, metadata
+            )
+
+            submitted_path_finder = PathFinder(schema, answer_store, metadata=metadata)
```
JamesGardiner, line 395:
```suggestion
            path_finder = PathFinder(schema, answer_store, metadata=metadata)
```

app/routes/questionnaire.py
```diff
@@ -379,11 +382,14 @@ def get_view_submission(schema):  # pylint: too-many-locals
...
+
+            summary_rendered_context = build_group_summary_context(
+                schema=schema, path_finder=submitted_path_finder
```
JamesGardiner:
```suggestion
                schema=schema, path_finder=path_finder
```

app/routes/questionnaire.py
```diff
@@ -396,10 +402,21 @@ def get_view_submission(schema):  # pylint: too-many-locals
                 }
             }
 
+            for group in context.get('summary').get('groups'):
```
JamesGardiner, line 413:
```suggestion
            for group in context['summary']['groups']:
```

app/routes/questionnaire.py
```diff
@@ -396,10 +402,21 @@ def get_view_submission(schema):  # pylint: too-many-locals
...
+                for this_block in group.get('blocks'):
```
JamesGardiner, line 414: This could be flattened:

```Python
args = (answer_store, list_store, metadata, schema, None)
this_block['question'] = Block.get_question(this_block['id'], *args)
```

JamesGardiner: With the args tuple defined outside of the loop

app/views/handlers/calculated_summary.py
```diff
@@ -1,14 +1,168 @@
...
+)
+from app.questionnaire.schema_utils import (
+    choose_question_to_display,
+    get_answer_ids_in_block,
+)
+from app.views.handlers.summary import Summary
 
 
-class CalculatedSummary(Content):
+def _get_formatted_total(
```
JamesGardiner, line 14: These huge functions need refactoring.

JamesGardiner (changes requested):
This looks like a positive change to me, but I do have some issues with the current changes:

1. The code slips too easily into heavily nested block of loops and if statements. Where it makes sense code should be flat not nested.
2. There are a number of functions and methods which are unacceptably large and obtuse, and should be refactored if they are going to be part of the pr

app/routes/questionnaire.py
iwootten: This conflicts with the global already imported in the file.

PR #2318, opened 2019-08-28
Add a script to check for @watch in tests

bitdivision:
I always forget to remove @watch statements from functional tests.

### How to review 
Add a watch statement and make sure the lint script fails.


scripts/check_watch_statements.sh
```diff
@@ -0,0 +1,7 @@
+if grep -r "@watch" tests/functional/spec; then
```
MebinAbraham: May as well check for `.only` and `@focus` as well?

bitdivision: Done

scripts/run_lint.sh
```diff
@@ -18,6 +18,9 @@ function display_result {
   fi
 }
 
+./scripts/check_watch_statements.sh
+display_result $? 1 "Functional tests watch check"
```
MebinAbraham: Not just watch now.

MebinAbraham (approved):

scripts/check_watch_statements.sh
```diff
@@ -0,0 +1,6 @@
+if grep -Er "@watch|@focus|\.only" tests/functional/spec; then
```
MebinAbraham: File name prob should be changed also. lint_spec_files?

scripts/run_lint.sh
bitdivision: Yeah, but I don't have a better name, and they all acheive a watch, so seems reasonable to me?

JamesGardiner (approved):

PR #2315, opened 2019-08-27
Census Household: Individual Repeating Section

MebinAbraham:
### What is the context of this PR?
Added the repeating section for the householder.

Most of the individual section is same as the existing individual section, however, the proxy question on individual vs household is the opposite. So to minimise duplication, I made use of jsonnet's `jpath` option to add an additional library path depending on the schema type.

The folder structure has been changed to suit the new changes, hence this PR also effects the individual census schema.

### How to review 
Ensure:
- Changes are as per instructions in Trello.
- Individual census schema has not been altered. Check diff between this and v3 build.

[Quick Launch for GB-ENG](https://v3-launcher.gcp.dev.eq.ons.digital/quick-launch?language_code=en&survey=CENSUS&case_type=HI&region_code=GB-ENG&url=https://raw.githubusercontent.com/ONSdigital/eq-survey-runner/eq-3197-add-individual-repeating-section-eng-wales-ni/data/en/census_household_gb_eng.json)
[Quick Launch for GB-WLS](https://v3-launcher.gcp.dev.eq.ons.digital/quick-launch?language_code=en&survey=CENSUS&case_type=HI&region_code=GB-WLS&url=https://raw.githubusercontent.com/ONSdigital/eq-survey-runner/eq-3197-add-individual-repeating-section-eng-wales-ni/data/en/census_household_gb_wls.json)
[Quick Launch for GB-NIR](https://v3-launcher.gcp.dev.eq.ons.digital/quick-launch?language_code=en&survey=CENSUS&case_type=HI&region_code=GB-ENG&url=https://raw.githubusercontent.com/ONSdigital/eq-survey-runner/eq-3197-add-individual-repeating-section-eng-wales-ni/data/en/census_household_gb_nir.json)

Depends on: https://github.com/ONSdigital/eq-schema-validator/pull/157

PR #2319, opened 2019-08-28
Fix issue with routing while in a repeating section

MebinAbraham:
This PR addresses 2 issues found within a repeating section.

1. Routing/Skip conditions
     
    Previously when evaluating when rules, it was checking whether the location for a given answer exists on the routing path. However, the location was created without taking into account a `list_item_id` and a `list_name` hence it evaluates False. This has now been addressed to ensure the location is built with repeats in mind.

2. Confirmation question

    When routing backwards from a confirmation page a new location is appended to the path, however, this new location was created without `list_item_id` etc hence it was considered not on the path. This PR has addressed this.

3. Fixed an issue where when submitting an empty answer value does not remove the answer from the answer store as the `list_item_id` was not passed. This has now been fixed.

**This is needed for Census Household Repeating Section** #2315 

### How to review 
Ensure:
- Routing and confirmation question work within a repeat.
- Answering a question, then changing the answer and submitting an EMPTY value le don't answer, removes the answer from the answer store.

The `test_confirmation_question_within_repeating_section.json` can be used to test both confirmation question and routing within a repeat.


MebinAbraham:
[Fix routing bug while within a repeat](https://trello.com/c/24CgN70U/3253-fix-routing-bug-while-within-a-repeat)

bitdivision (dismissed):
Looks good :+1:

bitdivision (approved):

ajmaddaford (approved):

bitdivision (approved):

PR #2321, opened 2019-08-29
 Add descriptions to relationship options

ajmaddaford:
Adds descriptions to relationship options.

### How to review 
- Check the test_relationship schema works
- Check the census schemas have the correct content


tests/functional/spec/relationships.spec.js
```diff
@@ -72,6 +72,13 @@ describe('Relationships', function() {
           .getUrl().should.eventually.contain(ListCollectorPage.pageName);
       });
 
+      it.only('And go to the first relationship, Then the \'Brother or Sister\' option should have the text \'Including half brother or half sister\'', function() {
```
JamesGardiner: You may want to remove this .only

bitdivision: If only we had a linter for this stuff...

bitdivision (approved):
:+1: All working.

berroar (approved):

PR #2331, opened 2019-08-30
Escape answers when rendered on a page

ajmaddaford:
Escapes answers when used in placeholders, answer fields and summaries.

Escaping of answers was mistakenly removed in https://github.com/ONSdigital/eq-survey-runner/pull/2040.

For some background, autoescaping was turned off in https://github.com/ONSdigital/eq-survey-runner/commit/159ecadfb97772cb813faf47f75a1e28852a7161 as html content can be defined in any string content in a survey schema. As the only user input we accept are answers, we instead ensure that anywhere we render answers they are escaped.

### How to review 
Use various schemas to enter content with characters that should be escaped (<>&"') and verify they are piped correctly into subsequent screens.


LJBabbage (dismissed):
Works as expected

MebinAbraham (approved):
Looks good, can't break it. As discussed, we should potentially look at escaping all sources that we don't have control of, ie metadata.

app/setup.py
```diff
@@ -178,8 +178,7 @@ def create_app(setting_overrides=None):  # noqa: C901  pylint: disable=too-compl
         # no cache and silence warning
         cache.init_app(application, config={'CACHE_NO_NULL_WARNING': True})
 
-    # Switch off flask default autoescaping as content is html encoded
-    # during schema/metadata/summary context (and navigition) generation
+    # Switch off flask default autoescaping as schema content can contain html
     application.jinja_env.autoescape = False
```
JamesGardiner, line 182: I didn't realise we turned this off globally. Flask docs recommend against this approach and instead recommend turning off autoescape on a per block basis. Won't block this PR but worth noting. https://flask.palletsprojects.com/en/1.1.x/templating/#controlling-autoescaping

JamesGardiner (dismissed):

MebinAbraham (changes requested):
Some issues with escaping.

Steps to reproduce the issue on Section Summary. 
1. Launch `test_placeholder_full` schema.
2. Enter some html into the first name field: Eg.
```html
<!DOCTYPE html>
<html>
<body>

<h1>This is heading 1</h1>
<h2>This is heading 2</h2>
<h3>This is heading 3</h3>
<h4>This is heading 4</h4>
<h5>This is heading 5</h5>
<h6>This is heading 6</h6>

</body>
</html>
```

You should see it rendered on SectionSummary when you progress through.
Screenshot 2019-09-02 at 10 16 38

MebinAbraham (commented):
On relationships, answer options don't escape the inputted name.

1. Launch `test_relationships`
2. Enter some html for the name fields (add 2 people): Eg.
```html
<!DOCTYPE html>
<html>
<body>

<h1>This is heading 1</h1>
<h2>This is heading 2</h2>
</body>
</html>

```

When you click on a Relationship, ie Husband or wife, the html should render.
Screenshot 2019-09-02 at 10 44 17

MebinAbraham (commented):
And as discussed, on answer inputs:
Entering the following closes the tags:
1. Launch `test_relationships`
2. For first name enter: `">Hello` 
3. Click submit

Result:
Screenshot 2019-09-02 at 10 47 45

ajmaddaford:
@MebinAbraham @JamesGardiner @LJBabbage This is ready for another review.

app/setup.py
ajmaddaford: Before autoescaping was turned off `safe` was used throughout the templates to turn off escaping for strings that could contain HTML. Most strings in templates used this approach. This lead to safe being used everywhere without any thought about whether the contents had been escaped or not. By making the choice explicit it made it clearer what should be escaped. 

Also, changing the default to autoescaping on now could have a performance impact - https://jinja.palletsprojects.com/en/2.10.x/faq/#why-is-autoescaping-not-the-default.

JamesGardiner (approved):

MebinAbraham (approved):
Nice 👍

PR #2332, opened 2019-08-30
Update hub 'view answers' to find section summary anywhere in a section

LJBabbage:
Currently if there is a section summary not as the last block in a section, clicking 'view answers' it returns to the last block in the section. This is incorrect and should go back to the last section summary if there is one, or the first block in the section if it doesn't. Additionally if there are no changes on the section then the next location should be the hub. 

N.B The default, back to first block can not return back to the hub if there are no changes as the pickup is section summary or last block. This shouldn't ever be an issue as the intention is always to have a section summary, however as this isn't enforce a default had to be written.

### How to review 
Using test_hub_section_summary.json make sure the 3 sections perform as expected

1. Section one completed- 'view answer' goes to section summary, if there are no changes which affect section completeness, continue returns back to hub. If there are changes which affect section completeness 'continue' will take you back through relationships

2. Section two completed - 'view answers' goes to section summary.

3. Section three completed - 'view answers' goes back to the first block

### DIFF
--- a/app/questionnaire/router.py
+++ b/app/questionnaire/router.py
@@ -44,10 +44,15 @@ class Router:
         current_block_type = self._schema.get_block(location.block_id)['type']
         last_block_location = routing_path[-1]
         last_block_type = self._schema.get_block(last_block_location.block_id)['type']
+        hub_enabled = self._schema.is_hub_enabled()
 
+        # A section summary doesn't always have to be the last block
         if (
-            self._schema.is_hub_enabled()
-            and location.block_id == last_block_location.block_id
+            hub_enabled
+            and (
+                location.block_id == last_block_location.block_id
+                or current_block_type == 'SectionSummary'
+            )
             and self._progress_store.is_section_complete(
                 location.section_id, location.list_item_id
             )
@@ -64,10 +69,9 @@ class Router:
         ):
             return last_block_location.url()
 
-        if self.is_survey_complete():
+        if self.is_survey_complete() and not hub_enabled:
             last_section_id = self._schema.get_section_ids()[-1]
             last_block_id = self._schema.get_last_block_id_for_section(last_section_id)
-
             return Location(section_id=last_section_id, block_id=last_block_id).url()
 
         location_index = routing_path.index(location)
@@ -162,6 +166,11 @@ class Router:
 
         return True
 
+    def get_section_return_location_when_section_complete(
+        self, routing_path
+    ) -> Location:
+        return self._get_location_of_section_summary(routing_path) or routing_path[0]
+
     def _get_allowable_path(self, routing_path):
         """
         The allowable path is the completed path plus the next location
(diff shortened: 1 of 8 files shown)


JamesGardiner (changes requested):
Change looks good but could do with a functional test

app/questionnaire/router.py
```diff
@@ -162,6 +166,9 @@ def is_survey_complete(self):
 
         return True
 
+    def get_location_for_hub_when_section_complete(self, routing_path) -> Location:
```
ajmaddaford: I don't think it's clear what this method does as it sounds like it would return the url of the hub. How about `get_section_return_location_when_section_complete`?

tests/app/questionnaire/test_router.py
```diff
@@ -395,3 +395,150 @@ def test_get_last_complete_location_in_section(self):
             last_complete_location,
             Location(section_id='property-details-section', block_id='insurance-type'),
         )
+
+    def test_get_location_for_hub_when_section_complete_section_summary_mid_block(self):
```
ajmaddaford: `_mid_section` rather than `_mid_block`

tests/app/questionnaire/test_router.py
```diff
@@ -395,3 +395,150 @@ def test_get_last_complete_location_in_section(self):
...
+
+        self.assertEqual(
+            location_when_section_complete,
+            Location(
+                section_id='vehicles-section',
+                block_id='number-of-vehicles-introduction',
+            ),
+        )
+
+    def test_get_next_location_url_for_hub_when_section_complete(self):
```
ajmaddaford: I think the method name would be clearer as `test_get_next_location_url_for_mid_section_summary_when_section_complete`. I think it's also worth adding a test for when the section isn't complete.

ajmaddaford (changes requested):
Changes look good. I think we should consider making the test schema simpler as I'm not sure it needs to be based on households and relationships?

app/questionnaire/router.py
```diff
@@ -64,10 +69,9 @@ def get_next_location_url(self, location, routing_path):
         ):
             return last_block_location.url()
 
-        if self.is_survey_complete():
+        if self.is_survey_complete and not hub_enabled:
```
MebinAbraham: Unless I'm missing something, `self.is_survey_complete` isn't being invoked and isn't a `property` either?

MebinAbraham (changes requested):

app/questionnaire/router.py
```diff
@@ -64,10 +69,9 @@ def get_next_location_url(self, location, routing_path):
...
             last_section_id = self._schema.get_section_ids()[-1]
             last_block_id = self._schema.get_last_block_id_for_section(last_section_id)
-
             return Location(section_id=last_section_id, block_id=last_block_id).url()
```
MebinAbraham, line 75: Existing but this wouldn't technically work for a repeating section but think for now that is fine as we intend to use repeats accompanied by a hub for the time being.

ajmaddaford (approved):

MebinAbraham (approved):

JamesGardiner (approved):

PR #2341, opened 2019-09-02
Toggle question text summary section

JamesGardiner:

Added a 'hide_on_section_summary' key to question objects in schema. When set to true this block will not be included in rendered summary sections.

### How to review 
Complete the Who Lives Here census sections. Assert that only list collectors are displayed on the seciton summary.


app/views/contexts/summary/group.py
```diff
@@ -28,7 +28,11 @@ def _build_blocks(
         block_ids_on_path = [location.block_id for location in path]
 
         for block in group_schema['blocks']:
-            if block['id'] in block_ids_on_path and block['type'] == 'Question':
+            if (
+                block['id'] in block_ids_on_path
+                and block['type'] == 'Question'
+                and not block.get('question', {}).get('hide_on_section_summary', False)
```
MebinAbraham: Does `hide_on_section_summary` need to live within `question`?. We won't have more than 1 question per block hence we can define it at the block level?. And this currently wouldn't work for `question_variants`?

data-source/jsonnet/england-wales/household/blocks/who-lives-here/any_visitors.jsonnet
```diff
@@ -18,6 +18,7 @@ function(census_date) {
     id: 'any-visitors-question',
     title: questionTitle(census_date),
     mandatory: true,
+    hide_on_section_summary: true,
```
MebinAbraham: `Does anyone usually live at` question need to also be hidden from the summary. Also these need to be hidden on all census schema, ie ENG/WLS/NIR.

Screenshot 2019-09-03 at 09 12 11

MebinAbraham (changes requested):
The functionality works as expected.

Just some minor comments.
Should we extend an existing functional test for this?

app/views/contexts/summary/group.py
JamesGardiner: Are you suggesting have this as an allowable key only on Question type blocks?

MebinAbraham: I think that would make sense?. Defined at the block level and validator can ensure it is defined on allowed types.

JamesGardiner: Yeah ok, it's a small change, and would extend the functionality to question_variants

scripts/test_schemas.sh
```diff
@@ -6,7 +6,7 @@ if [ "$1" == "--local" ] || [ "$2" == "--local" ]; then
 fi
 
 if [ "$run_docker" == true ]; then
-    branch=v3
+    branch=add-hide-on-section-summary-to-questions
```
JamesGardiner, line 9: Revert before merge

tests/app/views/contexts/summary/test_group.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+def test_group_serialize_hide_on_summary_false(
+    app, mocker, schema, path
+):  # pylint: disable=unused-argument
+    answer_store = mocker.MagicMock()
+    list_store = mocker.MagicMock()
+    metadata = mocker.MagicMock()
+
+    schema['blocks'][0]['hide_on_section_summary'] = False
```
LJBabbage: As discussed, as we would never see in a normal schema "hide_on_section_summary": False, not sure we should do this in a test

data-source/jsonnet/england-wales/household/blocks/who-lives-here/anyone_usually_live_at.jsonnet
```diff
@@ -23,6 +23,7 @@ function(census_date) {
       when: [rules.hasPrimary],
     },
   ],
+  hide_on_section_summary: true,
```
LJBabbage: Any reason we mix case, in the schema its capitalised or is this a jsonnet thing?

app/views/contexts/summary/group.py
```diff
@@ -28,7 +28,11 @@ def _build_blocks(
         block_ids_on_path = [location.block_id for location in path]
 
         for block in group_schema['blocks']:
-            if block['id'] in block_ids_on_path and block['type'] == 'Question':
+            if (
+                block['id'] in block_ids_on_path
+                and block['type'] == 'Question'
+                and not block.get('hide_on_section_summary', False)
```
LJBabbage: Enjoyable, 99% of the time this will default and then satisfy the condition :smiley:

LJBabbage (changes requested):

JamesGardiner:
[Remove summary of the questions from WLH summary page at the top (M)](https://trello.com/c/LPmBf6D9/3257-remove-summary-of-the-questions-from-wlh-summary-page-at-the-top-m)

tests/app/views/contexts/summary/test_group.py
JamesGardiner: That's not what I thought we agreed.

There are three scenarios possible:
  - "hide_on_section_summary": true
  - "hide_on_section_summary": false
  - Hide on section summary omitted defaulting to false.

I'll add a further test that covers the third scenario in addition to the above.

data-source/jsonnet/england-wales/household/blocks/who-lives-here/anyone_usually_live_at.jsonnet
JamesGardiner: Can you be more specific? Do you mean the case of the boolean value?

tests/app/views/contexts/summary/test_group.py
JamesGardiner: Scratch that. I've already tested all three scenarios... :man_facepalming:

data-source/jsonnet/england-wales/household/blocks/who-lives-here/anyone_usually_live_at.jsonnet
LJBabbage: Yep

JamesGardiner: JSON standard is booleans as `true` or `false` - same as JS :+1:

rmccar (approved):

Pipfile
```diff
@@ -30,6 +30,7 @@ fakeredis = "*"
 mypy = "*"
 black = "==19.3b0"
 pytest-flask = "*"
+pytest-mock = "*"
```
iwootten: Has this module been introduced as part of tests that have now been removed? 

Looks like its for wrapping mock.patch, which isn't used in any of these tests. I can switch to a standard mock and the tests still pass.

data-source/jsonnet/england-wales/household/blocks/who-lives-here/any_visitors.jsonnet
```diff
@@ -13,6 +13,7 @@ local questionTitle(census_date) = {
 function(census_date) {
   type: 'Question',
   id: 'any-visitors',
+  show_on_section_summary: false,
```
iwootten, line 16: I know this was discussed in the office, but didn't mention it as I felt it should be part of another PR. Has anyone actually requested we should flip this flag or what it should be named?

iwootten (changes requested):

data-source/jsonnet/england-wales/household/blocks/who-lives-here/any_visitors.jsonnet
JamesGardiner: I spoke to @ajmaddaford about it and agreed to flip it

Pipfile
JamesGardiner: Did you uninstall it and run the tests?

iwootten: Yes.

JamesGardiner: Looks like this is a result of using pipenv.

To see the actual dependency on this try:
```
$ pipenv --rm
$ pipenv install --dev
$ make test-unit
```

JamesGardiner: Removing the package from the pipfile and updating doesn't remove it from site-packages, so it's still available to pytest

LJBabbage (approved):
All good, tried it on a couple of other schemas and it worked as expected

Pipfile
iwootten: This still works for me when I've removed the module from the pipfile, lock, completely removed/reinstalled the env and replaced mocker with mock.

What behaviour do the new unit tests use from this module?

JamesGardiner: Using the mocker fixture

JamesGardiner: Using the steps you've outlined results in a failure for me locally

tests/app/views/contexts/summary/test_group.py
```diff
@@ -0,0 +1,132 @@
...
+        Location('section', 'number-question-one'),
+        Location('section', 'number-question-two'),
+    ]
+
+
+def test_group_serialize_hide_on_summary_true(
+    app, mocker, schema, path
+):  # pylint: disable=unused-argument
+
+    answer_store = mocker.MagicMock()
```
MebinAbraham: Can `MagicMock` from `mock` not be used?

MebinAbraham (approved):

tests/app/views/contexts/summary/test_group.py
JamesGardiner: Yes, as the MagicMock seen here _is_ a MagicMock from stdlib. However, if we have a long term goal of moving away from unittest to pytest, we should embrace where possible and sensible pytest plugins that can give cleaner test code. See https://github.com/pytest-dev/pytest-mock/ for examples of where the mocker fixture may help, especially when patching.

iwootten: Does this module actually make the code in the PR any cleaner over and above mock?

I don't think we should be introducing dependencies if we're not actually using the features they provide.

JamesGardiner: Having the mock available as a common fixture across unit tests _is_ using part of the functionality they provide.

Pipfile
iwootten: You don't need to use mock at all, these can be empty List/AnswerStore/dicts?

tests/app/views/contexts/summary/test_group.py
ajmaddaford: I think if we want to change the approach to mocks, we should do it across all tests in a separate pull request. Otherwise it's confusing when looking at the codebase why it's done differently.

MebinAbraham (changes requested):
I don't think we should have both `hide_on_section_summary` and `show_on_section_summary`. We should choose one?

JamesGardiner:
@MebinAbraham I was under the impression `show_on_section_summary` had been chosen?

MebinAbraham:
> @MebinAbraham I was under the impression `show_on_section_summary` had been chosen?

That's fine, but there are still references to `hide_on_section_summary` which makes things confusing.

tests/app/views/contexts/summary/test_group.py
JamesGardiner: Done :+1:

JamesGardiner:
@MebinAbraham - this should be changed in https://github.com/ONSdigital/eq-survey-runner/pull/2348#pullrequestreview-285384242

data-source/jsonnet/northern-ireland/household/blocks/who-lives-here/visitor_list_collector.jsonnet
```diff
@@ -190,5 +190,5 @@ function(census_date) {
     add_link_text: 'Add a visitor',
     empty_list_text: 'There are no visitors',
   },
-  hide_on_section_summary: false,
+  show_on_section_summary: false,
```
MebinAbraham: Should this not be `true`?

app/questionnaire/questionnaire_schema.py
```diff
@@ -163,8 +163,8 @@ def get_visible_list_blocks_for_section(section):
         list_collector_blocks = []
         for group in section['groups']:
             for block in group['blocks']:
-                hidden = block.get('hide_on_section_summary', False)
-                if block['type'] == 'ListCollector' and not hidden:
+                shown = block.get('show_on_section_summary', False)
```
iwootten: The default should be flipped as well.

MebinAbraham (approved):

iwootten (approved):

PR #2343, opened 2019-09-03
Ensure summary context is rendered using current language code

MebinAbraham:
The summary was not rendering context with the current `language_code` in mind. This PR ensures that the current `language_code` is used when rendering the context on the summary.

### How to review 
Ensure:
- When switching languages on the summary, Then the context is rendered in the correct language.

Can be tested using census schemas or `test_placeholder_full`.

### DIFF
--- a/tests/app/questionnaire/test_placeholder_renderer.py
+++ b/tests/app/questionnaire/test_placeholder_renderer.py
@@ -24,24 +24,32 @@ class TestPlaceholderRenderer(AppContextTestCase):
                     'options': [
                         {
                             'label': {
-                                'text': '{person_name} is {age_in_years} years old. Is this correct?',
+                                'text': '{person_name_possessive} age is {age_in_years}. Is this correct?',
                                 'placeholders': [
                                     {
-                                        'placeholder': 'person_name',
+                                        'placeholder': 'person_name_possessive',
                                         'transforms': [
                                             {
-                                                'transform': 'concatenate_list',
                                                 'arguments': {
+                                                    'delimiter': ' ',
                                                     'list_to_concatenate': {
-                                                        'source': 'answers',
                                                         'identifier': [
                                                             'first-name',
                                                             'last-name',
                                                         ],
+                                                        'source': 'answers',
                                                     },
-                                                    'delimiter': ' ',
                                                 },
-                                            }
+                                                'transform': 'concatenate_list',
+                                            },
+                                            {
+                                                'arguments': {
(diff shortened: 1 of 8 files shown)


MebinAbraham:
[Welsh language issues](https://trello.com/c/eDqVO3Mc/3268-welsh-language-issues)

JamesGardiner (approved):

tests/integration/questionnaire/test_questionnaire_placeholders.py
```diff
@@ -19,4 +19,32 @@ def test_title_placeholders_rendered_in_summary(self):
 
         self.post({'confirm-date-of-birth-answer-proxy': 'Yes'})
 
-        self.assertInBody('What is the date of birth for Kevin Bacon?')
+        self.assertInUrl('/summary/')
+        self.assertInBody('What is Kevin Bacon’s date of birth?')
+
+    def test_title_placeholders_rendered_in_summary_using_correct_language_en(self):
```
MebinAbraham:
```suggestion
    def test_title_placeholders_rendered_in_summary_using_correct_language(self):
```

rmccar (approved):

tests/app/questionnaire/test_placeholder_renderer.py
```diff
@@ -145,7 +154,40 @@ def test_renders_json(self):
...
+
+    def test_renders_json_uses_language(self):
+        mock_transform = {
+            'transform': 'calculate_years_difference',
+            'arguments': {
+                'first_date': {
+                    'source': 'answers',
+                    'identifier': 'date-of-birth-answer',
+                },
+                'second_date': {'value': '2019-02-01'},
```
ajmaddaford, line 167: Is this just changing the `second_date` value? If so it would be much clearer if it just changed that value 
```python
json_to_render['answers'][0]['options'][0]['label']['placeholders'][1]['transforms'][0]['arguments']['second_date']['value'] = '2019-02-01'
```

Although it would be better if we replaced the question json with a method that took the date as a parameter and returned the json.

Actually, can the date just be changed in the question json?

ajmaddaford (changes requested):

tests/app/questionnaire/test_placeholder_renderer.py
MebinAbraham: Too slow 😄 . Good shout, yes it can just be changed in the question JSON. Can address in a future PR.

PR #2346, opened 2019-09-05
Previous link on WLH interstitial only displayed when the hub is accessible

petechd:
This fixes previous link functional defect. On the first block it would normally return to the hub if the hub is enabled. Currently, since the hub is not accessible until the who lives here section is complete, the previous link would get redirect back to the page you are on.
It is now only displayed when the hub is enabled and is also accessible.

### How to review 
Load test_hub_complete sections and check that the previous page button doesn't appear on first question page.


petechd:
[Previous link on Who Lives Here Interstitial should only be displayed when the hub is accessible.](https://trello.com/c/dzwskIdl/3267-previous-link-on-who-lives-here-interstitial-should-only-be-displayed-when-the-hub-is-accessible)

tests/functional/spec/previous.spec.js
```diff
@@ -0,0 +1,50 @@
...
+const HubPage = require('../base_pages/hub.page.js');
+const schema = 'test_hub_complete_sections.json';
+
+describe('Choose another section link', function () {
+
+  it('When a user gets to initial question, then the previous location link should not be displayed', function () {
+    return helpers.openQuestionnaire(schema)
+      .then(() => {
+        return browser
+          .getSource().should.not.eventually.contain('Previous link click');
```
MebinAbraham: The previous link is available so we shouldn't need to do `getSource`. Something like 
```js
.isExisting(EmploymentStatusBlockPage.previous()).should.eventually.be.false;
```

Same comment for the other occurrences.

MebinAbraham (changes requested):
Functionality work as expected. Only a minor comment.

tests/functional/spec/previous.spec.js
petechd: Thanks, changed all tests to utilise .previous()

tests/functional/spec/previous.spec.js
```diff
@@ -0,0 +1,52 @@
...
+const EmploymentStatusBlockPage = require('../generated_pages/hub_and_spoke/employment-status.page.js');
+const EmploymentTypePage = require('../generated_pages/hub_and_spoke/employment-type.page.js');
+const HubPage = require('../base_pages/hub.page.js');
+const ProxyPage = require('../generated_pages/hub_and_spoke/proxy.page.js');
+const schema = 'test_hub_complete_sections.json';
+
+describe('Choose another section link', function () {
+
+  it('When a user gets to initial question, then the previous location link should not be displayed', function () {
+    return helpers.openQuestionnaire(schema)
```
JamesGardiner: It's worth doing a beforeEach in the describe function to open the question schema and reduce duplication

JamesGardiner (changes requested):

tests/functional/spec/previous.spec.js
petechd: I forgot about what you told me the other day, sorry for that. It's fixed now.

berroar (approved):
Tested the change, all looks OK

tests/functional/spec/previous.spec.js
```diff
@@ -0,0 +1,42 @@
+const helpers = require('../helpers');
+const EmploymentStatusBlockPage = require('../generated_pages/hub_and_spoke/employment-status.page.js');
+const EmploymentTypePage = require('../generated_pages/hub_and_spoke/employment-type.page.js');
+const HubPage = require('../base_pages/hub.page.js');
+const ProxyPage = require('../generated_pages/hub_and_spoke/proxy.page.js');
+const schema = 'test_hub_complete_sections.json';
```
MebinAbraham: Not sure if this test should be part of the /hub?. And it might be working checking that the previous link is as expected, ie they go where they should?.

MebinAbraham (approved):

JamesGardiner (approved):

tests/functional/spec/previous.spec.js
petechd: Moved it to /hub_and_spoke. Thanks!

PR #2352, opened 2019-09-06
Remove the account services url from the header

berroar:
Remove the account services url/"My Account" link from the header.

Amend the text on session expiry page so that it matches the design [here](https://ons-design-system.netlify.com/patterns/error-status-pages/#timeout).

### How to review 
Open any schema and check that the link no longer appears in the header.

Check to make sure that the text on the session expiry page has been amended to match the design above.


berroar:
[Remove account_services_url from header](https://trello.com/c/SH4t1liW/3286-remove-accountservicesurl-from-header)

rmccar (approved):

templates/layouts/_base.html
```diff
@@ -14,7 +14,6 @@
 } %}
 
 {% include 'layouts/configs/_languages.html' %}
-{% include 'layouts/configs/_service-links.html' %}
```
MebinAbraham, line 17: `_service-links.html` should be removed then?

MebinAbraham (commented):
Not sure, but maybe `My Account` link should be configurable to display or not instead of completely removing?

templates/layouts/_base.html
petechd: We left it because it'll be needed somewhere else in eq... We don't know where yet.

petechd: but if it's included it shows up in header element, when you give it url attribute which is the case.

petechd: It's been removed now, thanks.

package-lock.json
```diff
@@ -0,0 +1,14 @@
+{
```
MebinAbraham: Should be removed?

templates/errors/session-expired.html
```diff
@@ -19,7 +19,9 @@ <h1 {{helpers.track('visible', 'Timeout', 'Timeout expired')}}>{{ _("Your sessio
   {% endcall %}
 
   {% if account_service_url %}
-    <a href="{{account_service_url}}">{{ _("Return to your account") }}</a>
+    <p>{{ _("We have saved your progress and you will need to ") }}
```
MebinAbraham: This should help make the translation of this more meaningful. At the moment, they are split to 3 difference sentence which won't make sense?
```html
    <p>
    {{ _('We have saved your progress and you will need to %(enter_your_unique_code)s to access the survey again.', enter_your_unique_code=('<a href="' ~ account_service_url ~ '">%s</a>' % _('enter your unique code'))) }}
    </p>
```

berroar: Updated this.

package-lock.json
berroar: Removed.

templates/errors/session-expired.html
MebinAbraham: As discussed:
```
    {{ _('We have saved your progress and you will need to <a href="%(account_service_link)s">enter your unique code</a> to access the survey again.', account_service_link=account_service_url) }}
```

JamesGardiner (approved):

PR #2358, opened 2019-09-09
Update translations readme

rmccar:
This PR is to update the documentation around the translation process

### How to review 
Make sure the documentation makes sense and nothing is missing


README.md
```diff
@@ -311,12 +311,29 @@ We use flask-babel to do internationalisation. To extract messages from source a
 make translation-templates
 ```
 
+make translation-templates is a command that runs:
```
iwootten: I don't think this is correct. make translation-templates runs `pybabel extract` and only once to generate a single static messages.pot (which is the same for all languages). Then it uses eq-translations to extract a schema specific .pot template. 

Even so, I think this is going into more detail than is needed for the readme. Probably enough to say 'make translation templates is a command that uses pybabel to extract static messages and eq-translations to extract schema messages' or similar. If a dev is interested in the exact command they can dig into the script itself.

README.md
```diff
@@ -311,12 +311,29 @@ We use flask-babel to do internationalisation. To extract messages from source a
...
 
 These .pot files will then need to be translated. The translation process is documented in Confluence [here](https://collaborate2.ons.gov.uk/confluence/display/SDC/Translation+Process)
 
 Once we have the translated .po files they can be added to the source code and used by the application
 
+If the .po file being added is a new po file it will also need to be added to the translate_schemas.py build script
+
 ---
 
 ### Translating the schemas
```
iwootten, line 325: It should be noted here that translation happens automatically as part of the build process. The following paragraph should be updated to say 'make translate' instead of `scripts/translate_schemas.py`

README.md
```diff
@@ -311,12 +311,29 @@ We use flask-babel to do internationalisation. To extract messages from source a
...
+    To create the gaelic language files
+
+
 This will extract messages and place them in the .pot files ready for translation.
 
 These .pot files will then need to be translated. The translation process is documented in Confluence [here](https://collaborate2.ons.gov.uk/confluence/display/SDC/Translation+Process)
 
 Once we have the translated .po files they can be added to the source code and used by the application
 
+If the .po file being added is a new po file it will also need to be added to the translate_schemas.py build script
```
iwootten: This should probably be part of the 'Translating the schemas' section.

iwootten (changes requested):

iwootten (approved):

JamesGardiner (approved):

PR #2365, opened 2019-09-11
Handle languages for Census schemas

ajmaddaford:
Handles languages for all Census schemas, including NISRA.

The original intention was to add the allowed languages to the schema. This hasn't been done as the language currently needs to be known before loading the schema. Instead there is a hardcoded mapping in code of schema name to allowed languages. We will need to look at how we want to solve this going forward.

`en_US` has been used for Ulstér Scotch as there is no language code in the CLDR (http://cldr.unicode.org) for it and this is required for survey runner to work. This introduces two issues: 
- month names will be in English
- placeholders that have possessive names will have an `'s` added

### How to review
- Run the updated launcher - https://github.com/ONSdigital/eq-questionnaire-launcher/pull/13
- Use the test_language and Census surveys to test language switching.

### Checklist
* [-] New static content marked up for translation
* [-] Newly defined schema content included in eq-translations repo


app/helpers/language_helper.py
```diff
@@ -0,0 +1,55 @@
...
+from app.globals import get_metadata, get_session_store
+from app.questionnaire.questionnaire_schema import DEFAULT_LANGUAGE_CODE
+from app.utilities.schema import get_allowed_languages
+
+
+LANGUAGE_TEXT = {
+    'en': 'English',
+    'cy': 'Cymraeg',
+    'ga': 'Gaeilge',
+    'en_US': 'Ulstér Scotch',
```
MebinAbraham, line 17: Can maybe leave a comment to clarify this?

MebinAbraham (approved):
👍

iwootten:
There's 18 scenarios in the linked spreadsheet, and we cover 4 in the functional test. Should we add some more to at least cover a single scenario for gaelic/ulster scots?

ajmaddaford:
@iwootten I don't think so. The unit tests in test_schema_utils.py that test for language combinations given a schema name and launch language cover this.

app/jinja_filters.py
```diff
@@ -342,14 +329,6 @@ def map_select_config_processor():
     return dict(map_select_config=map_select_config)
 
 
-class LanguageConfig:
```
iwootten, line 345: 👍

iwootten (approved):

PR #2348, opened 2019-09-05
Add 'ListCollectorDrivingQuestion' block type

iwootten:
### What is the context of this PR?

For routes where a qualifying question is used to route directly to an add block within or around a list collector, the summary will no longer generate links correctly when a path around the list collector has been followed. 

This change adds a 'ListCollectorDrivingQuestion' to correctly link back to the driving question when required. Additionally we evaluate the list collectors on the section path to determine if they should be linked to in list items.

### How to review 

Use the test_list_collector_driving_question.json schema to confirm the fix. 

   - Answer "No, no one usually lives here" to the first (driving) question and follow through to summary. Confirm the add link for the section links to the driving question.
   - "Confirm the change/remove links for the section links to the appropriate list collector.
   - Answer 'Yes' to the first (driving) question, add someone and follow through to summary.  Remove the person via the list item link. Return to summary. Confirm the add link for the section links to the list collector add screen.

PR #2367, opened 2019-09-11
Update WLH to use ListCollectorDrivingQuestion

iwootten:
Due to the complexity of the scenarios for the household composition, we can't have a driving question when there are multiple list collectors for the same list. 

To fully resolve links on the summary for the household list, the current household schema should be modified to use variants over a driving question.

Additionally, we should update the visitors list to use a driving question single list collector.

https://github.com/ONSdigital/eq-survey-runner/pull/2348

### How to review 

Go through WLH within census for eng/nir. You should see different variants whether you have entered anyone to the household or not. Answering either no/yes to each additional person within each household/visitor list collector.

There shouldn't be any broken links on the summary.


iwootten:
[Add person to household on section summary doesn't work when "No, no one usually lives here" is selected](https://trello.com/c/o0W9MfTh/3304-add-person-to-household-on-section-summary-doesnt-work-when-no-no-one-usually-lives-here-is-selected)

JamesGardiner (approved):

data-source/jsonnet/england-wales/household/lib/rules.libsonnet
```diff
@@ -1,5 +1,17 @@
 local common_rules = import '../../lib/common_rules.libsonnet';
 
+local listIsEmpty(listName) = {
+  list: listName,
+  condition: 'equals',
+  value: 0,
+};
```
ajmaddaford, line 7: This condition is used in `central_heating.jsonnet` and `renewables.jsonnet`, we should replace  them with `listIsEmpty`.

ajmaddaford (changes requested):

ajmaddaford (approved):

ajmaddaford (changes requested):
The visitors fix hasn't been applied to the NIR survey.

ajmaddaford (approved):
👍

PR #2373, opened 2019-09-15
Return to summary after changing a list collector answer

ajmaddaford:

When changing (add/change/remove) a list collector answer from a section summary, clicking the previous link or the save answers button now returns the user to the summary rather than back to the list collector.

The existing `return_to` url parameter used by a list collector driving question has been renamed `previous`. This is because the previous url and next location url can be different:

- When coming from a driving question, previous location is the driving question, next location is the list collector
- When coming from a summary, previous location and next location are the summary.

### How to review 
- Use the `test_list_collector_section_summary` and `test_list_collector_driving_question` schemas to test the functionality.

### Checklist

* [-] New static content marked up for translation
* [-] Newly defined schema content included in eq-translations repo


app/views/handlers/list_action.py
```diff
@@ -29,17 +29,14 @@ def is_location_valid(self):
...
-
-        return self.parent_location.url()
+        block_id = self._request_args.get('previous') or self._request_args.get(
+            'return_to'
+        )
+        return self._get_location_url(block_id)
 
     def get_next_location_url(self):
-        return self.parent_location.url()
+        block_id = self._request_args.get('return_to')
```
iwootten, line 38: I can't follow what's going on here. Both the next and previous seem to be return_to if given?

iwootten: Isn't there a case where return_to may not be supplied?

iwootten: > I can't follow what's going on here. Both the next and previous seem to be return_to if given?

I guess this is actually what we want though 👍

iwootten (approved):

app/views/handlers/list_action.py
ajmaddaford: Yes. When coming from a summary both the previous link and submitting the change should return the user to the list summary.

MebinAbraham (approved):

PR #2379, opened 2019-09-17
Make gunicorn keep alive setting configurable

ajmaddaford:
We have been seeing 502 errors when load testing the app on GCP. The error `backend_connection_closed_before_data_sent_to_client` appears in the load balancer logs, and nothing in the application logs. 

This looks like it's due to gunicorn disconnecting connections prematurely. This [blog](https://blog.percy.io/tuning-nginx-behind-google-cloud-platform-http-s-load-balancer-305982ddb340) describes a similar problem being experienced with nginx and the [gunicorn docs](http://docs.gunicorn.org/en/stable/settings.html#keepalive) mention setting the keep-alive timeout to a higher value when deployed behind a load balancer. 

This pull request makes the keep-alive timeout setting configurable, and sets the keep-alive timeout (650s) to longer than the load balancer timeout (600s) for helm deploys.

It's also worth noting we had similar issues with Apache and ELBs on AWS - https://github.com/ONSdigital/eq-survey-runner/pull/1912.

### How to review 
- Deploy the new version to GCP and load test.


MebinAbraham (approved):
Tested and seems to be working as expected.

LJBabbage (approved):
Tested at 500 rps for an hour with no 502 or 401 generation, looks like it has done the job

pricem14pc:
[gunicorn keep alive config](https://trello.com/c/UIOmqsq8/3349-gunicorn-keep-alive-config)

PR #2383, opened 2019-09-18
Defects on routing

iwootten:

The default route through a schema is incorrect in certain cases for HH where a comparison is made on a non-mandatory answer. Additionally, the default route for the proxy question for household is incorrect, due being the reverse of what is in individual - 'Are you' vs 'Are you asking on behalf of someone else?'.

### How to review 
Evaluate the routes in the linked trello card and ensure they behave as described.

### DIFF
--- a/data-source/jsonnet/england-wales/household/blocks/individual/proxy.jsonnet
+++ b/data-source/jsonnet/england-wales/household/blocks/individual/proxy.jsonnet
@@ -17,7 +17,7 @@ local placeholders = import '../../../lib/placeholders.libsonnet';
       {
         id: 'proxy-answer',
         mandatory: false,
-        default: 'Yes',
+        default: 'No',
         options: [
           {
             label: 'Yes, I am',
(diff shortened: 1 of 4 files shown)


iwootten:
[Defects on routing](https://trello.com/c/7JGVWpPR/3332-defects-on-routing)

JamesGardiner (changes requested):
Doesn't NI Household term time location also need updating to be mandatory? Also needs a `yarn format`

iwootten:
@JamesGardiner - no, they don't need to be. You'll find that question is marked as mandatory already in the schemas.

JamesGardiner (approved):

iwootten:
Just to clarify - I went through the schemas with Donna on this branch end of play yesterday and she wasn't able to see the issues on the card. Not sure why the term-time-location originally seemed to be non-mandatory.

ajmaddaford (approved):

PR #2387, opened 2019-09-18
Update Irish Household translations

iwootten:

The Irish household translations (Approved on 17th Sept) have yet to be updated in runner causing warnings to be thrown on the household survey during translation. Ulster Scots were updated on the 17th so do not suffer the same issue. Neither individual survey throws warnings.

### How to review 

During household composition, check that the answers for 'Does anybody usually live here?' are correctly translated to gaelic.

### DIFF
--- a/app/translations/ga/census_household_gb_nir_ga.po
+++ b/app/translations/ga/census_household_gb_nir_ga.po
@@ -2,8 +2,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: eq-census-nisra\n"
 "Report-Msgid-Bugs-To: EMAIL@ADDRESS\n"
-"POT-Creation-Date: 2019-09-12 12:18+0100\n"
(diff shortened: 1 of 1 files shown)


rmccar (approved):
That problem with household being split into individual characters is still an issue but other than that looks good

iwootten:
> That problem with household being split into individual characters is still an issue but other than that looks good

Yes, this is because we've not yet updated the templates in crowdin to remove them. Given these letters aren't used they'll just be ignored during transformation in the build.

JamesGardiner (approved):

berroar:
> > That problem with household being split into individual characters is still an issue but other than that looks good
> 
> Yes, this is because we've not yet updated the templates in crowdin to remove them. Given these letters aren't used they'll just be ignored during transformation in the build.

I don't think it will build with the single letter message strings being populated - they need to be empty or it will fall over

PR #2391, opened 2019-09-19
Handle exception during exception handling

MebinAbraham:
Currently, if an exception is raised during the handling of an exception, you get shown a white `Internal Server Error` page and does not render our app error page. This happened in blacklodge because the state of the questionnaire has changed and the questionnaire state is used when actually logging errors.

This PR renders the app error page if an exception occurs during the current exception handling.

### How to review 
1. Create an exception that creates another exception and ensure the app error page is rendered.

If you want to test it quickly. you could make this key invalid:
https://github.com/ONSdigital/eq-survey-runner/blob/v3/app/data_model/progress_store.py#L61

Then launch a survey and try to proceed and ensure you see the app error page is rendered.


JamesGardiner (approved):

app/routes/errors.py
```diff
@@ -34,8 +33,18 @@ def forbidden(error=None):
 @errors_blueprint.app_errorhandler(SubmissionFailedException)
 @errors_blueprint.app_errorhandler(Exception)
 def internal_server_error(error=None):
-    log_exception(error, 500)
-    return _render_error_page(500)
+    try:
+        log_exception(error, 500)
+        return _render_error_page(500)
+    except Exception as ex:  # pylint:disable-broad-except
```
JamesGardiner:
```suggestion
    except Exception as ex:  # pylint:disable=broad-except
```

JamesGardiner (changes requested):

JamesGardiner (approved):

app/routes/errors.py
```diff
@@ -58,8 +67,10 @@ def log_exception(error, status_code):
     )
 
 
-def _render_error_page(status_code, template=None):
-    handle_language()
+def _render_error_page(status_code, template=None, store_language=True):
+    if store_language:
+        handle_language()
```
ajmaddaford: The `ua` user agent template variable is no longer used by any of the error templates and should be removed. Once that's removed there is no need for the `store_language` parameter as we can call `render_template` directly from the new exception handling code.

app/routes/errors.py
```diff
@@ -34,8 +33,18 @@ def forbidden(error=None):
...
-    log_exception(error, 500)
-    return _render_error_page(500)
+    try:
+        log_exception(error, 500)
+        return _render_error_page(500)
+    except Exception as ex:  # pylint:disable=broad-except
+        logger.error(
+            'an error has occurred',
+            initial_error=error,
+            final_error=str(ex),
```
ajmaddaford: This will lead to stack traces not being formatted correctly in JSON log lines. For other exceptions we use `exc_info` - http://www.structlog.org/en/stable/api.html#structlog.processors.format_exc_info. 

When `exc_info` gets rendered into the `exception` key we do some processing on it in `configure_logging`  to format appropriately for JSON logging - we convert double quotes to single quotes and each line of the exception is converted into an array item.

Replacing `initial_error` and `final_error` with `exc_info=True` seems to give a stacktrace that covers both the original exception and the exception while handling it.

ajmaddaford (changes requested):

app/routes/errors.py
MebinAbraham: Have addressed this. Thanks.

app/routes/errors.py
```diff
@@ -34,8 +32,14 @@ def forbidden(error=None):
...
 @errors_blueprint.app_errorhandler(Exception)
 def internal_server_error(error=None):
-    log_exception(error, 500)
-    return _render_error_page(500)
+    try:
+        log_exception(error, 500)
+        return _render_error_page(500)
+    except Exception:  # pylint:disable=broad-except
+        logger.error(
+            'an error has occurred', exc_info=True, url=request.url, status_code=500
```
ajmaddaford: This error message should be different to the other generic one, maybe something like "error when rendering 500 error".

ajmaddaford (changes requested):
Looks good, just one small change to the error message.

app/routes/errors.py
MebinAbraham: 👌

ajmaddaford (approved):

app/routes/errors.py
bitdivision: For future reference, `logger.exception()` should achieve the same thing. Handy method for exception handlers :)

Issue #2395, opened 2019-09-21
Log lines contain context from a previous request


ajmaddaford:
### Expected behaviour
Log lines should only contain context from the current request.

### Actual behaviour
Log lines under some circumstances can have the context of a previous request.

### Steps to reproduce the behaviour
- Run the application via Gunicorn with 1 worker and a keep-alive timeout of 20s - `gunicorn -w 1 --keep-alive 20 --worker-class gevent -b 0.0.0.0:5000 application:application`
- Start a survey
- Remove the session cookie
- Submit the page
- There should be a log line with "The CSRF session token is missing". This log line also includes other context from the previous request - note that it has a `ce_id` which would only be possible with a valid session, and the `request_id` is the `request_id` of the previous request.

### Technical information
Needs Gunicorn with a long keep-alive timeout. A long keep-alive timeout is necessary to work properly when behind a load balancer in GCP (https://github.com/ONSdigital/eq-survey-runner/pull/2379)

PR #2396, opened 2019-09-21
Fix per request logging issue with third party libraries

ajmaddaford:

Fixes an issue (https://github.com/ONSdigital/eq-survey-runner/issues/2395) where log lines output from the before request hooks of third party libraries could have context from a previous request.

The structlog documentation has a [warning](http://www.structlog.org/en/stable/thread-local.html#downsides-caveats) about shared state :
> If you can’t rule out that your application re-uses threads, you must remember to initialize your thread local context at the start of each request using new() (instead of bind()). Otherwise you may start a new request with the context still filled with data from the request before.

For our application, this should be dealt with when we do `logger.new(request_id=request_id)` in `@application.before_request`.  This creates a new logger context for each new request. The issue arises when other Flask libraries add before request hooks that run before ours. This means that any log lines they produce in their before request code will have the context of the previous logger.

We are actively using three libraries that could be affected by this (flask_login, flask_talisman and flask_wtf). Depending on environment settings, three other libraries could also be affected (flask_debugtoolbar, newrelic and aws_xray_sdk).

To fix this we now add our `before_request` hook before the other Flask libraries initialise. This means that any log lines those libraries output will have the new logger context for that request.

### How to review 
Follow the steps in the [issue](https://github.com/ONSdigital/eq-survey-runner/issues/2395) and check that log context is not shared between requests.

---
Fixes https://github.com/ONSdigital/eq-survey-runner/issues/2395

### DIFF
--- a/app/setup.py
+++ b/app/setup.py
@@ -132,6 +132,19 @@ def create_app(setting_overrides=None):  # noqa: C901  pylint: disable=too-compl
             version=application.config['EQ_APPLICATION_VERSION'],
         )
 
+    # IMPORTANT: This must be initialised *before* any other Flask plugins that add
+    # before_request hooks. Otherwise any logging by the plugin in their before
+    # request will use the logger context of the previous request.
+    @application.before_request
+    def before_request():  # pylint: disable=unused-variable
+
+        # While True the session lives for permanent_session_lifetime seconds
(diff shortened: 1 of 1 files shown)


MebinAbraham (approved):
Working as expected. 👍

JamesGardiner (approved):

PR #2406, opened 2019-09-24
Update gunicorn workers to 7

LJBabbage:
In performance tests it was found that 7 Gunicorns was more efficient than 9, which also ties in the with the idea that you should have 2 times + 1 in relation to CPU used by your application. We have 3 CPU for each runner app, so that also makes 7

### How to review 
Build through Concorse with this branch and make sure everything works

### DIFF
--- a/k8s/helm/templates/deployment.yaml
+++ b/k8s/helm/templates/deployment.yaml
@@ -76,7 +76,7 @@ spec:
             - name: EQ_SUBMISSION_BACKEND
               value: gcs
             - name: GUNICORN_WORKERS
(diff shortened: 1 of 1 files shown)


MebinAbraham (approved):

pricem14pc (approved):

pricem14pc:
[Decrease gunicorn workers to 7 from 9](https://trello.com/c/c858rLBC/3364-decrease-gunicorn-workers-to-7-from-9)

JamesGardiner (approved):

PR #2407, opened 2019-09-24
Ability to override helm deployment values

LJBabbage:
Added the ability to override helm deployment values.

### How to review 
Go to the concourse PR https://github.com/ONSdigital/census-eq-concourse/pull/27 and use this as your branch for eq-survey-runner

### DIFF
--- a/k8s/deploy_app.sh
+++ b/k8s/deploy_app.sh
@@ -9,6 +9,12 @@ fi
 
 DOCKER_REGISTRY="${DOCKER_REGISTRY:-eu.gcr.io/census-eq-ci}"
 IMAGE_TAG="${IMAGE_TAG:-latest}"
+REQUESTED_CPU_PER_POD="${REQUESTED_CPU_PER_POD:-3}"
+ROLLING_UPDATE_MAX_UNAVAILABLE="${ROLLING_UPDATE_MAX_UNAVAILABLE:-1}"
+ROLLING_UPDATE_MAX_SURGE="${ROLLING_UPDATE_MAX_SURGE:-1}"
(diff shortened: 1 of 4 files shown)


k8s/helm/values.yaml
```diff
@@ -26,6 +26,6 @@ resources:
     cpu: "3"
 
 autoscaler:
-    minReplicas: 3
-    maxReplicas: 10
+    minReplicas: ""
+    maxReplicas: ""
     targetCPUUtilizationPercentage: 50
```
ajmaddaford: It would be useful to make this configurable as well.

pricem14pc: We thought that too but maybe in another PR.

k8s/deploy_app.sh
```diff
@@ -9,6 +9,12 @@ fi
 
 DOCKER_REGISTRY="${DOCKER_REGISTRY:-eu.gcr.io/census-eq-ci}"
 IMAGE_TAG="${IMAGE_TAG:-latest}"
+REQUESTED_CPU="${REQUESTED_CPU:-3}"
+MAX_UNAVAILABLE="${MAX_UNAVAILABLE:-1}"
+MAX_SURGE="${MAX_SURGE:-1}"
```
ajmaddaford: It's not clear these are for rolling updates, it would be better to prefix the two with `ROLLING_UPDATE_`.

ajmaddaford (changes requested):

ajmaddaford (approved):

pricem14pc:
[Cluster min/max size](https://trello.com/c/RHIG9Sto/3369-cluster-min-max-size)

pricem14pc (approved):

PR #2420, opened 2019-10-02
Switch from persistent to session cookies

ajmaddaford:

After investigation into production issues it has been identified that some browsers (predominantly Edge) seem to drop our session cookie randomly. Although we still don't know why this happens (debugging has shown that a valid cookie is sent on the request previous to the one that errors), changing from a permanent cookie (one with an expiry time set) to a session cookie (one that expires when the browser is closed) seems to fix the issue.

Arguably the cookie should always have been a session one. The server side session times out after the timeout period (default is 45 minutes) and keeping the cookie client side means we will still have context like the theme or an account service url that are useful when rendering error screens. The cookie was set to expire prior to us having server side session expiry.

### How to review 
- Check that launching , signing out and resuming work as expected
- Use the `census-rehearsal` branch version of runner to start a survey, switch to this branch and confirm that there is no interruption to the journey.
- Test on a range of browsers

### Checklist

### DIFF
--- a/templates/errors/session-expired.html
+++ b/templates/errors/session-expired.html
@@ -16,14 +16,14 @@
     "classes": "u-mb-s"
   }) %}
     <p>{{ _("To help protect your information we have signed you out") }}.</p>
-    <p>You will need to <a href="https://census.gov.uk/start/">enter your unique access code</a> to access the census again.</p>
+    {% if account_service_url %}
+      <p>{{ _('We have saved your progress and you will need to <a href="%(account_service_link)s">enter your unique code</a> to access the survey again.', account_service_link=account_service_url) }}</p>
+    {% else %}
+      <p>{{ _('We have saved your progress and you will need to <a href="%(account_service_link)s">enter your unique code</a> to access the survey again.', account_service_link='https://census.gov.uk/start/') }}</p>
+    {% endif %}
     {% if using_edge %}
-    <p>If you have been timed out unexpectedly on more than one occasion, please try using a different browser. If you’re still having problems completing the census, <a href="https://census.gov.uk/contact-us/">contact us</a>.</p>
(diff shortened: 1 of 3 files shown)


templates/errors/session-expired.html
```diff
@@ -16,14 +16,14 @@ <h1 {{helpers.track('visible', 'Timeout', 'Timeout expired')}}>{{ _("Your sessio
...
     <p>{{ _("To help protect your information we have signed you out") }}.</p>
-    <p>You will need to <a href="https://census.gov.uk/start/">enter your unique access code</a> to access the census again.</p>
+    {% if account_service_url %}
+      <p>{{ _('We have saved your progress and you will need to <a href="%(account_service_link)s">enter your unique code</a> to access the survey again.', account_service_link=account_service_url) }}</p>
+    {% else %}
+      <p>{{ _('We have saved your progress and you will need to <a href="%(account_service_link)s">enter your unique code</a> to access the survey again.', account_service_link='https://census.gov.uk/start/') }}</p>
+    {% endif %}
     {% if using_edge %}
-    <p>If you have been timed out unexpectedly on more than one occasion, please try using a different browser. If you’re still having problems completing the census, <a href="https://census.gov.uk/contact-us/">contact us</a>.</p>
+      <p>If you have been timed out unexpectedly on more than one occasion, please try using a different browser. If you’re still having problems completing the census, <a href="https://census.gov.uk/contact-us/">contact us</a>.</p>
```
MebinAbraham, line 25: Worth wrapping these in `_()` now?

MebinAbraham (approved):
Tested for in-flight and new sessions when deploying this version. Also tested a rollback from this version to the previous version and all seems to work as expected. 👍

templates/errors/session-expired.html
ajmaddaford: I decided not as I'm hoping we will be removing this message in the next release.

rmccar (approved):

iwootten (approved):
Followed the instructions, works for me. 👍

PR #2421, opened 2019-10-03
Change summary panel

rmccar:
Update the styling and wording on the panels that are shown on summaries.

### How to review 
Check that all summaries in the household schema have blue panels with the correct wording from the Trello card


iwootten (approved):
👍

LJBabbage (approved):
All section summaries changed to the grey version when run

templates/summary.html
```diff
@@ -21,11 +21,10 @@ <h1>{{ _("Check your answers and submit") }}</h1>
 
     <h1>{{content.summary.title}}</h1>
       {% call onsPanel({
-        "type": "success",
```
iwootten, line 24: You shouldn't need to define the type as the default is information?

rmccar: Cool, sorted that. I tested it and it works for me

iwootten (changes requested):
Blocking this until someone senior makes the call whether we're allowed to fix this issue.

iwootten (approved):
Approving this based on the panel changes.

PR #2423, opened 2019-10-04
Add printlistener for gtm

martyncolmer:
The base html has been updated to include a print listener for google tag manager. 
This was requested by Andy Pitt. 

### How to review 
Run with docker-compose up.
Open a survey from Go Launcher.
Open the browser developer tool > console.  
Type dataLayer to view the dataLayer content
Print the webpage and re-review the dataLayer to see the print intent.
Repeat on a different page

###Expected outcome
For a print attempt there will be an entry in the data layer containing the url of the page.
Only one entry in the data layer should be recorded per print intent per page.


app/templates/layouts/_base.html
```diff
@@ -133,4 +133,26 @@
...
+          // matchMedia Support
+          if (window.matchMedia) {
+              var mediaQueryList = window.matchMedia('print');
+              mediaQueryList.addListener(function(mql) {
+                  if (mql.matches)
+                      onbeforePrinting();
+              });
+          }
+          window.onbeforeprint = onbeforePrinting;
+        }());
```
MebinAbraham: This should be moved into a file inside the app assets, so `app/assets/js/app/modules/`. Otherwise, this would not get transpiled by Babel. Once done you would also be able to use ES6 syntax like `let` instead of `var` which Babel will transpile when running `yarn compile`.

app/templates/layouts/_base.html
```diff
@@ -133,4 +133,26 @@
       document.write("<script async nonce='{{ csp_nonce() }}' src='http://HOST:5075/browser-sync/browser-sync-client.js?v=2.24.4'><\/script>".replace("HOST", location.hostname));
     //]]></script>
     {% endif %}
+
+    <script nonce="{{ csp_nonce() }}" >
+        (function() {
+          var printed = 0;
```
MebinAbraham: I think it would be clearer if this was a boolean instead of an integer.

MebinAbraham (changes requested):

app/templates/layouts/_base.html
martyncolmer: Now added to JS module

app/templates/layouts/_base.html
martyncolmer: changed to boolean

app/assets/js/app/modules/printlistener.js
```diff
@@ -0,0 +1,22 @@
...
+  const onafterPrinting = () => {
+    if (!printed) {
+      window.dataLayer.push({ 'event': 'onAfterPrint', 'currentURL': location.pathname })
+      printed = true
+    }
+  }
+  if (window.matchMedia) {
+    let mediaQueryList = window.matchMedia('print')
+    mediaQueryList.addListener(function(mql) {
+      if (mql.matches) {
```
MebinAbraham: Working as expected, however, the `onafterprint` with `mql` is when it is `if (!mql.matches) {`. <- this is how the current analytics event is done. 
https://github.com/ONSdigital/sdc-global-design-patterns/blob/fb4d79d15ffca96674c8c06e2245ea93ba5531a7/assets/js/analytics.js#L48-L64

```js
(function() {
    var beforePrint = function() {
        console.log('Functionality to run before printing.');
    };
    var afterPrint = function() {
        console.log('Functionality to run after printing');
    };

    if (window.matchMedia) {
        var mediaQueryList = window.matchMedia('print');
        mediaQueryList.addListener(function(mql) {
            if (mql.matches) {
                beforePrint();
            } else {
                afterPrint();
            }
        });
    }

    window.onbeforeprint = beforePrint;
    window.onafterprint = afterPrint;
}());
```
https://www.tjvantoll.com/2012/06/15/detecting-print-requests-with-javascript/

MebinAbraham: Not a blocker but `addListener ` is also deprecated, ideally, we should use https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList/onchange.

martyncolmer: Suggested changes now implemented

MebinAbraham (commented):
Have tried a print however the event does not seem to get pushed to the `dataLayer`?

MebinAbraham (approved):

SamGodwin2 (approved):

PR #2432, opened 2019-10-08
Gcp log formatting

JamesGardiner:
Logging now goes to stderr or stdout depending on what severity is logged.
When an X-Cloud-Trace-Header is present in a request, the span and trace ids will be present in each log line of the request.

### How to review 
Deploy to your local environment. Check span and trace in the log lines are empty and that log levels correctly go to stderr and stdout i.e. run under ` flask run 2> /dev/null` and check no stderr log messages are visible, then run `flask run > /dev/null` and check no stdout messages are visible.

Deploy to a GCP environment. Run through a survey, then check that requests have the correct span and trace ids in the logs. Check that GCP correctly categorises the severity of the log message.

### DIFF
--- a/.development.env
+++ b/.development.env
@@ -1,4 +1,5 @@
 FLASK_APP=application.py
+FLASK_ENV=development
 EQ_DEVELOPER_LOGGING=True
 EQ_SUBMITTED_RESPONSES_TABLE_NAME=dev-submitted-responses
 EQ_QUESTIONNAIRE_STATE_TABLE_NAME=dev-questionnaire-state
--- a/.development.env
+++ b/.development.env
@@ -1,5 +1,4 @@
 FLASK_APP=application.py
-FLASK_ENV=development
 EQ_DEVELOPER_LOGGING=True
 EQ_SUBMITTED_RESPONSES_TABLE_NAME=dev-submitted-responses
 EQ_QUESTIONNAIRE_STATE_TABLE_NAME=dev-questionnaire-state
--- a/.development.env
+++ b/.development.env
@@ -1,6 +1,5 @@
 FLASK_APP=application.py
 FLASK_ENV=development
-EQ_DEVELOPER_LOGGING=True
 EQ_SUBMITTED_RESPONSES_TABLE_NAME=dev-submitted-responses
 EQ_QUESTIONNAIRE_STATE_TABLE_NAME=dev-questionnaire-state
 EQ_SESSION_TABLE_NAME=dev-eq-session
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -18,10 +18,10 @@
     "default": {
         "asn1crypto": {
             "hashes": [
-                "sha256:2f1adbb7546ed199e3c90ef23ec95c5cf3585bac7d11fb7eb562a3fe89c64e87",
-                "sha256:9d5c20441baf0cb60a4ac34cc447c6c189024b6b4c6cd7877034f4965c464e49"
+                "sha256:0b199f211ae690df3db4fd6c1c4ff976497fb1da689193e368eedbadc53d9292",
+                "sha256:bca90060bd995c3f62c4433168eab407e44bdbdb567b3f3a396a676c1a4c4a3f"
             ],
-            "version": "==0.24.0"
+            "version": "==1.0.1"
         },
         "babel": {
             "hashes": [
@@ -39,18 +39,18 @@
         },
         "boto3": {
             "hashes": [
-                "sha256:9d588e23a8f6bbb1c71790d70190f67bd644aaf6a6010ffed40d63c1f38d6b9d",
-                "sha256:d695bb99dec048800fcba9657182d2db3a4cae5e6c3625a5b3a5cd785fb2d6d6"
+                "sha256:404acbecef8f4912f18312fcfaffe7eba7f10b3b7adf7853bdba59cdf2275ebb",
+                "sha256:c6e5a7e4548ce7586c354ff633f2a66ba3c471d15a8ae6a30f873122ab04e1cf"
             ],
             "index": "pypi",
-            "version": "==1.9.224"
+            "version": "==1.9.243"
         },
         "botocore": {
             "hashes": [
-                "sha256:197f118419b81c8e15db03b92220f9fa864de2f062d962dbbe1a30d42d080d50",
-                "sha256:e2009cc78a99d284aa40d8bdc6954a47f2afeeecd33492ba8ef9337b97d0044c"
+                "sha256:397585a7881230274afb8d1877ef69a661b0a311745cd324f14a052fb2a2863a",
+                "sha256:4496f8da89cb496462a831897ad248e13e431d9fa7e41e06d426fd6658ab6e59"
             ],
-            "version": "==1.12.224"
+            "version": "==1.12.243"
         },
         "cachetools": {
             "hashes": [
@@ -61,10 +61,10 @@
         },
         "certifi": {
             "hashes": [
-                "sha256:046832c04d4e752f37383b628bc601a7ea7211496b4638f6514d0e5b9acc4939",
-                "sha256:945e3ba63a0b9f577b1395204e13c3a231f9bc0223888be653286534e5873695"
+                "sha256:e4f3620cfea4f83eedc95b24abd9cd56f3c4b146dd0177e83a21b4eb49e21e50",
+                "sha256:fd7c7c74727ddcf00e9acd26bba8da604ffec95bf1c2144e67aff7a8b50e6cef"
             ],
-            "version": "==2019.6.16"
+            "version": "==2019.9.11"
         },
         "cffi": {
             "hashes": [
@@ -287,18 +287,18 @@
         },
         "google-cloud-storage": {
             "hashes": [
-                "sha256:1e1fce811753bd2570db3d468437d44052274641c7f637c50e1c2df16497b9f7",
-                "sha256:21f94c073eb9be34ca545c4eef28ae66947386687d99cd3a82400bf1504cfbd8"
+                "sha256:13a6a820311662eb91a99810568c2bca5ddc7e44e2163fed4cb3f4d47da132cf",
+                "sha256:2e7e2435978bda1c209b70a9a00b8cbc53c3b00d6f09eb2c991ebba857babf24"
             ],
             "index": "pypi",
-            "version": "==1.19.0"
+            "version": "==1.20.0"
         },
         "google-resumable-media": {
             "hashes": [
-                "sha256:46ee131e55d16e350cf9dc7de6fc0653b55314e8645e2deb0633394adcd7e9c0",
-                "sha256:d14dff24e0c6cab676f9eff0049c42b8cdae78dff723be095ef3af2c4ba06dfa"
+                "sha256:5fd2e641f477e50be925a55bcfdf0b0cb97c2b92aacd7b15c1d339f70d55c1c7",
+                "sha256:cdeb8fbb3551a665db921023603af2f0d6ac59ad8b48259cb510b8799505775f"
             ],
-            "version": "==0.4.0"
+            "version": "==0.4.1"
         },
         "googleapis-common-protos": {
             "hashes": [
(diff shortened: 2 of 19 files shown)


app/setup.py
```diff
@@ -136,8 +137,14 @@ def create_app(setting_overrides=None):  # noqa: C901  pylint: disable=too-compl
     # request will use the logger context of the previous request.
     @application.before_request
     def before_request():  # pylint: disable=unused-variable
+        span, trace = get_span_and_trace(flask_request)
```
ajmaddaford: `flask_request.headers` could be passed here rather than the full request object

app/setup.py
```diff
@@ -136,8 +137,14 @@ def create_app(setting_overrides=None):  # noqa: C901  pylint: disable=too-compl
...
+
+        # While True the session lives for permanent_session_lifetime seconds
+        # Needed to be able to set the client-side cookie expiration
+        cookie_session.permanent = True
+
         request_id = str(uuid4())
-        logger.new(request_id=request_id)
+        logger.new(request_id=request_id, span=span, trace=trace)
```
ajmaddaford: `span` and `trace` should only be added to the logger if they are set

app/setup.py
```diff
@@ -136,8 +137,14 @@ def create_app(setting_overrides=None):  # noqa: C901  pylint: disable=too-compl
     # request will use the logger context of the previous request.
     @application.before_request
     def before_request():  # pylint: disable=unused-variable
+        span, trace = get_span_and_trace(flask_request)
+
+        # While True the session lives for permanent_session_lifetime seconds
+        # Needed to be able to set the client-side cookie expiration
+        cookie_session.permanent = True
```
ajmaddaford: Is this branch up to date? This line was removed in https://github.com/ONSdigital/eq-survey-runner/pull/2420

app/setup.py
```diff
@@ -232,19 +239,6 @@ def response_minify(response):  # pylint: disable=unused-variable
...
-        # We're using the stringified version of the Flask session to get a rough
-        # length for the cookie. The real length won't be known yet as Flask
-        # serialises and adds the cookie header after this method is called.
-        logger.info(
-            'response',
-            status_code=response.status_code,
-            session_cookie_content_length=len(str(cookie_session)),
-            session_modified=cookie_session.modified,
-        )
-        return response
```
ajmaddaford, line 246: Are we sure we want to remove this logging?

Pipfile.lock
```diff
@@ -18,10 +18,10 @@
     "default": {
         "asn1crypto": {
             "hashes": [
-                "sha256:2f1adbb7546ed199e3c90ef23ec95c5cf3585bac7d11fb7eb562a3fe89c64e87",
```
ajmaddaford: Why have the Python modules been updated?

tests/app/conftest.py
```diff
@@ -0,0 +1,26 @@
...
+@fixture
+def app():
+    setting_overrides = {'LOGIN_DISABLED': True}
+    the_app = create_app(setting_overrides=setting_overrides)
+
+    return the_app
+
+
+@fixture
+def header():
```
ajmaddaford: This fixture should be named more clearly, possibly `cloud_trace_header`? 
Also, I'm not sure this should be a fixture as it's only used in one test.

application.py
```diff
@@ -15,24 +16,33 @@
...
-        'ERROR': logging.ERROR,
-        'WARNING': logging.WARNING,
-        'INFO': logging.INFO,
-        'DEBUG': logging.DEBUG,
-    }
-    handler = logging.StreamHandler()
+
+    info_handler = logging.StreamHandler(sys.stdout)
+    info_handler.setLevel(logging.getLevelName(EQ_LOG_LEVEL))
+    info_handler.addFilter(_LogLevelFilter(logging.WARNING))
```
ajmaddaford: This can be done without a class:
```python
info_handler.addFilter(lambda record: record.levelno <= logging.WARNING)
```

.development.env
```diff
@@ -1,5 +1,4 @@
 FLASK_APP=application.py
-FLASK_ENV=development
```
ajmaddaford, line 2: How come this has been removed?

ajmaddaford (changes requested):
Have you run any load tests with the logging changes?

ajmaddaford:
The exception traces aren't being formatted correctly (the exception should be split into a list in an `exception` element:
```
[error    ] an error has occurred          created=2019-10-09T09:01:28.192989Z request_id=8ffcfd60-3c71-4c58-9e65-b6a340fd156d service=eq-survey-runner span=  status_code=500 trace=  url=http://localhost:5000/flush
  File "/Users/maddaa/.local/share/virtualenvs/eq-survey-runner-lrST2HPR/lib/python3.7/site-packages/flask/app.py", line 1949, in full_dispatch_request
    rv = self.dispatch_request()
  File "/Users/maddaa/.local/share/virtualenvs/eq-survey-runner-lrST2HPR/lib/python3.7/site-packages/flask/app.py", line 1925, in dispatch_request
    self.raise_routing_exception(req)
  File "/Users/maddaa/.local/share/virtualenvs/eq-survey-runner-lrST2HPR/lib/python3.7/site-packages/flask/app.py", line 1907, in raise_routing_exception
    raise request.routing_exception
  File "/Users/maddaa/.local/share/virtualenvs/eq-survey-runner-lrST2HPR/lib/python3.7/site-packages/flask/ctx.py", line 350, in match_request
    result = self.url_adapter.match(return_rule=True)
  File "/Users/maddaa/.local/share/virtualenvs/eq-survey-runner-lrST2HPR/lib/python3.7/site-packages/werkzeug/routing.py", line 1798, in match
    raise MethodNotAllowed(valid_methods=list(have_match_for))
werkzeug.exceptions.MethodNotAllowed: 405 Method Not Allowed: The method is not allowed for the requested URL.
```

.development.env
JamesGardiner: My mistake - took it out for local dev to check logging under different scenarios

app/setup.py
JamesGardiner: Can leave it in if it's preferred - I was under the impression it wasn't being used anymore

app/setup.py
JamesGardiner: Removed now

Pipfile.lock
JamesGardiner: Will revert

JamesGardiner:
> The exception traces aren't being formatted correctly (the exception should be split into a list in an `exception` element:
> 
> ```
> [error    ] an error has occurred          created=2019-10-09T09:01:28.192989Z request_id=8ffcfd60-3c71-4c58-9e65-b6a340fd156d service=eq-survey-runner span=  status_code=500 trace=  url=http://localhost:5000/flush
> Traceback (most recent call last):
>   File "/Users/maddaa/.local/share/virtualenvs/eq-survey-runner-lrST2HPR/lib/python3.7/site-packages/flask/app.py", line 1949, in full_dispatch_request
>     rv = self.dispatch_request()
>   File "/Users/maddaa/.local/share/virtualenvs/eq-survey-runner-lrST2HPR/lib/python3.7/site-packages/flask/app.py", line 1925, in dispatch_request
>     self.raise_routing_exception(req)
>   File "/Users/maddaa/.local/share/virtualenvs/eq-survey-runner-lrST2HPR/lib/python3.7/site-packages/flask/app.py", line 1907, in raise_routing_exception
>     raise request.routing_exception
>   File "/Users/maddaa/.local/share/virtualenvs/eq-survey-runner-lrST2HPR/lib/python3.7/site-packages/flask/ctx.py", line 350, in match_request
>     result = self.url_adapter.match(return_rule=True)
>   File "/Users/maddaa/.local/share/virtualenvs/eq-survey-runner-lrST2HPR/lib/python3.7/site-packages/werkzeug/routing.py", line 1798, in match
>     raise MethodNotAllowed(valid_methods=list(have_match_for))
> werkzeug.exceptions.MethodNotAllowed: 405 Method Not Allowed: The method is not allowed for the requested URL.
> ```

As discussed IRL - this was because the `EQ_DEVELOPER_LOGGING` env var returns the original event dictionary without modifying the exception value.

app/setup.py
ajmaddaford: The cookie length can go, as it's not the real length anyway. I think the response status code and whether the session has been modified is useful.

tests/app/test_setup.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class AttrDict(dict):
+    def __init__(self, *args, **kwargs):
+        super(AttrDict, self).__init__(*args, **kwargs)
+        self.__dict__ = self
+
+
+def cloud_trace_header():
+    header = AttrDict()
```
ajmaddaford: I don't think `AttrDict` is needed anymore.

application.py
```diff
@@ -16,23 +17,24 @@
...
-        'CRITICAL': logging.CRITICAL,
-        'ERROR': logging.ERROR,
-        'WARNING': logging.WARNING,
-        'INFO': logging.INFO,
-        'DEBUG': logging.DEBUG,
-    }
-    handler = logging.StreamHandler()
+
+    info_handler = logging.StreamHandler(sys.stdout)
+    info_handler.setLevel(logging.getLevelName(EQ_LOG_LEVEL))
```
ajmaddaford: I think we should remove `EQ_LOG_LEVEL` as we use `INFO` everywhere, and turn debug logging on based on the application `DEBUG` mode.

ajmaddaford (changes requested):

.development.env
ajmaddaford: Is it going to go back in?

app/authentication/authenticator.py
```diff
@@ -22,19 +22,19 @@
 
 @login_manager.user_loader
 def user_loader(user_id):
-    logger.debug('loading user', user_id=user_id)
+    logger.info('loading user', user_id=user_id)
```
ajmaddaford: We should either leave these as debug, or review the usefulness of these log statements rather that just change them all to info. It would be better to have logging on the un-happy path so we log when something un-expected happens.

Also, by changing all of these statements to debug we will be adding a lot of extra logging to every request in production - from 2 to 8 log lines.

application.py
```diff
@@ -10,36 +11,36 @@
...
-        'DEBUG': logging.DEBUG,
-    }
-    handler = logging.StreamHandler()
+    log_level = logging.INFO
+    if os.getenv('FLASK_ENV') == 'development':
+        log_level = logging.DEBUG
+
+    info_handler = logging.StreamHandler(sys.stdout)
+    info_handler.setLevel(log_level)
+    info_handler.addFilter(lambda record: record.levelno <= logging.WARNING)
```
ajmaddaford: Given the `info_handler` will handle `WARNING` and `DEBUG` it might be clearer to name the handlers `log_handler` and `error_log_handler`

application.py
```diff
@@ -10,36 +11,36 @@
...
+
     logging.basicConfig(
-        level=levels[EQ_LOG_LEVEL], format=log_format, handlers=[handler]
+        level=log_level, format='%(message)s', handlers=[error_handler, info_handler]
     )
 
     # Set werkzeug logging level
     werkzeug_logger = logging.getLogger('werkzeug')
-    werkzeug_logger.setLevel(level=levels[EQ_WERKZEUG_LOG_LEVEL])
+    werkzeug_logger.setLevel(level=logging.getLevelName(EQ_WERKZEUG_LOG_LEVEL))
```
ajmaddaford: I think we can remove `EQ_WERKZEUG_LOG_LEVEL` and just set this as `INFO` as it's never overridden.

application.py
```diff
@@ -10,36 +11,36 @@
...
 
     def parse_exception(_, __, event_dict):
         if EQ_DEVELOPER_LOGGING:
```
ajmaddaford: We could remove `EQ_DEVELOPER_LOGGING` and use `FLASK_ENV` to set this as we do for the log level.

ajmaddaford (changes requested):

ajmaddaford:
Also, the Readme needs to be updated with the changes to the environment variables.

app/authentication/authenticator.py
JamesGardiner: We don't (or shouldn't be) manually parsing log lines given we use the GCP logging service. This negates the 'can't see the wood for the trees' problem. Even when we log the unhappy path, it can still be useful to see what steps were taken up to that point.

I couldn't say whether increasing from 2 to 8 is a lot of extra logging without running some performance tests (which I'll carry out and post results here - it would be interesting to see what impact the extra logs have in any case).

JamesGardiner:
> Also, the Readme needs to be updated with the changes to the environment variables.

Done

app/authentication/authenticator.py
ajmaddaford: I agree with what you're saying, just maybe not on the `census-rehearsal` branch.

README.md
```diff
@@ -342,7 +342,6 @@ The following env variables can be used
 | Variable Name                             | Default               | Description                                                                                   |
 |-------------------------------------------|-----------------------|-----------------------------------------------------------------------------------------------|
 | EQ_LOG_LEVEL                              | INFO                  | The default logging level (defaults to 'INFO' for local development)                          |
```
ajmaddaford: `EQ_LOG_LEVEL` also needs to go

app/authentication/authenticator.py
JamesGardiner: Does this refer to all the debug -> info changes or just the ones in this module?

ajmaddaford: All the debug -> info changes.

ajmaddaford:
I think we need to review the log levels used for 4xx responses as they are being logged as errors. We should change them to warning.

tests/functional/generate_pages.py
```diff
@@ -257,11 +257,9 @@ def generate_pascal_case_from_id(id_str):
     return name
```
MebinAbraham, line 257: How come this file changed in this PR?

JamesGardiner: Because I removed the debug line after discussing with @ajmaddaford - the other changes are small improvements

MebinAbraham (changes requested):
In production mode, the errors are logging as `info` for me. I tried 401 and 403, both are coming back as log level info.

Eg:
```json
{
  "request_id": "4013f1af-6f39-487c-86d2-1f73d7806b91",
  "status_code": 403,
  "session_modified": true,
  "event": "response",
  "level": "info",
  "created": "2019-10-16T11:13:16.459648Z",
  "service": "eq-survey-runner"
}
```

MebinAbraham (approved):

ajmaddaford (approved):

PR #2437, opened 2019-10-10
Allow POST requests to the /session endpoint

MebinAbraham:
This change allows `POST` requests to the `/session` endpoint to be treated the same as the existing `GET` request. Since go-live, there have been approx 60 eQ launches where the RH 302 request results in Chrome browsers making a POST request to the `/session` endpoint rather than a GET request. The root cause is still unclear but the intermediary fix is to allow `POST` to the session. The change should behave the same way as `GET`request since the user is still required to provide a genuine JWT token to authenticate successfully.

For our use case, we would be better suited to use 303 redirects instead of 302s for both RH and eQ, however, this change is only feasible post-rehearsal.

### How to review 
- Make `POST` requests to the `session` endpoint and ensure it works as expected. Test with both a valid and an invalid token.

### DIFF
--- a/tests/integration/routes/test_errors.py
+++ b/tests/integration/routes/test_errors.py
@@ -40,13 +40,8 @@ class TestErrors(IntegrationTestCase):
 
     def test_errors_500(self):
         # Given
-        payload_without_account_service_url = self.example_payload.copy()
-        del payload_without_account_service_url['account_service_url']
-        with patch(
-            'tests.integration.create_token.PAYLOAD',
-            payload_without_account_service_url,
-        ):
-            self.launchSurvey('test_percentage')
+        self.launchSurvey('test_percentage')
+
         # When / Then
         # Patch out a class in post to raise an exception so that the application error handler
         # gets called
--- a/tests/integration/session/test_login.py
+++ b/tests/integration/session/test_login.py
@@ -7,13 +7,13 @@ from tests.integration.create_token import PAYLOAD
 from tests.integration.integration_test_case import IntegrationTestCase
 
 
-class TestLogin(IntegrationTestCase):
+class TestLoginWithGetRequest(IntegrationTestCase):
     def test_login_with_no_token_should_be_unauthorized(self):
         # Given
         token = ''
 
         # When
-        self.get('/session?token=' + token)
+        self.get(url=f'/session?token={token}')
 
         # Then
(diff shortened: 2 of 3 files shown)


MebinAbraham:
[Understand POST to /session as method not allowed](https://trello.com/c/DN10djio/3394-understand-post-to-session-as-method-not-allowed)

tests/integration/routes/test_errors.py
```diff
@@ -40,13 +40,8 @@ def test_errors_404_with_payload(self):
...
     def test_errors_500(self):
         # Given
-        payload_without_account_service_url = self.example_payload.copy()
-        del payload_without_account_service_url['account_service_url']
-        with patch(
-            'tests.integration.create_token.PAYLOAD',
-            payload_without_account_service_url,
-        ):
-            self.launchSurvey('test_percentage')
+        self.launchSurvey('test_percentage')
```
LJBabbage, line 43: Is this part of this story? Just seems more like an unrelated tidy up unless I am wrong

tests/integration/session/test_login.py
```diff
@@ -149,7 +149,168 @@ def test_login_without_questionnaire_id_in_token_is_unauthorised(self):
...
+        # Given
+        token = self.token_generator.create_token('')
+
+        # When
+        self.post(url=f'/session?token={token}')
+
+        # Then
+        self.assertStatusForbidden()
+
+    def test_http_head_request_to_login_returns_successfully_and_get_still_works(self):
```
LJBabbage: How do you know get still works?

LJBabbage (changes requested):
Thorough tests, just a little bit of reworking unrelated to this story which perhaps shouldn't be in this PR

tests/integration/routes/test_errors.py
MebinAbraham: This test wasn't really written the way it needed to be and caused it to fail since my change. I have fixed it.

tests/integration/session/test_login.py
MebinAbraham: This should be `post`, have changed.

JamesGardiner (approved):

JamesGardiner:
I won't block on this point, but I think some of these tests could be parameterized

LJBabbage (approved):

PR #2441, opened 2019-10-14
Add ecommerce JSON for v3

LJBabbage:
As it says in the title, this is the JSON v3 for the e-commerce survey. It was produced by the converter(https://github.com/ONSdigital/eq-json-version-converter) so any adjustments will be made there and the JSON will then be reloaded  back into this PR. There are a few known issues which at present the converter can't amend or doesn't currently know how to

1. Odd ids in JSON.
2. trad_as_or_runame returning.
3. Primary content title missing on migration.

### How to review 
Launch Survey Runner and make sure the e-commerce JSON works, complete the survey and record any issues or missing features


ajmaddaford:
The placeholder for `trad_as_or_runame` should use the `first_non_empty_item` transform (see test_confirmation_question.json):

```
 {
    "placeholder": "company_name",
    "transforms": [
        {
            "transform": "first_non_empty_item",
            "arguments": {
                "items": {
                    "source": "metadata",
                    "identifier": ["trad_as", "ru_name"]
                }
            }
        }
    ]
}
```

LJBabbage (commented):
I guessing we could use the \_or\_ as a string splitter and update the metadata and the placeholder in the converter

JamesGardiner (approved):
Green based on Andrew's observation being addressed

MebinAbraham (commented):
So as disscused we need to look at merging this into a different branch?. Currently, em tags are specific to CCS:
Screenshot 2019-10-16 at 14 13 01


Also, not sure if we should change content, but there are a bunch of dumb quotes we should change?.

Apart from these, it works so.

ajmaddaford (approved):

PR #2446, opened 2019-10-17
Add autocomplete

rmccar:
This PR is to add autocomplete to input fields to comply with DAC

### How to review 
Fill in a questionnaire and check that autocomplete works for fields like name and address for example (you may need to turn on autocomplete settings in browser)


rmccar:
[Issue ID: DAC-Input-purpose](https://trello.com/c/XW4J2BK4/2960-issue-id-dac-input-purpose)

ajmaddaford (changes requested):
I'm not sure what the aim of this is, as the behaviour feels odd. All fields of the same property e.g. first name, are captured under the same autocomplete identifier. This means that as you add people you keep getting autocomplete suggestions with the names of the previous people. It's worse when you get to visitors and you get autocomplete suggestions with the householder names in it.

rmccar:
@ajmaddaford I understand your concerns. The way I saw it, it was either that or you just have autocomplete on fields affecting the primary person, which doesn't add that much value I was thinking

JamesGardiner:
While the change seems to be ok I agree with @ajmaddaford on this one - the autocomplete worked but made it less clear what needed selecting on the form. Have we done any user research on this?

PR #2447, opened 2019-10-17
Button improvements including reducing double submissions

rmccar:
This PR is to reduce the issue with double submissions. The change is to upgrade the design system to the version that will disable a submit button for a short period on submission of the form. This will then reduce the issue as double clicks by users wont now lead to two submissions.

It also amends the runner code to allow for the breaking change in version 13.8.0 of the design system around contextSuffix

### How to review 
Check that the submit button on the hub changes to the loading style and is disabled on click


rmccar:
[Temporarily disable submit button to guard against double clicks](https://trello.com/c/CvzQKCMM/3446-temporarily-disable-submit-button-to-guard-against-double-clicks)

ajmaddaford (changes requested):
Updating to the latest design system will break the interviewer instruction styling in CCS. This can't merge until the work is done to fix that.

rmccar:
The change that introduced the interviewer instruction styling was removed in v13.7.0 in this PR [#626](https://github.com/ONSdigital/design-system/pull/626). So it is back to how it was originally with it just being bold when an em tag is applied. The work to replace it with something else hasn't been done yet

ajmaddaford:
@rmccar yes, that's why this can't merge - it will break the expected styling for ccs.

templates/partials/question-definition.html
```diff
@@ -12,11 +12,12 @@
       },
       "button": {
         "close": _("Hide this"),
+        "contextSuffix": _("content"),
```
ajmaddaford: - This doesn't give enough context for a translator to translate. We can't do string concatenation in this way if it needs to be translated.
- This isn't related to this pull request, another pr should be raised for these changes.

ajmaddaford (changes requested):
The submit button changes look good. I think we need to move the unrelated guidance button changes out of this pr so we can get it merged. 
The pr title needs to change as this won't 'fix' the double submission issue. Something like 'Reduce double submission issues' would be more accurate.

templates/partials/question-definition.html
rmccar: As agreed I'll leave this in as it is needed for a breaking change to the design system but it'll be no longer translated. Then we will fix this in a further change to the design system.

ajmaddaford (approved):

MebinAbraham (approved):

PR #2461, opened 2019-10-22
Switch from persistent to session cookies

SamGodwin2:
After investigation into production issues it has been identified
that some browsers (predominantly Edge) seem to drop our session
cookie randomly. Although we still don't know why this happens
(debugging has shown that a valid cookie is sent on the request
previous to the one that errors), changing from a permanent cookie
(one with an expiry time set) to a session cookie (one that expires
when the browser is closed) seems to fix the issue.

Arguably the cookie should always have been a session one. The
server side session times out after the timeout period (default is
45 minutes) and keeping the cookie client side means we will still
have context like the theme or an account service url that are
useful when rendering error screens. The cookie was set to expire
prior to us having server side session expiry.

Solution based on: https://github.com/ONSdigital/eq-survey-runner/pull/2420

### DIFF
--- a/app/setup.py
+++ b/app/setup.py
@@ -3,14 +3,13 @@ import json
 import logging
 import os
 import sys
-from datetime import timedelta
 from uuid import uuid4
 
 import boto3
 import sqlalchemy
 import yaml
 from botocore.config import Config
-from flask import Flask, url_for, session as cookie_session
+from flask import Flask, url_for
 from flask_babel import Babel
 from flask_caching import Cache
 from flask_talisman import Talisman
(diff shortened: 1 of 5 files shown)


MattElcock (approved):

bethanbolton:
[Census session timeout fix](https://trello.com/c/IgZackos/1366-census-session-timeout-fix)

martyncolmer:
Need to removed the unused timedelta import from the test for the travis pylint tests to pass. @SamGodwin2

martyncolmer (approved):
Looks good. ok to merge.

MebinAbraham (approved):

PR #2462, opened 2019-10-23
Fix CCS translations

petechd:
CCS schema was not included in translation map in translate_schemas script and was only displaying static Welsh content. To make it use all Welsh translations from .pot file minor fix was added.

### How to review 
Run `make build`
It should translate 86% of ccs schema content.
Start launcher and choose "ccs_household_gb_eng" schema and Welsh language (don't change the region code).
Check if language toggle button is visible in top right corner.


berroar (approved):

ajmaddaford (changes requested):
The language switcher is missing when in runner (top right of the page). The CCS schema name needs to be added to `LANGUAGES_MAP` in `app/utilities/schema.py` to enable this.

ajmaddaford (approved):

PR #2463, opened 2019-10-24
Update functional tests to use latest webdriver

iwootten:

Our functional tests are unable to run in parallel due to the version of webdriver we're using and we cannot make use of synchronous style that has been added to more recent versions of it. This results in confusing call chains in each of our functional tests, which can't be run in parallel.

This PR drops use of chimp and replaces with webdriver. It allows us to write tests in a simpler synchronous style, which webdriver will convert and run as async.

### How to review 

Check the functional tests continue to pass. See if anything obvious stands out in the tests which have been converted.


README.md
```diff
@@ -224,13 +222,13 @@ If you have already built the generated pages, then the functional tests can be
...
+`./node_modules/.bin/wdio tests/functional/wdio.conf.js --spec tests/functional/spec/save_sign_out.spec.js`
 
-To run a single test, add `@watch` into the name of any `describe` or `it` function and run:
+To run a single test, add `.only` into the name of any `describe` or `it` function and run:
 
-`yarn test_functional --watch`
+`yarn test_functional`
 
 An example of adding `@watch` looks like this:
 `describe('@watch Skip Conditions', function() {...}` or
```
ajmaddaford: The `@watch` example needs to be reworded into a `.only` example or removed.

The Cypress mention below also needs to be removed.

ajmaddaford (changes requested):
Changes look good, just some minor Readme updates needed.

ajmaddaford (approved):

MebinAbraham (changes requested):
Just need to update some old references:
https://github.com/ONSdigital/eq-survey-runner/blob/update-functional-webdriver/tests/functional/generate_pages.py#L55-L67

MebinAbraham (commented):
I seem to get a bunch of errors. Have done `yarn`, `make build` and `make test-functional`. Have I missed something?.
Eg:
```
[0-4] Error in "Checkbox with a numeric "detail_answer" option "before each" hook for "Given a numeric detail answer options are available, When the user clicks an option, Then the detail answer input should be visible.""
Can't call click on element with selector "#checkbox-numeric-detail-answer-3" because element wasn't found
```

Also, I get the following warning:
```
[1573130738.262][WARNING]: This version of ChromeDriver has not been tested with Chrome version 78.
```
I can't see the chrome version being defined anywhere?

tests/functional/wdio.conf.js
```diff
@@ -0,0 +1,270 @@
...
+     * @param {Array.<String>} specs List of spec file paths that are to be run
+     */
+    before: function (capabilities, specs) {
+        const chai = require('chai');
+        const JwtHelper = require('./jwt_helper');
+
+        global.expect = chai.expect;
+
+        browser.addCommand('openQuestionnaire', async function (schema, { userId = JwtHelper.getRandomString(10), collectionId = JwtHelper.getRandomString(10), responseId = JwtHelper.getRandomString(16), periodId = '201605', periodStr = 'May 2016', region = 'GB-ENG', language = 'en', sexualIdentity = false, includeLogoutUrl = false } = {}){
+          token = await JwtHelper.generateToken(schema, {userId, collectionId, responseId: responseId, periodId: periodId, periodStr: periodStr, regionCode: region, languageCode: language, sexualIdentity: sexualIdentity, includeLogoutUrl: includeLogoutUrl})
```
MebinAbraham:
```suggestion
          const token = await JwtHelper.generateToken(schema, {userId, collectionId, responseId: responseId, periodId: periodId, periodStr: periodStr, regionCode: region, languageCode: language, sexualIdentity: sexualIdentity, includeLogoutUrl: includeLogoutUrl})
```

package-lock.json
```diff
@@ -0,0 +1,5775 @@
+{
```
MebinAbraham: Why has a package-lock been introduced?. We use yarn?

iwootten: Looks like I introduced it in the first commit accidentally.

MebinAbraham (approved):
👍

PR #2477, opened 2019-11-06
Remove dumb single and double quotes

berroar:
This PR fixes all dumb single and double quotes that have been identified in our all of our current schemas by the new method added to the schema validator which searches for dumb quotes.

We need to make this change before we add the new method to validator so that our build will not fail(see the validator PR [here](https://github.com/ONSdigital/eq-schema-validator/pull/166).

### How to review 
Check that there are no remaining dumb single or double quotes in our schemas.
Check that the changes made look OK.


berroar:
[Inconsistent quotes used in Welsh translations](https://trello.com/c/6LEeu41c/3228-inconsistent-quotes-used-in-welsh-translations)

JamesGardiner (approved):

ajmaddaford (approved):

PR #2435, opened 2019-10-10
Remove dumb quotes from the census schema

berroar:
### What is the context of this PR?
Removing dumb quotes in the census schema.

### How to review 
Build and check that dumb quotes have been removed from the census schemas. We should now only be using smart quotes.

PR #2478, opened 2019-11-06
Revert Quote Change

berroar:
This PR has been raised to revert a change from a dumb to a smart quote in an answer value. This was changed happened as part of [this PR](https://github.com/ONSdigital/eq-survey-runner/pull/2435).

We need to revert this change as we do not want to change any answer values while the rehearsal is still in flight.

### How to review 
Check that the change from the previous PR has been reverted.


berroar:
[Revert answer value change](https://trello.com/c/hxjOtvkC/3499-revert-answer-value-change)

data-source/jsonnet/england-wales/household/blocks/who-lives-here/primary_person_list_collector.jsonnet
```diff
@@ -50,7 +50,7 @@ local questionTitle = {
     title: questionTitle,
     definitions: [
       {
-        title: 'What does "usually live" mean?',
+        title: 'What does “usually live” mean?',
```
iwootten: How come these titles are changing, I thought this should only be changing values?

ajmaddaford (approved):

data-source/jsonnet/england-wales/household/blocks/who-lives-here/primary_person_list_collector.jsonnet
berroar: The build was failing because this line was different in the .po files. I'm not sure how or why these got out of sync because I hadn't changed this text. So I ended up having to make this change to allow the build to pass.

iwootten: I think that changing these titles and updating the .po files would be what we'd really want? The only thing changing in this PR should really be that single answer value.

berroar: I will close this PR and open another with a clean branch.

berroar:
Opened https://github.com/ONSdigital/eq-survey-runner/pull/2482 as the first commit in this branch was corrupt.

PR #2479, opened 2019-11-06
Optional account_service_url in templates

petechd:
This adds _account_service_url_ link as an option in business surveys, when available. New "Return to your account" link is displayed on two static pages: **signed-out.html** and **session-expired.html**. In the latter, there are multiple conditionals based on schema theme and value of the variable.

### How to review 
To check the changes it's best to launch both versions of census and business schemas ("census-nisra"/"northernireland" and "census"/"default") and manually visit pages( e.g [session expired](http://localhost:5000/session-expired)). Different content will be displayed while session data is not flushed.


petechd:
[Make account_service_url metadata optional (S)](https://trello.com/c/xiwXMbzb/3480-make-accountserviceurl-metadata-optional-s)

PR #2480, opened 2019-11-06
Fix playback confirmation

iwootten:

Answers which have lists as their value are being incorrectly escaped. This causes problems when trying to use placeholder transformations for checkbox answers and each character of the list content is escaped.

### How to review 

Use the new test_placeholder_playback_list.json and select some toppings 🍕. Confirm they are correctly formatted on the confirmation page.


iwootten:
[Fix playback confirmation feature](https://trello.com/c/nlmgzWE4/3487-fix-playback-confirmation-feature)

ajmaddaford:
I think we should have a functional test to demonstrate the feature. Probably worth waiting for #2463 to merge first.

app/questionnaire/placeholder_parser.py
```diff
@@ -40,9 +40,11 @@ def __call__(self, placeholder_list: Sequence[Mapping]) -> Mapping:
 
     def _lookup_answer(
         self, answer_id: str, list_item_id: str = None
-    ) -> Union[Markup, None]:
+    ) -> Union[Markup, Sequence[Markup], None]:
         answer = self._answer_store.get_answer(answer_id, list_item_id)
         if answer:
+            if isinstance(answer.value, list):
```
JamesGardiner, line 46: The use of isinstance is generally seen as a bad code smell or data choice. For example, in this example, if the value was always an iterable, the isinstance could be omitted. Instead we now have multiple return types for the placeholder_parser as well. This isn't a blocking issue for this PR, but it's worth mentioning as it is tech debt that could be refined.

JamesGardiner (approved):

JamesGardiner:
Approved based on the functional test requested by @ajmaddaford being added

app/questionnaire/placeholder_parser.py
iwootten: I'd considered the alternatives, but the fact is the value can be multiple types.

Refactoring the answerstore in this PR certainly seems overkill given that we frequently use isinstance elsewhere.

JamesGardiner: Yes, definitely overkill in this instance - are we tracking this kind of tech debt anywhere @ajmaddaford ?

MebinAbraham:
👍

app/questionnaire/placeholder_parser.py
ajmaddaford: Tech debt is tracked by raising a card with a 'Tech Debt' label. I don't think we have one for this so please add one. Ensure you add an explanation of why it's an issue and if possible a proposed fix.

ajmaddaford (approved):

PR #2487, opened 2019-11-11
Fix broken functional test

berroar:
A functional test has begun falling over as the string being asserted for is out of date (it references a date of birth which is now being calculated differently).

I have amended the test so that it no longer checks for the age within the string being matched.

### How to review 
Run and check that all functional tests are passing.

### DIFF
--- a/tests/functional/spec/features/repeating_sections/repeating_sections_with_hub_and_spoke.js
+++ b/tests/functional/spec/features/repeating_sections/repeating_sections_with_hub_and_spoke.js
@@ -252,7 +252,7 @@ describe('Feature: Repeating Sections with Hub and Spoke', function () {
         .setValue(DateOfBirthPage.year(), '1990')
         .click(DateOfBirthPage.submit())
 
-        .getText(ConfirmDateOfBirthPage.questionText()).should.eventually.equal('Samuel Clemens is 28 years old. Is this correct?')
+        .getText(ConfirmDateOfBirthPage.questionText()).should.eventually.contain('Samuel Clemens is')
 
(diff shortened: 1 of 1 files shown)


berroar:
[Fix broken functional test in runner](https://trello.com/c/7RXPeB0k/3514-fix-broken-functional-test-in-runner)

LJBabbage (approved):
Happy Birthday Samuel for today (11th November), you're now 29, oh crap that won't work. You could wildcard the age to make it a better test.

tests/functional/spec/features/repeating_sections/repeating_sections_with_hub_and_spoke.js
```diff
@@ -252,7 +252,7 @@ describe('Feature: Repeating Sections with Hub and Spoke', function () {
         .setValue(DateOfBirthPage.year(), '1990')
         .click(DateOfBirthPage.submit())
 
-        .getText(ConfirmDateOfBirthPage.questionText()).should.eventually.equal('Samuel Clemens is 28 years old. Is this correct?')
+        .getText(ConfirmDateOfBirthPage.questionText()).should.eventually.contain('Samuel Clemens is')
```
iwootten: I think this line should be removed completely - We shouldn't care what the text is midway through a test and the test will fail later if "Samuel's" name isn't present.

iwootten (changes requested):

tests/functional/spec/features/repeating_sections/repeating_sections_with_hub_and_spoke.js
berroar: I have removed this line.

iwootten (approved):

PR #2493, opened 2019-11-15
3521 fix flush feature

pricem14pc:
The /flush feature fails as the encryption is expecting a string representation of the message rather than a dict. This change aligns the flusher with submit_answers to use json.dumps to convert the dict to a string of json prior to encryption.

This doesn't get caught by the tests as the encryption is returned by a MagicMock implementation.

### DIFF
--- a/app/routes/flush.py
+++ b/app/routes/flush.py
@@ -1,3 +1,4 @@
+import simplejson as json
 from flask import Blueprint, Response, request, session, current_app
 from sdc.crypto.encrypter import encrypt
 from sdc.crypto.decrypter import decrypt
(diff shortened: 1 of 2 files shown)


pricem14pc:
[Fix /flush feature](https://trello.com/c/1qxZpTcA/3521-fix-flush-feature)

JamesGardiner (approved):

ajmaddaford (commented):
The changes look good. I think we should consider the logging. Currently there is no way to correlate a flush request someone makes with log lines. If we required `tx_id` for flush requests we could log it. We should also log when we successfully send a message.

ajmaddaford (approved):
As this is needed imminently I've raised a card for improving the logging.

PR #2494, opened 2019-11-25
Use set comprehension and flatten nested structures in questionnaire schema

JamesGardiner:
Use set comprehension in questionnaire schema where appropriate. Flatten deeply nested for loops.

### How to review 
Check all tests pass and you're happy with the approach.


JamesGardiner:
Opened against wrong repo

PR #2505, opened 2019-12-03
Fixed Decimal Rounding

martyncolmer:
Altered WTF Form config to display correct number of decimal places.


MattElcock (approved):
Looks good to me; works locally and fix makes sense.

SamGodwin2 (approved):

PR #2509, opened 2019-12-09
Fixed some Radio text in e-commerce

SamGodwin2:
Changed some radio text in both e-commerce surveys to bring them inline with sdx


martyncolmer (approved):
Looks good

heofs (approved):

PR #2515, opened 2020-01-24
Removed automatically adding language code query param on getting sch…

MattElcock:
…ema by url

The ability to get a schema by url automatically loads in the `?language=en` query parameter; this should not be the case when we add the ability to get a schema from the survey registry, as we use query parameters to do this. Ideally, we want to be able to just send Runner a URL and it will load in the schema from there.

Spoke to Torrence and to his knowledge RASRM don't use language codes

### How to review 
Ensure tests pass.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,33 +1,31 @@
 dist: xenial
 sudo: true
 addons:
-  apt:
-    sources:
-      - google-chrome
-    packages:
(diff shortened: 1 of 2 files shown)


heofs (approved):
LGTM!

zimboflyman (approved):
Coolio Matt!

PR #2551, opened 2020-04-06
added construction 001

martyncolmer:
Added the construction 0001 survey

### DIFF
--- a/data/en/construction_0001.json
+++ b/data/en/construction_0001.json
@@ -1,5 +1,5 @@
 {
-    "eq_id": "construction_0002",
+    "eq_id": "construction_0001",
     "form_type": "0001",
     "mime_type": "application/json/ons/eq",
     "schema_version": "0.0.1",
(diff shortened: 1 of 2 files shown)


jasehumphr (approved):

data/en/construction_0001.json
```diff
@@ -1117,7 +1088,7 @@
                             },
                             {
                                 "goto": {
-                                    "group": "group683013fe-88a1-47c3-975c-fbf36c3ec625"
+                                    "group": "groupe9d26530-ad8f-4527-b09c-afacfa0c820a"
```
MattElcock, line 1091: Should this be spelled `groupe`? Or is that part of the ID? Not sure if they always start with a number or not.

MattElcock (changes requested):

data/en/construction_0001.json
jasehumphr: Can start with a number or a letter. If it has to start with a number it will be less random

MattElcock (approved):

data/en/construction_0001.json
martyncolmer: It's part of the ID

PR #2632, opened 2020-09-03
Conditionally render and set cookies/cookiesbanner

tjamesmac:

Adds the option to accept all cookies and select individual cookies of the users choice. 

A cookie banner has been added to the top *any* page and is removed once a user accepts all cookies or modifies their preferences.

To modify their preferences a `/cookies-settings` page has been added which only gives the option to remove google analytics at this point. From what I can tell, the other options provided by the ONS cookie JS are relevant to `eq-survey-runner`. However I could be wrong.

At this point 99% of the functionality is complete with the supporting tests, the remaining work to be done is content and quality of life checks. Most of the templates have been taken from elsewhere in the ONS and some of the content doesn't necessarily reflect that.

One content issue I have noticed so far is that once the user has selected the cookies they want to allow. There is no way for them to move off of the settings page. At the moment I've left a placeholder for a link but not quite sure where to direct the link to.

The only major concern I have with this work is I have removed the google analytics scripts being added to the templates when the user has chosen to do so. However, there is an `analytics.js` script being brought in by the ONS cdn and I don't think I can modify it? There is a local version of the same file within runner but it doesn't seem to be used.

One minor concern and I'm not actually sure how to fix yet is cookies only being enabled on moving to a new page or after a refresh. This means after accepting all cookies, the analytics scripts will only be pulled in on moving to the next page.

### How to review 
This might take a while
#### Test one
1. Getting runner running locally (using docker provides the GA variables)
2. Launch a questionnaire and check that the cookie banner is present (check several different types of survey)
3. Open the chrome dev tools > application > cookies
4. Click the accept all cookies button and notice the `ons_cookies_message_displayed` cookie get added
5. Take a note of the value of the cookie policy
6. The banner should be replaced with another message with a hide able button
7. Hit refresh
8. The banner should now be gone
9. Clear the cookies

#### Test two
1. Launch a survey
2. Click the Set cookie preferences button
3. Read through the content to make sure if makes sense
4. The page is different from the ONS design system one because as far as I can tell we don't have that many cookies
5. Check the cookies
6. Press yes to the GA cookie and submit
7. Cookie confirmation banner should appear (need to decide where to send user after this)
8. Refresh the page and the options should represent the selected values

Also need to run the `js` integration tests to make sure they pass. They don't come up in the PR.

To run them use `./scripts/run_tests_functional.sh`

I won't be in the rest of the week but if there are any questions during reviews on this, message me on slack or leave a comment in the PR and I'll respond whilst I can.


app/assets/js/app/modules/cookies/cookiesfunctions.js
```diff
@@ -0,0 +1,180 @@
...
+  }
+}
+
+export function setDefaultConsentCookie() {
+  const defaultConsentCookie = JSON.stringify(DEFAULT_COOKIE_CONSENT).replace(/"/g, '"')
+  setCookie('ons_cookie_policy', defaultConsentCookie, { days: 365 })
+}
+
+export function approveAllCookieTypes() {
+  let approvedConsent = {
```
MattElcock, line 45:
```suggestion
  const approvedConsent = {
```

app/assets/js/app/modules/cookies/cookiesfunctions.js
```diff
@@ -0,0 +1,180 @@
...
+    cookieConsent = JSON.parse(JSON.stringify(DEFAULT_COOKIE_CONSENT).replace(/'/g, '"'))
+  }
+  for (let cookieType in options) {
+    cookieConsent[cookieType] = options[cookieType]
+    if (!options[cookieType]) {
+      for (let cookies in COOKIE_CATEGORIES) {
+        if (COOKIE_CATEGORIES[cookies] === cookieType) {
+          cookie(cookies, null)
+          if (cookie(cookies)) {
+            const cookieString = cookies + '=; expires=' + new Date() + '; domain=' + domain + '; path=/'
```
MattElcock, line 83:
```suggestion
            const cookieString = `${cookies}=; expires=${new Date()}; domain=${domain}; path=/`
```

app/assets/js/app/modules/cookies/cookiesfunctions.js
```diff
@@ -0,0 +1,180 @@
...
+            document.cookie = cookieString
+            // eslint-disable-next-line no-console
+            console.log('remove cookie:', cookieString)
```
MattElcock:
```suggestion

```

app/assets/js/app/modules/cookies/cookiesfunctions.js
```diff
@@ -0,0 +1,180 @@
...
+  if (!currentConsentCookie && COOKIE_CATEGORIES[cookieName]) {
+    return true
+  }
+
+  currentConsentCookie = getConsentCookie()
+  try {
+    return currentConsentCookie[cookieCategory]
+  } catch (e) {
+    // eslint-disable-next-line no-console
+    console.error(e)
```
MattElcock:
```suggestion
```

app/assets/js/app/modules/cookies/cookiesfunctions.js
```diff
@@ -0,0 +1,180 @@
...
+  if (domain.indexOf('localhost') === -1) {
+    setDomain = '; domain=' + domain
+  }
+
+  if (checkConsentCookie(name, value)) {
+    if (typeof options === 'undefined') {
+      options = {}
+    }
+
+    let cookieString = name + '=' + value + setDomain + '; path=/'
```
MattElcock, line 135: Could swap this to be a template string.

app/assets/js/app/modules/cookies/cookiesfunctions.js
```diff
@@ -0,0 +1,180 @@
...
+    if (options.days) {
+      const date = new Date()
+      date.setTime(date.getTime() + options.days * 24 * 60 * 60 * 1000)
+      cookieString = cookieString + '; expires=' + date.toGMTString()
```
MattElcock, line 139: Could swap this to be a template string.

app/assets/js/app/modules/cookies/cookiesfunctions.js
```diff
@@ -0,0 +1,180 @@
...
+
+export function getDomain(domain) {
+  let i = 0
+  let domainName = domain
+  let p = domainName.split('.')
+  let s = '_gd' + new Date().getTime()
+
+  while (i < p.length - 1 && document.cookie.indexOf(s + '=' + s) === -1) {
+    domainName = p.slice(-1 - ++i).join('.')
+    document.cookie = s + '=' + s + ';domain=' + domainName + ';'
```
MattElcock, line 172: Could swap this to be a template string.

app/assets/js/app/modules/cookies/cookiesfunctions.js
```diff
@@ -0,0 +1,180 @@
...
+  }
+  document.cookie = s + '=;expires=Thu, 01 Jan 1970 00:00:01 GMT;domain=' + domainName + ';'
```
MattElcock, line 174: Could swap this to be a template string.

app/assets/js/app/modules/cookies/cookiessettings.js
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    const currentConsentCookie = cookie('ons_cookie_policy')
+    let currentConsentCookieJSON = JSON.parse(currentConsentCookie.replace(/'/g, '"'))
+
+    try {
+      unset(currentConsentCookieJSON, 'essential')
+    } catch (e) {
+      // eslint-disable-next-line no-console
+      console.error(e)
```
MattElcock:
```suggestion
```

MattElcock (changes requested):
Hi Tom,

Functionally, it looks like there's some problems but it could just be the way Martyn and I are testing. Could you go through this with us when you're back in, please?

As for the code, it all looks pretty neat and tidy. There's a few refactoring things that I've noticed could be done, but nothing major. I've put some notes on those below. There's a bit about logging, too. We should have a chat about this, too, when you're back in, to decide on a way forward.

Reg. the code:

- You can use template strings on Runner's version of Node; I didn't think you could either but I tried it in my console on version 8 and it works.
- You're right to be logging things out, especially the errors. However, `console.*` should ideally only be used whilst deving as the output isn't amazing in production environments. Dedicated loggers are usually better. I'll hit you up about this when you're next in to go over how we do it in Author.

Reg. the feature:

When you're back in can we go through this with you? A lot seems to be broken so I'm not sure if we're doing something wrong.

- When I click 'Accept all cookies', visually nothing happens.
- On the preference page, my selection doesn't stay when I navigate away and come back, or after clicking an option and the page refreshes.
- The buttons in the banner are different sizes
- I don't think the padding or margins on the banner are quite right as the title text is really close to the top of the screen.
- The buttons on the preferences page are really spaced out
- The buttons on the preference page don't have the borders around it

Here's a screenshot of how it looks my end.

image

MattElcock:
Spoken with Tom and Martyn, managed to get the visual stuff sorted so that's all good. Agreed that the `console.*` statements will just be removed for now. Template literals can be ignored also.

MattElcock (approved):

farres1 (approved):

PR #2639, opened 2020-09-21
Changed leading 9 to 0 on MBS files

MattElcock:

From the JIRA ticket:

> Currently the MBS JSON URLs are listed with a 9 at the start of the form type e.g. mbs_*9*255. We need to change these to be 0 at the start e.g. mbs_*0*255. There are existing surveys with these IDs so they need to be deleted first before the "9" file is renamed.

### How to review 
Make sure the above has been done and that no files with a leading 0 have been deleted.

### Checklist


jasehumphr (approved):
VAN TICK

martyncolmer (approved):

PR #2696, opened 2020-11-06
Revert "EAR-999 MESS 999 or Less"

MattElcock:
Reverts ONSdigital/eq-survey-runner#2695


jasehumphr (approved):
So upset that wording was so much better as well

tjamesmac (approved):

PR #2706, opened 2020-11-25
Change to single quotes

tjamesmac:
### Talk to Andy Pitt before deploying to live so he can test the changes.

Quotes are killing google analytics in live

Live surveys don't have an `ga` working right now because the cookies `HTML` isn't getting pulled in.

We are checking for the presence of double-quotes to pull in `ga` where RAS are using single quotes

This means that in live surveys, if a user accepts cookies in RAS, `ga` isn't getting pulled in. Where as, if cookies are accepted within a survey, `ga` does get pulled in.

We need a solution to take into account both cases.

Sort of a big deal because rolling back breaks GDPR compliance and not doing anything means there is no analytics on any surveys.

#### Tom B edit 26/11, changes:

- Change our JS to emit single quotes in case downstream are expecting single to come back (since they initially set single-quote cookies)
- Refactor the cookie JS and python slightly so the quote-setting logic is just in one function in case we want to change it again
- Change runner so it's happy with either double or single quotes - it now converts to & parses the JSON for the cookie

### DIFF
--- a/app/assets/js/app/modules/cookies/cookiesbanner.js
+++ b/app/assets/js/app/modules/cookies/cookiesbanner.js
@@ -1,4 +1,4 @@
-import { cookie, setDefaultConsentCookie, approveAllCookieTypes, setConsentCookie } from './cookiesfunctions'
+import { cookie, setDefaultConsentCookie, approveAllCookieTypes, setConsentCookie, parseCookie } from './cookiesfunctions'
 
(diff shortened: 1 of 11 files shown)


zimboflyman (approved):
looks l;ike you've covered all the usage: true spots!

jasehumphr (approved):

PR #2712, opened 2020-12-14
EAR-1045 Implement UKIS_0001 and UKIS_0002 changes

farres1:
Updated UKIS content. ukis_9001 and ukis_9002 are temporary files that include the new content. ukis_0001 and ukis_0002 include the previous 2016-2018 versions of UKIS.

### How to review 
Check that the ukis_0001 and ukis_0002 surveys work as expected, and are not different from the previous versions. Check that ukis_9001 and ukis_9002 work as intended.

### Checklist

* [ ] Check that ukis_0001 is working as intended and the same as it was in previous versions
* [ ] Check that ukis_0002 is working as intended and the same as it was in previous versions
* [ ] Open ukis_9001
* [ ] Check that this survey is working as intended
* [ ] Open ukis_9002
* [ ] Check that this survey is working as intended
* [ ] Check that the values from the Jira ticket are correct in the ukis_9001 and ukis_9002 JSON files

### DIFF
--- a/data/en/ukis_0001.json
+++ b/data/en/ukis_0001.json
@@ -1,204 +1,84 @@
 {
-    "eq_id": "ukis",
+    "eq_id": "ukis_0001",
     "form_type": "0001",
     "mime_type": "application/json/ons/eq",
     "schema_version": "0.0.1",
     "data_version": "0.0.1",
     "survey_id": "144",
-    "title": "UK Innovation Survey 2016 - 2018",
-    "theme": "ukis",
-    "navigation": {
-        "visible": true
-    },
-    "view_submitted_response": {
-        "enabled": true,
-        "duration": 900
-    },
-    "metadata": [{
-            "name": "user_id",
-            "validator": "string"
-        },
-        {
-            "name": "period_id",
-            "validator": "string"
-        },
-        {
-            "name": "ru_name",
-            "validator": "string"
-        },
-        {
-            "name": "trad_as",
-            "validator": "optional_string"
-        }
-    ],
+    "title": "UK Innovation Survey 2018 - 2020",
     "sections": [{
-            "id": "introduction-section",
-            "title": "Introduction",
+            "id": "sectionfb9d2910-8ac7-40c6-90bd-ecafc58580b2",
+            "title": "Business strategy and practices",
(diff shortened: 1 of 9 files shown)


data/en/ukis_0001.json
```diff
@@ -1,277 +1,147 @@
 {
-    "eq_id": "ukis",
+    "eq_id": "ukis_0001",
```
MattElcock: Hi Sam. I see why this has changed; normally you're not incorrect - the eq_id is of that format. However, to avoid unpredicted errors, if the existing eq_id is different within the schema when we go to change it we normally leave it alone. Just in case.

```suggestion
    "eq_id": "ukis",
```

data/en/ukis_0001.json
```diff
@@ -4565,129 +4206,151 @@
...
+            "name": "trad_as",
+            "validator": "optional_string"
+        },
+        {
+            "name": "ref_p_start_date",
+            "validator": "date"
+        },
+        {
+            "name": "ref_p_end_date",
+            "validator": "date"
```
MattElcock: Double checked with the business via Jordan, we don't need these two bits of metadata in here.

MattElcock (changes requested):

data/en/ukis_0002.json
```diff
@@ -4547,129 +4206,151 @@
...
+            "name": "trad_as",
+            "validator": "optional_string"
+        },
+        {
+            "name": "ref_p_start_date",
+            "validator": "date"
+        },
+        {
+            "name": "ref_p_end_date",
+            "validator": "date"
```
MattElcock: The same as above reg metadata

data/en/ukis_0002.json
```diff
@@ -1,275 +1,147 @@
 {
-    "eq_id": "ukis_ni",
-    "form_type": "0001",
+    "eq_id": "ukis_0002",
```
MattElcock: The same as above reg eq_id

MattElcock (changes requested):

MattElcock (approved):

tbaggaley (approved):

PR #2732, opened 2021-02-10
Fix datetime breaking on 0 milliseconds

tjamesmac:
This fixes the thank-you page breaking when the submitted_time has 0 milliseconds.

### How to review 
Test it works by putting the format the other way round so the one with milliseconds is the exception.

### DIFF
--- a/app/utilities/format_submitted_time.py
+++ b/app/utilities/format_submitted_time.py
@@ -1,0 +1,9 @@
+from datetime import datetime
+
+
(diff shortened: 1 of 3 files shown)


jasehumphr (approved):

martyncolmer:
You could test it works by putting the format the other way round so the one with milliseconds is the exception.

tbaggaley (approved):

PR #2764, opened 2021-05-27
EAR-1321 Add PubSub to runner

martyncolmer:
Docker compose up --build will start runner with pubsub running in a container.
You can complete a simple questionnaire and review the console messages to see the message being published.


MattElcock (approved):
Nice one; works as expected. I can see it publishing in the console.

zimboflyman (approved):

PR #2770, opened 2021-06-16
Ear 1384 construction june 2021

MattElcock:
Added new construction surveys and kept old ones by putting 9 infront of the name.

### How to review 
Ensure that the old surveys are still accessibile and the new ones are there.

Ensure that the following AC is met on 0002 form type:

GIVEN I'm on the total number of employees breakdown question
and I cannot provide the answers
THEN I can continue to the next question without answering

### Checklist


tbaggaley (approved):
All seems well with the breakdown question on 0002

martyncolmer (approved):

